PR #14133, opened 2025-06-02
[WOOMOB-503] Fix retry behavior in order creation

malinajirka:
WOOMOB-503

### Description

This PR allows users to retry the recalculate operation as many times as they need.

### Steps to reproduce

1. Open Order Creation
2. Select a product
3. Turn on airplane mode
4. Tap on Recalculate
5. Tap on Retry
6. Notice the button is disabled and the only way to recover is to start over

### Testing information

1. Open Order Creation
2. Select a product
3. Turn on airplane mode
4. Tap on Recalculate
5. Tap on Retry
6. Notice the snackbar appears again when the operation fails
7. Tap on retry
8. Notice the snackbar appears again when the operation fails
9. Turn off airplane mode
10. Tap on retry
11. Notice the operation succeeds

- The above. I also tested that if I select different products inbetween retries, the app forces me to use `Recalculate` again before I can proceed with the order.

### Images/gif

https://github.com/user-attachments/assets/a5f7ac21-1a3a-4edf-a2eb-16a026f1386c

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 22.5
 -----
 - [**] Improved local search algorithms in POS [https://github.com/woocommerce/woocommerce-android/pull/14068]
+- [*] Fix a rare dead-end in the order creation flow. [https://github.com/woocommerce/woocommerce-android/pull/14133]
 
 22.4
 -----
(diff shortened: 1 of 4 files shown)


samiuelson (approved):
LGTM! Thanks for improving that, @malinajirka.

💡 Just a thought — can we make that snackbar non-dismissable? My understanding is that currently, a user can still end up in a dead-end if they accidentally dismiss the snackbar by swiping it out.

malinajirka:
Thanks for the review! 

> 💡 Just a thought — can we make that snackbar non-dismissable? My understanding is that currently, a user can still end up in a dead-end if they accidentally dismiss the snackbar by swiping it out.

Good catch, fixed in https://github.com/woocommerce/woocommerce-android/pull/14133/commits/db26ab8261c28e02fade611805339698902fb679

samiuelson (approved):
LGTM

PR #14134, opened 2025-06-02
[WOOMOB-207] Increase Decimal Places in Order Creation Requests

malinajirka:
WOOMOB-207

### Description
Fixes order creation rounding issues by adding the dp=8 (decimal places) parameter in create and update order API requests. This increases decimal precision to reduce discrepancies between client-side calculations and backend totals, addressing cases where item totals were incorrectly rounded (e.g., $6 product with 9% tax showing $11.00 instead of $11.01 for quantity of 2).

### Steps to reproduce

1. Add a product with price that results in non-terminating decimal after tax calculation (e.g., $6 with 9% tax = $5.504587 pre-tax) (Make sure the `Enter product prices including tax` settings is enabled)
3. Increase quantity to 2 or more in cart
5. Observe total calculation discrepancy between displayed rounded values and actual precise calculations
7. Complete order creation and verify totals match between app and WooCommerce backend

### Testing information

Repeat the above steps and make sure the price is correct and matches what is displayed in wpadmin.

### Images/gif

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/f4acd11b-2568-45f3-950e-88d78422df7f" controls width="400"></video> | <video src="https://github.com/user-attachments/assets/86717a1e-0a51-44cd-b339-2b8acfaa7ff2" controls width="400"></video> |


samiuelson (approved):
Thanks for fixing that! LGTM

PR #14139, opened 2025-06-03
Issue/wooprd 552 woo posbarcodes implement scanning send the data to the cart

malinajirka:
WOOMOB-552

### Description

This PR adds a naive implementation of barcode scanner support to the POS. 

It adds a custom modifier to PosItemsScreen that listens for keyboard inputs and passes them back to the ItemsVM when the input ends with `\n or \r`. Better detection of barcode inputs will be implemented in a separate task.

The ItemsVM passes the scanned barcode to HomeVM which passes it to CartVM. The CartVM takes care of searching the item and adding it into the cart.

The implementation in the CartVM is naive too
- Works only if the product is in the cache.
- Doesn't display any loading indicator when the search is in progress

Proper implementation will be implemented as part of a different task.

---------------

I considered listening for the key inputs in the WooPosActivity instead of adding the custom modifier. I like that the modifier is closer to the state/screen when we want to enable barcode scanning. However, I'm a bit worried it might cause troubles with TalkBack - we'll see, moving it to the Activity will be pretty simple if needed.

### Testing information

1. Open POS.
2. Scan a barcode - make sure the barcode is assigned as GTIN on one of the products in the cache.
3. Notice the item is added to the cart.
4. Open search.
5. Notice a software keyboard is visible or the OS displays a button to display it.
6. Check that you can search for items - scanning at this moment, enters the code into the search field instead of automatically adding the product into the cart as long as the search input has focus. I think this behavior is fine and might be even desired, especially considering we can't reliable differentiate between barcode scanner and keyboard. Wdyt?

I also tested
- scanning is disabled on checkout
- scanning is disabled when the FF is off
- back gesture/button works as expected

### Images/gif

https://github.com/user-attachments/assets/c86b3a96-ee16-45bc-85ac-5a6288923820

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
@@ -186,6 +186,10 @@ class WooPosCartViewModel @Inject constructor(
                     is ParentToChildrenEvent.RemoveCouponsClicked -> {
                         removeCouponsFromCart()
                     }
+
+                    is ParentToChildrenEvent.BarcodeScanned -> {
+                        onBarcodeScanned(event.barcode)
+                    }
                 }
             }
         }
@@ -315,6 +319,14 @@ class WooPosCartViewModel @Inject constructor(
 
     private fun clearCart() {
         _state.value = WooPosCartState()
+    }
+
+    private fun onBarcodeScanned(barcode: String) {
+        viewModelScope.launch {
+            val product = getProductByGtinOrSku.invoke(barcode)
+            val itemNumber = getItemNumber()
+            product.toCartListItem(itemNumber)
+        }
     }
 
     private fun getItemNumber(): Int {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
@@ -324,6 +324,9 @@ class WooPosCartViewModel @Inject constructor(
     }
 
     private fun onBarcodeScanned(barcode: String) {
+        if (_state.value.cartStatus == CHECKOUT) {
+            return
+        }
         viewModelScope.launch {
             if (_state.value.body == WooPosCartState.Body.Empty) {
                 analyticsTracker.track(InteractionWithCustomerStarted)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
@@ -325,9 +325,16 @@ class WooPosCartViewModel @Inject constructor(
 
     private fun onBarcodeScanned(barcode: String) {
         viewModelScope.launch {
+            if (_state.value.body == WooPosCartState.Body.Empty) {
+                analyticsTracker.track(InteractionWithCustomerStarted)
+            }
+            // TBD display a loading state when searching for a product
             val product = getProductByGtinOrSku.invoke(barcode)
+            // TBD handle cases when the barcode is not found
+            // TBD handle cases when the product is a variation
             val itemNumber = getItemNumber()
-            product.toCartListItem(itemNumber)
+            val cartListItem = product.toCartListItem(itemNumber)
+            _state.value = updateStateWithNewItem(cartListItem)
         }
     }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
@@ -10,6 +10,7 @@ import com.woocommerce.android.R
 import com.woocommerce.android.model.Product
 import com.woocommerce.android.model.ProductVariation
 import com.woocommerce.android.ui.woopos.common.data.WooPosGetCouponById
+import com.woocommerce.android.ui.woopos.common.data.WooPosGetProductByGtinOrSku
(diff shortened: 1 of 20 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>WooPosGetProductByGtinOrSku</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -317,6 +323,24 @@ class WooPosCartViewModel @Inject constructor(
         _state.value = WooPosCartState()
     }
 
+    private fun onBarcodeScanned(barcode: String) {
+        if (_state.value.cartStatus == CHECKOUT) {
```
kidinov: np: better to check what states are supported then which one is not, e.g.

```
if (_state.value.cartStatus !in listOf(CHECKOUT, EMPTY)) return
```

So if a new one introduced it's still handled correctly

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsScreen.kt
```diff
@@ -98,11 +102,16 @@ private fun MainItemsList(
     onSearchEvent: (WooPosSearchUIEvent) -> Unit,
     onTabClicked: (WooPosItemsToolbarViewState.Tab) -> Unit,
     onAddCouponEvent: () -> Unit,
+    onBarcodeScanned: (String) -> Unit,
     onBackClicked: () -> Unit,
 ) {
     Box(
         modifier = modifier
             .fillMaxSize()
+            .listenForBarcodes(
```
kidinov, line 111: Do we have any reasons to not put it in the cart itself so we don’t have to send the event via home?

kidinov (approved):
LGTM! And it works cool! 

I left a couple of comments, please take a look.

> I think this behavior is fine and might be even desired, especially considering we can't reliably differentiate between a barcode scanner and a keyboard. What do you think?

Not sure if it's desired, but it’s probably not a problem. 
Two considerations though:
* Not clear what to do with other search fields, e.g. coupons.
* We don't search for GTIN in the search right now, so this may be confusing for the merchants that nothing is found when they scan with the search in focus, and an item is found when it's not in focus.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
kidinov: I meant something else 😮‍💨 
This:
```
if (_state.value.cartStatus in listOf(EDITABLE, EMPTY)) {
....
```

malinajirka: Updated in https://github.com/woocommerce/woocommerce-android/pull/14139/commits/e51407a0a8595f34ed64fb7e974a1d67fa3c4436

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsScreen.kt
malinajirka: The reason is only conceptual. It feels off that cart would be responsible for barcode scanning. 

I also think that anyone looking for the barcode scanning code might have troubles finding it if it's in the cart view. The fact that we go through home is an unrelated architectural decision and I don't think we should drive code-placement based on that. 

Wdyt?

malinajirka:
Thanks for the review @kidinov !

> We don't search for GTIN in the search right now, so this may be confusing for the merchants that nothing is found when they scan with the search in focus, and an item is found when it's not in focus.

I agree that if the search searched also by gtin, it would have been much better UX. Do you know if there is something blocking us from searching by gtin (is it supported on the backend?)?

> Not clear what to do with other search fields, e.g. coupons.

Yeah, I guess this is one of the downsides of using HID. It's especially tricky, since I can imagine some users might theoretically scan barcode which in reality is the coupon code => so it might potentially be desired to input the code into the coupon search. But of course, it's hard to guess without user feedback and it might be individual.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsScreen.kt
kidinov: I’m curious, why the items (product, coupons, customers etc) responsible for barcode scanning and not the cart? I don’t think either of them are directly related to the barcode scanning, but at least, the products are added to the cart suggests that there might be a connection between them. Also, we want to make it possible to scan a product when the cart is visible, so there is another logical (not direct though) connection to it.

And the practical advantage is that we don't need to add extra code to send what's scanned to the cart

 Wdyt?

kidinov:
@malinajirka 

> I agree that if the search searched also by gtin, it would have been much better UX. Do you know if there is something blocking us from searching by gtin (is it supported on the backend?)?


We've just added name + sku search recently as it was not supported. I believe name + sku + gtin are not supported even now

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsScreen.kt
malinajirka: Hmm, my train of thoughts was that the left part of the screen is responsible for selecting products/coupons/customers, while the right side is responsible for displaying the selected items. So that's why I feel the scanning part belongs more to the left side than the right side. Having said that, I guess we could make arguments for both approaches. If you feel it's better to move it to the cart, feel free to spin up a PR.

PR #14152, opened 2025-06-05
[Orders] UI improvements for small tablets

samiuelson:
## Fix support for large font scale size on small tablets

WOOMOB-537

### Description

This PR adds a couple of improvements for Orders screen on small tablets, especially with increased size of font and increased pixel density:
1. Ellipsizing texts
2. Shrinking horizontal margins (only on small tablets in portrait)
3. Optimizing widths ration of Orders | Order Details screen

### Testing information

Check if you can reproduce any UI glitches with accessibility size settings (screen size, font size).

Tested **Orders | Order** details screen on different devices and configurations:
1. Small tablet (10.2" 1280x800px)
4. Medium tablet (10.9" 1440x2304px)
5. Increased screen size (AS –> Running Devices –> Device UI Shortcuts)
6. Increased font size (AS –> Running Devices –> Device UI Shortcuts)

Results below.

### Images/gif

* **Small tablet** with increased screen size scale and font size scale in accessibility settings
Untitled1
* **Small tablet** with very increased screen size scale and font size scale in accessibility settings
Untitled2
* **Normal tablet** landscape – bigger margins in details pane
Untitled3
* **Normal tablet** portrait - smaller margins in details pane
Untitled4


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailFragment.kt
```diff
@@ -108,6 +110,7 @@ class OrderDetailFragment :
     companion object {
         val TAG: String = OrderDetailFragment::class.java.simpleName
         private const val MARGINS_FOR_TABLET: Float = 0.1F
+        private const val MARGINS_FOR_SMALL_TABLET: Float = 0.01F
```
kidinov: I believe we should use something like 16dp in this instance; otherwise, it may not adhere to the 4dp grid guideline and become noticeable as this is small padding

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListFragment.kt
```diff
@@ -97,7 +97,7 @@ class OrderListFragment :
         const val FILTER_CHANGE_NOTICE_KEY = "filters_changed_notice"
 
         private const val JITM_FRAGMENT_TAG = "jitm_orders_fragment"
-        private const val TABLET_LANDSCAPE_WIDTH_RATIO = 0.3f
+        private const val TABLET_LANDSCAPE_WIDTH_RATIO = 0.4f
```
kidinov: While it's name "landscape" it's actually set 0.4 to both Portrait and landscape orientations. So:
* Maybe name it properly
* It feels that 0.3 is better for landscape, but 0.4 for portrait

kidinov (commented):
LGTM! I left a couple of suggestions please take a look

Also, I'll leave this up to you, but now, in some cases we show 2 panes in portrait and 1 pane in landscape, which is odd

https://github.com/user-attachments/assets/cd4355d2-bcfa-41fa-bf70-024b548e48c4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListFragment.kt
kidinov: image

Yeah, it looks odd now in landscape on big tablets

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailFragment.kt
samiuelson: Makes sense. Set to fixed 16dp: f2a761dc567229b29acf8ea40e1d90035452ad16

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListFragment.kt
samiuelson: Switched to .4 in portrait and .3 in landscape: a3d096fc0004bf5d54c18476ad634fb1a0885a3b

samiuelson:
Thanks for review, @kidinov. 

> Also, I'll leave this up to you, but now, in some cases we show 2 panes in portrait and 1 pane in landscape, which is odd

Good catch. Since this is a separate bug, and may require some more tests and discussion, I raised a separate PR with the fix: https://github.com/woocommerce/woocommerce-android/pull/14160. You can take a look.

kidinov (commented):
@samiuelson I get this on a small tablet now (notice large paddings)


image

image

image

samiuelson:
> @samiuelson I get this on a small tablet now (notice large padding)

👋 @kidinov, I can't reproduce such large margins on a similar configuration. 

Could you check the actual **dp** size of the screen, for example, by measuring the dp size of the root container using the layout inspector in AS? It's the dp value of the shorter dimension that is being classified in `Context.windowWidthSizeClass`. 

I suspect your device has larger dp screen width class than `MEDIUM_SCREEN_MAX_WIDTH` and that's why it shows larger margins.

kidinov:
@samiuelson 

image

image

kidinov:
@samiuelson  👋

I just realised that I tested it on the products screen, not orders. Sorry about that. We'd need to implement this on the products though, otherwise it looks inconsistent.

kidinov (approved):

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14154, opened 2025-06-06
[WOOMOB-558] Connect barcode scanning results with ProductSearcher

malinajirka:
WOOMOB-558

### Description

This PR replaces the barcode-simulation flow in the POS cart with a real searchByIdentifier implementation and updates related tests, utilities, and resources.
- display a loading item when a barcode is scanned,
- replace it with product item if the product is found (either locally or remotely by either gtin or sku)
- replace it with error item if the product is not found

This PR also refactors WooPosSearchByIdentifierRemote to support parallel tasks. Previously the class was broken since invoking it multiple times would override its local values ultimately ending up in a deadlock situations. The update implementation stores a map of continuations that are resumed when a result is received. Moreover, it doesn't send duplicated requests and instead if a request for the same identifier is in progress, it just waits for it's competion.

P.S. I can extract changes to WooPosSearchByIdentifierRemote into a separate PR, let me know if you prefer that. It felt like it might be an overkill, considering most of the other changes are removals of unused classes and basic logic for adding items to the cart.

### Testing information

Test the following cases:
1. Product found by gtin in remote
2. Product found by sku in remote
3. Product found by gtin locally
4. Product found by sku locally
5. Multiple parallel remote searches work as expected

### The tests that have been performed

The above + various simulations of tens of products being added in parallel.

### Images/gif

https://github.com/user-attachments/assets/592cc700-3bd3-43bd-8c8b-74b156097235

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14155, opened 2025-06-06
[WOOMOB-560] Prevent users from adding unsupported products by scanning barcode

malinajirka:
WOOMOB-560

**Do not merge label**
- ~We haven't decided on the desired behavior - see the slack discussion referenced below.~
- https://github.com/woocommerce/woocommerce-android/pull/14154 needs to be merged first.

### Description

This PR update the Barcode Product Search logic to ensure the app doesn't add products that are not supported in the POS into the cart upon barcode scan. We discussed whether we should pass the filters to the API to ensure the product is not even returned from the API, however, we decided to go with local filtering instead (p1749200146020459-slack-C070SJRA8DP).

### Testing information

1. Set GTIN on an unsupported product type.
2. Scan the GTIN in the app.
3. Notice the product is not added and `Unknown scanned item` error is displayed.

The above + scan of an existing supported product.

https://github.com/user-attachments/assets/29421c8a-8d85-48d9-ac9f-26d8e3e3778b


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
```diff
@@ -1,22 +1,55 @@
...
+            result
+        } else {
+            WooPosSearchByIdentifierResult.Failure(WooPosSearchByIdentifierResult.Error.UnsupportedProduct)
+        }
+    }
+
+    private fun meetsFilterRequirements(product: Product): Boolean {
+        val hasValidStatus = product.status?.value == filterConfig.filters[ProductFilterOption.STATUS]
+
+        val meetsDownloadableRequirement = !(
```
kidinov: I am surprised that detekt doesn't complain about this formatting. Looks odd 😀

malinajirka: It actually complained when it was on the same line 🤷  - I agree it looks odd.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
```diff
@@ -1,22 +1,55 @@
...
-            return WooPosSearchByIdentifierResult.Success(localProduct)
+            return filterOutUnsupportedProducts(WooPosSearchByIdentifierResult.Success(localProduct))
         }
 
-        return remoteSearcher(identifier, format)
+        val remoteResult = remoteSearcher(identifier, format)
+        return filterOutUnsupportedProducts(remoteResult)
+    }
+
+    private fun filterOutUnsupportedProducts(result: WooPosSearchByIdentifierResult): WooPosSearchByIdentifierResult {
```
kidinov: The naming a bit confusing. To me it felt like it takes list of products and filter out unsupported, while what it actually does it determines if given product (wrapped in the result) valid or not

Also, maybe extract this to a separate class, e.g.
[Extract_validation_to_a_separate_class.patch](https://github.com/user-attachments/files/20655264/Extract_validation_to_a_separate_class.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
kidinov: Maybe to open this up to avoid this

```
val meetsDownloadableRequirement = !product.isDownloadable || 
    filterConfig.filters[ProductFilterOption.DOWNLOADABLE] != DownloadableOptions.FALSE.toString()
```

kidinov (approved):
LGTM! Thanks @malinajirka !


image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
malinajirka: Fixed in e9cfb4b971698cac34c7d823ec26963a1398267a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
malinajirka: Good point regarding the name, it sounds like if it was a method for a collection. I renamed it to filterUnsupportedProductResult in 87527b53c1844a2ef77c1dae873285601da0b41a.

Thanks for the patch as well, however, it feels like an overkill to me. Filtering out unsupported products still feels like part of the search. Considering the class is quite simple and we don't need to re-use the logic anywhere else I don't see much value from extracting it in this case.

PR #14157, opened 2025-06-06
[Better Tablet Products] Fix keyboard hiding unexpectedly on product tablet search

toupper:
Closes WOOMOB-540

#### Summary
This PR addresses an issue where the keyboard would unexpectedly hide when searching for products on a tablet. The root cause was that navigating to an empty product detail view was triggered whenever the user typed characters and the query was shorter than 3 characters.

#### Fix
To resolve this, we prevent navigation to an empty product detail view while a search is active. This aligns with the existing behavior in the Orders screen. Supporting navigation during search would require more complex logic, which we deem unnecessary at this point.

#### Changes
- Prevent `OpenEmptyProduct` during search:
The `ProductListViewModel` no longer triggers the `ProductListEvent.OpenEmptyProduct` event if a search is in progress.
- Adjust execution order in `onSearchOpened`:
We now update the viewStatus before resetting the product list. Previously, the keyboard would hide immediately because `isSearching` was still false during the navigation trigger.

### Steps to reproduce

On a tablet

1. Go to the Products tab
2. Start searching for a product
3. See that the keyboard is not hidden whenever we type the first characters

Tablet portrait, phone.

#### Before

https://github.com/user-attachments/assets/95ccb0d1-4e31-439a-aaa4-2d06e14d4719

#### After

https://github.com/user-attachments/assets/962c164c-cbb8-481b-8627-8d537f8dab5e

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
@@ -224,12 +224,12 @@ class ProductListViewModel @Inject constructor(
     }
 
     fun onSearchOpened() {
-        _productList.value = emptyList()
         viewState = viewState.copy(
             isSearchActive = true,
             displaySortAndFilterCard = false,
             isAddProductButtonVisible = false
         )
+        _productList.value = emptyList()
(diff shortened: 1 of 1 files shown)


malinajirka (approved):
LGTM, thanks!

PR #14160, opened 2025-06-09
[Bug] Fix determineIfTwoPanesShouldBeUsed condition

samiuelson:
### Description

This PR fixes issue reported [here](https://github.com/woocommerce/woocommerce-android/pull/14152#pullrequestreview-2909577923) — in some cases, in portrait orientation, on tablet, we are showing two two-pane layout and in landscape, a single-pane layout.

The updated logic takes into account device orientation when checking screen size against the minimal value.

### Steps to reproduce

1. Run app on small tablet (e.g. 10", 1280x800px)
2. Go to Orders and switch between portrait and landscape
3. Observe single pane layout in landscape.

### Testing information

Verify that the dual pane layout is visible both in portrait and landscape.

As above.

### Images/gif


WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/ContextExt.kt
```diff
@@ -33,14 +34,19 @@ val Context.windowWidthSizeClass: WindowSizeClass
...
     val minScreenShortSizeDP = MIN_SCREEN_SHORT_SIZE_DP
     val minScreenLongSizeDP = MIN_SCREEN_LONG_SIZE_DP
 
     val shortSize = min(widthDp, heightDp)
     val longSize = max(widthDp, heightDp)
 
-    return shortSize >= minScreenShortSizeDP && longSize >= minScreenLongSizeDP
+    val orientation = resources.configuration.orientation
+    return when (orientation) {
+        Configuration.ORIENTATION_PORTRAIT -> shortSize >= minScreenShortSizeDP
```
kidinov: Sorry, not following, what does it do? For instance, you have 800x1200. Short size will be always 800 doesn't matter in what orientation it is, right? The same for landscape. 800x1200 will be always bigger than 674 x 800


https://github.com/user-attachments/assets/2a0a59e9-a97b-4317-addc-6872c09c7165

kidinov (commented):
@samiuelson Thanks for working on this! Please take a look into the comment I left

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/ContextExt.kt
samiuelson: > For instance, you have 800x1200. Short size will be always 800 doesn't matter in what orientation it is, right? The same for landscape. 800x1200 will be always bigger than 674 x 800

Both `resources.configuration.screenWidthDp` and `resources.configuration.screenHeightDp` return screen size values excluding window insets. I tested the values behind these properties on emulated tablet and physical Pixel 7 Pro devices – in portrait and landscape. The result is that screen size is different in portrait and landscape orientation even though no additional settings were changed. 

The OS calculates these values as effective available size left for the app—excluding staff like status and nav bars. Apparently the insets sizes are different in landscape and portrait orientations for the same device. This also explains why the 2-pane layout was shown in portrait and 1-pane layout was shown in landscape—even though the width of landscape is bigger than width of portrait, the height in landscape was falling below minimum value.

After some thinking, I lean toward simplifying the logic behind `isTwoPanesShouldBeUsed` to take into account only the current width of the screen. In the end, the width is the most important factor to decide if the screen can fit  two panes together. I did this in these commits: 78db1ef, ecf073e. 

Let me know what you think, @kidinov.

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/ContextExt.kt
```diff
@@ -15,34 +15,17 @@ import androidx.annotation.ColorRes
 import androidx.core.content.ContextCompat
 import com.woocommerce.android.util.SystemVersionUtils
 import kotlinx.parcelize.Parcelize
-import kotlin.math.max
-import kotlin.math.min
 
+private const val MIN_WIDTH_DP_FOR_TWO_PANES = 674
```
kidinov: @samiuelson should it be 800 as it used to be for the long side?

kidinov: Also, while I like the approach it means that even on small phones in landscape mode we can show 2 pane layout which as far as I remember people complained about in the past


Maybe something simple like that?


```
private fun determineIfTwoPanesShouldBeUsed(widthDp: Int, heightDp: Int): Boolean {
    if (widthDp < MIN_SCREEN_WIDTH_DP) {
        return false
    }
    
    if (heightDp < MIN_SCREEN_HEIGHT_DP) {
        return false
    }
    
    return true
}
```

samiuelson: Yeah, this would make it equivalent to the original condition:
```kotlin
private fun determineIfTwoPanesShouldBeUsed(widthDp: Int, heightDp: Int): Boolean {
    return widthDp >= MIN_SCREEN_WIDTH_DP && heightDp >= MIN_SCREEN_HEIGHT_DP
}
```
Which is probably the best way to go. I'd just slightly reduce the `MIN_SCREEN_HEIGHT_DP` value, but would keep `MIN_SCREEN_WIDTH_DP` as it is.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14161, opened 2025-06-09
[WOOPRD-568][Woo POS][Barcodes] Handle search of the variations

kidinov:
WOOPRD-568

### Description

The PR adds an extra network call when scanned product is variation, so the variation object is fetched and cached

### Steps to reproduce

* Open POS
* Scan a variation -> Notice that it's been added 
* Scan it again -> Notice that it's been added instantaneously
* Scan a normal product to validate that no regression introduced 

Above

### Images/gif

https://github.com/user-attachments/assets/2df37dce-3d48-4591-9bad-f04aea4be5a9


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
```diff
@@ -50,6 +62,22 @@ class WooPosSearchByIdentifier @Inject constructor(
         return hasValidStatus && meetsDownloadableRequirement && hasValidType
     }
 
+    private fun meetsVariationFilterRequirements(variation: ProductVariation): Boolean {
+        val requiredStatus = variationFilterConfig.filters[VariationFilterOption.STATUS]
+        val hasValidStatus = when (requiredStatus) {
```
malinajirka, line 67: ⚠️ I'm either missing something or this code compares the status in the filter (static value) with the string "publish" (static value) => it doesn't use the dynamic variation value at all.

🔍 I would consider using the same approach for detection as we do in meetsFilterRequirements.

P.S. I'd personally avoid hardcoding `publish` here, we could use WooPosVariationsTypesFilterConfig.VARIATION_STATUS_PUBLISH or `CoreProductStatus.Publish.value`. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
```diff
@@ -50,6 +62,22 @@ class WooPosSearchByIdentifier @Inject constructor(
...
+            "publish" -> variation.isVisible
+            else -> false
+        }
+
+        val meetsDownloadableRequirement = !(
```
malinajirka: 🔍  We had the same code in `meetsFilterRequirements` and you shared a concern it's not as readable so we changed it. I think we should keep these two consistent 
```
val meetsDownloadableRequirement = !variation.isDownloadable ||
            variationFilterConfig.filters[VariationFilterOption.DOWNLOADABLE] != DownloadableOptions.FALSE.toString()
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -548,6 +548,19 @@ class WooPosCartViewModel @Inject constructor(
                 )
             }
 
+            is WooPosSearchByIdentifierResult.VariationSuccess -> {
+                val product = getProductById(variation.remoteProductId)!!
```
malinajirka, line 559: ❓ Are we truly guaranteed the product is in the cache? I'd personally avoid using `!!` here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierRemote.kt
```diff
@@ -173,18 +176,61 @@ class WooPosSearchByIdentifierRemote @Inject constructor(
...
+        val parentId = product.parentId
+        val variationId = product.remoteId
+
+        if (parentId <= 0) {
+            return WooPosSearchByIdentifierResult.Failure(
+                WooPosSearchByIdentifierResult.Error.ProductNotFound
+            )
+        }
+
+        val variationResult = productStore.fetchSingleVariation(
```
malinajirka, line 206: ❓ If I understand it correctly, if the results of the `/product?global_....` endpoint returns an object that has `product.type` equal to `variation`, we send a new request fetching just the variation. 

Are there any data missing in the response from the server for the variation or why do we need to re-fetch it? Have you considered implementing a custom gson parser that would look at the product.type field when parsing and would parse it either into Product or ProductVariation object in the FluxC layer?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
```diff
@@ -1,28 +1,51 @@
...
 import android.util.LruCache
+import com.woocommerce.android.model.ProductVariation
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import javax.inject.Inject
+import javax.inject.Singleton
 
-class WooPosVariationsLRUCache<K, V> @Inject constructor() {
-
+@Singleton
```
malinajirka, line 10: 📓 I don't see anything wrong with it, but using singleton here will retain the cache even if the user leaves the POS section of the app. Is this an intentional change?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
```diff
@@ -1,28 +1,51 @@
...
         mutex.withLock {
             cache.put(key, value)
         }
     }
+
+    suspend fun add(key: Long, value: ProductVariation) {
+        mutex.withLock {
+            val list = cache.get(key)
+            if (list != null) {
+                val updatedList = list.toMutableList().apply {
```
malinajirka: 📓 AFAICT this code will change the order of the items which I assume might result in items jumping around as we re-fetch the list. Have you considered using `.map` instead?

```
val updatedList = list.toMutableList().map {
                    if(it.remoteVariationId == value.remoteVariationId) {
                        value
                    } else {
                        it
                    }
                }
```

The code isn't identical to your code, since your code removes all items with the remoteVariationId, however, I'm not sure if we can truly have more than one in the cache - feels like a bug if we do, doesn't it?

malinajirka (changes requested):

I haven't tested the code yet, since I found a couple potential issues that might result in changes that will need to be retested again. Looking forward to your thoughts!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
kidinov: Theoretically in what order we display the list is responsibility of the VM, but I don't see that we do the sorting there. So just to be sure I reimplemented the add method here

316ae1294cb6075f02efd712ef61bf6c6fc65b38

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
kidinov: Without singleton another instance of the cache is injected into `WooPosSearchByIdentifierLocal`, not the same one which keeps the cache itself. `WooPosProductsInMemoryCache` is singletone as well, so yes, intentional

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierRemote.kt
kidinov: > ❓ If I understand it correctly, if the results of the /product?global_.... endpoint returns an object that has product.type equal to variation, we send a new request fetching just the variation.

Yes, that's correct!

> Are there any data missing in the response from the server for the variation or why do we need to re-fetch it? Have you considered implementing a custom gson parser that would look at the product.type field when parsing and would parse it either into Product or ProductVariation object in the FluxC layer?

I haven’t checked for any missing data, but my assessment is that rewriting half of the app would be necessary if the products endpoint started returning Variation OR Product types. This is because Variation type is separated both on the database level and the model level. Variations also have it's own end points

Having said that, I haven’t delved deeply into this, and this is just my general understanding.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
malinajirka: > Without singleton another instance of the cache is injected into WooPosSearchByIdentifierLocal, not the same one which keeps the cache itself. 

This could be solved by using @ActivityScope instead. 

> WooPosProductsInMemoryCache is singletone as well, so yes, intentional

👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
malinajirka: I think we usually rely on the order from the API instead of manually sorting the items in the VM as it can cause weird side-effects with pagination and re-fetches. Either case, the updated code LGMT, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
kidinov: >  Are we truly guaranteed the product is in the cache?

By the current logic it must be guaranteed as we fetch a product firstly to fetch the variation afterwards

>  I'd personally avoid using !! here.

What would you do then? We have to have product object to show variations, the same as a variation is selected via `handleVariationClicked`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
kidinov: Done here - 79233018db9dd141b191712ea5d4a228a7d12c58

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
kidinov: > ⚠️ I'm either missing something or this code compares the status in the filter (static value) with the string "publish" (static value) => it doesn't use the dynamic variation value at all.

It checks if filter require only published variations and if it is then we check if the variation `isVisible` which means status is `publish`. 

> 🔍 I would consider using the same approach for detection as we do in meetsFilterRequirements.

We can't as those are 2 different models and `Variation` doesn't have `status` field, only `isVisible`. Why? I don't know but probably changing the model because of that is not reasonable

> P.S. I'd personally avoid hardcoding publish here, we could use WooPosVariationsTypesFilterConfig.VARIATION_STATUS_PUBLISH or CoreProductStatus.Publish.value. Wdyt?


Done here - d0f31fb04f1388e2a0217c16e86502ee32597403

kidinov:
@malinajirka please take a look again!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
kidinov: > This could be solved by using @ActivityScope instead.

If the system kills an activity (e.g., it can be simulated using “don’t keep activities”), will it recreate all its dependencies? If so, will the cache be lost? 🤔

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
kidinov: > I think we usually rely on the order from the API instead of manually sorting the items in the VM as it can cause weird side-effects with pagination and re-fetches. Either case, the updated code LGMT, thanks!

No, not really. At least for the search we specifically sort all locally to avoid buggy/wrong sorting from the API, as they may not match the order we push the data from the cache

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifier.kt
malinajirka: > It checks if filter require only published variations and if it is then we check if the variation isVisible which means status is publish.

🤦 I read it too early in the morning :P.

> We can't as those are 2 different models and Variation doesn't have status field, only isVisible. Why? I don't know but probably changing the model because of that is not reasonable

👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
malinajirka: Good question, yes, it would recreate all the dependencies and the cache would be lost.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsLRUCache.kt
malinajirka: TIL

malinajirka (approved):
Thanks for working on this and the changes! I've tested this and it works like a charm 🚢

PR #14162, opened 2025-06-09
[WOOMOB-547] Add Heuristics for differentiating between barcode scanner and HW keyboard

malinajirka:
WOOMOB-373

### Description

This PR introduces the BarcodeInputDetector class that helps differentiate between input from a hardware keyboard and input from a barcode scanner.

**How It Works:**
The detector differentiates between manual keyboard typing and automatic barcode scanner input by analyzing timing patterns:

- Barcode scanners typically input characters very rapidly in sequence.
- The detector tracks time between keystrokes to identify scanner patterns
- When a barcode is detected, it's passed to the provided callback function

**Time-based detection:** 
Uses two timing thresholds 
- MAX_SCANNER_INTER_CHAR_DELAY_MS (100ms): Maximum delay between characters from scanner (my scanner inputs characters between 5-60ms). If two keystrokes are further apart than 100ms it's considered as human input and ignored.
- MAX_SCANNER_TOTAL_SCAN_TIMEOUT_MS (1500ms): Maximum total time for a complete scan (I set it to 1500 as a reasonably sized barcode won't be longer than 25 characters if we assume average character inter-delay 60ms - my barcode scanner has average delay around 20ms so this feels like a huge enough buffer). If the timeout elapses, the input is validated and callback is invoked (this is handling for scanners that do not pass end-of-line character.

**Character validation:** Only allows alphanumeric and common special characters
**Minimum length:** Rejects potential barcodes shorter than 4 characters
**Auto-completion:** Scan completes on Enter key or after timeout
**Reset mechanism:** Automatically resets state when human typing is detected

### Testing information

- Try scanning barcodes and verify items are added to the cart.
- Try typing on a hardware keyboard and verify items are not added to the cart unless you type at least 4 characters extremely fast.

- I've tested this with with a hardware scanner -  I tried disabling the end-of-line suffix to verify items are added to the cart even when the end-of-line character is not provided.

### Images/gif

No user-facing changes.

cc @joshheald

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
@@ -62,9 +62,6 @@ class BarcodeInputDetector(
     private val currentTimeProvider: CurrentTimeProvider,
 ) {
     companion object {
-        @Suppress("SpellCheckingInspection")
-        const val ALLOWED_BARCODE_CHARS =
-            "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-._~:/?#[]@!$&'()*+,;="
         const val MAX_SCANNER_TOTAL_SCAN_TIMEOUT_MS = 1500L
         const val MAX_SCANNER_INTER_CHAR_DELAY_MS = 100L
         const val MIN_BARCODE_LENGTH = 4
@@ -75,7 +72,6 @@ class BarcodeInputDetector(
     private var lastCharTime: Long = -1L
     private var timeoutJob: Job? = null
 
-    @Suppress("ReturnCount")
     fun handleKeyInput(char: Char): Boolean {
         val currentTime = currentTimeProvider.currentDate().time
 
@@ -86,10 +82,9 @@ class BarcodeInputDetector(
                     (currentTime - lastCharTime <= MAX_SCANNER_INTER_CHAR_DELAY_MS)
                 if (isLikelyScanner) processBarcodeBuffer()
                 clear()
-                return true
             }
 
-            in ALLOWED_BARCODE_CHARS -> {
+            else -> {
                 cancelTimeout()
 
                 if (scanStartTime == -1L) {
@@ -97,14 +92,9 @@ class BarcodeInputDetector(
                 } else {
                     handleContinuedScan(char, currentTime)
                 }
-                return true
-            }
-
-            else -> {
-                clear()
-                return false
             }
         }
+        return true
     }
 
     private fun startNewScan(char: Char, currentTime: Long) {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
@@ -61,7 +61,7 @@ class BarcodeInputDetector(
     private val coroutineScope: CoroutineScope,
     private val currentTimeProvider: CurrentTimeProvider,
 ) {
-    companion object {
+    private companion object {
         const val MAX_SCANNER_TOTAL_SCAN_TIMEOUT_MS = 1500L
         const val MAX_SCANNER_INTER_CHAR_DELAY_MS = 100L
         const val MIN_BARCODE_LENGTH = 4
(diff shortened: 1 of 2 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeInputDetectorTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        // GIVEN
+        val detector = setupDetector(this)
+        val barcode = "1234567"
+
+        // WHEN
+        for (char in barcode) {
+            detector.handleKeyInput(char)
+            advanceTestTimeBy(10)
+        }
+        advanceTimeBy(BarcodeInputDetector.MAX_SCANNER_TOTAL_SCAN_TIMEOUT_MS)
```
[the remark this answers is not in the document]
malinajirka: Fixed

joshheald:
Thanks for the ping @malinajirka 

The heuristics as you describe them look good. My only question is why 4 characters?

> Minimum length: Rejects potential barcodes shorter than 4 characters

UPC-E is the shortest of the common symbologies we're looking to support, and that uses 6 characters. Is there some other use case I'm missing for shorter codes? 

That said, with custom barcodes in QR codes, I guess someone could have shorter codes. I think it comes down to: what's going to give us the best accuracy in scanning, without compromising common barcode format support?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
```diff
@@ -37,18 +48,115 @@ fun Modifier.listenForBarcodes(
...
+@VisibleForTesting
+class BarcodeInputDetector(
+    private val onBarcodeScanned: (String) -> Unit,
+    private val coroutineScope: CoroutineScope,
+    private val currentTimeProvider: CurrentTimeProvider,
+) {
+    companion object {
+        @Suppress("SpellCheckingInspection")
+        const val ALLOWED_BARCODE_CHARS =
+            "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-._~:/?#[]@!$&'()*+,;="
```
joshheald: What's the source for this character set?

malinajirka: Good question, I realized it doesn't include all the characters supported by some of the allowed symbologies so I removed this check completely 3e5c7726c99f255ab1e2339aca78b1ba417b2602 - for example Code 39 Ext. or Code 128 support all ASCII characters, QR codes support even more.

malinajirka:
Thanks for looking into this @joshheald !

> The heuristics as you describe them look good. My only question is why 4 characters?
> Minimum length: Rejects potential barcodes shorter than 4 characters
> UPC-E is the shortest of the common symbologies we're looking to support, and that uses 6 characters. Is there some other use case I'm missing for shorter codes?
> That said, with custom barcodes in QR codes, I guess someone could have shorter codes. I think it comes down to: what's going to give us the best accuracy in scanning, without compromising common barcode format support?

As you mentioned for example QR codes support codes of any length, however, I don't think we can come up with a heuristic which would reliable allow such scans. I tried to find a reasonably but as short as possible length to give us decent accuracy. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
```diff
@@ -37,18 +48,105 @@ fun Modifier.listenForBarcodes(
...
+        }
+}
+
+@VisibleForTesting
+class BarcodeInputDetector(
+    private val onBarcodeScanned: (String) -> Unit,
+    private val coroutineScope: CoroutineScope,
+    private val currentTimeProvider: CurrentTimeProvider,
+) {
+    companion object {
```
kidinov, line 65: np: probably should be private

kidinov:
@malinajirka 👋

Looks good, but I don't understand what specifically it gives as? That if a new line char is not coming, we'll still submit it as a barcode? The search in focus will still put scanned data in it, correct?

kidinov (approved):
The code looks good but I am 100% understand what it is for 😀. I asked it in the comment

malinajirka:
Thanks for the review!

> Looks good, but I don't understand what specifically it gives as? That if a new line char is not coming, we'll still submit it as a barcode? The search in focus will still put scanned data in it, correct?

It gives us two things:
1. support for scanners that do not put the `end-of-line` character after the scan,
2. typing on the keyboard and pressing enter won't be detected as scan.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
malinajirka: Updated in db700f6ecf96cd6b94bdcfe02cb696067701c226

PR #14164, opened 2025-06-10
[WOOPOS-577] Play sound when fetching product by identifier fails

malinajirka:
WOOMOB-577

### Description

This PR plays a sound when fetching a product after a successful barcode scan fails. I decided to use soundPool since it plays the sound faster reducing the delay between displaying the error and playing the sound.

I updated also the chaching sound to use the same approach for the same reason as well as for consistency.

### Testing information

1. Scan a barcode which isn't assigned to any product.
2. Notice the failure sound plays.

- I have tested the above + I tested successful scan doesn't play the sound.

### Images/gif

https://github.com/user-attachments/assets/b5f0dfa8-3f18-43ac-b7d1-9133b488c671


kidinov (approved):
LGTM! Tests good, too!

PR #14165, opened 2025-06-10
[WOOMOB-577] Propagate barcodes only in cart state

malinajirka:
WOOMOB-577

### Description

This PR ensures barcode scan results are processed only on the Cart screen/state.

### Testing information

1. Scan barcode on the cart screen and verify it works
2. Tap on checkout
3. Scan barcode and verify the result is ignored
4. Collect a payment
5. Scan barcode on the success screen and verify the barcode is ignored
6. Tap on New Order
7. Scan barcode on the cart screen and verify it works

The above.

### Images/gif


kidinov (commented):
@malinajirka 👋

I can consistently reproduce this. It clicks cash payment button when I scan a barcode on the checkout screen


https://github.com/user-attachments/assets/3d4e495e-3a8f-4f42-84de-9734c34d733b

**False alarm - it looks like that this behavior caused by Android studio connected to stream the video from the device and it puts focus on the buttons**

kidinov (approved):
LGTM!

PR #14166, opened 2025-06-10
Fix coroutines test deprecations

hichamboushaba:
### Description
After updating coroutines-test library in #14087, a bunch of warnings about using deprecated functions is thrown in fluxc unit tests and breaking the [CI](https://buildkite.com/automattic/woocommerce-android/builds/29749#019756bf-1bb4-4277-87b5-a0735a457e3b), this PR fixes those deprecations.

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
@@ -1,4 +1,4 @@
-@file:Suppress("DEPRECATION", "ForbiddenComment")
+@file:Suppress("DEPRECATION_ERROR", "ForbiddenComment")
 // TODO: @malinajirka Issue: https://github.com/woocommerce/woocommerce-android/issues/6899
 
 package com.woocommerce.android.ui.reviews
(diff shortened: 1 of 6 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
```diff
@@ -1,4 +1,4 @@
-@file:Suppress("DEPRECATION", "ForbiddenComment")
+@file:Suppress("DEPRECATION_ERROR", "ForbiddenComment")
```
hichamboushaba, line 1: They updated the deprecation level from warning to error in the new update, so I updated the suppress statement.

I'll try to find some time to update these tests to stop using deprecated APIs.

irfano (approved):
Thanks for taking care of this. I'm curious why it didn't fail in #14087. I'm approving this because it fixes the current failure, and addressing the deprecations isn't as urgent, so we can handle them later. Thank you!

hichamboushaba:
> I'm curious why it didn't fail in https://github.com/woocommerce/woocommerce-android/pull/14087

I think it could be because the PR's branch was outdated, I know we did some changes to the FluxC modules to enable `detekt` recently, maybe we also updated the build to treat warnings as errors recently, and this wasn't applied in the PR's branch.

PR #14167, opened 2025-06-10
[POS as a Tab] Add feature flag for the first iteration

toupper:
Part of WOOMOB-569

### Description

This brief PR adds a feature flag for the first iteration of the POS as Tab project.


WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -12,6 +12,7 @@ enum class FeatureFlag {
     ORDER_CREATION_AUTO_TAX_RATE,
     NEW_SHIPPING_SUPPORT,
     REVAMP_WOO_SHIPPING,
+    POS_AS_A_TAB_I1,
```
kidinov: np: another FF is called `WOO_POS_` 😀

kidinov (approved):
LGTM! Feel free to merge

PR #14168, opened 2025-06-10
`WCCustomerModel` to `Room`

wzieba:
Closes [AINFRA-550](https://linear.app/a8c/issue/AINFRA-550/migrate-wccustomermodel)

### Description

This PR migrates `WCCustomerModel` to `Room`.

### Testing information

1. On web go to `Users` > `Add User` (on top)
2. Fill the details, select role `Customer`, click `Add User`
3. Go to `WooCommerce` > `Customers`
4. Click on new customer name
5. Fill some details like `Address` or `City` in both `billing address` and `shipping address`
6. Go to mobile app 
7. Create new order
8. Tap on `Add customer details`
9. See the new customer is shown on the screen
10. Tap on it
11. Assert that the details of the customer are applied to the order (including shipping/billing addresses)

### Images/gif

Screenshot 2025-06-10 at 18 38 37


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -2190,6 +2190,9 @@ open class WellSqlConfig : DefaultWellConfig {
                 215 -> migrateAddOn(ADDON_WOOCOMMERCE, version) {
                     db.execSQL("DROP TABLE IF EXISTS WCProductSettingsModel")
                 }
```
ParaskP7, line 2192: Minor (🔍): Just because I see us leaving a space, consider doing the same between `215` and `216`.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerModel.kt
```diff
@@ -1,98 +1,49 @@
...
-    @Column var shippingState: String = ""
-    var analyticsCustomerId: Long? = null
-
-    override fun getId() = id
-
-    override fun setId(id: Int) {
-        this.id = id
-    }
-
-    override fun toString(): String {
```
ParaskP7, line 60: Question (❓): I really wonder why this `toString()` was created back then (see [FluxC#1920](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1920), including this `@JvmName("setIsPayingCustomer")`... 🤷

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CustomerDao.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.wordpress.android.fluxc.model.customer.WCCustomerModel
+
+@Dao
+internal abstract class CustomerDao {
+    @Query(
+        """
+        SELECT * FROM CustomerEntity
+        WHERE localSiteId = :siteId
+        AND remoteCustomerId = :customerId
+        LIMIT 1
```
ParaskP7: Question (❓): Is this `LIMIT 1` really needed here, with `localSiteId` and `remoteCustomerId` being the primary key, there is no chance that we'll get 2 results, right? 🤔

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/CustomerDaoTest.kt
```diff
@@ -0,0 +1,149 @@
...
+        sut.upsertCustomer(customerTwo)
+        sut.deleteCustomersForSite(LocalId(site.id))
+
+        // then
+        val storedCustomers = sut.getCustomersForSite(LocalId(customerOne.localSiteId.value))
+        assertThat(storedCustomers).containsExactly(customerOne)
+    }
+
+    @After
+    fun closeDb() {
```
ParaskP7: Minor (🔍): I think having this closer to `@Before` is better, otherwise it's easier to miss it and think of the database open.

PS: Same applies for `WCCustomerStoreTest`.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/store/WCCustomerStoreTest.kt
```diff
@@ -0,0 +1,238 @@
...
+        assertThat(result.isError).isTrue
+        assertThat(customerDao.getCustomersForSite(DEFAULT_SITE_MODEL.localId())).isEmpty()
+    }
+
+    @After
+    fun closeDb() {
+        roomDb.close()
+    }
+
+    private companion object {
```
ParaskP7, line 122: Praise (❤️): Nicely done, thanks for replacing all those mocks with the actual models, so much better!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
```diff
@@ -308,20 +308,23 @@ class OrderFilterCategoriesViewModel @Inject constructor(
...
-            OrderFilterCategoryUiModel(
-                categoryKey = CUSTOMER,
-                displayName = resourceProvider.getString(R.string.orderfilters_customer_filter),
-                displayValue = getCustomerDisplayValueFrom(customer.customerId),
-                orderFilterOptions = listOf(
-                    OrderFilterOptionUiModel(
-                        key = customer.customerId?.toString() ?: error("Customer ID is null"),
-                        displayName = getCustomerDisplayValueFrom(customer.customerId),
-                        isSelected = true
+        launch {
```
ParaskP7, line 311: Question (❓): I wonder why sometimes we chose doing this here (using `launch { ... }`), versus adding `runBlocking { ... }` on the store side (see `WCProductStore.getProductByRemoteId(...)` for example).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
```diff
@@ -308,20 +308,23 @@ class OrderFilterCategoriesViewModel @Inject constructor(
...
+            val displayValue = getCustomerDisplayValueFrom(customer.customerId)
```
ParaskP7, line 312: Praise (❤️): Nicely done, less duplicates, less db calls!

ParaskP7 (commented):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything seems to be working as expected (ish), good job on getting this migration done! 🌟 

-----

I have left few questions (❓) and some minor (🔍) comments for you to consider. Also, I have this testing question (🧪) below, and thus, I am **NOT** going to approve this PR atm, just to make sure we're not missing anything.

-----

Testing (🧪): While testing, following the testing instructions, but also, navigating to this `Menu` -> `Customers` screen, I noticed something quite weird. I want to discuss all with you, see below:

Before the db migration, I had the following:

before

After the db migration, I now have the following:

after

With the before and after of this `Customers` screen of mine remaining the same, 5 customers are shown at all times:

image

Any idea what is happening here? 🤔

FYI: I understand that the `remoteCustomerId = 0` of those 3 entries before the migration are now merged, then only 1 entry manages to persist on db, overriding the other 2, and that the screen gets populated with the online data first. 🤔

PS: When offline the screen is actually showing an error, not using the db at all, which by itself is very interesting. 🤔

image

Let me know what you think of all that, maybe my customer data is screwed and we shouldn't have a different customer with the same `remoteCustomerId`, or else, all those `remoteCustomerId = 0` customers on my side were created by mistake... 🤷

wzieba:
> Any idea what is happening here? 🤔

> FYI: I understand that the remoteCustomerId = 0 of those 3 entries before the migration are now merged, then only 1 entry manages to persist on db, overriding the other 2, and that the screen gets populated with the online data first. 🤔

👋 I'm not sure if I follow - what's the question here? I think you answered it correctly yourself. Do you think the end result should be different?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerModel.kt
wzieba: Probably by mistake, I think there's no action here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
wzieba: I think in this case it was easy to migrate to asynchronous behavior as I could clearly see the consequences and that it seems to work alright.

wzieba:
Clarifying question: how many customers do you have on the web? 3 or 5? Or 2 actually?

ParaskP7:
👋 @wzieba !

> 👋 I'm not sure if I follow - what's the question here? I think you answered it correctly yourself.

The problem here is that I am not liking my answer, this 1 entry overriding the other 2, one-by-one... 🤷

> Do you think the end result should be different?

That's the problem, I am very unsure about what I think should be the end result, thus I wanted to start this discussion with you. Maybe we should involve the product teams in that discussion here too. So, here are my itemized questions for all:

1. Is a customer with `remoteCustomerId = 0` a valid customer?
    - Is this somehow an inconsistent state I found myself with my site?
2. How come we can see 3 customer entries in the db, while 5 customers on screen?
    - What I see when visiting this screen is that it starts with 3 customers (from db) and then it add those extra 2 (missing) customers (from web).
3. Why isn't this screen working offline with data from db?

> Clarifying question: how many customers do you have on the web? 3 or 5? Or 2 actually?

I have 5 customers on the web, yes, of course, otherwise, this thing would have been quite weird.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
ParaskP7: Gotcha, okay then, it is a case-by-case scenario that you apply to such changes. 👍

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/CustomerDaoTest.kt
wzieba: 👍 1f5ff8d0028bda0311951bf89fde5b55af008354

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CustomerDao.kt
wzieba: Yes, redundant. Fixed in 9ea75007f659baef08e6e8f392665c9a58ee85af

wzieba:
@ParaskP7 I come back with some answers:

1. The “Add customer details” screen displays customers from two sources:
- Registered WooCommerce customers (from `CustomerEntity` with `remoteCustomerId` != 0)
- Guest customers fetched from the Analytics API (`CustomerFromAnalyticsEntity` with `remoteCustomerId = 0`)

For some reason, the decision was that Analytics API customers are saved in 2 places: both `CustomerEntity` (former `WCCustomerModel` in WellSql) and in `CustomerFromAnalyticsEntity`.

So yes - a customer with `remoteCustomerId = 0` is valid in this context. It represents a guest customer (someone who checked out without registering an account), and it’s not an inconsistent state - it’s actually expected behavior when the site has guest purchases.

> Why 3 in DB and 5 in UI?

You have 2 users on your website, plus 1 from `Analytics API`, that was merged with the rest from `Analytics API` as they all had the same primary key.

> Why doesn’t it work offline?

The implementation uses `CustomerListRepository#searchCustomerListWithEmail` which calls the API first - if it fails, no results are shown. But this PR didn't modify this behavior.

---

My fix for the above is to bring back the synthetic primary key. I was thinking about refactoring the code to keep "standard" customers in `CustomerEntity` and these from `Analytics API` in `CustomerFromAnalyticsEntity`, but it actually requires non-trivial amount of work and I don't think I should focus on this in scope of this PR. Proposed solution: 708993adaf92a59731c5c51a2057a8c1c97334c3

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CustomerDao.kt
ParaskP7: Thanks for the fix @wzieba ! 🚀

ParaskP7:
👋 @wzieba and thanks for coming back with answers, nice investigation! 🥇 

> So yes - a customer with remoteCustomerId = 0 is valid in this context. It represents a guest customer (someone who checked out without registering an account), and it’s not an inconsistent state - it’s actually expected behavior when the site has guest purchases.

This all is good to know! 🔖 

> You have 2 users on your website, plus 1 from Analytics API, that was merged with the rest from Analytics API as they all had the same primary key.

👍

> The implementation uses CustomerListRepository#searchCustomerListWithEmail which calls the API first - if it fails, no results are shown. But this PR didn't modify this behavior.

👍

---

> My fix for the above is to bring back the synthetic primary key.

Make total sense to me! 💯 

> I was thinking about refactoring the code to keep "standard" customers in CustomerEntity and these from Analytics API in CustomerFromAnalyticsEntity, but it actually requires non-trivial amount of work and I don't think I should focus on this in scope of this PR.

I agree, we shouldn't focus on this is scope of this PR. 👍

> Proposed solution: https://github.com/woocommerce/woocommerce-android/commit/708993adaf92a59731c5c51a2057a8c1c97334c3

On it... 👀

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/CustomerDaoTest.kt
```diff
@@ -0,0 +1,155 @@
...
+            localSiteId = LocalId(site.id),
+            username = username
+        )
+
+        // when
+        sut.upsertCustomer(expectedCustomer)
+
+        // then
+        val storedCustomer = sut.getCustomerByRemoteId(LocalId(site.id), RemoteId(remoteCustomerId))
+        assertThat(storedCustomer).usingRecursiveComparison().ignoringFields("id").isEqualTo(expectedCustomer)
```
ParaskP7: Minor (🔍): Consider split that into multiple lines, just like you did with those other 2 instances of such code below.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerModel.kt
```diff
@@ -1,98 +1,57 @@
...
+    /**
+     * Synthetic primary key used to uniquely identify a customer.
+     *
+     * Some customers, such as guest customers from the Analytics API (represented by
+     * [org.wordpress.android.fluxc.persistence.entity.CustomerFromAnalyticsEntity]), have `remoteCustomerId = 0`,
+     * which means they cannot be uniquely identified by (remote id x local site id) pair.
+     *
+     * To ensure uniqueness across both registered and guest customers, this synthetic ID is used.
+     */
+    @PrimaryKey(autoGenerate = true) val id: Int = 0,
```
ParaskP7, line 22: Praise (❤️): Thanks for [taking a closer look at how this all works](https://github.com/woocommerce/woocommerce-android/pull/14168#issuecomment-2984261681) and then bringing back this auto-generated synthetic `id`, making is a primary key to ensure the uniqueness of both registered and guest customers. 💪

ParaskP7 (approved):
FYI: I review and re-tested this change, with updated logic and all, everything LGTM, thanks for the fix @wzieba, we can proceed with the merge! 🥇 🚀 🙇

PR #14183, opened 2025-06-12
[POS as a Tab] Tab & Visibility

toupper:
### Description

With this PR, we add a tab for POS and show it when:

- Woo POS is enabled (same conditions previously used in the menu)
- The feature flag for "POS as a Tab" is enabled

### When visibility is updated

We update the POS tab visibility in the following cases:

- On app startup and when returning from the background
- When the active site changes
- On user login

### Navigation behavior

When the POS tab is tapped:

- The POS screen is launched
- The bottom navigation bar is hidden
- The tab is not visually selected (since it transitions to a new Activity)
#### On app startup and coming to foreground, when logged in

Open the app with a site where POS is enabled. The tab should be visible.
Open the app with a site where POS is not enabled. The tab should not be visible.

#### On site changed

Change your site in menu and check that the tab visibility is updated according to the the new site POS availability.

#### On login

Login and check that the tab visibility is shoen according to the the login site POS availability.

### Testing information

See above

See above

#### Switching sites

https://github.com/user-attachments/assets/76f977df-0c40-44f5-9e22-cfe32c354ab2

#### Login

https://github.com/user-attachments/assets/278a5f84-8d3a-4b10-8ee9-3f004fcdb007


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosIsPosAsTabEnabled</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosTabController</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

jaclync:
@toupper Nice progress with small amount of changes! From the screencasts, it looks like the POS tab is displayed right after logging in / switching stores, is it because it's shown by default, or the eligibility check results are cached for each connected site? Or the internet was so fast that all API calls from site settings (for store country/currency), remote feature flag, active plugins just returned with minimal delay?

In iOS, I'm exploring ways to optimize the POS eligibility check for the POS tab so that we can show the tab as soon as possible. This includes caching the eligibility result for each site for like 3 days in WOOMOB-599.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosIsPOSAsATabEnabled.kt
```diff
@@ -0,0 +1,10 @@
+package com.woocommerce.android.ui.woopos.tab
+
+import com.woocommerce.android.util.FeatureFlag
+import javax.inject.Inject
+
+class WooPosIsPOSAsATabEnabled @Inject constructor() {
```
malinajirka: 🔍 `POS` might be cleaner with camel case `Pos`. I'd also consider dropping the article `A`, we usually don't include it in the names even though it's dramatically incorrect => WooPosIsPosAsTabEnabled. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -0,0 +1,78 @@
...
+     */
+    fun refreshPOSTabVisibility() {
+        if (isPOSAsATabEnabled()) {
+            updatePOSTabVisibility()
+        }
+    }
+
+    private fun updatePOSTabVisibility() {
+        activity.lifecycleScope.launch {
+            setPOSTabVisibility(isWooPosEnabled())
```
malinajirka, line 69: ⚠️ This is not a flaw of this PR, but if my evaluation is correct, I think we should fix it here. I believe this code will start a coroutine on the main/UI thread. The isWooPosEnabled method should take care of switching context for long running operations, but it doesn't. For example `GetWooCorePluginCachedVersion` doesn't change the dispatcher and it hits the database. Similarly woocommerceStore.getSiteSettings also hits the database.

I see three options:
1. Change the coroutine context here to the IO dispatcher.
2. Change the coroutine context in the `WooPosIsEnabled.invoke()`. (The solution I'd personally go with here)
3. Or the most correct but the most time consuming approach, update all the coresponding sites and change the context as close to the time consuming operation as possible (eg. right before we hit the database).

Wdyt? Have I missed something or it truly runs on the UI thread?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -388,6 +394,8 @@ class MainActivity :
         // Track if App was opened from a widget
         trackIfOpenedFromWidget()
 
+        posTabController.refreshPOSTabVisibility()
```
malinajirka: 💡 I think we might want to consider registering the WooPosTabController to [the activity lifecycle ](https://developer.android.com/topic/libraries/architecture/lifecycle)and calling this method within the WooPosTabController. I think we should be able to move there even the two methods we invoke in `onCreate`, ultimately keeping the MainActivity completely clean of the controller's logic -> all the logic would be in the controller and the only code in the activity would be the registration for the lifecycle. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -725,6 +733,7 @@ class MainActivity :
         )
         finish()
         startActivity(intent)
+        posTabController.refreshPOSTabVisibility()
```
malinajirka: ❓ Honest question as I haven't tested this and I'm not sure. Is this truly necessary even though we invoke the same method in `onResume`? I'd expect both onCreate and onResume should be called when we finish the activity and start a new one 🤔.

malinajirka (commented):
@toupper Great job, the PR seems short but I'm sure this one was tricky to figure out!

I haven't tested the app yet as I left a few comments which might result in changes and a need for a re-test. Let me know what you think about them - the only one I believe we should fix is the database access on the UI thread, the others are more of a food for thought.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosIsPOSAsATabEnabled.kt
toupper: Good point, done in e0769c1e1336f5def43642005584b389a0705c69

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
toupper: That's great, I didn't know about that, thanks a lot! Done in a75db86b28059c5369883d49d9b372220db328c3

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
toupper: Indeed that works, done in a75db86b28059c5369883d49d9b372220db328c3

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
toupper: Good point, I implemented option 2 in 2aab7066cc6c0d4f5294d7c989b00dfe21bd7ef8

toupper:
Thanks for your message @jaclync!

> Nice progress with small amount of changes! From the screencasts, it looks like the POS tab is displayed right after logging in / switching stores, is it because it's shown by default, or the eligibility check results are cached for each connected site?

It was still showing the previous value, but that wasn't correct, so I implement hiding it by default here a4decdf2ac8e2985b18d157b4945199201f2407e. In next PRs we can consider caching the value as you suggested.

jaclync:
> It was still showing the previous value, but that wasn't correct, so I implement hiding it by default here [a4decdf](https://github.com/woocommerce/woocommerce-android/commit/a4decdf2ac8e2985b18d157b4945199201f2407e). In next PRs we can consider caching the value as you suggested.

Got it, I created an issue for caching the eligibility value in Android in WOOMOB-620.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -0,0 +1,95 @@
...
+    fun initialize(
+        activity: MainActivity,
+        binding: ActivityMainBinding,
+        navController: NavController
+    ) {
+        this.activity = activity
+        this.binding = binding
+        this.navController = navController
+
+        // Register this controller as a lifecycle observer
```
malinajirka: 🔍 Unnecessary comment, it says exactly the same thing as the code.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -0,0 +1,95 @@
...
+        refreshPOSTabVisibility()
+    }
+
+    override fun onDestroy(owner: LifecycleOwner) {
+        super.onDestroy(owner)
+        owner.lifecycle.removeObserver(this)
+    }
+
+    /**
+     * Initializes the POS tab with default settings and navigation handling.
```
malinajirka: 🔍 Unnecessary comment, it says exactly the same thing as the code.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private fun setupPOSTab() {
+        setPOSTabVisibility(false)
+        if (isPosAsTabEnabled()) {
+            setupPOSTabNavigation()
+        }
+    }
+
+    /**
+     * Refreshes the visibility of the POS tab based on current conditions.
```
malinajirka: 🔍 Unnecessary comment - the method name says enough imo.

malinajirka (changes requested):
@toupper Thanks for the changes, I believe the encapsulation will help with long-term maintenance. However, the performance changes seem to be affecting the behavior - there is some race condition. The Tab isn't visible after login or after (some) site switches.

I added some logs and it seems the WooIsPosEnabled doesn't have access to the site settings and therefore returns false:
```
11:44:25.734  D  WooPosTabController: POS as tab enabled, updating visibility
11:44:25.734  D  WooPosTabController: updatePOSTabVisibility - checking isWooPosEnabled
11:44:25.734  D  WooPosIsEnabled: checking if POS is enabled
11:44:25.748  D  WooPosIsEnabled: WooCommerce version 9.8.5 supports product filtering: true
11:44:25.753  D  WooPosIsEnabled: WooCommerce version 9.8.5 supports feature switch: false
11:44:25.755  D  WooPosIsEnabled: Failed - couldn't retrieve site settings
11:44:25.824  D  WooPosTabController: isWooPosEnabled returned: false
11:44:25.824  D  WooPosTabController: Setting POS tab visibility to false
11:44:25.824  D  WooPosTabController: POS menu item found: true, visibility set to false
```

It seems it currently loads the siteSettings from the local cache and if it's not there, it fails. I think we might need to either add some delay and check if the site settings is available after the delay and/or fetch the site settings if they are not available. 

This issue was likely present even before, but it was almost impossible to reproduce since the user would need to switch to `More` menu very quickly, before the siteSettings request finishes.


Patch with logs 
[Logs_for_POS_tab.patch](https://github.com/user-attachments/files/20773071/Logs_for_POS_tab.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
toupper: Thanks, removed in 4c96fc515509a02ed3bec35bf04af75ed6c9faed

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
toupper: Indeed, removed in 4c96fc515509a02ed3bec35bf04af75ed6c9faed

toupper:
> It seems it currently loads the siteSettings from the local cache and if it's not there, it fails. I think we might need to either add some delay and check if the site settings is available after the delay and/or fetch the site settings if they are not available.
> 
> This issue was likely present even before, but it was almost impossible to reproduce since the user would need to switch to More menu very quickly, before the siteSettings request finishes.

Thanks @malinajirka, good catch and debug! I fixed that here 10853d021a21b9a0a6c0c6826c3aebdd25465b9b by indeed fetching the settings remotely if the local version fails. Let me know what you think.

malinajirka (approved):
LGTM, thanks!

PR #14184, opened 2025-06-12
[WOOMOB-597] Fix NPE when scanning variations

malinajirka:
WOOMOB-597
### Description

This PR fixes NPE which might happen when the user adds a variation using a barcode scanner - given the variation product is not in cache.

I tried couple solutions but I ended up with the following approach:
- Always fetch both the Variation Model + parent Product model (unless they are already in cache of course)

### Steps to reproduce

1. Change PRE_POPULATION_PAGE_SIZE in WooPosProductsDataSource to 1. Also make sure the variable product is not loaded on the first page of the list.
2. Scan a barcode that is assigned to a variation.
3. Notice the app crashes

### Testing information

Repeat the above steps and notice the variation is successfully added to the cart.

The above


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierRemote.kt
```diff
@@ -303,6 +291,58 @@ class WooPosSearchByIdentifierRemote @Inject constructor(
         searchByGlobalUniqueIdContinuations.clear()
         coroutineScope.cancel()
     }
+
+    @Suppress("ReturnCount")
+    private suspend fun getOrFetchVariationAndUpdateCache(variationId: Long, parentId: Long): ProductVariation? {
+        val cachedVariation = variationsCache.get(variationId)?.find { it.remoteVariationId == variationId }
```
kidinov, line 297: Do we need this? I think it's handled in the `WooPosSearch...Local`

malinajirka: Good question - I removed it at one point but than decided to revert it, since I felt like having both variation and product use the same code structure might be better. I also thought that we might want to extract both of these into use-cases in which case I believe they should handle both local and remote, wdyt?

kidinov (approved):
Looks and tests good. I'll add a ticket to refactor this and add better error handling

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierRemote.kt
kidinov: If we refactor this in something different then yes, but now we kind of have 2 use cases already
* Search in local data
* Search in the remote

Now it adds local search to the remote one which is redundant and goes slightly off with the original idea

malinajirka: I wouldn't say it goes against the original idea:
- Remote search - hits the remote search endpoint (the fact that it checks cache, after it receives weird Product with variable type result from remote search endpoint doesn't change its idea IMO)
- Local search - doesn't hit remote search endpoint.

Having said that, feel free to remove this line if you believe is cleaner.

PR #14185, opened 2025-06-12
[POS as a Tab] Remove POS entry point from Menu

toupper:
**~Do not merge label - the parent PR https://github.com/woocommerce/woocommerce-android/pull/14183 needs to be merged first.~**

…re flag is enabled.

Closes WOOMOB-596

### Description

With this PR we hide the POS entry point in the menu if the POS as a Tab is Feature Flag is enabled, as we won't be needing that anymore: the POS can accessed through a tab.

### Steps to reproduce

With the `WOO_POS_AS_A_TAB_I1` feature flag enabled, and a site eligible for POS, go to menu and check that the POS entry point is not displayed. Check also that when the `WOO_POS_AS_A_TAB_I1` feature flag is disabled we show the entry point as before.


malinajirka (approved):
LGTM! 🚢 


Note: It seems this PR targets another PR, so we likely don't want to merge it before the parent PR is merged. In those cases, we usually leave a comment at the top of the PR description and add `do not merge` label to prevent us from accidentally merging it before it's ready to go.

toupper:
Thanks for the review @malinajirka 

> Note: It seems this PR targets another PR, so we likely don't want to merge it before the parent PR is merged. In those cases, we usually leave a comment at the top of the PR description and add do not merge label to prevent us from accidentally merging it before it's ready to go.

Awesome, I didn't know that, thanks for the hint 👍

PR #14192, opened 2025-06-13
[WOOMOB-601] Make Barcode info dialog scrollable

malinajirka:
### Description

Makes the barcode info dialog scrollable so it works even with increased font size.

### Testing information

1. Open POS
2. Tap on More menu
3. Tap on Barcode scanning
4. Increase font size
5. Notice the dialog is scrollable

The above

### Images/gif

https://github.com/user-attachments/assets/adff904a-4ab6-47c1-984b-50777df8f0b1


kidinov (approved):
LGTM! Also, it makes it possible to read it all on the smallest screens

image

PR #14194, opened 2025-06-13
[WOOPRD-572][Woo POS][Barcodes] Use cart layout to assign focus for scanning

kidinov:
WOOPRD-572
### Description

The PR moves the barcode scanning modifier attachment to the cart, which allows to remove quite some code.

It has one disadvantage though - as we clear the cart when the order has been successfully paid it means that the cart is in the state when it is possible to scan items again. To change that will require some extra complexity, so:
* We can close this PR
* I can do the changes to make it impossible to scan on payment success screen which maybe slightly negate the value of this PR
* Merge this PR allowing scanning on the payment success screen. It will work fine in terms of adding items to the cart, it just not something directly visible to a user

@malinajirka wdyt?

### Testing information

Validate that scanning works as before except that it is possible to scan on on the payments success screen

Above


malinajirka (commented):

> Merge this PR allowing scanning on the payment success screen. It will work fine in t erms of adding items to the cart, it just not something directly visible to a user

I'd personally vote against this one, since it's extremely likely this will be at some point reported as a bug. Fixing it at that point will be more expensive than fixing/not-introducing it now.


> We can close this PR
> I can do the changes to make it impossible to scan on payment success screen which maybe slightly negate the value of this PR

Between these two, it's a tough choice. I like the smaller code-footprint, however, from a conceptual point of view, I think scanning belongs more to the `item selection section` than the cart. I'd also say finding the code responsible for scanning can be tricky for future developers, so perhaps having the extra event will ultimately increase efficiency even though it's more code.

With all that said, I'd personally vote for keeping the code as is. Having said that, if you prevent items from being added to the cart, I'm completely fine merging it.

kidinov:
@malinajirka 👋

> I'd personally vote against this one, since it's extremely likely this will be at some point reported as a bug. Fixing it at that point will be more expensive than fixing/no-introducing it now.

I don’t believe this situation is entirely black and white. For instance, a merchant can use the scanner, hear the “beeep” sound, move the items while a customer can put all the items in their bags. They might scan all 20 (or whatever) items only to realize that the merchant forgot to click on the “new order” button after the previous customer. Then, they click “new order” and become quite frustrated that nothing was actually added to the cart. They’ll have to rescan everything again, removing all the items from the bags, which probably makes them very unhappy about the POS. 😀

I believe this scenario is quite plausible and could potentially occur in real life. It seems that we should proactively handle this situation and automatically redirect the user to the “pre-checkout” screen if any items are scanned. Wdyt?

malinajirka:
I see what you mean, but I'd be personally surprised if this happened - with one item for sure, but I think users will be looking at the screen after they scan. That's at least what I see even in retail stores where they scan tens of items per minute - they are actually even asked to do it to avoid users replacing  barcodes with barcodes of cheaper items.

Redirecting users upon scan feels a bit off, what about playing the `scan failed sound` instead?

kidinov:
As discussed, lets close it for now!

malinajirka:
As we discussed in the follow up on Slack, other POS solutions add the item to the cart and automatically navigate to the new-order screen when the merchant scans an item on the success screen. Let's re-open this PR, handle the navigation and merge it.

kidinov:
@malinajirka 

Here is the commit that implements that 

e3e132904d8786ff80db3911c7dd571e24778a24



https://github.com/user-attachments/assets/aade85b4-99a0-4a47-b249-133df9a870e7

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -341,7 +342,13 @@ class WooPosCartViewModel @Inject constructor(
     }
 
     private fun onBarcodeScanned(barcode: String) {
+        if (_state.value.cartStatus !in listOf(EDITABLE, EMPTY)) {
+            return
+        }
+
         viewModelScope.launch {
+            childrenToParentEventSender.sendToParent(OnNewTransactionStarted)
```
malinajirka: ❓ This feels a bit off to me - when we scan a barcode we don't always start a new transaction, we might be just adding items to an existing transaction. I understand the code behaves as expected, since we ignore the event in the cart state, but it still reads wrong to me.
I think we should either rename the event or move it inside the `if` statement below. Wdyt?

malinajirka (approved):
Thanks @kidinov, I've left one comment to consider, but feel free to merge it as is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
kidinov: Good point!

Moved to send it only when the cart is empty

c493fbb2f3b513c0b0045bb45c12ba40486fc030

PR #14202, opened 2025-06-16
[WOOMOB-615] Remove mentions of GTIN in code

malinajirka:
WOOMOB-615

### Description

This PR renames all mentions of gtin to GlobalUniqueIdentifier, since gtin is just one of the standards which can be put into the `GlobalUniqueIdentifier` field. Calling it GTIN can be misleading, as one might think the value follows the format of GTIN even when it doens't.

### Testing information

Green CI is enough.

### Images/gif

No user facing changes.


kidinov (approved):
LGTM!

PR #14203, opened 2025-06-16
[WOOPRD-589] Update the user facing string for RequestCancelled state

malinajirka:
WOOPRD-589

### Description

Updates the user facing string of `request cancelled` state to `Unexpected error`.

### Testing information

I'm not aware of any flow which ends up in the `Request cancelled` state. Previously, it was possible to reproduce it due to a bug in the FluxC layer which was fixed. Either way, I believe displaying `Unexpected error` is a better user-facing text than `Request cancelled` which can be quite confusing:
1. User has no idea what request or cancellation means in this context, those feel more like developer facing words
2. User didn't request any cancellation

While displaying Unexpected error doesn't help the user much, they will at least understand they didn't do anything wrong, the app just entered an unexpected state.

### Images/gif


kidinov:
I guess we can close this PR as the error itself is removed here

https://github.com/woocommerce/woocommerce-android/pull/14204

Having said that we'd need to improve error handling and separate network errors from the backend non 200 responses

PR #14208, opened 2025-06-17
[WOOPRD-596][Woo POS][Barcodes] Remove SKU and check digit handling

kidinov:
WOOPRD-596

### Description

The PR removes search by SKU when barcode scanned. It also removes double search with and without check digit

### Steps to reproduce

* POS
* Scan a barcode
* Using flipper notice that only 1 request is being done now

### Testing information

Regression testing is needed
* Simple products
* Variations
* Variables

Above

### Images/gif

No visual changes


malinajirka (approved):
LGTM, thanks!

PR #14209, opened 2025-06-17
[WOOPRD-590][Woo POS][Barcodes] Improve error handling for barcodes

kidinov:
### Description

Adds better error handling for backend related errors. Some of the errors don't make sense at all, e.g. `INVALID_COUPON` but I guess better to show at least something relevant than nothing or not relevant

Also, the localization is not implemented if the message is empty/null. Not sure if it worth it

### Steps to reproduce

We probably can test this only by modifing `WCProductStore::searchProducts`,  `WCProductStore::getVariationByRemoteId` and `WCProductStore::fetchSingleProduct` to return specific error

E.g. 

```
            if (Math.random() > 0.5) {
                return@withDefaultContext WooResult(
                    WooError(WooErrorType.RESOURCE_ALREADY_EXISTS, original = GenericErrorType.NETWORK_ERROR, "dadada")
                )
            }
```

I didn't test all the case. Just some plus unit tests

### Images/gif

image
image


libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
```diff
@@ -329,12 +329,12 @@ protected Cache.Entry createCacheEntry(NetworkResponse response) {
     @NonNull
     private BaseNetworkError getBaseNetworkError(VolleyError volleyError) {
         // No connection
-        if (volleyError.getCause() instanceof NoConnectionError) {
+        if (volleyError instanceof NoConnectionError || volleyError.getCause() instanceof NoConnectionError) {
```
kidinov, line 332: No Connection error was not propogated because of that

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierGlobalUniqueSearch.kt
```diff
@@ -20,27 +33,50 @@ class WooPosSearchByIdentifierGlobalUniqueSearch @Inject constructor(
...
-                WooPosSearchByIdentifierResult.Error.UnknownError(
-                    "Results not found for Global Unique ID: $globalUniqueId"
+                WooPosSearchByIdentifierResult.Error.ServerError(
+                    "Empty response from server for global unique ID: $globalUniqueId"
                 )
             )
         }
     }
+
+    private fun WooError.mapError(): WooPosSearchByIdentifierResult.Error =
```
malinajirka, line 63: 🔍 We could consider adding a resource id instead of a String to ensure the string are translated. Having said that, AFAICT none of the errors with custom message should happen, so I guess it's fine to fallback to an untranslated string.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierGlobalUniqueSearch.kt
```diff
@@ -20,27 +33,50 @@ class WooPosSearchByIdentifierGlobalUniqueSearch @Inject constructor(
...
+        when (type) {
+            TIMEOUT, NO_CONNECTION -> WooPosSearchByIdentifierResult.Error.NetworkError
+            API_ERROR -> WooPosSearchByIdentifierResult.Error.ServerError(
+                message?.ifEmpty { null } ?: "API error occurred"
```
malinajirka, line 67: 🔍 We can consider replacing these with `orNullIfEmpty()`.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosProductSearchByIdentifierErrorMapperTest.kt
```diff
@@ -0,0 +1,163 @@
+package com.woocommerce.android.ui.woopos.common.data.searchbyidentifier
+
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.wordpress.android.fluxc.store.WCProductStore.ProductError
+import org.wordpress.android.fluxc.store.WCProductStore.ProductErrorType
+
+class WooPosProductSearchByIdentifierErrorMapperTest {
```
malinajirka: 🗒️  I'd personally consider returning a generic error for all the server responses we don't expect and removing the tests. The more tests we have the longer it takes to run the tests on PRs and during the release process. These cases we don't expect aren't IMO worth the CI time.

malinajirka (approved):
Thanks @kidinov!

I've left a two super small comments for consideration, feel free to merge as is.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosProductSearchByIdentifierErrorMapperTest.kt
kidinov: Removed the test class - 4c1591cc9f6c89bc3de0851e7478aa27dc546c0f

> I'd personally consider returning a generic error for all the server responses we don't expect

IMO it doesn't require us much to have this extra handling, but those messages could a hint to the merchant to point out that the issue is with their backend

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierGlobalUniqueSearch.kt
kidinov: I saw this extension, but this is from the main app code base, and considering that it's just a simple wrapper of the same I decided to use the SDK as probably not worth of extra  "code dependency"

```
fun String.orNullIfEmpty(): String? = this.ifEmpty { null }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierGlobalUniqueSearch.kt
kidinov: Yes, that was my conscious decision to not have localization for those, to reduce complexity. If the error happens (which should not), it’s probably still better than having no information at all than having information in English

PR #14210, opened 2025-06-17
[WOOMOB-538] Fix duplicated requests in order creation flow

malinajirka:
### Description

This PR makes two main changes
1. It removes unnecessary requests in the order creation flow in order to speed it up.
- The app pushes all changes to the order model to the remote. However, it does so even when the changes are coming from the backend. When the user creates a new order in the app, the app creates a draft-order send syncs with the backend which is desired. However, when a result comes back, the app identifies an ID was assigned to the order and totals/subtotals/... was assigned => it considers it as a change and pushes the changes to the backend, ignoring the fact that the changes came from backend and the backend knows about them. This PR improves this behavior by not sending this unnecessary request.

2. It adds an unclickable overlay in the order creatino when a remote request is in progress to prevent users from editing the order.
- We have been disabling buttons which aren't supposed to be editable when a request is in progress. However, this solution is extremely prone to error as we often forget to disable newly added buttons. This PR adds an unclickable overlay over the whole section to ensure the order is not editable when it's being synced.

Note: The overlay is added with 500ms delay, since there is a debounce implemented to ensure users can increase quantity of items in the cart by clicking repeatedly on the plus button.

### Testing information

- Verify the order is sent to the backend just once after each change - eg. using network inspector.
- Verify the overlay is displayed when the order is being synced.
- Verify the overlay is unclickable - clicks are ignored.
- Verify the overlay disappears when the sync finishes - in both error and success.

Test the above on both a tablet and a phone.

- I've tested the above + I tested order editing mode + dark and light mode.

### Images/gif

https://github.com/user-attachments/assets/c0112681-3131-4cec-9c8c-e536f14744a8

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
@@ -565,14 +565,12 @@ class OrderCreateEditFormFragment :
             }
             new.isUpdatingOrderDraft.takeIfNotEqualTo(old?.isUpdatingOrderDraft) { isUpdatingOrderDraft ->
                 if (isUpdatingOrderDraft) {
-                    // Show overlay with fade in animation
                     binding.orderUpdateOverlay.isVisible = true
                     binding.orderUpdateOverlay.animate()
                         .alpha(1f)
                         .setDuration(ANIMATION_DURATION)
                         .start()
                 } else {
-                    // Hide overlay with fade out animation
                     binding.orderUpdateOverlay.animate()
                         .alpha(0f)
                         .setDuration(ANIMATION_DURATION)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
@@ -561,6 +561,25 @@ class OrderCreateEditFormFragment :
                 binding.taxRateSelectorSection.isVisible = it.isShown
                 binding.setTaxRateButton.text = it.label
             }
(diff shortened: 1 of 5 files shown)


WooCommerce/src/main/res/layout/fragment_order_create_edit_form.xml
```diff
@@ -182,4 +182,17 @@
...
         app:layout_constraintTop_toBottomOf="@id/two_pane_mode_toolbar" />
 
+    <View
+        android:id="@+id/order_update_overlay"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:visibility="gone"
+        android:background="@color/order_update_overlay_background"
+        android:clickable="true"
+        android:focusable="true"
```
samiuelson, line 192:
```suggestion
        android:focusable="true"
        android:focusableInTouchMode="true"
```
💡 I think it's worth adding `focusableInTouchMode`, otherwise external keyboard can potentially be used to access the buttons.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
```diff
@@ -561,6 +563,25 @@ class OrderCreateEditFormFragment :
                 binding.taxRateSelectorSection.isVisible = it.isShown
                 binding.setTaxRateButton.text = it.label
             }
+            new.isUpdatingOrderDraft.takeIfNotEqualTo(old?.isUpdatingOrderDraft) { isUpdatingOrderDraft ->
+                if (isUpdatingOrderDraft) {
+                    // Show overlay with fade in animation
```
samiuelson: 💡 Those comments seem redundant to me because the code is easy to follow.

samiuelson (approved):
LGTM! I added a couple of nitpick suggestions.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditFormFragment.kt
malinajirka: Removed 👍

PR #14216, opened 2025-06-19
Fix for the broken trunk

kidinov:
Apparently, after conflict resolution, we got a broken trunk. Green CI is enough.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
@@ -35,7 +35,7 @@ import com.woocommerce.android.ui.payments.taptopay.isAvailable
 import com.woocommerce.android.ui.plans.domain.SitePlan
 import com.woocommerce.android.ui.plans.repository.SitePlanRepository
 import com.woocommerce.android.ui.woopos.WooPosIsEnabled
-import com.woocommerce.android.ui.woopos.tab.WooPosIsPOSAsATabEnabled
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
```diff
@@ -471,7 +469,7 @@ class MoreMenuViewModel @Inject constructor(
             doCheckAvailability(MoreMenuItemButton.Type.GoogleForWoo) { isGoogleForWooEnabled() },
             doCheckAvailability(MoreMenuItemButton.Type.Inbox) { moreMenuRepository.isInboxEnabled() },
             doCheckAvailability(MoreMenuItemButton.Type.Settings) { moreMenuRepository.isUpgradesEnabled() },
-            doCheckAvailability(MoreMenuItemButton.Type.WooPos) { isWooPosEnabled() && !isPOSAsATabEnabled() }
```
malinajirka: @kidinov Could you please elaborate on this change? How come it's safe to remove the feature flag? It was added here yesterday - https://github.com/woocommerce/woocommerce-android/pull/14185.

kidinov: It's a good question. There is no `WooPosIsPOSAsATabEnabled` class in the code base. It's probably named differently. Let me see

kidinov: @malinajirka yep, that was the naming issue. Renamed it

malinajirka (approved):
LGTM!

PR #14217, opened 2025-06-19
[WOOMOB-625][Order List] Order and product filtering is not consistent on the tablets

kidinov:
WOOMOB-625

### Description

The PR makes the filter options be full screen to match the products filters on tablets

### Steps to reproduce

* Order list -> filters -> select all the filters

### Testing information

Test both on tablets and phones

Above

### Images/gif

https://github.com/user-attachments/assets/3c40e922-cf49-4a96-a4ae-f220266d1420


irfano (approved):
Thanks for addressing this. LGTM and works as expected. 👍🏻

PR #14218, opened 2025-06-19
[POS as a Tab] Persist Value for Site

toupper:
Closes WOOMOB-620

### Description

With this PR, we persist whether the POS tab was shown for a site in the last run, so we can show it as soon as possible. This makes it easier to jump to POS faster. Read more about this idea in the original issue or this PR https://github.com/woocommerce/woocommerce-ios/pull/15753

### Steps to reproduce

1. Open a site that has POS enabled. The tab should be shown.
2. Switch to a site that doesn't have POS enabled
3. Switch to the first one. See that the tab is shown right away, without any delay.

Site switching, default values, login.

#### Before

Even if it happens quite fast, note how the tab takes a while to be shown.

https://github.com/user-attachments/assets/9aba265c-42a3-4831-b209-fd6958db7f18

#### After

https://github.com/user-attachments/assets/06399fec-cb19-49d2-ae6f-5f28c9d2fb37


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -60,13 +66,25 @@ class WooPosTabController @Inject constructor(
...
 
-    private fun updatePOSTabVisibility() {
+    private fun updatePOSTabVisibilityFromPrefs() = setPOSTabVisibility(
+        appPrefs.isPOSTabVisibleForSite(selectedSite.getSelectedSiteId())
+    )
+
+    private fun updateTabVisibilityFromRemoteAndPersist() {
         activity.lifecycleScope.launch {
-            setPOSTabVisibility(isWooPosEnabled())
+            val isWooPosEnabledValue = withContext(Dispatchers.Default) {
```
malinajirka: 🔍 I noticed `isWooPosEnabled` already switches to `IO` context, so I think this `withContext(Dispatchers.Default)` is unnecessary and adds a bit of overhead. Wdyt?

malinajirka (approved):
Thanks @toupper, the code works as expected. I left one minor suggestions.

I noticed detekt is failing and you recently asked about a pre-commit/pre-push hooks. Here is a pre-push hook I created after our conversation. It essentially runs detekt and then out-loud reads `Detekt Passed` or `Detekt Failed`, which is quite convenient if you want to switch to a different context but want to make sure you don't forget. (if you are pushing through AS, make sure to mark enable-hooks checkbox).

```
#!/bin/bash

echo "Running detekt check..."
./gradlew detektAll
EXIT_CODE=$?
if [ $EXIT_CODE -ne 0 ]; then
  echo "***********************************************"
  echo "              Detekt failed                    "
  echo " Please fix the above issues before pushing    "
  echo "***********************************************"
  say "Detekt failed" 2>/dev/null || true
  exit 1  # This will block the push
else
  echo "Detekt passed successfully!"
  say "Detekt passed" 2>/dev/null || true
  exit 0  # Allow the push to proceed
fi

```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
toupper: Thanks @malinajirka! I don't know why but the compiler was complaining about that. But once I run, it builds ok. Strange :)

toupper:
Thanks for the pre-push hook @malinajirka! That looks great, to avoid forgetting it as in this case.

PR #14221, opened 2025-06-19
[Shipping Labels] Handle adding new payment methods in the WebView

hichamboushaba:
Closes WOOMOB-438

### Description
This PR adds logic to handle adding new payment methods in the WebView, the logic is as follows:
- Tapping on the Add button will open the WebView.
- Like the old flow, when we detect redirection to the `/me/payment-methods` URL, we treat it as success.
- When we detect success, we reload the account settings to get the new payment methods.
- In the new flow, I added logic to auto-select the newly added payment method.

> [!IMPORTANT]
> I was planning to use the existing `AuthenticatedWebViewFragment`, but given the screen is shown as a `BottomSheetDialogFragment`, this wasn't possible, as after navigating to another `Fragment`, it gets dismissed automatically (a limitation to being a `DialogFragment`), I tried various solutions, and at the end I opted to use the `WebView` in `Compose`, please feel free to share any thoughts about this topic, or to suggest any other solutions. 

### Steps to reproduce
1. Disable the "WCS staging" plugin on your store to point to production servers (The mobile WebView will always point to the productions server)
2. Open the app, and start creating a new shipping label.
3. Open the payment settings.
4. Tap on "Add new payment method"
5. Add a card.

Make sure to delete the card if you don't need it.

### Testing information
- Confirm that the WebView is correctly opened.
- Confirm that after adding the card, the WebView is automatically dismissed, and we start loading the account settings.
- Confirm that the added card is automatically selected.

The above.

##### Adding first payment card

https://github.com/user-attachments/assets/2a4e2527-429e-495a-9b5e-69c8354f17f1

##### Adding additional payment cards

https://github.com/user-attachments/assets/dba643aa-7a1f-4b80-bee5-9407dfd952c5

##### Error

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
@@ -7,6 +7,9 @@ import android.view.ViewGroup
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.fragment.app.viewModels
+import androidx.navigation.fragment.findNavController
+import com.woocommerce.android.NavGraphMainDirections
+import com.woocommerce.android.ui.common.webview.AuthenticatedWebViewViewModel
 import com.woocommerce.android.ui.compose.theme.WooTheme
 import com.woocommerce.android.widgets.WCBottomSheetDialogFragment
 import dagger.hilt.android.AndroidEntryPoint
@@ -27,5 +30,30 @@ class WooShippingEditPaymentDialogFragment : WCBottomSheetDialogFragment() {
                 }
             }
         }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        handleEvents()
+    }
+
+    private fun handleEvents() {
+        viewModel.event.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is WooShippingEditPaymentViewModel.ShowPaymentMethodAddWebView -> {
+                    showWebView(event.url, event.successUrl)
+                }
+            }
+        }
+    }
+
+    private fun showWebView(url: String, exitUrl: String) {
+        findNavController(R.id.nav_host_fragment_main).navigate(
+            NavGraphMainDirections.actionGlobalAuthenticatedWebViewFragment(
+                urlToLoad = url,
+                urlsToTriggerExit = arrayOf(exitUrl),
+                urlComparisonMode = AuthenticatedWebViewViewModel.UrlComparisonMode.PARTIAL
+            )
+        )
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
@@ -7,8 +7,11 @@ import android.view.ViewGroup
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.fragment.app.viewModels
-import androidx.navigation.fragment.findNavController
 import com.woocommerce.android.NavGraphMainDirections
+import com.woocommerce.android.R
+import com.woocommerce.android.extensions.findNavController
+import com.woocommerce.android.extensions.handleNotice
+import com.woocommerce.android.ui.common.webview.AuthenticatedWebViewFragment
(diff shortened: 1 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
```diff
@@ -28,4 +37,28 @@ class WooShippingEditPaymentDialogFragment : WCBottomSheetDialogFragment() {
...
+
+    private fun disableDraggingOnWebViewState() {
+        viewModel.viewState.map { it.javaClass }
+            .distinctUntilChanged()
+            .observe(viewLifecycleOwner) {
+                val dialog = requireDialog() as BottomSheetDialog
+                dialog.behavior.isDraggable =
+                    it != WooShippingEditPaymentViewModel.ViewState.AddPaymentMethodWebView::class.java
+            }
+    }
```
hichamboushaba, line 63: When the WebView is shown, scrolling the WebView can dismiss the bottom sheet, and working on the nested scrolling behavior will be tricky given how things are organized, I opted for disabling the draggability of the bottom sheet instead, and given we force showing the WebView as full screen, I think this is fine, let me know what you think.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentViewModelTest.kt
```diff
@@ -0,0 +1,221 @@
...
+import kotlinx.coroutines.test.advanceUntilIdle
+import org.assertj.core.api.Assertions.assertThat
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.wordpress.android.fluxc.network.UserAgent
+import kotlin.test.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class WooShippingEditPaymentViewModelTest : BaseUnitTest() {
```
hichamboushaba, line 27: These tests were completely generated by Jetbrain's AI agent [Junie](https://www.jetbrains.com/junie/), while I didn't have good results with AI in the past, this is impressive, it followed my patterns to the line, I asked him to do so explicitly.

hichamboushaba:
@irfano I just noticed that the snackbar is not shown after adding the payment method, given there is another PR stacked on this PR, please ignore this part, I'll make sure to fix in a separate PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
```diff
@@ -44,33 +41,24 @@ class WooShippingEditPaymentDialogFragment : WCBottomSheetDialogFragment() {
...
-    private fun showWebView(url: String, exitUrl: String) {
-        findNavController(R.id.nav_host_fragment_main).navigate(
-            NavGraphMainDirections.actionGlobalAuthenticatedWebViewFragment(
-                urlToLoad = url,
-                urlsToTriggerExit = arrayOf(exitUrl),
-                urlComparisonMode = AuthenticatedWebViewViewModel.UrlComparisonMode.PARTIAL
-            )
-        )
+    private fun disableDraggingOnWebViewState() {
+        viewModel.viewState.map { it.javaClass }
```
irfano, line 56: I couldn't understand why we're mapping to `javaClass`? Wouldn’t simply observing `viewState` achieve the same result? Like this:

```
   viewModel.viewState
            .distinctUntilChanged()
            .observe(viewLifecycleOwner) {
                val dialog = requireDialog() as BottomSheetDialog
                dialog.behavior.isDraggable = it !is WooShippingEditPaymentViewModel.ViewState.AddPaymentMethodWebView
            }
```

irfano (approved):
Good job! LGTM! I just added a minor comment.

irfano:
> @irfano I just noticed that the snackbar is not shown after adding the payment method, given there is another PR stacked on this PR, please ignore this part, I'll make sure to fix in a separate PR.

Displaying the card after adding it is also not testable in this PR; I tested it in #14227.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentDialogFragment.kt
hichamboushaba: The above would mean that we'll invoke the observer each time a state **content** change happens (for example when the user changes the selected card, or toggles the email receipts option), while when mapping to the javaClass before invoking `distinctUntilChanged`, we then won't invoke the observer until it's necessary: when the **type** of the state changes.

PR #14223, opened 2025-06-20
[WOOPRD-603][Woo POS][Barcodes] Home as focus listener for barcodes scanning

kidinov:
### Description

WOOPRD-603

When I moved the barcode focus listener to the cart, we got into the situation when another, not child view takes focus, it's not getting back; therefore no keyboard input sent to the cart.

Moving it back to the items sub-screen also is not a good idea, as if anything that takes focus in the cart will be added, we would have the same problem

This PR moves it to the home screen, so all the keyboard input goes through it

### Steps to reproduce

* Scan a product – works as expected
* Tap search, and leave search (even without doing a search)
* Scan a product – this scan is ignored, and all future scans are also ignored until I quit and re-open POS.

### Testing information

* Try that scanning works after/during something in the focus on the cart screen
* Do checkout, go back
* Do payment, go back
* Notice that the scanning still works

Above

### Images/gif

https://github.com/user-attachments/assets/00b210aa-8a80-4fc9-b813-115129372ad1


malinajirka (approved):
LGTM, thanks!

PR #14224, opened 2025-06-20
[WOOMOB-626] Simplify Barcode Detection Logic

malinajirka:
WOOMOB-626

  ### Description
  Simplified the BarcodeInputDetector implementation to remove unnecessary complexity. The changes implement the agreed-upon requirements:
  - Remove the 1.5s timeout for barcode scanning
  - Only process scans with \r or \n terminator characters
  - Allow maximum 200ms between keystrokes (increased from 100ms)

  The logic is now much cleaner and easier to understand - instead of complex state management with multiple helper methods, we simply check at the beginning of each input if too much time has passed and clear the buffer
  if needed.

  ### Testing information
  Smoke test barcode scanning

I have smoked tested barcode scanning. I also tested entering characters using a hardware keyboard - with 200ms delay and without, all worked fine.

  ### Images/gif
  N/A - Internal code simplification

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
@@ -5,7 +5,6 @@ import androidx.compose.foundation.focusable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.composed
 import androidx.compose.ui.focus.FocusRequester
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
```diff
@@ -68,78 +62,38 @@ fun Modifier.listenForBarcodes(
 @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
 class BarcodeInputDetector(
     private val onBarcodeScanned: (String) -> Unit,
-    private val coroutineScope: CoroutineScope,
     private val currentTimeProvider: CurrentTimeProvider,
 ) {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
kidinov: Btw, it's probably not needed. I don't see that you use or even should use anything in the tests from here

kidinov (approved):
Looks and tests good. I didn't test on a real scanner though, I was using adb:

```
adb shell input text "21321313123123411"
adb shell input keyevent 66
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/modifier/BarcodeScannerModifier.kt
malinajirka: Good point, originally I was accessing it in tests instead of hardcoding `250` and similar values. However, I agree it's likely an overkill. Removed in https://github.com/woocommerce/woocommerce-android/pull/14224/commits/92da21c60d67eb291ca3c8cbb0cf3123b64196cf.

kidinov: > I was accessing it in tests instead of hardcoding 250 and similar values

By the way, I believe using hardcoded values in the tests makes more sense because then you can actually check the value itself, not the reference. This way, the test will detect if the value in the constant has changed mistakenly, while if a reference is used, it won’t

PR #14225, opened 2025-06-20
[POS as a Tab] Analytics

toupper:
### Description

With this PR we add analytics support for the first iteration of the POS as a Tab project, as well as fix the callback logic for the tab bar that was broken through https://github.com/woocommerce/woocommerce-android/pull/14183

#### Broken callback logic

With `WooPosTabController`, we had overridden the entire BottomNavigation listener, which caused some methods in MainActivity.kt not to be called as expected. To fix this, I removed the listener override from `WooPosTabController` and now invoke that logic directly from `MainActivity.kt's onNavItemSelected.`

To support this change, I also made a slight adjustment to the logic in `MainBottomNavigationView` to handle the POS mode activation, which doesn't follow the same pattern as the other tabs. See the in-code comment for details.

Since this logic was tightly coupled with analytics, I decided to include the fix in the analytics PR—otherwise, analytics wouldn't function correctly. I hope that's okay.
### Steps to reproduce

Prerequisite: a WPCOM account that has at least two connected stores, one eligible for POS and the other one not eligible.

- Log in to the account in the prerequisite with the store that is **eligible for POS** --> after the POS tab is shown, an event should be tracked ` 🔵 Tracked: woocommerceandroid_pos_tab_visibility_checked, Properties: {"is_visible":"true"...]`
- Tap on the POS tab --> it should launch POS in full-screen with an event tracked `🔵 Tracked: woocommerceandroid_main_tab_pos_selected`.
- Exit POS
- Go to Menu tab and switch to a store **ineligible for POS** --> after switching stores, the POS tab should not appear even after a while. An event should be tracked `🔵 Tracked pos_tab_visibility_checked, properties: [is_visible: false, ...]`

### Testing information

I also verified that the rest of the main bottom navigation logic continues to work as expected.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -769,8 +780,11 @@ class MainActivity :
             ORDERS -> AnalyticsEvent.MAIN_TAB_ORDERS_RESELECTED
             PRODUCTS -> AnalyticsEvent.MAIN_TAB_PRODUCTS_RESELECTED
             MORE -> AnalyticsEvent.MAIN_TAB_HUB_MENU_RESELECTED
+            else -> null
```
toupper: The POS tab cannot be reselected.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainBottomNavigationView.kt
```diff
@@ -135,11 +135,8 @@ class MainBottomNavigationView @JvmOverloads constructor(
 
     override fun onNavigationItemSelected(item: MenuItem): Boolean {
         navController?.let { navController ->
-            val navSuccess = NavigationUI.onNavDestinationSelected(item, navController)
```
toupper, line 138: I don't think it's necessary to call `listener.onNavItemSelected` only when `onNavDestinationSelected` is successful. The listener's name suggests that it's triggered when an item is selected, regardless of whether navigation occurred. We need to adjust this for the POS tab, where navigation follows a custom flow and doesn't succeed through the usual mechanism.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
malinajirka: 💡 I'd consider explicitly listing `POS` instead of relying on `else`. 
If someone introduces a new BottomNavigationPosition, they might forget to update this `when`, however, if the `when` doesn't have `else` statement, the compiler will inform them about their mistake.

malinajirka (approved):
Thanks @toupper! The changes look good to me. 

I've noticed the event is tracked even on phones (with is_visible=false). I assume that is expected and desired, but just noting it to be sure. 

🚢

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
toupper: Good point! Done in 10aff47abc5106337ff9af0746b4fac641f80b96

PR #14227, opened 2025-06-20
[Shipping Labels] Handle reading permissions and emailReceipts options from the network + other changes

hichamboushaba:
Closes WOOMOB-436

### Description
This PR has two main changes:
- Exposes more fields from the `/account/settings` response to handle in the ViewMode, those include the permissions settings, and the store owner details and the `emailReceipts` option.
- Add logic for the `emailReceipts` toggle.

I also did some refactoring which I'll explain below in the comments.

> [!NOTE]
> Saving is still not implemented. 

### Steps to reproduce
1. Open the shipping label creation form.
2. Open payment bottom sheet.

### Testing information
- Confirm that the payment options are enabled for store owner, and disabled for other users.
- Confirm that the footer caption mentions the correct store owner name and WPCom username.
- Toggle the emailReceipts and confirm it works.

The above.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentScreen.kt
@@ -392,7 +392,7 @@ private fun PaymentMethodsList(
             text = stringResource(R.string.woo_shipping_payment_email_receipt_toggle),
             checked = viewState.emailReceipts,
             enabled = viewState.canEditSettings,
-            onCheckedChange = { TODO() },
+            onCheckedChange = viewState.onEmailReceiptsChanged,
             modifier = Modifier.fillMaxWidth()
         )
         Spacer(Modifier.height(16.dp))
@@ -506,7 +506,8 @@ private fun ContentScreenEmptyPreview() {
                 storeOwnerName = "John Doe",
                 storeOwnerUsername = "johndoe",
                 selectedPaymentMethodId = null,
-                currentPaymentOptions = ShippingLabelSampleData.getPaymentOptions(countOfPaymentMethods = 0)
+                currentPaymentOptions = ShippingLabelSampleData.getPaymentOptions(countOfPaymentMethods = 0),
+                onEmailReceiptsChanged = {}
             ),
             onAddNewPaymentMethod = {},
             onPaymentMethodSelected = {},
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentScreen.kt
```diff
@@ -390,15 +378,21 @@ private fun PaymentMethodsList(
...
-            onClick = onSaveClicked,
-            text = stringResource(R.string.woo_shipping_payment_use_card_button),
+            onClick = viewState.onSaveClicked,
+            text = stringResource(
+                if (viewState.canManagePaymentMethods) {
+                    R.string.woo_shipping_payment_use_card_button
+                } else {
+                    R.string.save
+                }
+            ),
```
hichamboushaba, line 387: When the user doesn't have permissions to manage payments, this button used to say "Use this card" like this:



This didn't make sense as the only thing the user could modify is the `emailReceipts` option, so I updated it to say "Save" in this context.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/payment/WooShippingEditPaymentViewModel.kt
```diff
@@ -172,11 +192,15 @@ class WooShippingEditPaymentViewModel @Inject constructor(
...
+            val emailReceipts: Boolean,
             val selectedPaymentMethodId: Int?,
             val storeOwnerName: String,
             val storeOwnerUsername: String,
-            val currentPaymentOptions: PaymentMethodOptions
+            val currentPaymentOptions: PaymentMethodOptions,
+            val onAddNewPaymentMethod: () -> Unit,
+            val onPaymentMethodSelected: (Int) -> Unit,
+            val onEmailReceiptsChanged: (Boolean) -> Unit,
+            val onSaveClicked: () -> Unit
```
hichamboushaba, line 203: I moved the callbacks to the ViewState, I like this pattern when possible (I mean when the ViewState is not `Parcelable` like this case for us), as this allows us to limit the callbacks to the correct state, for example for this screen, calling `onPaymentMethodSelected` or the other callbacks when the state is `Loading` or `AddPaymentMethodWebView` is a logic issue, and there was no protection against it, while now it's not possible.

irfano (approved):
LGTM! I just couldn’t test whether the email was sent successfully since the purchase endpoint isn’t working at the moment. All other parts work as expected. 👍🏻

irfano:
I noticed that the payment section is editable for purchased shipments. Since we don’t allow editing address fields for purchased shipments, we should follow the same rule for payment methods as well. Feel free to include this in this PR or in any follow-up PRs.

hichamboushaba:
> I noticed that the payment section is editable for purchased shipments. Since we don’t allow editing address fields for purchased shipments, we should follow the same rule for payment methods as well. Feel free to include this in this PR or in any follow-up PRs.

Yes, this part is not included yet, it's part of a different issue WOOMOB-642

PR #14232, opened 2025-06-23
`WCUserModel` to `Room`

wzieba:
Closes: [AINFRA-566](https://linear.app/a8c/issue/AINFRA-566/migrate-wcusermodel)

### Description

This PR migrates `WCUserModel` from `WellSql` to `Room`.

### Testing information

Install the app from this PR, run `App Inpsection` and check `UserEntity` value - I think it should be enough for this PR.

### Images/gif

Screenshot 2025-06-23 at 14 50 54


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserModel.kt
```diff
@@ -1,43 +1,19 @@
...
-import com.yarolegovich.wellsql.core.annotation.Table
-import org.wordpress.android.fluxc.persistence.WellSqlConfig
+import androidx.room.Entity
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 
-@Table(addOn = WellSqlConfig.ADDON_WOOCOMMERCE)
-@RawConstraints(
-        "FOREIGN KEY(LOCAL_SITE_ID) REFERENCES SiteModel(_id) ON DELETE CASCADE",
-        "UNIQUE (EMAIL, LOCAL_SITE_ID) ON CONFLICT REPLACE"
```
ParaskP7, line 15: Question (❓): I wonder why this constraint existed, it does seems better to me to have `localSiteId` and `remoteUserId` as the primary key, but still, I wonder, was it just because of `getUserBySiteAndEmail(...)`... 🤷

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/UserDao.kt
```diff
@@ -0,0 +1,22 @@
+package org.wordpress.android.fluxc.persistence.dao
+
+import androidx.room.Dao
+import androidx.room.Query
+import androidx.room.Upsert
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.user.WCUserModel
+
+@Dao
+abstract class UserDao {
```
ParaskP7, line 10: Warning (⚠️): Just so that we keep testing quality high, please consider adding an associated `UserDaoTest` that replaces the deleted `WCUserSqlUtilsTest`. 🙏

ParaskP7 (changes requested):
👋 @wzieba  !

I have reviewed and tested this PR as per the instructions, everything works as expected, good job, minus one table from our list! 🌟 

-----

I have left one test related warning (⚠️) and a question (❓) for you. I am **NOT** going to approve this PR for now, just because of this warning (⚠️), but, nothing serious. 🙏

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/UserDao.kt
wzieba: I didn't introduce tests for this class intentionally and with the high quality in mind: `UserDao` doesn't have any logic whatsoever. Introducing tests here would be testing the database framework, and we shouldn't do this.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserModel.kt
wzieba: Good catch, but I'm not sure how to proceed. If we both agree that `localSiteId` x `remoteUserId` is the primary key, then I don't think we should introduce a different one.

ParaskP7: Nah, I am on your side @wzieba, I am just not sure we're missing anything, probably not, just wanted to note that, in case you could think anything we might be missing... 🤷

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/UserDao.kt
ParaskP7: I am a bit confused @wzieba, and I do understand where you're coming from, still, on all the previous such migration PRs we merged, we did add an associated DAO test, what changed? 🤔

> UserDao doesn't have any logic whatsoever.

Again, I understand what you say, but maybe this is not a yes or no answer, rather it an it depends. 🤔

- You could test `upsertUser` and then `getUserBySiteAndEmail` using a different email seeing that it will give you a `null` result, or you could use the same email to the user you inserted and it will again give you a `null` result, just because you made a mistake in the query on that DAO, in which case the test will fail.
- Then, you could test `upsertUser` twice, but instead of the `@Upsert` annotation you mistakenly used `@Insert`, in which case the test should fail, trying to add the same row constraint by primary key twice.

Someone might consider that logic... 🤷

---

PS: I really thought this was an oversight, thus I blocked this PR with a warning (⚠️). If you really okay with merging this without an associated DAO test, or you think that those DAO tests are actually not giving us any benefit, let's do it and continue as such with the rest of the migrations. Maybe even removing the existing DAO tests to help convey this message. I just wanted clarity on what process we should be follow from here onwards.

ParaskP7:
FYI: I am now approving this to unblock you ([context](https://github.com/woocommerce/woocommerce-android/pull/14232#discussion_r2163962179)).

ParaskP7 (approved):

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/UserDao.kt
wzieba: Good arguments: I've introduced tests in 9fb046c 👍

ParaskP7: Awesome, thanks for adding this test suite @wzieba ! 🙇 ❤️ 🚀

PS: Feel free to merge this! ✅ 

---

FYI: I would personally use `sampleUser.email` instead of `val userEmail = "user@example.com"` on those 3 x tests that you did create this `val`, but I am sure you did that on-purpose, and I would love to know why... 😊

wzieba: I thought it was more readable, but actually I agree `sampleUser.email` looks better, thanks!

ParaskP7: FYI: To me, it is not actually about readability, but rather making sure to provide the same `email` during the `then` phase, same to the `sampleUser` being upserted, similar to what you did on that 3rd out of 4 tests there. The next alternative would have been to use `SAMPLE_USER_MAIL` instead, since this is the email that `sampleUser` gets instantiated with. 🤔

PR #14236, opened 2025-06-24
[POS as a Tab] Remove i1 Feature Flag and Unnecessary Code

toupper:
### Description

With this PR we remove the POS as a Tab i1 feature flag and the unnecessary code, mostly relate to the POS entry in the menu.

### Steps to reproduce

Check that the tab is properly shown for eligible stores and do a quick general testing in the More Menu to ensure that it's still working.


malinajirka (approved):
LGTM, thanks! 🚢

PR #14247, opened 2025-06-26
[WOOPRD-612][Woo POS][Barcodes] Barcode scanned tracking event 

kidinov:
WOOPRD-612

### Description

The PR adds tracking of the tracking of

```
            override val name: String = "barcode_scanned"

            init {
                addProperties(
                    mapOf(
                        "barcode_length" to barcodeLength.toString(),
                        "scan_duration_ms" to scanDurationMs.toString(),
                        "is_numeric_only" to isNumericOnly.toString(),
                        "scanner_info" to (scannerInfo ?: "unknown")
                    )
                )
            }
        }
```

### Steps to reproduce

* Use logcat
* Scan a barcode in the POS
* Notice tracking:

```
🔵 Tracked: woocommerceandroid_pos_barcode_scanned, Properties: {"barcode_length":"12","scan_duration_ms":"44","is_numeric_only":"false","scanner_info":"BSH209900679-(BLUETOOTH)","blog_id":198349371,"is_wpcom_store":true,"was_ecommerce_trial":false,"plan_product_slug":"business-bundle","store_id":"534e18db-8439-411e-9dfc-467cf73a450f","is_debug":true,"site_url":"https:\/\/paymentwithoutaddress.wpcomstaging.com","cached_woo_core_version":"9.9.4"}
```

### Testing information

* We'd need to retest it on other barcode scanners when the folks in the team will have them
* The scanner info won't be there if BT permission is not given

I tested it only with my BSH scanner. The scanner detection code is largely LLM based


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/util/WooPosScannerDetectionUtil.kt
```diff
@@ -0,0 +1,150 @@
...
+import android.bluetooth.BluetoothDevice
+import android.bluetooth.BluetoothManager
+import android.content.Context
+import android.hardware.input.InputManager
+import android.view.InputDevice
+import androidx.annotation.RequiresPermission
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
malinajirka: 🔍 Singleton annotation should IMO be used only when we need to guarantee there will be just one instance. I believe in this case @Reusable might be more appropriate.  Tbh in this particular case, I don't see a reason why to have either of those, so no-scope would IMO be the best solution (solution with the lowest performance/memory cost).

malinajirka (approved):
Thanks @kidinov!

I've tested it without and with BT permissions and it all seems to be working as expected. I've left one minor suggestion.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/util/WooPosScannerDetectionUtil.kt
kidinov: Good catch. That was not on purpose. Removed

a67b48bfa55ce0100326605a9738969b1e200083

PR #14251, opened 2025-06-26
`WCSettingModel` to `Room`

wzieba:
Closes: [AINFRA-561](https://linear.app/a8c/issue/AINFRA-561/migrate-wcsettingsmodel-wcsettingsbuilder)

### Description

This PR migrates `WCSettingsModel` from `WellSql` to `Room`.

Furthermore, it adjusts naming to the convention of the project and introduces a novelty: `internal`-only database model.

Database models should be `internal` for `fluxc` module, as this is the, let's say, `data` layer for the app. Usually, it's not possible as domain modules live in `WooCommerce` module, e.g. `com.woocommerce.android.model.Product`, so they can't be mapped in FluxC - therefore, their db counterparts has to be `public`.

In this PR, the `org.wordpress.android.fluxc.model.settings.Settings` domain model, is placed in FluxC module, making it possible to make `WCSettingsModel` `internal`.

In the future, I think it'd be best if we moved `Settings` to another, domain-only module so both WooCommerce and FluxC can share the same models, but for now I think it's a good, incremental solution.

### Testing information

1. On the web, go to `WooCommerce` > `Settings` > `General` > `Currency options`
2. Change settings and reopen the mobile app back and forth. Make sure the mobile app reflects the changes (usually you'll have to kill the app and run it again).

### Images/gif

Screenshot 2025-06-27 at 08 48 39


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/settings/CurrencyPosition.kt
```diff
@@ -0,0 +1,5 @@
+package org.wordpress.android.fluxc.model.settings
+
+enum class CurrencyPosition {
```
wzieba, line 3: `CurrencyPosition` is not inherently connected to "settings" - it can exist without them, hence extraction to a separate file.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/items/coupons/CachedCouponEnabledCheckerTest.kt
```diff
@@ -38,9 +37,7 @@ class CachedCouponEnabledCheckerTest {
     @Test
     fun `returns true when coupons are enabled in site settings`() = runTest {
         // GIVEN
-        val siteSettings = mock<WCSettingsModel> {
-            on { couponsEnabled } doReturn true
-        }
```
wzieba, line 43: There's no point in mocking data models.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -200,15 +203,15 @@ open class WooCommerceStore @Inject constructor(
     /**
      * Given a [SiteModel], returns its WooCommerce site settings, or null if no settings are stored for this site.
      */
-    fun getSiteSettings(site: SiteModel): WCSettingsModel? =
-        WCSettingsSqlUtils.getSettingsForSite(site)
+    fun getSiteSettings(site: SiteModel): Settings? =
+        runBlocking { settingsDao.getSettings(site.localId())?.let { WCSettingsMapper.mapToDomain(it) } }
```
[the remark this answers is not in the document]
wzieba: As in all other migrations, it's expected - the project is about migration, not fixing the threading issues 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/CurrencyPositionConverter.kt
```diff
@@ -0,0 +1,17 @@
...
+class CurrencyPositionConverter {
+
+    @TypeConverter
+    fun fromCurrencyPosition(value: CurrencyPosition): String {
+        return value.name
+    }
+
+    @TypeConverter
+    fun toCurrencyPosition(value: String): CurrencyPosition {
+        return CurrencyPosition.valueOf(value.uppercase())
```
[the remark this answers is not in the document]
wzieba: As we control what we save to the database, I can't say it's a likely scenario. Nevertheless, I've decided to address this to be completely safe in case of future development over `CurrencyPosition`. Addressed in: c03d1fef0321afa8ae7811d6f2454d01bc151498

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
```diff
@@ -24,8 +25,10 @@ import javax.inject.Singleton
 )
 interface WCDatabaseModule {
     companion object {
-        @Singleton @Provides fun provideDatabase(context: Context): WCAndroidDatabase {
-            return WCAndroidDatabase.buildDb(context)
+        @Singleton
+        @Provides
+        fun provideDatabase(context: Context, currencyPositionConverter: CurrencyPositionConverter): WCAndroidDatabase {
```
ParaskP7, line 30: Question (❓): I read the commit on the converter crashing, but why was this change necessary? 🤔

I don't see any other such converter initialize this way, or (kind of) twice, as you already added it into the list of `@TypeConverters`. Please help me understand because I think I am missing an important piece of the puzzle. 🙏

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/CurrencyPositionConverter.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import androidx.room.ProvidedTypeConverter
+import androidx.room.TypeConverter
+import org.wordpress.android.fluxc.model.settings.CurrencyPosition
+import org.wordpress.android.fluxc.utils.AppLogWrapper
+import org.wordpress.android.util.AppLog
+import javax.inject.Inject
+
+@ProvidedTypeConverter
+class CurrencyPositionConverter @Inject constructor(
```
ParaskP7, line 11: Praise (❤️): Nicely done! 🥇

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SettingsDao.kt
```diff
@@ -0,0 +1,24 @@
...
+        SELECT * FROM SettingsEntity
+        WHERE localSiteId = :siteId
+    """
+    )
+    suspend fun getSettings(siteId: LocalId): WCSettingsModel?
+
+    @Upsert
+    suspend fun upsertSettings(settings: WCSettingsModel)
+
+    @Query("UPDATE SettingsEntity SET couponsEnabled = :couponsEnabled WHERE localSiteId = :siteId")
```
ParaskP7, line 22: Praise (❤️): This is our first such `UPDATE` query in this whole project, interesting and nicely done! 🥇

PS: I guess otherwise we would need to do the following instead, right?

```kotlin
                    response.result?.let { couponsEnabled ->
                        val settings = settingsDao.getSettings(site.localId())
                        settings?.let { settingsDao.upsertSettings(settings.copy(couponsEnabled = couponsEnabled)) }
                        couponsEnabled
                    } ?: false
```

For which, I am sure we have lot of :this: above throughout our codebase that can benefit form such an `UPDATE` query, thus mentioning it... 💭

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -14,12 +15,12 @@ import org.wordpress.android.fluxc.annotations.action.Action
...
-import org.wordpress.android.fluxc.model.WCSettingsModel
-import org.wordpress.android.fluxc.model.WCSettingsModel.CurrencyPosition.LEFT
-import org.wordpress.android.fluxc.model.WCSettingsModel.CurrencyPosition.LEFT_SPACE
-import org.wordpress.android.fluxc.model.WCSettingsModel.CurrencyPosition.RIGHT
-import org.wordpress.android.fluxc.model.WCSettingsModel.CurrencyPosition.RIGHT_SPACE
 import org.wordpress.android.fluxc.model.plugin.SitePluginModel
+import org.wordpress.android.fluxc.model.settings.CurrencyPosition.LEFT
+import org.wordpress.android.fluxc.model.settings.CurrencyPosition.LEFT_SPACE
+import org.wordpress.android.fluxc.model.settings.CurrencyPosition.RIGHT
+import org.wordpress.android.fluxc.model.settings.CurrencyPosition.RIGHT_SPACE
```
ParaskP7, line 22: Minor (🔍): I know we've discussed this and probably I care more than needed about these imports, but I see on some classes we just kept this single `CurrencyPosition` import, while on other classes we kept repeating that import per enum. Plus, losing the extra `CurrencyPosition` information during usage. Maybe we should consider using a single `CurrencyPosition` import everywhere (else just ignore me). 😊

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/persistence/TestDatabase.kt
```diff
@@ -0,0 +1,15 @@
+package org.wordpress.android.fluxc.persistence
+
+import android.content.Context
+import androidx.room.Room
+import androidx.room.RoomDatabase
+import org.mockito.Mockito
+import org.wordpress.android.fluxc.persistence.converters.CurrencyPositionConverter
+
+object TestDatabase {
```
ParaskP7: 1. Praise (❤️): Irrespective your need to not repeat adding the `CurrencyPositionConverter` for every test, I think this is quite a nice addition to help keep such test setup to minimum, nicely done! 🥇
2. Suggestion (💡): Why not include the actual `.build()` be part of this? This will avoid having to type `.build()` for every test, also one less import saved. Then, instead of `provideTestDatabase(...)` we rename it to `buildTestDatabase(...)`.
3. Suggestion (💡): Now that you updated all tests, why not doing the below instead of having that setup as part of the `@Before` loop?

Instead of:
```kotlin
    private lateinit var db: WCAndroidDatabase

    @Before
    fun createDb() {
        val context = ApplicationProvider.getApplicationContext<Application>()
        db = TestDatabase.provideTestDatabase(context).build()
        ...
    }

    ...
}
```

Do:

```kotlin
class XyzDaoTest {
    val context = ApplicationProvider.getApplicationContext<Application>()
    val db = TestDatabase.provideTestDatabase(context)

    ...
}
```

This way the `@Before` related `setup()` test function would be much more focused on the actual test setup required for this test, which is everything related to `sut` and not initializing the actual in-memory test database. 

---

Now that I am thinking about it, we could even create an abstract base `DaoUnitTest` which will further reduce all this noise. Wdyt?

```kotlin
open class DaoUnitTest {
    private val context = ApplicationProvider.getApplicationContext<Application>()
    val db = Room.inMemoryDatabaseBuilder(context, WCAndroidDatabase::class.java)
        .addTypeConverter(CurrencyPositionConverter(Mockito.mock()))
        .allowMainThreadQueries()
        .build()

    @Rule
    @JvmField
    val databaseTestRule = DatabaseTestRule(db)
}

class DatabaseTestRule(val db: WCAndroidDatabase) : TestWatcher() {
    override fun finished(description: Description?) {
        super.finished(description)
        db.close()
    }
}
```

Look at all the de-duplication and code savings, a much better/focused test, no? 🤔

PS: With the above you no longer need the `TestDatabase` object.

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
```diff
@@ -1,13 +1,38 @@
...
 import org.wordpress.android.fluxc.network.discovery.RootWPAPIRestResponse
 import org.wordpress.android.fluxc.network.discovery.RootWPAPIRestResponse.Authentication
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.SiteSettingOptionResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.SiteSettingsResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.system.WooSystemRestClient.SSRResponse
+import org.wordpress.android.fluxc.persistence.entity.WCSettingsModel
 
 object WCSettingsTestUtils {
+
+    internal fun generateSettingsModel(siteId: LocalId) = WCSettingsModel(
```
ParaskP7, line 16: Question (❓): I am getting this error-ish 🔴 warning on the `SettingsDaoTest` import on that, don't you get the same? 🤔

`Cannot access 'fun generateSettingsModel(siteId: LocalOrRemoteId.LocalId): WCSettingsModel': it is internal in 'org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils'.`

image

PS: The test still run as expected... 🤷

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
```diff
@@ -1,13 +1,38 @@
...
+        countryCode = "US",
+        stateCode = "CA",
+        address = "123 Main St",
+        address2 = "Apt 4",
+        city = "San Francisco",
+        postalCode = "94105",
+        couponsEnabled = false
+    )
+
+    fun generateSettings(siteId: LocalId) = generateSettingsModel(siteId).let {
```
ParaskP7, line 32: 1. Praise (❤️): Great work on removing all those duplicated, plus replacing all those mocks! 🥇 
2. Suggestion (💡): Unrelated, but I've been thinking on all those data class mocks, maybe we can create a custom `Detekt` rule to warn when such data classes are mocked on test. Maybe such rules even exists already... 🤔

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, awesome work, only 11 such `WC` tables left to migrate, with 11 more of the "shared" ones, plus `EcryptedLogModel` and finally the dangerous `SiteModel`, we're getting there! 🎉 x 🎉  ^ 🎉  

-----

I have left few questions (❓), a couple of suggestions (💡) and one minor (🔍) comment for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

-----

`EXTRAS`

1. Question (❓): As part of other such PRs, you explicitly avoided making more changes than necessary, to avoid blow-up the size of the PR itself. What changed here, was all those changes necessary, or some of them, which then drive the rest... 🤔

PS: I am asking the above to try and understand the strategy moving forward, to know what to expect. 🙏

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
wzieba: I provided a dependency to the `CurrencyPositionConverter` (`AppLogWrapper`), so I had to inject this in DI graph.

ParaskP7: Aaaaaa, okay, okay, I see @wzieba, thanks for clarifying that. Does that really needed though, what if we use pure `android.util.Log` instead to avoid all that, do we really need `AppLog` here? 🤔

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
wzieba: Yeah, it's an IDE issue 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
wzieba: Yes - using static `android.util.Log` would not allow us to run a unit test on the machine, as it's an Android dependency. We would have to move it to Firebase Test Lab or use Robolectric, which is significantly slower.

ParaskP7: Ah, true, you've convinced me, let's keep it as is! 👍 

Question (❓): A follow-up, do we need the duplicate setup of `@TypeConverters` or `addTypeConverter(currencyPositionConverter)` for this converted during `buildDb(...)` is enough? 🤔

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/persistence/TestDatabase.kt
wzieba: Thanks! 

I don't really like the idea of `DaoUnitTest` as I'd try to reduce inheritance to minimum, especially in tests. This can backfire, when e.g. we'll want to have a test which is both `DaoUnitTest` but also a `BaseUnitTest` and by using this pattern, we could give a signal for other "base test classes" which will be even more problematic as tests grow larger.

But, I'm fine with the test rule idea, as it's a composition. So, partially applying your suggestion, WDYT about this POC? Does it look good to you? 7bc96545a4b690b56838a2abaf9d6c3ba23cbd6f

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
wzieba: sorry, I'm not sure what you mean, could you please rephrase?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
wzieba: I think in some cases it was useful to add origin class context, and in some cases it wasn't - we could go through one-by-one and discuss all of them, but I'm not sure if it's really productive, WDYT?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
ParaskP7: I mean this line here: [CurrencyPositionConverter::class,](https://github.com/woocommerce/woocommerce-android/pull/14251/files#diff-c1dc6c021892393397fd1b7b7c26e9a85a1eca0081730fa066a0c537348042c6R183)

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
ParaskP7: Nah, thanks, not productive at all, I just wanted to understand how you are thinking about it, for me, you know, I am all in favor of not having multiple imports referring to the same thing... 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
wzieba: Sorry, I'm still not getting your point. If you think something is duplicated, please feel free to remove it and verify if the build works and tests are passing. Alternatively, here's [the documentation](https://developer.android.com/training/data-storage/room/referencing-data) which can answer some of your questions.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SettingsDao.kt
wzieba: Yes, the provided snippet would work as well 👍 !

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/persistence/TestDatabase.kt
ParaskP7: > I don't really like the idea of DaoUnitTest as I'd try to reduce inheritance to minimum, especially in tests. This can backfire, when e.g. we'll want to have a test which is both DaoUnitTest but also a BaseUnitTest and by using this pattern, we could give a signal for other "base test classes" which will be even more problematic as tests grow larger.

Yea, actually, I don't like the whole base class thing either... 😅 

> But, I'm fine with the test rule idea, as it's a composition. So, partially applying your suggestion, WDYT about this POC? Does it look good to you? https://github.com/woocommerce/woocommerce-android/commit/7bc96545a4b690b56838a2abaf9d6c3ba23cbd6f

Love it @wzieba, much better, yes, no base class involved.

Suggestion (💡): Just one minor thing, you don't need `TestDatabase` anymore, all the setup can be on this one `DatabaseTestRule` class, right? 🙏

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
ParaskP7: 😊 okay, thanks, this [here](https://developer.android.com/training/data-storage/room/referencing-data#converter-initialization) answers my question:

> Then, in addition to declaring your converter class in @TypeConverters, use the [RoomDatabase.Builder.addTypeConverter()](https://developer.android.com/reference/androidx/room/RoomDatabase.Builder#addTypeConverter(java.lang.Object)) method to pass an instance of your converter class to the RoomDatabase builder: ...

The "in addition" is the key here, which means to me that adding `@TypeConverters` is needed, no matter if you also use `addTypeConverter(currencyPositionConverter)` explicitly.

Thanks for throwing at me the docs @wzieba ! 🙏

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/persistence/TestDatabase.kt
wzieba: Sure, thanks! I'll proceed with the change 👍 .

You asked somewhere about returning `Builder` instead of `Database` right? I can't find the comment 😵 but anyway, I think that I agree to return `Database` - with the rule, it can be easy later to return a builder (or introduce another rule) if it'll be necessary.

ParaskP7: > Sure, thanks! I'll proceed with the change, though.

Thanks! 🌟 

> You asked somewhere about returning Builder instead of Database right? I can't find the comment 😵 but anyway, I think that I agree to return Database - with the rule, it can be easy later to return a builder (or introduce another rule) if it'll be necessary.

Yea, it was a suggestion on this exact [review comment](https://github.com/woocommerce/woocommerce-android/pull/14251#discussion_r2174825619) that started all this discussion, thanks:

> 2. Suggestion (💡): Why not include the actual .build() be part of this? This will avoid having to type .build() for every test, also one less import saved. Then, instead of provideTestDatabase(...) we rename it to buildTestDatabase(...).

wzieba: ah I see, thanks! 🙌

wzieba: Done in: 93b0fd75bdb1b6d6e771278af6aac9652807a772

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
wzieba: I'll think about it 👍 I don't think Detekt will be able to do this effectively (as it's static), nor it has any out-of-the-box rule, but we can probably use a custom Lint rule 👍

wzieba: > This also detects common types that should never be mocked, such as Kotlin data classes or AutoValue classes.

This looks promissing 😄 https://github.com/slackhq/slack-lints?tab=readme-ov-file#do-not-mock

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/persistence/TestDatabase.kt
ParaskP7: Awesome, so much duplicate code removed, thank YOU @wzieba ! 🙇 ❤️ 🚀

libs/fluxc-plugin/src/testFixtures/kotlin/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
ParaskP7: 😄 Oh nice, we can try that yes! 🤞

wzieba: fyi: https://linear.app/a8c/issue/AINFRA-857/introduce-donotmockdataclass-lint-rule

ParaskP7: Awesome, thanks! 💯

PR #14103, opened 2025-05-23
[Login] Fix a crash during login after Activity or process death

hichamboushaba:
WOOMOB-381

### Description
This PR fixes a crash that happens during the site credentials login, the crash happens if a Activity or process death occurred when the Application Passwords tutorial screen was shown, and the user dismissed it and retried login.

> [!IMPORTANT]
> I'm targeting `release/22.4` branch with the fix as this impacted 268 users just in the last month, but if you feel this should go to `trunk`, please let me know. 

### Steps to reproduce
1. Enable "Don't keep activities" in the developer options of your device.
2. Prepare a site that can't use native site credentials login (for example using this [plugin](https://wordpress.org/plugins/captcha-for-contact-form-7/), and enabling captcha for "WordPress login")
3. Open the app and start login.
4. Enter the site address, then credentials, then tap on **Continue**.
5. In the Application Passwords tutorial screen, tap on **Contact support**
6. Go back to the site credentials screen.
7. Tap on **Continue** again.

### Testing information
- Confirm the app doesn't crash.

### The tests that have been performed
Tested on an emulator, and confirmed that the app crashs because of this on `trunk` and `release/22.4` branches, this branch fixes it.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14252, opened 2025-06-26
Fix rare crash during login with site credentials.

hichamboushaba:
Closes WOOMOB-381

### Description
This PR fixes a crash that seems happens during site credentials, we had a previous attempt in https://github.com/woocommerce/woocommerce-android/pull/14103 that reduced the number of occurrences quite a lot, now, the crash is rarely happening, but as it's still happening, this PR updates the logic to show an error instead of crashing.

### Steps to reproduce
I couldn't reproduce the crash, so just code review should be enough.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

irfano (approved):
LGTM! I couldn’t reproduce the crash either. However, I noticed that in the crash logs, users receive a `Status Code: 302` error, which is related to URL redirection. So it’s likely a third-party plugin issue. Showing an error seems like the best approach.

I verified that it will display this error screen instead of crashing:

hichamboushaba:

> However, I noticed that in the crash logs, users receive a Status Code: 302 error, which is related to URL redirection. So it’s likely a third-party plugin issue

The 302 is a normal redirection during the login flow, so I don't think this is related.

PR #14259, opened 2025-06-30
[WOOMOB-725] Fix NPE when adding variation item by scanning a barcode in the POS

malinajirka:
#WOOMOB-725

### Description

This PR fixes a NPE in the POS. When a parent of a variation isn't found in the cache, the app crashes.

The solution simple returns `NotFound` from WooPosSearchByIdentifierLocal when the parent product isn't found.

### Steps to reproduce

1. Open POS
2. Scan barcode of a variation (the parent product must not be in the first 200 products that are loaded into the cache. Alternatively, you can reduce the size of the page we load upon opening the POS)
3. Notice the variation is added to the cart.
4. Exit the POS
5. Re-enter the POS
6. Scan the same barcode again - 💥 

### Testing information

- Test the above flow and make sure it doesn't crash.

The above.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierLocal.kt
```diff
@@ -9,20 +9,28 @@ class WooPosSearchByIdentifierLocal @Inject constructor(
...
+
+    private suspend fun findProductByIdentifier(identifier: String) =
+        productsCache.getAll().firstOrNull { product ->
             product.globalUniqueId.equals(identifier, ignoreCase = true)
-        }?.let { return WooPosSearchByIdentifierResult.Success(it) }
+        }
 
-        val allVariations = variationsCache.getAll()
-        return allVariations.firstOrNull { variation ->
+    private suspend fun findVariationWithParentByIdentifier(identifier: String):
```
kidinov: Shell we return the sealed type instead of nullable? Otherwise it's weird that we cover this in the type structure but still made it nullable


```
    private suspend fun findVariationWithParentByIdentifier(identifier: String): WooPosSearchByIdentifierResult {
        val variation = variationsCache.getAll().firstOrNull { variation ->
            variation.globalUniqueId.equals(identifier, ignoreCase = true)
        } ?: return WooPosSearchByIdentifierResult.Failure(
            WooPosSearchByIdentifierResult.Error.NotFound
        )

        return productsCache.getProductById(variation.remoteProductId)?.let { parentProduct ->
            WooPosSearchByIdentifierResult.VariationSuccess(variation, parentProduct)
        } ?: WooPosSearchByIdentifierResult.Failure(
            WooPosSearchByIdentifierResult.Error.NotFound
        )
    }
```

kidinov (approved):
LGTM! I left one code improvement suggestion

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierLocal.kt
malinajirka: Good call, that's even more readable! Fixed in 72eae5e724388b9f9d2cee599db67b72374ab458

PR #14233, opened 2025-06-23
[Shipping Labels] Improvements to the UI shipment details for purchased labels

hichamboushaba:
Closes WOOMOB-642
Part of WOOMOB-668

### Description
This PR updates the UI to purchased labels with the following changes:
- Hide the payment section.
- Hide the address validation status.
- Hide all the notices.
- Update the address formatting to match the other platforms.

> [!NOTE]
> WOOMOB-668 is not closed yet because we need to display the price of the label, but I couldn't implement it yet.

### Steps to reproduce
1. Open an order with a purchased shipping label.
2. Tap on "Create shipping labels"
3. Expand the shipment details bottom sheet.

### Testing information
- Confirm the payment section is hidden for purchased labels and shown for non-purchased labels.
- Confirm that the address section matches the other platforms behavior.
- Confirm that no notices are shown for purchased labels.

### The tests that have been performed
- The above

### Images/gif
| Before | After |
| --- | --- |
|Screenshot_20250623_193134 | Screenshot_20250623_193340 |
| Screenshot_20250623_193142 | Screenshot_20250623_203350 |

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14260, opened 2025-06-30
[Shipping Labels] Parse and display UPS additional rate options

hichamboushaba:
Closes WOOMOB-729

### Description
This PR adds logic to handle parsing UPS additional rates, and displaying them. This includes the calculation of the total cost based on the selected options.

The idea here was to have two fields `selectedOption` which represents the "base" rate, this can either be `DEFAULT`, `SIGNATURE` or `ADULT_SIGNATURE` (as the three are mutually exclusive), then we have an `additionalOptions` which can then be added to the selected rate.

With the above idea, I also refactored the models a bit by moving most of the common fields to the `ShippingRateUI` instead of the `ShippingRateOptionUI` (these include the `basePrice`, the estimated days...)

> [!IMPORTANT]
> The purchase part is not implemented yet, the additional options will be ignored, this will be implemented in the next PR. 

### Steps to reproduce
1. Open Shipping Labels creation form.
2. Select a UPS package.
3. Check UPS rates.
4. Select various options.

### Testing information
- Confirm the `signature required` and `adult signature required` are mutually exclusive.
- Confirm that the other additional rates can be added combined.
- Compare the rates and the calculation with the web version and confirm they behave the same. 

The above

### Images/gif

https://github.com/user-attachments/assets/3dab6d36-3baf-47c6-a291-632d582a6fd9


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -327,9 +358,11 @@ class WooShippingLabelCreationViewModel @Inject constructor(
...
                 } else {
                     ShippingRatesState.DataState(
-                        selectedRatesSortOrders[index],
-                        sortShippingRates(selectedRatesSortOrders[index], map),
-                        selectedRates[index]
+                        selectedRatesSortOrder = selectedRatesSortOrders[index],
+                        shippingRates = sortShippingRates(selectedRatesSortOrders[index], map),
+                        selectedRate = selectedRates[index],
+                        onSelectedShippingRateChanged = ::onSelectedShippingRateChanged,
+                        onSelectedRateOptionChanged = ::onSelectedRateOptionChanged
```
hichamboushaba, line 365: I moved the callbacks to the state, this makes it easier to reference them from nested Compose views, the only downside is that it makes tests more verbose, but also more robust IMO as we can't call a callback without setting the correct state before.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -1200,11 +1226,11 @@ data class ShippingLineSummaryUI(
 ) : Parcelable
 
 @Parcelize
-data class ShippingRateSummaryUI(
+data class ShipmentCostUI(
```
hichamboushaba, line 1229: I renamed this, now this is completely decoupled from the rates, and this will help us in implementing the last part of this issue: WOOMOB-668

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/ShipmentDetails.kt
```diff
@@ -513,30 +513,34 @@ private fun PaymentSection(
...
+        } else {
+            shipmentCostUI?.serviceName
+        }
         ShipmentCostRow(
-            title = shippingRateSummary?.serviceName ?: stringResource(R.string.subtotal),
-            total = shippingRateSummary?.total,
-            modifier = Modifier.padding(top = dimensionResource(R.dimen.major_100))
+            title = serviceName ?: stringResource(R.string.subtotal),
+            total = shipmentCostUI?.formattedBasePrice
         )
```
hichamboushaba, line 532: This fixes two issues with the previous implementation:
- Shows the base price and not the total price, as previously we show the price of the selected option here, which was equal to the total price.
- We didn't have the `base fee` suffix when needed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/rates/domain/WooShippingRatesDomainMapper.kt
```diff
@@ -194,4 +166,13 @@ class WooShippingRatesDomainMapper @Inject constructor(
...
     }
+
+    private fun Option.getTitle() = when (this) {
+        Option.DEFAULT -> ""
+        Option.SIGNATURE -> resourceProvider.getString(R.string.woo_shipping_rate_option_signature_required)
+        Option.ADULT_SIGNATURE -> resourceProvider.getString(R.string.woo_shipping_rate_option_adult_signature_required)
+        Option.CARBON_NEUTRAL -> resourceProvider.getString(R.string.woo_shipping_rate_option_carbon_neutral)
+        Option.ADDITIONAL_HANDLING -> resourceProvider.getString(R.string.woo_shipping_rate_option_additional_handling)
+        Option.SATURDAY_DELIVERY -> resourceProvider.getString(R.string.woo_shipping_rate_option_saturday_delivery)
+    }
```
irfano, line 177: I’d like to expose only the resource ID from this mapper class, but doing so would require a lot of changes for a minor benefit. I’m just sharing the thought, we can skip it for now.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4694,6 +4695,13 @@
     <string name="wpp_shipping_ups_tos_condition_technology_agreement">I also agree to the &lt;a href=\"ups-technology-agreement\"&gt;UPS® Technology Agreement&lt;/a&gt;.</string>
     <string name="wpp_shipping_ups_tos_accept">Confirm and continue</string>
 
+    <string name="woo_shipping_rate_surcharge_description_template">%1$s (%2$s)</string>
```
irfano, line 4698: Let’s add the `+` symbol in parentheses, like `Signature required (+$5.90)`. The Figma design and the web version also include this `+` symbol.

irfano (changes requested):
I noticed a scroll issue and it was not reproducible before 22.6. We modified some scroll settings in UPS PRs, might they be the reason?



- In landscape mode, the details screen isn’t scrollable, so I can’t see anything below the “SHIPMENT COST” section.



- In portrait mode, "After purchasing a label,.." button is hidden behind the “Purchase label” button and is not visible.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -327,9 +358,11 @@ class WooShippingLabelCreationViewModel @Inject constructor(
                     ShippingRatesState.NoAvailable
                 } else {
                     ShippingRatesState.DataState(
-                        selectedRatesSortOrders[index],
-                        sortShippingRates(selectedRatesSortOrders[index], map),
-                        selectedRates[index]
+                        selectedRatesSortOrder = selectedRatesSortOrders[index],
```
irfano, line 361: `selectedRatesSortOrders[index]` crashes with ` java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1` when following these steps:
- Select UPS options for the last shipment
- On  thesplit shipment screen, remove the last shipment.
- Tap the DONE button.

hichamboushaba: Great catch, but this code is the same on trunk, I just added the name of the argument.
Also the crash happens also on trunk and regardless of using UPS or any other carrier, this why we should try to work on having a single wrapper object of all the data, and then have a single `Flow` that has all the information, this multiple `Lists` in the different flows is not easy to maintain.

I'll try to look for a quick solution, otherwise given this is not caused by this PR, we can work on it on a separate issue.

hichamboushaba: I submitted this commit 5068d0d375fa37984b99ab919c4da0efc02d4212 to fix this, it's a sort of a workaround the issue of the various Flows, what happens here is that when we update the sizes of the flows in `adjustFlowSizesToShipmentCount`, it doesn't happen atomically, so we trigger the collection in `observeShippingRatesState` when the first flow is updated (`selectedRatesSortOrdersFlow` in this case) and then it results in the crash.

As I suggested above, IMO the better solution is to use a single `StateFlow<List<ShipmentDataWrapper>>`, where `ShipmentDataWrapper` will contain all the items, this way we can benefit from `StateFlow#update` to make atomic updates and have a more robust logic.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/rates/domain/WooShippingRatesDomainMapper.kt
hichamboushaba: I agree, this was already existing before, so I kept it, I just moved it to this function and added new names instead of using the ones from the legacy code.

WooCommerce/src/main/res/values/strings.xml
hichamboushaba: Good eye, I should've mentioned that I didn't add it for now because we need to have a formatting logic that works with both positive and negative values (see here p1750994700834159-slack-C05VBLKHHV1), let's keep it as it's now, and I'll work on it on a separate PR.

hichamboushaba:
> I noticed a scroll issue and it was not reproducible before 22.6. We modified some scroll settings in UPS PRs, might they be the reason?
> In landscape mode, the details screen isn’t scrollable, so I can’t see anything below the “SHIPMENT COST” section.
> In portrait mode, "After purchasing a label,.." button is hidden behind the “Purchase label” button and is not visible.

Great catches, but I don't think this is related to this PR or any other UPS PR, as they didn't modify the shipment details bottom sheet, the only change was for a nested view that shouldn't cause these issues, this could probably be caused by one the payment PRs or this [PR](https://github.com/woocommerce/woocommerce-android/pull/14233), if you agree with me, then let's log a ticket for those issues, and I'll take a look at them today.

hichamboushaba:
I tested version 22.6, and I see the same scrolling issues:
</br>




So this is not caused by any recent PR, I'll log an issue and check this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
irfano: > this multiple Lists in the different flows is not easy to maintain.

I agree. 

Thanks for fixing this quickly, even though it's not related to this PR. 🙏🏻

irfano (approved):
Thank you for investigating the issues that I noticed. This PR is good to merge. 👍🏻 🚢

PR #14261, opened 2025-07-01
Update Compose BOM to 2025.06.01

hichamboushaba:
Closes WOOMOB-736
Closes WOOMOB-442

### Description
This PR updates Compose to the latest version, which fixes two issues for us:
- Fixes this [issue](https://issuetracker.google.com/issues/391378120) that was blocking the target update to 35, **I confirmed the updated fixed the issue for us when using Arabic**.
- Brings this [change](https://android-review.googlesource.com/c/platform/frameworks/support/+/3260658) that should fix one of our main crashes now.

I didn't update the target SDK version in this PR, I'll do it in a separate PR, because I want to get the benefits of this update ASAP, and the target SDK would require more testing than this change.

### Steps to reproduce
Open few screens in the app and confirm there are no regressions.

- I tested the app in Arabic language after setting the target SDK to 35 and confirmed it works as expected now.
- I tested various screens to confirm there are no regressions.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/DashboardContainer.kt
```diff
@@ -70,21 +69,12 @@ fun DashboardContainer(
     blazeCampaignCreationDispatcher: BlazeCampaignCreationDispatcher,
 ) {
     dashboardViewModel.dashboardCardsState.observeAsState().value?.let { state ->
-        WooLog.d(
-            WooLog.T.DASHBOARD,
-            "DashboardCardsState observed: widgets=${state.widgets.size}, refreshing=${state.isRefreshing}"
-        )
```
hichamboushaba, line 76: I removed these logs as we no longer need them.

gradle/libs.versions.toml
```diff
@@ -7,7 +7,7 @@ androidx-arch-core = '2.1.0'
 androidx-browser = '1.5.0'
 androidx-camera = '1.2.3'
 androidx-cardview = '1.0.0'
-androidx-compose-bom = '2024.09.01'
+androidx-compose-bom = '2025.06.01'
```
JorgeMucientes, line 10: I'm glad that updating to the latest BOM fixed the Arabic issue. However, I must be missing something, because in the reported [bug link](https://issuetracker.google.com/issues/391378120), Google mentions the fix was applied in: 
> androidx.compose.ui:ui-text:1.9.0-alpha01
androidx.compose.ui:ui-text-android:1.9.0-alpha01
androidx.compose.ui:ui-text-jvmstubs:1.9.0-alpha01
androidx.compose.ui:ui-text-linuxx64stubs:1.9.0-alpha01

However, the [BOM](https://developer.android.com/develop/ui/compose/bom/bom-mapping) version `2025.06.01` is using version `1.8.3` of the `ui-text` libraries. Did they apply the fix to earlier versions? What am I missing?

JorgeMucientes (approved):
Smoked tested multiple composable screens across the app and everything is looking good! Thanks for addressing this update. I left a question in case you may know the answer to it. :shipit:

gradle/libs.versions.toml
hichamboushaba: There are conflicting information on the ticket, initially they [said](https://issuetracker.google.com/issues/391378120#comment12) it will target 1.8, then later on they said it will be fixed in 1.9.

I think there may be multiple issues that cause the same issue, for our case, this [change](https://android-review.googlesource.com/c/platform/frameworks/support/+/3466624) that landed in 1.8 fixed the issue, as it explicitly calls `disableUseBoundsForWidth`.

PR #14263, opened 2025-07-01
[WOOMOB-737] Fix a few memory leaks

malinajirka:
Fixes WOOMOB-737

### Description

This PR fixes a couple memory leaks in custom views:
- References are being stored for longer than they should through invalid handling of OnGlobalLayoutListener. Our implementation assumes that OnLayout is always invoked at least once, as that's where we remove the OnGlobalLayoutListener. However, that's not guaranteed and in certain cases the OnGlobalLayoutListener is registered, but the view is destroyed before OnLayout is called leading to a leak.
- There is a second leak in WCEmptyView, which is caused by holding an implicit reference in delayed Handler.
- It completely removes WCProductPropertyReadMoreView since it contains a leak but it's unused code.

### Steps to reproduce

I don't know how to reproduce these leaks - I was browsing the app when leak canary detected an issue with the GlobalLayoutListener in WCEmptyView and then I was looking for similar issues through the codebase.

### Testing information

Smoke test
- EmptyView - it is used through out the app when we don't have data - eg. order list, product list, analytics, search, filters, network error, ... .
- WCReadMoreView - open order detail => add a very long customer note => tap on Edit => notice the note has Learn More button at the end => tap on it

The above

### Images/gif

N/A - no user facing changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson (approved):
Thanks, @malinajirka. LGTM!

PR #14264, opened 2025-07-01
Fix leaking view in ProductDetailsToolbarHelper

malinajirka:
Fixes WOOMOB-739

### Description

Fixes memory leak on the ProductDetail screen. The issue is that we store references to view in onViewCreated but we relieve them in onDestroy instead of onDestroyView. Fragment might be alive even after its view is destroyed, so holding reference to a view leads to a memory leak.

### Steps to reproduce

1. Open Product Detail
2. Go to the Order List
3. Notice the leak

### Testing information

Smoke test product detail and product list.

I've smoke tested opening the product detail, navigation product list and configuration changes.

### Images/gif

No user facing changes


kidinov (approved):
LGTM, thanks!

PR #14265, opened 2025-07-02
[WOOMOB-740][Woo POS] Top margin for the search is too small

kidinov:
WOOMOB-740

### Description

The PR sets top margin for the toolbars in the POS to use `statusBarsPadding` so it consistent across different devices

### Steps to reproduce

* Use the POS
* Notice top margin for the toolbars in the products, search, cart, cash, email
* All should be consistent and all should be bellow the status bar

above

#### before
Screenshot 2025-07-02 at 09 43 21

#### after
Screenshot 2025-07-02 at 09 49 56


malinajirka (approved):
LGTM, thanks @kidinov!

One thing I noticed on your screenshots is that the purple outline isn't exactly on top of the input field. It looks fine on my device 🤔. Quite weird tbh, almost seems like the padding for the field itself and the outline is different.
Screenshot 2025-07-02 at 10 33 41

kidinov:
@malinajirka that the buggy video translation from my device. In the reality it looks good

PR #14266, opened 2025-07-02
[WOOMOV-745][Woo POS] Navigation from the cash payment back to the checkout causes products list to be open

kidinov:
WOOMOV-745
### Description

Currently, the search state is only reset when the order was successfully paid. But navigation back from the cash screen is not handled, and the focus is taken by the search input field which causes items screen to be shown

We have 3 options to fix that:
* On navigation back from the cash, reset the search state
* On click on the cash button reset the search state
* On click on the check-out reset the search state

The last one is chosen here:
* It covers all the potential navigation issues similar to this one
* It easiest to implement
* **BUT** the UX maybe is not great as in some cases a merchant wants to go back and continue the search
Wdyt? Which option would you go with here?

### Steps to reproduce

* Open POS
* Open search
* Add any amount of products
* Select cash payment
* Navigate back
* Notice that the search is in focus and that caused the products screen to be partially shown

### Testing information

* Try different cases of the navigation from check out screen and back to the items
* Notice that nothing is weird is going on except that the search is closed after check out is clicked

Above

#### before

https://github.com/user-attachments/assets/40d7ff5c-f7ed-4c64-b15b-d2e656f3c817

#### after

https://github.com/user-attachments/assets/241590b5-6782-45db-83d9-a716ac9ba39e

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsSearchHelper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsSearchHelper.kt
@@ -67,7 +67,7 @@ class WooPosItemsSearchHelper @Inject constructor(
                     is ParentToChildrenEvent.CouponsRemoved -> Unit
                     is ParentToChildrenEvent.RemoveCouponsClicked -> Unit
                     is ParentToChildrenEvent.CouponsValidationFailed -> Unit
-                    is ParentToChildrenEvent.OrderSuccessfullyPaid  -> Unit
+                    is ParentToChildrenEvent.OrderSuccessfullyPaid -> Unit
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsSearchHelper.kt
```diff
@@ -59,16 +55,19 @@ class WooPosItemsSearchHelper @Inject constructor(
                             onSearchChanged("", 0)
                         }
                     }
+                    is ParentToChildrenEvent.CheckoutClicked -> {
+                        onCloseSearchClicked()
```
malinajirka: 🔍 I'd consider renaming this method or extracting the logic into `closeSearch/clearSearch` and calling that instead here.
```
                    is ParentToChildrenEvent.CheckoutClicked -> { closeSearch() }
                    
                    
                    fun onCloseSearchClicked() = closeSearch()
                    fun closeSearch() { ... }

malinajirka (approved):

> BUT the UX maybe is not great as in some cases a merchant wants to go back and continue the search

I would have gone with the same solution as you. I think it's a reasonable behavior and if we find out that it's common for users to go back and forth, I think we need to fix the underlying issue (totals are not displayed on the selector screen).

P.S. I'd consider adding a comment to `CheckoutClicked` handling, since the knowledge why we are closing search might get lost and we might re-introduce this issue in the future. Alternatively, we could somehow include in the name of the test. Wdyt?

kidinov:
@malinajirka Thanks for the review!

Both comments are implemented in this commit

1a5943d993cf5cff1f38b85289b079c7d4516a81

PR #14267, opened 2025-07-02
[Shipping Labels] Update purchase networking to support UPS additional rates (also fixes some issues)

hichamboushaba:
Closes WOOMOB-729

### Description
This PR updates the purchase endpoint to support the UPS additional rates, the request was updated to exactly match the plugin request, and the main changes are:
- Pass the selected options in the `selected_rate_options` field, we pass the values with their surcharge.
- Update the `package` to include the selected options (see `PackagePurchaseDTO` changes)
- Update the `selected_rate` field to match the plugin format: using a single object with `rate` and `parent` rates.
- Pass the `parent` rate when selecting a non-default rate.
- Add a field `type` to the `RateDTO` that we use for the purchase to indicate whether this is a signature rate.

The PR also updates the formatting of the `hazmat` and `customs` fields to make sure we pass the correct shipment ID in the map, I'll explain this further in the comments below.

### Steps to reproduce
1. Open the shipping labels form.
2. Select a UPS package.
3. Include some UPS specific options.
4. Purchase the label.

### Testing information
- Confirm the purchase is successful.
- Open the label in wp-admin, and confirm the cost includes the `signature` fees (it should also include the other fees, but there is a bug in the plugin (p1751303271937159-slack-C05VBLKHHV1)

The above


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jetpack/JetpackActivationEligibilityErrorViewModel.kt
```diff
@@ -40,7 +40,7 @@ class JetpackActivationEligibilityErrorViewModel @Inject constructor(
     ) { (username, role), isRetrying ->
         ViewState(
             username = username,
-            role = role.replace("_", " ").toCamelCase(),
+            role = role.replace("_", " ").capitalizeAllWords(),
```
hichamboushaba, line 43: The `toCamelCase` extension was misleading as it capitalized the letters and added spaces between them, I replaced it with `capitalizeAllWords`, and created a second extension `snakeToCamelCase` that best suits my need for this PR (the name is clearer also in this case).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingLabelRestClient.kt
```diff
@@ -134,31 +136,32 @@ class WooShippingLabelRestClient @Inject constructor(
...
+                    "rate" to selectedRate,
+                    "parent" to parentRate
                 ),
-                // TODO: `selected_rate_options` will be updated while adding UPS support PaJDVv-2Gf-p2
-                "selected_rate_options" to "",
-                "hazmat" to mapOf(selectedPackage.boxId to hazmat),
-                "customs" to customs,
+                "selected_rate_options" to selectedRateOptions.mapKeys { it.key.typeId },
+                "hazmat" to mapOf(shipmentKey to hazmat),
+                "customs" to customs?.let { mapOf(shipmentKey to it) }.orEmpty(),
```
hichamboushaba, line 163: We were having some issues here:
- For `hazmat` we were passing the package `boxId` as the `key`.
- For `customs` we were passing a List with a single item to the mapper, which then results in having a constant `key` set to `shipment_0`

Both issues results in issues:
- The `boxId` is not expected in the backend, so when no custom is passed, the backend can't read the stored data the label, and it results in displaying some wrong information (especially the cost details)
- When customs are included, we still had issues when we have multiple labels, as the last created label will override the data of the previously created labels, and this was a serious issue, as we can show the wrong destination address or the wrong cost.

Both issues can be shown in the following screenshot:
Screenshot 2025-07-02 at 17 05 56

Notice how we have two created labels (`currentOrderLabels`) but only a single entry for the other details.

The fix is to match what the plugin does: we create a key based on the shipment index in the format `shipment_<index>`, this way each shipment will have its data correctly stored.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingLabelRestClient.kt
```diff
@@ -136,30 +136,31 @@ class WooShippingLabelRestClient @Inject constructor(
...
-        customs: Map<String, CustomsDTO>,
+        hazmat: HazmatDTO,
+        customs: CustomsDTO?,
     ): WooPayload<PurchasedShippingLabelResponseDTO> {
         val url = "/wcshipping/v1/label/purchase/$orderId/"
         return wooNetwork.executePostGsonRequest(
             site = site,
             path = url,
             body = mapOf(
+                "async" to true,
```
irfano: What does this `async` parameter do?

irfano (approved):
LGTM and it works as expected. I just added a question about `async` parameter out of curiosity.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingLabelRestClient.kt
hichamboushaba: In the past, the purchase was by default synchronous, meaning that you get a success or failure after making the request, but this was slow, so the plugin team introduced the asynchronous purchase flow (that we already use before this PR), more on this here p91TBi-4K3-p2

For this PR, I added the argument to be consistent with the web request (while investigating some issues), but looking at the code (384a1-pb/), I see the plugin ignores the argument and hardcodes the value to `true` behind the scenes, so we don't need it, I'll remove it before merging the PR.

Thanks @irfano

PR #14271, opened 2025-07-03
[POS as a tab i2] Enable POS tab for stores in eligible countries behind i2 feature flag

toupper:
Closes WOOMOB-719

### Description

This PR implements the new logic for showing the POS tab, which remains behind a feature flag. The logic for checking the feature flag is also implemented in this PR.

The new visibility conditions for the POS tab are:

- The remote feature flag is enabled.
- The store’s country is one of the eligible countries (US, UK).

The rest of the checks will be performed now once the tab is tapped. This logic will be implemented on future PRs.

### Steps to reproduce

Prerequisite: a WPCOM account with at least two stores where one store is eligible for POS, and the other in a country that is not eligible for POS (e.g. CA)

- Log in to the store eligible for POS in the prerequisite --> the POS tab should be shown shortly
- Tap on the POS tab --> POS should be launched as before
- In wp-admin of this store, update the currency in WC settings to a non-eligible value
- Relaunch the app --> the POS tab should still be shown even after a while
- Tap on the POS tab --> POS should be launched as before. in a future task, a loading & eligible UI will be shown
- Go to the Menu tab and switch to the store in a country ineligible for POS --> the POS tab should not be shown even after a while
- Relaunch the app --> the POS tab should not be shown even after a while

### Images/gif

For a store in an eligible country but ineligible for other reason (e.g. currency):

i2 feature flag enabled | i2 feature flag disabled
-- | --
Screenshot_20250703_124458_Woo (Pre-Alpha) | Screenshot_20250703_124420_Woo (Pre-Alpha)

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
@@ -35,6 +35,6 @@ class WooPosTabShouldBeVisible @Inject constructor(
     private fun isCountrySupported(countryCode: String) = SUPPORTED_COUNTRIES.contains(countryCode.lowercase())
 
     private companion object {
-        val SUPPORTED_COUNTRIES = listOf("us", "gb")
+        private val SUPPORTED_COUNTRIES = listOf("us", "gb")
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
@@ -8,7 +8,7 @@ import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.store.WooCommerceStore
 import javax.inject.Inject
 
-class WooPosTabIsVisible @Inject constructor(
+class WooPosTabShouldBeVisible @Inject constructor(
     private val selectedSite: SelectedSite,
     private val wooCommerceStore: WooCommerceStore,
     private val isRemoteFeatureFlagEnabled: IsRemoteFeatureFlagEnabled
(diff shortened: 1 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosIsPosAsTabM2Enabled</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        return@withContext isCountrySupported(
+            countryCode = siteSettings.countryCode
+        )
+    }
+
+    private fun isCountrySupported(countryCode: String) = SUPPORTED_COUNTRIES.contains(countryCode.lowercase())
+
+    private companion object {
+        val SUPPORTED_COUNTRIES = listOf("us", "gb")
```
samiuelson:
```suggestion
        private val SUPPORTED_COUNTRIES = listOf("us", "gb")
```
It's not currently used outside this class.

samiuelson: ❓ In the existing `WooPosIsEnabled` use case, we're checking currencies in addition to country codes. I just want to confirm this is intentional to skip the currency check.

samiuelson (approved):
LGTM! I added non-blocking comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
toupper: > ❓ In the existing WooPosIsEnabled use case, we're checking currencies in addition to country codes. I just want to confirm this is intentional to skip the currency check.

Thanks for confirming this one! Yeah, we show the tab for those stores in eligible countries even if their currency is not the allowed one. We want to let them know once they visit the tab that by simply changing their currency, they can access the POS.

PR #14274, opened 2025-07-04
[POS as a Tab i2] Add logic to determine if POS can be launched in the tab

toupper:
⚠️ This PR depends builds on top of https://github.com/woocommerce/woocommerce-android/pull/14271 Please do not merge until that one is merged too

### Description

This PR adds the logic to determine whether POS mode can be launched once the user visits the tab. It introduces a new class, WooPosCanBeLaunchedInTab, which checks the launch conditions and returns either that POS can be launched or, if not, the specific reason why it cannot.

This class is not yet used in the app; a future PR will integrate it to either launch POS or display an error explaining why it cannot be launched.

### Steps to reproduce

As it's not used yet, we just have to review the code and verify that tests pass.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
@@ -9,8 +9,13 @@ import org.wordpress.android.fluxc.store.WooCommerceStore
 import javax.inject.Inject
 import javax.inject.Singleton
 
+/**
+ * Determines if POS can be launched *from within the POS tab* based on launch conditions,
+ * e.g., currency support, WooCommerce version, feature flags, etc.
+ * This is only checked once the POS tab is already visible.
+ */
 @Singleton
-class WooPosCanBeLaunched @Inject constructor(
+class WooPosCanBeLaunchedInTab @Inject constructor(
     private val selectedSite: SelectedSite,
     private val isScreenSizeAllowed: WooPosIsScreenSizeAllowed,
     private val getWooCoreVersion: GetWooCorePluginCachedVersion,
@@ -23,12 +28,6 @@ class WooPosCanBeLaunched @Inject constructor(
             ?: return@withContext WooPosLaunchability.NotLaunchable(
                 WooPosLaunchability.Reason.NoSiteSelected
             )
-
-        if (!isScreenSizeAllowed()) {
-            return@withContext WooPosLaunchability.NotLaunchable(
-                WooPosLaunchability.Reason.NotTablet
-            )
-        }
 
         if (!isWooCoreSupportsOrderAutoDraftsAndExtraPaymentsProps()) {
             return@withContext WooPosLaunchability.NotLaunchable(
@@ -85,7 +84,6 @@ sealed class WooPosLaunchability {
     data class NotLaunchable(val reason: Reason) : WooPosLaunchability()
 
     enum class Reason {
-        NotTablet,
         UnsupportedWooCommerceVersion,
         SiteSettingsUnavailable,
         FeatureSwitchDisabled,
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
@@ -17,7 +17,6 @@ import javax.inject.Singleton
 @Singleton
 class WooPosCanBeLaunchedInTab @Inject constructor(
     private val selectedSite: SelectedSite,
-    private val isScreenSizeAllowed: WooPosIsScreenSizeAllowed,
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -0,0 +1,93 @@
...
+        return@withContext if (isCurrencySupported(siteSettings.currencyCode)) {
+            WooPosLaunchability.Launchable
+        } else {
+            WooPosLaunchability.NotLaunchable(
+                WooPosLaunchability.Reason.UnsupportedCurrency
+            )
+        }
+    }
+
+    private fun isCurrencySupported(currency: String) = SUPPORTED_CURRENCIES.contains(currency.lowercase())
```
samiuelson, line 62: 💡 Just sharing some thoughts for awareness. 

I'm not sure if the condition related to currency and country is strict enough. For example, for the sake of test, I changed currency in my UK-based store to USD. As a result, I wasn't able to connect to WisePad reader. Error message:
`TerminalException: The BBPOS WisePad 3 reader is not supported in the United States (US) as determined by the Location this reader is registered to, tml_Fzk2Fw1prKkucT. For more information about which countries each Terminal reader may be used in, see https://stripe.com/docs/terminal/choosing-reader-and-sdk#availability.`

I was able to connect and collect payment with the M2 reader, though, so maybe the flow is fine 🤔

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.wordpress.android.fluxc.store.WooCommerceStore
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Determines if POS can be launched *from within the POS tab* based on launch conditions,
+ * e.g., currency support, WooCommerce version, feature flags, etc.
+ * This is only checked once the POS tab is already visible.
+ */
+@Singleton
```
samiuelson, line 18: 💡 Unless we have a good reason to make it`@Singleton`, I'd prevent making it a singleton. This may potentially cause issues with instance surviving store switching, etc. Alternatives – `@ActivityRetaindedScoped` which makes the class instance survive configuration changes, but destroys it when Activity is destroyed (e.g. we exit POS) or no  annotation at all.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -0,0 +1,93 @@
...
+class WooPosCanBeLaunchedInTab @Inject constructor(
+    private val selectedSite: SelectedSite,
+    private val getWooCoreVersion: GetWooCorePluginCachedVersion,
+    private val wooCommerceStore: WooCommerceStore,
+    private val isRemotelyEnabled: WooPOSIsRemotelyEnabled
```
samiuelson, line 23: 💡 I can see there's internal 10 sec value caching inside WooPOSIsRemotelyEnabled -> WCSSRModelCachingFetcher. This may potentially cause some frustration if the merchant changes the value remotely but it's not reflected in the app.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Determines if POS can be launched *from within the POS tab* based on launch conditions,
+ * e.g., currency support, WooCommerce version, feature flags, etc.
+ * This is only checked once the POS tab is already visible.
+ */
+@Singleton
+class WooPosCanBeLaunchedInTab @Inject constructor(
+    private val selectedSite: SelectedSite,
+    private val getWooCoreVersion: GetWooCorePluginCachedVersion,
```
samiuelson, line 21: 💡 `GetWooCorePluginCachedVersion` is checking for the plugin version only locally (DB-cached value). In case we plan to offer some kind of "Retry" option, to let the user update WC version and come back to the app, this will not work.

samiuelson (approved):
Looks good overall, @toupper! For awareness, I shared a few concerns about potential bugs and inconveniences in some edge-case situations. Use your best judgment and context to decide if they are valid and worth addressing.

toupper:
Thanks for the comments @samiuelson! These are for sure quite insightful. I merged this PR and will tackle your remarks in future PRs when we have more context about it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
toupper: Thanks for the comment @samiuelson ! That is fixed here https://github.com/woocommerce/woocommerce-android/pull/14291

PR #14278, opened 2025-07-04
[Shipping Labels] Multiple fixes

hichamboushaba:
Closes WOOMOB-690
Closes WOOMOB-671

### Description
This PR includes multiple fixes:
- It fixes WOOMOB-690 by fixing the logic of inferring purchased labels, we were checking just the value of `refund` and ignoring the `status`, which means failed labels were treated as purchased, and thus preventing the creation of new labels.
- It fixes WOOMOB-671 by auto-filling the email address after the normalization. This is needed because the normalization API doesn't return an email, so we should set it back from the order, this follows what iOS does too.
- It fixes an issue that I identified when working on this: we weren't sending multiple address fields when fetching rates and making a purchase, I updated the mapping logic from `Address` to `ShippingAddressDTO` to include all accepted fields, and removed the `email` field as it's not allowed by the purchase API.

##### WOOMOB-690 and WOOMOB-671
1. Open shipping labels creation form.
2. Open destination address and enter a phone number with 5 digits.
3. Select a UPS package.
4. Purchase a label, the purchase will fail.
5. Go back to order details, then re-open shipping labels form.

##### Other issues
1. Create an order.
6. Edit the shipping address from order details to include a second address line (for example "Other line"), and enter a company name. (We edit here because the shipping labels address editor doesn't allow adding a second line)
7. Open the shipping labels creation form.
8. Purchase a label.
9. Check the created label and confirm it shows information company and other line.

### Testing information
- For WOOMOB-690, confirm that you can purchase a new label after the failure.
- For WOOMOB-671, when editing the destination address, confirm that the email was autofilled too.
- For the address missing information, confirm that the created label has all passed information.

The above.

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/datasource/WooShippingConfigDataStore.kt
```diff
@@ -35,9 +36,10 @@ class WooShippingConfigDataStore @Inject constructor(
             ?.let { mapper(it) }
     }
 
-    fun getShippingLabels(orderId: Long): Flow<List<ShippingLabelModel>?> = observeConfig(orderId).map { config ->
-        config?.shippingLabelData?.currentOrderLabels?.let {
-            it.map { shippingLabelDTO -> mapper(shippingLabelDTO) }
+    fun getPurchasedLabels(orderId: Long): Flow<List<ShippingLabelModel>?> = observeConfig(orderId).map { config ->
```
hichamboushaba, line 39: I renamed this to better reflect its new role: returning only purchased labels.

This change is to fix an issue where failed labels were shown in the order details.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingNetworkingMapper.kt
```diff
@@ -106,7 +106,7 @@ class WooShippingNetworkingMapper @Inject constructor(
             firstName = name.getOrElse(0) { "" },
             phone = destinationAddressDTO.phone.orEmpty(),
             address2 = destinationAddressDTO.address2.orEmpty(),
-            email = destinationAddressDTO.email.orEmpty(),
+            email = "", // TODO handle email in destination address
```
JorgeMucientes: Is this a TODO leftover? You are already handling setting the email after destination address is validated.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingNetworkingMapper.kt
```diff
@@ -221,12 +221,15 @@ class WooShippingNetworkingMapper @Inject constructor(
 
     fun toDestinationAddressDTO(address: Address): DestinationAddressDTO {
         return DestinationAddressDTO(
+            company = address.company,
+            name = "${address.firstName} ${address.lastName}",
+            phone = address.phone,
```
JorgeMucientes, line 226: I wonder why the error we get when attempting to purchase a label with a 5-digit phone number is not happening when validating the destination address. Ideally, the backend should be validating this when verifying the address, especially if this is a potential cause for the purchase request to fail. 

I see the API returns an `error` field with a string showing the cause of the purchase failure, I wonder if we could reuse this text to handle the purchase error better. In the web they are actually using this error value to display a message to the user: 
```
      "error": "ShipTo PhoneNumber must be at least 10 alphanumeric characters",
```
I'd say it is not ideal, I'd probably prefer to get some sort of `errorType` field from the API so that we could manage the error ourselves in the app. 

Another thing I noticed is that iOS validates locally the entered phone number in the shipping addresses screen. We might want to do that too: 

Screenshot 2025-07-08 at 19 17 59

JorgeMucientes (approved):
Nice work @hichamboushaba 
Approving (and will merge soon so we can include this fixes for running M4 tests plan). If you confirm the TODO is a leftover I can remove it myself before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingNetworkingMapper.kt
hichamboushaba: Good catch, sorry for the leftover, I updated the comment to better reflect what we do.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/networking/WooShippingNetworkingMapper.kt
hichamboushaba: The error comes from EasyPost for specific carriers, that's why it happens only for UPS and not other carriers.

> Another thing I noticed is that iOS validates locally the entered phone number in the shipping addresses screen. We might want to do that too:

Yes, I'm aware, this is not always a good behavior, as validating international phone numbers can be tricky, so we should be careful to make sure we don't block valid phone numbers.
In the old flow, IIRC we discussed this, and we agreed to not validate destination phone numbers to follow what the web was doing back then (and it's still doing it from what I see), anyway, before following what the iOS does, I would suggest checking the side effects carefully, as having a late error (what happens now) is better than not allowing the purchase for valid phone numbers.

hichamboushaba:
@JorgeMucientes thanks for the review, I saw the Github notification on my phone and decided to check the PR to unblock you.

Please feel free to make any other needed modifications to merge this 🙏

PR #14286, opened 2025-07-07
[WOOMOB-779][POS Orders] Sales Channel filter option

samiuelson:
## Sales Channel filter option

WOOMOB-779
### Description

This PR adds new filter option in Orders screen, allowing to filter by orders created in Point Of Sale.

### Steps to reproduce

N/A

### Testing information

1. Go to Orders -> Filters -> Sales Channel
2. Select Point Of Sale
3. Tap "Show Orders"

* Verify that sales channel filter works in combination with other filter options as well
* Verify that sales channel filter is not available in case WC version is < 9.9.0

Above.

### Images/gif

https://github.com/user-attachments/assets/2beace2b-33ef-4461-8e7e-d49cac2a01d4


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/data/OrderFilterDataModels.kt
```diff
@@ -37,3 +38,9 @@ data class DateRangeFilterOption(
     val startDate: Long,
     val endDate: Long,
 ) : OrderFilterOption()
+
+data class SalesChannelOption(
```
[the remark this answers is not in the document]
samiuelson: Removed

samiuelson:
💡 Added "do not merge" label to make sure we merge this PR together with the upcoming analytics tracking PR.

.gitignore
```diff
@@ -116,3 +116,4 @@ google-upload-credentials.json
 
 # Cursor AI Editor
 /.cursor/
+/CLAUDE.md
```
malinajirka, line 119: ❓ Is this change intentional? We discussed the opposite approach - adding the file to the repo. Wdyt?

malinajirka (approved):
Thanks @samiuelson ! LGTM, great job! :shipit:

.gitignore
samiuelson: I was thinking that until we decide about the details (e.g., contents of that file, location, etc), it may be helpful to ignore it. I can revert that if you think it's not a good idea.

malinajirka:
Considering this PR is failing on CI and it has `do not merge` label, I'm moving the target version to 22.9 to unblock the codefreeze. If you want to include these changes in 22.8, please let me know.

PR #14287, opened 2025-07-07
Introduce `DoNotMockDataClass` lint rule

wzieba:
Close: AINFRA-951
Internal: paqN3M-1w9-p2

### Description

This PR introduces `DoNotMockDataClass` from Slack's lint (provided by WordPress lint wrapper). It also adds current violations to the baseline - I hope that in the future we'll address them.

And, on top of that, it improves configuration of Lint in this project. From now on:
- Custom lint checks are added to _all_ Library and App modules. If checks are not added, then rules from them don't run in a module. E.g., in this project "WordPress lint checks" were _not_ running in `fluxc-tests`. Now, they do.
- `WooCommerce-Wear`'s lint will now send sarif report to GitHub. This project will check only its own codebase.
- `WooCommerce`'s lint will check itself + all dependencies

The above is IMO optimal configuration:
- we benefit from performance of `checkDependencies=true` (see [Only analyze app/leaf modules](https://googlesamples.github.io/android-custom-lint-rules/user-guide.md.html))
- we don't duplicate violations from subprojects (`WooCommerce-Wear`'s lint e.g. doesn't include `FluxC`violations)

### Testing information

See: https://github.com/woocommerce/woocommerce-android/pull/14288 . There are 3 violations:

- From `WooCommerce-Wear` - this means that sending `.sarif` from this module works 👍 
- From `fluxc-plugin` - this means that `checkDependencies` works fine from `WooCommerce` 👍 
- From `fluxc-test` - this is not detected as we use `abortOnError = true` (default). So whenever the violation from `fluxc-plugin` is detected, the Gradle job is canceled.

I was thinking about setting `abortOnError = false`, so all lint violations would be commented, but this also means that the Gradle task would return "success" every time (even if there are violations), and this doesn't sound good for dev experience.

### Screenshot

Screenshot 2025-07-17 at 16 35 39


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/commands/lint.sh
```diff
@@ -20,5 +20,6 @@ if [ $app_lint_exit_code -ne 0 ] || [ $wear_lint_exit_code -ne 0 ]; then
 fi
 
 upload_sarif_to_github 'WooCommerce/build/reports/lint-results-jalapenoDebug.sarif'
+upload_sarif_to_github 'WooCommerce-Wear/build/reports/lint-results-jalapenoDebug.sarif'
```
ParaskP7, line 23: Praise (❤️): Nice!

WooCommerce-Wear/build.gradle
```diff
@@ -120,6 +120,10 @@ android {
             }
         }
     }
+
+    lintOptions {
+        checkDependencies = false
```
ParaskP7, line 125: > [...but I want to be elaborative here that `WooCommerce-Wear` should only lint itself](https://github.com/woocommerce/woocommerce-android/pull/14287/commits/f8eaa4a76462261095154bf14c1d910474bc4618)

Praise (❤️): I agree with being explicit here!

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, nicely done, let's stop the mock bleeding! 🧪 x 🧪 ^ 🧪 

---

`EXTRAS`

> And, on top of that, it improves configuration of Lint in this project. From now on:

All great improvements! 🥇

> The above is IMO optimal configuration:

I agree! 👍

> I was thinking about setting abortOnError = false, so all lint violations would be commented, but this also means that the Gradle task would return "success" every time (even if there are violations), and this doesn't sound good for dev experience.

👍

PR #14291, opened 2025-07-08
[POS as a Tab i2] Fix currencies check

toupper:
Part of WOOMOB-715

### Description

From this comment https://github.com/woocommerce/woocommerce-android/pull/14274#discussion_r2185169418

With this PR we fix the logic for checking the POS eligibility based on currency. Before we just checked the currencies, but we should actually verify that the pair store country and currency is the right one. It should be:

- UK Store, British Pound
- US Store, US Dollar

Any other combination won't be allowed

### Steps to reproduce

As this class is still not being used, we should verify that we cover the cases in the unit tests and that they pass succesfully.


samiuelson (approved):
LGTM :shipit:

PR #14296, opened 2025-07-09
[WOOPRD-644][Woo POS][Barcodes] Add Bluetooth settings button to barcode scanner pairing screen

kidinov:
WOOPRD-644

### Description
This PR adds a dedicated Bluetooth settings button to the barcode scanner pairing screen, allowing users to easily access Android's Bluetooth settings during the setup process.

The design is not final and is based on prototype.

**Problem:**
Users need to manually navigate to Bluetooth settings to pair their barcode scanner, which interrupts the setup flow and creates a poor user experience.

**Solution:**
- **Add "Open Bluetooth Settings" button** to the PairOnYourDevice screen that launches Android's Bluetooth settings
- **Create dedicated `PairOnYourDeviceContent` composable** for better code organization and separation of concerns
- **Remove barcode image** from PairOnYourDevice step as it's not needed for this screen

### Steps to reproduce
1. Navigate to WooCommerce POS home screen
2. Open menu and select "Barcode scanning"
3. Click "Set up a barcode scanner"
4. Select a scanner device and proceed through the flow
5. Reach the "Pair on your device" screen
6. Verify the "Open Bluetooth Settings" button is displayed with a settings icon
7. Tap the button and verify it opens Android's Bluetooth settings
8. Verify the screen layout looks clean without the barcode image
9. Test that Next/Back navigation still works correctly

above

### Images/gif

https://github.com/user-attachments/assets/b710b129-f1bf-4875-859d-d8508b9f94ba

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
@@ -148,7 +148,7 @@ fun WooPosScanningSetupDialog(
                         onSecondaryClick = { viewModel.onUiEvent(WooPosScanningSetupUiEvent.OnSecondaryButtonClicked) }
                     )
 
-                    is ScanningSetupStep.PairOnYourDevice -> BarcodeStepContent(
+                    is ScanningSetupStep.PairOnYourDevice -> PairOnYourDeviceContent(
                         step = step,
                         onPrimaryClick = { viewModel.onUiEvent(WooPosScanningSetupUiEvent.OnPrimaryButtonClicked) },
                         onSecondaryClick = { viewModel.onUiEvent(WooPosScanningSetupUiEvent.OnSecondaryButtonClicked) },
@@ -309,7 +309,6 @@ private fun BarcodeStepContent(
     step: ScanningSetupStep,
     onPrimaryClick: () -> Unit,
     onSecondaryClick: () -> Unit,
-    onOpenBluetoothSettings: (() -> Unit)? = null,
 ) {
     val title: String
     val message: String
@@ -370,16 +369,6 @@ private fun BarcodeStepContent(
             modifier = Modifier.padding(bottom = WooPosSpacing.Large.value.toAdaptivePadding())
         )
 
-        if (step is ScanningSetupStep.PairOnYourDevice && onOpenBluetoothSettings != null) {
-            WooPosOutlinedButton(
-                text = step.bluetoothSettingsButtonText,
-                onClick = onOpenBluetoothSettings,
-                modifier = Modifier
-                    .fillMaxWidth()
-                    .padding(bottom = WooPosSpacing.Large.value.toAdaptivePadding()),
-            )
-        }
-
         Box(
             modifier = Modifier
                 .size(200.dp)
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
```diff
@@ -80,6 +82,14 @@ fun WooPosScanningSetupDialog(
             ChromeCustomTabUtils.launchUrl(context, url, enableSlideAnimation = true)
         }
     }
+
+    LaunchedEffect(Unit) {
+        viewModel.openBluetoothSettingsEvent.collect {
+            val intent = Intent(Settings.ACTION_BLUETOOTH_SETTINGS)
```
malinajirka: 💡 [The documentation](https://developer.android.com/reference/android/provider/Settings#ACTION_BLUETOOTH_SETTINGS) mentions `In some cases, a matching Activity may not exist, so ensure you safeguard against this.`. I'm not sure whether it's truly required, but perhaps better to be sure, wdyt?

malinajirka (approved):
Looks and works as expected, thanks! I've left one comment to consider.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
kidinov: I’ve added basic error catching here. However, adding specific handling probably isn’t worth considering because it’s either a specific ROM or a device without Bluetooth (which practically doesn’t exist).

86f4479a04c901c1692876b768826f87af826469

malinajirka: Thanks! I agree regarding the handling, one thing I'd consider is calling here `crashLogging.sendReport(...)` so we are notified in case this behavior becomes common.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
kidinov: We'd need to inject `crashLogging` into the VM, and then add another UI event in order to trigger that. I am not sure that it's worth it. Wdyt? Maybe crashing is even better than that -> in both cases the merchant won't be able to open settings from this button, but in the case with crashing, at least we'd know that without extra code complexity.

malinajirka: I think you might be able to access it directly, something like:
```
(LocalContext.current.applicationContext as WooCommerce).appInitializer.get().crashLogging
```

It's not a beautiful code, but I think considering what we need it for, the reduced complexity + app not crashing, it's worth it. Wdyt?

kidinov: Added that here b476c8d32ed545bdc98dddd901f80778dc0b8280 !

PR #14299, opened 2025-07-09
[WOOPRD-645][Woo POS][Barcodes] Add "Other" device option and streamline device selection

kidinov:
WOOPRD-645

### Description
This PR adds an "Other" device option to the barcode scanner setup flow and streamlines the device selection experience by removing navigation buttons and implementing automatic dialog transitions.

**Problem:**
Users with unsupported barcode scanners need a way to access general barcode information without being forced through the specific device setup flow.

**Solution:**
- Add "Other" device option to the scanner selection list
- Remove Back/Next buttons from device selection screen for cleaner UX
- Auto-transition to BarcodeInfoDialog when "Other" is selected

### Steps to reproduce
1. Navigate to WooCommerce POS home screen
2. Open menu and select "Barcode scanning"
3. Click "Set up a barcode scanner"
4. Verify "Other" option appears in device list
5. Select any supported device - should immediately proceed to next step
6. Go back and select "Other" - should close setup dialog and open barcode info dialog
7. Verify no Back/Next buttons are shown on device selection screen

- Verified device selection flow works for supported devices
- Confirmed "Other" selection opens barcode info dialog
- Tested removal of navigation buttons doesn't break UX
- Checked dialog transitions work smoothly

### UI

https://github.com/user-attachments/assets/e3e320f4-1476-4570-843b-b5bc2aa23185


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeScreen.kt
```diff
@@ -194,6 +194,9 @@ private fun Dialogs(
         isVisible = dialogState is WooPosHomeState.DialogState.ScanningSetupDialog,
         onDismissRequest = {
             onHomeUIEvent(WooPosHomeUIEvent.DismissScanningSetupDialog)
+        },
+        onShowBarcodeInfoDialog = {
+            onHomeUIEvent(WooPosHomeUIEvent.ShowBarcodeInfoDialog)
```
malinajirka, line 199: ❓ Out of curiosity - have you considered treating the InfoDialog as an alternative final step in `WooPosScanningSetupState`? Feels like, we are losing some of the benefits of the sealed class by treating InfoDialog differently. Wdyt?

malinajirka (approved):
Thanks @kidinov, I've left one question, but it looks and works good to me!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeScreen.kt
kidinov: I think it's a great idea. Both UX wise and code wise. But I think I can not easy to do right now as we still have the "old" dialog as the main flow, and this one is behind feature flag. I'll create a ticket to move it here latter then

kidinov: Actually I can just copy paste the UI here. I'll do that in the separate PR

PR #14305, opened 2025-07-10
[WOOPRD-639][Woo POS][Barcodes] Scanning setup flow UI based on the final design

kidinov:
Part of WOOPRD-645

## Description

Implements the complete UI flow for WooPOS barcode scanner setup with device selection, pairing steps, testing, and success screens.

No tests yet to keep the PR smaller. I'll add them a bit latter

**Note**: This contains UI only for the happy flow without actual barcode/QR functionality. The test screen automatically navigates to success after 3 seconds since real scanning logic is not implemented yet.

**Note 2**: There are some changes in the design... so it will be another PR to address them
p1752226922338979/1752225605.633679-slack-C070SJRA8DP

## Steps to reproduce

1. Open WooPOS scanner setup
2. Select a scanner device 
3. Follow setup flow: HID mode → Pair mode → Pair scanner → Test → Success → Info
4. Test screen auto-advances after 3 seconds

## Testing information

- ✅ Complete setup flow navigation
- ✅ Auto-navigation from test to success screen
- ✅ Information dialog with setup guidance
- ✅ Bluetooth settings integration
- ✅ All strings extracted to resources

---

https://github.com/user-attachments/assets/9b8ffa4a-a8fa-4cbe-9964-6652f8fc2423


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4470,6 +4470,36 @@
...
+    <string name="woopos_scanning_setup_success_title">Scanner set up!</string>
+    <string name="woopos_scanning_setup_success_message">You are ready to start scanning products.\nRead more about barcode and QR code scanner support.</string>
+    <string name="woopos_scanning_setup_more_information">More information</string>
+    <string name="woopos_scanning_setup_info_title">Scanner set up</string>
+    <string name="woopos_scanning_setup_info_message">You can scan barcodes using an external scanner to quickly build a cart.</string>
+    <string name="woopos_scanning_setup_info_bullet_1">Refer to your bluetooth barcode scanner in System Bluetooth settings.</string>
+    <string name="woopos_scanning_setup_info_bullet_2">Scan barcodes while on the item list to add products to the cart.</string>
+    <string name="woopos_scanning_setup_info_bullet_3">Ensure the search field is not enabled while scanning barcodes.</string>
+    <string name="woopos_scanning_setup_info_text">The scanner emulates a keyboard, so sometimes it will prevent the software keyboard from showing, e.g. in search. Double-press the scan button to bring the software keyboard back.</string>
```
malinajirka: > Double-press the scan button to bring the software keyboard back.

I believe this is an iOS only feature. We should remove the sentence or replace it with something like `Re-enable it by tapping on the keyboard button if available or in OS settings under "On-Screen Keyboard"`.

malinajirka (approved):
Thanks @kidinov!

The flow feels great overall. I've left one in-code comment. 

🔍 It's a bit subjective, but I find the transition animation a bit off -> it seems we play fade out + fade in in parallel, so the content is drawn over each other. I'm not sure how to fix it, but perhaps just using DecelerateInterpolator might make it look smoother.

Screenshot 2025-07-14 at 10 50 45

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4489,7 +4489,7 @@
     <string name="woopos_scanning_setup_info_bullet_1">Refer to your bluetooth barcode scanner in System Bluetooth settings.</string>
     <string name="woopos_scanning_setup_info_bullet_2">Scan barcodes while on the item list to add products to the cart.</string>
     <string name="woopos_scanning_setup_info_bullet_3">Ensure the search field is not enabled while scanning barcodes.</string>
-    <string name="woopos_scanning_setup_info_text">The scanner emulates a keyboard, so sometimes it will prevent the software keyboard from showing, e.g. in search. Double-press the scan button to bring the software keyboard back.</string>
+    <string name="woopos_scanning_setup_info_text">Tap the keyboard button if visible or enable it in OS settings under \"On-Screen Keyboard\"</string>
```
malinajirka: I was suggesting to remove just the part about `Double-press the ....` not the whole string. I think the `The scanner emulates a keyboard, so sometimes it will prevent the software keyboard from showing` is important for context, wdyt?

WooCommerce/src/main/res/values/strings.xml
kidinov: I made it shorter, and I think more clear

```
<string name="woopos_scanning_setup_info_text">Barcode scanner may hide keyboard. Tap keyboard icon or enable in OS settings under \"Physical Keyboard\"</string>
```

wdyt?

image

WooCommerce/src/main/res/values/strings.xml
kidinov: Sorry, that was WIP change =)

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4470,6 +4470,42 @@
...
+    <string name="woopos_scanning_setup_success_message">You are ready to start scanning products.\nRead more about barcode and QR code scanner support.</string>
+    <string name="woopos_scanning_setup_more_information">More information</string>
+    <string name="woopos_scanning_setup_info_title">Scanner set up</string>
+    <string name="woopos_scanning_setup_info_message">You can scan barcodes using an external scanner to quickly build a cart.</string>
+    <string name="woopos_scanning_setup_info_bullet_1">Refer to your bluetooth barcode scanner in System Bluetooth settings.</string>
+    <string name="woopos_scanning_setup_info_bullet_2">Scan barcodes while on the item list to add products to the cart.</string>
+    <string name="woopos_scanning_setup_info_bullet_3">Ensure the search field is not enabled while scanning barcodes.</string>
+    <string name="woopos_scanning_setup_info_text">The scanner emulates a keyboard, so sometimes it will prevent the software keyboard from showing, e.g. in search. Double-press the scan button to bring the software keyboard back.</string>
+    <string name="woopos_scanning_setup_test_scanner_title">Test your scanner</string>
```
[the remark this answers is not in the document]
kidinov: That string us used

WooCommerce/src/main/res/values/strings.xml
malinajirka: I'm always in favor of shorter guidance/tips! LGTM

kidinov:
@malinajirka 

> 🔍 It's a bit subjective, but I find the transition animation a bit off -> it seems we play fade out + fade in in parallel, so the content is drawn over each other. I'm not sure how to fix it, but perhaps just using DecelerateInterpolator might make it look smoother.

Yep, I agree. It feels off

The sliding animation is logical but maybe a bit too complex. 

https://github.com/user-attachments/assets/4231c15b-e2a1-4e8a-8031-9d0b49a242f5

I also tried a simpler one, without "crossfade" (when both states are visible), but it feels off on the states that are almost identical.

https://github.com/user-attachments/assets/38a2ea1a-840d-4d54-9e64-6df4772c009e

Wdyt?

malinajirka:
Thanks @kidinov!

> The sliding animation is logical but maybe a bit too complex.

I kind of like it but at the same time I kind of don't  :D. Perhaps, it feels a bit off since even the buttons are animating.

> I also tried a simpler one, without "crossfade" (when both states are visible), but it feels off on the states that are almost identical.

I'd personally go with this one. I know what you mean about the `almost identical states`, but when each state will have its own QR code, I think it'll look better.

kidinov:
@malinajirka end up with the fade in/fade out with delay

PR #14315, opened 2025-07-11
`WCOrderSummary` to `Room`

wzieba:
Closes: AINFRA-558 

### Description

This PR migrates `WCOrderSummary` from `WellSql` to `Room`

### Testing information

1. Create a new Jurassic test site - check "Smooth Generator" option for WooCommerce
2. Open the site on web, go to `Tools` > `Smooth generator`
3. Generate some products and a great number of orders, e.g. 1000 (it takes a few minutes)
4. Open the mobile app and smoke test browsing the list of orders

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/263">22.9</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
```diff
@@ -17,20 +15,14 @@ import org.wordpress.android.fluxc.persistence.WellSqlConfig
...
-    override fun setId(id: Int) {
-        this.id = id
-    }
+data class WCOrderSummaryModel(
+    val siteId: LocalId,
+    val orderId: RemoteId, // The unique identifier for this order on the server
+    val dateCreated: String, // ISO 8601-formatted date in UTC, e.g. 1955-11-05T14:15:00Z
+) {
+    @Ignore var dateModified: String? = null
 }
```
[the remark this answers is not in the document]
wzieba: Not really - the previous WellSql implementation didn't had this field in db table as well:

https://github.com/woocommerce/woocommerce-android/blob/c68a743bedcd6b5903d93f9ad898ec199eb0f243/libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt#L877-L882

so I guess it's intentional

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListItemDataSource.kt
```diff
@@ -111,20 +112,22 @@ class OrderListItemDataSource @Inject constructor(
...
-
-                if (!networkStatus.isConnected()) {
-                    // The network is not connected so remove any order summaries from the list where
-                    // a matching order has not yet been downloaded. This prevents the user from seeing
-                    // a "loading" view for that item indefinitely.
-                    val cachedOrders = orderStore.getOrdersForDescriptor(listDescriptor, orderIds)
-                    summaries.filter { cachedOrders.containsKey(it.orderId) }
-                } else {
-                    summaries
+        val orderSummaries = runBlocking {
```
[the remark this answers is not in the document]
wzieba: Well known comment. As always - it's intentional, as these migrations are not meant to resolve threading issues (sometimes they do when I'm sure there are no side effects)

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
```diff
@@ -17,20 +15,14 @@ import org.wordpress.android.fluxc.persistence.WellSqlConfig
  * [org.wordpress.android.fluxc.store.ListStore], but since we need the `dateCreated` field to group the
  * orders into time-based groups, this extra table is necessary.
  */
-@Table(addOn = WellSqlConfig.ADDON_WOOCOMMERCE)
-@RawConstraints(
-        "FOREIGN KEY(LOCAL_SITE_ID) REFERENCES SiteModel(_id) ON DELETE CASCADE",
-        "UNIQUE (REMOTE_ORDER_ID, LOCAL_SITE_ID) ON CONFLICT REPLACE"
+@Entity(
```
ParaskP7, line 19: Suggestion (💡): Just like you did elsewhere, consider adding a `TODO` similar to what we have ie. on `WCProductCategoryModel`:

`todo: as soon as SiteModel is migrated to Room, add foreign key constraint`

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
```diff
@@ -17,20 +15,14 @@ import org.wordpress.android.fluxc.persistence.WellSqlConfig
...
+    tableName = "OrderSummaryEntity",
+    primaryKeys = ["siteId", "orderId"]
```
ParaskP7, line 21: Info (ℹ️): Just an FYI that after I migrated I noticed that, when navigating to the `Orders` screen, the list of orders was empty and wouldn't auto-refresh. I then had to manually trigger pull-to-refresh to get the orders to show up.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDao.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    @Upsert
+    abstract suspend fun upsertOrderSummaries(orderSummaries: List<WCOrderSummaryModel>)
+
+    /**
+     * Returns a list of [WCOrderSummaryModel]s that match the [remoteOrderIds] provided. This method uses
+     * Kotlin's chunked functionality to ensure we don't crash with the "SQLiteException: too many SQL variables"
+     * exception.
+     */
+    @Transaction
```
ParaskP7, line 22: Praise (❤️): 👍 

> The need for "chunk" is an implementation constraint from SQL - it shouldn't leak to consumer of dao. Therefore this change name from "how" to simple "what".

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDao.kt
```diff
@@ -0,0 +1,42 @@
...
+    open suspend fun getOrderSummaries(siteId: LocalId, remoteOrderIds: List<RemoteId>): List<WCOrderSummaryModel> {
+        return remoteOrderIds.chunked(CHUNK_SIZE)
+            .flatMap { chunk ->
+                getOrderSummariesForSubset(siteId, chunk)
+            }
+    }
+
+    @Query("SELECT * FROM OrderSummaryEntity WHERE siteId = :siteId AND orderId IN (:remoteOrderIds)")
```
ParaskP7: Minor (🔍): Consider splitting this and that other `DELETE` query below that into multiple lines. I am finding it easier to read when it is like that:

```kotlin
    @Query(
        """
        SELECT * FROM OrderSummaryEntity
        WHERE siteId = :siteId
        AND orderId IN (:remoteOrderIds)
        """
    )
```

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDaoTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        val result = sut.getOrderSummaries(
+            site.localId(),
+            listOf(1, 2, 3).asRemoteIds()
+        )
+        assertThat(result).containsOnlyIds(1, 2, 3)
+    }
+
+    @Test
+    fun testGetOrderSummariesForRemoteIds() = runTest {
```
ParaskP7, line 42: Question (❓): Isn't this test an exact (ish) duplicate of the previous one: `testUpsertOrderSummaries`

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions (almost), everything works as expected, good job! 🌟 x 🌟 ^ 🌟 

-----

I have left one question (❓), one suggestions (💡) and one minor (🔍) comment for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

-----

`EXTRAS`
1. Copilot wrote this and it made me 🤔 a bit: `The chunking logic splits IDs into subsets of CHUNK_SIZE—it would be prudent to add a test case that verifies getOrderSummaries handles lists larger than CHUNK_SIZE correctly.`
2. On this second step of testing information `2. Open the site on web, go to Tools > Smooth generator`, I could proceed as I was getting this `critical error` below:

image

Any ideas, did that happen to you too? 🤔

FYI: I tried that 3 times already, 3 different sites... 🤷

ParaskP7:
> Any ideas, did that happen to you too? 🤔

Btw @wzieba , maybe this here and me being just unlucky: p1752505897072599-slack-C02TDTD1GCX 🤷 

PS: Will try again tomorrow... 🤞

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
wzieba: Could you please confirm that it doesn't happen on `trunk`?

ParaskP7: 👋 @wzieba , how do you suggest I confirm that this doesn't happen on `trunk`? 🤔

FYI: To my understanding, I should uninstall/install the app to get a similar behavior, an empty `WCOrderSummaryModel` table, correct? And, as far as I remember, whenever I uninstall/install the app from scratch, when navigating to the `Orders` screen, the list of orders gets auto-populated for me, no need for a manual pull-to-refresh from my side.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDaoTest.kt
wzieba: The idea was to test if `upsertOrderSummaries` won't return _all_ summaries every time, but I agree the naming here is confusing and anyway `testUpsertOrderSummaries` looks redundant, so I'll remove it 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDao.kt
wzieba: 👍 cdb80ffafa4ea09c9a04325a817a13e1684d8dc5

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
wzieba: Ah sorry, I understand now. I wonder, we can either:
- leave it as it is (one time pull-to-refresh shouldn't be bothering)
- move the data from WellSql to Room database before removing the table (maybe worth experimenting with a solution for this as we'll have to do it anyway for `SiteModel`)
- request summaries fetch _if_ database has orders, but not summaries (it'll be one-time case, only after initial run after this migration)

I'll check how time-consuming would be the second point .

wzieba:
> Copilot wrote this and it made me 🤔 a bit: The chunking logic splits IDs into subsets of CHUNK_SIZE—it would be prudent to add a test case that verifies getOrderSummaries handles lists larger than CHUNK_SIZE correctly.

I was thinking about test for this method, but rather to verify if it really runs in a transaction - I wasn't sure that `@Transaction` annotation was enough to make it happen. So I experimented with a test, but eventually decided to not add it as I thought it's too complex. But this comment made me think that this could be actually useful, so here it is: fe950c8

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
ParaskP7: FYI: This is now done (thanks): 0302babe3d8b63344dfa9a09b0170a610c1f8f38

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
ParaskP7: Thanks for the reply @wzieba ! 🙏

> leave it as it is (one time pull-to-refresh shouldn't be bothering)

1. Can't disagree too much, it shouldn't be too bothering indeed.
2. Then again, in trying to think as a non-tech user, I wonder if they'll get scared not having any orders shown after an app update... And how quickly would they find pull-to-refresh to get their soul back into their body... 😅 🤷

FYI: I think personally this was they first time I used pull-to-refresh on this screen, but can't be too sure, thinking that the order screen got auto-refreshed for me every time I visit this screen, again, not sure... 🤔

> move the data from WellSql to Room database before removing the table (maybe worth experimenting with a solution for this as we'll have to do it anyway for SiteModel)

Maybe this is indeed worth experimenting and get some initial feedback on the process... 🤔

> request summaries fetch if database has orders, but not summaries (it'll be one-time case, only after initial run after this migration)

Interesting idea, maybe that would work... 🤔

---

> I'll check how time-consuming would be the second point .

👍 and btw, maybe it's a good time to ask someone from the product team for some extra feedback on that, wdyt? 💭

ParaskP7:
👋 @wzieba !

> I was thinking about test for this method, but rather to verify if it really runs in a transaction - I wasn't sure that @Transaction annotation was enough to make it happen. So I experimented with a test, but eventually decided to not add it as I thought it's too complex. But this comment made me think that this could be actually useful, so here it is: https://github.com/woocommerce/woocommerce-android/commit/fe950c85d2a3a37ba366e63849d0ad2d91493a3c

I just reviewed this `OrderSummaryDaoChunksTest`, thanks for doing that, an awesome way you found there! 🥇 

1. Question (❓): Is this ` assertThat(normalizedLogs).contains(buildString { ... }` really needed? I am just trying to understand why you added it. Isn't this testing implementation details of how Room does it? Wouldn't just testing that adding 250 order summaries and then retrieving them be enough? Understand that you also want to test the `chunked` functionality, and if want to keep that please do, I was just not expecting that when writing/pointing to that comment. 🤔
2. Suggestion (💡): Just to keep this test simpler, I would extract the whole `chucked` test login into its own assert method, maybe that will further help:

```kotlin
    @Test
    fun `given more than 200 order IDs when getOrderSummaries is called then it chunks input and returns all results in a single transaction`() =
        runTest {
            val orderSummaries = (1..250).asOrderSummaries(siteId)
            dao.upsertOrderSummaries(orderSummaries)

            val remoteIds = orderSummaries.map(WCOrderSummaryModel::orderId)

            // Call your method that chunks with a for loop and is annotated with @Transaction
            val result = dao.getOrderSummaries(siteId, remoteIds)

            assertChunked()
            assertThat(result).containsExactlyInAnyOrderElementsOf(orderSummaries)
        }

    private fun assertChunked() {
        val expectedFirstChunk = createExpectedQuery(200)
        val expectedSecondChunk = createExpectedQuery(50)

        val normalizedLogs = logs.lines()
            .map { it.trim() }
            .filter { it.isNotEmpty() }
            .joinToString("\n")
        assertThat(normalizedLogs).contains(
            buildString {
                appendLine(expectedFirstChunk)
                appendLine(expectedSecondChunk)
                appendLine("TRANSACTION SUCCESSFUL")
                appendLine("END TRANSACTION")
            }
        )
    }
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
wzieba: This is addressed via: 2131fd791d3febb78e017909fe0474379ae94cb1 and 0b4bb345e45b64e7adf41d3eb06dbd5d47c19ddc. I think this is an okay trade off, but I'll consult the product team for the opinion

wzieba:
> Wouldn't just testing that adding 250 order summaries and then retrieving them be enough?

IMO it wouldn't. With this kind of test, if you remove completely the chunking logic, the test will still be successful.

---

I understand your concern though, this was an unconventional proposal and not a unit test, bur rather an integration one (that's why it was in a different class). To make things simpler, here's a different proposition: 4b9b09b

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
wzieba: p1753213556619899-slack-CGPNUU63E

ParaskP7: Thanks @wzieba , I guess we are going with option 2 after all, which I am too, in favor of! 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/ShouldUpdateOrdersList.kt
```diff
@@ -9,9 +10,15 @@ import javax.inject.Inject
 
 class ShouldUpdateOrdersList @Inject constructor(
     private val lastUpdateDataStore: LastUpdateDataStore,
-    private val listStore: ListStore
+    private val listStore: ListStore,
+    private val appPrefs: AppPrefsWrapper
 ) {
     suspend operator fun invoke(listDescriptor: ListDescriptor): Boolean {
+        if (!appPrefs.orderSummaryMigrated) {
```
ParaskP7, line 19: 1. Praise (❤️): Nicely done! 🥇 
2. Suggestion (💡): Consider adding a `TODO` comment, with a new associated AINFRA sub-task, a child of [AINFRA-558](https://linear.app/a8c/issue/AINFRA-558), to remove this whole logic few versions from now.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/OrderSummaryDaoTest.kt
```diff
@@ -0,0 +1,105 @@
...
+
+        val result = sut.getOrderSummaries(
+            site.localId(),
+            listOf(RemoteId(1L), RemoteId(2L), RemoteId(3L))
+        )
+        assertThat(result).isEmpty()
+    }
+
+    @Test
+    fun testQueryingLargeNumberOfOrderSummaries() = runTest {
```
ParaskP7, line 81: 👍 ([context](https://github.com/woocommerce/woocommerce-android/pull/14315#issuecomment-3104605936))

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/orders/list/ShouldUpdateOrdersListByStoreIdTest.kt
```diff
@@ -86,4 +93,14 @@ class ShouldUpdateOrdersListByStoreIdTest : BaseUnitTest() {
...
     }
+
+    @Test
+    fun `when fetch after migration wasn't yet done, then list should be updated`() = runTest {
+        appPrefs.orderSummaryMigrated = false
+        val listDescriptor = WCOrderListDescriptor(SiteModel())
+
+        val result = sut.invoke(listDescriptor = listDescriptor)
+
+        assertTrue(result)
```
ParaskP7, line 104: 1. Suggestion (💡): Consider adding `assertTrue(appPrefs.orderSummaryMigrated)`, which also asserts that the boolean value of `orderSummaryMigrated` is also being updated to `true` as part of `sut.invoke(...)` from when `false`. 
2. Suggestion (💡): Consider adding a complimentary test for the case when `appPrefs.orderSummaryMigrated = true`, where at the end you'll assert that the `result` can be `false` with `orderSummaryMigrated` remaining `true`.

ParaskP7 (approved):
👋 @wzieba !

I reviewed and testing this PR once more and everything works as expected, thanks for the update, nicely done! 🚀 x 🚀 ^ 🚀 

---

FYI: As per my first attempt [here](https://github.com/woocommerce/woocommerce-android/pull/14315#pullrequestreview-3016862951), after generating a `Jurassic.Ninja` site, with `WooCommerce Smooth Generator` checked (☑️), I still can't generate a great number of orders using `Tools` > `Smooth generator`. I keep getting the:

`There has been a critical error on this website. Please check your site admin email inbox for instructions. If you continue to have problems, please try the support forums.`

PS: I trust you did that yourself and it worked as expected, with the migrated `chunked` logic and all. Still, if you want, we can pair so that you show me how its done... 🙏

wzieba:
Just to document, here's the demo of the "first run after migration"


https://github.com/user-attachments/assets/bcb1280f-e0b6-4655-a2d9-0ae19580a04e

PR #14318, opened 2025-07-14
Fix crash when navigating after double click

malinajirka:
Fixes #WOOMOB-790

### Description

This PR replaces `navigate` with `navigateSafely` to avoid crashes after double-clicking on one of the items in the PaymentSelection screen.

### Steps to reproduce

1. Open detail of an unpaid order
2. Tap on Collect Payment
3. Tap on `Card Reader` twice very fast - I kind of tapped with two fingers, but one finger was few milliseconds late. 
4. Notice the crash

### Testing information

Repeat the above steps and verify the navigation works but the app doesn't crash after double click

The above

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,7 +3,7 @@
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 22.9
 -----
-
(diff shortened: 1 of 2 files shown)


kidinov (approved):
LGTM!

PR #14302, opened 2025-07-10
Temporarily disable end-to-end tests in build configuration

kidinov:
Related to WOOMOB-789

Check the CI to ensure that the end-to-end tests were not run in instrumented-tests

PR #14319, opened 2025-07-14
Fix flaky e2e tests

wzieba:
Closes: AINFRA-903
Reverts: https://github.com/woocommerce/woocommerce-android/pull/14302

### Description

This PR is an attempt to address flakiness of e2e tests that the team lately observed. It will
- update Firebase Test Lab device to non-deprecated
- enable Android orchestrator
- bump test dependencies
- decrease tests timeout 
- repeat a test once if it fails

### Testing information

You can repeat instrumentation tests job on CI, or run `./gradlew assembleVanillaDebug assembleVanillaDebugAndroidTest && ./gradlew runFlank` and see that tests don't fail.

I've executed instrumented tests 5 times (+ more during work on this PR) and they all passed

Screenshot 2025-07-14 at 13 28 01


WooCommerce/build.gradle
```diff
@@ -19,13 +19,17 @@ fladle {
...
-            "notPackage com.woocommerce.android.e2e"
+            "notPackage com.woocommerce.android.e2e.tests.screenshot"
     ]
     devices = [
-            ["model": "Pixel2.arm", "version": "30"]
+            ["model": "MediumPhone.arm", "version": "33"]
     ]
     localResultsDir = "$rootDir/build/instrumented-tests"
 
+    useOrchestrator = true
```
wzieba, line 29: Orchestrator will help with test runs isolation. Any performance drawbacks should be mitigated by tests sharding.

WooCommerce/build.gradle
```diff
@@ -19,13 +19,17 @@ fladle {
...
+    testTimeout = "5m"
```
wzieba, line 30: Our longest e2e tests are no longer than ~2m30s so I think 5m timeout is good enough (instead of default 15m).

Image

WooCommerce/build.gradle
```diff
@@ -19,13 +19,17 @@ fladle {
...
+    flakyTestAttempts = 1
```
wzieba, line 31: [docs](https://cloud.google.com/sdk/gcloud/reference/firebase/test/ios/run#--num-flaky-test-attempts)

I think it's _acceptable_ to repeat an e2e test _once_ in our case: they rely (at least partially) on backend sites and this might be one of the sources of flakiness.

gradle/libs.versions.toml
```diff
@@ -39,7 +39,7 @@ apache-commons-text = '1.10.0'
 apache-http-client-android = '4.3.5.1'
 automattic-about = '1.7.0'
 automattic-measure-builds = '3.2.1'
-automattic-tracks = '6.0.3'
+automattic-tracks = '6.0.4'
```
wzieba, line 42: This bumps the underlying Sentry SDK in hope to address a native crash during tests:

Image

kidinov:
@wzieba Thanks for taking care of this!

I reviewed the changes, and they make sense to me. However, I’m puzzled as to why they were failing - could it be an issue with the outdated emulator and its network settings?

wzieba:
No problem @kidinov ! 

> I’m puzzled as to why they were failing - could it be an issue with the outdated emulator and its network settings?

Yes, I think you're right. As @ParaskP7 mentioned in a Slack thread, this flakiness started to appear around the time of some other changes in FTL - e.g. `Nexus5` device was completely decommissioned (we had to fix this in Aztec). It _seems_ there might've been some changes to `Pixel2.arm`, but I can't find the source of this.

I hope that changing device to default + other changes will help eliminate (or reduce to reasonable levels) flakiness of these e2e tests. If no, we'll dig more 👍 .

Do you mind approving the PR?

kidinov (approved):
LGTM!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

wzieba:
Thanks! I'll wait a bit with merging this, as we discussed with Petros on DM.

WooCommerce/build.gradle
```diff
@@ -19,13 +19,17 @@ fladle {
     variant = "vanillaDebug"
     serviceAccountCredentials = rootProject.file(".configure-files/firebase.secrets.json")
     testTargets = [
-            "notPackage com.woocommerce.android.e2e"
+            "notPackage com.woocommerce.android.e2e.tests.screenshot"
     ]
     devices = [
-            ["model": "Pixel2.arm", "version": "30"]
+            ["model": "MediumPhone.arm", "version": "33"]
```
ParaskP7: 1. Question (❓): Why not using the latest version `35` instead? 🤔

image

2. Question (❓): Why not using a `PHYSICAL` device instead? 🤔

Is this all (maybe) because of [Fladle](https://runningcode.github.io/fladle/)? 💭

ParaskP7:
> Thanks! I'll wait a bit with merging this, as we discussed with Petros on DM.

I just checked all the changes in more detail and all LGTM, nicely done @wzieba , just have this one [comment](https://github.com/woocommerce/woocommerce-android/pull/14319#discussion_r2209522952), nothing else for you. 🙏

WooCommerce/build.gradle
wzieba: We used a virtual device before. I don't know if there are any advantages of using physical in our case.

Sure, we can use `35`

ParaskP7: > I don't know if there are any advantages of using physical in our case.

Maybe a physical device is more robust, not sure as well, but wouldn't hurt to try, wdyt? 🤔 Worse case we revert that to a virtual again...

> Sure, we can use 35

👍

wzieba: Let's try physical if the test suite will continue to be flaky 👍

PR #14320, opened 2025-07-14
[WOOMOB-839] Implement Test Scanner screen logic + error states

malinajirka:
WOOMOB-839

### Description

This PR implements the logic for detecting scanned barcode on the Test Scanner step of the setup flow + the UI for the error states.

Note 1 - I haven't added the barcode itself on the UI yet, since I assume we'll receive a vector drawable from Wagner.
Note 2 - I noticed the setup flow restarts automatically when I connect a reader in the middle of the flow - I'll report this as a separate issue.

### Testing information

1. Open POS
2. Tap on More menu -> Setup a barcode scanner
3. Select one of the scanners
4. Move to the Test your scanner step
5. Wait 11 seconds -> Notice the screen is replaced with `No scan data found yet` state.
6. Scan an invalid barcode, eg:
Screenshot 2025-07-14 at 15 57 34

7. Notice an error screen is shown
8. Tap on retry and move to the Test your scanner step
9. Scan a valid barcode
Screenshot 2025-07-14 at 15 57 40

10. Notice the success screen is shown

I've tested the above + I also tested moving back and forth between the screens, scanning valid barcode on the first try, scanning valid barcode on the initial Test your scanner step, and few more combinations.

### Images/gif

https://github.com/user-attachments/assets/f1469bd6-2876-4df1-95ab-faa23c467a8d


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModel.kt
```diff
@@ -227,19 +278,42 @@ class WooPosScanningSetupViewModel @Inject constructor(
...
+                    currentStep = createTestYourScannerTimeoutStep()
                 )
             }
         }
     }
 
     companion object {
-        private const val AUTO_NAVIGATION_DELAY_MS = 3000L
+        private const val AUTO_NAVIGATION_DELAY_MS = 10000L
+        private const val TEST_BARCODE_EAN13 = "1234567890128"
```
kidinov, line 276: Have you considered using `barcode-scanning` which we have in the dependencies already to generate the barcode (and QR codes too)? Advantages I see:
* We can change the values easily
* No need to manage all different QR/barcodes based on a reader type. We just need to keep values in the states
* Probably easy to adapt to dark/light mode
* It's easy to make it with our own icon inside (to make it look cool, the same as in the scan to pay)
* We can generate appropriate sizes for specific screens
* APK size

Check `QRCode` composable - we do this already there, so for QRs we can reuse it

kidinov (approved):
LGTM! I would just recommend generating both bar and qr codes in the code due to the reasons given in the comment. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModel.kt
kidinov: I actually noticed that we use `zxing` in `QRCode` but I guess it doesn't matter

malinajirka: Thanks for the suggestion, I like the approach! I'll give that a try and see if/how it works.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka:
Thanks for the review and suggestions @kidinov!

I've implemented some changes:
- The test barcode is generated in compose now.
- TestScanner and TestScannerTimeout steps reuse the same composable.
- Barcode is properly rendered in both light and dark mode.

Please, make sure to also test the QR Payment flow for regression.

Thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/QRCode.kt
```diff
@@ -70,23 +74,60 @@ fun QRCode(
...
+            ),
+        painter = rememberBarcodeBitmapPainter(
+            content,
+            barcodeFormat = BarcodeFormat.EAN_13,
+            widthDp = widthDp,
+            heightDp = heightDp,
+            codeColor = codeColor,
+            overlayId = overlayId,
+        ),
+        contentDescription = "EAN13 Barcode Code",
```
kidinov: Won't be localised. Shell we extract this?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/QRCode.kt
```diff
@@ -3,7 +3,6 @@ package com.woocommerce.android.ui.compose.component
 import android.content.res.Configuration
```
kidinov, line 3: Shell we rename the file as it's not only QRcode anymore?

kidinov (approved):
Thanks for the change! I left 2 minor comments; feel free to merge

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/QRCode.kt
malinajirka: Good catch, renamed in https://github.com/woocommerce/woocommerce-android/pull/14320/commits/2584905b157199f1f2691941f24183bf9dd0afb2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/QRCode.kt
malinajirka: Fixed in https://github.com/woocommerce/woocommerce-android/pull/14320/commits/fd461b2f0e8f2747c27b00fad1ef6f1254ede6fb

PR #14322, opened 2025-07-14
[WOOMOB-842][Woo POS][Barcodes] Other device opens Scanning info state instead of the dialog

kidinov:
WOOMOB-842

### Description

This PR adds navigation to the Scanning info state instead of scanning info dialog when "other" device is selected

The main changes include:

- Removed barcode info dialog handling from the setup flow and instead navigate directly to the Scanner setup info step when "Other" device is selected
- Refactored navigation to use a state-based approach with previousStep tracking for proper back navigation
- Updated the success step to use primary button instead of secondary button for "More Information" action
- Moved the sealed class for scanning setup UI events to a separate file for better organization
- Added some unit tests of the VM

### Testing information
1. Open POS
2. Select "Other" device - verify it goes directly to Scanner setup info step instead of showing barcode dialog
3. Test back navigation throughout the flow - verify each step goes back to the correct previous step
4. Test the complete flow with different device selections to ensure proper navigation

Above

https://github.com/user-attachments/assets/7fa551f6-4c5b-4964-9097-1e8d099c43d1

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModel.kt
@@ -217,10 +217,3 @@ class WooPosScanningSetupViewModel @Inject constructor(
         private const val AUTO_NAVIGATION_DELAY_MS = 3000L
     }
 }
-
-sealed class WooPosScanningSetupUiEvent {
-    data object OnPrimaryButtonClicked : WooPosScanningSetupUiEvent()
-    data object OnSecondaryButtonClicked : WooPosScanningSetupUiEvent()
-    data object OnOpenBluetoothSettings : WooPosScanningSetupUiEvent()
-    data class OnDeviceSelected(val device: BarcodeReaderDevice) : WooPosScanningSetupUiEvent()
-}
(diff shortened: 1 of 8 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModel.kt
```diff
@@ -110,7 +108,7 @@ class WooPosScanningSetupViewModel @Inject constructor(
             }
 
             is ScanningSetupStep.ScannerSetupSuccess -> {
-                error("Primary button should not be available on ScannerSetupSuccess step")
+                error("Not implemented yet")
```
kidinov, line 111: We'd need to open a new state here which doesn't exist yet

image

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModelTest.kt
```diff
@@ -0,0 +1,332 @@
...
+            assertThat(previousStep.previousStep)
+                .isInstanceOf(ScanningSetupStep.DeviceSelection::class.java)
+        }
+
+    private fun createViewModel(): WooPosScanningSetupViewModel {
+        setupMockResourceProvider()
+        return WooPosScanningSetupViewModel(resourceProvider)
+    }
+
+    private fun setupMockResourceProvider() {
```
malinajirka, line 278: 🔍 Considering we are not using any of these values in the tests, it just adds a maintenance burden. I'd consider replacing these with `whenever(resourceProvider.getString(any()).thenReturn("")`. Wdyt?

malinajirka (approved):
LGTM :shipit:

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModelTest.kt
kidinov: I think to add some tests in the following PRs that actually validate that the correct resources are passed, so the actual strings will be used

PR #14331, opened 2025-07-15
[WOOMOB-852][Woo POS][Barcodes] Disable barcode listener on cart when set up flow is visible

kidinov:
WOOMOB-852

### Description

* Disables barcode listener in the cart when set up flow is visible
* Fixes the issue when on configuration change the flow was reseted to the beginning

### Steps to reproduce

* Scan a barcode when scanning set up flow is visible
* Change dark/light mode when scanning set up flow is visible

### Testing information

Check both cases from the steps to reproduce

### Images/gif

https://github.com/user-attachments/assets/68821522-b09a-487f-a8b5-2d166683a0e0

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupState.kt
```diff
@@ -8,7 +8,6 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class WooPosScanningSetupState(
-    val isVisible: Boolean = false,
```
kidinov, line 11: Was unused

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
```diff
@@ -80,8 +82,20 @@ fun WooPosScanningSetupDialog(
     val viewModel = hiltViewModel<WooPosScanningSetupViewModel>()
     val context = LocalContext.current
 
-    LaunchedEffect(isVisible) {
-        viewModel.resetToInitialState()
+    val isClosing = remember { mutableStateOf(false) }
+
+    LaunchedEffect(isClosing.value) {
```
malinajirka, line 87: Nice catch!

malinajirka (approved):
LGTM!

PR #14332, opened 2025-07-15
Issue/woomob 843 woo posbarcodes barcodes for pos scanner setup

malinajirka:
WOOMOB-868

### Description

I ended up making more changes than I originally planned. However, vast majority of the changed lines are test and move of the UiStates definitions. The PR should hopefully still be relatively easy to follow. Let me know if not and I'll split it.

This PR refactors the WooPosScanningSetupViewModel
- Moves state definitions into the state class.
- Extracts the next/previous step logic into a separate class + adds a bunch of tests to verify all the combinations.
- Replaces strings for UI with R.string.[id] to avoid having to use companion objects helping keep the state definitions clean.
- Adds some missing tests to the VM.

All these changes have two main goals:
1. It'll be easy to have a different flow for different readers.
2. The VM was growing and splitting helped keeping even the test classes easier to get oriented in.

### Testing information

Requires full re-testing of the Barcode Setup flows - having said that, I assume we'll do a test plan testing so I don't think it's necessary to spent more than 10 minutes on the testing.

I've tried to test all the states of the setup flow that came to my mind.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupViewModelTest.kt
```diff
@@ -34,299 +35,305 @@ class WooPosScanningSetupViewModelTest {
...
+            // THEN
+            assertThat(viewModel.state.value.currentStep).isEqualTo(mockNextStep)
         }
 
     private fun createViewModel(): WooPosScanningSetupViewModel {
-        setupMockResourceProvider()
-        return WooPosScanningSetupViewModel(resourceProvider)
-    }
-
-    private fun setupMockResourceProvider() {
```
malinajirka, line 278: These are not longer needed since the UIState contains just the id not the string itself.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
```diff
@@ -0,0 +1,79 @@
...
+        val nextIndex = currentIndex + 1
+        if (nextIndex >= stepSequence.size) {
+            error("No next step available for device: $device, current step: $currentStep")
+        }
+
+        return stepSequence[nextIndex]
+    }
+
+    fun getPreviousStep(device: BarcodeReaderDevice, currentStep: ScanningSetupStep): ScanningSetupStep {
+        // Handle special branching steps that aren't in the main sequences
```
kidinov, line 28: np: probably redundant

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
```diff
@@ -0,0 +1,79 @@
...
+        return when (currentStep) {
+            is ScanningSetupStep.TestYourScannerTimeout,
+            is ScanningSetupStep.TestYourScannerScanFailed -> {
+                getPreviousStep(device, ScanningSetupStep.TestYourScanner)
+            }
+            else -> {
+                val stepSequence = ScannerConfigurations.getStepSequence(device)
+                val currentIndex = stepSequence.indexOf(currentStep)
+
+                if (currentIndex <= 0) {
```
kidinov, line 38: We should probably crash here if `currentIndex`, the same as in `getNextStep`? -1 - means not found.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        return stepSequence[nextIndex]
+    }
+
+    fun getPreviousStep(device: BarcodeReaderDevice, currentStep: ScanningSetupStep): ScanningSetupStep {
+        // Handle special branching steps that aren't in the main sequences
+        return when (currentStep) {
+            is ScanningSetupStep.TestYourScannerTimeout,
+            is ScanningSetupStep.TestYourScannerScanFailed -> {
+                getPreviousStep(device, ScanningSetupStep.TestYourScanner)
```
kidinov: Shell it be on navigation back 

* TestYourScannerTimeout -> TestYourScanner
* TestYourScannerScanFailed -> TestYourScannerTimeout

?

Now it opens the step which is before `TestYourScanner`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupState.kt
```diff
@@ -18,92 +19,231 @@ data class WooPosScanningSetupState(
...
         data class PairYourScanner(
-            val title: String,
-            val message: String,
-            @DrawableRes val iconRes: Int,
-            val primaryButtonText: String,
-            val secondaryButtonText: String,
-            val bluetoothSettingsButtonText: String,
-            override val previousStep: ScanningSetupStep
-        ) : ScanningSetupStep()
+            @DrawableRes val iconRes: Int = R.drawable.ic_woopos_bluetooth_settings
```
kidinov: Do you expect this icon to be specific per reader? It looks like all the resources that are not specific are hardcoded, so maybe it also should be an object?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
malinajirka: Good question, and I'm not sure what is expected. I was working with the assumption that all `TestYourScanner...` are the same step, so if we want to go back, we go back to Pairing. 

Thinking about this more, I'd say the Failed state shouldn't have back at all - only retry => if we read an invalid barcode, we are "sure" the scanner isn't correctly setup. The TestYourScannerTimeout should go back to pairing as it does now, since it's still the same step with the same code as TestYourScanner. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
malinajirka: I think this one is valuable since it might not be clear that there is a main sequence and then branches. I'll keep it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
```diff
@@ -165,21 +165,21 @@ fun WooPosScanningSetupDialog(
...
+                        secondaryButtonText = stringResource(step.secondaryButtonTextRes),
                         onPrimaryClick = { viewModel.onUiEvent(WooPosScanningSetupUiEvent.OnPrimaryButtonClicked) },
                         onSecondaryClick = { viewModel.onUiEvent(WooPosScanningSetupUiEvent.OnSecondaryButtonClicked) }
                     )
 
                     is ScanningSetupStep.ScannerPairModeSetup -> ScannerModeSetupContent(
-                        title = step.title,
-                        message = step.message,
+                        title = stringResource(step.titleRes),
+                        message = stringResource(step.messageRes),
```
kidinov, line 194: It's a parameterised string here, therefore Int probably cannot be used. Initially, they were used everywhere for this reason - I thought that there would be many places like that where we mention the device name.

image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
malinajirka: Good point, fixed in 48eab7a693f058fb26dd7e2775417bab2d81dccc

kidinov (commented):
Looks and tests well! Thanks for the improvement!

I left a couple of comments; please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupState.kt
malinajirka: I missed this, fixed in 9f72bb0c9a09f30fb309601aca1cd3c59f2cf202

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
malinajirka: Good catch! Not sure how I could have missed this - I'll try to come up with a solution that will still allow us to avoid companion objects.

malinajirka: What do you think about this approach @kidinov ?
Screenshot 2025-07-16 at 9 36 01

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
kidinov: I agree with your thinking except that:
* As we implemented it, it's now visually a new step. We play animation, and the size of the dialog changes so it feels and looks like just a new step visually, the same as the others
* While we mostly sure that failed step means something is wrong with the scanner, there is still a small chance that it was an actual keyboard used by mistake

All together, considering that these are rather edge cases, I'd keep it simple implementation and flow-wise and always just navigate back

```
    fun getPreviousStep(device: BarcodeReaderDevice, currentStep: ScanningSetupStep): ScanningSetupStep {
        val stepSequence = ScannerConfigurations.getStepSequence(device)
        val currentIndex = stepSequence.indexOf(currentStep)
        return stepSequence[currentIndex - 1]
    }
```

malinajirka: I believe `val currentIndex = stepSequence.indexOf(currentStep)` will return -1 if the step is one of the side-branches, right?

> As we implemented it, it's now visually a new step. We play animation, and the size of the dialog changes so it feels and looks like just a new step visually, the same as the others
While we mostly sure that failed step means something is wrong with the scanner, there is still a small chance that it was an actual keyboard used by mistake

Good points! Perhaps going to TestYourScanner makes the most sense.

kidinov: > I believe val currentIndex = stepSequence.indexOf(currentStep) will return -1 if the step is one of the side-branches, right?

Oh yes, it will. I sort of liked how it was before - linked list for back navigation. Stack also would be nice. Having said that, I don't know if it's worth a change now

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
kidinov: I guess it's fine, although tbh I am not a big fan of it. It feels like it's better if the VM exposes ready to show data to the view

malinajirka: I know what you mean, but technically I believe it's an anti-pattern => the view might have Themed composable or resources might change during a configuration change so getting resources should be IMO done in the view layer. Having said that, it's more in the sphere of theory since both approaches are likely fine and the downsides are edge-cases not worth worrying about. Having said all that, in this case, since you are not strongly against, I'll opt for the UIString because I want to avoid having to pass `resourceProvider` to all the ScanningSetupSteps - that will save a bunch of boilerplate.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigator.kt
malinajirka: I kind of liked the original approach as well - however, having the definition of all the steps at the compile time IMO outweighs the downsides. I wish we could somehow elegantly include the side-branches in the definition of the flow, that would be cleanest. I didn't come up with a reasonable solution though since the side-branches are used only when certain conditions are met. There will always need to be some business logic behind it => this logic IMO belongs to the VM not the navigator or the state definition.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupState.kt
```diff
@@ -18,92 +20,240 @@ data class WooPosScanningSetupState(
...
+                    ScanningSetupStep.ScannerPairModeSetup(),
+                    ScanningSetupStep.PairYourScanner(deviceName = device.displayNameRes),
+                    ScanningSetupStep.TestYourScanner,
+                    ScanningSetupStep.ScannerSetupSuccess
+                )
+
+                BarcodeReaderDevice.STAR_BSH_20B -> listOf(
+                    ScanningSetupStep.DeviceSelection,
+                    ScanningSetupStep.ScannerHIDModeSetup(qrCodeImageRes = R.drawable.ic_barcode),
+                    ScanningSetupStep.PairYourScanner(deviceName = device.displayNameRes),
```
kidinov, line 38: ‼️ `ScannerPairModeSetup` is missing for this reader. Expected?

malinajirka: Yes, this is expected since we have all-in-one barcode for this reader and that's displayed on the ScannerHidModeSetup => this was actually one of the reasons I realized each scanner will need a different flow.

malinajirka:
Thanks @kidinov for the review! I've implemented the suggestions and fixed the issues you found. It should be all good now, but definitely worth re-reviewing. Thanks!

kidinov (approved):
LGTM!

PR #14237, opened 2025-06-24
`WCTaxClassModel` to `Room`

wzieba:
Fixes: [AINFRA-564](https://linear.app/a8c/issue/AINFRA-564/migrate-wctaxclassmodel)

### Description

This PR migrates `WCTaxClassModel` from `WellSql` to `Room`.

### Testing information

1. Make sure "taxes feature" is enabled on your Woo web instance: `WooCommerce` -> `Settings` -> `General` -> `Enable taxes`
2. Launch the mobile app
3. Go to details of any product, then `Price`
4. Tap on `Tax class`
5. Assert you can see 3, default tax classes: `Reduced rate`, `Standard rate`, `Zero rate`
6. Go to web Woo, `WooCommerce` -> `Settings` -> `Tax`.
7. Add a new tax class to `Additional tax classes`
8. On mobile, close product details, open them again and assert you can see the new class.

### Images/gif

Screenshot 2025-06-25 at 14 21 00

- [ ] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14335, opened 2025-07-16
Fix crash when tax classes are duplicated

hichamboushaba:
Closes WOOMOB-871

### Description
This PR fixes a crash on product details, the crash happens when the store includes duplicate tax classes, and is caused by the recent migration to Room in the PR https://github.com/woocommerce/woocommerce-android/pull/14237

### Steps to reproduce
`wp-admin` doesn't allow adding duplicate taxes, so the duplication is most probably caused by a plugin or manual handling of the site DB, so we'll use Api Faker to simulate this.
1. Open the Api Faker tool from the developer options menu.

<details>
<summary>2. Import the following configuration using the "import from clipboard" feature</summary>

```json
[
  {
    "request": {
      "httpMethod": "GET",
      "id": 0,
      "path": "/wc/v3/taxes/classes",
      "queryParameters": [],
      "type": {
        "type": "wp-api"
      }
    },
    "response": {
      "body": "[\n  {\n      \"slug\": \"standard\",\n      \"name\": \"Standard rate\",\n      \"_links\": {\n        \"collection\": [\n          {\n            \"href\": \"https://hichamwootest.blog/wp-json/wc/v3/taxes/classes\"\n          }\n        ]\n      }\n    },\n  {\n      \"slug\": \"standard\",\n      \"name\": \"Standard rate\",\n      \"_links\": {\n        \"collection\": [\n          {\n            \"href\": \"https://hichamwootest.blog/wp-json/wc/v3/taxes/classes\"\n          }\n        ]\n      }\n    }\n]",
      "endpointId": 0,
      "statusCode": 200
    }
  }
]
```
</details>

3. Open details screen of any product.

### Testing information
- Confirm the app doesn't crash on this branch.

The above.

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TaxClassDao.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TaxClassDao.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.persistence.dao
 
 import androidx.room.Dao
 import androidx.room.Insert
+import androidx.room.OnConflictStrategy
 import androidx.room.Query
 import androidx.room.Transaction
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
@@ -31,6 +32,6 @@ internal abstract class TaxClassDao {
     )
     protected abstract suspend fun deleteAll(siteId: LocalId): Int
 
-    @Insert
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
     protected abstract suspend fun insertAll(classes: List<WCTaxClassModel>)
 }
--- a/libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/TaxClassDaoTest.kt
+++ b/libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/TaxClassDaoTest.kt
@@ -121,4 +121,17 @@ class TaxClassDaoTest {
         val otherSiteResult = sut.getTaxClasses(LocalId(999))
         assertThat(otherSiteResult).containsExactly(differentSiteTaxClass)
     }
+
+    @Test
+    fun `when there are duplicate tax classes, replaceAll keeps the last one`() = runTest {
+        // given
+        val taxClasses = listOf(sampleTaxClass1, sampleTaxClass2, sampleTaxClass1)
+        sut.replaceAll(site.localId(), taxClasses)
+
+        // when
+        val result = sut.getTaxClasses(site.localId())
+
+        // then
+        assertThat(result).containsExactlyInAnyOrderElementsOf(listOf(sampleTaxClass1, sampleTaxClass2))
+    }
 }


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/266">22.7.1</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TaxClassDao.kt
```diff
@@ -31,6 +32,6 @@ internal abstract class TaxClassDao {
     )
     protected abstract suspend fun deleteAll(siteId: LocalId): Int
 
-    @Insert
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
```
hichamboushaba, line 35: This matches the previous behavior we had in WellSql https://github.com/woocommerce/woocommerce-android/blob/13421ddde3f3ee398d45318245478633ab516955/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/taxes/WCTaxClassModel.kt#L13

wzieba: Thanks. We can also use `@Upsert` - something I've used in other DAOs. The difference is subtle in this case.

wzieba (approved):
I thought simple `@Insert` here shouldn't be the problem as we remove all taxes before inserting, but I guess you're right with

> duplication is most probably caused by a plugin or manual handling of the site DB

Thanks for handling this 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TaxClassDao.kt
hichamboushaba: Actually I didn't know about `@Upsert`, it seems this was added a while ago (version 2.5) but I missed it 🤦.

I was curious about the differences between using `@Upsert` and `@Insert(onConflict = OnConflictStrategy.REPLACE)`, and after looking at the generated code and Room source code, it seems that `@Insert(onConflict = OnConflictStrategy.REPLACE)` makes Sqlite handle the conflict itself, as it [generates](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-compiler/src/main/kotlin/androidx/room/writer/DaoWriter.kt;l=443-458;drc=2774c91c93b8573989f8450ccd307330fa0735db;bpv=0;bpt=1) an `INSERT OR REPLACE...` query, while when using `@Upsert`, we generate a simple `INSERT...` query, and then [handle](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-runtime/src/androidMain/kotlin/androidx/room/EntityUpsertionAdapter.android.kt;l=57-64;bpv=0;bpt=0) the conflict on the code side, I'm not sure what's the best approach, but from a performance perspective, I think `INSERT OR REPLACE...` could be better, but not that it matters in this scenario.

Anyway, I was just curious about this and decide to look into, thanks for sharing the `@Upsert` annotation 🙏

wzieba: Sure! Thanks for deep dive in the generated code, I wasn't aware that `@Upsert` will address this in code rather than SQL. 

As long as we don't rely on synthetic primary keys or have foreign keys constraints on this entity I think `@Insert(onConflict = OnConflictStrategy.REPLACE)` is perfectly fine 👍 .

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/TaxClassDaoTest.kt
```diff
@@ -121,4 +121,17 @@ class TaxClassDaoTest {
         val otherSiteResult = sut.getTaxClasses(LocalId(999))
         assertThat(otherSiteResult).containsExactly(differentSiteTaxClass)
     }
+
+    @Test
+    fun `when there are duplicate tax classes, replaceAll keeps the last one`() = runTest {
```
ParaskP7, line 126: Praise (❤️): I great idea to cover this scenario with a test! 🥇

PR #14340, opened 2025-07-16
[WOOMOB-843] Render real setup barcodes in the Barcode Reader Setup Flows

malinajirka:
### Description

This PR:
- Remove the Inateck reader since it likely won't be one of the recommended readers.
- ~Renders actual setup barcodes for the two remaining readers.~ I ended up using vector drawables since the generated QR/DataMatrixes were not accepted by the STAR BSH scanner even though the value of the code was identical. It's safer to have these "hardcoded" so a library update or something like that can't break the setup flow.

### Testing information

Test the setup flows - if you have one of the readers, try to test the barcodes as well.

I've tested the setup flow for the Star BSH reader.

### Images/gif

Uploading Screen Recording 2025-07-16 at 15.49.03.mov…


kidinov (approved):
I reset my BSH reader and set it up again using the steps. It works really great, and the UI feels super fluid and responsive on a 120Hz tablet. It took like 30 seconds to do all the steps.

We'll need to test it with the other recommended readers during the call for testing.

PR #14343, opened 2025-07-16
[POS as a Tab i2] Keep eligibility reason visible during loading state

toupper:
Closes WOOMOB-718

### Description
This PR improves the UX of the POS eligibility screen by preserving and displaying the last known ineligibility reason even when the state transitions to `Loading`. Previously, the error message would disappear during retry, resulting in a subpar user experience.

We now store the last `Ineligible` reason using a `remember + LaunchedEffect` pattern and continue to show the relevant suggestion text while the retry check is in progress.

This ensures a smoother and more informative experience for users attempting to access POS again after a failure.

### Steps to reproduce
1. Set up a site that will trigger an ineligible POS state (e.g. unsupported currency).
2. Open the POS app and reach the eligibility screen.
3. Tap the "Retry" button.
4. Observe that the message is preserved while the retry is loading.

### Testing information
- Alternate workflows: Verified behavior across different ineligibility reasons.
- Edge cases: Confirmed that the message remains visible in `Loading` and disappears only when navigation occurs.
- Affected areas: Eligibility screen for Woo POS.
- Critical flows: Retry flow on eligibility check.
- Not tested: Eligibility logic itself (out of scope).
- Remaining unknowns: None known.

- Manually triggered retry and verified text preservation.
- Confirmed that navigation triggers correctly when state becomes `Eligible`.
- Verified that "Exit POS" still works normally during loading.

### Images/gif

https://github.com/user-attachments/assets/0c9124ff-ed6f-4106-afd4-42624e62bd00

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
@@ -13,6 +13,10 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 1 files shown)


samiuelson (approved):
Tests good 👍

PR #14344, opened 2025-07-16
[Bug][Orders] Keyboard not shown in landscape

samiuelson:
## [Bug] Keyboard doesn't show the first time we tap on the Orders Search on Tablets

WOOMOB-622 

### Description

This PR fixes bug in Orders screen:
> Unlike Products, the keyboard is not shown when tapping on the Orders Search; the user has to tap again on the text field. It would be easier if we just show the keyboard right away.

### Steps to reproduce

1. Go to orders and ensure device is in landscape
2. Tap search icon 
3. Observe the search field expands, however, keyboard is not showing

💡 There was a logic proactively hiding the keyboard in landscape in Ordrers screen. I tested it on both phone and tablet and don't see any sense doing this anymore. Also, there's no such behavior in Products screen.

### Testing information

Verify that keyboard is shown as expected after tapping "search" icon.

Above. 

### Images/gif

—


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListFragment.kt
```diff
@@ -1141,13 +1138,6 @@ class OrderListFragment :
             }, HANDLER_DELAY)
         }
     }
-
-    private fun checkOrientation() {
```
kidinov, line 1145: Wow, 6 years old code

image

Not making it more clear why it's done

kidinov (approved):
I didn't find any explanation why it was done so I think it's fine to remove

LGTM!

PR #14347, opened 2025-07-17
[POS as a Tab i2] Handle woocommerce plugin not found

toupper:
Closes WOOMOB-897

### Description

With this PR we add the case when the WooCommerce plugin is not found when checking the eligibility for POS. We add as well the text to display in this case, as well as the analytics strings.

### Steps to reproduce

This is an edge case, so it's difficult to create the context for not finding the WooCommerce plugin. An easy way is to force showing the eligibility screen by not meeting other requirements, e.g. currency, and then intercepting the plugins call with a proxy and edit the response to remove the woocommerce plugin.

- Open the app
- Tap on the POS tab. If you don't meet the currency requirements, the eligibility screen will be shown.
- Tap on Retry. If you intercepted the network call fetching the plugins and edited to remove the woocommerce plugin, an error should be shown (See screenshot)
`🔵 Tracked: woocommerceandroid_pos_ineligible_ui_retry_tapped, Properties: {"reason":"unknown_wc_plugin",...}
`
should be logged.

### Images/gif

Screenshot_20250717_163823_Woo (Dev)


samiuelson (approved):
LGTM

PR #14354, opened 2025-07-18
[Encrypted Logging] Integrate `automattic-enryptedlogging` 1.0.0

ParaskP7:
Closes:
- [AINFRA-865](https://linear.app/a8c/issue/AINFRA-865)
- [AINFRA-957](https://linear.app/a8c/issue/AINFRA-957)
- [AINFRA-539](https://linear.app/a8c/issue/AINFRA-539)

---

### Description

This PR migrates the current implementation of encrypted logging for this project to the newly versioned [1.0.0](https://github.com/Automattic/EncryptedLogging/releases/tag/1.0.0) standalone [EncryptedLogging](https://github.com/Automattic/EncryptedLogging) library.

FYI: This update makes this library fully 16 KB page size compatible ([context](https://android-developers.googleblog.com/2025/05/prepare-play-apps-for-devices-with-16kb-page-size.html)).

PS:
- `Dependencies Diff`: Note that as part of this change lots of dependencies are being updated, but transitively, because this project is quite behind in terms of dependency updates. 528c7a6bd33d41f9ce7bb6c757f213bd2bcc5f5e + See [Project dependencies changes](https://github.com/woocommerce/woocommerce-android/pull/14354#issuecomment-3089225338) comment
- `Manifest Diff`: I inspected the [diff](https://github.com/woocommerce/woocommerce-android/pull/14354#issuecomment-3089225772) myself and everything LGTM, it just that the elements got reordered due to all the dependency updates. It is indeed unfortunate that we have to do that manually and I just created [AINFRA-966](https://linear.app/a8c/issue/AINFRA-966) to keep track of this work. 

---

### Testing information

0. Using `App Inspection` from AS find the `wp-fluxc.db` and notice that `EncryptedLogModel` table no longer exists.
1. Using `App Inspection` from AS find the `encrypted-log.db` and its sole `EncryptedLogEntity` table. Notice that the table is empty, there are no log to be sent.
2. Prepare the app for a crash. Navigate to `Menu` -> `Settings` -> `Privacy settings` and then switch-on the `Report Crashes` option.
3. Using the patch below, make the app crash. You could navigate to `Menu` -> `Settings` -> `Developer Options` and then click `Crash The App`.

<details>
    <summary>patch</summary>

```diff
Subject: [PATCH] Build: Remove unnecessary encrypted logging proguard rules

The 'automattic-logging' library defines these Proguard rules already,
via its 'consumer-rules.pro' file, and thus those are automatically 
applies to any module that uses this dependency.

consumer-rules.pro: https://github.com/Automattic/EncryptedLogging/blob/
trunk/encryptedlogging/consumer-rules.pro
---
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DeveloperOptionsViewModel.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DeveloperOptionsViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DeveloperOptionsViewModel.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DeveloperOptionsViewModel.kt	(revision 2b8d77c042fff8f0df4629edf9a0e84e1961b97f)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DeveloperOptionsViewModel.kt	(date 1752838777405)
@@ -90,12 +90,25 @@
         )
     )
 
+    private val crashTheApp = flowOf(
+        NonToggleableListItem(
+            icon = R.drawable.ic_cancel,
+            iconTint = R.color.color_primary,
+            label = UiString.UiStringText("Crash The App"),
+            isEnabled = true,
+            onClick = {
+                throw Exception("Developer caused a crash!")
+            }
+        )
+    )
+
     val viewState = combine(
         simulatedCardReaderFlow,
         readerUpdateFrequencyFlow,
         interacPaymentEnabledFlow,
         savedPrivacySettingsOnDialogFlow,
-        apiFakerFlow
+        apiFakerFlow,
+        crashTheApp
     ) { items ->
         DeveloperOptionsViewState(
             rows = items.filterNotNull()
```

</details>

4. You might now notice a new db entry within `App Inspection` from AS on that `DETACHED` app (or not, it depends). Now, switch to the `ATTACHED` app and notice the db empty. This is because this db entry/log has been already sent to Sentry already, on app start and after the app crashed. To really notice the entry, trying crashing the app again, this time put the device on offline mode first. Doing that you'll be able to see the entry on the db, even when switching to the newly created `ATTACHED` app.
5. Open [Sentry](https://a8c.sentry.io) and navigate to `Issues`. Find `woocommerce-android` and check the latest `debug` exceptions. You could search for `DiagnosticCoroutineContextException` (or `Developer caused a crash!`) and check the latest crash. Now navigate at the bottom and check the `uuid`. Verify it is the same `UUID` that you saw on that db entry before.
6. Open [Read Encrypted Logs](https://mc.a8c.com/encrypted-logs.php), copy-paste that `UUID` into `Log UUID`, click `View` and make sure you can read the newly uploaded and now decrypted log.
7. Finally, you could inspect `Logcat` and notice/verify the below entries:
    - Success: `Successfully uploaded encrypted log with uuid: <UUID>`
    - Failure: `Failed to upload encrypted log with uuid: <UUID> [Reason: <ERROR>]`

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/test/kotlin/com/woocommerce/android/util/crashlogging/EnqueueSendingEncryptedLogsTest.kt
```diff
@@ -2,29 +2,27 @@ package com.woocommerce.android.util.crashlogging
...
-import org.wordpress.android.fluxc.action.EncryptedLogAction.UPLOAD_LOG
-import org.wordpress.android.fluxc.store.EncryptedLogStore.UploadEncryptedLogPayload
 import java.io.File
 
 @RunWith(MockitoJUnitRunner::class)
 class EnqueueSendingEncryptedLogsTest {
     private lateinit var sut: EnqueueSendingEncryptedLogs
 
-    private val eventBusDispatcher: Dispatcher = mock()
+    private val encryptedLogging: EncryptedLogging = mock()
```
wzieba, line 22: Pocket Casts has an interesting idea of [FakeEncryptedLogging](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/crashlogging/src/test/java/au/com/shiftyjelly/pocketcasts/crashlogging/fakes/FakeEncryptedLogging.kt) - maybe this is something we could use instead of mock here?

If so, maybe we should put `FakeEncryptedLogging` in a new, test-only library in Automattic-Tracks-Android dependency, to help running tests that use `EncryptedLogging` in all clients.

wzieba (approved):
Works well! I like that this solution is smarter with the minimal API. I leave one non-blocking suggestion, other than that :shipit:

WooCommerce/src/test/kotlin/com/woocommerce/android/util/crashlogging/EnqueueSendingEncryptedLogsTest.kt
ParaskP7: 👋 @wzieba good suggestions! 🥇

> Pocket Casts has an interesting idea of [FakeEncryptedLogging](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/crashlogging/src/test/java/au/com/shiftyjelly/pocketcasts/crashlogging/fakes/FakeEncryptedLogging.kt) - maybe this is something we could use instead of mock here?

This is a good idea, and an opportunity to showcase fakes versus mocks. Having said that, I would like us to do such a change on a separate PR, and thus, making the case for fakes, showcasing how it's done. For that reason, and the fact that I would like to proceed with this change asap, without too many distractions, I am considering this as out-of-scope. 😊

> If so, maybe we should put FakeEncryptedLogging in a new, test-only library in Automattic-Tracks-Android dependency, to help running tests that use EncryptedLogging in all clients.

Oh, that's an excellent idea, I think I like it! 💯

Question (❓): Btw, why using [Automattic-Tracks-Android](https://github.com/Automattic/Automattic-Tracks-Android) for that, and not [EncryptedLogging](https://github.com/Automattic/EncryptedLogging) itself? 🤔

ParaskP7:
👋 @samiuelson !

Just and FYI from my side that in addition to Wojtek ✅ I would like for a product engineer to review this change as well, thus, I'll wait for that extra ✅ before merging this. This is just so you become aware of such change, and then can share that with the rest of the WCAndroid team. 🙏

samiuelson (approved):
LGTM!

WooCommerce/src/test/kotlin/com/woocommerce/android/util/crashlogging/EnqueueSendingEncryptedLogsTest.kt
wzieba: > Question (❓): Btw, why using [Automattic-Tracks-Android](https://github.com/Automattic/Automattic-Tracks-Android) for that, and not [EncryptedLogging](https://github.com/Automattic/EncryptedLogging) itself? 🤔

I mixed up the repos: I agree it should be in EncryptedLogging :)

ParaskP7: Got it, thanks for the clarification on that @wzieba ! 🙏

PR #14357, opened 2025-07-21
[WOOMOB-908] Improve scanning performance in dark mode for Scanner setup flow

malinajirka:
### Description

This PR improves scanners ability to properly scan setup codes by:
- Increasing the empty area around the barcodes
- Automatically increasing brightness for the setup flow - it increases it to the maximum for the whole setup flow, let me know if you believe it'd be better to increase it just for the steps with a barcode. I tried that and it felt a bit off as the brightness was going up and down depending on the step.

### Testing information

Verify the brightness increases when you start the flow and restores back to the original when you dismiss it or background the app.

The above

### Images/gif

https://github.com/user-attachments/assets/4128c701-989d-4c90-adf9-5b1ec0c098ea

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
@@ -16,14 +16,22 @@ fun BrightnessControl(temporarilyIncreaseToMax: Boolean = true) {
     var originalBrightness by remember { mutableStateOf<Float?>(null) }
 
     LaunchedEffect(temporarilyIncreaseToMax) {
+        val activity = context as? Activity ?: return@LaunchedEffect
+        val window = activity.window
+
         if (temporarilyIncreaseToMax) {
-            val activity = context as? Activity ?: return@LaunchedEffect
-            val window = activity.window
-
-            originalBrightness = window.attributes.screenBrightness
-
+            if (originalBrightness == null) {
+                originalBrightness = window.attributes.screenBrightness
+            }
             window.attributes = window.attributes.apply {
                 screenBrightness = 1.0f
+            }
+        } else {
+            originalBrightness?.let {
+                window.attributes = window.attributes.apply {
+                    screenBrightness = it
+                }
+                originalBrightness = null
             }
         }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
@@ -1,0 +1,41 @@
+package com.woocommerce.android.ui.compose.component
+
+import android.app.Activity
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+
+@Composable
+fun BrightnessControl(temporarilyIncreaseToMax: Boolean = true) {
+    val context = LocalContext.current
+    var originalBrightness by remember { mutableStateOf<Float?>(null) }
+
+    LaunchedEffect(temporarilyIncreaseToMax) {
+        if (temporarilyIncreaseToMax) {
+            val activity = context as? Activity ?: return@LaunchedEffect
+            val window = activity.window
+
+            originalBrightness = window.attributes.screenBrightness
+
+            window.attributes = window.attributes.apply {
+                screenBrightness = 1.0f
+            }
+        }
+    }
+
+    DisposableEffect(Unit) {
+        onDispose {
+            if (temporarilyIncreaseToMax && originalBrightness != null) {
+                val activity = context as? Activity ?: return@onDispose
+                activity.window.attributes = activity.window.attributes.apply {
+                    screenBrightness = originalBrightness ?: -1.0f
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+
+@Composable
+fun BrightnessControl(temporarilyIncreaseToMax: Boolean = true) {
+    val context = LocalContext.current
+    var originalBrightness by remember { mutableStateOf<Float?>(null) }
+
+    LaunchedEffect(temporarilyIncreaseToMax) {
```
harrytmthy: This increases brightness when `temporarilyIncreaseToMax` is true, but doesn't restore it when the value is changed back to false, leaving the screen at max brightness permanently across recompositions.

Would it make sense to handle both branches in this effect to ensure the brightness is always restored based on the latest flag?

```kotlin
LaunchedEffect(temporarilyIncreaseToMax) {
    val activity = context as? Activity ?: return@LaunchedEffect
    val window = activity.window

    if (temporarilyIncreaseToMax) {
        if (originalBrightness == null) {
            originalBrightness = window.attributes.screenBrightness
        }
        window.attributes = window.attributes.apply {
            screenBrightness = 1.0f
        }
    } else {
        originalBrightness?.let {
            window.attributes = window.attributes.apply {
                screenBrightness = it
            }
            originalBrightness = null
        }
    }
}
```

malinajirka: Thank you for the review @harrytmthy, external contributions are always very much appreciated!

Your suggestion makes sense to me, I've updated the code in ca91609683a24e5b0b5607cec2dab49ccc0e87ec. I believe it doesn't have any impact on the behavior, since the DisposableEffect takes care of restoring the brightness. Having said that, it makes the component prepared for cases in which we change the `temporarilyIncreaseToMax` between compositions, so it's definitely a good improvement. 🙇

harrytmthy (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+
+@Composable
+fun BrightnessControl(temporarilyIncreaseToMax: Boolean = true) {
```
kidinov: Should be probably named `WooPosBrightnessControl` to follow the convention

kidinov: Nope, should not - not in the POS. I'd put this into the POS though - fewere dependencies and nobody apparently uses this anyway for now except us

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
```diff
@@ -141,6 +142,8 @@ fun WooPosScanningSetupDialog(
             id = R.string.woopos_scanning_setup_dialog_content_description
         )
     ) {
+        BrightnessControl(temporarilyIncreaseToMax = true)
```
kidinov, line 145: That works only once per POS:
* Open the dialog
* Notice brightness goes up
* Close it and open again
* Notice the brightness stays
*

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
```diff
@@ -0,0 +1,49 @@
...
+    val context = LocalContext.current
+    var originalBrightness by remember { mutableStateOf<Float?>(null) }
```
kidinov: It looks like that if during the dialog usage the brightness was changed manually then max brightness is not restored even if the dialog closed and reopened again

kidinov (approved):
I had left a couple of comments, but now scanning experience is much better. It appears that the most significant improvements are the larger border, as even with lower brightness, it scans very well.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
malinajirka: I'm not able to reproduce this behavior. The way it behaves on my tablet:

1. Open POS, brightness 0.5
2. Open dialog, brightness 1.0
3. Change manually to brightness 0.7
4. Close dialog, restored back to brightness 0.5
5. Open dialog, brightness 1.0
6. Close dialog, brightness 0.5

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
malinajirka: Update in 433dfc8b20123c81c17490ec29bca77ba8b0dd67

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/BrightnessControl.kt
kidinov: The video doesn't represent the brightness, but after the dialog was closed once it always was low even when it was open again. I think this maybe even bug on my tablet (airplane mode was just a mistake and doesn't influence the behavior)

https://github.com/user-attachments/assets/868ae0aa-5ecf-44b7-9d0c-baea57ab1250

malinajirka: Hmm, seems like something specific to your device - I'm not able to reproduce this and tbh I'm quite surprised by this behavior. When the dialog is closed, the composable is disposed and a brand new one is created - it almost seems like the system on your tablet doesn't allow programmatically changing brightness after it was changed manually.

PR #14358, opened 2025-07-21
[WOOMOB-841][Woo POS][Barcodes] Keyboard settings hint when software keyboard is not shown

kidinov:
WOOMOB-841

### Description
The PR adds a hint that explains what to do when the SW is not visible and a reader connected

@malinajirka @samiuelson there is not any reliable way of determining if the floating toolbar keyboard is visible; therefore, I show the hint in this case too. Considering that, I am thinking if this is even worth it. Wdyt?

Btw, settings for this is in "languages input -> manage keyboards -> Gboard (likely) -> physical keyboard (when connected)."

### Steps to reproduce
1. Open WooPos items screen
2. Open search with hardware keyboard (reader) connected
3. Observe the keyboard hint appears with animation
4. Dismiss the hint and observe smooth disappearance
5. Click on the hint

### Testing information
- Ensure hint appears when hardware keyboard is connected and search is open
- Confirm hint dismisses properly
- Confirm the settings do open the setting that do the change
- Validate the copy

https://github.com/user-attachments/assets/f0b2a5f7-a4cf-4c57-961f-ebd56fa60b3f

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFloatingKeyboardHint.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFloatingKeyboardHint.kt
@@ -7,136 +7,175 @@ import androidx.compose.animation.slideInVertically
 import androidx.compose.animation.slideOutVertically
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Close
 import androidx.compose.material.icons.filled.Info
-import androidx.compose.material.icons.outlined.Close
-import androidx.compose.material3.Card
-import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.LinkAnnotation
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.text.withLink
+import androidx.compose.ui.text.withStyle
 import androidx.compose.ui.unit.dp
+import androidx.constraintlayout.compose.ConstraintLayout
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosKeyboardStatus.kt
```diff
@@ -0,0 +1,79 @@
...
+}
+
+@Composable
+fun rememberKeyboardStatus(): WooPosKeyboardStatus {
+    val view = LocalView.current
+    val context = LocalContext.current
+    var keyboardStatus by remember { mutableStateOf(getInitialKeyboardStatus(context)) }
+
+    DisposableEffect(view) {
+        val listener = ViewTreeObserver.OnGlobalLayoutListener {
```
malinajirka, line 33: 💬 Hmm, I hoped there would be an API which would make this easy. Using a globalLayout listener makes me a bit worried - it's easy to mess it up and cause significant performance issues down the road.

Have you tried checking if the keyboard is visible when the input fields gets focus instead of observing the keyboard status at all times? I realize, it might not be a bullet-proof approach, but perhaps checking with a 1s delay if the keyboard is visible after the input gets focus might be fine. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFloatingKeyboardHint.kt
```diff
@@ -0,0 +1,191 @@
...
+import com.woocommerce.android.R
+import com.woocommerce.android.ui.woopos.common.composeui.WooPosPreview
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosCornerRadius
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosElevation
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosSpacing
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTheme
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
+
+@Composable
+fun WooPosFloatingKeyboardHint(
```
malinajirka, line 41: 📓 I was originally thinking about displaying some kind of a floating button somewhere - small keyboard icon. On tap we could display a dialog explaining the situation. This hint feels quite strong considering we are not even sure something is wrong - they might have a hw keyboard connected. Having said that, perhaps I'm overthinking this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosKeyboardStatus.kt
kidinov: I added a 1-second delay to the calculations to avoid redundant work. But don’t worry, the calculations themselves are pretty basic, so they shouldn’t be too heavy
11cecfe3a258e55aad88c19c24c4db927bc0a43e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFloatingKeyboardHint.kt
kidinov: I thought about this too, but I’m not sure how to make it clear what’s going on if the space is limited. So, I went with this instead. If we decide to handle this case, one thing we could consider is showing this hint only once per app installation

malinajirka: Yeah, if this was shown once per app installation, I think it might be fine. Let's discuss on the call.

kidinov:
Closing as we are going with documentation for now 

pdfdoF-7GF-p2

PR #14368, opened 2025-07-23
Improve Plugins matching

samiuelson:
## Improve Plugins matching

WOOMOB-895

### Description

This PR changes the way version of WC plugin is determined – both cached (check in DB) and hot (REST API).

Previously, the matching logic for obtaining the WC plugin instance from DB and from server relied on the following implementation:
* The plugin was matched by the name, inclusive of its parent directory, e.g., `woocommerce/woocommerce`. 

As it turned out, in some cases, that directory may be different. For example, if you install woocommerce dev build, it looks like: `woocommerce-dev/woocommerce`.

Additionally, it turned out that it's possible to have multiple versions of the same plug-in installed, e.g.,  `woocommerce/woocommerce`, and `woocommerce-dev/woocommerce`. While it's possible to activate only one, the original logic fetching the WC plugin from DB and REST API, would not care about which one is activated, returning the first-found one, even if it was the one not activated.

This PR updates 2 use case classes: `FetchWooCorePluginVersion`, and `GetWooCorePluginCachedVersion` to address the edge-cases described above.

### Steps to reproduce

* To reproduce the scenario when WC plugin is installed in non-default dir you can install dev build of woocommerce. You can use a dev build I compiled locally: [woocommerce-dev.zip](https://github.com/user-attachments/files/21416796/woocommerce-dev.zip), and install on your site. 
* To reproduce multiple instances of WC plugin, you can install the release version of WC from your WP site's Plugins page (in addition to the dev one).

### Testing information

* Smoke test the app trying to spot any regression
* Verify that it's possible to access POS even if user switched to WC-dev plugin
* Verify it's impossible to access POS if the WC is not activated (mind the cache, though!)

The same as above.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/test/kotlin/com/woocommerce/android/util/FetchWooCorePluginVersionTest.kt
```diff
@@ -47,9 +50,11 @@ class FetchWooCorePluginVersionTest : BaseUnitTest() {
         val siteModel = mock<SiteModel>()
         whenever(selectedSite.getOrNull()).thenReturn(siteModel)
 
-        val fetchResult = mock<WooResult<List<SitePluginModel>>> {
-            on { isError }.thenReturn(true)
-        }
+        val error = WooError(
```
samiuelson, line 53: Just switched to fake instead of mock

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PluginSqlUtils.kt
```diff
@@ -71,9 +71,32 @@ object PluginSqlUtils {
...
+    @JvmStatic
+    fun getActiveSitePluginByName(site: SiteModel, pluginName: String): SitePluginModel? {
+        val allPlugins = WellSql.select(SitePluginModel::class.java)
+            .where()
+            .equals(SitePluginModelTable.LOCAL_SITE_ID, site.id)
+            .equals(SitePluginModelTable.IS_ACTIVE, true)
+            .endWhere()
+            .asModel
+
+        val extractedRequestedPluginName = pluginName.substringAfterLast('/')
```
samiuelson, line 93: 💡 Re `substringAfterLast`, if the name does not contain `/ `, the original value is returned.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FetchActiveWCPluginVersion.kt
```diff
@@ -24,7 +24,15 @@ class FetchWooCorePluginVersion @Inject constructor(
...
 
-    private fun List<SitePluginModel>?.getWooPlugin() = this?.firstOrNull {
-        it.name.endsWith(WooCommerceStore.WooPlugin.WOO_CORE.pluginName)
+    private fun List<SitePluginModel>?.getWooPlugin(): SitePluginModel? {
+        if (this.isNullOrEmpty()) return null
+
+        val pluginName = WooCommerceStore.WooPlugin.WOO_CORE.pluginName.substringAfterLast('/')
+
+        val activePlugin = this.firstOrNull { plugin ->
+            plugin.name.substringAfterLast('/').endsWith(pluginName) && plugin.isActive
```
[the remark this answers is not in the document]
kidinov: @samiuelson Isn’t this suggestion logical, considering that you only take the part after the last “/“?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -249,6 +249,10 @@ open class WooCommerceStore @Inject internal constructor(
         return pluginSqlUtils.getSitePluginByName(site, plugin.pluginName)
     }
 
+    fun getActiveSitePlugin(site: SiteModel, plugin: WooPlugin): SitePluginModel? {
```
kidinov, line 252: We have a new method that works correctly, but `getSitePlugin` doesn’t, along with other places where `WooCommerceStore.WooPlugin` may be matched by name. I believe we need to fix this issue once and for all, otherwise, it will only lead to more and more chaos and bugs

kidinov (approved):
The fix for the POS is working fine, but imo we should fix it in generic way (maybe in another PR). Maybe we need to store the actual name of the plugin (after last "/") in the DB so no string manipulation needed at all?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
samiuelson: I was investigating how we're querying for WooCommerceStore.WooPlugin in other places. In some cases, we want to get a plugin by name and check if it's active or not manually. Then, in case it's not active, we offer action button to activate it (e.g. in the collect payment -> card reader -> `woocommerce-payments` check). So we will need to keep `getSitePlugin()` for cases like this. I agree we should audit all usages of `getSitePlugin()` and analyze if they should be replaced by `getActiveSitePlugin`. I added task for that: WOOMOB-927

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FetchActiveWCPluginVersion.kt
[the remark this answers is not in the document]
samiuelson: Updated the string matching logic: 4595f7b271bfb941e3700c1a15a791bc9c84d43b

PR #14375, opened 2025-07-24
[POS as a tab i2][Bug] Analytics event not tracked in some cases

samiuelson:
## Analytics event not tracked in some cases

WOOPRD-693

### Description

This PR fixes the following bug:
> In the store ineligible for two reasons scenario -> after I fix the first issue and tap on retry, the pos_ineligible_ui_shown event is not tracked => the event is tracked only for the first reason.

### Steps to reproduce

1. Change currency to CZK and disable POS in WPAdmin
2. Open POS in the app -> notice pos_ineligible_ui_shown is tracked with reason `feature_switch_disabled`
3. Enable POS in WPAdmin
4. Tap on Retry in the app

Notice store_currency error is shown, but `woocommerceandroid_pos_ineligible_ui_shown` event is NOT tracked.

### Testing information

Verify that the event is tracked even in case the "retry" operation fails.

Above. 

### Images/gif

```
🔵 Tracked: woocommerceandroid_pos_ineligible_ui_shown, Properties: {"reason":"feature_switch_disabled", ...}
```


malinajirka (approved):
Thanks @samiuelson, works as expected! :shipit:

PR #14387, opened 2025-07-28
HACK: Removing product image background on-device with MLKit | Final PR

samiuelson:
## Product image "background remove with AI" feature in Product images viewer.

💡 This PR is the base for the "Product image background remove" feature and contains now of the following partial PRs:
* https://github.com/woocommerce/woocommerce-android/pull/14393
* https://github.com/woocommerce/woocommerce-android/pull/14397

### Description

This PR contains complete feature code - on-device image background removal, based on MLKit subject-segmentation.

##### 💡 ML-Kit subject segmentation dependency:
* .jar size is 110kB 
* ML model is dynamically downloaded first time the API is called (another 100kB)
* Dependency is released as beta; no stable release is available yet (https://developers.google.com/ml-kit/vision/subject-segmentation)

### Steps to reproduce

Try out the background remove feature in the Product details screen.

### Testing information

Verify it's possible to remove background in the specific Product image and save the processed image on the backend.

1. Go to Products and select a product
2. Pick a product and go to pictures
3. Tap on a picture, and scroll to select a desired picture
4. Tap on the "Remove background" button in the top toolbar
5. Save the processed picture

### The tests that have been performed

Above

### Images/gif

Demo:
<video src="https://github.com/user-attachments/assets/afdc05a2-7bd2-4081-964a-2dc773d4888d"/>

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14393, opened 2025-07-29
HACK: Removing product image background on-device | Part 2 

samiuelson:
### Description

This PR focuses on applying subject-segmentation APIs to remove the background from the product image (see video). 
The base use case responsible for removing the background is the `PerformImageBackgroundRemoval` class.

💡 `Do not merge` until the parent PR is approved: https://github.com/woocommerce/woocommerce-android/pull/14387/

💡 Out of scope: 
* `ProductImageRemoveBackgroundFragment` and its composable layouts. Please ignore this large class with messy code - it will be cleaned up in the next PR.
* The processed image is not saved or uploaded, just shown on the screen, for now.

### Steps to reproduce

1. Go to Products
2. Select Product and go to Product photos
3. Tap on a photo or add a new one
4. Tap "remove background" menu item
Observe background is removed from the image

### Testing information

Verify that background is being removed in different scenarios: From existing photos, images from devices gallery, or camera.

Above

### Images/gif

https://github.com/user-attachments/assets/a32fee1f-18c2-402b-a828-b1e8e6e1c9d3


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          5 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>CreateInputImageFromUrl</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>PerformImageBackgroundRemoval</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>RemoveBackground</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>SaveProcessedImageToTheProduct</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/ProductImageRemoveBackgroundFragment.kt
```diff
@@ -19,28 +22,34 @@ import androidx.compose.material3.Text
...
-import java.time.Instant
-import java.util.Date
+import kotlinx.coroutines.launch
 import javax.inject.Inject
+import kotlin.math.cos
+import kotlin.math.sin
+import kotlin.random.Random
 
 @AndroidEntryPoint
 class ProductImageRemoveBackgroundFragment : BaseFragment() {
```
samiuelson: 💡 Please ignore this Fragment in the review – it will be cleaned up in the next PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/CreateInputImageFromUrl.kt
```diff
@@ -0,0 +1,37 @@
...
+import coil.request.ImageRequest
+import coil.request.SuccessResult
+import com.google.mlkit.vision.common.InputImage
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+class CreateInputImageFromUrl @Inject constructor(
+    @ApplicationContext private val context: Context
+) {
+    suspend operator fun invoke(imageUrl: String): Result<InputImage> {
```
malinajirka: 💡 We might want to consider defining the coroutine dispatcher here. I believe this is a leaf node and as such it's usually a best practice to make sure the heavy operation is run on a safe-thread. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/CreateInputImageFromUrl.kt
```diff
@@ -0,0 +1,37 @@
...
+                    val bitmap = result.drawable.toBitmap()
+                    val inputImage = InputImage.fromBitmap(bitmap, 0)
+                    Result.success(inputImage)
+                }
+                else -> Result.failure(Exception("Image loading failed"))
+            }
+        } catch (e: IllegalArgumentException) {
+            Result.failure(e)
+        } catch (e: OutOfMemoryError) {
+            Result.failure(Exception("Out of memory loading image", e))
```
malinajirka: ❓ Is out-of-memory something the app can recover from? Feels a bit unconventional to return a Result instead of re-throwing the exception. I'd be even worried the JVM's memory might be in a corrupted state in this case, so it might be unsafe to recover.

We could potentially try to check whether the image will fit the memory before creating the bitmap. Wdyt?

malinajirka (approved):
Thanks @samiuelson!

LGTM - I've left a couple comments which would be great to consider before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/CreateInputImageFromUrl.kt
samiuelson: Good catch. I updated `CreateInputImageFromUrl` to run on `IO`: c9e88835a14f49ca8e81861d92e60e404dda5f55, also did the same for `RemoveBackground`: 3affb6601bbdd3af49cd38d352a1858b274bd058

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/CreateInputImageFromUrl.kt
samiuelson: Good point. I updated the coil `ImageRequest` to resize the image to `2048 × 2048`, which should result in max ~16.7 MB memory usage, which I think is acceptable. Also stopped swallowing the `OOM` exception: 34ba6ed8a9e74f458f785b8d9c091f67977662a9

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ai/CreateInputImageFromUrl.kt
malinajirka: Noice, thank you!

I haven't tried it but I remember testing Dispatchers.XYZ was a bit tricky, we usually inject `private val dispatchers: CoroutineDispatchers` so the test can provide any dispatcher it wants. I'm not 100% sure this hasn't changed, I have been using this approach out of habit but I haven't questioned it for a while.

samiuelson: This makes sense, thanks. Switched to the `CoroutineDispatchers` wrapper in both classes: fba145642942c84b8343f9ed242b67ea308e7f2f

PR #14394, opened 2025-07-29
Add MaterialTheme 3 definition to our WooTheme

hichamboushaba:
Part of WOOMOB-933

### Description
Currently, we are using both material 2 and material 3 components, but we didn't have an explicit definition of the material 3 theme, which meant that when using material 3 components, we'll be using their default look.
This PR adds a definition of the Material 3 theme, that aims to stay consistent as much as possible with our current theme: defining colors, typography and also shapes.

### Steps to reproduce
From what I saw, the only visible component from material 3 that we use is the `ModalBottomSheet`, so for testing:
- Open Blaze campaign creation screen.
- Edit budget or destination URL parameters.
- Check that the bottom sheet looks as before.

Feel free to explore other screens of the app for a quick sanity check.

### Testing information
Confirm there is no changes to the app UI.

The above.

### Images/gif
| Before | After |
| --- | --- |
| Screenshot_20250729_104815 | Screenshot_20250729_105049 |
| Screenshot_20250729_104806 | Screenshot_20250729_105101 |
| Screenshot_20250729_104729 | Screenshot_20250729_105135 |
| Screenshot_20250729_104754 | Screenshot_20250729_105739 |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
```diff
@@ -2,63 +2,118 @@ package com.woocommerce.android.ui.compose.theme
...
+        outline = colorResource(R.color.divider_color),
+        inverseOnSurface = colorResource(R.color.color_surface),
+        inverseSurface = colorResource(R.color.color_on_surface),
+        inversePrimary = colorResource(R.color.color_on_primary),
+        surfaceBright = colorResource(R.color.color_surface),
+        surfaceContainer = colorResource(R.color.color_surface_elevated),
+        surfaceContainerHigh = colorResource(R.color.color_surface_elevated),
+        surfaceContainerHighest = colorResource(R.color.color_surface_elevated),
+        surfaceContainerLow = colorResource(R.color.color_surface_elevated),
+        surfaceContainerLowest = colorResource(R.color.color_surface_elevated),
```
hichamboushaba, line 81: While Material2 used the elevation of components to apply an overlay to elevated components, Material 3 seems to have moved from this, and the elevation is not even used as often as before, so to keep a consistent look with before, I'm using the `elevated` surface color for all the `surfaceContainer` variations for now, IMO this is fine as a temporary solution, but it would be great if we can get Wagner to provide us with an updated color system for Material 3 like what he did for the POS project 1qcjzXitBHU7xPnpCOWnNM-fi-162_18857

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
```diff
@@ -2,63 +2,118 @@ package com.woocommerce.android.ui.compose.theme
...
+        primaryContainer = colorResource(R.color.color_primary),
+        onPrimaryContainer = colorResource(R.color.color_on_primary),
+        secondary = colorResource(R.color.color_secondary),
+        onSecondary = colorResource(R.color.color_on_secondary),
+        secondaryContainer = colorResource(R.color.color_secondary),
+        onSecondaryContainer = colorResource(R.color.color_on_secondary),
+        tertiary = colorResource(R.color.color_secondary),
+        onTertiary = colorResource(R.color.color_on_secondary),
+        tertiaryContainer = colorResource(R.color.color_secondary),
+        onTertiaryContainer = colorResource(R.color.color_on_secondary),
```
hichamboushaba, line 61: We don't have a tertiary color for now, so I'm applying the secondary one for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
```diff
@@ -2,63 +2,118 @@ package com.woocommerce.android.ui.compose.theme
...
+        secondaryVariant = colorResource(R.color.color_secondary_variant),
+        background = colorResource(R.color.default_window_background),
+        surface = colorResource(R.color.color_surface),
+        error = colorResource(R.color.color_error),
+        onPrimary = colorResource(R.color.color_on_primary),
+        onSecondary = colorResource(R.color.color_on_secondary),
+        onBackground = colorResource(R.color.color_on_background),
+        onSurface = colorResource(R.color.color_on_surface),
+        onError = colorResource(R.color.color_on_error),
+    )
```
hichamboushaba, line 27: This update enables us to maintain a single source of truth for colors, making it easier to update them as needed, and it ensures the consistency of the Compose and View themes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Shapes.kt
```diff
@@ -0,0 +1,19 @@
...
+    medium = RoundedCornerShape(8.dp),
+    large = RoundedCornerShape(8.dp),
+    extraLarge = RoundedCornerShape(8.dp)
+)
+
+val Material2Shapes = Material2Shapes(
+    small = RoundedCornerShape(4.dp),
+    medium = RoundedCornerShape(8.dp),
+    large = RoundedCornerShape(8.dp)
+)
```
hichamboushaba, line 19: I defined the shapes following what we use in our theme today, this will allow to need less overriding for the shapes in code.
This allowed removing the shape override in `WCModalBottomSheet` too.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
```diff
@@ -2,63 +2,118 @@ package com.woocommerce.android.ui.compose.theme
...
-    onSurface = WooColors.md_theme_light_onSurface,
-    onError = WooColors.md_theme_light_onError,
-)
+// Material 3 color schemes
+val Material3LightColorScheme
+    @Composable
+    get() = lightColorScheme(
+        primary = colorResource(R.color.color_primary),
+        onPrimary = colorResource(R.color.color_on_primary),
+        primaryContainer = colorResource(R.color.color_primary),
```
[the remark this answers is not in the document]
hichamboushaba: For now, our components that use the primaryContainer color should use the `primary` color, so this is intended.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Theme.kt
```diff
@@ -25,24 +27,39 @@ fun WooTheme(
...
 @Composable
 private fun SurfacedContent(
     content: @Composable () -> Unit
 ) {
-    Surface(color = MaterialTheme.colors.background) {
-        content()
+    Surface(color = MaterialTheme.colorScheme.background) {
+        Material2Surface(color = Material2Theme.colors.background) {
+            content()
+        }
```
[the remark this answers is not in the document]
hichamboushaba: The background colors are the same.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
```diff
@@ -2,63 +2,118 @@ package com.woocommerce.android.ui.compose.theme
...
+        surface = colorResource(R.color.color_surface),
+        onSurface = colorResource(R.color.color_on_surface),
+        surfaceVariant = colorResource(R.color.color_surface_variant),
+        onSurfaceVariant = colorResource(R.color.color_on_surface),
+        outline = colorResource(R.color.divider_color),
+        inverseOnSurface = colorResource(R.color.color_surface),
+        inverseSurface = colorResource(R.color.color_on_surface),
+        inversePrimary = colorResource(R.color.color_on_primary),
+        surfaceBright = colorResource(R.color.color_surface),
+        surfaceContainer = colorResource(R.color.color_surface_elevated),
```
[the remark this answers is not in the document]
hichamboushaba: We don't have any other colors for now, so we use the same color.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
irfano: I’m not asking for this PR but can’t we use the same colors he used for POS?  I remember Wagner wanted the design system to be consistent across POS and all other parts of the app.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Typography.kt
```diff
@@ -93,3 +92,98 @@ val WooTypography = Typography(
...
+        fontWeight = FontWeight.Normal,
+        fontSize = 34.sp,
+        lineHeight = 36.sp
+    ),
+    headlineMedium = TextStyle(
+        fontFamily = robotoFamily,
+        fontWeight = FontWeight.Normal,
+        fontSize = 24.sp,
+        lineHeight = 24.sp
+    ),
```
irfano: `headlineMedium` is equivalent to `h4`. So fontSize should be 34, `lineHeight` should be 36.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Typography.kt
```diff
@@ -93,3 +92,98 @@ val WooTypography = Typography(
...
+    headlineSmall = TextStyle(
+        fontFamily = robotoFamily,
+        fontWeight = FontWeight.Medium,
+        fontSize = 20.sp,
+        lineHeight = 24.sp,
+        letterSpacing = 0.5.sp
+    ),
```
irfano: `headlineMedium` is equivalent to `h5`. So `fontSize` should be 24 `fontWeight` should be Normal.

I used the table in this [page](https://developer.android.com/develop/ui/compose/designsystems/material2-material3#typography). I wonder how did you determine these sizes but I believe we should set these same as Material 2 by using the table.

Currently I found some font changes and there can be more.

|Before|After|
|-|-|
|before-refund|after-refund|
|before-light-notice|after-light-notice|
|before-notice|after-notice|
|before-customs|after-customs|

irfano (changes requested):
This is a great addition, thanks for handling it, Hicham.
I noticed some changes in fonts and colors, and I believe this PR shouldn’t introduce any visual changes. I’ve explained the font differences I observed. Also, please check the background color on the notices.

|Before|After|
|-|-|
|before-light-notice|after-light-notice|
|before-notice|after-notice||

hichamboushaba:
Thanks @irfano for the great review and for sharing the official typography mapping, even though I visited the above link, I didn't notice the differences because I used AI to generate the typography mapping, thinking it will be one of the tasks where it will do a good job, but it wasn't the case 🤦.
I'll fix the discrepancies, except for `h4`, let's discuss it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Typography.kt
hichamboushaba: Even though the official mapping [here](https://developer.android.com/static/develop/ui/compose/images/migration-typography-update.png) states that h4 maps to `headlineMedium`, I don't think it makes sense for us, `headlineMedium` by default has `28px`, while `h4` has `34px`, we use the default `h4` size in our app, so IMO we should map it to `headlineLarge` which by default has `32px`, and keep `headlineMedium` to its default `28px`, please let me know what you think about this.

I don't think we use `headlineMedium` anywhere in the app anyway, so this doesn't seem to have any impact for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Typography.kt
hichamboushaba: Can you please check the fonts now after the last commits? I updated the mappings, and for the Hazmat type, I set a specific text style, because the default style after applying the Material Theme [is](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/MaterialTheme.kt;l=108?q=MaterialThem) `bodyLarge`.

**Important**: Some small changes are still expected, when I said in the PR description that this PR shouldn't cause any visual changes, it was because I didn't know we were already using Material3 text styles in our screens. Given we were already using them, then updating the styles to match ours will in some cases cause some changes, but these are wanted changes, as they bring more consistency to the app. One noticeable example here is `titleLarge`, the default size for this is `22px`, but in our styles we use `20px`, so whenever it's used, the text will be slightly smaller.

-------

For the color changes, I'll discuss them in a separate comment.

hichamboushaba:
For the Snackbar color changes, we can divide the discussion two areas:

#### The Snackbar background
The new color is what we should be using, the old one was dark on dark, which didn't provide enough contrast. The old color was used because as we didn't provide a specific theme definition, the default theme was light by default.
The new color is also closer to what we use in other screens where Material 2 is used, for example here:
Screenshot_20250730_103212

#### The action button color
I updated it in addition to other updates to make the Snackbar more closer to the default material look in 1678e7a, with these updates, it looks now like this:

Screenshot_20250730_103352
Screenshot_20250730_110809

Please let me know what you think.

-------
These issues are what I wanted to avoid when I suggested p1740493990545489/1740491444.393149-slack-C6H8C3G23 to not use Material 3 components before theming the app.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/Typography.kt
irfano: Good point! It's better to set them to their default sizes, since they’re not being used in the app and won’t cause any changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/components/ShippingLabelsSnackbar.kt
```diff
@@ -22,13 +22,21 @@ import com.woocommerce.android.R
 
 @Composable
 fun ShippingLabelsSnackbar(visuals: ShippingLabelsSnackbarVisuals, action: () -> Unit) {
-    Surface(
-        color = SnackbarDefaults.color,
+    Snackbar(
```
irfano, line 25: As I see it, we used to use a `Surface` as a custom snackbar, but you replaced it with a real `Snackbar`. We’re using Snackbars in other parts of the shipping labels flow, so this change improves consistency. I especially appreciate that.

irfano (approved):
Thanks for addressing the feedback. I agree with all the points in your comments. I checked all the screens where we use Material 3, including the WooPOS screens, and couldn’t find any other differences.
Great job! We can now safely use Material 3 moving forward. LGTM! 👍🏻

PR #14395, opened 2025-07-29
[WOOMOB-937][Woo POS][Barcodes] Add software keyboard setup step for barcode scanner configuration

kidinov:
WOOMOB-937

### Description
Adds a new setup step in the barcode scanner configuration flow to address an issue where Android recognizes barcode scanners as external keyboards, causing on-screen keyboards to be hidden. The new step provides clear instructions on how to enable software keyboards when a physical barcode scanner is connected, improving the user experience during POS setup.

The step includes:
- Clear instructions for enabling on-screen keyboard in Android settings
- An input field for users to test keyboard visibility
- Proper integration into the existing scanner setup flow for all device types

### Steps to reproduce
1. Navigate to WooPOS settings
2. Start barcode scanner setup
3. Complete device selection and scanner pairing steps
4. After "Test Your Scanner" step, observe the new "Configure software keyboard" step
5. Follow the provided instructions to enable keyboard settings
6. Test keyboard visibility using the input field

### Testing information
- Tested on all supported scanner device types (TERA_1200, STAR_BSH_20B, NETUM_1228BC, OTHER)
- Verified proper navigation flow between steps
- Confirmed input field functionality with WooPosInputField component
- Validated string resources and UI layout

https://github.com/user-attachments/assets/2f28ff19-a594-4728-8daa-13990276fa9b


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigatorTest.kt
```diff
@@ -99,15 +99,15 @@ class WooPosScannerSetupNavigatorTest {
     }
 
     @Test
-    fun `given TestYourScanner step, when TERA_1200 device, then should navigate to ScannerSetupSuccess`() {
+    fun `given TestYourScanner step, when TERA_1200 device, then should navigate to SoftwareKeyboardSetup`() {
```
malinajirka, line 102: 🔍 We might consider adding a test to check the next step for the keyboard setup screen is the success screen, wdyt?

malinajirka (approved):
Thanks @kidinov! It works as expected. One potential issue I realized is that it might not be obvious to some users, how to hide the keyboard. The content is not scrollable. I think we should either add a 'hide keyboard' button next to the input or we should make the content scrollable, wdyt?
Screenshot 2025-07-31 at 8 29 49

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScannerSetupNavigatorTest.kt
kidinov: Added here 4e16768b6383676d0902eff2fd7b78adea488b90

kidinov:
> Thanks @kidinov! It works as expected. One potential issue I realized is that it might not be obvious to some users, how to hide the keyboard. The content is not scrollable. I think we should either add a 'hide keyboard' button next to the input or we should make the content scrollable, wdyt?


Thanks @malinajirka for the review

There is no clear path to make it scrollable when the keyboard becomes visible except modifying the whole activity behavior, or maybe if you know how to do this please share.

I added hiding the keyboard on any click to this state 2503dce272e901ae896fe9f6f300286e5df3d443

https://github.com/user-attachments/assets/deb66169-44cc-4bfc-9ecd-8afa8a76a717

Wdyt?

malinajirka:
> I added hiding the keyboard on any click to this state https://github.com/woocommerce/woocommerce-android/commit/2503dce272e901ae896fe9f6f300286e5df3d443

This is perfect! Even better than a button or scrolling behavior. Thanks!


@staskus Shared some suggestions for different phrasing - we might want to update the PR with those. Either case, I think this is good to be merged from the behavioral stand point of view.

kidinov:
@malinajirka I addressed the copy changes as well

image

PR #14396, opened 2025-07-29
[WOOMOB-934] Fix unintentional horizontal scroll on Payment Success screen when going through Barcode Setup flow

malinajirka:
#WOOMOB-934

### Description

This issue turned out to be more tricky than I anticipated. The problem is that when we display the Barcode Scanner Setup flow on the Payment Success screen, for some reason when you scan a barcode the search icon gets focus -> the `/n/r` at the end of the scan triggers submit action => the app automatically scrolls horizontally a bit (not fully) to make the button/search-input visible.

I wanted to fix the root cause and prevent the app from focusing the search field/search icon, however, compose works in mysterious ways to me so I wasn't able to implement such fix. Instead, I opted for handling the scanning logic for the scanner setup flow a bit differently -> the flow now consumes all scanning events and ignores them unless the app is on the Test Scan or Test Scan Timeout steps.

**Note: Considering this flow doesn't block the user and is quite improbable to happen - initial scanner setup on payment success screen, I'm thinking about merging this into trunk. Mostly because, the change has a risk of breaking something else. The current branch was created on top of the release branch so it contains your other fix, but I can update that after we make a final call. Wdyt?**

### Steps to reproduce

1. Open POS
2. Add an item to the order
3. Go to checkout
4. Tap on Cash payment
5. Enter amount and confirm
6. Tap on more menu
7. Tap on `Setup barcode scanner`
8. Scan something with your scanner and notice the view in the background scroll roughly 500px.

The same happens if you scan a barcode on steps that do not have the scanning modifier applied.

### Testing information

1. Repeat the above steps and double check the app works as expected.

A bunch of tests but it's very weird - sometimes I get a scanner into a state in which even when I repeat the above steps before the fix, the issue isn't reproducible. I haven't figured out how that happens - the only clue I have is that it's way more probable to happen when the scanner doesn't send the `end of line` character after scan.

### Images/gif

No user facing changes.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,7 +8,7 @@
 22.9.1
 -----
 - [*] POS: Barcode scanning on the order success screen now opens the cart and adds an item to the cart [https://github.com/woocommerce/woocommerce-android/pull/14382]
-
+- [*] POS: Fixes an issue with the Barcode Scanner Setup flow that could result in an inconsistent UI state on the Checkout and Payment Success Screens [https://github.com/woocommerce/woocommerce-android/pull/14396]
 22.9
 -----
 - [*] Fixed infinite loading animation in Order details pane when order list is empty [https://github.com/woocommerce/woocommerce-android/pull/14316]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/scanningsetup/WooPosScanningSetupDialog.kt
@@ -150,11 +150,7 @@ fun WooPosScanningSetupDialog(
                 .padding(WooPosSpacing.XLarge.value.toAdaptivePadding())
                 .listenForBarcodes(
                     { barcodeResult ->
-                        if (state.currentStep is ScanningSetupStep.TestYourScanner
-                            || state.currentStep is ScanningSetupStep.TestYourScannerTimeout
-                        ) {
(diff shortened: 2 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

RELEASE-NOTES.txt
```diff
@@ -5,6 +5,9 @@
 23.0
 -----
 
+22.9.1
```
kidinov, line 8: If it goes to the trunk then it should be 23.0

malinajirka: This is part of your PR I mentioned in the description - including the change log item. If we agree to target trunk, I'll rebase this PR.

kidinov (approved):
I tried a couple of other options and this looks the best to me.


> Note: Considering this flow doesn't block the user and is quite improbable to happen - initial scanner setup on payment success screen, I'm thinking about merging this into trunk. Mostly because, the change has a risk of breaking something else. The current branch was created on top of the release branch so it contains your other fix, but I can update that after we make a final call. Wdyt?

I believe it’s ok to merge it into the release branch. The risk of breaking anything is minimal, and we’ll release a new version in any case as there is another PR merged. And by the way, this PR also addresses the issue where we scan QR/barcodes from the checkout screen, which currently redirects to the cart screen, so the positive impact is slight higher

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14398, opened 2025-07-29
Exclude resources from PR size warnings

malinajirka:
This PR excludes `/src/main/res` files from the PR size calculation to avoid false warnings about large PRs when the majority of changes are resource files (vector drawables, layouts, etc.).


AliSoftware (approved):
👌

PR #14401, opened 2025-07-30
[Hack Week] Migrate some of our Compose components to Material 3

hichamboushaba:
Closes WOOMOB-933

### Description
This PR migrates some of the reusable Compose components to Material 3:
#### Buttons
The buttons defined in `Buttons.kt` were migrated to use Material 3.

#### DatePickerDialog
Currently, we have our custom DatePickerDialog that uses the Android `CalendarView` under the hood, in a custom Compose layout for the other parts. With this change, we will now use the Material 3 `DatePicker`.
For this part, I opted to keep using the same API for `DatePickerDialog` instead of using the provided one, as our version supports a `neutralButton` that can be shown in the dialog, while the default one doesn't, and also this makes the migration easier as the consumers are not impacted.

#### WCExposedDropDown and WCOverflowMenu
Both were migrated to Material 3

#### Buttons
Run the `ButtonsPreview` and check that the buttons look and behave the same.

#### Date Picker
1. Open Blaze campaign creation form.
2. In the Budget screen, tap on Edit, then on Start Date
3. Check the date picker.

#### Dropdown
1. Open products list.
2. Select multiple products.
3. Tap on "Update stock status" in the overflow menu.

#### Overflow menu
1. Open a product or order.
2. Open the custom fields screen.
3. Add a new custom field.
4. Check the overflow menu.

### Testing information
- Confirm the new components look and behave correctly. Small differences with previous Material 2 theme are to be expected, but should be minimal and barely noticeable, except for the date picker.

The above.

### Images/gif
| Before | After |
| --- | --- |
| Screenshot_20250730_134635 |Screenshot_20250730_144847 |
| Screenshot_20250730_134646 | Screenshot_20250730_144840 |
| Screenshot_20250730_134853 | Screenshot_20250730_135442 |
| Screenshot_20250730_134801 | Screenshot_20250730_135455 |
| Screenshot_20250730_134928 |  Screenshot_20250730_135514 |
| Screenshot_20250730_135122 | Screenshot_20250730_135523 |
| Screenshot_20250730_134949 | Screenshot_20250730_135539 |
| Screenshot_20250730_135106 | Screenshot_20250730_135548 |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

irfano (changes requested):
I noticed that some button colors have changed on the dashboard. I’m listing some screenshots below. Can you please check them?

| Before | After |
|--------|-------|
|||
|||
|||

hichamboushaba:
Thanks @irfano for the review, I'm not sure how I missed this in the dashboard 🤦.

This is an interesting issue that shows also how mixing M2 and M3 components can lead to weird issues, for this case, the issue is that the buttons impacted are using the M2 `Text` in the `content` argument, the M3 buttons we define will provide `androidx.compose.material3.LocalContentColor` and `androidx.compose.material3.LocalTextStyle`, but the M2 `Text` won't use them, because it expects `androidx.compose.material.LocalContentColor` and `androidx.compose.material.LocalTextStyle`.
I used a hack in 8d955cb to fix this, this is needed during the migration, as the button consumer can use M2 components, using the overload that accepts the String `text` argument won't have this issue, but I don't think we should enforce this.

TL;DR: the issue is fixed, and ready for another round.

irfano (approved):
Thanks for the update and additional explanations. It looks good now! LGTM! 👍🏻 Good to merge after fixing the Detekt error.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/dao/OrderStatusDaoTest.kt
```diff
@@ -16,6 +16,7 @@ import org.wordpress.android.fluxc.persistence.DatabaseTestRule
 
 @Config(manifest = Config.NONE)
 @RunWith(RobolectricTestRunner::class)
+@Suppress("UnitTestNamingRule")
```
hichamboushaba, line 19: @wzieba please note that I had to suppress this detekt issue in this newly added test, thanks to @kidinov, the app now has a detekt rule that validates the test names (#14402)

PR #14404, opened 2025-07-30
[WOOMOB-946] Remove deprecated testing approach

### LINKED ISSUE #6899: Resolve suppressed Coroutines related `DEPRECATION` warning for `ReviewModerationHandlerTests`
As part of #6902 and this [commit](https://github.com/woocommerce/woocommerce-android/pull/6902/commits/4a6304d4c8f3e6644bc3e42f0e88c99189a451bd), the `ReviewModerationHandlerTests` test and all its warnings got suppressed with `DEPRECATION`.

The warnings occur because this `ReviewModerationHandlerTests` test suite is the only such suite that still uses the old way of testing with Coroutines. As part of the [Coroutine 1.6 PR](https://github.com/woocommerce/woocommerce-android/pull/6292), migrating this test suite to the new way didn't happen as different approaches using both 'Standard' and 'Unconfined' dispatchers didn't work, 3 tests were always failing. One would need to remove this figure out how to use the new way of testing Coroutines on this `ReviewModerationHandlerTests`, maybe even rewriting it if need be.

malinajirka:
Closes https://github.com/woocommerce/woocommerce-android/issues/6899
Close WOOMOB-946

### Description

Goal of this PR is to remove the long deprecated approach to testing coroutines. I originally planned to update the tests one by one, but ultimately decided to completely rewrite them and add a few more tests with the help of AI.

### Testing information

Green CI is enough

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
@@ -1,25 +1,17 @@
-@file:Suppress("DEPRECATION_ERROR", "ForbiddenComment")
-// TODO: @malinajirka Issue: https://github.com/woocommerce/woocommerce-android/issues/6899
-
 package com.woocommerce.android.ui.reviews
 
 import com.woocommerce.android.model.ActionStatus
 import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.reviews.ProductReviewStatus.HOLD
-import kotlinx.coroutines.Dispatchers
+import com.woocommerce.android.viewmodel.BaseUnitTest
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.test.TestCoroutineDispatcher
-import kotlinx.coroutines.test.TestCoroutineScope
-import kotlinx.coroutines.test.resetMain
-import kotlinx.coroutines.test.runBlockingTest
-import kotlinx.coroutines.test.setMain
+import kotlinx.coroutines.test.runTest
 import org.assertj.core.api.Assertions.assertThat
-import org.junit.After
-import org.junit.Before
 import org.junit.Test
 import org.mockito.kotlin.any
 import org.mockito.kotlin.doReturn
@@ -35,7 +27,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
 import org.wordpress.android.fluxc.store.WCProductStore
 
 @ExperimentalCoroutinesApi
-class ReviewModerationHandlerTests {
+class ReviewModerationHandlerTests : BaseUnitTest()  {
     private val selectedSite: SelectedSite = mock {
         on { get() } doReturn SiteModel()
     }
@@ -45,31 +37,18 @@ class ReviewModerationHandlerTests {
 
     private lateinit var handler: ReviewModerationHandler
 
-    private val testDispatcher = TestCoroutineDispatcher()
-
-    @Before
-    fun starting() {
-        Dispatchers.setMain(testDispatcher)
-    }
-
-    @After
-    fun ending() {
-        Dispatchers.resetMain()
-        testDispatcher.cleanupTestCoroutines()
-    }
-
-    suspend fun setup(initMocks: suspend () -> Unit = {}) {
+    suspend fun setup(coroutineScope: CoroutineScope, initMocks: suspend () -> Unit = {}) {
         initMocks()
         handler = ReviewModerationHandler(
             selectedSite = selectedSite,
             productStore = productStore,
-            appCoroutineScope = TestCoroutineScope(testDispatcher)
+            appCoroutineScope = coroutineScope,
         )
     }
 
     @Test
-    fun `when moderating a review, then start with a pending state`() = testDispatcher.runBlockingTest {
-        setup()
+    fun `when moderating a review, then start with a pending state`() = runTest {
+        setup(this)
 
         val status = runTestAndCollectLastStatus {
             handler.postModerationRequest(review, HOLD)
@@ -81,8 +60,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `given moderating a review, when the undo delay is passed, then change status to submitted`() =
-        testDispatcher.runBlockingTest {
-            setup()
+        runTest {
+            setup(this)
 
             val latestStatus = runTestAndCollectLastStatus {
                 handler.postModerationRequest(review, HOLD)
@@ -97,8 +76,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `given moderating a review, when the undo delay is passed, then submit status to the API`() =
-        testDispatcher.runBlockingTest {
-            setup {
+        runTest {
+            setup(this) {
                 whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                     WooResult(
                         WCProductReviewModel(
@@ -122,8 +101,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `when review status update succeeds, then status removed after a given time`() =
-        testDispatcher.runBlockingTest {
-            setup {
+        runTest {
+            setup(this) {
                 whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                     WooResult(
                         WCProductReviewModel(
@@ -150,8 +129,8 @@ class ReviewModerationHandlerTests {
         }
 
     @Test
-    fun `when review status update fails, then change status to error`() = testDispatcher.runBlockingTest {
-        setup {
+    fun `when review status update fails, then change status to error`() = runTest {
+        setup(this) {
             whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                 WooResult(
                     WooError(GENERIC_ERROR, UNKNOWN, "")
@@ -171,8 +150,8 @@ class ReviewModerationHandlerTests {
     }
 
     @Test
-    fun `when review status update fails, then status removed after a given time`() = testDispatcher.runBlockingTest {
-        setup {
+    fun `when review status update fails, then status removed after a given time`() = runTest {
+        setup(this) {
             whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                 WooResult(
                     WooError(GENERIC_ERROR, UNKNOWN, "")
@@ -196,8 +175,8 @@ class ReviewModerationHandlerTests {
     }
 
     @Test
-    fun `when moderation is canceled, then submit success with original status`() = testDispatcher.runBlockingTest {
-        setup()
+    fun `when moderation is canceled, then submit success with original status`() = runTest {
+        setup(this)
 
         val latestStatus = runTestAndCollectLastStatus {
             handler.postModerationRequest(review, HOLD)
@@ -214,8 +193,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `when queuing a second moderation request, then skip delay for the previous one`() =
-        testDispatcher.runBlockingTest {
-            setup {
+        runTest {
+            setup(this) {
                 whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                     WooResult(
                         WCProductReviewModel(
@@ -239,8 +218,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `when adding a second moderation request, then queue it after the previous one`() =
-        testDispatcher.runBlockingTest {
-            setup {
+        runTest {
+            setup(this) {
                 whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                     WooResult(
                         WCProductReviewModel(
@@ -264,8 +243,8 @@ class ReviewModerationHandlerTests {
 
     @Test
     fun `when queuing moderations, then make sure the pending status list is ordered`() =
-        testDispatcher.runBlockingTest {
-            setup {
+        runTest {
+            setup(this) {
                 whenever(productStore.updateProductReviewStatus(any(), any(), any())).thenReturn(
                     WooResult(
                         WCProductReviewModel(
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
@@ -2,14 +2,11 @@ package com.woocommerce.android.ui.reviews
 
 import com.woocommerce.android.model.ActionStatus
 import com.woocommerce.android.tools.SelectedSite
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.APPROVED
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.HOLD
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.SPAM
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.TRASH
 import com.woocommerce.android.viewmodel.BaseUnitTest
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.advanceTimeBy
 import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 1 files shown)


ParaskP7:
Nice @malinajirka and thanks ❤️ this actually then closes #6899 and we can add that to the description to have that auto-closed when this gets merged, I am on-it with the review... 👀

ParaskP7 (approved):
👋 @malinajirka , the tests change LGTM, and thus I am unblocking this with a ✅ no matter, many thanks! 🥇

Also, I tried making the test suite a bit more robust, by initializing `handler` once and using `TestScope(coroutinesTestRule.testDispatcher)` instead of a `backgroundScope` scope on it. See patch below, let me know what you think! 🙏 

<details>
    <summary>patch</summary>

```diff
Index: WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt	(revision 91fc5d42ed6d1f357b98a0ae358ef7b3170a9941)
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewModerationHandlerTests.kt	(date 1753964076107)
@@ -2,14 +2,11 @@
 
 import com.woocommerce.android.model.ActionStatus
 import com.woocommerce.android.tools.SelectedSite
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.APPROVED
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.HOLD
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.SPAM
-import com.woocommerce.android.ui.reviews.ProductReviewStatus.TRASH
 import com.woocommerce.android.viewmodel.BaseUnitTest
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.advanceTimeBy
 import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
@@ -38,7 +35,11 @@
 
     private val review = ProductReviewTestUtils.generateProductReview(0L, 0L)
 
-    private lateinit var handler: ReviewModerationHandler
+    private val handler = ReviewModerationHandler(
+        selectedSite = selectedSite,
+        productStore = productStore,
+        appCoroutineScope = TestScope(coroutinesTestRule.testDispatcher),
+    )
 
     @Before
     fun setup() = runTest {
@@ -46,7 +47,7 @@
             WooResult(
                 WCProductReviewModel(
                     remoteProductReviewId = RemoteId(review.remoteId),
-                    status = HOLD.toString(),
+                    status = ProductReviewStatus.HOLD.toString(),
                 )
             )
         )
@@ -56,11 +57,6 @@
     fun `when moderating a review, then start with a pending state`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val collectJob = launch {
             handler.pendingModerationStatus.collect { statuses ->
@@ -70,7 +66,7 @@
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         // THEN
@@ -86,11 +82,6 @@
     fun `given moderating a review, when the undo delay is passed, then change status to submitted`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val collectJob = launch {
             handler.pendingModerationStatus.collect { statuses ->
@@ -99,7 +90,7 @@
         }
 
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         val pendingStatuses = statusUpdates.flatten().filter { it.actionStatus == ActionStatus.PENDING }
@@ -124,11 +115,6 @@
     fun `given moderating a review, when the undo delay is passed, then submit status to the API`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val collectJob = launch {
             handler.pendingModerationStatus.collect { statuses ->
@@ -137,7 +123,7 @@
         }
 
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         // WHEN
@@ -154,14 +140,8 @@
     @Test
     fun `when review status update succeeds, then status removed after a given time`() = runTest {
         // GIVEN
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
-
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         advanceTimeBy(ReviewModerationHandler.UNDO_DELAY + 100)
@@ -184,11 +164,6 @@
         )
 
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val collectJob = launch {
             handler.pendingModerationStatus.collect { statuses ->
@@ -197,7 +172,7 @@
         }
 
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         // WHEN
@@ -218,14 +193,8 @@
             WooResult(WooError(GENERIC_ERROR, UNKNOWN, ""))
         )
 
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
-
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         advanceTimeBy(ReviewModerationHandler.UNDO_DELAY + 100)
@@ -243,14 +212,8 @@
     @Test
     fun `when moderation is canceled, then submit success with original status`() = runTest {
         // GIVEN
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
-
         runCurrent()
-        handler.postModerationRequest(review, HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
         runCurrent()
 
         val pendingStatus = handler.pendingModerationStatus.first()
@@ -271,11 +234,6 @@
     fun `when queuing a second moderation request, then skip delay for the previous one`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val review1 = review.copy(remoteId = 1L)
         val review2 = review.copy(remoteId = 2L)
@@ -287,12 +245,12 @@
         }
 
         runCurrent()
-        handler.postModerationRequest(review1, HOLD)
+        handler.postModerationRequest(review1, ProductReviewStatus.HOLD)
         advanceTimeBy(100)
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review2, HOLD)
+        handler.postModerationRequest(review2, ProductReviewStatus.HOLD)
         advanceTimeBy(100)
         runCurrent()
 
@@ -309,11 +267,6 @@
     fun `when adding a second moderation request, then queue it after the previous one`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val review1 = review.copy(remoteId = 1L)
         val review2 = review.copy(remoteId = 2L)
@@ -327,8 +280,8 @@
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review1, HOLD)
-        handler.postModerationRequest(review2, HOLD)
+        handler.postModerationRequest(review1, ProductReviewStatus.HOLD)
+        handler.postModerationRequest(review2, ProductReviewStatus.HOLD)
         advanceTimeBy(100)
         runCurrent()
 
@@ -342,20 +295,14 @@
     @Test
     fun `when queuing moderations, then make sure the pending status list is ordered`() = runTest {
         // GIVEN
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
-
         val review1 = review.copy(remoteId = 1L)
         val review2 = review.copy(remoteId = 2L)
 
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review1, HOLD)
-        handler.postModerationRequest(review2, HOLD)
+        handler.postModerationRequest(review1, ProductReviewStatus.HOLD)
+        handler.postModerationRequest(review2, ProductReviewStatus.HOLD)
         advanceTimeBy(100)
         runCurrent()
 
@@ -369,7 +316,11 @@
     @Test
     fun `when moderating to different review statuses, then handle each correctly`() = runTest {
         // GIVEN
-        val statuses = listOf(APPROVED, SPAM, TRASH)
+        val statuses = listOf(
+            ProductReviewStatus.APPROVED,
+            ProductReviewStatus.SPAM,
+            ProductReviewStatus.TRASH
+        )
 
         for (status in statuses) {
             // GIVEN
@@ -382,12 +333,6 @@
                 )
             )
 
-            handler = ReviewModerationHandler(
-                selectedSite = selectedSite,
-                productStore = productStore,
-                appCoroutineScope = backgroundScope,
-            )
-
             runCurrent()
 
             // WHEN
@@ -408,11 +353,6 @@
     fun `when posting multiple requests for same review rapidly, then only process the latest`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val collectJob = launch {
             handler.pendingModerationStatus.collect { statuses ->
@@ -423,9 +363,9 @@
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review, HOLD)
-        handler.postModerationRequest(review, APPROVED)
-        handler.postModerationRequest(review, SPAM)
+        handler.postModerationRequest(review, ProductReviewStatus.HOLD)
+        handler.postModerationRequest(review, ProductReviewStatus.APPROVED)
+        handler.postModerationRequest(review, ProductReviewStatus.SPAM)
         runCurrent()
         advanceTimeBy(100)
         runCurrent()
@@ -438,18 +378,12 @@
         }
 
         val latestPending = pendingForSameReview.lastOrNull()
-        assertThat(latestPending?.newStatus).isEqualTo(SPAM)
+        assertThat(latestPending?.newStatus).isEqualTo(ProductReviewStatus.SPAM)
     }
 
     @Test
     fun `when undoing operation that doesn't exist, then handle gracefully`() = runTest {
         // GIVEN
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
-
         runCurrent()
         val nonExistentReview = review.copy(remoteId = 999L)
 
@@ -466,11 +400,6 @@
     fun `when handling concurrent operations on different reviews, then process all correctly`() = runTest {
         // GIVEN
         val statusUpdates = mutableListOf<List<ReviewModerationStatus>>()
-        handler = ReviewModerationHandler(
-            selectedSite = selectedSite,
-            productStore = productStore,
-            appCoroutineScope = backgroundScope,
-        )
 
         val review1 = review.copy(remoteId = 1L)
         val review2 = review.copy(remoteId = 2L)
@@ -485,9 +414,9 @@
         runCurrent()
 
         // WHEN
-        handler.postModerationRequest(review1, APPROVED)
-        handler.postModerationRequest(review2, SPAM)
-        handler.postModerationRequest(review3, TRASH)
+        handler.postModerationRequest(review1, ProductReviewStatus.APPROVED)
+        handler.postModerationRequest(review2, ProductReviewStatus.SPAM)
+        handler.postModerationRequest(review3, ProductReviewStatus.TRASH)
         runCurrent()
         advanceTimeBy(100)
         runCurrent()
```

</details>

FYI: There's only this last test that fails when you'll apply this patch, and for the life of me, I just couldn't fix it without altering the assert logic at the very end, maybe you can fix it... 🤞 

PS: Again, many thanks for this! 🙇 ❤️ 🚀

malinajirka:
Noice, thank you for the review as well as the patch! That's definitely much nicer ❤️.

malinajirka:
> FYI: There's only this last test that fails when you'll apply this patch, and for the life of me, I just couldn't fix it without altering the assert logic at the very end, maybe you can fix it... 🤞

Ahh, I missed this part of your comment and I applied the patch without thinking about it more deeply.

1. Initializing the handler on the class level makes it shared across all the tests which is a bit risky. At first I thought moving the initializing to @Before method would work, but it won't because of `2`.
2. I don't have 100% clarity on this, so take this with a grain of salt. But I think the issue with the patch is that we create a TestScope which is unrelated to the test scope created by `runTest`. Moreover, I think the usage of `backgroundScope` is correct here since the production code uses Application Level scope and AFAIU that's exactly what `backgroundScope` is for.

I extracted creation of the SUT to a method, but I reverted back to calling it in each method to make sure we use scope of the `runTest`. 

Let me know what you think @ParaskP7

ParaskP7:
👋 @malinajirka !

> Ahh, I missed this part of your comment and I applied the patch without thinking about it more deeply.

😊 yea, I should have made that more clear, apologies... 🫣

> 1. Initializing the handler on the class level makes it shared across all the tests which is a bit risky.

Actually, the interesting thing about this and unit tests, is that using this `val` like that, actually any test class field, would not be shared across all the tests, this is a common misconception which I too forget from time-to-time.

FYI: At the property level, this handler will be initialized when the test class instance is created.
- JUnit (and typical JVM test frameworks) create a new instance of a test class for each @Test method.
- Therefore, for each test, a fresh instance of a test class is created, and a property initializer runs for that instance.
- ...and thus:
    - `handler` is not shared between tests.
    - It’s initialized per test method, just like it would be if you set it up in a `@Before`-annotated function.
    - Each test gets a new handler.

PS: If one however declares a `handler` as `companion object`, `object`, or `@JvmStatic` then this would be shared across all test methods. But a regular property (even with private val) will be re-initialized per test method.

> At first I thought moving the initializing to https://github.com/before method would work, but it won't because of 2.

Yea, I tried that too, using `backgroundScope` on that `@Before` `runTest { ... }` level is not the same as using `backgroundScope` on a `@Test` `runTest { ... }`, I struggled with that myself... 😊

> 2. I don't have 100% clarity on this, so take this with a grain of salt. But I think the issue with the patch is that we create a TestScope which is unrelated to the test scope created by runTest.

Yea, I think you're right here and we should indeed use the test scope created by `runTest` to do what we want with these tests. 🤔

> Moreover, I think the usage of backgroundScope is correct here since the production code uses Application Level scope and AFAIU that's exactly what backgroundScope is for.

Yea, I haven't see usage of this `backgroundScope` much, so this through me off. I also didn't see it used throughout the codebase so it made me think if that's the correct approach. I really have limited knowledge about `backgroundScope` and where you really need to use that.

Thus, I proceeded with our usual `TestScope(coroutinesTestRule.testDispatcher)` approach and see how far I'll get with it, thinking that I am missing something on the last test that you could simply fix (or update accordingly). 😊

> I extracted creation of the SUT to a method, but I reverted back to calling it in each method to make sure we use scope of the runTest.

I do think we should indeed accept this solution and forward, apologies for all the back-and-forward! 👍  💯 😊

malinajirka:
> JUnit (and typical JVM test frameworks) create a new instance of a test class for each @test method.

TIL very interesting - I thought it calls Before on each test but didn't realize it creates a new instance of the test class 🤔.

> I also didn't see it used throughout the codebase so it made me think if that's the correct approach.

Might not be the correct approach, however, I think this case is a bit specific since we are injecting app level scope into the SUT instead of just dispatchers as we do in vast majority of the codebase.

> I do think we should indeed accept this solution and forward, apologies for all the back-and-forward! 👍 💯 😊

No reason to apologize, I learned something new so definitely worth it! 💟

ParaskP7:
> TIL very interesting - I thought it calls Before on each test but didn't realize it creates a new instance of the test class 🤔.

Yea, I know, this is one of the TILs I get myself multiple times a couple of years, as I forget that and re-learn that again, not intuitive at all... 🤷 

> Might not be the correct approach, however, I think this case is a bit specific since we are injecting app level scope into the SUT instead of just dispatchers as we do in vast majority of the codebase.

💯 it's indeed a specific case! 👍

> No reason to apologize, I learned something new so definitely worth it! 💟

🙇 ❤️ 🚀

PR #14406, opened 2025-07-30
[HACK] Migrate WooLogViewerScreen to Material 3

kidinov:
### Description
Migrated the WooLogViewerScreen from Material 2 to Material 3 components to align with the app's design system modernization. This change updates the log viewer screen to use Material 3 Scaffold, TopAppBar, and other components while maintaining the same functionality and visual appearance.

Key changes:
- Updated WooLogViewerActivity to use ComponentActivity with edge-to-edge support
- Converted all Material 2 components to Material 3 equivalents
- Toolbar uses material 3 now

### Steps to reproduce
1. Open the WooCommerce app
2. Navigate to Settings > Help & Support > Application logs
3. Verify the screen displays correctly in both light and dark modes
4. Test toolbar interactions (back button, copy, share)
5. Test search functionality with navigation arrows
6. Verify edge-to-edge display behavior

### Testing information
- Test on devices with different screen sizes and densities
- Verify proper window insets handling with system bars
- Test orientation changes
- Verify accessibility features still work correctly
- Test with different system themes (light/dark)

- Visual verification of Material 3 styling
- Functional testing of all toolbar actions
- Search functionality testing
- Dark mode compatibility testing
- Edge-to-edge display verification

### Images/gif

https://github.com/user-attachments/assets/437e14c1-e186-4bf7-af43-938a51308b53


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/ThemeMaterial3.kt
```diff
@@ -0,0 +1,120 @@
...
+    inversePrimary = WooColors.md_theme_light_primary
+)
+
+/**
+ * This theme should be used to support light/dark colors if the composable root of the view tree
+ * does not support the use of contentColor.
+ * @see <a href="https://developer.android.com/jetpack/compose/themes/material#content-color</a> for more details
+ */
+@Composable
+fun WooThemeM3WithBackground(
```
hichamboushaba: @kidinov there is no need for defining a new theme, we just added a Material 3 theme definition to our `WooTheme` declaration, it was just merged today (https://github.com/woocommerce/woocommerce-android/pull/14394), so you can just rebase your branch onto trunk, and remove these new declarations, and then continue to use `WooThemeWithBackground` as before, it will supply the correct colors.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/ToolbarMaterial3.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import com.woocommerce.android.R
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun ToolbarM3(
```
hichamboushaba: Just asking a question here, WDYT about updating the current `Toolbar` to use the Material 3 components under the hood instead of adding a new component?

This is the approach I followed in this PR https://github.com/woocommerce/woocommerce-android/pull/14401, and to me, this will make the migration easier, as the screens that use the components will be migrated correctly, and with the correct theming, the components will look the same, so no UI impact.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/ThemeMaterial3.kt
kidinov: Didn't know that. Using that theme now.

Btw, are you sure that it's a good idea to wrap one theme into another?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/ToolbarMaterial3.kt
kidinov: I tried that and seems like it works fine

0c8e97f2f73ee1c8626a3bb2fee85a4cdf2a3579

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/ThemeMaterial3.kt
hichamboushaba: > Btw, are you sure that it's a good idea to wrap one theme into another?

Yes, it should be fine, the themes provide composition locals for different classes (for example: `androidx.compose.material.LocalShapes` vs `androidx.compose.material3.LocalShapes`, and `androidx.compose.material.LocalColors` vs `androidx.compose.material3.LocalColorScheme`), so they don't conflict with each other. The themes will override two common composition locals: `LocalIndication` and `LocalTextSelectionColors`, for those, the last one will override the other one, but from my testing, both themes define almost the same colors for those.

Also, from my testing and Irfan's testing when reviewing the PR, this approach didn't cause any issues. But if you have any concerns, please share them 🙏.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
```diff
@@ -44,40 +47,42 @@ fun ToolbarWithHelpButton(
...
     onNavigationButtonClick: (() -> Unit),
     navigationIcon: ImageVector = Icons.AutoMirrored.Filled.ArrowBack,
     navigationIconContentDescription: String = stringResource(id = string.back),
-    windowInsets: WindowInsets = WindowInsets(0),
+    windowInsets: WindowInsets = TopAppBarDefaults.windowInsets,
 ) {
     Toolbar(
         modifier = modifier,
-        title = { Text(title) },
+        title = { Text(title, style = MaterialTheme.typography.headlineSmall) },
```
hichamboushaba:
```suggestion
        title = { Text(title) },
```

The `headlineSmall` style is causing a significant change from the old implementation, and thus causing inconsistencies with other screens:

| Before | After |
| --- | --- |
| Screenshot_20250731_145419 | Screenshot_20250731_145317 |

I tested removing the explicit style to use the default style provided by the `TopAppBar`, and it seems to be closer to what had before:

Screenshot_20250731_145728

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
```diff
@@ -44,40 +47,42 @@ fun ToolbarWithHelpButton(
...
-    windowInsets: WindowInsets = WindowInsets(0),
+    windowInsets: WindowInsets = TopAppBarDefaults.windowInsets,
     actionButtonIcon: ImageVector,
     onActionButtonClick: (() -> Unit),
     actionIconContentDescription: String
 ) {
     Toolbar(
         modifier = modifier,
-        title = { Text(title) },
+        title = { Text(title, style = MaterialTheme.typography.headlineSmall) },
```
hichamboushaba: Same as above

```suggestion
        title = { Text(title) },
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
```diff
@@ -86,26 +91,28 @@ fun Toolbar(
             IconButton(onClick = onActionButtonClick) {
                 Icon(
                     imageVector = actionButtonIcon,
-                    contentDescription = actionIconContentDescription
+                    contentDescription = actionIconContentDescription,
+                    tint = MaterialTheme.colorScheme.onSurface
```
hichamboushaba: I don't think we need this `tint` override here, `TopAppBar` will tint the icons using the `actionIconContentColor` we are providing below.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
```diff
@@ -140,19 +148,25 @@ fun Toolbar(
...
     actions: @Composable RowScope.() -> Unit = {}
 ) {
     TopAppBar(
         windowInsets = windowInsets,
-        backgroundColor = colorResource(id = R.color.color_toolbar),
+        colors = TopAppBarDefaults.topAppBarColors(
+            containerColor = colorResource(id = R.color.color_toolbar),
+            titleContentColor = MaterialTheme.colorScheme.onSurface,
+            navigationIconContentColor = MaterialTheme.colorScheme.onSurface,
+            actionIconContentColor = MaterialTheme.colorScheme.onSurface
```
hichamboushaba: We use primary color for Toolbar actions (you can confirm it in other app screens)
```suggestion
            actionIconContentColor = MaterialTheme.colorScheme.primary
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
```diff
@@ -2,33 +2,36 @@ package com.woocommerce.android.ui.compose.component
...
+@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun ToolbarWithHelpButton(
     modifier: Modifier = Modifier,
     title: String = "",
     onNavigationButtonClick: (() -> Unit)? = null,
     navigationIcon: ImageVector? = Icons.AutoMirrored.Filled.ArrowBack,
     navigationIconContentDescription: String = stringResource(id = string.back),
-    windowInsets: WindowInsets = WindowInsets(0),
+    windowInsets: WindowInsets = TopAppBarDefaults.windowInsets,
```
hichamboushaba: The default insets we get from this change is `WindowInsets.systemBarsForVisualComponents.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Top)`,This change have some side effects depending on where the Toolbar is used and whether the insets were consumed or not, an example where this breaks the app is screens that use the Toolbar and the keyboard is shown:

Screenshot_20250731_151923
Screenshot_20250731_152109

I think we should revert to `WindowInsets(0)` for now, and then update the `WooLogViewerScreen` to either revert to applying the `WindowInsetsSides.Top` direction of the padding, or pass a specific inset to the Toolbar.

RELEASE-NOTES.txt
```diff
@@ -4,13 +4,14 @@
 
 23.0
 -----
+- [*] Add search functionality to application log viewer [https://github.com/woocommerce/woocommerce-android/pull/14386]
+- [*] Edge to edge support to application log viewer [https://github.com/woocommerce/woocommerce-android/pull/14406]
```
hichamboushaba: Just wondering, what part was broken here? 
From comparing the before and after, all I can notice is that the background doesn't expand to edge-to-edge anymore, IMO both approaches are valid.

| Before | After |
| --- | --- |
| Screenshot_20250731_151252 | Screenshot_20250731_151139 |

hichamboushaba (changes requested):
Thanks @kidinov, great work here, but there are some issues that need fixing before merging it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: Good catch!

51a9bb178df2bf9870f7273d7b1e2e25f46744cb

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: 51a9bb178df2bf9870f7273d7b1e2e25f46744cb

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: Good catch!

1f1aaa753919f0363e6bce084ff977e79db26fbc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: Interestingly, I cannot reproduce this on my emulator. I tried both SDK 35 and 33. What device/emulator you use? 

image

image

image


What I really don't like about `windowInsets: WindowInsets = WindowInsets(0),` is that we override the default behavior of the Material 3 TopAppBar, which is supposed to handle edge-to edge out of the box

RELEASE-NOTES.txt
kidinov: I am sure that it had bottom padding taht was covering the navigation bar, but now I tested on trunk and even on 22.8 and it's not there. So weird =). I removed that from the release notes then

0c9edf7ddb9300a49f2c64ce8537e002c3aec78c

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: Removed redundant redeclaration

fdbeaca6010ec99fdb0baf33c359053b9320e9fd

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
hichamboushaba: Thanks @kidinov for testing, and sorry for not making it clearer (I mentioned it above, but it was easy to miss). To reproduce the issue, you need to trigger displaying the keyboard, then hide it in one of the mentioned screens. Did you do this?

After looking more into this, I see it is related to this [issue](https://issuetracker.google.com/issues/411868840) that I reported to Google, which seems to have been fixed recently, and will be [released](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.9.0-alpha02) with the upcoming Compose 1.9, but for now, we'll have to avoid using consumed insets in Compose views.

> What I really don't like about windowInsets: WindowInsets = WindowInsets(0), is that we override the default behavior of the Material 3 TopAppBar, which is supposed to handle edge-to edge out of the box

I understand this, but I think the default that the team uses is more geared towards full Compose apps, for apps that still use the View system as the root layout (like what we do), we have some insets handling in our layouts which collides with the Compose system, and this is what happens here, that's why I think using `WindowInsets(0)` like the default that was in Material 2 still works best for our app, and given it's applied on a Component, we can easily update it in the future when the bug is fixed.

RELEASE-NOTES.txt
hichamboushaba: Thank you 🙏

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Toolbar.kt
kidinov: @hichamboushaba 

Make sense!

Implemented here f9dda37b00926578d7ba497054da538d31a964f4

hichamboushaba (approved):
Nice work @kidinov, thanks for taking the time to fix the issues.

I took the liberty to push one last commit to fix the compilation issue, I hope that's fine.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14407, opened 2025-07-30
[Hack Week] App logs improvements: persist logs on disk

hichamboushaba:
Part of WOOMOB-948

### Description
This PR is the first part of the improvements for app logs, it adds logic to persist app logs of the last seven days on disk to follow what iOS does. The list of changes of the PR:

1. Remove `WooLogWrapper`, I don't think the Wrapper had any added value, we can inject `WooLog` directly, and when needed, Mockito can mock it. The removal makes maintenance work easier. **It's better to review this change independently, as it's not related to the other changes of the PR, the change is in the commit 3d43168**
2. Introduce `WooFileLogger`, this class is the main component of the logs storage, it acts as a facade for writing logs, and handles reading logs from storage. For log writing, it uses an in-memory buffer to minimize disk access.
3. Update `WooLog` to use `WooFileLogger` as the source of truth for logs, and remove the in-memory logs.
4. Update the Encrypted Logs to use `WooFileLogger` directly, and get rid of the `WooLogFileProvider` that copied logs to a temporary file.

> [!NOTE]
> - Sorry for the size of the changes, half of them are for tests, and I didn't find a way to split this further. The main changes of the PR are in two files: `WooFileLogger` and `LogFileWriter`
> - For now, only the logs of the current day are shown in the logs screen, I'll work on allowing viewing the other logs in a separate PR.
> - **Important**: Copy and share buttons are broken, please ignore this in this PR, I will fix them in the next PR.

##### Logs viewing
1. Open the app.
2. Perform some actions to fill the logs.
3. Open the log viewer (from the support screen)
4. Confirm logs are shown.
5. Restart the app.
6. Re-open the log viewer.
7. Confirm logs were persisted (you can look for the first tracked line `NotificationStore onRegister` and confirm it was logged more than once)

##### Encrypted logs non-regression
1. Apply the following [patch](https://github.com/user-attachments/files/21548295/patch.patch)
2. Run the app
3. Open any order.
4. Open Sentry, and look for an event `Test encrypted logs`
5. Grab the Encrypted Logs UUID
6. Open the Encrypted Logs viewer, and confirm the file is correctly uploaded.
7. Delete the Sentry issue.

### Testing information
- Check the logging works as expected, and that logs persist different app launches.
- Confirm that encrypted logs work as expected.

The above.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
```diff
@@ -0,0 +1,19 @@
+package com.woocommerce.android.util.crashlogging
+
+import com.woocommerce.android.util.WooLog
+import kotlinx.coroutines.runBlocking
+import java.io.File
+import javax.inject.Inject
+
+class EncryptedLogsFileProvider @Inject constructor(private val wooLog: WooLog) {
```
hichamboushaba: This class was just renamed from the `WooLogFileProvider`, Github is marking it as a new file wrongly.

You can exclude the commit eea06d7188a51bbd4f5eb3972a9c880dbc916d85 during the review to avoid the noise of this.

WooCommerce/src/test/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProviderTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.junit.MockitoJUnitRunner
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(MockitoJUnitRunner::class)
+class EncryptedLogsFileProviderTest : BaseUnitTest() {
```
hichamboushaba, line 14: Same as [above](https://github.com/woocommerce/woocommerce-android/pull/14407/files#r2247887427)

WooCommerce/src/main/kotlin/com/woocommerce/android/util/logs/WooFileLogger.kt
```diff
@@ -0,0 +1,151 @@
...
+        private const val LOG_FILE_DIRECTORY = "logs"
+        private const val MAX_LOG_FILES = 7
+        private const val LOG_ENTRY_PREFIX = "--"
+    }
+}
+
+/**
+ * Buffers elements from a ReceiveChannel until a trigger Flow emits, then emits the buffered elements as a List.
+ */
+private fun <T> ReceiveChannel<T>.bufferUntil(trigger: Flow<Unit>): Flow<List<T>> {
```
hichamboushaba, line 155: This extension was inspired by the one shared [here](https://github.com/Kotlin/kotlinx.coroutines/issues/2193#issuecomment-705655058), our version just simpler, the other version is intended for "natural batching" depending on the backpressure, but our version uses a given trigger, and we already have a `Channel`, the logic is quite simple: On each round (outer `while`), wait for the trigger, then drain the channel buffer.

WooCommerce/src/test/kotlin/com/woocommerce/android/util/logs/LogFileWriterTest.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import java.io.File
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class LogFileWriterTest : BaseUnitTest() {
```
hichamboushaba, line 16: This was fully AI generated, I think it did a good job in covering all cases.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
```diff
@@ -0,0 +1,19 @@
...
+    fun provide(): File {
+        return File.createTempFile("log", "").apply {
+            val logs = runBlocking { wooLog.getCurrentLogEntries().take(LOG_ENTRIES_LIMIT) }
```
hichamboushaba: I'm using `runBlocking` here as the `EnqueueSendingEncryptedLogs` is not suspendable, and also their consumers. We were already writing to disk on the main thread before, so I hope this change is safe.

hichamboushaba: Given our focus on avoiding blocking calls, I updated the logic in f84551a to make sure we don't use `runBlocking`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/categories/AddProductCategoryFragment.kt
```diff
@@ -112,7 +114,7 @@ class AddProductCategoryFragment :
         }
 
         with(binding.productCategoryParent) {
-            runBlocking { viewModel.getSelectedParentCategoryName()?.let { setText(it) } }
+            viewLifecycleOwner.lifecycleScope.launch { viewModel.getSelectedParentCategoryName()?.let { setText(it) } }
```
JorgeMucientes: Is there a reason for adding this change in this PR? Seems completely unrelated to the logging changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/WooLog.kt
```diff
@@ -52,14 +58,21 @@ object WooLog {
     enum class LogLevel { v, d, i, w, e }
 
     const val TAG = "WooCommerce"
-    private const val MAX_ENTRIES = 999
-    val logEntries = RollingLogEntries(MAX_ENTRIES)
 
-    init {
+    private lateinit var fileLogger: WooFileLogger
```
JorgeMucientes, line 64: I just wanted to surface that the initialization of `WooFileLogger` will make the app crash when Leak Canary is enabled: 
```
java.lang.RuntimeException: Unable to get provider com.woocommerce.android.LeakCanaryInstaller: kotlin.UninitializedPropertyAccessException: lateinit property fileLogger has not been initialized
at android.app.ActivityThread.installProvider(ActivityThread.java:7770)
```
I didn't look too much into the details; I just disabled LeakCanary for testing this. But look like the call `WooLog.init(context)` from `AppConfigModule` has not happened by the time LeakCanaryInstalled is initialised.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
```diff
@@ -0,0 +1,26 @@
...
+            val logs = wooLog.getCurrentLogEntries().take(LOG_ENTRIES_LIMIT)
+
+            File.createTempFile("log", "").apply {
+                appendText(logs.joinToString("\n"))
+            }
+        }
+    }
+
+    companion object Companion {
+        private const val LOG_ENTRIES_LIMIT = 1000
```
JorgeMucientes, line 24: Curious what the prior limit was before this was added?

WooCommerce/src/main/kotlin/com/woocommerce/android/util/logs/WooFileLogger.kt
```diff
@@ -0,0 +1,151 @@
...
+                trigger = merge(
+                    flow {
+                        delay(FLUSH_PERIOD_MS)
+                        emit(Unit)
+                    }, // periodic flush
+                    writeTrigger // flush on demand
+                )
+            )
+                .collect { processLogs(it) }
+        }
```
JorgeMucientes: I'm not sure I fully understand, but will the current logic batch logs only for 500ms? Meaning will be writing into a file log batches every 500ms? Isn't that too frequent, considering we are doing costly operations (writing into a file) all the time while the app is running?

JorgeMucientes (approved):
Hey @hichamboushaba great work here! I just left a few questions to better understand the changes and also reported one issue that I think should be addressed. The issue will only impact us devs, but still might be worth checking.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/logs/WooFileLogger.kt
hichamboushaba: > I'm not sure I fully understand, but will the current logic batch logs only for 500ms?

Yes, that's the actual logic, we'll keep buffering for 500ms, then we'll flush the log, if the buffer is empty, we'll keep waiting.

I agree that 500ms is a bit aggressive, but there is some tradeoff between performance and loss of logs here, as the app can be closed at any time, the longer the flushing period, the higher the chances we might lose some logs, and this can happen in two occasions:
- The app crashes.
- The app gets killed by the system.

In the commit 0bec0d7, I made some adjustments to try to keep the performance impact lower: we'll now use `5 seconds` when the app is in the foreground, as the chances of the app being killed are quite low, and when the app enters the background, we'll reduce the period to `500ms`, and given we don't write that many logs when the app is in the background, we won't even hit this limit that often.

This change means that we might lose more logs if the app gets crashed, I think that's fine for now, and it's still a big improvement over what we had before. I'll try to find some time to work on an additional improvement to get crash logs logged too, and to flush during crashes.

WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
hichamboushaba: The same (just minus 1 😅), we had a limit of `999` entries in `WooLog` itself ([here](https://github.com/woocommerce/woocommerce-android/blob/64209047a54ec98340e97970bbff5501607e116f/WooCommerce/src/main/kotlin/com/woocommerce/android/util/WooLog.kt#L55))

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/categories/AddProductCategoryFragment.kt
hichamboushaba: This was an accidental change as I was investigating the DB blocking calls 😄, I'll revert it before merging this, as I already fixed this spot in a separate PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/WooLog.kt
hichamboushaba: Nice catch, I addressed this in 5930af61ecb1660c3e5fc3f127f5a32b82773be0

hichamboushaba:
Thanks @JorgeMucientes for the review, I addressed your comments, this is ready for another round to check my replies and last changes.
I'll address the conflcits later.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/logs/WooFileLogger.kt
JorgeMucientes: Thanks for looking again into this. I think these changes https://github.com/woocommerce/woocommerce-android/commit/0bec0d7ce605ee1939ac52e81ecf4bdbdc54a51e make complete sense. 

> This change means that we might lose more logs if the app gets crashed, I think that's fine for now,
I think that's a totally ok tradeoff. I was concerned about the impact the prior flush rate could have on the app's performance, given that logs are a dev tool for us. With the current approach, I think we are on the safe side of not impacting the app's performance at all.

JorgeMucientes (approved):

PR #14415, opened 2025-07-31
Fix input background color on the shipping selection screen

malinajirka:
Fixes WooMob-952

### Description

This PR updates background color for the `Name` input field on the Add Shipping screen. 

### Testing information

1. Start order creation
2. Add a product to the cart
3. Tap on Add shipping
4. Notice the background for `Name` input looks fine in both light and dark mode

The above

### Images/gif

###Before

https://github.com/user-attachments/assets/470519b2-9140-403d-89f4-3c1efe4b05ce

### After

https://github.com/user-attachments/assets/dc146e9a-90b0-423d-9c83-e2335abef533


hichamboushaba (approved):
Thanks @malinajirka, the fix works as expected.

Just two minor notes, just for awareness, nothing needs to be addressed:
- This is not related to shipping labels, it's about the shipping method of orders.
- I'm not sure if we need to use BasicTextField and not one of the standard TextFields, I didn't check this deeper; it's just that the BasicTextField should be used rarely as it's not themed correctly, which can lead to inconsistencies like what we had here, it's generally a building block for apps not wishing to use the Material themed ones.

PR #14419, opened 2025-08-01
[Dependabot] Add `automattic-encryptedlogging` to Ignore Configuration

ParaskP7:
Similar: [JP/WPAndroid#22082](https://github.com/wordpress-mobile/WordPress-Android/pull/22082)

---

### Description

Having #14354 merged, we now have one more of our libraries that are stored in S3 and have a custom versioning scheme which doesn't work with :dependabot:. Thus this library should be ignored as well, just like the rest.

FYI: At least for now and until we make this whole :dependabot: process work for such S3 libraries of ours as well.

---

### Testing information

Not needed, we just make sure that no :dependabot: PRs (like [this](https://github.com/wordpress-mobile/WordPress-Android/pull/22063)) gets opened from here onwards and for this repo. 

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,6 +23,7 @@ updates:
       - dependency-name: "org.wordpress:utils"
       - dependency-name: "com.automattic:Automattic-Tracks-Android"
       - dependency-name: "com.automattic.tracks:experimentation"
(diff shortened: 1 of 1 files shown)


wzieba (approved):
thanks!

PR #14420, opened 2025-08-01
Issue/woomob 735 update compile sdk version to 36

JorgeMucientes:
Closes: WOOMOB-735

## Description
Please don't panic about the number of file changes. Most of them are really small ones. It is the testing of the changes that will be more cumbersome, that's why I've assigned 2 of you for the testing phase. 

This PR upgrades Android `compileSdk` value from 15 to 16. Additionally, it adds the following changes to "comply" with the `compileSdk` upgrade: 
- Upgrade AGP to the latest version (consequently had to update Gradle to `8.13` as well). This will fix the following warnings:
> We recommend using a newer Android Gradle plugin to use compileSdk = 36
This Android Gradle plugin (8.8.0) was tested up to compileSdk = 35.

- Fixed all the "Deprecated API use" lint errors after upgrading compileSdk to API 16. Some of them were straightforward, some others weren't. And there is one change I was not able to fix as I think it might require a significant refactor. I'll share more about that in the code. 

Given the number of test scenarios, I'll assign different testers. 

### Deprecated Locale() usages

**Scenario 1**
Commit: [2be78ff](https://github.com/woocommerce/woocommerce-android/pull/14420/commits/2be78ffa3a0884307a19bd883a2a95b92c543956)
Tester: @irfano 
Steps:
1. Apply this patch [Test_new_Locale_constructor1.patch](https://github.com/user-attachments/files/21584444/Test_new_Locale_constructor1.patch)
2. In a physical device with tap to pay, navigate to More Menu > Payments > About Tap to Pay
3. Check that the Store's country name is properly displayed

 

**Scenario 2**
Commit: [8e96e77](https://github.com/woocommerce/woocommerce-android/pull/14420/commits/8e96e776ef781a92f3bd820c02c38bcbcfee8be6)
Tester: @irfano 
Steps:
1. Open any order detail and check that shipping and billing addresses are properly displayed and that the country name is displayed correctly:

 

### Deprecated Accessibility Usages
Most of the changes are related to the Android accessibility talkback service. So enable TalkBack first: 
- Go to Device Settings > Search for TalkBack. If nothing is shown in results, you'll need to install Android Accessibility Suite from Google Play.
- Enable TalkBack. I suggest enabling `TalkBack Shortcut` option for convenience to turn on/off Talkback just when needed. 
- I also suggest testing this with two devices side by side, one with this PR deployed and another with `trunk` deployed, and compare the talkback output. 

**Scenario 1**
Commit: [f9d9fed](https://github.com/woocommerce/woocommerce-android/pull/14420/commits/f9d9fed9e66e0d7ead7d7a20c03c3829124d4d41)
Tester: @irfano 
Steps
1. Navigate to the orders tab
3. Tap create order
5. Tap "Add customer details" and select a customer
6. Tap the edit icon in the customer section. 
7. With talkback enabled, tap on `Billing Address` and `Shipping Address` headers and check that talkback announces them as headers

 

 **Scenario 2**
Commit: [b88b9ed](https://github.com/woocommerce/woocommerce-android/pull/14420/commits/b88b9edbbce338323808ba4faee9ae5789f1b524)
Tester: @irfano 
Steps
1. With TalkBack enabled, open the order list
2. Tap on any of the headers from the list 
4. 
 

 **Scenario 3**
Overall, check the talkback output for card reader connection management makes sense. You can compare the prior implementation for reference. There are some slight differences with the prior implementation. 
Commit: Multiple commits here.
Tester: @kidinov 
Steps
1. Enable simulated card reader from settings if needed and set "Update Simulated Card Reader" to always. 
2. Navigate to More Menu > Payments > Manage Card Reader
3. Tap Connect Reader and ensure TalkBack output makes sense and describes the different statuses

 **Scenario 4**
Commit: [31c340a](https://github.com/woocommerce/woocommerce-android/pull/14420/commits/31c340a30c3b3b4602825b5122ce2b6d8ea59977)
Tester: @kidinov 
Steps
1. Open POS
2. Add and remove products and coupons to the cart
3. Check TalkBack announces when a product/coupon is added or removed from the cart as it was before

The above ☝🏼 

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditCustomerAddFragment.kt
```diff
@@ -203,12 +203,9 @@ class OrderCreateEditCustomerAddFragment :
         )
     }
 
-    private fun LayoutAddressFormBinding.setAccessibilityHeaders(detailsHeading: Int, addressHeading: Int) {
-        detailsHeaderContainer.announceForAccessibility(detailsHeading.toString())
+    private fun LayoutAddressFormBinding.setAccessibilityHeaders(addressHeadingTextRes: Int) {
+        addressSectionHeader.setText(addressHeadingTextRes)
         ViewCompat.setAccessibilityHeading(detailsHeaderContainer, true)
```
JorgeMucientes, line 208: When testing this on `trunk`, you'll notice that this announceForAccessibility is really not needed. Talkback will already read the section title and inform its a header. Btw I think informing that the text is a header and re-reading it is redundant imho, but I have kept the `setAccessibilityHeading` calls to maintain the same behavior.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/ext/WooPosContextExt.kt
```diff
@@ -32,6 +32,7 @@ fun Context.announceForAccessibility(text: String) {
             @Suppress("DEPRECATION")
             AccessibilityEvent.obtain()
         }
+        @Suppress("DEPRECATION")
```
JorgeMucientes: Hey @kidinov, this `announceForAccessibility()` function is heavily used for POS. I didn't find an easy way to resolve this deprecation, and I think it might require a significant refactor on how announcements for talkback are triggered. 
So, for now, in order to avoid blocking merging the rest of the PR changes, I'm simply suppressing this deprecation. However, I think a Linear issue should be created to fix this.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/local/LocalNotificationWorker.kt
```diff
@@ -35,13 +36,16 @@ class LocalNotificationWorker @AssistedInject constructor(
...
+        if (!wooNotificationBuilder.isNotificationsEnabled()) return Result.failure()
+
         val type = inputData.getString(LOCAL_NOTIFICATION_TYPE)
         val notificationId = inputData.getInt(LOCAL_NOTIFICATION_ID, -1)
         val title = inputData.getString(LOCAL_NOTIFICATION_TITLE)
         val description = inputData.getString(LOCAL_NOTIFICATION_DESC)
         val data = inputData.getString(LOCAL_NOTIFICATION_DATA)
         val siteId = inputData.getLong(LOCAL_NOTIFICATION_SITE_ID, 0L)
 
+        @SuppressLint("MissingPermission")
```
JorgeMucientes: I've added a check right above in line 39 to ensure that when local notifications are triggered, we have permission to display them. The suppress warning is needed because Lint is unable to determine that our function `wooNotificationBuilder.isNotificationsEnabled()` is checking for the permission. 

Same for the above [change](https://github.com/woocommerce/woocommerce-android/pull/14420/files#diff-47efb07bf9de6fd10988a566a6a5465c091b83b390620e47b5119d0323a3df03R187) in `WooNotificationBuilder.kt`

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/local/LocalNotificationWorker.kt
kidinov: Done - WOOMOB-1005 !

JorgeMucientes: Thanks for opening the issue @kidinov 🙏🏼 
I just noticed that you also removed yourself from the PR request, but I'd still like you or someone with knowledge on CardReader and POS code to review the current changes. In the PR description, I've tagged you in the specific scenarios and commits that should be reviewed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/update/CardReaderUpdateDialogFragment.kt
```diff
@@ -57,9 +57,6 @@ class CardReaderUpdateDialogFragment : PaymentsBaseDialogFragment(R.layout.card_
                 is SoftwareUpdateProgress ->
                     announceSoftwareUpdateProgress(event.progress, binding)
 
-                is SoftwareUpdateAboutToStart ->
```
kidinov, line 60: `SoftwareUpdateAboutToStart` is unused now and can be deleted

kidinov (approved):
All looks good to me! I left one comment, though

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/local/LocalNotificationWorker.kt
irfano: We are actually already checking the permission in [PreconditionCheckWorker](https://github.com/woocommerce/woocommerce-android/blob/8d0a8c56c1f40c6f707e1279ebd0dffdf1826fa2/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/local/PreconditionCheckWorker.kt#L33). I suspected that `PreconditionCheckWorker` might be running **before** scheduling, or **at the time** that notification is triggered. It's the latter. 

I tested a local notification that was triggered 8 seconds after the app launched, and both `PreconditionCheckWorker` and `LocalNotificationWorker` ran at the same time. So, I believe we can remove this new condition added here.

image

irfano (changes requested):
Thanks for handling this task and for the detailed instructions, @JorgeMucientes. Everything works as expected. I just noticed a behavioral difference and would like you to test it as well. The "Connecting to reader" header is announced twice. I couldn’t reproduce this on `trunk`.

[after.webm](https://github.com/user-attachments/assets/b4fc82ac-ba38-4840-9d8d-662a89ed9433)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartScreen.kt
```diff
@@ -293,40 +293,41 @@ private fun CartBodyWithItems(
...
                         currentItem.message
                     )
 
                 is WooPosCartItemViewState.Loading ->
-                    localView.context.getString(R.string.woopos_cart_searching_for_item)
+                    context.getString(R.string.woopos_cart_searching_for_item)
             }
 
-            localView.announceForAccessibility(message)
+            context.announceForAccessibility(message)
```
JorgeMucientes: Hey folks @kidinov @toupper , I think [this PR](https://github.com/woocommerce/woocommerce-android/pull/14440/files)  that removes the function `fun Context.announceForAccessibility()` hasn't handled migrating this part. So now when I update this PR with `trunk` the code won't compile. 
I see 2 options here: 
- Create a new branch from this PR to migrate how talkback updates are handled in `WooPosCartScreen.kt`
- I can revert the changes added to `WooPosCartScreen.kt` in this PR so that everything compiles again. And then you can start a new branch from `trunk` to migrate the existing deprecated usage of `localView.announceForAccessibility(message)` ini WooPosCartScreen.kt 

Let me know how you want to proceed and I'll update this PR accordingly 🙏🏼

toupper: Thanks for the ping @JorgeMucientes! Yeah, In the previous [PR](https://github.com/woocommerce/woocommerce-android/pull/14440) WooPosCartScreen wasn't changed as it relied on on the `localView::announceForAccessibility` instead of the context, which was where the deprecation came.
I lean towards number 2, I can migrate it in `WooPosCartScreen` as well. But before I wanted to understand, why do we need here to change from calling the `localView` property to the `context` one? Thanks a lot!

JorgeMucientes: Hey @toupper 

> I lean towards number 2, I can migrate it in WooPosCartScreen as well.

Perfect I've reverted the changes and added a @Suppress deprecation warning to the deprecated code so the PR and code compiles correctly. 

> why do we need here to change from calling the localView property to the context one?

The reason I did that was to be able to reuse the extension function `Context.announceForAccessibility()` and remove direct usages of the deprecated function `localView::announceForAccessibility`. But then `Context.announceForAccessibility()` turned out to be also using a deprecated API that had to be replaced. So the change wasn't very useful in the end.

toupper: I see, thanks for adding more context @JorgeMucientes Could you please revert the changes here as described in your second point above and create a Linear issue for the next task? I will then take care of it. Thanks a lot

JorgeMucientes:
Hey @irfano good eye! Thanks for the review. 
I've tested the bug you mentioned and couldn't reproduce it consistently. There seems to be a race condition related to the simulated card reader. I don't have a real card reader with me right now, but I'd say this is very unlikely to happen with a real connection. 
My hypothesis after debugging this minor issue is that with the simulated card reader, it transitions almost instantly from this status: 
Screenshot 2025-08-11 at 17 19 55

To this other one: 
Screenshot 2025-08-11 at 17 21 32

Because the header of that dialog is now an accessibility "Live Region" any updates to those headers will be enqueued for talkback to read. By the time the talkback start talking the first value "Scanning readers" is already gone, however there are 2 events in talkback queue to read that header value. 

If you apply the following patch, you'll be able to simulate a more realistic connection and see that the issue is gone because talkback has some time to read the first value for the header:
[Fix_detekt_indentation_issues.patch](https://github.com/user-attachments/files/21717985/Fix_detekt_indentation_issues.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartScreen.kt
JorgeMucientes: Sure! Done @toupper : https://linear.app/a8c/issue/WOOMOB-1024/remove-deprecated-use-of-localviewannounceforaccessibility

As soon as this PR is merged feel free to tackle removing that deprecated function 🙏🏼

toupper: Thanks @JorgeMucientes!

irfano (approved):
Thanks for digging deeper.
I tested your patch, but the headers are still announced twice:

[Screen_recording_20250812_115648.webm](https://github.com/user-attachments/assets/1da19304-7d6c-4730-b6da-6d8b60ac39d7)

But I noticed that on the first screen, the focused view is the description, so TalkBack announces the header first and then the focused view. So, we don't have any issue on this screen.

image

On subsequent screens, the header receives focus instead of the description. As a result, TalkBack first announces the header as the first element and then announces it again as the focused view.

image

I’m not entirely sure, but this may be expected behavior: the user first hears which screen is shown, then the currently focused view. We could adjust which view receives initial focus, but it’s not a major issue. I’m approving the PR, also thanks for addressing the other feedback. 🚢

JorgeMucientes:
Hey @irfano, thanks for looking into this again. In my case with the patch, none of the headers are read twice:

[Screen_recording_20250812_182136.webm](https://github.com/user-attachments/assets/6bd97423-8649-4a50-9666-c6427c3cf900)


It is weird to see how Talkback behaves differently on your device than on mine, especially considering Talkback is an external app, so it's not dependent on device OS. 

Anyway, as we talked, this is not a critical issue at all, so I don't think it is worth spending more time on it. I'll proceed with the merge.

PR #14422, opened 2025-08-01
[WOOMOB-961] Remove unused libsapachehttpclientandroid dependency

malinajirka:
Fixes WOOMOB-961

### Description
This PR removes the unused libs.apache.http.client.android dependency from the project. The dependency was identified as no longer needed and removing it helps reduce the app size and potential security surface area.

### Testing information

Green CI should be enough.

I ran `./gradlew :WooCommerce:assembleWasabiDebugAndroidTest` to verify the dependency truly wasn't used.

### Images/gif 
Not applicable for dependency removal.

### DIFF
--- a/WooCommerce/build.gradle
+++ b/WooCommerce/build.gradle
@@ -369,6 +369,7 @@ dependencies {
             because("version shipped with WireMock 2.26.3 contains security vulnerabilities")
         }
     }
+    androidTestImplementation(libs.apache.http.client.android)
 
     implementation(libs.zendesk.support) {
         exclude group: 'com.android.support', module: 'support-annotations'
--- a/WooCommerce/build.gradle
+++ b/WooCommerce/build.gradle
@@ -369,7 +369,6 @@ dependencies {
             because("version shipped with WireMock 2.26.3 contains security vulnerabilities")
         }
     }
-    androidTestImplementation(libs.apache.http.client.android)
 
     implementation(libs.zendesk.support) {
         exclude group: 'com.android.support', module: 'support-annotations'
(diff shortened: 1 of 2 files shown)


samiuelson (approved):
LGTM! Thanks.

samiuelson (changes requested):
Hm, actually it looks like the libs.apache.http.client.android dependency is [needed](https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.edfd947f2636efe3/matrices/8644592129489850253/details?stepId=bs.57c380e1708fb961&testCaseId=2) for UI tests.

malinajirka:
Interesting, I tried to build the UI tests but it's likely using reflection or something like that. @ParaskP7  is there a way how to mark the dependency as "used" in the dependency dashboard?

ParaskP7:
👋 @malinajirka !

> Interesting, I tried to build the UI tests but it's likely using reflection or something like that. @ParaskP7 is there a way how to mark the dependency as "used" in the dependency dashboard?

Do you mean something like [that](https://github.com/woocommerce/woocommerce-android/blob/trunk/libs/fluxc/build.gradle#L123-L130)?

```gradle
dependencyAnalysis {
    issues {
        onUnusedDependencies {
            // This dependency is actually needed otherwise the app will crash with a runtime exception.
            exclude(libs.eventbus.android.get().module.toString())
        }
    }
}
```

malinajirka:
Nice, that's exactly what I mean - thank you 🙇 !

malinajirka (commented):
@samiuelson Good catch, I've reverted the removal and suppressed the `unused` warning.

samiuelson (approved):
LGTM. Thanks! :shipit:

PR #14425, opened 2025-08-01
[Hack Week] App logs improvements: display all stored log files

hichamboushaba:
Closes WOOMOB-948

> [!IMPORTANT]
> Please don't merge, we have a PR that migrates the `WooLogViewerScreen` to Material 3 #14406, and merging this will cause conflicts, so let's wait until it's merged, and I'll rebase my branch. 

### Description
This PR updates the `WooLogViewerActivity` to display a list of the available log files, and tapping on each log file will open its content, this is similar to what the iOS app does.

> [!NOTE]
> For time constraints, I took some shortcuts in the implementation, we have a single ViewModel with both the list and detail states, this was simpler to implement, and I think it does the job well. 

### Steps to reproduce
1. To simulate having multiple log files, change the device date, then open the app, this will create a log file for each date.
2. Open the application log viewer from the support screen.

### Testing information
- Confirm that the list of last seven log files is shown.
- Tap on the log files, and confirm the correct contents are shown.
- Test tapping on Copy and Share.

The above.

### Images/gif

https://github.com/user-attachments/assets/464ade8a-4b48-4858-8af7-5e305c447c4f

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
@@ -75,27 +75,18 @@ class WooLogViewerViewModel @Inject constructor(
     }
 
     private suspend fun prepareLogFileContentState(logFile: LogFile): UiState.LogFileContent {
-        val logContent = wooFileLogger.getLogFileContent(logFile.name).orEmpty()
+        fun getShareLogs(entries: List<LogEntry>): String {
+            return (entries.takeLast(MAX_LOG_ENTRIES_TO_SHARE) + getDeviceInfo())
+                .joinToString("\n")
+        }
+
+        val logEntries = wooFileLogger.getLogFileContent(logFile.name).orEmpty()
         return UiState.LogFileContent(
             logFile = logFile,
-            logContent = logContent,
-            onBackPressed = {
-                selectedLogFile.value = null
-            },
-            onShareClicked = {
-                triggerEvent(
-                    ShareLogs(
-                        logs = (logContent + getDeviceInfo()).joinToString("\n")
-                    )
-                )
-            },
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooLogViewerViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+            },
+            onLogFileSelected = { selectedFile ->
+                selectedLogFile.value = selectedFile
+            }
+        )
+    }
+
+    private suspend fun prepareLogFileContentState(logFile: LogFile): UiState.LogFileContent {
+        fun getShareLogs(entries: List<LogEntry>): String {
+            return (entries.takeLast(MAX_LOG_ENTRIES_TO_SHARE) + getDeviceInfo())
```
hichamboushaba, line 82: Android enforces a limit on the size of data we pass to Intents, so we limit here to the last 1000 entries, I think this limit will be rarely hit by users, but even if does, getting the 1000 last entries should be fine, let me know what you think?

JorgeMucientes: Sounds good to me 👍🏼

JorgeMucientes (approved):
Excellent work @hichamboushaba, works like a charm. :shipit:

PR #14446, opened 2025-08-08
Active plugin info fetching improvements

samiuelson:
# Active plugin info fetching improvements

Closes: Woomob-927

### Description

This PR fixes the places where the plugin info is fetched, but we've not been checking if the plugin is activated, which was incorrect:
* [DetermineProductPasswordApi.kt](https://github.com/woocommerce/woocommerce-android/pull/14446/files#diff-52b05e246de600745346bd65a56911fd4f6f4aff351fa1ea240854186b602fa1)
* [ProductListViewModel.kt](https://github.com/woocommerce/woocommerce-android/pull/14446/files#diff-2e7178ef01897149331a3724ea96c05908d2b147d0d9f2a559ca09cb5ba81612) - the logic has been correct, just a cosmetic change.

### Steps to reproduce

N/A

### Testing information

Check the code. Optionaly some test the app.

I smoke tested the app. 

### Images/gif

—


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/DetermineProductPasswordApi.kt
```diff
@@ -14,7 +14,7 @@ class DetermineProductPasswordApi @Inject constructor(
 ) {
     suspend operator fun invoke(): ProductPasswordApi {
         val wcVersion = withContext(Dispatchers.IO) {
-            wooCommerceStore.getSitePlugin(
+            wooCommerceStore.getActiveSitePlugin(
```
kidinov, line 17: I’d like to confirm this with @hichamboushaba - is the active plugin required?

kidinov (approved):
LGTM, but lets wait confirmation from @hichamboushaba before merging

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/DetermineProductPasswordApi.kt
hichamboushaba: Thanks for the ping @kidinov, as the plugin we are checking here is `WOO_CORE`, then `getSitePlugin` and `getActiveSitePlugin` would generally behave the same, the user can't access the app without having Woo itself active.

So for me, it's fine to change it.

hichamboushaba: Looking at [this ](https://github.com/woocommerce/woocommerce-android/pull/14368), I understand now better the origin of the change, so a 👍 from me.

Just a question, did we confirm if checking just the plugin name and not the `directory/name` won't have any unwanted side effects? I mean, can we be sure that plugins can't have duplicate `names` in WordPress.org (for example two plugins `dir1/plugin` and `dir2/plugin`), I guess that using `woocommerce` won't be allowed for sure, but I'm wondering about other plugins.

samiuelson: In case of WC plugin, we experienced the opposite - sometimes the `dir` was modified by the host or because we were using `dev` plugin build. So in the case here, I believe this is a correct solution. 

The same (potentially) applies to Square plugin which is maintained by A8c. 

I tested a scenario of multiple plugins with the same name (different dir) in WP-admin. It's impossible to activate more than one plugin with the same name.

PR #14449, opened 2025-08-11
[WOOMOB-1015] Fix WooPos cash payment screen behavior consistency

malinajirka:
Fixes WOOMOB-1015

### Description
This PR implements consistent behavior for the POS cash payment screen:

1. **Pre-filled total amount**: The total amount is now pre-entered and selected when the cash payment screen opens, allowing users to easily replace it by typing.
2. **Button state validation**: The "Mark as completed" button is disabled unless the entered amount is equal to or higher than the total amount.
3. **Analytics tracking**: Updated tracking

### Testing information

   - Open the POS and add items to the cart.
   - Navigate to checkout and select "Cash payment" button.
   - Verify "checkout_cash_payment_tapped" is tracked.
   - Verify the total amount is pre-filled and selected in the input field.
   - Verify the "Mark as complete" button is enabled.
   - Start typing and verify the pre-filled amount is replaced. Also verify the button is disabled unless the amount is equal or higher than total.
   - Turn on airplane mode and tap on "Mark as complete" - verify "cash_payment_tapped" and "cash_payment_failed" is tracked.
   - Turn off airplane mode and tap on "Mark as complete" - verify "cash_payment_tapped" and  "cash_collect_payment_success" is tracked.

- The above

### Images/gif

https://github.com/user-attachments/assets/33472a2c-9ddb-4626-b302-ae95038ad65c

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
@@ -84,7 +84,7 @@ fun WooPosMoneyInputField(
         stateSaver = TextFieldValue.Saver,
     ) {
         currentValue = value
-        mutableStateOf(TextFieldValue(valueMapper.printValue(value)))
+        mutableStateOf(TextFieldValue(valueMapper.printValue(value), TextRange(0, value.toString().length)))
     }
 
     var labelWidth by remember { mutableIntStateOf(0) }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
@@ -56,6 +56,7 @@ fun WooPosMoneyInputField(
     keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
     keyboardActions: KeyboardActions = KeyboardActions.Default,
     contentAlignment: Alignment = Alignment.CenterStart,
+    preselectText: Boolean = false,
 ) {
     val visualTransformation = remember {
         CurrencyVisualTransformation(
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
```diff
@@ -84,7 +84,7 @@ fun WooPosMoneyInputField(
         stateSaver = TextFieldValue.Saver,
     ) {
         currentValue = value
-        mutableStateOf(TextFieldValue(valueMapper.printValue(value)))
+        mutableStateOf(TextFieldValue(valueMapper.printValue(value), TextRange(0, value.toString().length)))
```
kidinov: I think this is should be optional and by default the previous behavior should be kept

kidinov (approved):
LGTM! I left one suggestion, please take a look

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
malinajirka: I believe providing more options makes the POS more complex - I'd personally vote for opinionated design over flexibility, unless the behavior could make the POS unusable for some folks, which isn't the case here I believe. We could remove it, if we believe it won't bring value for most folks.

kidinov: Yes, but this is POS wide component, and in many/most cases the `WooPosMoneyInputField` should not be selected by default, for instance you editing order or creating coupon and those are screens with many input fields, therefore I think we should support "autoSelectAll" option on this component, and it should be:
* An Option
* By default it should be false

malinajirka: Ahh, I see - you mean making it optional on the code level, not on the user level. Sure, I'll update it.

malinajirka: Fixed in https://github.com/woocommerce/woocommerce-android/pull/14449/commits/9466a4fb736cf0fe5dc9d17228e5a378af28836c

PR #14457, opened 2025-08-12
Handle null values when API returns null user roles

JorgeMucientes:
Closes: WOOMOB-471

### Description

Fixes NullPointerException crash https://a8c.sentry.io/issues/4727853296/?project=1459556
Makes user roles a nullable field and ensures we handle the case where the API returns a null value for the roles by providing a new error message. 

### Steps to reproduce

I don't know in which rare cases the API may end up returning a null value for roles. I suspect a custom roles plugin like https://easydigitaldownloads.com/ my be messing up with this. 
Apply the following patch to simulate the API returning null.
[Simulate_API_returning_null_value.patch](https://github.com/user-attachments/files/21735374/Simulate_API_returning_null_value.patch)

### Testing information
1. Apply the above patch
2. Log into the app using WP.com credentials
3. Check the below error message is displayed

 

Repeat the same steps but log in with site credentials. 

The above. 

### DIFF
--- a/WooCommerce/src/main/res/values/strings.xml
+++ b/WooCommerce/src/main/res/values/strings.xml
@@ -290,7 +290,7 @@
     <string name="login_troubleshooting_tips">Read our troubleshooting tips</string>
     <string name="login_discovery_error_option_select">Select an option</string>
     <string name="login_no_wpcom_account_found">Your email isn\'t used with a WordPress.com account.</string>
-    <string name="user_role_access_error_user_roles_null">This app supports only Administrator and Shop Manager user roles. We were unable to fetch user roles for your account: %1$s. Please contact support.</string>
+    <string name="user_role_access_error_user_roles_null">This app supports only Administrator and Shop Manager user roles. We were unable to fetch user roles for your account. Please contact support.</string>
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/res/values/strings.xml
```diff
@@ -290,6 +290,7 @@
     <string name="login_troubleshooting_tips">Read our troubleshooting tips</string>
     <string name="login_discovery_error_option_select">Select an option</string>
     <string name="login_no_wpcom_account_found">Your email isn\'t used with a WordPress.com account.</string>
+    <string name="user_role_access_error_user_roles_null">This app supports only Administrator and Shop Manager user roles. We were unable to fetch user roles for your account: %1$s. Please contact support.</string>
```
irfano: I’m not sure if we really need to display the username in this message, since it’s already shown in the header.
I’ll suggest an alternative approach, let me know what you think:

- Remove `: %1$s` from `user_role_access_error_user_roles_null`
- In the line after the header, where it currently displays `null`, replace it with `@username`.

Screenshot_20250812_154449

irfano (approved):
I think this is a good approach to resolve the issue. I’m pre-approving it, but I’ve also shared a suggestion in the comments.

JorgeMucientes:
Thanks for the review and suggestion @irfano. Changes applied:

PR #14463, opened 2025-08-13
[POS] Better error handling for eligibility

toupper:
Closes WOOMOB-1016

Only one review required, but I add you both as reviewers in case you are interested in the solution.

### Description

With this PR we improve the logic for handling errors when opening POS, including when there is no connection. We implement this logic:

POS tab eligibility
- Positive → Show POS tab.
- Negative → Hide POS tab.
- Unknown (e.g., backend error) → Show tab using last cached value.
When entering POS tab and eligibility check errors again:
- Cached Positive → Allow access to POS.
- No Cache → Deny access and show error message.

For that we:
- Update `AppPrefs` to cache also whether POS was launchable. We also update the tab logic to only make it possible to cache positive cases.
- Update `WooPOSIsRemotelyEnabled`, `WooPosCanBeLaunchedInTab` and `WooPosTabShouldBeVisible` to return a result instead of a boolean
- Refactor `WooPosCanBeLaunchedInTab` with the new logic, including setting the value in cache.

### Steps to reproduce

With all conditions met:

- Start the app with internet connection
- Verify the POS tab is visible
- Tap on it and access POS
- Kill the app
- Turn on airplane mode
- Start the app
- Notice the POS tab is visible (before it was not)
- Tap on the tab and access POS

With the POS feature switch disabled in wp-admin:

- Start the app with internet connection
- Verify the POS tab is visible
- Go to the POS tab
- Verify that you get an error about the feature switch disabled
- Kill the app
- Turn on airplane mode
- Start the app
- Notice the POS tab is visible
- Go to the tab
- You cannot enter POS and a genetic message is shown

With a store in a country other than US or UK:

- Start the app with internet connection
- Verify the POS tab is not visible
- Kill the app
- Turn on airplane mode
- Start the app
- Notice the POS tab is not visible

See above

### Images/gif

On this video we can see how the tab is shown in airplane mode because it was shown before, but given that we never cached whether the POS was launched, we prevent accessing it:

https://github.com/user-attachments/assets/2228bf6f-8818-4843-ad8a-2fbed57987a4

Here we show how if previously the user accessed POS, it can access it even if there's no connection:

https://github.com/user-attachments/assets/9d05d4b9-0d97-4da0-a88f-80ae6cdaa398

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
@@ -210,6 +210,8 @@ object AppPrefs {
         APPLICATION_STORE_SNAPSHOT_TRACKED_FOR_SITE,
 
         POS_TAB_VISIBILITY,
+
+        POS_LAUNCHABLE
     }
 
     fun init(context: Context) {
@@ -1193,6 +1195,20 @@ object AppPrefs {
     fun isPOSTabVisibleForSite(siteId: Int): Boolean {
         return getBoolean(
             key = PrefKeyString("${UndeletablePrefKey.POS_TAB_VISIBILITY}:$siteId"),
+            default = false
+        )
+    }
+
+    fun setPOSLaunchableForSite(siteId: Int) {
+        setBoolean(
+            key = PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"),
+            value = true
+        )
+    }
+
+    fun isPOSLaunchableForSite(siteId: Int): Boolean {
+        return getBoolean(
+            key = PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"),
             default = false
         )
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
@@ -1199,6 +1199,10 @@ object AppPrefs {
         )
     }
 
+    fun clearPOSTabVisibilityForSite(siteId: Int) {
+        remove(PrefKeyString("${UndeletablePrefKey.POS_TAB_VISIBILITY}:$siteId"))
+    }
+
     fun setPOSLaunchableForSite(siteId: Int) {
         setBoolean(
             key = PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"),
@@ -1211,6 +1215,10 @@ object AppPrefs {
             key = PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"),
             default = false
         )
+    }
+
+    fun clearPOSLaunchableForSite(siteId: Int) {
+        remove(PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"))
     }
 
     /**
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
@@ -1183,10 +1183,10 @@ object AppPrefs {
         return getInt(PrefKeyString(channel.name), 0).takeIf { it != 0 }
     }
 
-    fun setPOSTabVisibilityForSite(siteId: Int, visible: Boolean) {
+    fun setPOSTabVisibilityForSite(siteId: Int) {
         setBoolean(
             key = PrefKeyString("${UndeletablePrefKey.POS_TAB_VISIBILITY}:$siteId"),
-            value = visible
+            value = true
         )
     }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
@@ -42,35 +42,67 @@ class WooPosCanBeLaunchedInTab @Inject constructor(
     @Suppress("ReturnCount")
     private suspend fun checkLaunchability(forceRefresh: Boolean = false): WooPosLaunchability {
         val site = selectedSite.getOrNull()
-            ?: return WooPosLaunchability.NotLaunchable(WooPosLaunchability.NonLaunchabilityReason.NoSiteSelected)
+            ?: return WooPosLaunchability.NotLaunchable(
+                reason = WooPosLaunchability.NonLaunchabilityReason.NoSiteSelected
+            )
 
         val cachedPositive = appPrefs.isPOSLaunchableForSite(site.id)
 
-        val wooCoreVersion = getWooCoreVersion(forceRefresh)
-        if (wooCoreVersion != null) {
-            getNonLaunchabilityReasonFromWooCoreVersion(wooCoreVersion)?.let {
-                return WooPosLaunchability.NotLaunchable(it)
-            }
-            getNonLaunchabilityReasonFromFeatureSwitch(wooCoreVersion, forceRefresh, cachedPositive)?.let {
-                return WooPosLaunchability.NotLaunchable(it)
-            }
-        } else if (!cachedPositive) {
-            return WooPosLaunchability.NotLaunchable(
-                WooPosLaunchability.NonLaunchabilityReason.UnknownNoPositiveCache
-            )
+        getNonLaunchabilityReasonFromVersionAndFeatureSwitch(forceRefresh, cachedPositive)?.let {
+            return prepareNotLaunchableStateWithCacheUpdate(site.id, it)
         }
 
-        val siteSettings = resolveSiteSettings(site, forceRefresh)
-            ?: return fallbackUnknownOrCache(cachedPositive)
+        getNonLaunchabilityReasonFromSiteSettingsAndCurrency(site, forceRefresh, cachedPositive)?.let {
+            return prepareNotLaunchableStateWithCacheUpdate(site.id, it)
+        }
 
-        if (isCountryAndCurrencySupported(siteSettings.countryCode, siteSettings.currencyCode)) {
-            appPrefs.setPOSLaunchableForSite(site.id)
-            return WooPosLaunchability.Launchable
+        appPrefs.setPOSLaunchableForSite(site.id)
+        return WooPosLaunchability.Launchable
+    }
+
+    private fun prepareNotLaunchableStateWithCacheUpdate(
+        siteId: Int,
+        reason: WooPosLaunchability.NonLaunchabilityReason
+    ): WooPosLaunchability.NotLaunchable {
+        if (reason != WooPosLaunchability.NonLaunchabilityReason.UnknownNoPositiveCache) {
+            appPrefs.clearPOSLaunchableForSite(siteId)
+        }
+
+        return WooPosLaunchability.NotLaunchable(reason)
+    }
+
+    private suspend fun getNonLaunchabilityReasonFromVersionAndFeatureSwitch(
+        forceRefresh: Boolean,
+        cachedPositive: Boolean
+    ): WooPosLaunchability.NonLaunchabilityReason? {
+        val wooCoreVersion = getWooCoreVersion(forceRefresh)
+            ?: return reasonIfNoPositiveCache(cachedPositive)
+
+        return getNonLaunchabilityReasonFromWooCoreVersion(wooCoreVersion)
+            ?: getNonLaunchabilityReasonFromFeatureSwitch(wooCoreVersion, forceRefresh, cachedPositive)
+    }
+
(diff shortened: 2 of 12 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
```diff
@@ -1183,10 +1185,10 @@ object AppPrefs {
         return getInt(PrefKeyString(channel.name), 0).takeIf { it != 0 }
     }
 
-    fun setPOSTabVisibilityForSite(siteId: Int, visible: Boolean) {
+    fun setPOSTabVisibilityForSite(siteId: Int) {
         setBoolean(
             key = PrefKeyString("${UndeletablePrefKey.POS_TAB_VISIBILITY}:$siteId"),
-            value = visible
+            value = true
```
malinajirka, line 1191: ❓ It seems we are storing always `true` but we never clear the cache, what if the server returns a valid response but it's `false`? Shouldn't we clear the cache in that case?

WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
```diff
@@ -1197,6 +1199,20 @@ object AppPrefs {
         )
     }
 
+    fun setPOSLaunchableForSite(siteId: Int) {
+        setBoolean(
+            key = PrefKeyString("${UndeletablePrefKey.POS_LAUNCHABLE}:$siteId"),
+            value = true
```
malinajirka, line 1209: ❓ Same as above.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -38,51 +42,75 @@ class WooPosCanBeLaunchedInTab @Inject constructor(
...
-                WooPosLaunchability.NonLaunchabilityReason.NoSiteSelected
-            )
+            ?: return WooPosLaunchability.NotLaunchable(WooPosLaunchability.NonLaunchabilityReason.NoSiteSelected)
 
-        val wooCoreVersion = if (forceRefresh) {
-            fetchWooCoreVersion()
+        val cachedPositive = appPrefs.isPOSLaunchableForSite(site.id)
+
+        val wooCoreVersion = resolveWooCoreVersion(forceRefresh)
+            ?: return fallbackForMissingWooCore(cachedPositive)
```
malinajirka: ❓ If I'm reading this correctly, this returns `launchable` if the request fails and we have positive value in the cache. Is this intentional? Shouldn't we continue checking the site settings related conditions in such case?

Something like:
```
resolveWooCoreVersion(forceRefresh)?.let { wooCoreVersion ->
            validateWooCoreVersion(wooCoreVersion)?.let { return it }
            evaluateFeatureSwitch(wooCoreVersion, forceRefresh, cachedPositive)?.let { return it }

        } ?: run {
            if (!cachedPositive) {
                return WooPosLaunchability.NotLaunchable(
                    WooPosLaunchability.NonLaunchabilityReason.WooCommercePluginNotFound
                )
            }
        }
```

P.S. Perhaps, it's an over-optimization, I'm not sure, so take this comment with a grain of salt.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -38,51 +42,75 @@ class WooPosCanBeLaunchedInTab @Inject constructor(
...
+        if (!isFeatureSwitchSupported(wooCoreVersion)) return null
+        val remote = isRemotelyEnabled(forceRefresh)
+        return when {
+            remote.isSuccess && !remote.getOrThrow() ->
+                WooPosLaunchability.NotLaunchable(WooPosLaunchability.NonLaunchabilityReason.FeatureSwitchDisabled)
+            remote.isFailure && cachedPositive ->
+                WooPosLaunchability.Launchable
+            remote.isFailure && !cachedPositive ->
+                WooPosLaunchability.NotLaunchable(WooPosLaunchability.NonLaunchabilityReason.UnknownNoPositiveCache)
+            else -> null
```
malinajirka: ❓ When would this `else` be evaluated? It seems like it's an unexpected state, right? I think we should crash or at the very least log an exception to Sentry. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
```diff
@@ -64,10 +66,18 @@ class WooPosTabController @Inject constructor(
...
         activity.lifecycleScope.launch {
-            val tabShouldBeVisible = shouldPosTabBeVisible()
-            setPOSTabVisibility(tabShouldBeVisible)
-            appPrefs.setPOSTabVisibilityForSite(selectedSite.getSelectedSiteId(), tabShouldBeVisible)
-            analyticsTracker.track(WooPosAnalyticsEvent.Event.TabVisibilityChecked(tabShouldBeVisible))
+            val result = shouldPosTabBeVisible()
+
+            result.onSuccess { tabShouldBeVisible ->
+                setPOSTabVisibility(tabShouldBeVisible)
+                if (tabShouldBeVisible) appPrefs.setPOSTabVisibilityForSite(selectedSite.getSelectedSiteId())
```
malinajirka: Related to the first comment - I think we need to clear the cache in case of `!tabShouldBeVisible`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
```diff
@@ -104,15 +104,16 @@ sealed class WooPosAnalyticsEvent : IAnalyticsEvent {
...
-        data class TabVisibilityChecked(val isVisible: Boolean) : Event() {
+        data class TabVisibilityChecked(val isVisible: Result<Boolean>) : Event() {
             override val name: String = "tab_visibility_checked"
 
             init {
-                addProperties(
-                    mapOf(
-                        "is_visible" to isVisible.toString()
-                    )
+                val value: String = isVisible.fold(
```
malinajirka, line 111: ❓ This look like a change to the tracking logic - is iOS handling errors the same way? If not, I think we should consider making the platforms consistent, wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/WooPOSIsRemotelyEnabledTest.kt
```diff
@@ -30,65 +31,67 @@ class WooPOSIsRemotelyEnabledTest {
...
-    fun `given empty feature list when invoked then returns false`() = runTest {
+    fun `given empty feature list, when invoked, then returns success false`() = runTest {
         val jsonSettings = """{"enabled_features": []}"""
         whenever(ssrModel.settings).thenReturn(jsonSettings)
         whenever(cacheResult.isError).thenReturn(false)
         whenever(cacheResult.model).thenReturn(ssrModel)
         whenever(fetcher.load(siteModel)).thenReturn(cacheResult)
 
-        val result = sut.invoke()
+        val r = sut.invoke()
```
malinajirka: 💡 Is this rename to `r` intentional in these tests? We usually call it `sut` or `result`.

malinajirka (changes requested):
Thanks for working on this @toupper! I've reviewed the code and left some questions/suggestions. I haven't tested the behavior yet. Let me know what you think and I'll test it when we resolve these discussions.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/WooPOSIsRemotelyEnabledTest.kt
toupper: Good point, changed in 46d5b690400b77f0b8d3ddd28a8cc7299cf91b39

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
toupper: Yeah, I'm also taking care of the IOS issue, I added a comment about [here](https://linear.app/a8c/issue/WOOMOB-1035/woo-pos-pos-tab-not-visible-when-network-not-available-during-app#comment-5739cc2c).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabController.kt
toupper: Indeed, done in 798c607502b54fb9db15fe13ba4428e18a5a6a1e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
toupper: I refactored it here 60680c898da34a294203c1e016e05d92d64e8376 to make it more clear. However, that can change depending on the other comments

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
toupper: Yeah that makes sense, I updated it 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
toupper: That makes sense, I added that too

WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
toupper: Same

toupper:
Thanks for your comments @malinajirka! I addressed them and refactored the code a bit to make it cleaner and easier to read. Please let me know what you think

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
```diff
@@ -39,51 +43,115 @@ class WooPosCanBeLaunchedInTab @Inject constructor(
...
             )
 
-        val wooCoreVersion = if (forceRefresh) {
-            fetchWooCoreVersion()
-        } else {
-            getWooCoreCachedVersion()
-        } ?: return WooPosLaunchability.NotLaunchable(
-            WooPosLaunchability.NonLaunchabilityReason.WooCommercePluginNotFound
-        )
+        val cachedPositive = appPrefs.isPOSLaunchableForSite(site.id)
```
malinajirka, line 49: 💡 I'm wondering if this should be set to false when `forceRefresh` is true, wdyt?

malinajirka (approved):
Thanks for all the changes! The tests are failing, other than that LGTM :shipit:. I've left question but feel free to ignore it.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt
```diff
@@ -41,41 +43,51 @@ class WooPosTabShouldBeVisibleTest : BaseUnitTest() {
...
-        assertTrue(sut())
+    fun `given feature flag enabled and supported country, when invoked, then return success true`() = testBlocking {
+        val r = sut()
+        assertTrue(r.isSuccess)
+        assertTrue(r.getOrThrow())
     }
 
     @Test
-    fun `given feature flag disabled, when invoked, then return false`() = testBlocking {
+    fun `given feature flag disabled, when invoked, then return success true (tab visible anyway)`() = testBlocking {
```
malinajirka: ❓ Is this test correct, why do we want to return tabVisible when remote FF is disabled?

toupper: Good catch, fixed in c609905e623b543dd5494491b3ef7270cab328c1

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosCanBeLaunchedInTab.kt
toupper: That's a good question. I was considering that, and I think it's ok to keep the logic simple and use the same rationale when `forceRefresh` is true. If the previous check gave false eligibility we refresh the cache, so it would give us false anyways

PR #14469, opened 2025-08-14
[Shipping Labels] Customs Form: autofill country code and improve validation logic

hichamboushaba:
Part of WOOMOB-878

> [!NOTE]
> Despite what looks like a big number of lines, a big part of the changes is unit tests, and also the new files contain just previous logic that was moved from the ViewModel. 

### Description
This PR is part of the improvements we need for the Customs Form, while the main task is just adding autofilling for the country code, I found out that to support the same user experience we support in web and iOS, we need to rework how we handle the customs form validation, we need to validate the form without depending on the Customs screen, and this is what this PR achieves.
To help with the review, here is a description of the main changes:
1. Introduce the usecase `CreateDefaultCustomsData` that replaces the previous extension we had before, this allows us to get the country name from the DB during the creation.
2. Introduce `WooShippingCustomsValidator`: This class encapsulates all the validation logic of the custom data, and as part of the change, the `WooShippingCustomsFormViewModel` was updated to use it instead of handling the validation itself.
3. Update `WooShippingLabelCreationViewModel` with the following changes:
    - Create a default `CustomsData` object when needed: during package split, and after updating the destination address.
    - Validate the `CustomsData` using the `WooShippingCustomsValidator`.

##### Valid default customs form: The user doesn't need to add any additional details
This is where the Android app lagged behind the web and iOS versions. When the user doesn't need to enter any additional details, in Android, we always showed a "Missing information" alert, and the user needed to open the form to trigger the validation.
1. Create an international order, with a destination country not part of the EU, and with a value less than 2500. Ensure the products define a valid weight.
2. Start shipping labels creation.
3. Confirm the customs form is shown as "Completed"
4. Open the customs form and confirm the origin country matches the origin address of the label.

##### Missing ITN
1. Create an international order, with a value higher than 2500.
2. Start shipping labels creation.
3. Confirm the customs form is shown as "Missing info", and that there is a notice "ITN is required"

##### Add missing information
1. Start from the previous test case.
2. Add the missing information.
3. Confirm the customs form is shown as "Missing info"

##### National order
1. Create a national order.
2. Start shipping labels creation.
3. Confirm the customs form is not shown.

### Testing information
- Confirm the above test cases work as expected.
- Create shipping labels after some of the test cases and confirm it works as expected.

The above.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

irfano (approved):
Nice improvement! The refactor was a bit hard to follow, and I couldn’t find any issues. I tested various cases, and everything worked as expected. LGTM! 👍🏻

PR #14472, opened 2025-08-14
[Log in] Parse error related to public-api.wordpress.com availability

samiuelson:
# WP API (public-api.wordpress.com) availability error handling 

WOOMOB-924

### Description

This PR adds parsing of the API error related to `*.wordpress.com` resources being unavailable, which can be related to DNS service blocking the WP domain.

### Steps to reproduce

In order to block `public-api.wordpress.com`, you can use various tools, like local DNS Service implementation or specialized apps, e.g. [Rethink DNS](https://play.google.com/store/apps/details?id=com.celzero.bravedns&hl=en).
Important part is to block access to `*.wordpress.com` domains.

Next, try to log-in to the app and notice the error message is not quite appropriate.

### Testing information

1. Block `*.wordpress.com` domain
2. Try logging in to the app
3. Verify the error message is appropriate

Above

### Images/gif

* Before
image

* After
image

image


malinajirka (approved):
LGTM, thanks!

PR #14473, opened 2025-08-15
Add GitHub warning to constraint layout dependency updates

malinajirka:
This PR adds a github action that should drop a comment/warning on PRs created by dependabot that update the constraintlayout-compose dependency. We found out that the new version breaks the POS, specifically the items in the cart start disappearing during screen transitions. 

The code was vibe coded and hasn't been tested. I've checked it and it looks fine.

I noticed it uses github-script@v7 while we use v4 in the other scripts - not sure if that's fine or not, but according to Claude v7 should be more secure.

Let me know what you think, thanks!

### DIFF
--- a/.github/workflows/dependabot-constraintlayout-check.yml
+++ b/.github/workflows/dependabot-constraintlayout-check.yml
@@ -1,0 +1,67 @@
+name: Dependabot POS Reminder
+
+on:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  add-pos-reminder:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+
+    steps:
+      # Check if PR contains the constraintlayout dependency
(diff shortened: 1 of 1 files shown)


ParaskP7 (approved):
👋 @malinajirka , thanks for this, I just reviewed, vibe-reviewed it too, and it all LGTM, it should work as expected, let's merge this and see, worse-case we quickly revert it! 🙇 ❤️ 🚀 

FYI: After merging this, we could (potentially) re-open this #14400 PR and see it working, right... 🤞

> I noticed it uses github-script@v7 while we use v4 in the other scripts - not sure if that's fine or not, but according to Claude v7 should be more secure.

👍

malinajirka:
> FYI: After merging this, we could (potentially) re-open this https://github.com/woocommerce/woocommerce-android/pull/14400 PR and see it working, right... 🤞

Great idea!

PR #14493, opened 2025-08-20
[Shipping Labels] Show refund notice about refunded labels in Order Details screen

hichamboushaba:
Closes WOOMOB-972

### Description
This PR adds a refund notice to the shipment details section in order details screen.

This PR required changing the logic of `GetShipments` to make sure we map the labels even when they are refunded.

##### Refund notice
Open an order that has a refunded label, or create a new one and refund it.

##### No regression
Test buying a new label from the app.

### Testing information
- Confirm the shipment details in the order details screen shows the refund notice.
- Confirm the `GetShipments` changes don't cause any regression.

The above.

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20250820_133544 | Screenshot_20250820_133555 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
@@ -68,13 +68,14 @@ class GetShipments @Inject constructor(
         shipmentUIModelList: List<ShipmentUIModel>,
         currentOrderLabels: List<ShippingLabelModel>
     ) = shipmentUIModelList.map { shipmentUIModel ->
-        val purchasedNonRefundedLabel = currentOrderLabels.find {
-            it.shipmentId == shipmentUIModel.remoteId && it.status == ShippingLabelStatus.PURCHASED && it.refund == null
+        val shipmentLabels = currentOrderLabels.filter {
+            it.shipmentId == shipmentUIModel.remoteId
+        }.sortedByDescending { it.createdDate?.time }
+
+        val purchasedLabel = shipmentLabels.find {
+            it.status == ShippingLabelStatus.PURCHASED && it.refund == null
         }
-        if (purchasedNonRefundedLabel == null) {
-            shipmentUIModel
-        } else {
-            shipmentUIModel.copy(purchased = true, label = purchasedNonRefundedLabel)
-        }
+
+        shipmentUIModel.copy(label = purchasedLabel ?: shipmentLabels.firstOrNull())
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
@@ -8,10 +8,17 @@ data class ShipmentUIModel(
     val localId: String,
     val remoteId: String? = null,
     val items: List<ShippableItemModel>,
-    val purchased: Boolean = false,
     val purchaseState: PurchaseState = PurchaseState.NoStarted,
     val label: ShippingLabelModel? = null,
-) : Parcelable
+) : Parcelable {
+    /**
+     * Whether the shipment has been purchased or not.
+     * A shipment is considered purchased if the label was already purchased or is in the process of being purchased.
+     */
+    val purchased: Boolean
+        get() = label?.status == ShippingLabelStatus.PURCHASE_IN_PROGRESS ||
+            (label?.status == ShippingLabelStatus.PURCHASED && label.refund == null)
+}
 
 @Parcelize
 sealed class PurchaseState : Parcelable {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailWooShippingShipmentListView.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailWooShippingShipmentListView.kt
@@ -1,6 +1,7 @@
 package com.woocommerce.android.ui.orders.details.views
 
 import androidx.annotation.StringRes
+import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
@@ -12,6 +13,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.LocationOn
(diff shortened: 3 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
```diff
@@ -68,13 +68,12 @@ class GetShipments @Inject constructor(
...
-        if (purchasedNonRefundedLabel == null) {
-            shipmentUIModel
-        } else {
-            shipmentUIModel.copy(purchased = true, label = purchasedNonRefundedLabel)
-        }
+        val purchasedLabels = currentOrderLabels.filter {
+            it.shipmentId == shipmentUIModel.remoteId && it.status == ShippingLabelStatus.PURCHASED
+        }.sortedByDescending { it.createdDate?.time }
+
+        val nonRefundedLabel = purchasedLabels.find { it.refund == null }
```
hichamboushaba, line 75: I'm not sure this part is really needed, given we sort the labels by createdDate above, I saw the iOS app uses it ([here](https://github.com/woocommerce/woocommerce-ios/blob/732cda3e60d87946c22e93c078ebf88ac6ce1431/Modules/Sources/Networking/Model/ShippingLabel/Shipments/WooShippingConfigResponse.swift#L66-L72)) and I copied it to be safe.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
```diff
@@ -68,13 +68,14 @@ class GetShipments @Inject constructor(
...
-            shipmentUIModel
-        } else {
-            shipmentUIModel.copy(purchased = true, label = purchasedNonRefundedLabel)
+        val shipmentLabels = currentOrderLabels.filter {
+            it.shipmentId == shipmentUIModel.remoteId
+        }.sortedByDescending { it.createdDate?.time }
+
+        val purchasedLabel = shipmentLabels.find {
+            it.status == ShippingLabelStatus.PURCHASED && it.refund == null
         }
```
hichamboushaba, line 77: I'm not sure this part is really needed, given we sort the labels by createdDate above, I saw the iOS app uses it ([here](https://github.com/woocommerce/woocommerce-ios/blob/732cda3e60d87946c22e93c078ebf88ac6ce1431/Modules/Sources/Networking/Model/ShippingLabel/Shipments/WooShippingConfigResponse.swift#L66-L72)) and I copied it to be safe.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
```diff
@@ -8,10 +8,17 @@ data class ShipmentUIModel(
...
-) : Parcelable
+) : Parcelable {
+    /**
+     * Whether the shipment has been purchased or not.
+     * A shipment is considered purchased if the label was already purchased or is in the process of being purchased.
+     */
+    val purchased: Boolean
+        get() = label?.status == ShippingLabelStatus.PURCHASE_IN_PROGRESS ||
+            (label?.status == ShippingLabelStatus.PURCHASED && label.refund == null)
+}
```
hichamboushaba, line 21: This change is not related to the PR in itslef, but having two sources of truth for the purchase status was confusing to me and a source of potential bugs, so I updated it when I updated `GetShipments`.

Please feel free to share any thoughts about this change.

hichamboushaba: For information, I'm planning to rename this to `isPurchasedOrInProgress` in my next PR, as it will better relay its role.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
JorgeMucientes: Hmm there's a couple things I don't fully understand from the new logic: 

- For `currentOrderLabels.filter {}` why are we using filter instead of find? Do we expect more than one label associated with the same shipment? When can this happen?
-   ~~`currentOrderLabels` already contains purchased labels right? Do we need the check `it.status == ShippingLabelStatus.PURCHASED` ? Or can we simply ensure the label hasn't been refunded?~~ Nevermind I see there's also the status `ShippingLabelStatus.PURCHASE_IN_PROGRESS` so I guess checking for only `PURCHASED` labels would make sense. 
- Would something like the following code work? 

```kotlin
    private fun mergePurchaseData(
        shipmentUIModelList: List<ShipmentUIModel>,
        currentOrderLabels: List<ShippingLabelModel>
    ) = shipmentUIModelList.map { shipmentUIModel ->
        val shipmentLabels = currentOrderLabels.find {
            it.shipmentId == shipmentUIModel.remoteId
        }
        
        shipmentUIModel.copy(label = shipmentLabels)
    }
```

Sorry if I'm talking nonsense I still feel there's details I'm missing here.

JorgeMucientes (approved):
Thanks for working on this @hichamboushaba. 
I left a couple questions to better understand the new logic as I'm confused about a couple things.

Also while testing this I noticed a new bug, that I believe is unrelated to the changes from this PR: 
Given an order with splited shipments and some labels already purchased.
1. Purchase a label
2. Navigate back to order details
3. The new label is show AND a new shipment is created out of the blue
4. Refund the label you've just purchased. The newly created shipment is gone: 

https://github.com/user-attachments/assets/15684f87-2ada-46dd-8dc7-9f26e6b6f807

Is this bug under your radar?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/GetShipments.kt
hichamboushaba: > For currentOrderLabels.filter {} why are we using filter instead of find? Do we expect more than one label associated with the same shipment? When can this happen?

Yes, there can be multiple labels attached to the same shipment, after a failure for example.

> Would something like the following code work?

Given the above, I don't think it would work, personally I think the following should work (given the sorting logic):

```Kotlin
        val shipmentLabel = currentOrderLabels.sortedByDescending { it.createdDate?.time }
             .find {  it.shipmentId == shipmentUIModel.remoteId }
       shipmentUIModel.copy(label = shipmentLabel)
```

But given iOS and also the web from what I saw in the code specifically check for the purchased label, I decided to keep the above part.

hichamboushaba:
> Also while testing this I noticed a new bug, that I believe is unrelated to the changes from this PR:
Given an order with splited shipments and some labels already purchased.

This seems like a backend bug, and reminds me of the issue where the plugin was creating additional shipments for refunds and failure, can you reproduce it always? And if yes, can you check the woocommerce-shipping plugin is up-to-date? 
I can't reproduce it on my test site.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
@JorgeMucientes I'm going to merge this, please check my comment above about the bug, I don't think this is related to this PR anyway.

PR #14494, opened 2025-08-20
[POS] Remove hacky and deprecated accessibility announcement calls to and align with Android a11y guidelines

toupper:
### Description

This PR removes our current **`AccessibilityAnnouncement`** helper and its usages in WooPOS Totals, as well as the deprecated function call in WooPOS Cart.

The previous solution was **hacky and buggy**:  
- It created a hidden/dummy `Text` composable with changing content just to trick TalkBack into re-announcing.  
- This dummy view was rendered into the UI tree even though it had no real purpose, making the approach fragile.  
- It also led to **crashes and recomposition issues** when the helper was called in certain states.  
- On top of that, it relied on a deprecated pattern (`announceForAccessibility`) and was not aligned with Android accessibility guidelines, which explicitly discourage synthetic announcements. (More [info](https://developer.android.com/reference/android/view/View?authuser=1#announceForAccessibility(java.lang.CharSequence)))

**What changed**
- Deleted `com.woocommerce.android.ui.woopos.common.composeui.component.AccessibilityAnnouncement`.
- Removed calls to this helper in:
  - `ui.woopos.home.totals` (payment success/failure/in-progress, reader status)
- Remove deprecated call in  `ui.woopos.home.cart` (cart item add/error announcements)
- No user-visible UI added; we’re cleaning up a non-compliant and crash-prone implementation.

**Why**
- The platform now expects apps to **expose meaningful state changes in the semantics tree** instead of firing arbitrary announcements. Examples:
  - Use `paneTitle` for **pane/screen** changes.
  - Use `stateDescription` (optionally with `liveRegion = Polite/Assertive`) for **dynamic state changes** (e.g., cart count).
  - Use component-level error semantics for **validation errors**.
- Removing this hack avoids rendering unnecessary dummy nodes, prevents crashes, and keeps us aligned with the Android a11y model.

**Follow-ups (separate PRs)**
I will create a Linear ticket once this PR is approved. The suggested approach is:

- Cart:
  - Expose the **cart count** label with `stateDescription = "N items in cart"` and mark it `liveRegion = Polite` so count changes are announced naturally.
  - (Optional) Use a **Snackbar** (`"$product added to cart"`) for explicit feedback; TalkBack announces Snackbars automatically.
- Totals:
  - Wrap state containers with `Modifier.semantics { paneTitle = "<Screen title>" }` so navigation/state transitions are announced as **pane changes**.
  - Keep error surfaces visible and rely on component error semantics.

**Known trade-off (temporary)**
- Users won’t hear an “action confirmation” message until we add compliant semantics. They can still review the state via UI (e.g., cart count).

### Steps to reproduce

_Pre-PR (current trunk):_
1. Enable TalkBack.
2. Add a product to the cart or trigger Totals state changes.
3. A synthetic announcement is forced via the dummy `Text`; may fail to repeat, may cause crashes, and always renders an unnecessary hidden node.

_Post-PR:_
1. Enable TalkBack.
2. Perform the same actions.
3. No synthetic “Added to cart / Payment …” messages are forced. Announcements will be re-introduced in a **guideline-compliant** way in follow-up PRs.

### Testing information

- Built and navigated Cart and Totals with TalkBack ON/OFF; verified no runtime errors after removing the helper and imports.
- Added/removed items; applied an invalid coupon; walked through Totals states to confirm no regressions aside from removed announcements.
- Verified that no dummy `Text` nodes are rendered in the layout inspector.

### Images/gif

N/A (no visual changes)


kidinov (approved):
LGTM!

I am just thinking maybe not merge it before we offer good replacement, otherwise merchants who'll use this version won't have notification about payment statuses changes or wdyt?

toupper:
> I am just thinking maybe not merge it before we offer good replacement, otherwise merchants who'll use this version won't have notification about payment statuses changes or wdyt?

Thanks for the review @kidinov! The problem is that with the current approach, it _randomly_ renders a Text in the screen, which is obviously not acceptable. The other solution I was thinking of is to the revert to the original approach, but that is deprecated. With that in mind, given that we are still in an _early_ stage of POS with not so many users, I vote for removing the announcements altogether and approaching it from scratch following Android guidelines.

PR #14496, opened 2025-08-21
Issue/woomob 1013 we mention manual refund on woopayments ipp payments

JorgeMucientes:
Fixes WOOMOB-1013

### Description
Fixes 2 issues when handling order refunds: 
- Instead of showing "Manual Refund" texts while we load card details for refund we'll show skeletons. 
- Show new text "Credit card" when refund is automatically made to the credit card used when paying the order. This is consistent with iOS behaviour. 

I'll give some more context on why I opted for this 2 options

### Testing information
1. Open a full filled order paid with a credit card
2. Tap on issue refund
3. Select a product to refund and tap Next
4. Check skeletons are shown while card details are fetched. 
5. Check card details are shown and new text "Credit card" is prefixed. 

https://github.com/user-attachments/assets/c4192a19-c771-4425-8e1f-11da0e7e00f6

The above

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
@@ -91,6 +91,7 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
                         )
                     findNavController().navigateSafely(action)
                 }
+
                 is NavigateToCardReaderScreen -> {
                     val action = RefundSummaryFragmentDirections.actionRefundSummaryFragmentToCardReaderFlow(
                         cardReaderFlowParam = CardReaderFlowParam.PaymentOrRefund.Refund(
@@ -100,6 +101,7 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
                     )
                     findNavController().navigateSafely(action)
                 }
+
                 else -> event.isHandled = false
             }
         }
@@ -129,6 +131,15 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
                     binding.refundSummaryMethodDescription.show()
                 } else {
                     binding.refundSummaryMethodDescription.hide()
+                }
+            }
+            new.isFetchingCardData.takeIfNotEqualTo(old?.isFetchingCardData) { loading ->
+                if (loading) {
+                    binding.refundMethodLayout.hide()
+                    binding.loadingPaymentMethodSkeletons.show()
+                } else {
+                    binding.refundMethodLayout.show()
+                    binding.loadingPaymentMethodSkeletons.hide()
                 }
             }
         }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
@@ -56,7 +56,8 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
         _binding = FragmentRefundSummaryBinding.bind(view)
         setupToolbar()
         initializeViews()
-        setupObservers()
+        setupUiStateObservers()
+        handleEvents()
         handleResults()
     }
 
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
```diff
@@ -131,6 +104,47 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
...
+                    binding.loadingPaymentMethodSkeletons.show()
+                } else {
+                    binding.refundMethodLayout.show()
+                    binding.loadingPaymentMethodSkeletons.hide()
+                }
+            }
+        }
+    }
+
+    private fun handleEvents() {
```
JorgeMucientes, line 119: Had to extract this code into a new function to fix Detect issues. Nothing changed here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryFragment.kt
```diff
@@ -131,6 +104,47 @@ class RefundSummaryFragment : BaseFragment(R.layout.fragment_refund_summary), Ba
...
                 }
             }
+            new.isFetchingCardData.takeIfNotEqualTo(old?.isFetchingCardData) { loading ->
+                if (loading) {
+                    binding.refundMethodLayout.hide()
+                    binding.loadingPaymentMethodSkeletons.show()
+                } else {
+                    binding.refundMethodLayout.show()
+                    binding.loadingPaymentMethodSkeletons.hide()
+                }
```
JorgeMucientes, line 114: These are the only new changes added in this file. 

I opted for the skeleton solution because gateway details don't provide any relevant text we can show while fetching card details. 

Screenshot 2025-08-21 at 10 49 24

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt
```diff
@@ -205,7 +205,9 @@ class RefundSummaryViewModel @Inject constructor(
                     val refundMethodWithCard = result.run {
                         val brand = result.cardBrand.orEmpty().replaceFirstChar { it.uppercase() }
                         val last4 = result.cardLast4.orEmpty()
-                        "$refundMethod ($brand **** $last4)"
+                        val creditCardRefundDefaultText =
+                            resourceProvider.getString(R.string.order_refunds_credit_card_refund)
+                        "${refundMethod.ifBlank { creditCardRefundDefaultText }} ($brand **** $last4)"
```
JorgeMucientes, line 210: Using the alternative value coming from `method_title` as fallback didn't feel safe to me. For example, for most of the tests I've ran, the value for `method_title: WooPayments` wouldn't make much sense in the screen UI. Given that at this point in the code we know we are managing a refund to a credit card, I felt the "Credit card" text would work better (this is also consistent with what iOS does)

Screenshot 2025-08-21 at 11 02 57

hichamboushaba: > For example, for most of the tests I've ran, the value for method_title: WooPayments wouldn't make much sense in the screen UI

Can you explain why? To me, saying that the refund will be handled via WooPayments still makes sense, as it's the actual payment gateway.

If we don't set a default value here, we might show an empty text there if we end up on line 193 https://github.com/woocommerce/woocommerce-android/blob/95540988bd7afb8deec1019761a834f3f72cda2a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt#L193

I'm not sure in which conditions the `chargeId` would be null, but given the `if`, I think it's a possibility, and in this case, to me saying `Refund via WooPayments` would be better than an empty value, I think we can have a middle ground here, we use the default value only for line 193, but not for the other path, WDYT?

hichamboushaba (approved):
Nice work @JorgeMucientes, I left one comment about using `method_title` as default value, please check it out, but to me it's not a blocker.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt
JorgeMucientes: > Can you explain why? To me, saying that the refund will be handled via WooPayments still makes sense, as it's the actual payment gateway.

Fair point. My main concern is not knowing what other possible values can be set for `method_title` field. Additionally it felt semantically more intuitive to prepend "Credit card" before the credit card details, instead of `WooPayments`. But that's just my opinion not really a proper argument. 

Given your points I'll set `gateway.methodTitle` as general fallback for both cases in inside `enrichRefundMethodWithCardDetails`. And I'll leave the "Credit card" only to be displayed when both `title` and `methodTitle` are empty. [50cc97b](https://github.com/woocommerce/woocommerce-android/pull/14496/commits/50cc97b0e156bbd2230a3e3450058e6abb853a6a)

hichamboushaba: I think I didn't relay my idea clearly, sorry. In my last paragraph, I meant to use the fallback only in the line 193 (when `chargeId` is null), which means we'll use `Credit card` when we know it's a credit card, and we'll use `method_title` for the other cases, to make it clearer, I mean the following patch, let me know what you think.

```patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt	(revision 50cc97b0e156bbd2230a3e3450058e6abb853a6a)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt	(date 1755803715633)
@@ -180,7 +180,7 @@
                 }
                 updateRefundSummaryState(paymentTitle, isMethodDescriptionVisible = true)
             } else {
-                enrichRefundMethodWithCardDetails(gateway.title.ifBlank { gateway.methodTitle })
+                enrichRefundMethodWithCardDetails(gateway.title)
             }
         }
     }
@@ -190,7 +190,7 @@
         if (chargeId != null) {
             loadCardDetails(chargeId, refundMethod)
         } else {
-            updateRefundSummaryState(refundMethod, isMethodDescriptionVisible = false)
+            updateRefundSummaryState(refundMethod.ifBlank { gateway.methodTitle }, isMethodDescriptionVisible = false)
         }
     }
 ```

hichamboushaba: Given it's too late today, I'll re-enable auto-merging (I disabled it temporarly), and if you agree with my opinion above, we can add it in a separate PR, as regardless of which approach we take, the change of this PR is huge improvement, and the sooner we release it, the better.

PR #14498, opened 2025-08-21
[Shipping Labels] Update the purchase in-progress and failure screens

hichamboushaba:
Closes WOOMOB-873

### Description
This PR introduces updated UI for the purchase in-progress and failure screens, and below is a detailed list of the added changes:
1. Update the purchase in-progress UI to match the web counterpart closely.
2. Update the purchase failure UI to match the web counterpart closely.
3. Allow purchasing a new label after an async failure. This is a **bug** that we had, as the failure was also treated like success, with the only difference being the use of different text, but the whole UI was read-only, and there was no way to retry.
4. Don't show the green check mark on the shipment tab when the purchase is in progress.
5. Disable splitting shipments when there is a purchase in progress (to match the web behavior).

> [!NOTE]
>  Some notes about the changes:
> - The PR appears larger than it's, because GitHub is marking the file `ShippingLabelPurchaseStatusSection` as a new file, but it's just a renaming of the old file `PrintShippingLabelSection`, and the file has `423` lines.
> - I included some refactorings that caused the PR to touch more places, it would be better to review them separately:
>     - The commit 61dcbc1dd41df089befb7623922e8df9f8fcba19 updated the logic to introduce `ShipmentUI#isReadOnly`, and a separate `isPurchased`, this was needed as previously in progress was treated like success, while it should be just a flag for read-only form.
>     - b3ee1f41d627f29a2728ec5d4a85ccb68ec92d26 I renamed the `ShipmentUIModel#purchased` to `ShipmentUIModel#isPurchasedOrInProgress` to better clarify its role.
> 

> [!IMPORTANT]
> This PR will be followed by a second PR that:
> - Updates the tracking logic to match iOS, I'll explain more on the PR.
> - Fix an issue with in-progress labels. 
> 
> Even though the issues above have been here before this PR, it would be better not to merge this until the second PR is done too.

##### Success
1. Start creating a new shipping label.
2. Check the in-progress UI and the success UI.

##### Failure
1. To trigger a failure, use the tip that was suggested here p1750956225617989/1750246257.598949-slack-C05VBLKHHV1 (ping me if you need more details)
2. Buy a new shipping a label.
3. Check the in-progress and failure UI.

### Testing information
- Confirm the updated UI looks as expected.
- Confirm that the split button is hidden when there is a purchase in-progress.
- Confirm that the UI allows you to re-purchase a new label after failure.

The above.

### Images/gif
|  | Light | Dark |
| --- | --- | --- |
| **Purchase then success** | <video src=https://github.com/user-attachments/assets/d2e477b5-21ef-4446-9dda-3cbf90f7b9fb /> | <video src=https://github.com/user-attachments/assets/f1b70eef-6d10-4e94-a86a-ac0d0f05c982 />
| **Purchsae then failure** | <video src=https://github.com/user-attachments/assets/d52ad0bc-c84b-40ac-b5af-5869e3a28bfe /> | <video src=https://github.com/user-attachments/assets/bd2e8abc-079c-46fe-9084-c4944ce72548 />


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
```diff
@@ -8,22 +8,10 @@ data class ShipmentUIModel(
     val localId: String,
     val remoteId: String? = null,
     val items: List<ShippableItemModel>,
-    val purchaseState: PurchaseState = PurchaseState.NoStarted,
+    val isPurchaseAPILoading: Boolean = false,
```
hichamboushaba, line 11: The `PurchaseState` was confusing, as we already have `ShippingLabelStatus` that reflects purchase status.
The confusing part here is that the Shipping Label purchase happens in two steps:
- We make an API request to the backend to trigger the purchase, then the purchase will continue async.
- We observe the purchase status which is represented by `ShippingLabelStatus`.

`PurchaseState` was only used to show the loading UI for the first step, but its other statuses weren't used at all, so I opted to remove it and replace it with the new Boolean `isPurchaseAPILoading`, I hope the mention of `API` there makes this less confusing for future code readers.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
```diff
@@ -8,22 +8,10 @@ data class ShipmentUIModel(
...
     val label: ShippingLabelModel? = null,
 ) : Parcelable {
-    /**
-     * Whether the shipment has been purchased or not.
-     * A shipment is considered purchased if the label was already purchased or is in the process of being purchased.
-     */
-    val purchased: Boolean
+    val isPurchasedOrInProgress: Boolean
```
hichamboushaba, line 14: As mentioned in the PR description, this change is not related to the PR itself, but an update to clarify the property role.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShippingLabelModel.kt
```diff
@@ -39,7 +39,8 @@ data class ShippingLabelModel(
     val refund: Refund?,
     val products: List<Order.Item> = emptyList(),
     val originAddress: Address? = null,
-    val destinationAddress: Address? = null
+    val destinationAddress: Address? = null,
+    val error: String? = null,
```
hichamboushaba, line 43: We don't cache this property, as it's used only after a purchase failure.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -704,7 +702,7 @@ class WooShippingLabelCreationViewModel @Inject constructor(
             hazmatStatesFlow
         ) { accountSettings, order, shipments, addresses, shippingRatesList,
             packageSelections, uiState, customsState, hazmatStates ->
-            if (accountSettings == null || shipments.any { it.purchaseState is PurchaseState.Error }) {
```
hichamboushaba, line 707: The condition `shipments.any { it.purchaseState is PurchaseState.Error }` wasn't used at all, as we never set the `purchaseState` to the `Error` state.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -704,7 +702,7 @@ class WooShippingLabelCreationViewModel @Inject constructor(
...
+            if (accountSettings == null) {
```
hichamboushaba, line 712: The condition shipments.any { it.purchaseState is PurchaseState.Error } wasn't used at all, as we never set the purchaseState to the Error state.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -969,12 +967,21 @@ class WooShippingLabelCreationViewModel @Inject constructor(
...
+                    updateShipment(
+                        selectedShipmentIndex,
+                        shipments.value[selectedShipmentIndex].copy(
+                            isPurchaseAPILoading = false,
+                            label = it.labels.firstOrNull()
+                        )
+                    )
+                    observeShippingLabelPurchaseStatus(selectedShipmentIndex)
+                    // TODO check if we need to track this here or in the observeShippingLabelPurchaseStatus method
+                    analyticsTracker.track(AnalyticsEvent.WCS_PURCHASE_STEP, mapOf(KEY_STATE to "purchase_success"))
```
hichamboushaba, line 986: I removed `handlePurchaseSuccess` and moved its logic here, I did this because the `Success` part of the function was a source of confusion, as this is just an API success, and the label is yet to be purchased.

I'll revisit this in my next PR to see if we can avoid the `LongMethod` detekt issue.

hichamboushaba, line 986: This is the tracking issue that I mentioned in the PR description, we track the label as a success before it's actually purchased, this is different than iOS.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/models/ShipmentUIModel.kt
irfano: Yes, I agree, this simplification makes it much better.

irfano (approved):
Great job! The changes look good. I also tested the API error and status code 500 error, all cases work as expected. LGTM! 👍🏻

PR #14501, opened 2025-08-21
[Shipping Labels] Update tracking and handle in-progress purchase of existing shipments 

hichamboushaba:
### Description
This PR fixes two issues:
1. It updates the shipping label purchase tracking to be consistent with iOS:
    - Track success only when the async purchase finishes successfully.
    - Track a failure when the async purchase fails.
2. Handle the case of existing shipments with in-progress purchase: The async purchase means that the user can leave the app or wp-admin while a purchase is still ongoing, if they do this, the label will keep the `PURCHASE_IN_PROGRESS` status. We weren't handling this case; we had two different issues:
    - Before this [PR](https://github.com/woocommerce/woocommerce-android/pull/14493), the in-progress label was dismissed, and this was a major issue I think, as they user could be charged multiple times.
    - After the above PR, the app now shows correctly the in-progress label, but it will stay like this indefinitely, unless the user opens wp-admin, which will update the status correctly.

#### Tracking success
1. Purchase a label from the app.
2. Check logcat and confirm the following line `Tracked: woocommerceandroid_wcs_purchase_step, Properties: {"state":"purchase_success"...}` is tracked only after the label is completely purchased.

#### Optional: tracking failure
1. Check the testing steps to trigger a failure in this [PR](https://github.com/woocommerce/woocommerce-android/pull/14498) description, or use ApiFaker.
2. Check logcat and confirm the following line `Tracked: woocommerceandroid_wcs_purchase_step, Properties: {"state":"purchase_failed","error":..."...}` is tracked.

#### In-progress purchase from mobile
1. Start purchase of a shipping label in the app.
2. Leave the screen while the purchase is in progress.
3. Re-open the shipping labels form.
4. Confirm the purchase continues.

#### In-progress purchase from web
1. Start purchase of a shipping label in wp-admin.
2. Close the browser tab while the purchase is in progress.
3. Open the shipping labels form in the mobile app.
4. Confirm the purchase continues.

### Testing information
- Confirm the tracking works as expected.
- Confirm the in-progress labels are handled correctly.

The above.

### Images/gif
| Before #14493 | After #14493 | After this PR |
| --- | --- | --- |
| <video src=https://github.com/user-attachments/assets/b063a49d-f20f-454c-a98d-5e6d64bad067 /> | <video src=https://github.com/user-attachments/assets/cec05470-5085-4fb6-86ca-ba23daa5a4c6 /> | <video src=https://github.com/user-attachments/assets/b030750a-3b08-43b3-9687-bb2ddf29ae5a /> |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
@@ -955,7 +955,7 @@ class WooShippingLabelCreationViewModel @Inject constructor(
         launch { refreshShippingRates.emit(Unit) }
     }
 
-    @Suppress("ComplexCondition", "LongMethod")
+    @Suppress("ComplexCondition")
     fun onPurchaseShippingLabel() {
         val selectedShipmentIndex = selectedShipmentIndex
         val selectedPackage = selectedPackagesFlow.value[selectedShipmentIndex]
@@ -994,13 +994,11 @@ class WooShippingLabelCreationViewModel @Inject constructor(
                 hazmatSelection
             ).fold(
                 onSuccess = {
-                    updateShipment(
-                        selectedShipmentIndex,
-                        shipments.value[selectedShipmentIndex].copy(
-                            isPurchaseAPILoading = false,
-                            label = it.labels.firstOrNull()
-                        )
+                    val updatedShipment = shipments.value[selectedShipmentIndex].copy(
+                        isPurchaseAPILoading = false,
+                        label = it.labels.firstOrNull()
                     )
+                    updateShipment(selectedShipmentIndex, updatedShipment)
                 },
                 onFailure = { exception ->
                     updateShipment(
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -269,16 +268,38 @@ class WooShippingLabelCreationViewModel @Inject constructor(
...
-            val labelId = shipment.label?.labelId ?: return@launch
-            observeShippingLabelStatus(orderId = navArgs.orderId, labelId = labelId).onEach { result ->
-                val originAddress = shippingAddresses.value.getOrNull(shipmentId)?.shipFrom?.toAddress()
+    private suspend fun observeShippingLabelAsyncPurchaseStatus() {
+        val shipmentWithIndex = uiState.map { it.selectedIndex }.distinctUntilChanged()
+            .flatMapLatest { index ->
+                shipments.map { it.getOrNull(index) }
+                    .filterNotNull()
+                    .map { Pair(it, index) }
+            }
```
hichamboushaba, line 278: While using `combine` could work here too, IMO and to follow the reactive streams best practices, what I used here is better, and I'll try to explain better here: `combine(index, shipments)` woule emit whenever one of the two flows change, even if the thipment of the current index doesn't change, so changing the responsibilities a bit to what I used here means we'll observe only the selected shipment, which is safer.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

irfano (changes requested):
Great job on identifying the issue and fixing it. It works as expected. and LGTM! 👍🏻 
_Edit: I requested changes by mistake._

irfano (approved):

PR #14507, opened 2025-08-22
Show disabled button state when missing required fields

JorgeMucientes:
Closes: WOOMOB-940

### Description

The current support form is a bit confusing because the "Submit Support Request" is shown as enabled even when it is disabled behind scenes because there's missing required info in the form. 
This PR simply adds a disabled state to the button so that it reflects it's current enabled/disabled status clearly

### Testing information
1. Navigate to Settings > Help & Support > Contact Support
2. Verify the button is shown as disabled in both light and dark mode
3. Enter all the required field
4. Check how the button is shown as enabled
5. Erase any of the required fields and confirm the button is shown as disabled again. 

The above

https://github.com/user-attachments/assets/42e2cc53-95ff-4935-8dd1-3248bca63d5d


irfano (approved):
LGTM! 👍🏻

PR #14511, opened 2025-08-22
[POS][Local Catalog] Repo for incremental sync timestamps

samiuelson:
# POS Incremental Sync Timestamp Storage

#####  Summary

Implemented a repository system for storing and retrieving site-specific incremental sync timestamps for POS products and variations using DataStore preferences.

#####   Changes

  - `WooPosSyncTimestampRepository`: Low-level repository for string timestamp storage with site-specific keys
  - `WooPosSyncTimestampManager`: High-level manager providing Date object API with GMT timezone handling

#####   Key Features
  - Site-specific storage: Timestamps are isolated per site using dynamic key generation
  - Type-safe API: Date objects for convenience, strings for raw access
  - Error handling: Graceful null site handling with logging

### Testing information

This code is not used at the moment. Check if CI is green and if the code looks good.

Above 

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
@@ -14,6 +14,7 @@ class WooPosSyncTimestampManager @Inject constructor(
 ) {
     private val gmtDateFormat = SimpleDateFormat(GMT_DATE_FORMAT, Locale.US).apply {
         timeZone = TimeZone.getTimeZone("GMT")
+        isLenient = false
     }
 
     suspend fun storeProductsLastSyncTimestamp(timestamp: Date) {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
@@ -1,0 +1,69 @@
+package com.woocommerce.android.ui.woopos.util.datastore
+
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class WooPosSyncTimestampManager @Inject constructor(
+    private val timestampRepository: WooPosSyncTimestampRepository
+) {
+    private val gmtDateFormat = SimpleDateFormat(GMT_DATE_FORMAT, Locale.US).apply {
+        timeZone = TimeZone.getTimeZone("GMT")
+    }
+
+    suspend fun storeProductsLastSyncTimestamp(timestamp: Date) {
+        val timestampGmt: String = gmtDateFormat.format(timestamp)
+        timestampRepository.storeProductsLastSyncTimestamp(timestampGmt)
+    }
+
+    suspend fun getProductsLastSyncTimestamp(): Date? {
+        val timestampString = timestampRepository.getProductsLastSyncTimestamp()
+        return timestampString?.let { parseGmtTimestamp(it) }
+    }
+
+    suspend fun clearProductsLastSyncTimestamp() {
+        timestampRepository.clearProductsLastSyncTimestamp()
+    }
+
+    suspend fun storeVariationsLastSyncTimestamp(timestamp: Date) {
+        val timestampGmt: String = gmtDateFormat.format(timestamp)
+        timestampRepository.storeVariationsLastSyncTimestamp(timestampGmt)
+    }
+
+    suspend fun getVariationsLastSyncTimestamp(): Date? {
+        val timestampString = timestampRepository.getVariationsLastSyncTimestamp()
+        return timestampString?.let { parseGmtTimestamp(it) }
+    }
+
+    suspend fun clearVariationsLastSyncTimestamp() {
+        timestampRepository.clearVariationsLastSyncTimestamp()
+    }
+
+    suspend fun clearAllSyncTimestamps() {
+        timestampRepository.clearAllSyncTimestamps()
+    }
+
+    fun formatTimestampForApi(timestamp: Date): String {
+        return gmtDateFormat.format(timestamp)
+    }
+
+    fun parseTimestampFromApi(timestampString: String): Date? {
+        return parseGmtTimestamp(timestampString)
+    }
+
+    private fun parseGmtTimestamp(timestampString: String): Date? {
+        return try {
+            gmtDateFormat.parse(timestampString)
+        } catch (e: Exception) {
+            null
+        }
+    }
+
+    private companion object {
+        const val GMT_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+    }
+}
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
@@ -1,5 +1,7 @@
 package com.woocommerce.android.ui.woopos.util.datastore
 
+import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
+import java.text.ParseException
 import java.text.SimpleDateFormat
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    suspend fun clearAllSyncTimestamps() {
+        dataStore.edit { preferences ->
+            preferences.remove(productsTimestampKey)
+            preferences.remove(variationsTimestampKey)
+        }
+    }
+
+    private fun buildSiteSpecificKey(key: String): Preferences.Key<String> =
+        stringPreferencesKey("${selectedSite.getOrNull()?.siteId}-$key")
```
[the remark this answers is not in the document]
samiuelson: Improved null site handling: be5980d1709941e62097330c951b40656f4334cd

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    fun parseTimestampFromApi(timestampString: String): Date? {
+        return parseGmtTimestamp(timestampString)
+    }
+
+    private fun parseGmtTimestamp(timestampString: String): Date? {
+        return try {
+            gmtDateFormat.parse(timestampString)
+        } catch (e: Exception) {
```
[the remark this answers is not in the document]
samiuelson: Improved exception handling: 35b6cd1e9dd7a08c4aee32eb678078189eb42bfa

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    suspend fun clearAllSyncTimestamps() {
+        dataStore.edit { preferences ->
+            preferences.remove(productsTimestampKey)
+            preferences.remove(variationsTimestampKey)
+        }
+    }
+
```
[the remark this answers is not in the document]
samiuelson: I think this is false positive (the process is restarted after changing the site). I'm however going to remove singleton annotation because it does not bring any improvement: 1ce3f7a19e6bd31f2fb8989ef021f6bb7f27af15

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
```diff
@@ -0,0 +1,72 @@
...
+    private val logger: WooPosLogWrapper
+) {
+    private val gmtDateFormat = SimpleDateFormat(GMT_DATE_FORMAT, Locale.US).apply {
+        timeZone = TimeZone.getTimeZone("GMT")
+        isLenient = false
+    }
+
+    suspend fun storeProductsLastSyncTimestamp(timestamp: Date) {
+        val timestampGmt: String = gmtDateFormat.format(timestamp)
+        timestampRepository.storeProductsLastSyncTimestamp(timestampGmt)
```
malinajirka: 💡 I think storing formatted dates has some drawbacks, the main one is that we can't safely change the format without clearing the storage. I'd suggest storing just timestamps - milliseconds since 1970. 

Considering Samuel is AFK, @kidinov could you double check my train of thoughts and share whether you agree with this suggestion before I make the change? Thanks!

kidinov: Agree. Also timestamp just feels cleaner as we don't need to format anything, And btw, `gmtDateFormat` is not thread safe

Not related to this, but I just realized that even if the date is retrieved from the backend, it’s not something we control. Therefore, there’s no guarantee that the date will be accurate and can change in some cases

malinajirka:
@kidinov I have updated the API to store timestamps (Long) instead of dates. I also replaced the thread-unsafe DateFormatter with the new java.time API. Since I made quite a few changes and Sam is AFK, would you mind taking over of the review of this PR?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun parseTimestampFromApi(dateFromApi: String): Long? {
+        return parseGmtTimestamp(dateFromApi)
+    }
+
+    private fun parseGmtTimestamp(dateFromApi: String): Long? {
+        return try {
+            val localDateTime = LocalDateTime.parse(dateFromApi, gmtFormatter)
+            return localDateTime.toInstant(ZoneOffset.UTC).toEpochMilli()
+        } catch (e: DateTimeParseException) {
+            logger.e("Failed to parse GMT timestamp: '$dateFromApi'", e)
```
kidinov, line 52: If the date format returned by the API doesn’t match our expectations, we return null. What happens next in the expected flow then?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    fun parseTimestampFromApi(dateFromApi: String): Long? {
+        return parseGmtTimestamp(dateFromApi)
+    }
+
+    private fun parseGmtTimestamp(dateFromApi: String): Long? {
+        return try {
+            val localDateTime = LocalDateTime.parse(dateFromApi, gmtFormatter)
+            return localDateTime.toInstant(ZoneOffset.UTC).toEpochMilli()
```
kidinov, line 50: The formatter is already in UTC so I guess we can just 

```
    private fun parseGmtTimestamp(dateFromApi: String): Long? {
        return try {
            Instant.from(gmtFormatter.parse(dateFromApi)).toEpochMilli()
        } catch (e: DateTimeParseException) {
            logger.e("Failed to parse GMT timestamp: '$dateFromApi'", e)
            null
        }
    }
```
?

kidinov (approved):
LGTM! I left a couple of questions

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
malinajirka: I was thinking we'd fall back to local timestamp instead of using the remote one. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
malinajirka: I find this quite weird as well, but from what I can tell + what AI can tell, we need to specify the timezone here again.  There is a chance it's not needed, but likely not worth spending time on verifying this and testing it. Since the method explicitly says `gmt` I think it's fine to explicitly use UTC even if it might not be necessary.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
kidinov: I suppose that would work if the API returns a timestamp, but please correct me if I’m mistaken - we expect them to return a formatted date with timezone and so on. So, if there’s something wrong in parsing this date, and we fall back to the local timestamp, we end up with a problem: those times could be in different time zones?

malinajirka: This is handled by the API since it provides `lastModified` and `lastModifiedGmt` dates, so as long as we shift our local time to GMT/UTC and use the `lastModifiedGmt` date, we should be just fine. Of course, if the local time is incorrect, then it won't work as expected until the full catalog sync, but we could potentially detect that. Moreover, chances the parsing fails and we need to fallback to local time are quite low I believe.

PR #14513, opened 2025-08-25
Propagate Headers to RestClients

malinajirka:
Considering this change impacts all endpoints implemented in the app, I believe it could benefit from two reviewers, thanks!

**Draft PR - I need to update the tests.**

### Description

This PR ensures headers from network responses are propagated to RestClients and can be used. It doesn't change any other behavior and doesn't actual use the headers in any of the responses, it just builds the infrastructure to propagate them to the RestClient layer.

This change is needed for the POS Local Catalog project as the headers will drive some of the logic - e.g. determining when we should do a full sync vs an incremental sync, based on the total number of changed items.

The main idea is implemented in the GsonRequest:
- We pass `ResponseWithHeaders<T>` instead of just `T` to the parent BaseRequest => `parseNetworkResponse(..)` parses the headers and passes them back to the parent which delivers the result through `deliverResponse()` where we put the headers into the actual response instance depending on the network type (Jetpack/WPCom/..).

### Testing information

This change impacts all rest clients and all endpoints. However, I believe green CI on it's own gives us pretty good confidence. Having said that, I think we should test the different connection types
- .com
- self-hosted .org api
- self-hosted rest api
- jetpack

P.S. I believe the biggest risk is nullability handling.

I've smoke tested the app on a few different sites and everything seemed to work just fine.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/plans/networking/SitePlanRestClient.kt
```diff
@@ -61,7 +61,8 @@ class SitePlanRestClient @Inject constructor(
                                         type = determineSitePlanType(id)
                                     )
                                 }
-                            }
+                            },
+                        headers = originalResponse.headers
```
hichamboushaba, line 65: @malinajirka not a review yet, but just wanted to share something here, there are cases where we can have the same header [defined](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi) multiple times (for example the `Link` header can be defined multiple times, and it was causing an issue until we updated the [implementation](https://github.com/woocommerce/woocommerce-android/blob/9792a91b9c78cee612bc29b99821abc5ec7a362a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/discovery/WPAPIHeadRequest.kt#L23-L29) to check all headers).

Volley's `headers` field will [contain](https://github.com/google/volley/blob/ef8ac3f434e18b1f7137f58c43a62cc701652ed5/core/src/main/java/com/android/volley/NetworkResponse.java#L178-L180) only the last value, so to support all cases, I think we have two choices:
1. Expose the headers as a `List`, like what Volley does with the `allHeaders` property.
2. Expose the headers as a `Map<String, List<String>>`.

malinajirka: Ohh, thanks for sharing! I didn't realize headers can be duplicated. I'll expose a list instead of Map then.

malinajirka: Implemented in https://github.com/woocommerce/woocommerce-android/pull/14513/commits/31f63af3ec1fb579ce3c90a27e42be0fd615ab28

kidinov (approved):
LGTM

I tested JP and direct connection with printing headers in the rest client, and the headers are present

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/WPAPIResponse.kt
```diff
@@ -1,8 +1,11 @@
 package org.wordpress.android.fluxc.network.rest.wpapi
 
+import com.android.volley.Header
+
 sealed class WPAPIResponse<T> {
     data class Success<T>(
         val data: T?,
+        val headers: List<Header>,
```
hichamboushaba, line 8: np, I think we could benefit in the future from reducing dependency on Volley on upper layers, so it would be good to not leak Volley's components here, WDYT about creating a simple `data class Header(val key: String, val value: String)` instead of using Volley's one?

hichamboushaba (approved):
Nice work @malinajirka. I didn't test the changes as Andrey already did, but the code looks good. I left one np comment for you to check, but it's not a blocker.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/WPAPIResponse.kt
malinajirka: Good point, updated in 0b27a820c9cf89574079a4afc91129182c2e5eaa.

malinajirka:
Thank you both for the review 🙇 !

PR #14521, opened 2025-08-26
[WOOMOB-1175][Woo POS][Settings] Update empty cart state UI with integrated barcode scanning option

kidinov:
WOOMOB-1175

### Description
Updated the empty cart state UI in the POS to provide a cleaner, more integrated design. The barcode scanning option is now part of the main message as a clickable link instead of a separate button.

~~Do not merge as I am still waiting confirmation on the design~~

### Steps to reproduce
1. Open the POS application
2. Navigate to the cart screen when cart is empty
3. Verify the new empty state UI displays correctly
4. Click on "scan a barcode" text link
5. Verify the scanner setup flow opens

### Testing information
- The empty cart state now shows a shopping cart icon with the message "Tap on a product or scan a barcode to add it to the cart"
- The "scan a barcode" text is styled as a clickable link (underlined, primary color)
- Clicking the link opens the same scanner setup flow as before
- The functionality remains unchanged, only the UI presentation has been updated

- Verified empty cart state displays correctly
- Tested clicking on the barcode link opens scanner setup
- Checked that the UI adapts properly for different screen sizes
- Confirmed string resources are used for all text (no hardcoded strings)

### Images/gif

https://github.com/user-attachments/assets/2a77cb34-8fdd-4f69-b4d9-eb956f5a9200


malinajirka:
Thanks @kidinov !

I'm still in the process of the review. I noticed one thing though - when I turn off the FF, the cart says only `Tap on a product to add it to the cart` without any mentions about the barcode scanning. Previously, the cart was empty I believe. Is this an intentional behavior?

malinajirka:
My two cents: The 'scan a barcode' clickable element works differently than I'd expect. It starts the barcode setup flow, while I'd expect it to show some information about barcode scanning. Not sure if it's natural for users to expect this button would start the setup flow, wdyt? Moreover, the highlighted text takes a lot of focus - it's quite prominent. I'd personally remove this link completely, and just keep the text. Having said that, if you all decided on this behavior, fine by me.

kidinov:
> I'm still in the process of the review. I noticed one thing though - when I turn off the FF, the cart says only Tap on a product to add it to the cart without any mentions about the barcode scanning. Previously, the cart was empty I believe. Is this an intentional behavior?

No it's never been empty, it was the same. The icon is slightly different compared to what it was before the project. Before it was a custom icon without proper dark mode support which we were hacking it with putting overlay on top of that therefore I changed the icon to use one of the standard ones

>  My two cents: The 'scan a barcode' clickable element works differently than I'd expect. It starts the barcode setup flow, while I'd expect it to show some information about barcode scanning. Not sure if it's natural for users to expect this button would start the setup flow, wdyt? Moreover, the highlighted text takes a lot of focus - it's quite prominent. I'd personally remove this link completely, and just keep the text. Having said that, if you all decided on this behavior, fine by me.

Thanks for the feedback! Keep the text without the link without clickability or without clickability? 

To me it sounds like:
* It's not clickable -> no link
* It's clickable -> link

Otherwise it doesn't make sense to keep it clickable as it's not clear that it is clickable and nobody will click it anyway

I am fine to leave it not clickable at all. @iamgabrielma  wdyt?

malinajirka:
I meant not clickable at all. Just a hint that a barcode scanner is supported.

> No it's never been empty, it was the same. 

Ohh, in that case I'm just confused :P.

iamgabrielma:
> Otherwise it doesn't make sense to keep it clickable as it's not clear that it is clickable and nobody will click it anyway
> I am fine to leave it not clickable at all. @iamgabrielma wdyt?

IMHO we should keep it clickable for now just because we remove the direct access to "initial barcode scanner setup". But I'm happy to remove it entirely if the design feedback says otherwise, for now let's keep it just because is what was proposed in the wireframes.

malinajirka (approved):

PR #14522, opened 2025-08-26
[Woo POS][Historical Orders] Order List and Details Split view layout

toupper:
Closes WOOMOB-1147

### Description

Implement the navigaton and the layout with dummy content for POS Orders.

### Steps to reproduce

1. Open POS
2. Select Menu ... -> Orders
3. Confirm the view appears in a split view
4. Confirm selection changes details view

### Testing information

Tested on Galaxy Tab A8 Android 14

### Images/gif

https://github.com/user-attachments/assets/db91b3bf-2909-4898-8bff-2732321495e7


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosOrdersViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -0,0 +1,190 @@
...
+            WooPosOrdersListPaneScreen(
+                orders = state.orders,
+                selectedOrderId = state.selectedOrderId,
+                onOrderSelected = viewModel::onOrderSelected,
+                modifier = Modifier.fillMaxSize()
+            )
+        }
+
+        WooPosOrdersDetailPaneScreen(
+            order = state.orders.firstOrNull { it.id == state.selectedOrderId },
```
kidinov: That's should not be in the UI, but the state should contain `selectedOrder` (order not id)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+class WooPosOrdersViewModel @Inject constructor() : ViewModel() {
+    private val _state = MutableStateFlow(WooPosOrdersState())
+    val state: StateFlow<WooPosOrdersState> = _state.asStateFlow()
+
+    fun onOrderSelected(orderId: Long) {
+        _state.update { it.copy(selectedOrderId = orderId) }
+    }
+}
+
+enum class WooPosOrdersDetailDestination { LIST, DETAIL }
```
kidinov: Unused?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -0,0 +1,190 @@
...
+            modifier = Modifier
+                .weight(0.7f)
+                .fillMaxHeight()
+                .background(MaterialTheme.colorScheme.surfaceContainerLow)
+        )
+    }
+}
+
+@Composable
+private fun OrdersToolbar(
```
kidinov: Probably better to reuse `WooPosToolbar` and I'd say we must have back button here as well in contrary to the settings

kidinov (approved):
LGTM!

I understand that this is largely WIP, but just in case some recommendations for the feature PRs:
* There is should be a back icon/button
* I don't think that we need to have chevron on every order on android
* The colors of the selection and style in general should match the settings. And it doesn't mean that we'd need to follow the settings, it maybe the opposite that we'd need to change the settings screen
* I think the state should be defined something like:

```
data class WooPosOrder(
    val id: Long,
    val title: String,
    val isSelected: Boolean,
)

sealed class WooPosOrdersState {
    object Loading : WooPosOrdersState()
    object Empty : WooPosOrdersState()
    data class Success(
        val leftPaneState: LeftPaneState,
        val rightPaneState: RightPaneState,
    ) : WooPosOrdersState() {
        sealed class LeftPaneState {
            object Loading : LeftPaneState()
            data class Loaded(
                val orders: List<WooPosOrder>
            ) : LeftPaneState()
            object Empty : LeftPaneState()
            data class Error(...) : LeftPaneState()
        }

        sealed class RightPaneState {
            object None : RightPaneState()
            data class Selected(val order: WooPosOrder) : RightPaneState()
        }
    }
    data class Error(...) : WooPosOrdersState()
}
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Interesting! Thanks for the comment. Why do you think here we need a back button other than the system one, but not in settings?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Yes, thanks, removed in 9e62f1be1f3377843986a6c91a176505a80b0733

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: Thanks for asking! After more consideration and discussing it with the AI, I believe we actually need to have a back button on both.

Initially why I went without it on the settings due to 2 factors:
* The Android's OS settings don't have back icon and I got confused by that. But now I understood as it's a separate app so in context of the app it's a root therefore it make sense that there is no back icon
image


* 2-pane layout in the settings has nested navigation in the second pane, so you can go on a level deeper, the back icon needs to be shown on the second pane. I thought that it would be weird to have two back navigation at the same time and not clear how to handle the system back navigation. But now I think that it's fine and makes sense while being not used much as mostly 2-pane screens are root screens of the apps. System back navigation can be handled by just giving priority to the second (right) pane navigation first, and if it's root in the second pane, then we navigate back from the whole screen. 

So all in all, I'll add back navigation to the settings

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, done in f86abf9a280dbc5459c6415216c321565f3aa37a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, done in 08f23483d6f356bd5a3894013041502f3d6443c3

toupper:
Thanks for the suggested next steps @kidinov! I removed the chevron in c38a081e09c73c29271d154e1b6fa31344f18b60 I will tackle these other improvements in future PRs

PR #14523, opened 2025-08-26
Issue/woomob 1062 woo poslocal catalog add new generate endpoint

malinajirka:
#WOOMOB1062

### Description

This PR integrates two new endpoints:
- POST /catalog
- GET /catalog/status/<id>

These endpoints aren't finalized on the server side, so we are parsing just the vital fields in the response for now. 

The code introduced in this PR isn't used yet, so green CI is a good enough test. We'll test the integration more thoroughly when we start using the endpoints.

### Steps to reproduce

N/A the code isn't used in the app.

N/A

### Images/gif

N/A


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/pos/PosCatalogStatusResponse.kt
```diff
@@ -0,0 +1,10 @@
+package org.wordpress.android.fluxc.model.pos
+
+import com.google.gson.annotations.SerializedName
+
+data class PosCatalogStatusResponse(
+    @SerializedName("status")
+    val status: String? = null,
```
kidinov, line 7: The plan to handle nulls in the app's level?

kidinov (approved):
LGTM!

Not related to the PR. To keep it more organized I'd suggest to put it inside of `localcatalog` package and extract there from `PosLocalCatalogStore` all the classes like `PosLocalCatalogError` and the results as well

image

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/pos/PosCatalogStatusResponse.kt
malinajirka: The plan was to handle the ones which are specified by the API contract on the store level, however, I forgot about it. I updated it in 1a20111fd2893fe337a723c7f4b9348848f9437f and added unit tests in 9fa2eda1514bf90324f14faa135a850c46eeeb95. Thanks for bringing it up!

malinajirka:
> Not related to the PR. To keep it more organized I'd suggest to put it inside of localcatalog package and extract there from PosLocalCatalogStore all the classes like PosLocalCatalogError and the results as well

Good idea, done in 04ff5e7b98b1130cdc002a2ee62713f82e1d8b95.


Thanks for the review!

PR #14524, opened 2025-08-26
[Login] Improve root endpoint response deserialization to fix login issues

hichamboushaba:
Closes WOOMOB-1179

> [!NOTE]
> Given we are going to depend more on the root API response for the Jetpack Performance project, I prioritized working on this task.

### Description 
When we released the Jetpack Performance experiment, we got a spike in the reported ParseErrors, and after investigation we found out this was related to the `namespaces` field in the root endpoint response, some sites return an object instead of the expected array. An example can be found here https://a8c.sentry.io/issues/6775905263/events/9b0ff9afc5804035ac31d0e5cfa2fed1.

After checking the API response, I found out that the response is a dictionary with numeric keys instead of array, like this:
```json
"namespaces": {
    "0": "akismet/v1",
    "1": "jetpack/v4",
    "2": "jetpack-boost/v1",
    "3": "my-jetpack/v1",
    "4": "jetpack/v4/explat",
    "5": "wc/v3",
    "6": "wc/v1",
    "7":  "wp/v2",
  },
```

So I updated our deserialization logic to handle this case too.

I was also able to identify a plugin that causes this behavior, so we now have exact steps to reproduce it. 

### Steps to reproduce
1. Install the [CoCart API plugin](https://wordpress.org/plugins/cart-rest-api-for-woocommerce/)
2. Open the app and sign in to the site, or switch to it (after switching to a different site)

### Testing information
- In `trunk` the above steps will lead to an error, but should succeed in this branch.

The above.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -9,6 +9,7 @@
 - [*] Shipping Labels: Fix handling of existing shipping labels with in-progress purchase status [https://github.com/woocommerce/woocommerce-android/pull/14501]
 - [Internal] [WEAR] Update horologist to 0.6.23 [https://github.com/woocommerce/woocommerce-android/pull/13395]
 - [WEAR] Updated UserAgent of API requests to use `http.agent` System Property to fix performance issues related to WebView usage on app launch [https://github.com/woocommerce/woocommerce-android/pull/14431]
+- [*] [Login] Improved the detection of WooCommerce API version during login flow [https://github.com/woocommerce/woocommerce-android/pull/14524]
 
 23.1
 -----
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/discovery/NamespacesDeserializer.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/discovery/NamespacesDeserializer.kt
@@ -1,0 +1,45 @@
+package org.wordpress.android.fluxc.network.discovery
+
+import com.google.gson.JsonDeserializationContext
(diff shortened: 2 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>NamespacesDeserializer</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

RELEASE-NOTES.txt
```diff
@@ -9,6 +9,7 @@
 - [*] Shipping Labels: Fix handling of existing shipping labels with in-progress purchase status [https://github.com/woocommerce/woocommerce-android/pull/14501]
 - [Internal] [WEAR] Update horologist to 0.6.23 [https://github.com/woocommerce/woocommerce-android/pull/13395]
 - [WEAR] Updated UserAgent of API requests to use `http.agent` System Property to fix performance issues related to WebView usage on app launch [https://github.com/woocommerce/woocommerce-android/pull/14431]
+- [*] [Login] Improved the detection of WooCommerce API version during login flow [https://github.com/woocommerce/woocommerce-android/pull/14524]
```
hichamboushaba, line 12: This is currently used only for WooCommerce API detection, so that's what I'm mentioning in the release note.

kidinov (approved):
I have not tested it e2e but the code and the tests LGTM

PR #14527, opened 2025-08-27
[WOOMOB-1048][Woo POS][Settings] Remove POS settings feature flag and reorganize product info dialog

kidinov:
WOOMOB-1048

### Description

Remove the WOO_POS_SETTINGS feature flag and make the code behave as if the flag is always enabled. Move the product information dialog from the home package to the settings package where it's actually used.

Changes include:
- Remove WOO_POS_SETTINGS feature flag and related code
- Always show Settings menu item in toolbar
- Always show barcode scanner option in empty cart state
- Move WooPosProductInfoDialog to settings package with dedicated package structure
- Clean up unused UI events and dialog states
- Remove obsolete string resources

### Steps to reproduce

1. Build and run the app
2. Navigate to WOO POS
3. Verify Settings menu item is always visible in toolbar
4. Navigate to cart and verify empty state shows barcode scanner option
5. Navigate to Settings and verify product info dialog works from there

### Testing information

- Empty cart state should always show barcode scanning option
- Settings menu should always be available in main toolbar
- Product information dialog should work from settings screen
- All existing functionality should remain unchanged

- Verified Settings menu appears correctly
- Verified empty cart UI shows scanner option
- Verified product info dialog works from settings

### Images/gif

https://github.com/user-attachments/assets/a1c68f71-2d87-465e-bb04-b738ea6b51d6


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
```diff
@@ -123,9 +123,6 @@ sealed class WooPosAnalyticsEvent : IAnalyticsEvent {
         data object ExitConfirmed : Event() {
             override val name: String = "exit_confirmed"
         }
-        data object GetSupportTapped : Event() {
```
malinajirka, line 126: ❓ Are we intentionally removing these two tracking events? Seems like we don't have any other events with the pos_ prefix that would replace these.

malinajirka (approved):
Thanks @kidinov! The changes LGTM. I've left one question.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
kidinov: It was intentional, but now that I think about it, I believe it’s actually not a good idea. We haven’t completely removed those features; instead, we’ve moved it to the settings.

I’ve returned the tracking to the settings

b5c0d2b4f436ae0c5348b56b3e2ffdc477a1b2df

PR #14529, opened 2025-08-27
Add ReaderNotConnected enum value

malinajirka:
### Description

This PR is just a technicality -> it explicitly adds `ReaderNotConnected` error as part of [the Stripe SDK update](https://github.com/woocommerce/woocommerce-android/pulls), however, it's still mapped to `Generic` error so the UX remains unchanged. 

### Testing information

Smoke testing IPP with a reader is enough.

I tested:
- Card reader in order creation (on phone and tablet)
- Card reader in POS (one tablet)
- Tap to pay in order creation (on phone)
- Simulated update of card reader (on tablet)

### Images/gif

N/A


JorgeMucientes (approved):
Thanks for addressing this update @malinajirka 

I tested the IPP flow using card reader and everything worked as expected. However, I couldn't test tap to pay flow as it always shows this dialog: 

https://github.com/user-attachments/assets/544ff4bc-cf29-42bb-adae-621c797becae

I notice this happens in `trunk` too. So it's unrelated to the Stripe SDK update. However it doesn't happen for me when I try to use tap to pay from production app for the same order failing in the video. Do I have to enable something specific to use tap to pay in Debug versions?

malinajirka:
@JorgeMucientes Thanks for testing, TTP works only on release builds, so you'd have to build vanillaRelease variant or so. Having said that, considering the changes are simple, the fact that I tested it, and the fact that we'll re-test during smoke testing, I think it's unnecessary to retest TTP here.

JorgeMucientes:
Sounds good! 

I've pre-approved the PR. Lint and instrumented tests are failing, though.

JorgeMucientes:
I think we can merge this right away into my branch, and let the other PR handle running the check. Feel free to merge without waiting @malinajirka

PR #14533, opened 2025-08-28
[Woo POS][Historical Orders] Orders Fetching

toupper:
## Description

Part of WOOMOB-1145

Infrastructure for remotely fetching required data from API, with the implementation of a dummy UI.

For consistency, I implemented a similar approach that we use with the items. 

## Steps to reproduce

1. Open POS -> Menu -> Orders
2. Confirm the loading state appears when orders are loading
3. Confirm correct POS orders are loaded, compare with wp-admin and Woo app orders

UI is out of scope, generated to facilitate testing. In-memory cache coming next.

## Testing information

## Screenshots

https://github.com/user-attachments/assets/330efadb-8bfc-4bbd-87dc-6f77f9be6b69

---


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -0,0 +1,41 @@
...
+            WooResult(result.error)
+        } else {
+            WooResult(result.model.toAppModels())
+        }
+    }
+
+    private suspend fun fetchOrdersFromStore(
+        page: Int
+    ): WooResult<List<OrderEntity>> {
+        return orderStore.fetchOrders(
```
kidinov: That's removes all the orders from the DB, including the orders from the main app.

I think we should probably use `OrderRestClient`

kidinov: Also, shell we filter out auto drafts here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
```diff
@@ -1,19 +1,13 @@
...
 
-@Suppress("MagicNumber")
 data class WooPosOrdersState(
-    val orders: List<WooPosOrder> = listOf(
-        WooPosOrder(1, "Order 1"),
-        WooPosOrder(2, "Order 2"),
-        WooPosOrder(3, "Order 3")
-    ),
-    val selectedOrderId: Long? = 1
+    val orders: List<Order> = emptyList(),
```
kidinov, line 6: I mentioned [the state](https://github.com/woocommerce/woocommerce-android/pull/14522#pullrequestreview-3158630427) in the previous PR. In my opinion, the foundation of any feature is a well-defined state and it's better to start with this firstly

Each piece of state should have one clear definition, eliminating ambiguity and preventing conflicting states from existing simultaneously. Currently your state can be defined, for instance, as:

`isLoading = true` && `error = "something` && `orders = list_of_orders` which is clearly should not be possible. This is the source of many bugs

So I'd recommend to define state firstly and then go from here otherwise a lot of will be missed resulting not in well implemented feature

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -55,12 +64,55 @@ fun WooPosOrdersScreen(
                 onBackClicked = onBackClicked,
             )
 
-            WooPosOrdersListPaneScreen(
-                orders = state.orders,
-                selectedOrderId = state.selectedOrderId,
-                onOrderSelected = viewModel::onOrderSelected,
-                modifier = Modifier.fillMaxSize()
-            )
+            when {
```
kidinov, line 62: I think this `when` is good illustration why badly defined state is a problem. You have many places here to make a mistake here just by not declaring (or declaring wrongly) one of the conditions (and no compiler check!) so you can end up in "else" case when it's not needed.

Also, `state.error != null` and `state.orders.isEmpty()` is kinda not view level logic - should be in the VM

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -55,12 +64,55 @@ fun WooPosOrdersScreen(
...
+                state.isLoading -> {
+                    Column(
```
kidinov, line 64: I recommend extracting all states here into separate composable functions with proper naming for better readability and code maintainability

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -0,0 +1,41 @@
...
+        return if (result.isError) {
+            WooResult(result.error)
+        } else {
+            WooResult(result.model.toAppModels())
+        }
+    }
+
+    private suspend fun fetchOrdersFromStore(
+        page: Int
+    ): WooResult<List<OrderEntity>> {
```
kidinov: I think it'd better to return either manually defined result or `kotlin.Result`. Ideally your own sealed class result, as `kotlin.Result` has it's own design issues as it is implemented as an inline value class wrapping Any?, but conceptually behaves like a sealed class with Success and Failure cases. This creates friction in the type system similar to `WooResult`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -40,6 +45,10 @@ fun WooPosOrdersScreen(
     val viewModel: WooPosOrdersViewModel = hiltViewModel()
     val state by viewModel.state.collectAsState()
 
+    LaunchedEffect(Unit) {
```
kidinov: That will refetch all the orders on any configuration change. Any reason to not fetch them in VM's `init`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -86,29 +138,51 @@ fun WooPosOrdersListPaneScreen(
...
             ) {
+                Column(
+                    verticalArrangement = Arrangement.spacedBy(WooPosSpacing.XSmall.value)
+                ) {
+                    WooPosText(
+                        "Order #${order.number}",
+                        style = WooPosTypography.BodyMedium
+                    )
+                    WooPosText(
+                        text = order.dateCreated.formatToDDMMMYYYY(),
```
kidinov, line 169: Should not be part of the view. All the data that comes to the view should be already ready to be displayed

kidinov (commented):
@toupper Overall, it looks great! Thanks! I have left a few suggestions. But firstly, I would suggest taking a step back and redesigning the state before proceeding further. This will save a lot of rework eventually

toupper:
Thanks for the review and comments @kidinov! They are very instructive and I will use them in future PRs.

As noted in the description, the UI (including state and view model) isn’t meant to be part of this PR. The UI code was added quickly just to help test fetching, not meant to be production-ready.

That said, I understand if you’d prefer not to merge code that feels suboptimal. One option would be to close this PR and open a new one focused only on the fetching logic, with the bare minimum needed to verify it works (e.g., logging the results).

Would you prefer that approach?

kidinov:
@toupper, I’m completely fine with merging it if it works for you this way. However, there are comments not only about the state but also about fetching and modifying the persistence of the main app, which we probably shouldn’t do

toupper:
@kidinov Could you please review it again? I’ve addressed all the comments except those specifically related to UI and state, which I plan to handle in the next PRs.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -0,0 +1,63 @@
...
+            page = page,
+            orderBy = OrderBy.DATE,
+            sortOrder = OrderRestClient.SortOrder.DESCENDING,
+            statusFilter = null,
+            createdVia = "pos-rest-api"
+        )
+
+        return if (result.isError) {
+            FetchOrdersResult.Error(
+                WooError(API_ERROR, SERVER_ERROR, result.error.message)
```
kidinov: How do you know that it's API error and not lets say network? I think if you don't need to expose specific error it's fine but at least we should not return incorrect/misleading error

Also, not sure if we actually need to have 2 return types for this, do we? Maybe something simple like that:

```
sealed class LoadOrdersResult {
    data class Success(val orders: List<Order>) : LoadOrdersResult()
    data class Error(val message: String) : LoadOrdersResult()
}

class WooPosOrdersDataSource @Inject constructor(
    private val restClient: OrderRestClient,
    private val orderMapper: OrderMapper,
) {
    suspend fun loadOrders(): LoadOrdersResult {
        val result = restClient.fetchOrders(
            site = selectedSite.get(),
            count = 25,
            page = 1,
            orderBy = OrderBy.DATE,
            sortOrder = OrderRestClient.SortOrder.DESCENDING,
            statusFilter = null,
            createdVia = "pos-rest-api"
        )

        return if (result.isError) {
            LoadOrdersResult.Error(result.error.message)
        } else {
            LoadOrdersResult.Success(result.orders.toAppModels())
        }
    }

    private suspend fun List<OrderEntity>.toAppModels(): List<Order> = map { 
        orderMapper.toAppModel(it) 
    }
}
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
```diff
@@ -1,19 +1,13 @@
...
+    val selectedOrderId: Long? = null,
+    val isLoading: Boolean = false,
+    val error: String? = null
 ) {
-    val selectedOrder: WooPosOrder?
+    val selectedOrder: Order?
```
kidinov, line 11: Not for this PR, so just to make sure that it won't be forgotten - we should have our `WooPosOrder` that contains info only needed to be shown. The order is data layer model

kidinov (approved):
LGTM! I left a couple of minor suggestions, please take a look!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Good point! I changed it to use `OrderRestClient`. We'll filter auto drafts in a future PR

PR #14534, opened 2025-08-28
WOOMOB-1057 - Introduce basic implementation of POS local catalog repository

malinajirka:
#WOOMOB-1057

### Description

This PR introduces a basic implementation of Local Catalog repository. It has several TBD comments, however, I decided to split the work to ensure the PRs remain reasonably long.

### Steps to reproduce

N/A The code isn't used yet, green CI is enough.

### Testing information

N/A

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14535, opened 2025-08-28
WOOMOB-1057 - Introduce basic implementation of background worker for POS local catalog

malinajirka:
#WOOMOB-1057

[Do not merge label - the parent PR needs to be merged first.](https://github.com/woocommerce/woocommerce-android/pull/14534)

### Description

This PR introduces a basic implementation of background worker for POS Local Catalog. It has several TBD comments, however, I decided to split the work to ensure the PRs remain reasonably long.

The scheduler uses static methods from the Android SDK, so I decided not to introduce any tests for it.

### Steps to reproduce

N/A The code isn't used yet, green CI is enough.

### Testing information

N/A

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
@@ -1,0 +1,102 @@
+package com.woocommerce.android.ui.woopos.localcatalog
+
+import android.content.Context
+import androidx.work.BackoffPolicy
+import androidx.work.Constraints
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.ExistingWorkPolicy
+import androidx.work.NetworkType
+import androidx.work.OneTimeWorkRequestBuilder
+import androidx.work.PeriodicWorkRequestBuilder
+import androidx.work.WorkInfo
+import androidx.work.WorkManager
+import com.woocommerce.android.util.WooLog
+import com.woocommerce.android.util.WooLog.T
+import dagger.hilt.android.qualifiers.ApplicationContext
+import java.util.Calendar
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class PosLocalCatalogSyncScheduler @Inject constructor(
+    @ApplicationContext private val context: Context
+) {
+
+    private companion object {
+        private const val ONE_TIME_WORK_NAME = "PosLocalCatalogSyncOneTime"
+
+        const val REFRESH_INTERVAL_HOURS = 24L
+        const val TIME_OF_DAY_FOR_PERIODIC_SYNC = 23 // 11 PM
+    }
+
+    private val workManager = WorkManager.getInstance(context)
+
+    fun schedulePeriodicFullCatalogSync() {
+        val syncWorkRequest = PeriodicWorkRequestBuilder<PosLocalCatalogSyncWorker>(
+            REFRESH_INTERVAL_HOURS,
+            TimeUnit.HOURS
+        )
+            .setInitialDelay(calculateDelayToNight(), TimeUnit.MILLISECONDS)
+            .setConstraints(getConstraints())
+            .setBackoffCriteria(
+                BackoffPolicy.EXPONENTIAL,
+                1,
+                TimeUnit.MINUTES
+            )
+            .build()
+
+        workManager.enqueueUniquePeriodicWork(
+            PosLocalCatalogSyncWorker.WORK_NAME,
+            ExistingPeriodicWorkPolicy.KEEP,
+            syncWorkRequest
+        )
+
+        WooLog.d(T.POS, "POS local catalog full sync scheduled.")
+    }
+
+    fun triggerManualFullCatalogSync() {
+        val oneTimeWorkRequest = OneTimeWorkRequestBuilder<PosLocalCatalogSyncWorker>()
+            .setConstraints(getConstraints())
+            .build()
+
+        workManager.enqueueUniqueWork(
+            ONE_TIME_WORK_NAME,
+            ExistingWorkPolicy.REPLACE,
+            oneTimeWorkRequest
+        )
+
+        WooLog.d(T.POS, "Manual POS local catalog sync triggered")
+    }
+
+    fun isPeriodicWorkRunning(): Boolean {
+        val periodicWork = workManager.getWorkInfosForUniqueWork(PosLocalCatalogSyncWorker.WORK_NAME).get()
+
+        return periodicWork.any { it.state == WorkInfo.State.RUNNING }
+    }
+
+    fun isOneTimeWorkRunning(): Boolean {
+        val oneTimeWork = workManager.getWorkInfosForUniqueWork(ONE_TIME_WORK_NAME).get()
+
+        return oneTimeWork.any { it.state == WorkInfo.State.RUNNING }
+    }
+
+    private fun getConstraints(): Constraints {
+        return Constraints.Builder()
+            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .setRequiresBatteryNotLow(true)
+            .build()
+    }
+
+    private fun calculateDelayToNight(): Long {
+        val now = Calendar.getInstance()
+        val night = Calendar.getInstance().apply {
+            set(Calendar.HOUR_OF_DAY, TIME_OF_DAY_FOR_PERIODIC_SYNC)
+            set(Calendar.MINUTE, 0)
+            set(Calendar.SECOND, 0)
+
+            if (before(now)) add(Calendar.DAY_OF_YEAR, 1)
+        }
+        return night.timeInMillis - now.timeInMillis
+    }
+}
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
@@ -10,8 +10,7 @@ import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkInfo
 import androidx.work.WorkManager
-import com.woocommerce.android.util.WooLog
-import com.woocommerce.android.util.WooLog.T
+import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
 import dagger.hilt.android.qualifiers.ApplicationContext
 import java.util.Calendar
 import java.util.concurrent.TimeUnit
@@ -20,7 +19,8 @@ import javax.inject.Singleton
 
 @Singleton
 class PosLocalCatalogSyncScheduler @Inject constructor(
-    @ApplicationContext private val context: Context
+    @ApplicationContext private val context: Context,
+    private val logger: WooPosLogWrapper,
 ) {
 
     private companion object {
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>PosLocalCatalogSyncScheduler</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
```diff
@@ -0,0 +1,102 @@
...
+        return periodicWork.any { it.state == WorkInfo.State.RUNNING }
+    }
+
+    fun isOneTimeWorkRunning(): Boolean {
+        val oneTimeWork = workManager.getWorkInfosForUniqueWork(ONE_TIME_WORK_NAME).get()
+
+        return oneTimeWork.any { it.state == WorkInfo.State.RUNNING }
+    }
+
+    private fun getConstraints(): Constraints {
```
malinajirka, line 89: 📓 The constraints aren't final - we'll sync these with iOS. We'll likely limit it to non-roaming or at least add some safeguard not to download large data sets in background on roaming.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.work.WorkerParameters
+import com.woocommerce.android.tools.SelectedSite
+import com.woocommerce.android.ui.login.AccountRepository
+import com.woocommerce.android.util.WooLog
+import com.woocommerce.android.util.WooLog.T
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedInject
+
+@HiltWorker
+class PosLocalCatalogSyncWorker @AssistedInject constructor(
```
malinajirka, line 14: 📓 The worker isn't scheduled yet, so the code is unused. This will be added when we finish all the subtasks.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val accountRepository: AccountRepository,
+    private val selectedSite: SelectedSite,
+    private val syncRepository: PosLocalCatalogSyncRepository,
+) : CoroutineWorker(appContext, workerParams) {
+
+    companion object {
+        const val WORK_NAME = "PosLocalCatalogSyncWork"
+    }
+
+    @Suppress("ReturnCount")
```
malinajirka, line 27: 📓 I added the suppress here, since having 3 return statements feels valid for something like this. Having said that, if you think it'd be better to refactor it, let me know.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
```diff
@@ -0,0 +1,65 @@
...
+    override suspend fun doWork(): Result {
+        if (!accountRepository.isUserLoggedIn()) {
```
kidinov, line 29: Is there a plan to only perform the work for the actual POS users?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
```diff
@@ -0,0 +1,102 @@
...
+            )
+            .build()
+
+        workManager.enqueueUniquePeriodicWork(
+            PosLocalCatalogSyncWorker.WORK_NAME,
+            ExistingPeriodicWorkPolicy.KEEP,
+            syncWorkRequest
+        )
+
+        WooLog.d(T.POS, "POS local catalog full sync scheduled.")
```
kidinov: There is `WooPosLogWrapper` which we wanted to use to avoid to have multiple places where we depend on the app's logger, so maybe worth reusing

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
```diff
@@ -0,0 +1,65 @@
...
+                    T.POS,
+                    "Local catalog sync completed successfully. Products: ${syncResult.productsSynced}, " +
+                        "Variations: ${syncResult.variationsSynced}, Duration: ${syncResult.syncDurationMs}ms"
+                )
+                Result.success()
+            }
+
+            is PosLocalCatalogSyncResult.Failure.UnexpectedError -> {
+                WooLog.e(T.POS, "Local catalog sync failed: ${syncResult.error}. Retrying ...")
+                Result.retry()
```
kidinov, line 55: We should probably set `setBackoffCriteria` manually unless there’s a reasonable default, otherwise, it will retry endlessly draining the battery.

kidinov: Just in case, probably it's planned. 

Overal we probably need to have different type of errors:
* network -> retry
* something 4XX -> one handling
* 5xx -> anotherone
* etc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
kidinov, line 14: Is there a plan to ensure that the worker is not started if it’s already in progress? Or is it handled already?

kidinov (approved):
LGTM! I left a couple of questions/ideas!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
malinajirka: Yes, but it made me realize we didn't have a ticket for it so I created on - https://linear.app/a8c/issue/WOOMOB-1219/woo-poslocal-catalog-constraint-syncing-to-pos-usersstores.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncScheduler.kt
malinajirka: Good point, updated in 6d671d82e86faa312f63c79c1d89ab25a2544316

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
malinajirka: > We should probably set setBackoffCriteria manually unless there’s a reasonable default, otherwise, it will retry endlessly draining the battery.

I set the exponential backoff for the periodic task, but your comment made me realize we need to set it even for the one-time task. Fixed in eab65add7df31a7e03c0d20c40e0c984a48dbb98.

Does exponential backoff starting on 1 minute sound good to you?

> Overal we probably need to have different type of errors:

We currently have different handling for `catalog too large`. I'm not sure if we should differentiate between network/5xx/4xx => all of them might be temporary. Depends on plugins setup/server state/auth/... Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
malinajirka: This is handled in the scheduler:
- Periodic won't be replaced and will just continue running if it runs or continues to be scheduled if it's scehduled.
- Manually triggered sync will be restarted.

The systems ensures there aren't two workers of the same type running at the same time.

Having said that, periodic and manual might be running next to each other - not sure if we need to handle this though - it's an edge case that might not even happen and even if it happens, nothing should go wrong since we are using transactions. Wdyt?

malinajirka:
Thanks for the review @kidinov! I've replied to your comments, let me know what you think 🙇

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
kidinov: > Does exponential backoff starting on 1 minute sound good to you?

Tbh I don't know. Probably

> I'm not sure if we should differentiate between network/5xx/4xx => all of them might be temporary. Depends on plugins setup/server state/auth/... Wdyt?

it depends. For instance `429 - Too Many Requests (rate limiting)` probably not a good idea to retry right away or at least we should wait longer

There is also header `Retry-After: XXX` - which we may need to consider. Even if WP doesn't send it, the reverse proxy like Cloudflare could send it (I didn't check myself)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
kidinov: > Having said that, periodic and manual might be running next to each other - not sure if we need to handle this though - it's an edge case that might not even happen and even if it happens, nothing should go wrong since we are using transactions. Wdyt?

I actually don't know what's going to happen then. Lets say you start transcation 1 and then transaction 2. Both transaction has some overlap with the same product ids. If transaction 1 committed then what would happen to transaction 2? Will it commit and we will end up with products with the same id or it will be reverted and we loose results of the second worker?

malinajirka: It depends on the transaction setting and I don't know what is the default for SQLite. But I believe one of the transaction will simply fail since the first one to start gets precedence. I believe that should be an ok behavior for our use-case. Both transactions are attempting to upsert/delete the same data or data that are a few seconds apart. We are going to run an incremental update after that, so we'll have up-to-date catalog.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/PosLocalCatalogSyncWorker.kt
malinajirka: I see your point, but I personally think this would introduce unnecessary complexity. The exponential backoff should handle such cases fine even if it's not the most optimal solution. I'll keep this in mind though, in case we'd realize another type of error that would require us to handle it properly.

Having said that, if you believe it's worth implementing, please feel free to add a ticket to the board and I'll implement it. Thanks!


> Does exponential backoff starting on 1 minute sound good to you?
> Tbh I don't know. Probably

I'll sync on this with iOS folks and keep it consistent across platforms.

PR #14538, opened 2025-08-29
[Jetpack Performance] Fetch app passwords support on app launch

hichamboushaba:
Closes WOOMOB-1181

> [!IMPORTANT]
> Please read the description thoroughly to gain a better understanding of the changes and their rationale.
> In the description, the root endpoint is the WordPress root endpoint which is generally at `/wp-json/`.
 
### Description
This PR's main task is to add the fetching of App Passwords support on app launch as discussed pe5sF9-4ye-p2#comment-4756, but while doing so, I added other improvements to optimize the number of requests that we make when fetching sites, and I'll explain below the changes.

#### Before this PR
Before this PR, the logic for fetching sites was as following:
- **For Jetpack Sites**: we were making a single API request to the WPCom endpoint `/sites/<site-id>`
- **For Jetpack CP sites**: since the endpoint `/sites/<site-id>` does not work, we made two separate requests — first, `/wc/v3/settings` to check if WooCommerce is installed, and then `/wp/v2/settings` to retrieve site metadata.

The above wasn't optimal, as we made two requests for Jetpack CP sites and self-hosted sites, where there is a better way that I'll explain below.

#### After this PR
With the changes of this PR, we'll make the following requests for each type of site:
- **For Jetpack Sites**: we'll continue to use `/sites/<site-id>`, and then we'll fetch whether the site supports App Passwords from the root endpoint.
- **For Jetpack CP sites**: we'll make a single request to the root endpoint, and this will allow us to get site metadata, whether WooCommerce is installed or not, and whether the site supports App Passwords or not.

With this change, we optimized the number of requests for all sites, and we added only a single request for Jetpack sites to detect whether App Passwords are supported or not.

**Requests when opening site picker**
With a single Jetpack CP site
| Before | After |
| --- | --- |
| Screenshot 2025-08-29 at 19 43 13 | Screenshot 2025-08-29 at 19 44 02 |

Also, the logic now is in `SiteStore` completely, so we simplified `WooCommerceStore`, as this is generally outside of its scope. Before, when FluxC was reused between us and the WPAndroid app, we couldn't make big modifications to `SiteStore`, now we have more freedom.

### Steps to reproduce
Test login with the three types of sites and confirm there is no regression:
1. A Jetpack-connected site.
2. A Jetpack CP site (by using the WooCommerce Payments plugin, for example).
3. A self-hosted site and site credentials login.

### Testing information
- Confirm signing in works without issues.
- Confirm switching sites using site picker works without issues.
- Relaunch the app and confirm there are no issues logged in logcat.

- The above.

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
@@ -3,9 +3,6 @@ package org.wordpress.android.fluxc.store
 import android.content.Context
 import com.wellsql.generated.SiteModelTable
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.async
-import kotlinx.coroutines.awaitAll
-import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.runBlocking
@@ -524,9 +521,5 @@ open class WooCommerceStore @Inject internal constructor(
         applyDecimalFormatting: Boolean
     ): String {
         return formatCurrencyForDisplay(amount.toString(), site, currencyCode, applyDecimalFormatting)
-    }
-
-    private suspend fun <A, B> Iterable<A>.pmap(f: suspend (A) -> B): List<B> = coroutineScope {
-        map { async { f(it) } }.awaitAll()
     }
 }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
@@ -92,9 +92,6 @@ open class WooCommerceStore @Inject internal constructor(
         const val WOO_API_NAMESPACE_V3 = "wc/v3"
     }
 
-    private val SiteModel.needsAdditionalCheckForWooInstallation
-        get() = origin == SiteModel.ORIGIN_XMLRPC || (origin == SiteModel.ORIGIN_WPCOM_REST && isJetpackCPConnected)
-
     override fun onRegister() = AppLog.d(T.API, "WooCommerceStore onRegister")
 
     @Subscribe(threadMode = ThreadMode.ASYNC)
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -92,9 +89,6 @@ open class WooCommerceStore @Inject internal constructor(
         const val WOO_API_NAMESPACE_V3 = "wc/v3"
     }
 
-    private val SiteModel.needsAdditionalCheckForWooInstallation
-        get() = origin == SiteModel.ORIGIN_XMLRPC || (origin == SiteModel.ORIGIN_WPCOM_REST && isJetpackCPConnected)
```
hichamboushaba, line 96: We don't support XMLRPC anymore (for quite some time), so the first condition here was just redundant.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -177,6 +187,28 @@ class SiteRestClient @Inject constructor(
...
+                                AppLog.T.API,
+                                "Error fetching root endpoint for Jetpack CP connected site: ${site.url}, error: ${it.error}"
+                            )
+                            site
+                        } else {
+                            it
+                        }
+                    }
+                }
+            }.awaitAll()
```
hichamboushaba, line 208: We do this to fetch requests for Jetpack CP sites in parallel, like we did before in `WooCommerceStore`.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -1373,7 +1373,7 @@ open class SiteStore @Inject constructor(
             site.webEditor = freshSiteFromDB.webEditor
 
             // The WPCom REST API doesn't return info about the application passwords authorize URL.
-            if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
+            if (site.origin == SiteModel.ORIGIN_WPCOM_REST && !includesAppPasswordsUrl) {
                 site.applicationPasswordsAuthorizeUrl = freshSiteFromDB.applicationPasswordsAuthorizeUrl
             }
```
hichamboushaba, line 1378: We preserve the `applicationPasswordsAuthorizeUrl` value from the cached `SiteModel` when fetching multiple sites using `/me/sites`, as we don't fetch the root endpoint in this case.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -222,6 +267,54 @@ class SiteRestClient @Inject constructor(
...
+    private suspend fun fetchSiteUsingRootEndpoint(site: SiteModel): SiteModel {
+        val result = jetpackTunnelGsonRequestBuilder.syncGetRequest(
+            restClient = this,
+            site = site,
+            url = "/",
+            params = mapOf("_fields" to ROOT_ENDPOINT_FIELDS),
+            clazz = RootWPAPIRestResponse::class.java
+        )
+
+        return when {
```
irfano, line 279: np suggestion: For binary conditions, I'd prefer to use `if` instead `when`. [Kotlin reference](https://kotlinlang.org/docs/coding-conventions.html#if-versus-when)

irfano (approved):
Good job, @hichamboushaba!  I tested the cases you mentioned in the description, as well as some additional login cases. Everything is working as expected. LGTM! 👍🏻

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
hichamboushaba: Thanks @irfano, I agree with you, but for this specific case, and to stay consistent with other usages in the RestClients, I'll keep the `when`.

PR #14543, opened 2025-09-01
[WOOMOB-1043] Extract scanning setup dialog from home to POS root

kidinov:
WOOMOB-1043

### Description
Moved the scanning setup dialog from the home package to the root POS package to make it placed correctly as it's used from both the home screen and settings screen now

### Steps to reproduce
1. Open POS home screen
2. Verify scanning setup dialog works as before
3. Navigate to POS settings
4. Verify scanning setup dialog is accessible from settings

### Testing information
Tested on tablet and phone devices. Both home and settings screens can now access the scanning setup dialog.

- Unit tests pass for the moved components
- Manual testing of scanning setup from both entry points


malinajirka (approved):
LGTM, thanks!

PR #14551, opened 2025-09-02
Fix UI tests after introducing logic to fetch app passwords on app launch

hichamboushaba:
### Description
The changes I introduced in #14538 seem to cause an intermittent UI tests failure because of an umatched request, this PR updates the UI tests to improve the matching of the root endpoint to fix the tests.

### Steps to reproduce
Green CI.

### DIFF
--- a/WooCommerce/src/androidTest/assets/mocks/mappings/jetpack-blogs/rest_api.json
+++ b/WooCommerce/src/androidTest/assets/mocks/mappings/jetpack-blogs/rest_api.json
@@ -6,8 +6,8 @@
       "json": {
         "equalTo": "true"
       },
-      "query": {
-        "matches": "(.*)fields(.*)authentication,namespaces(.*)"
+      "path": {
+        "equalTo": "/&_method=get"
       },
       "locale": {
         "matches": "(.*)"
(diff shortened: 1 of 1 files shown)


WooCommerce/src/androidTest/assets/mocks/mappings/jetpack-blogs/rest_api.json
```diff
@@ -6,8 +6,8 @@
       "json": {
         "equalTo": "true"
       },
-      "query": {
-        "matches": "(.*)fields(.*)authentication,namespaces(.*)"
+      "path": {
+        "equalTo": "/&_method=get"
```
hichamboushaba, line 10: I updated the logic to match against the path and not the query, which should be more robust for different cases and query arguments.

WooCommerce/src/androidTest/assets/mocks/mappings/jetpack-blogs/rest_api.json
```diff
@@ -19,13 +19,6 @@
     "jsonBody": {
       "data": {
         "namespaces": ["oembed\/1.0", "wpcomsh\/v1", "akismet\/v1", "jetpack\/v4", "wc\/v1", "wc\/v2", "amp\/v1", "wc-admin", "wc-analytics", "wc-product-add-ons\/v1", "wpcom\/v2", "wc\/v3", "wc\/store", "wccom-site\/v1", "wc-calypso-bridge\/v1", "wp\/v2", "wp-site-health\/v1"],
-        "authentication": {
-          "application-passwords": {
-            "endpoints": {
-              "authorization": "https:\/\/automatticwidgets.wpcomstaging.com\/wp-admin\/authorize-application.php"
-            }
-          }
-        },
```
hichamboushaba, line 28: I removed this part to avoid issues with more unmatched requests related to creating app passwords.

irfano (approved):
Your reasonings makes sense to me. I didn't run all UI tests multiple times for quicker review. I trust that CI has passed. LGTM! 👍🏻

PR #14556, opened 2025-09-03
Auto-merge `PATCH` Dependabot updates

wzieba:
Closes: AINFRA-1224

### Description

This PR enables auto-merge of `PATCH` updates from Dependabot. Before merging, it'll also:
- assign the closest milestone (that is not `due`)
- approve the PR

### Testing information

I've tested this on my fork:

#### 1. When the update is `PATCH`

- https://github.com/wzieba/woocommerce-android/pull/35
- Action: https://github.com/wzieba/woocommerce-android/actions/runs/17430367838/job/49487171350

#### 2. When the update is not `PATCH`

- https://github.com/wzieba/woocommerce-android/pull/56
- Action: https://github.com/wzieba/woocommerce-android/actions/runs/17430265089/job/49486859570

### DIFF
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -44,8 +44,12 @@ jobs:
 
       - name: Approve the PR
         if: steps.metadata.outputs.update-type == 'version-update:semver-patch'
-        run: gh pr review --approve "$PR_URL"
+        run: |
+          set -e
+          gh pr review --approve "$PR_URL"
 
       - name: Enable auto-merge for Dependabot PRs
         if: steps.metadata.outputs.update-type == 'version-update:semver-patch'
-        run: gh pr merge --auto --merge "$PR_URL"
+        run: |
+          set -e
+          gh pr merge --auto --merge "$PR_URL"
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -16,7 +16,7 @@ jobs:
       PR_URL: ${{ github.event.pull_request.html_url }}
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v5
 
       - name: Fetch Dependabot metadata
         id: metadata
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   auto-merge:
     runs-on: ubuntu-latest
-    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'wzieba/woocommerce-android'
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'woocommerce/woocommerce-android'
     env:
       PR_URL: ${{ github.event.pull_request.html_url }}
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -1,0 +1,48 @@
+name: Automerge Dependabot PATCH updates
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  assign-milestone:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'woocommerce/woocommerce-android'
+    env:
+      PR_URL: ${{ github.event.pull_request.html_url }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Fetch Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Exit if no Dependabot metadata found
+        if: steps.metadata.outputs.update-type != 'version-update:semver-patch'
+        run: |
+          echo "The update is not PATCH. Exiting."
+          exit 0
+
+      - name: Assign closest open milestone to PR
+        run: |
+          set -e
+          number=$(gh api repos/$GITHUB_REPOSITORY/milestones \
+            --jq '[.[]
+              | select(.state=="open" and .due_on!=null and (.due_on | fromdateiso8601) >= now)
+            ]
+            | sort_by(.due_on)
+            | .[0].number')
+
+          if [ -n "$number" ]; then
+            echo "Assigning milestone #$number to PR #${{ github.event.pull_request.number }}"
+            gh pr edit "${{ github.event.pull_request.number }}" --milestone "$number"
+          else
+            echo "No future open milestones found."
+          fi
+
+      - name: Approve the PR
+        run: gh pr review --approve "$PR_URL"
+
+      - name: Enable auto-merge for Dependabot PRs
+        run: gh pr merge --auto --merge "$PR_URL"
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -1,30 +1,31 @@
-name: Automerge Dependabot PATCH updates
+name: Auto-merge Dependabot PATCH updates
 
 on:
   pull_request:
-    types: [opened]
+    types: [opened, reopened]
+
+permissions:
+  contents: write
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.github/workflows/dependabot_automerge.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'woocommerce/woocommerce-android'
+    env:
+      PR_URL: ${{ github.event.pull_request.html_url }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v3
```
ParaskP7: Suggestion (💡): Maybe we should consider using the most recent `@v5` version of this [checkout](https://github.com/actions/checkout) action.

.github/workflows/dependabot_automerge.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Fetch Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Assign closest open milestone to PR
+        if: steps.metadata.outputs.update-type == 'version-update:semver-patch'
```
ParaskP7, line 28: Suggestion (💡): Just to avoid this `if` logic duplication, and thus potential future mistake where one is updated and not the other, maybe we could do something like the below:

1. Have another `name` step like:

```yml
      - name: Set patch update flag
        id: patch_flag
        run: |
          if [[ "${{ steps.metadata.outputs.update-type }}" == "version-update:semver-patch" ]]; then
            echo "This is a patch update."
            echo "is_patch=true" >> $GITHUB_OUTPUT
          else
            echo "This is not a patch update."
            echo "is_patch=false" >> $GITHUB_OUTPUT
          fi
```

2. Then using the below `if` logic instead:

```yml
if: steps.patch_flag.outputs.is_patch == 'true'
```

Wydt? 🤔

.github/workflows/dependabot_automerge.yml
```diff
@@ -0,0 +1,51 @@
...
+            | .[0].number')
+
+          if [ -n "$number" ]; then
+            echo "Assigning milestone #$number to PR #${{ github.event.pull_request.number }}"
+            gh pr edit "${{ github.event.pull_request.number }}" --milestone "$number"
+          else
+            echo "No future open milestones found."
+          fi
+
+      - name: Approve the PR
```
ParaskP7, line 45: Suggestion (💡): Just like you've used `set -e` for the `Assign closest open milestone to PR` step, which aborts on errors, this step, as well as the next one could benefit from more explicit error reporting for better debugging.

.github/workflows/dependabot_automerge.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: steps.metadata.outputs.update-type == 'version-update:semver-patch'
+        run: gh pr review --approve "$PR_URL"
+
+      - name: Enable auto-merge for Dependabot PRs
```
ParaskP7, line 51: Suggestion (💡): Maybe we should check that `auto-merge` is enabled for the repository. If not, this merge step will fail silently, right? 🤔

ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR as per the instructions, everything LGTM, nicely done! 🌟 x 🌟 ^ 🌟 

PS: Thanks for doing the testing part of your personal repo, this helped me a lot! 🥇 

(Soft) Warning (⚠️): I wonder how this GitHub Action would manage group updates like on this [kotlin-ksp](https://github.com/woocommerce/woocommerce-android/blob/trunk/.github/dependabot.yml#L11) group, or other such group updates where you could have both a minor + patch update, or major + patch update, or all 3 for that matter (me thinking of DOAndroid atm). Wdyt? 🤔

-----

I have left few suggestions (💡) for you to consider. I am going to approve this PR anyway, since none is blocking, give of take this (soft) warning (⚠️) above. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

.github/workflows/dependabot_automerge.yml
wzieba: Thanks but to me, the "Set patch update flag" looks more complex and less readable. It doesn't remove code, but adds a new step. We also don't decrease "`if` logic duplication" - we still have to keep it, just the `if` condition is changed. I'd leave this as it is now.

ParaskP7: Yea @wzieba , I was thinking about it too, I decided to post this comment anyway because:

1. You get a nice (debugging) log: `This is (not) a patch update.` 🗒️
2. `steps.metadata.outputs.update-type == 'version-update:semver-patch'` felt less complicate than `steps.patch_flag.outputs.is_patch == 'true'`. My end goal was to somehow remove this duplicate `if` check and this is the best I could do. 🤷

I agree with your reasoning nevertheless. 👍

.github/workflows/dependabot_automerge.yml
wzieba: > If not, this merge step will fail silently, right? 🤔

I don't know - have you checked it? I'm guessing it won't be silent, why would it be? Anyway, I think there's no point in making this more complex with additional checks - I assume that if a project wants auto-merge, they will have auto-merge enabled in GitHub settings.

wzieba:
> (Soft) Warning (⚠️): I wonder how this GitHub Action would manage group updates like on this [kotlin-ksp](https://github.com/woocommerce/woocommerce-android/blob/trunk/.github/dependabot.yml#L11) group, or other such group updates where you could have both a minor + patch update, or major + patch update, or all 3 for that matter (me thinking of DOAndroid atm). Wdyt? 🤔

I assume that the version from Dependabot will match the highest update in the group. So if an update will have minor + patch, then the version of the group will be minor.

.github/workflows/dependabot_automerge.yml
ParaskP7: > I don't know - have you checked it? I'm guessing it won't be silent, why would it be?

Yea, no @wzieba , I haven't checked it, it is just something my friend `@ai` suggested and I thought I should mention, just in case you knew better.

>  Anyway, I think there's no point in making this more complex with additional checks - I assume that if a project wants auto-merge, they will have auto-merge enabled in GitHub settings.

I guess so, and if the project doesn't have `Allow auto-merge` enabled for their project in [Settings/General](https://github.com/woocommerce/woocommerce-android/settings), then I guess we'll enable that for them before creating such a GitHub Action, correct? 🤔

image

ParaskP7:
> I assume that the version from Dependabot will match the highest update in the group. So if an update will have minor + patch, then the version of the group will be minor.

Is it (maybe) possible to test this assumption on your private repo @wzieba ? 😊

FYI: My biggest fear with that is that it will merge major or minor updates just because of an associated batch update in a group, again, maybe this couldn't (possibly) apply to this one `kotlin-ksp` group on this repo, but for sure it could affect DOAndroid because they have multiple such groups.

PS: Another think I wonder is whether it will handle group updates at all, maybe it will not be able to derive that single value from the metadata, because it is now a list of values when on groups updates, not sure...

wzieba:
> I wonder is whether it will handle group updates at all

It will, here's an example: https://github.com/wzieba/woocommerce-android/actions/runs/17429610191/job/49484709765#step:3:14

> Is it (maybe) possible to test this assumption on your private repo

Sure 👍

wzieba:
So I confirmed the assumption. I couldn't easily do this with my public fork, but I privately forked DOAndroid and for updates like:

Screenshot 2025-09-04 at 20 36 07

which includes patch, minor and major updates, the semver of the group is the highest one: major:

Screenshot 2025-09-04 at 20 35 49

.github/workflows/dependabot_automerge.yml
wzieba: Sure 👍

ParaskP7:
> So I confirmed the assumption. I couldn't easily do this with my public fork, but I privately forked DOAndroid and for updates like:

Thank you for the effort of testing this @wzieba ! 🙇 ❤️ 

> which includes patch, minor and major updates, the semver of the group is the highest one: major:

Nice, success! 🎯

PR #14559, opened 2025-09-03
[Jetpack Performance] Add a preference toggle for enabling app passwords for Jetpack sites

hichamboushaba:
Closes WOOMOB-1183

### Description
This PR adds a toggle that allows disabling using Application Passwords for Jetpack sites, it's meant as a temporary solution for any potential issues that we may miss.
The toggle is displayed when the feature flag (will be replaced with a remote feature flag in a future PR) is enabled, and when the current connection is using a WordPress.com account.

I also renamed the section from "Beta features" to "Experimental features" to match iOS platform, and to make the section more suitable for this toggle.

### Steps to reproduce
1. Open the app settings.
2. Tap on Experimental features.
3. Notice the toggle.

### Testing information
- Confirm the toggle is enabled by default.
- Inspect network calls, and change the value if the toggle and confirm it works as expected.

The above.


irfano (approved):
Works as expected! LGTM! 👍🏻

PR #14517, opened 2025-08-25
Updates or removes dependecies to prepare app for 16KB page size requirement

JorgeMucientes:
Closes  WOOMOB-515

### Description

To make our codebase compatible with the new 16KB page size requirement I've applied the following changes: 

- Updated Stripe Terminal SDK to 4.6.0 (latest one): https://github.com/stripe/stripe-terminal-android/issues/577#issuecomment-3145860110.
- Adds an upgraded version of `org.aomedia.avif.android:avif` that Glide library was using. I though of removing media avif file format all together, but it looks like we need to support it: https://github.com/woocommerce/woocommerce-android/issues/12810
- Removes Flipper from the codebase. Flipper use multiple native libraries that are not compatible with 16KB page sizes. Moreover Flipper is not maintained anymore. I checked and there doesn't seem to be any plans on fixing this incompatibility issues. https://github.com/facebook/flipper/issues/5634#issuecomment-2447723072. Most of Flipper functionality is covered with Android Studio App Inspector. I'm open to alternative approaches to keep Flipper and make the app compatible with 16KB, but from my initial research this wasn't straightforward. 

### Testing information
Test 16KB page size compatibility. 2 options: 

1. Build the [project and use APK analyzer](https://developer.android.com/guide/practices/page-sizes) and check there's no warning like this:
Screenshot 2025-08-25 at 19 47 21

2. [Set up a 16KB based Android 15+ emulator](https://developer.android.com/guide/practices/page-sizes#16kb-emulator) and check the following dialog is not show:

Screenshot 2025-08-25 at 19 46 32

**Regression tests:** 
1. Smoke test all the IPP flows and ensure the new version of the Stripe SDK doesn't introduce any bug. I pretty much tested paying and order using the card reader. 
3. To test avif file format follow [this PR instructions](https://github.com/woocommerce/woocommerce-android/pull/12835): 

> To test this, you will need at least one product configured with an image in the AVIF format. I highly recommend using [this one](https://raw.githubusercontent.com/link-u/avif-sample-images/refs/heads/master/fox.profile0.8bpc.yuv420.avif) from [an open source AVIF samples repository](https://github.com/link-u/avif-sample-images?tab=readme-ov-file).
Please bear in mind that the AVIF format supports multiple different metadata and configurations, and the Android OS doesn't support them all, some AVIF images will load, while others may not.
The testing should ensure that sections like Orders, Product Details, Product List, Product Gallery, and related sections successfully display the image in the AVIF format.

### The tests that have been performed

[...]

PR #12835, opened 2024-10-29
Introduce AVIF displaying support

ThomazFB:
Summary
==========
Depends on https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/154

Fix issue #12810 by introducing an update from the PhotonUtils in the `WordPress-Utils-Android` and adding the AVIF integration to the Glide dependency. 

There's no code to cover with unit tests in the PR changelog.

How to Test
==========
To test this, you will need at least one product configured with an image in the AVIF format. I highly recommend using [this one](https://raw.githubusercontent.com/link-u/avif-sample-images/refs/heads/master/fox.profile0.8bpc.yuv420.avif) from [an open source AVIF samples repository](https://github.com/link-u/avif-sample-images?tab=readme-ov-file).

Please bear in mind that the AVIF format supports multiple different metadata and configurations, and the Android OS doesn't support them all, some AVIF images will load, while others may not.

The testing should ensure that sections like Orders, Product Details, Product List, Product Gallery, and related sections successfully display the image in the AVIF format.

## Reviewer (or Author, in the case of optional code reviews):

Please make sure these conditions are met before approving the PR, or request changes if the PR needs improvement:

- [X] The PR is small and has a clear, single focus, or a valid explanation is provided in the description. If needed, please request to split it into smaller PRs.
- [X] Ensure Adequate Unit Test Coverage: The changes are reasonably covered by unit tests or an explanation is provided in the PR description.
- [X] Manual Testing: The author listed all the tests they ran, including smoke tests when needed (e.g., for refactorings). The reviewer confirmed that the PR works as expected on big (tablet) and small (phone) in case of UI changes, and no regressions are added.

PR #14562, opened 2025-09-03
Upgrade Glide to v5

JorgeMucientes:
### Description

Coming from this other PR: https://github.com/woocommerce/woocommerce-android/pull/14517
Glide has been recently updated to a version compatible with the 16KB page sizes on Android. So, part of the changes introduced in the above linked PR to update AVIF dependency manually are not needed anymore. 

1. Test codebase is still compatible with the 16KB page size. Build the [project and use APK analyzer](https://developer.android.com/guide/practices/page-sizes) and check there's no warning like this:

Screenshot 2025-08-25 at 19 47 21 

2. To test that avif file format still works after the update follow [this PR instructions](https://github.com/woocommerce/woocommerce-android/pull/12835): 

> To test this, you will need at least one product configured with an image in the AVIF format. I highly recommend using [this one](https://raw.githubusercontent.com/link-u/avif-sample-images/refs/heads/master/fox.profile0.8bpc.yuv420.avif) from [an open source AVIF samples repository](https://github.com/link-u/avif-sample-images?tab=readme-ov-file).
Please bear in mind that the AVIF format supports multiple different metadata and configurations, and the Android OS doesn't support them all, some AVIF images will load, while others may not.
The testing should ensure that sections like Orders, Product Details, Product List, Product Gallery, and related sections successfully display the image in the AVIF format.

### DIFF
--- a/libs/login/src/main/java/org/wordpress/android/login/util/AvatarHelper.kt
+++ b/libs/login/src/main/java/org/wordpress/android/login/util/AvatarHelper.kt
@@ -16,52 +16,58 @@ import com.gravatar.types.Email
 import org.wordpress.android.login.R
 
 object AvatarHelper {
-    @JvmStatic fun loadAvatarFromEmail(
+    @JvmStatic
+    fun loadAvatarFromEmail(
         fragment: Fragment,
         email: String?,
         avatarView: ImageView,
         listener: AvatarRequestListener
     ) {
         val avatarSize = fragment.resources.getDimensionPixelSize(R.dimen.avatar_sz_login)
-        val avatarUrl = email?.let { AvatarUrl(Email(email),
-                AvatarQueryOptions(preferredSize = avatarSize, defaultAvatarOption = Status404)).url().toString() }
+        val avatarUrl = email?.let {
+            AvatarUrl(
+                Email(email),
+                AvatarQueryOptions(preferredSize = avatarSize, defaultAvatarOption = Status404)
+            ).url().toString()
+        }
         loadAvatarFromUrl(fragment, avatarUrl, avatarView, listener)
     }
 
-    @JvmStatic fun loadAvatarFromUrl(
+    @JvmStatic
+    fun loadAvatarFromUrl(
         fragment: Fragment,
         avatarUrl: String?,
         avatarView: ImageView,
         listener: AvatarRequestListener
     ) {
         Glide.with(fragment)
-                .load(avatarUrl)
-                .apply(RequestOptions.circleCropTransform())
-                .apply(RequestOptions.placeholderOf(R.drawable.ic_user_circle_no_padding_grey_24dp))
-                .apply(RequestOptions.errorOf(R.drawable.ic_user_circle_no_padding_grey_24dp))
-                .listener(object : RequestListener<Drawable?> {
-                    override fun onLoadFailed(
-                        e: GlideException?,
-                        model: Any?,
-                        target: Target<Drawable?>,
-                        isFirstResource: Boolean
-                    ): Boolean {
-                        listener.onRequestFinished()
-                        return false
-                    }
+            .load(avatarUrl)
+            .apply(RequestOptions.circleCropTransform())
+            .apply(RequestOptions.placeholderOf(R.drawable.ic_user_circle_no_padding_grey_24dp))
+            .apply(RequestOptions.errorOf(R.drawable.ic_user_circle_no_padding_grey_24dp))
+            .listener(object : RequestListener<Drawable?> {
+                override fun onLoadFailed(
+                    e: GlideException?,
+                    model: Any?,
+                    target: Target<Drawable?>,
+                    isFirstResource: Boolean
+                ): Boolean {
+                    listener.onRequestFinished()
+                    return false
+                }
 
-                    override fun onResourceReady(
-                        drawable: Drawable,
-                        model: Any,
-                        target: Target<Drawable?>,
-                        dataSource: DataSource,
-                        isFirstResource: Boolean
-                    ): Boolean {
-                        listener.onRequestFinished()
-                        return false
-                    }
-                })
-                .into(avatarView)
+                override fun onResourceReady(
+                    resource: Drawable?,
+                    model: Any?,
+                    target: Target<Drawable?>?,
+                    dataSource: DataSource?,
+                    isFirstResource: Boolean
+                ): Boolean {
+                    listener.onRequestFinished()
+                    return false
+                }
+            })
+            .into(avatarView)
     }
 
     interface AvatarRequestListener {
--- a/WooCommerce/build.gradle
+++ b/WooCommerce/build.gradle
@@ -292,10 +292,7 @@ dependencies {
     implementation(libs.bumptech.glide.main)
     ksp(libs.bumptech.glide.compiler)
     implementation(libs.bumptech.glide.volley.integration)
-    implementation libs.avif
-    implementation(libs.bumptech.glide.avif.integration) {
-        exclude group: 'org.aomedia.avif.android', module: 'avif'
-    }
+    implementation(libs.bumptech.glide.avif.integration)
     implementation(libs.google.play.app.update)
     implementation(libs.google.play.review)
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -45,8 +45,7 @@ automattic-encryptedlogging = '1.0.0'
 automattic-measure-builds = '3.2.1'
 automattic-tracks = '6.0.4'
 assertj = '3.27.3'
-avif = "1.3.0.841110fd"
-bumptech-glide = '4.16.0'
+bumptech-glide = '5.0.4'
 cashapp-turbine = '1.0.0'
 coil = '2.7.0'
 commons-fileupload = '1.6.0'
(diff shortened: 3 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/login/src/main/java/org/wordpress/android/login/util/AvatarHelper.kt
```diff
@@ -16,52 +16,58 @@ import com.gravatar.types.Email
...
+                    resource: Drawable?,
+                    model: Any?,
+                    target: Target<Drawable?>?,
+                    dataSource: DataSource?,
+                    isFirstResource: Boolean
+                ): Boolean {
+                    listener.onRequestFinished()
+                    return false
+                }
+            })
```
JorgeMucientes, line 69: Only real changes on this file. Had to update the function signature to the Glide provided interface.

ParaskP7:
👋 @JorgeMucientes and thanks for this change! 🥇

After skim reviewing/testing this, I have the following:

1. I noticed the [diff](https://github.com/woocommerce/woocommerce-android/pull/14562#issuecomment-3250508205) on this change and saw that:
    - With this Glide `v5` update, we are actually now downgrading AVIF from `1.3.0.841110fd` (`Jul 17, 2025`) to `1.1.1.14d8e3c4` (`Aug 20, 2024`).
    - But, looking at AVIF's release notes I see that they added support for 16KB page size as part of [1.2.0](https://github.com/AOMediaCodec/libavif/releases/tag/v1.2.0) release (`Feb 25, 2025`). 🤷
    - No matter, Glide does mention the AVIF update to align with 16KB page size nevertheless on their [v5-rc5](https://github.com/bumptech/glide/releases/tag/v5.0.0-rc05) release. This is a bit confusing since Glide is using AVIF's `1.1.1.14d8e3c4` transitive dependency via their `avif-integrations` dependency, which shouldn't be compatible, and I would have expected them to use the latest `1.3.0.841110fd` transitive dependency instead. 🤷
    - Still, analyzing the APK it indeed showing as 16KB page size compatible. 🤷
    - All that to say, I am a bit confused and skeptical as to why this `v5` release of Glide is 16KB page size compatible. 🤔
2. I also tried testing AVIF following the testing instructions on #12835, downloading [this one](https://raw.githubusercontent.com/link-u/avif-sample-images/refs/heads/master/fox.profile0.8bpc.yuv420.avif) (`fox`) and another image from this [open source AVIF samples repository](https://github.com/link-u/avif-sample-images?tab=readme-ov-file) (`kimono`). Still, I couldn't manage to upload such an AVIF image on a product. I was constantly getting upload failures.
    - Image upload errors: `media must define a valid MIME type`
    - Any idea if that AVIF feature works as expected after that #14517 change, or in general, even before that chagne? 🤔

AdamGrzybkowski:
@ParaskP7 About the AVIF versioning difference - [it looks like they used an unofficial release before the 1.2.0 was published](https://github.com/AOMediaCodec/libavif/issues/2464#issuecomment-2486890946)? 🤔 Anyway, I can also confirm that this build looks to be 16 KB compliant.

ParaskP7:
> @ParaskP7 About the AVIF versioning difference - [it looks like they used an unofficial release before the 1.2.0 was published](https://github.com/AOMediaCodec/libavif/issues/2464#issuecomment-2486890946)? 🤔

🤔 and nice find on that @AdamGrzybkowski , interesting indeed, and also, so confusing, oh well... 😡 🤷 😡

FYI: To be honest, I actually (kind of) real safer if we keep the extra `avif` dependency we've already declared and replaced the transitive one provided automatically (via `Glide`), but also, don't want to derail further away from the official use of `AVIF` with `Glide`, especially if it works and is 16KB page size compatible... 🤔 🤷 🤔 

> Anyway, I can also confirm that this build looks to be 16 KB compliant.

👍 Thanks! 🙇

AdamGrzybkowski (approved):
As @ParaskP7 I couldn't upload a new product AVIF image, the upload fails. I tested on trunk and I observed the same behavior. This PR is also about displaying images so I don't think it's related. Other than that looks good! 

Here's what I tested:

1. 16 KB compatibility - ✅ 
2. Loading an avatar - ✅ 
Screenshot_20250908_173901

2. Displaying AVIF images - ✅ 

| Screen | Screenshot |
|---|---|
| Product list | Screenshot_20250908_173327 |
| Product details | Screenshot_20250908_173334 |
| Product gallery | Screenshot_20250908_173354 |
| Full screen image | Screenshot_20250908_173402 |
| Product selection | Screenshot_20250908_173422 |
| Blaze campaign card (?) | Screenshot_20250908_173446 |

ParaskP7:
> As @ParaskP7 I couldn't upload a new product AVIF image, the upload fails. I tested on trunk and I observed the same behavior. This PR is also about displaying images so I don't think it's related. Other than that looks good!

Awesome @AdamGrzybkowski and thanks (🚀), about the `AVIF` image, I am guessing you uploaded the `AVIF` image via the web interface, and then tested that it is displayed okay, right? 🙏

AdamGrzybkowski:
> about the AVIF image, I am guessing you uploaded the AVIF image via the web interface, and then tested that it is displayed okay, right? 🙏

Yes.

JorgeMucientes:
Thanks so much for the exhaustive review @ParaskP7 and @AdamGrzybkowski. 

And yes, I confirm the AVIF support added to the app back then was only for displaying AVIF images. I think we never worked on adding support to upload AVIF images from the app, and I suspect the reason for that is that officially WordPress doesn't support [AVIF images](https://wordpress.com/support/accepted-filetypes/). So, we currently support the following types (which replicate the ones from [WordPress docs](https://wordpress.com/support/accepted-filetypes/)). [MimeTypes.kt](https://github.com/woocommerce/woocommerce-android/blob/f62c9f8455ddd8286811e9154e32131e3ee091e8/libs/fluxc/src/main/java/org/wordpress/android/fluxc/utils/MimeTypes.kt#L77)
```kotlin
    /*
     * The WordPress supported image types based on https://wordpress.com/support/accepted-filetypes/ are:
     * .jpg, .jpeg, .png, .gif, .webp
     * This translates (based on https://android.googlesource.com/platform/frameworks/base/+/cd92588/media/java/android/media/MediaFile.java) to:
     * .jpg, .jpeg - "image/jpeg"
     * .png - "image/png"
     * .gif - "image/gif"
     * .webp - "image/webp"
     * .heic - "image/heic"
     * .heif - "image/heif"
     */
    @Suppress("MaxLineLength")
    private val imageTypes = listOf(
            MimeType(IMAGE, Subtype.JPEG, listOf("jpg", "jpeg")),
            MimeType(IMAGE, Subtype.PNG, listOf("png")),
            MimeType(IMAGE, Subtype.GIF, listOf("gif")),
            MimeType(IMAGE, Subtype.WEBP, listOf("webp")),
            MimeType(IMAGE, Subtype.HEIC, listOf("heic")),
            MimeType(IMAGE, Subtype.HEIF, listOf("heif"))
    )
```

> FYI: To be honest, I actually (kind of) real safer if we keep the extra avif dependency we've already declared and replaced the transitive one provided automatically (via Glide), but also, don't want to derail further away from the official use of AVIF with Glide, especially if it works and is 16KB page size compatible... 🤔 🤷 🤔

After your analysis above I'm also a bit hesitant about merging this PR as the AVIF version used by Glide doesn't explicitly support 16KB size, although it looks like it is compatible from all the checks. Derailing from official Glide updates doesn't seem like a big concern given Glide usage is deprecated and we are using Coil now ad default lib for image loading. 

I'm good to close this PR without merging. But I'd like to know what @AdamGrzybkowski thinks as well :)

ParaskP7:
Thanks for the reply @JorgeMucientes and providing all this extra context on everything AVIF! 🙏 

> After your analysis above I'm also a bit hesitant about merging this PR as the AVIF version used by Glide doesn't explicitly support 16KB size, although it looks like it is compatible from all the checks. Derailing from official Glide updates doesn't seem like a big concern given Glide usage is deprecated and we are using Coil now ad default lib for image loading.

Given this new Glide v5 version comes with more improvements than just AVIF, I would personally keep this update, but maybe, a middle ground could be, to also keep the explicit AVIF dependency declaration/configuration.

image

PS: Having said that I am not 100% sure if that's the safest thing to do and maybe just using Glide's v5 version of AVIF is safer, although yes, a bit weird what they did there with AVIF... 🤷

AdamGrzybkowski:
There's always a risk when overriding a transitive dependency, so unless we have a good reason, I would opt out of doing that. 

In this case, the one that Glide v5 uses solves the 16 KB compatibility issue.

...but I see that we are using this dependency directly, so in that case, we might have another reason to use the latest version (bugfixes). At this point, I don't have enough context about how we use that dependency, so it's hard for me to suggest the best way forward.

ParaskP7:
> There's always a risk when overriding a transitive dependency, so unless we have a good reason, I would opt out of doing that.

This is true, and also, since I don't see any `avif` related import anywhere, there is no reason to have this transitive dependency declared directly.

> In this case, the one that Glide v5 uses solves the 16 KB compatibility issue.

👍

> ...but I see that we are using this dependency directly, so in that case, we might have another reason to use the latest version (bugfixes). At this point, I don't have enough context about how we use that dependency, so it's hard for me to suggest the best way forward.

There is none other reason @AdamGrzybkowski , and actually, we added this dependency directly as part of #14517, just to make WCAndroid 16KB page size compatible. At that point, 2/3 weeks back, this was the only feasible alternative.

---

Ok, then let's merge this PR as is @JorgeMucientes , using Glide v5 with the AVIF dependency it provides and take it from there. 🚀

JorgeMucientes:
Thanks for the good points provided folks 🙏🏼 
I'll proceed with merging then.

PR #14565, opened 2025-09-04
POS Product Migration Step 1

malinajirka:
Fixes WOOMOB-1243

Let me know if this PR is too big - roughly 400 lines of logic. I can extract rename of the entity model and nullability change to the database into a separate PR.

### Description
This PR introduces a new POS-specific product model (`WooPosProductModelVersion2`) for the UI/VM layer to replace direct usage of the database entity. The change improves separation of concerns by:
- Renaming the existing `WCPosProductModel` to `WCPosProductEntity` to clarify it's a database entity
- Creating a new `WooPosProductModelVersion2` as a proper domain model for the UI layer
- Implementing a mapper to convert between the entity and the model
- Making `parentId` and `stockQuantity` nullable to better represent optional product attributes

**Note:** The `Version2` suffix in the model name is temporary to avoid unnecessary conflicts during migration. Once the full migration is complete, this suffix will be removed.

### Testing information
The code isn't used yet, so there is nothing to test at the moment.

### Images/gif
N/A - This is an internal refactoring with no UI changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModelVersion2.kt
```diff
@@ -0,0 +1,118 @@
...
+    val sku: String,
+    val globalUniqueId: String,
+    val type: WooPosProductType,
+    val status: WooPosProductStatus,
+    val pricing: WooPosPricing,
+    val description: String,
+    val shortDescription: String,
+    val isDownloadable: Boolean,
+    val lastModified: String,
+    val images: List<WooPosProductImage> = emptyList(),
```
malinajirka, line 25: 🤔 I have been wondering whether these attributes should by parsed lazily - however, we'd have to have an access to a gson instance. Passing it in feels off and creating it per model feels of as well. Which leaves us with a single instantiation within companion object, wdyt?

samiuelson: I know we already store attributes in `WCPosProductEntity` table, but do we really need them in POS? Can we remove them, maybe, instead of passing them down to the domain layer?

malinajirka: Good question. I have been following pdfdoF-7VS-p2 and plan to remove unnecessary columns in https://linear.app/a8c/issue/WOOMOB-1120/woo-poslocal-catalog-remove-unnecessary-columns-from-pos-products. It's difficult to say now which columns will be needed and which not. What do you think about this approach?

malinajirka: But yeah, I agree it would remove the need to think about lazy loading now. Let me add a note to that ticket to also consider the lazy loading when we know more.

samiuelson: Anyway, I'm not a fan of lazy parsing, and would propose parsing JSON while mapping the entity. This would give us:
* Type safety and clean models
* Fail fast advantage: JSON parsing errors caught early during mapping, not during UI rendering
* Performance?: I'm not 100% sure, but maybe in some cases, lazy parsing would cause worse performance when the same data is accessed multiple times.

Wdyt?

malinajirka: > Performance?: I'm not 100% sure, but maybe in some cases, lazy parsing would cause worse performance when the same data is accessed multiple times.

Lazy parsing never parses it twice - it always parses it at most once, but only when it's needed and caches the result => so lazy parsing definitely provides a better performance. 

> Fail fast advantage: JSON parsing errors caught early during mapping, not during UI rendering

Having said that, I agree with the fail-fast advantage. Even though we might just want to consume parsing errors in these attributes anyway. But still feels slightly better. Let's keep this current approach and migrate to lazy if we notice performance issues.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModelVersion2MapperTest.kt
```diff
@@ -0,0 +1,350 @@
...
+
+    @Test
+    fun `when mapProductStatus is called with status strings, then returns correct product statuses`() {
+        assertThat(mapper.mapProductStatus("publish")).isEqualTo(WooPosProductModelVersion2.WooPosProductStatus.PUBLISH)
+        assertThat(mapper.mapProductStatus("PUBLISH")).isEqualTo(WooPosProductModelVersion2.WooPosProductStatus.PUBLISH)
+        assertThat(mapper.mapProductStatus("draft")).isEqualTo(WooPosProductModelVersion2.WooPosProductStatus.DRAFT)
+        assertThat(mapper.mapProductStatus("unknown")).isEqualTo(WooPosProductModelVersion2.WooPosProductStatus.UNKNOWN)
+    }
+
+    // Test helper methods
```
samiuelson:
```suggestion
```

samiuelson (approved):
LGTM!

PR #14567, opened 2025-09-04
POS Product Migration Step 4

malinajirka:

### Description

This is part of the migration to POS specific Product Model. There will be many changes and I want to avoid huge PRs. Therefore I created a feature branch - I'll create a PR for each set of changes. The CI won't be able to build this PR, but I personally wouldn't worry about it -> we can ensure everything is green when we are merging the feature branch back to trunk.

This PR migrates product cache/product provider/product data source/product index.

Let me know if you agree with this approach. Ideas are welcome.

The changes are pretty straight-forward in this case - we just change the reference and update the mapping.

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14569, opened 2025-09-04
[POS - Historical Orders] In-memory Cache

toupper:
### Description

Closes WOOMOB-1145

With this PR, we add an in-memory cache when fetching the POS Orders. The cache content is removed when leaving POS.

### Steps to reproduce

1.	Open the app.
2.	Navigate to POS.
3.	Go to Orders.
4.	Confirm that a loading state is shown while the orders are being fetched.
5.	Once the orders are fetched, they are displayed.
6.	Return to the POS home, then go back to Orders.
7.	Confirm that cached orders are shown immediately (no loading state). Once the new fetch completes, the list refreshes.
8.	Exit POS.
9.	Open POS again and go to Orders.
10.	Confirm that, since cached orders are cleared on exit, the loading state is shown first, followed by the fetched orders.

### Testing information

### Images/gif

https://github.com/user-attachments/assets/011e6152-fd81-43f9-aedc-cfca684d0440


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosOrdersInMemoryCache.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.woocommerce.android.ui.woopos.common.data.searchbyidentifier.WooPosOrdersCache.Companion.MAX_CACHE_SIZE
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import javax.inject.Inject
+
+class WooPosOrdersInMemoryCache @Inject constructor() : WooPosOrdersCache {
+    private val mutex = Mutex()
+
+    companion object {
+        private const val INITIAL_CAPACITY = MAX_CACHE_SIZE
```
kidinov: As the goal is to have this equal to the page size, I think we should reference the same constant used in WooPosOrdersDataSource and here.

Maybe adding `const val ORDERS_PAGE_SIZE = 25` to `WooPosOrdersDataSource.kt` and then

```
private const val INITIAL_CAPACITY = ORDERS_PAGE_SIZE
```
?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosOrdersInMemoryCache.kt
```diff
@@ -0,0 +1,41 @@
+package com.woocommerce.android.ui.woopos.common.data
```
kidinov: I think it's better to place it into `woopos.orders`. The product-related code here is because it's used in the list and the search.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosOrdersInMemoryCache.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val mutex = Mutex()
+
+    companion object {
+        private const val INITIAL_CAPACITY = MAX_CACHE_SIZE
+        private const val LOAD_FACTOR = 0.75f
+    }
+
+    private val ordersCache = LinkedHashMap<Long, Order>(INITIAL_CAPACITY, LOAD_FACTOR, true)
+
+    override suspend fun addAll(orders: List<Order>) = mutex.withLock {
```
kidinov: As discussed on the call, this should probably be `setAll` that just replaces all you have in the cache with the new list.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosOrdersInMemoryCache.kt
```diff
@@ -0,0 +1,41 @@
...
+
+class WooPosOrdersInMemoryCache @Inject constructor() : WooPosOrdersCache {
+    private val mutex = Mutex()
+
+    companion object {
+        private const val INITIAL_CAPACITY = MAX_CACHE_SIZE
+        private const val LOAD_FACTOR = 0.75f
+    }
+
+    private val ordersCache = LinkedHashMap<Long, Order>(INITIAL_CAPACITY, LOAD_FACTOR, true)
```
kidinov: Any specific reason to use hash map here? I think we can have just a list of orders here

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/root/WooPosActivity.kt
```diff
@@ -53,6 +58,13 @@ class WooPosActivity : AppCompatActivity() {
             }
         }
     }
+
+    override fun onDestroy() {
+        runBlocking {
```
kidinov: Not sure about that for 2 reasons:
* You are blocking the main thread here
* Overall, are you sure that we want to clean the cache here? Keep in mind that on Android, activity can be destroyed on configuration change. I wouldn't do that at all, especially since currently there are up to 25 orders in memory

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -30,25 +30,28 @@ class WooPosOrdersViewModel @Inject constructor(
...
-                            isLoading = false,
-                            error = result.message ?: "Unknown error"
-                        )
+            ordersDataSource.loadOrders().collect { result ->
+                when (result) {
+                    is LoadOrdersResult.Error -> {
+                        _state.update {
+                            it.copy(
+                                isLoading = false,
+                                error = result.message ?: "Unknown error"
```
kidinov: Just keep in mind to not show this error to a user, as this won't be localized

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -29,14 +36,16 @@ class WooPosOrdersDataSource @Inject constructor(
...
         } else {
-            LoadOrdersResult.Success(result.orders.toAppModels())
+            val mapped = result.orders.toAppModels()
+            ordersCache.addAll(mapped)
+            emit(LoadOrdersResult.Success(result.orders.toAppModels()))
         }
     }
 
-    private suspend fun List<OrderEntity>.toAppModels(): List<Order> = map {
+    private suspend fun List<OrderEntity>?.toAppModels(): List<Order> = this?.map {
```
kidinov: I’m a bit confused about this change. I thought the list here was always non-nullable, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: Ah, and btw `message` is never null so `?: "Unknown error"` not needed

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosOrdersCache.kt
```diff
@@ -0,0 +1,15 @@
+package com.woocommerce.android.ui.woopos.common.data.searchbyidentifier
+
+import com.woocommerce.android.model.Order
+
+interface WooPosOrdersCache {
```
kidinov: np: I think we don't really need an interface here. For the products, it was introduced because we were planning to migrate to persistent storage; therefore, we declared one interface with potential for two implementations. Here, at least for now, there are no plans for that, so maybe keeping it simple would be easier.

kidinov (approved):
LGTM. I left a few comments; please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/root/WooPosActivity.kt
toupper: Thanks for the comment @kidinov!

> You are blocking the main thread here

If we do it asynchronously, we lose the reference to the cache, so we end up not deleting the cache. 

> Overall, are you sure that we want to clean the cache here? Keep in mind that on Android, activity can be destroyed on configuration change. I wouldn't do that at all, especially since currently there are up to 25 orders in memory

What other place would be more suitable for this action?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Awesome, I removed it 👍

PR #14577, opened 2025-09-05
Add `cooldown` of Dependabot PRs for minor & major updates

wzieba:
Closes: AINFRA-1233

### Description

This PR introduces a 30-day cooldown period for minor & major updates, as discussed with the team internally (p91TBi-dpu-p2#comment-14339 .

The documentation for this feature is available here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

### Testing information

1. Create a fork
2. Downgrade a random dependency with minor version
3. Wait 30 days...
4. Just kidding :) This is just a configuration option, there's no way to meaningfully test it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,9 @@ updates:
           - "com.google.devtools.ksp*"
         exclude-patterns:
           - "org.jetbrains.kotlinx*"
+    cooldown:
+      semver-major-days: 30
+      semver-minor-days: 30
     ignore:
       # Bumping 2.26.3 to 2.27.2 will break the mocks. For more details, see
       # https://github.com/wiremock/wiremock/issues/1345#issuecomment-656060968


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/267">23.2</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

ParaskP7 (approved):
Awesome @wzieba , I am eager to try this out and see how it works! 🥇 ❤️ 🚀 

> Testing information
> - Create a fork
> - Downgrade a random dependency with minor version
> - Wait 30 days...
> ...

We really got me there, no kidding! 🤣

ParaskP7:
Btw, and just as an FYI @wzieba , my understanding of this feature is the below:

1. Instead, we can had `schedule:interval: "monthly"` instead, this would create an update for a dependency even if that dependency update was available a few days ago, as long as dependabot has waited a month, as per schedule.
2. While, now that we have `schedule:interval: "daily"` with `cooldown: semver-major/minor-days: 30`, it will daily try to create an update for a dependency, and when that `major/minor` dependency update get available, it will wait a month until it creates it.

Does the above sounds right to you? 🤔

PS: What I am not too clear about, is what will happen on this scenario below:

1. We get a `minor` dependency update, the `1.1.0`, say on `08/09/25`.
2. Dependabot with wait until `08/10/25` to create that update.
3. In the meanwhile, we get another `minor` dependency update on that same dependency, the `1.2.0`, say on `22/09/25`, 2 weeks after.
4. What will Dependabot do:
    1. Still wait until `08/10/25` to create that `1.1.0` update?
    2. Still wait until `08/10/25` but create the `1.2.0` update instead?
    3. Reset on `22/09/25` for that dependency update and wait util `22/10/25` to create the `1.2.0` update.

Do you have any idea? 🤔 (maybe I should have asked AI)

wzieba:
I assume it'll be `i.`, but the documentation doesn't specify details. Copilot (from GH docs) leans into option `ii.` but I can't make it quote the source of its claim :/.

wzieba:
Actually 

> After a cooldown ends for a dependency, Dependabot resumes updating the dependency following the standard update strategy defined in dependabot.yml.

And the "standard update strategy" is "newest version available". So in this case it'll be indeed `ii.`([source, point 5](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-))

ParaskP7:
Thanks for the reply @wzieba ! 🙇 

> And the "standard update strategy" is "newest version available". So in this case it'll be indeed ii.([source, point 5](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-))

Yes, I have the feeling that you're right on this and that this is what will happen, let's see... 🤷

I am just a bit confused atm at what is our actual goal with this, are you able to put that in a brief sentence, just so that we measure this for success, a bit later-on and don't forget what's the goal. For example: `Product engineers would like to hold-off on updating a dependency for 30 days since it was first published to increase confidence on that update` 🤔

FYI: I am asking all these questions because me re-reading this p91TBi-dpu-p2#comment-14339 comment, the product team want to setup a cooldown period to increase confidence in updates, which to me, seems that if `ii.` is what will happen this might produce a false-positive sense of confidence in updates, right? 🤔

PS: I don't expect that to happen too often, so not a big deal, but still something to keep in mind, right? 🤔

wzieba:
> which to me, seems that if ii. is what will happen this might produce a false-positive sense of confidence in updates, right? 🤔

Yes, it might, but only in a specific scenario. At the end of the day, if we merge this PR, we'll have:
- _some_ cooldown (from 30 days to minimum 1 day)
- no cooldown if a new release was released **just** after a cooldown period passed.

Currently, we don't have _any_ cooldown whatsoever.

So I think the `cooldown` introduce in this PR _does_ increase confidence of updates, achieving the goal described in P2.

---

Also, this is about minor & major versions - I really doubt that many of our dependencies update minor more than once per month. Let's look at the release cycle of a sample dependency: https://mvnrepository.com/artifact/androidx.lifecycle/lifecycle-viewmodel

ParaskP7:
Yea 💯 @wzieba I agree with ALL your argument and I am all-in, with `cooldown` versus without `cooldown` we'll anyway achieve that extra level of confidence that the product team needs, let's do it! :shipit:

wzieba:
Thanks for asking good questions and raising the concern! Let's see how the `cooldown` will work in practice 🙂

PR #14578, opened 2025-09-05
[POS Historical Orders] Pull to Refresh

toupper:
### Description

With this PR we implement Pull to Refresh on POS Orders.

### Steps to reproduce

1. Open the app
2. Go to POS
3. Go to Orders
4. Pull to Refresh
5. Verify that we request orders remotely and that we update the list

### Testing information

### Images/gif

https://github.com/user-attachments/assets/4781cab9-e2c5-40de-98db-1a517a8814c7


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -48,6 +53,7 @@ fun WooPosOrdersScreen(
     BackHandler { onNavigationEvent(WooPosNavigationEvent.GoBack) }
 
     Row(modifier = Modifier.fillMaxSize()) {
+        // Left pane
```
kidinov: NP: Instead of a comment, it is better to have a composable with a nice name that represents the left and right panes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -43,10 +44,12 @@ class WooPosOrdersDataSource @Inject constructor(
         } else {
             val mapped = result.orders.toAppModels()
             ordersCache.setAll(mapped)
-            emit(LoadOrdersResult.Success(result.orders.toAppModels()))
+            emit(LoadOrdersResult.SuccessRemote(result.orders.toAppModels()))
         }
     }
 
+    fun clearCache() = ordersCache.clear()
+
```
kidinov, line 54: Not related to the PR: on line 55 there is still  `?:` applied to a non-nullable type
image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -10,9 +10,10 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderRestClient.O
 import org.wordpress.android.fluxc.persistence.entity.OrderEntity
 import javax.inject.Inject
 
-sealed class LoadOrdersResult {
-    data class Success(val orders: List<Order>) : LoadOrdersResult()
-    data class Error(val message: String) : LoadOrdersResult()
+sealed interface LoadOrdersResult {
```
kidinov: Any reason specifically for `interface` here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -59,59 +65,98 @@ fun WooPosOrdersScreen(
...
-                    ) {
-                        WooPosText(
-                            text = stringResource(R.string.loading),
-                            style = WooPosTypography.BodyMedium,
-                            color = MaterialTheme.colorScheme.onSurfaceVariant,
-                            modifier = Modifier.padding(WooPosSpacing.Large.value)
-                        )
-                    }
+            val refreshing =
+                when (val s = state) {
```
kidinov: It’s best to keep the logic separate from the view. The view should simply display the data without any complex calculations, the logic should be in the VM. Also, you seem to ignore disabled state here.

So maybe something like that would be cleaner:

```
            val pullRefreshState = rememberPullRefreshState(
                refreshing = state.pullToRefreshState == Refreshing,
                onRefresh = { viewModel.refresh() },
            )

            Box(
                modifier = Modifier
                    .fillMaxSize()
                    .pullRefresh(
                        pullRefreshState,
                        enabled = state.pullToRefreshState != WooPosPullToRefreshState.Disabled
                    )
            ) {
```


```
@Immutable
sealed class WooPosOrdersState(
    open val pullToRefreshState: WooPosPullToRefreshState,
) {
    @Immutable
    data class Content(
        val items: List<OrderItemViewState>,
        override val pullToRefreshState: WooPosPullToRefreshState,
        val paginationState: WooPosPaginationState,
        val selectedOrderId: Long?
    ) : WooPosOrdersState(pullToRefreshState)

    @Immutable
    data class Error(
        val message: String,
        override val pullToRefreshState: WooPosPullToRefreshState = WooPosPullToRefreshState.Disabled,
    ) : WooPosOrdersState(pullToRefreshState)

    @Immutable
    data class Loading(
        override val pullToRefreshState: WooPosPullToRefreshState = WooPosPullToRefreshState.Disabled,
    ) : WooPosOrdersState(pullToRefreshState)

    @Immutable
    data class Empty(
        override val pullToRefreshState: WooPosPullToRefreshState = WooPosPullToRefreshState.Enabled,
    ) : WooPosOrdersState(pullToRefreshState)
}
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: The same applies to other comments here; it appears to be a workaround for a decent composition of the code. And some maybe don't have value, e.g. 

```
              // PTR indicator at the top of the list area
                PullRefreshIndicator(
```
I think that's a redundant comment

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -15,46 +18,102 @@ class WooPosOrdersViewModel @Inject constructor(
...
 ) : ViewModel() {
 
-    private val _state = MutableStateFlow(WooPosOrdersState())
+    private val _state = MutableStateFlow<WooPosOrdersState>(WooPosOrdersState.Loading)
     val state: StateFlow<WooPosOrdersState> = _state.asStateFlow()
 
-    init {
-        refreshOrders()
-    }
+    private var orders: List<Order> = emptyList()
```
kidinov: Why to save orders in 2 places? The same about selectedId. You have to have 1 source of the truth - `_state` variable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -15,46 +18,102 @@ class WooPosOrdersViewModel @Inject constructor(
...
+        return current?.takeIf { id -> newOrders.any { it.id == id } }
+            ?: newOrders.firstOrNull()?.id
+    }
 }
+
+private fun List<Order>.toOrderItems(selectedId: Long?): List<OrderItemViewState> =
+    map { o ->
+        OrderItemViewState(
+            id = o.id,
+            title = "Order #${o.number}",
```
kidinov: Localization?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: It feels that all this code can be simpler: 

```

@HiltViewModel
class WooPosOrdersViewModel @Inject constructor(
    private val ordersDataSource: WooPosOrdersDataSource
) : ViewModel() {

    private val _state = MutableStateFlow<WooPosOrdersState>(WooPosOrdersState.Loading())
    val state: StateFlow<WooPosOrdersState> = _state.asStateFlow()

    init {
        loadOrders()
    }

    fun onOrderSelected(orderId: Long) {
        val currentState = _state.value
        if (currentState is WooPosOrdersState.Content) {
            _state.value = currentState.copy(
                items = currentState.items.map { it.copy(isSelected = it.id == orderId) },
                selectedOrderId = orderId
            )
        }
    }

    fun refresh() {
        val currentState = _state.value
        _state.value = when (currentState) {
            is WooPosOrdersState.Content -> currentState.copy(
                pullToRefreshState = WooPosPullToRefreshState.Refreshing
            )

            is WooPosOrdersState.Empty -> currentState.copy(
                pullToRefreshState = WooPosPullToRefreshState.Refreshing
            )

            is WooPosOrdersState.Error -> currentState
            is WooPosOrdersState.Loading -> currentState
        }

        ordersDataSource.clearCache()
        loadOrders()
    }

    private fun loadOrders() {
        viewModelScope.launch {
            ordersDataSource.loadOrders().collect { result ->
                when (result) {
                    is LoadOrdersResult.Error -> {
                        _state.value = WooPosOrdersState.Error(
                            message = result.message,
                            pullToRefreshState = WooPosPullToRefreshState.Disabled
                        )
                    }

                    is LoadOrdersResult.SuccessCache -> {
                        updateContentState(result.orders)
                    }

                    is LoadOrdersResult.SuccessRemote -> {
                        if (result.orders.isEmpty()) {
                            _state.value = WooPosOrdersState.Empty()
                        } else {
                            updateContentState(result.orders)
                        }
                    }
                }
            }
        }
    }

    private fun updateContentState(orders: List<Order>) {
        val currentSelectedId = (_state.value as? WooPosOrdersState.Content)?.selectedOrderId
        val newSelectedId = currentSelectedId?.takeIf { id -> orders.any { it.id == id } }
            ?: orders.firstOrNull()?.id

        _state.value = WooPosOrdersState.Content(
            items = orders.map { order ->
                OrderItemViewState(
                    id = order.id,
                    title = "Order #${order.number}",
                    date = order.dateCreated.formatToDDMMMYYYY(),
                    total = "${order.total} ${order.currency}",
                    isSelected = order.id == newSelectedId
                )
            },
            selectedOrderId = newSelectedId,
            pullToRefreshState = WooPosPullToRefreshState.Enabled,
            paginationState = WooPosPaginationState.None
        )
    }
}

```

I think it make sense if `WooPosOrdersDataSource` won't return `SuccessCache` if there is nothing in the cache as well:


```
        val cached = ordersCache.getAll()
        if (cached.isNotEmpty()) {
            emit(LoadOrdersResult.SuccessCache(cached))
        }
```

kidinov (commented):
Thanks @toupper !

I left a few comments! Please take a look. Especially I think we should not have 2 different states in the view model as it source of many bugs having to sync them.

Also, it’s generally not a good idea to use `when` with `else` in a sealed class. This can make it harder to catch bugs when a new type is added, as the compiler won't complain

I didn't test the implementation yet, as I want to see what you think about my comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Thanks for the hints. Done in e4d24ab8347ce91ce1a18e9bb3643522620464db

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: True, removed in 32927808d5320b9bd8162804196d4ce83a59a0c3

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Not a big reason, changed back to sealed class as that's the convention in our code 43400830f82f9cb240ef33530fd419b28144c022

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Enabled state in a5378e88e50e3607234486dc34d07d9a89e5dbd8

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Awesome, done in 73bec2d154eac7213f66d861944d44fd2dac8650

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: As we talked in other PRs, the UI is still dummy. I would prefer that we add localizable strings when implementing the design, so we don't have to change it afterwards.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -48,19 +53,59 @@ fun WooPosOrdersScreen(
     BackHandler { onNavigationEvent(WooPosNavigationEvent.GoBack) }
 
     Row(modifier = Modifier.fillMaxSize()) {
-        Column(
+        WooPosOrdersLeftPane(
+            state = state,
+            onBackClicked = onBackClicked,
+            onRefresh = viewModel::refresh,
+            isRefreshing = viewModel.isRefreshing(),
```
kidinov: The issue here is that `isRefreshing` is a function call that derives state from the existing state flow, rather than being part of the state itself. This violates the principle of having a single source of truth for UI state. So basically during composition, in some moment you go to the VM and ask for another piece of state, and there is no guarantee that it will  still be the same state that you are rendering at the moment, which may end up with inconsistent state being displayed to the user

Something like that would solve this potential issue:

[Refactor_pull-to-refresh_logic_in_WooPosOrders_components.patch](https://github.com/user-attachments/files/22248415/Refactor_pull-to-refresh_logic_in_WooPosOrders_components.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -15,46 +18,91 @@ class WooPosOrdersViewModel @Inject constructor(
...
     }
 
-    fun refreshOrders() {
-        viewModelScope.launch {
-            _state.update { it.copy(isLoading = true, error = null) }
+    fun isRefreshing(): Boolean =
+        (state.value as? WooPosOrdersState.Content)
+            ?.pullToRefreshState == WooPosPullToRefreshState.Refreshing
+
+    fun refresh() {
```
kidinov: np: `onRefresh` would follow the convention when the functions are called as a reaction to the user’s action

kidinov (approved):
LGTM! Before merging, please take a look at the comments I left especially the one about `isRefreshing` method

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Thanks for that @kidinov! I applied it in 7cf9f9380dffd2b9e5e531eb349b3b0c710da29b

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Good point, done in ac705b3e7f718a538aeaa02e6a88e07f62b6ed97

PR #14585, opened 2025-09-05
[Jetpack Performance] Refresh site status periodically

hichamboushaba:
Closes WOOMOB-1182

### Description
This PR adds logic to make sure we refresh site status periodically, it implements the approach suggested here pe5sF9-4ye-p2#comment-4801:
- Store the timestamp when flagging a website as not supporting app passwords.
- When the timestamp is older than 7 days, we ignore it.

### Steps to reproduce
Follow the TC7 test steps from here pe5sF9-4Am-p2

### Testing information
- Confirm that after the duration, we check the site again, and we can use app passwords.

The above.

### DIFF
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
@@ -18,6 +18,7 @@ class JetpackApplicationPasswordsSupport @Inject constructor(
 ) {
     private val fluxCPreferences by lazy { PreferenceUtils.getFluxCPreferences(context) }
 
+    @Synchronized
     fun supportsAppPasswords(siteModel: SiteModel): Boolean {
         val siteFlag = fluxCPreferences.getStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, null)
             ?.firstOrNull { it.startsWith("${siteModel.siteId}:") }
@@ -38,6 +39,7 @@ class JetpackApplicationPasswordsSupport @Inject constructor(
         return siteModel.isApplicationPasswordsSupported
     }
 
+    @Synchronized
     fun flagAsUnsupported(siteModel: SiteModel) {
         val unsupportedSites = fluxCPreferences.getStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, null) ?: setOf()
         if (unsupportedSites.any { it.startsWith("${siteModel.siteId}:") }) {
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
@@ -64,7 +64,7 @@ class JetpackApplicationPasswordsSupport @Inject constructor(
 
     companion object {
         @VisibleForTesting
-        internal const val FLAG_REFRESH_DURATION_DAYS = 7
+        internal const val FLAG_REFRESH_DURATION_DAYS = 14
 
         @VisibleForTesting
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
```diff
@@ -1,28 +1,72 @@
...
         fluxCPreferences.edit {
-            putStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, unsupportedSites + siteModel.siteId.toString())
+            putStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, updatedSites)
         }
     }
 
     companion object {
-        private const val UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES = "unsupported_jetpack_app_passwords_sites"
+        @VisibleForTesting
+        internal const val FLAG_REFRESH_DURATION_DAYS = 7
```
hichamboushaba: The duration is not confirmed yet, I will confirm it with Huong, and update if needed in a separate PR.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
```diff
@@ -1,28 +1,72 @@
...
     fun supportsAppPasswords(siteModel: SiteModel): Boolean {
-        return siteModel.isApplicationPasswordsSupported &&
-            fluxCPreferences.getStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, null)
-                ?.contains(siteModel.siteId.toString()) != true
+        val siteFlag = fluxCPreferences.getStringSet(UNSUPPORTED_JETPACK_APP_PASSWORDS_SITES, null)
+            ?.firstOrNull { it.startsWith("${siteModel.siteId}:") }
+
+        if (siteFlag != null) {
+            val flagTimestamp = siteFlag.substringAfter(":").toLongOrNull() ?: 0L
+            if (System.currentTimeMillis() - flagTimestamp < FLAG_REFRESH_DURATION_DAYS.days.inWholeMilliseconds) {
```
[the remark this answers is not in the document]
hichamboushaba: I think using `0L` in this case is still accepted, we just want to handle migration for entries added without the timestamp, using `0L` will cause the flag to be removed in next lines.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordsSupport.kt
hichamboushaba: We agreed to increase this to 14 days pe5sF9-4ye-p2#comment-4815, this was implemented in 7e3bc60

JorgeMucientes (approved):
Good job! It works as expected @hichamboushaba. 

However, one remark about the testing instructions for our future selves when beta testing this: 
Instructions tell to test TC7 from pe5sF9-4Am-p2. TC7 point to use TC3 snippet in order to mark the site as application passwords unsupported. However, with the current logic and the provided snippet in TC3 the site won't be flagged as unsupported because the API error returned with TC3 snippet is `application_passwords_disabled` whereas the app expects `application_passwords_disabled_for_user`: 

Image

In the real case scenario, what would WP API return? `application_passwords_disabled_for_user` or `application_passwords_disabled`. If the former, then we just need to update the TC3 snippet to fix the testing instructions; if the latter, then we should fix the issue in the code asap. Also why not match agains the `httpStatusCode` 501 instead of matching against `apiErrorCode` ?

hichamboushaba:
Good eye @JorgeMucientes, yes this is an issue, I already fixed it in my next PR in this commit https://github.com/woocommerce/woocommerce-android/commit/81de96d7b7ad3f4098e6448760b303b9f4e05038, and updated the error handling plan to include both error codes: pe5sF9-4ye-p2#comment-4787

PR #14588, opened 2025-09-08
POS Product Migration Step 5

malinajirka:

### Description

This PR migrates WooPosProductsViewModel and a couple search related classes.

### Images/gif

N/A


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/util/WooPosProductUtils.kt
```diff
@@ -0,0 +1,31 @@
+package com.woocommerce.android.ui.woopos.util
```
kidinov, line 1: Should we name the file with “test” in it to avoid cluttering the main namespace when searching for products-related files in the production code? `WooPosProductTestUtil` maybe

kidinov (approved):
LGTM! A lot of nice ideas with this model 👏

I left one np - feel free to merge!

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/util/WooPosProductUtils.kt
malinajirka: Good idea, I've done this in the last PR in the chain - 44471ed4acd807825468a7434a828c31c25144ea.

PR #14589, opened 2025-09-08
POS Product Migration Step 6

malinajirka:
Step 5 needs to be merged first.

### Description

This PR migrates WooPosCartItemsUpdater, WooPosSearchByIdentifierProcessVariationResult and WooPosSearchByIdentifierResultConverter.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierResultConverter.kt
```diff
@@ -15,7 +15,7 @@ class WooPosSearchByIdentifierResultConverter @Inject constructor(
         return when (result) {
             is WooPosSearchByIdentifierResult.Success -> {
                 val product = result.product
-                if (product.type.equals("variation", ignoreCase = true)) {
+                if (product.type.value.equals("variation", ignoreCase = true)) {
```
kidinov: Shell it be `product.type == WooPosProductModelVersion2.WooPosProductType.VARIATION` ?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartItemsUpdater.kt
```diff
@@ -108,11 +109,26 @@ class WooPosCartItemsUpdater @Inject constructor(
         updatedItem: WooPosCartItemViewState.Product,
         updatedProduct: ParentToChildrenEvent.OrderCreated.ProductInfo
     ) {
+        // TBD Local Catalog The app should update cache based on the data coming from backend not from the view layer
```
kidinov, line 112: Is the data coming here from the view layer?

kidinov (approved):
LGTM! I left a couple of minor comments here

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartItemsUpdater.kt
malinajirka: Yes, or at least we are receiving `WooPosCartItemViewState.Product` which is a ViewState.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/searchbyidentifier/WooPosSearchByIdentifierResultConverter.kt
malinajirka: Fixed in f5884ec487f2902f352ebac3ea40f1fa5b7b85d7

PR #14590, opened 2025-09-08
POS Product Migration Step 7

malinajirka:
Step 6 needs to be merged first.

### Description

This PR migrates [WooPosCartViewModel](https://github.com/woocommerce/woocommerce-android/commit/b9ed82452b0119c5fe8e0c8c16e0c12ea3dbff1e), [WooPosVariationMapper](https://github.com/woocommerce/woocommerce-android/commit/893a50c59b6ddca5ee2635fe3709b4790fa4a6f0), [WooPosProductSearchPredicate](https://github.com/woocommerce/woocommerce-android/commit/3cea6b5f17a74a44340a3aa5245e89cda645709a), [WooPosSearchProductsDataSource](https://github.com/woocommerce/woocommerce-android/commit/50ce0d4042e285f351f39fb30dfa4106b393fcf5),[WooPosSearchResultsIndex](https://github.com/woocommerce/woocommerce-android/commit/cd121007e8b25b9cd5739d86ed680bf9b259b114)

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModelTest.kt
```diff
@@ -72,6 +74,13 @@ class WooPosCartViewModelTest {
     private val parentToChildrenEventReceiver: WooPosParentToChildrenEventReceiver = mock {
         on { events }.thenReturn(parentToChildrenMutableSharedFlow)
     }
+
+    private val mockedEventForTracking: WooPosAnalyticsEvent.Event.ItemAddedToCart =
```
kidinov, line 78: I’m curious, why do we need this one? And if we do, why is it called “mocked” when it seems to be the real object?

kidinov (approved):
LGTM! I made a comment about the changes in the test

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModelVersion2.kt
```diff
@@ -30,6 +30,9 @@ data class WooPosProductModelVersion2(
     val variationIds: List<Long> = emptyList(),
 ) : Parcelable {
 
+    val variationEnabledAttributes
+        get() = attributes.filter { it.isVariation }
```
kidinov: It's an immutable data class, so do we need to refilter the list every time when it's accessed, or maybe it will be more efficient to have it as a field?

```
 @IgnoredOnParcel
  val variationEnabledAttributes = attributes.filter { it.isVariation }
```
?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModelTest.kt
malinajirka: Detekt wouldn't let me keep the mocked data object - there is a relatively new rule, that data classes can't be mocked. I kept the word `mocked` in the name to highlight, the object is kind of random - isn't actually testing which event we are passing/getting, just that it's a WooPosAnalyticsEvent.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModelVersion2.kt
malinajirka: Good point, I'll create a lazily calculated property - that gives us the best of both worlds.

malinajirka: Done in 4bb389f4852243059ef5e859933399974b16843d

PR #14595, opened 2025-09-09
[WOOMOB-1309] Fix WooPos local catalog API response parsing

malinajirka:
Fixes WOOMOB-1309

### Description
This PR fixes API response parsing issues in the WooPos local catalog functionality. The changes include updating the WooPosSyncTimestampManager to handle different timestamp formats properly and fixing a typo in query parameters that was affecting API responses.

### Testing information
- Green CI is enough for this PR, since the code isn't used yet.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
@@ -2,10 +2,10 @@ package com.woocommerce.android.ui.woopos.util.datastore
 
 import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
 import java.time.Instant
-import java.time.LocalDateTime
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeParseException
+import java.time.temporal.TemporalAccessor
 import java.util.Locale
 import javax.inject.Inject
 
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
```diff
@@ -37,24 +36,41 @@ class WooPosSyncTimestampManager @Inject constructor(
...
+        return null
     }
 
     private companion object {
-        const val GMT_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+        private const val DEFAULT_API_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+        private val defaultApiDateFormatter = DateTimeFormatter.ofPattern(DEFAULT_API_DATE_FORMAT, Locale.US)
+            .withZone(ZoneOffset.UTC)
+
+        private val PARSING_FORMATTERS = listOf(
```
kidinov, line 68: So the API can return the date in different formats, and is this kind of expected, or is this a fallback in cases when plugins mess things up? Just curious

kidinov (approved):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampManager.kt
malinajirka: Good question! 

I came up with this solution when I saw a parsing error on one of my sites. However, exploring this more, I realized the issue is that we are parsing dates from two different sources - header and body. The issue isn't that a site/plugin would change the date, but that the format we use in headers differs from the format we use in body.

I still believe having this fallback might be a good idea, just to be sure. However, I will create a ticket to ensure we always use the most probable format based on the location/source of the date.

https://linear.app/a8c/issue/WOOMOB-1325/woo-poslocal-catalog-date-parsing

PR #14596, opened 2025-09-09
[WOOMOB-1206] Schedule WooPos local catalog sync on installation and run

malinajirka:
Fixes WOOMOB-1206

⚠️ [This issue needs to be fixed before we merge this PR](https://linear.app/a8c/issue/WOOMOB-1324/woo-poslocal-catalog-the-database-gets-locked-when-we-sync-pos).

### Description
This PR implements scheduled synchronization of the WooPos local catalog on app installation and startup. The implementation includes:

- Schedule nightly local catalog sync on app startup.
- Manually trigger sync with a small delay on site change.
- Skip local catalog sync when the feature flag is disabled

Known issue - [localSiteId is always set to 0 in the PosProduct table.](https://linear.app/a8c/issue/WOOMOB-1312/woo-poslocal-catalog-localsiteid-is-always-0)

Note: This will start the worker even before we release the project, since the FF check is done in the worker. However, I think that's fine - it'll allow us to easily add a remote feature flag, which might not be initialized on app startup.

### Testing information
- Verify the periodic job is scheduled on app startup - use App Inspection -> Background Task Inspector in AS
- Verify manual sync is invoked after login
- Verify manual sync is invoked on site switch
- Verify the data are available in the database - use App Inspection -> Database Inspector in AS

- The bbove

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosLocalCatalogSyncWorker</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
```diff
@@ -92,6 +94,7 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
         private const val SECONDS_BETWEEN_SITE_UPDATE = 60 * 60 // 1 hour
         private const val UNAUTHORIZED_STATUS_CODE = 401
         private const val CARD_READER_USAGE_THIRTY_DAYS = 30
+        private const val POS_LOCAL_CATALOG_SYNC_INITIAL_DELAY_SECONDS = 20L
```
malinajirka: 📓 I added this to ensure we don't do too many requests right after login/site-switch. What do you think, does it make sense/is it worth it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -7,17 +7,20 @@ import androidx.work.WorkerParameters
 import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.login.AccountRepository
 import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
+import com.woocommerce.android.ui.woopos.featureflags.WooPosLocalCatalogM1Enabled
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 
 @HiltWorker
+@Suppress("LongParameterList")
```
kidinov, line 17: I believe the intention was to suppress the constructor, not the entire class?

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
```diff
@@ -424,6 +435,19 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
         }
     }
 
+    fun triggerLocalCatalogSyncOnSiteChange() {
+        appCoroutineScope.launch {
+            selectedSite.observe()
+                .drop(1) // invoke only on site change not on app initialization
+                .collect { selectedSite ->
+                    if (selectedSite != null) {
+                        delay(POS_LOCAL_CATALOG_SYNC_INITIAL_DELAY_SECONDS)
```
kidinov, line 444: It's 20 ms (not seconds) delay here

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
```diff
@@ -238,6 +244,9 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
 
         // Schedule worker to refresh FCM token periodically
         FCMRefreshWorker.schedule(application)
+
+        posLocalCatalogScheduler.schedulePeriodicFullCatalogSync()
+        triggerLocalCatalogSyncOnSiteChange()
```
kidinov: The naming could be a bit confusing because you’re not triggering it here but rather subscribing to changes to trigger it later. Maybe `observeSiteChangesForCatalogSync`?

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
kidinov: I think it's a good idea

kidinov: I actually can always reproduce now that if we start the sync right away (as now it's 20 ms), the store switch is blocked by that. So I have to wait till `local catalog sync completed successfully. Products: 531, Variations: 80, Duration: 14393ms` appears in the logs, and only then the switch happens.

I conducted another experiment and changed the delay to 20s. The store has approximately 500 products. And after 20s, during the sync process, the entire app is blocked network-wise. In my case, the orders were completely blocked till sync finished. So I guess we need to have some local throttling for the sync?

kidinov (approved):
LGTM! I left a few minor comments, but some of them are important; please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
malinajirka: Shoot I think I know what is going on 😢 ... it's not about network, it's about the db transaction -> it prevents all writes across the whole database.

I think we might need to store the network results into a file and create a transaction only when we have all the data locally. We might also consider splitting POS into it's own database, so it doesn't impact the rest of the app.

Another alternative is not to worry about transaction for the very first sync - since we can simulate transaction by wiping the data.

kidinov: Oh wow. I thought it was supposed to lock on a table, not the whole database. 😬

Another option could be to implement transactions on our side by remembering product IDs that were inserted and rolling them back (deleting by the IDs) manually if something goes wrong. 

Not related to this: in `WooPosLocalCatalogSyncRepository` the classes still called with `Pos...`. Looks like the file was renamed but the classes not

malinajirka: > Oh wow. I thought it was supposed to lock on a table, not the whole database. 😬

This is one of the drawbacks of sqlite compared to mysql and similar.


> Another option could be to implement transactions on our side by remembering product IDs that were inserted and rolling them back (deleting by the IDs) manually if something goes wrong.

That's impossible to do safely I believe - the app might crash for example. Moreover, some rows might be updated and we won't know which column to revert and what was the previous value.

> Not related to this: in WooPosLocalCatalogSyncRepository the classes still called with Pos.... Looks like the file was renamed but the classes not

👍 I'll update that.

malinajirka: > Not related to this: in WooPosLocalCatalogSyncRepository the classes still called with Pos.... Looks like the file was renamed but the classes not

Fixed in c01050df7718b6ba383f16f6f6fcc1fcb35c6ba6

> I conducted another experiment and changed the delay to 20s. The store has approximately 500 products. And after 20s, during the sync process, the entire app is blocked network-wise. In my case, the orders were completely blocked till sync finished. So I guess we need to have some local throttling for the sync?

I've created an issue for this, since it won't be a trivial change - we need to update the store layer to return the response or move the syncing logic into the store layer. Issue here - https://linear.app/a8c/issue/WOOMOB-1324/woo-poslocal-catalog-the-database-gets-locked-when-we-sync-pos. I also added a `do not merge` label to this PR, so we merge both PRs at the same time. It wouldn't be a big deal, since this code is run only when the FF is enabled, but still blocking the whole database for tens of seconds might not be great for other developers who work on trunk.

-----
I also noticed we are storing timestamp of the server date of the last request instead of the first request, updated in 88d8eabf2c0b87aa1aa9f3c2817c76cc5230358f.

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
malinajirka: > (as now it's 20 ms)


🤦  Good catch, fixed in 34f8dcd1a944cba95786ed0068fe3a041f6af897.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
malinajirka: Fixed in c930f54320353f02f0aac7834808e96c576955f9

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
malinajirka: Good idea, update in ba7111ea0a06b2793391e40a503cb1cfcba0db6d

PR #14600, opened 2025-09-10
[Jetpack Performance] Tracking

hichamboushaba:
Closes WOOMOB-1321

### Description
This PR adds tracking for the Jetpack performance project, following the plan agreed on pe5sF9-4ye-p2#comment-4787, the PR has the following changes:

- Removes the property request_type from the event `order_list_load_error`, this was added in the experiment, but we don't need it anymore.
- Add the new event `woocommerceandroid_jetpack_site_eligible_for_app_password_support`
- Add the new event `woocommerceandroid_jetpack_site_flagged_unsupported_for_app_passwords`

The last commit of the PR c1c7334118214759cba08d1ace899a2c0ccc5544 is not related to the tracking changes, it just adds a code error that I missed when working on error handling (pe5sF9-4ye-p2#comment-4820)

### Steps to reproduce
I'm not adding testing steps for the `woocommerceandroid_orders_list_loaded` event, as this behavior was already in the app and reviewed in this PR https://github.com/woocommerce/woocommerce-android/pull/13919.

For the new events, please follow these steps:

1. Use a site that supports App Passwords.
2. Sign in to the app using a WordPress.com account.
3. After login, confirm the event `woocommerceandroid_jetpack_site_eligible_for_app_password_support` was tracked (check Logcat for Tracked: ... entries)
4. Restart the app, and confirm the event `woocommerceandroid_jetpack_site_eligible_for_app_password_support` is tracked on launch.
5. Install this [plugin](https://wordpress.org/plugins/disable-application-passwords/), or apply the snippet from TC4 here pe5sF9-4Am-p2#tc4-app-passwords-feature-is-broken-case-2
6. **Without restarting the app** (this is needed if you installed the plugin), refresh some screens (to trigger network requests)
7. Confirm the event `woocommerceandroid_jetpack_site_flagged_unsupported_for_app_passwords` is tracked, with the properties `"flow":"app_password_generation", "cause":"major_error"`

### Testing information
Confirm the events are tracked correctly.

The above


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/ApplicationPasswordsNotifier.kt
```diff
@@ -66,6 +67,18 @@ class ApplicationPasswordsNotifier @Inject constructor(
...
     }
 
+    override fun onJetpackSiteFlaggedAsUnsupported(event: JetpackSiteFlaggedAsUnsupported) {
+        analyticsTrackerWrapper.track(
+            stat = AnalyticsEvent.JETPACK_SITE_FLAGGED_UNSUPPORTED_FOR_APP_PASSWORDS,
+            properties = mapOf(
+                AnalyticsTracker.KEY_FLOW to event.flow.name.lowercase(),
+                AnalyticsTracker.KEY_CAUSE to event.cause.name.lowercase(),
+                "api_error_code" to event.apiErrorCode,
+                "http_status_code" to event.httpStatusCode,
```
hichamboushaba: I didn't define these new keys in `AnalyticsTracker`, personally, I think adding all key values as constants there is not a scalable solution, but please let me know if you think I should add them.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -516,11 +516,13 @@ class SitePickerViewModel @Inject constructor(
...
                 launch {
                     val siteVerificationResult = repository.verifySiteWooAPIVersion(it.site)
+                    val siteVerificationModel = siteVerificationResult.model
                     when {
                         siteVerificationResult.isError -> onSiteVerificationError(siteVerificationResult, it)
-                        siteVerificationResult.model?.apiVersion == WooCommerceStore.WOO_API_NAMESPACE_V3 -> {
+                        siteVerificationModel?.apiVersion == WooCommerceStore.WOO_API_NAMESPACE_V3 -> {
                             experimentTracker.log(ExperimentTracker.SITE_VERIFICATION_SUCCESSFUL_EVENT)
-                            selectedSite.set(it.site)
+                            selectedSite.set(siteVerificationModel.siteModel)
```
hichamboushaba, line 524: This is not related to this PR, and for now it doesn't have any impact because `SiteModel` is a mutable class, and after the fetch we in `verifySiteWooAPIVersion` we modify the values in the same instance.

But given the goal to migrate `SiteModel` to Room, and that it could mean refactoring `SiteModel` and making it `immutable`, changes like this will be needed, so I'm handling this part now 🙂 .

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackSiteFlaggedAsUnsupported.kt
```diff
@@ -0,0 +1,19 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPINetworkError
+
+data class JetpackSiteFlaggedAsUnsupported(
+    val flow: Flow,
```
AdamGrzybkowski: Nitpick, but my first thought after reading this line was "Why do we need a Flow in a data class here?". Only later, I saw that this is an internal enum name. Given this is pretty isolated here, I'm okay with it, but maybe it could be changed to something else that won't be confused with Kotlin's Flow. `ConnectionFlow` maybe?

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/ApplicationPasswordsNotifier.kt
```diff
@@ -66,6 +67,18 @@ class ApplicationPasswordsNotifier @Inject constructor(
...
     }
 
+    override fun onJetpackSiteFlaggedAsUnsupported(event: JetpackSiteFlaggedAsUnsupported) {
+        analyticsTrackerWrapper.track(
+            stat = AnalyticsEvent.JETPACK_SITE_FLAGGED_UNSUPPORTED_FOR_APP_PASSWORDS,
+            properties = mapOf(
+                AnalyticsTracker.KEY_FLOW to event.flow.name.lowercase(),
+                AnalyticsTracker.KEY_CAUSE to event.cause.name.lowercase(),
+                "api_error_code" to event.error.errorCode,
+                "http_status_code" to event.error.volleyError?.networkResponse?.statusCode,
```
AdamGrzybkowski: ❓ Do we have any rules about when we decide to extract a fixed value to a property?

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/ApplicationPasswordsNotifier.kt
AdamGrzybkowski: Got it. I agree that defining keys for use in one place is not scalable. I would prefer to have `AnalyticEvent` sealed classes where each one would define whatever key props it needs. 

On the other hand, if that's how this code base is doing it, maybe it's worth sticking to it until there's an agreed alternative.

AdamGrzybkowski (commented):
Code changes look good to me. I've left some comments but nothing blocking. 

Before I approve, can you please tell me if we have, and if yes, where can I find a spreadsheet with events? The question is coming from the Danger comment:

>Verify the internal Tracks spreadsheet has also been updated.

I've verified the events were registered for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/ApplicationPasswordsNotifier.kt
AdamGrzybkowski: I saw your comment after, so I think this can be ignored.

hichamboushaba:
> Before I approve, can you please tell me if we have, and if yes, where can I find a spreadsheet with events? The question is coming from the Danger comment:
> 
> > Verify the internal Tracks spreadsheet has also been updated.

Thanks @AdamGrzybkowski for the review.

Actually we stopped adding to the spreadsheet a long time ago p91TBi-aNE-p2#comment-11755, I think we just missed updating the Danger rule here.

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/ApplicationPasswordsNotifier.kt
hichamboushaba: > On the other hand, if that's how this code base is doing it, maybe it's worth sticking to it until there's an agreed alternative

Done in b9a7dbce28df314679d2ff70a0e8c6162dc45b8f

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackSiteFlaggedAsUnsupported.kt
hichamboushaba: I renamed it to `Scenario` in b4dc01c0d5a76c9960655f6c3c495b453032ab87

AdamGrzybkowski (approved):

PR #14602, opened 2025-09-11
[WOOMOB-28] Use product details for default Blaze campaign values

AdamGrzybkowski:
Closes WOOMOB-28

### Description

When the request to get the Blaze campaign values fails, the Android app shows an empty Ad preview. We can prepopulate some fields based on the product details in such cases. 

From the issue description:
- Use product name for ad title
- Use product short description if available or Description for ad description.

This was done in https://github.com/woocommerce/woocommerce-android/commit/aa3e0e55fdc5cec81b85c2024fba418a2b05884b

Additionally, [I looked at iOS and noticed they do shows the product image without the network](https://github.com/woocommerce/woocommerce-ios/pull/14868). The Android app logic didn't allow for that, becuase there's an extra check if the image is valid (certain minimum size), and without the internet the image couldn't be loaded.

In https://github.com/woocommerce/woocommerce-android/commit/9e272ef30cc40aa81742d0c3d0e10c820fb286e9 I've added an extra check to use an already cached image as a backup. 

It uses some experimental Coil functions, but I think this is mostly because we use a pretty old Coil version. Also, I'm not entirely sure if that is enough given the app uses Glide as well.

Alternatively, we could remove the `isValidAdImage()` check from the `BlazeRepository.generateDefaultCampaignDetails()`. After all, this is just a default, which should be treated as backup, so maybe not doing this check is better than not displaying an image at all. 

Thoughts?

One more thing:

The iOS app sets the default `ctaText` as `Shop now`. Due to where the `CampaignDetails` is created, we can't do that. We would have to pass the translated string from the View. But given that this text should match the language of the shop, I'm not sure that using the device's language is the way to go. At this point, I kept this as empty.

### Steps to reproduce

1. Open the app
2. Go to products
3. Tap on a product
5. Disable the internet connection
6. Tap "Promote with Blaze"
7. Tap "Start your campaign"

### Testing information

Follow the steps above with different products, some with short description and some without.

I performed the above tests with different products.

### Images/gif
| Before | After |
|---|---|
| Screenshot_20250911_123736 | Screenshot_20250911_123207 |

Technically, this is visible, but it's a "backup" so I'm not sure it's worth mentioning.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -101,7 +101,7 @@ class MediaFilesRepository @Inject constructor(
                 }
                 return@withContext ImageDimensions(options.outWidth, options.outHeight)
             } catch (e: IOException) {
-                e.printStackTrace()
+                WooLog.e(T.MEDIA, "MediaFilesRepository > Error getting image dimensions", e)
             }
             return@withContext ImageDimensions(width = 0, height = 0)
         }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -5,8 +5,6 @@ import android.graphics.BitmapFactory
 import android.graphics.BitmapFactory.Options
 import android.net.Uri
 import android.util.Patterns
-import coil.annotation.ExperimentalCoilApi
-import coil.imageLoader
 import com.woocommerce.android.OnChangedException
 import com.woocommerce.android.R
 import com.woocommerce.android.extensions.isNotNullOrEmpty
@@ -44,7 +42,6 @@ import java.io.File
 import java.io.FileDescriptor
 import java.io.IOException
 import java.net.URL
-import java.net.UnknownHostException
 import javax.inject.Inject
 
 class MediaFilesRepository @Inject constructor(
@@ -90,7 +87,6 @@ class MediaFilesRepository @Inject constructor(
         }
     }
 
-    @OptIn(ExperimentalCoilApi::class)
     suspend fun getImageDimensions(uri: String): ImageDimensions {
         return withContext(dispatchers.io) {
             try {
@@ -104,16 +100,6 @@ class MediaFilesRepository @Inject constructor(
                     parcelFileDescriptor.close()
                 }
                 return@withContext ImageDimensions(options.outWidth, options.outHeight)
-            } catch (e: UnknownHostException) {
-                e.printStackTrace()
-                // Image could not be loaded from internet, let's try to load a cached version
-                context.imageLoader.diskCache?.openSnapshot(uri)?.let { snapshot ->
-                    val source = snapshot.data.toFile()
-                    val options = Options().apply { inJustDecodeBounds = true }
-                    BitmapFactory.decodeStream(source.inputStream(), null, options)
-                    snapshot.close()
-                    return@withContext ImageDimensions(options.outWidth, options.outHeight)
-                }
             } catch (e: IOException) {
                 e.printStackTrace()
             }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -5,6 +5,8 @@ import android.graphics.BitmapFactory
 import android.graphics.BitmapFactory.Options
 import android.net.Uri
 import android.util.Patterns
+import coil.annotation.ExperimentalCoilApi
+import coil.imageLoader
 import com.woocommerce.android.OnChangedException
 import com.woocommerce.android.R
 import com.woocommerce.android.extensions.isNotNullOrEmpty
@@ -42,6 +44,7 @@ import java.io.File
 import java.io.FileDescriptor
 import java.io.IOException
 import java.net.URL
+import java.net.UnknownHostException
 import javax.inject.Inject
 
 class MediaFilesRepository @Inject constructor(
@@ -87,6 +90,7 @@ class MediaFilesRepository @Inject constructor(
         }
     }
 
+    @OptIn(ExperimentalCoilApi::class)
     suspend fun getImageDimensions(uri: String): ImageDimensions {
         return withContext(dispatchers.io) {
             try {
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
```diff
@@ -100,6 +104,16 @@ class MediaFilesRepository @Inject constructor(
                     parcelFileDescriptor.close()
                 }
                 return@withContext ImageDimensions(options.outWidth, options.outHeight)
+            } catch (e: UnknownHostException) {
+                e.printStackTrace()
```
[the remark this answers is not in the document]
hichamboushaba: @AdamGrzybkowski I know we've been doing this before this PR, but I think this is a good advice, using `WooLog` will add an entry of the issue to the app log, which could be useful when checking support requests.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/blaze/BlazeRepositoryTest.kt
```diff
@@ -89,12 +107,79 @@ class BlazeRepositoryTest : BaseUnitTest() {
...
+            val product = ProductTestUtils.generateProduct(
+                productId = productId,
+                productName = "My Product",
+                imageUrl = imageUrl
+            ).copy(
+                shortDescription = "<p>Hello<br>World</p>"
+            )
 
-            verify(blazeCampaignsStore).createCampaign(any(), createCampaignRequestCaptor.capture())
-            assertThat(createCampaignRequestCaptor.firstValue.budget.amount).isEqualTo(TOTAL_BUDGET.toDouble())
```
hichamboushaba: Unless I'm missing something here, we deleted the test that had this checks, is this expected?

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
```diff
@@ -100,6 +104,16 @@ class MediaFilesRepository @Inject constructor(
...
+                // Image could not be loaded from internet, let's try to load a cached version
+                context.imageLoader.diskCache?.openSnapshot(uri)?.let { snapshot ->
+                    val source = snapshot.data.toFile()
+                    val options = Options().apply { inJustDecodeBounds = true }
+                    BitmapFactory.decodeStream(source.inputStream(), null, options)
+                    snapshot.close()
+                    return@withContext ImageDimensions(options.outWidth, options.outHeight)
+                }
```
hichamboushaba: Thanks for working on this @AdamGrzybkowski, I'll share some remarks below 🙂 :

> Additionally, [I looked at iOS and noticed they do shows the product image without the network](https://github.com/woocommerce/woocommerce-ios/pull/14868). The Android app logic didn't allow for that, becuase there's an extra check if the image is valid (certain minimum size), and without the internet the image couldn't be loaded.

I think the iOS app uses the same logic as the Android app, it checks the product image dimensions before using it [here](https://github.com/woocommerce/woocommerce-ios/blob/3cbee10a7a041005003d7f9f69e18e87b9df8549/WooCommerce/Classes/ViewRelated/Blaze/CampaignCreation/BlazeCampaignCreationFormViewModel.swift#L480C10-L490).
The linked PR was only about the title and description.

> In https://github.com/woocommerce/woocommerce-android/commit/9e272ef30cc40aa81742d0c3d0e10c820fb286e9 I've added an extra check to use an already cached image as a backup.
>
> It uses some experimental Coil functions, but I think this is mostly because we use a pretty old Coil version. Also, I'm not entirely sure if that is enough given the app uses Glide as well.

I'm fine with using this, but IMHO it's a bit of overengineering for what should be an edge case.
Unlike the `title` and `description` where the AI generation could fail even when the network is fine, being able to download and check the image dimensions is needed for the campaign creation, which will need a working network connection anyway.

> Alternatively, we could remove the isValidAdImage() check from the BlazeRepository.generateDefaultCampaignDetails(). After all, this is just a default, which should be treated as backup, so maybe not doing this check is better than not displaying an image at all.

Just an opinion here, I'm not sure we can consider the product image as a backup, I think many merchants will use it in their campaigns directly.
And regardless of the number of merchants that will use it and given what was shared above, I believe we need to validate the dimensions of the image, as otherwise the campaign creation will fail later, so displaying an empty image to force the merchant to choose an image is a better scenario.

hichamboushaba (commented):
Nice work @AdamGrzybkowski, the fix for title and description works well.

I left some remarks about the product image, let's discuss them in the comments, thank you 👍.

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
AdamGrzybkowski: >The linked PR was only about the title and description.

Correct. I referred to that PR, as it showed the behavior I was implementing, and I noticed the difference in default images between the two (that were not mentioned in the issue).

>I'm fine with using this, but IMHO it's a bit of overengineering for what should be an edge case.

To be honest, I agree. I've added it to try to replicate iOS behavior so that platforms are aligned. Anyway, the user still needs access to the internet to complete the flow, so I wasn't sure about including this.

I get the impression that you would rather drop this and not have any "backup" image. Is that it? (Personally, I'm okay with this)

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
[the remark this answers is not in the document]
AdamGrzybkowski: Yes, definitely! I will update this once we have a final decision in https://github.com/woocommerce/woocommerce-android/pull/14602#discussion_r2349716778

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/blaze/BlazeRepositoryTest.kt
AdamGrzybkowski: 🤦 Thanks for that... not sure what happened. I mean, Junie happened, and I missed that the test was removed. Fixed in https://github.com/woocommerce/woocommerce-android/pull/14602/commits/34eb28d009ee9e573e0f10d8b0096b866be04750

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
hichamboushaba: > I get the impression that you would rather drop this and not have any "backup" image. Is that it? (Personally, I'm okay with this)

Yes, not a strong opinion, but I think using experimental API to access the cache for this scenario is not needed, when the internet loading fails, we'll just require the user to select an image.

But I'll leave the final decision to you 👍

hichamboushaba (approved):
Nice work, the changes look good after fixing the test.

For the image point, I'm OK with both choices, so I'm pre-approving this.

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
AdamGrzybkowski: I will revert that code. Thanks!

PR #14604, opened 2025-09-11
[WOOMOB-1146][Woo POS][Orders] Add search functionality to Woo POS orders

kidinov:
WOOMOB-1146

### Description
Implemented search functionality for orders in Woo POS. The enhancement allows users to search through orders using a search query, with proper handling of empty queries to load all orders. The implementation includes UI updates with a search bar integrated into the toolbar and state management for search operations. The UI is mocked and not final. There is no pagination, and only 1 page is loaded yet.

### Steps to reproduce
1. Navigate to Woo POS orders screen
2. Use the search bar in the toolbar to enter a search query
3. Observe filtered orders appear based on the search
4. Clear the search to see all orders reload

### Testing information
- Tested search functionality with various queries
- Verified empty query behavior reloads all orders
- Verified error state

### Images/gif

https://github.com/user-attachments/assets/55debae6-f1af-41e4-8315-c64ece78267c

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
@@ -4,9 +4,13 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.woocommerce.android.extensions.formatToDDMMMYYYY
 import com.woocommerce.android.model.Order
+import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearchInputState
+import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearchUIEvent
 import com.woocommerce.android.ui.woopos.home.items.WooPosPaginationState
 import com.woocommerce.android.ui.woopos.home.items.WooPosPullToRefreshState
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.Job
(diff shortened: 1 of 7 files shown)


toupper (approved):
LGTM! @kidinov There are some tests failing in CI, feel free to merge the PR after fixing those

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -51,15 +71,112 @@ class WooPosOrdersViewModel @Inject constructor(
...
+            }
+        }
     }
 
     private fun loadOrders() {
-        viewModelScope.launch {
+        cancelJobs()
+        loadingJob = viewModelScope.launch {
+            val currentState = _state.value
+            _state.value = WooPosOrdersState.Loading(searchInputState = currentState.searchInputState)
```
toupper, line 172: @kidinov I’m working on orders pagination and noticed this line may have some unintended effects. Because of this, we always set the state to Loading when loading orders, which hides the list and shows the loading message also during pull-to-refresh. In other parts of the app (e.g. the POS product list), we instead keep the list visible and rely on the PTR indicator to convey that we're loading something. We may want to align orders with that approach, keeping the state to `content`. What do you think?

kidinov: @toupper, you’re right. If it’s the PTR, we should display content with the PTR indicator visible. If it’s the initial loading or retry on the error screen it should be in full-screen loading mode, and similarly for the search query change

PR #14601, opened 2025-09-11
Issue/woomob 1324 woo poslocal catalog the database gets locked when we sync

malinajirka:
#WOOMOB-1324

### Description

This PR updates the approach to inserting products during Local Catalog sync. Previously, the app would start a transaction, fetch products, page at a time, while inserting them as they come in. However, this locks the database for the duration of the whole operation, including the network requests.

The new approach, fetches all products first, and then insert them.

### Steps to reproduce

1. Start the app
2. Switch to a site with a lot of products
3. Wait 20 s (you can use App inspector to check when the service starts running)
4. Go to order list and notice you can't sync orders until the local catalog refresh finishes

### Testing information

1. Start the app
2. Switch to a site with a lot of products
3. Wait 20 s (you can use App inspector to check when the service starts running)
4. Go to order list and notice you can fetch orders

### The tests that have been performed

The above

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14605, opened 2025-09-11
[WooMob-1323] Insert Pos Variations in transaction

malinajirka:
[Parent needs to be merged first](https://github.com/woocommerce/woocommerce-android/pull/14601).

### Description

This PR updates Local Catalog variations action to ensure we fetch all the variations and then insert all of them into the db within a single transaction. It also refactors the class a bit to match the Products related action.

### Testing information

- Verify the PosVariations table is updated when you log in or switch to a different site

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CatalogTooLargeException</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosSyncVariationsAction.kt
```diff
@@ -20,59 +22,134 @@ class WooPosSyncVariationsAction @Inject constructor(
...
+            return when (error) {
+                is CatalogTooLargeException -> error.toSyncResult()
+                else -> WooPosSyncVariationsResult.Failed.UnexpectedError(
+                    error?.message ?: "Failed to fetch variations"
+                )
+            }
+        }
+
+        val (variations, serverDate) = fetchResult.getOrThrow()
+        return posLocalCatalogStore.executeInTransaction {
```
kidinov, line 47: Just thinking out loud, we insert variations and products in different transactions. There’s a `remoteProductId` in the variations. Do you think there’s any possibility of inconsistency?

malinajirka: Good question!

We don't have SQL constraint on the remoteProductId, so even if there was, we should be fine. I also added ticket to move it into a single transaction here - https://linear.app/a8c/issue/WOOMOB-1334/woo-poslocal-catalog-run-both-products-and-variations-within-single.

kidinov (approved):
Looks and tests good!

The only feedback I would give is to consider moving from using the exceptions here to functions that return results. Both here and for products. It would make the code easier to follow since all possible outcomes would be visible in the function signatures. Plus, we'd get compile-time checks to make sure we handle all error cases.

malinajirka:
Thanks for the review and feedback! I've recorded this suggestion here https://linear.app/a8c/issue/WOOMOB-1336/woo-poslocal-catalog-consider-refactoring-the-actions.

PR #14608, opened 2025-09-12
Fix broken trunk after merge of local catalog PRs

malinajirka:
This PR fixes broken trunk after we merged:
- https://github.com/woocommerce/woocommerce-android/pull/14605
- https://github.com/woocommerce/woocommerce-android/pull/14601
- https://github.com/woocommerce/woocommerce-android/pull/14596


ParaskP7 (approved):
👋 @malinajirka and thanks for this change, I've compiled/reviewed/tested (on basic flows) this app and everything LGTM, I think the broken `trunk` problem is now resolved! ✅

FYI: I also searched for any leftover occurrences of `WCPosProductEntity` (versus the new `WooPosProductEntity` entity) and I don't see any. ☑️

PS: We can merge this when CI goes 🟢 ... 🤞

PR #14611, opened 2025-09-12
[Site Picker] Fix issue with the visibility of the Continue button

hichamboushaba:
Closes WOOMOB-1337

### Description
This PR fixes an issue with the visibility of the Continue button: The button is not shown when fetching sites from the API, even when we already have sites cached.

The issue started happening after this PR https://github.com/woocommerce/woocommerce-android/pull/12603, when we added logic to hide the button during the fetch in an attempt to fix a crash, the issue with that change is that it was too broad, as there are cases when the button should still be shown. This PR adjustes the logic so that the button is hidden only when we need to show the skeleton loading view, which means the cache is empty or we need to enforce a refresh.

### Steps to reproduce
1. To better see the issue, sign in using your A8C account to the app.
2. Ensure your account is not connected to any Jetpack CP sites, as we enforce a refresh for those sites.
3. Open the site picker.

### Testing information
- Confirm the continue button is shown at the start.
- Good to test: add a Jetpack CP site to your account and confirm that the button is hidden when we open the screen, and the skeleton is shown.

The above.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 - [Internal] Migration to POS specific product model [https://github.com/woocommerce/woocommerce-android/pull/14598]
 - [*] Enhance the shipping label creation screen by making the selected hazardous box tappable [https://github.com/woocommerce/woocommerce-android/pull/14576]
 - [*] Shipping Labels: Enhance selected package UI in the create shipping label flow [https://github.com/woocommerce/woocommerce-android/pull/14579]
+- [*] Fix an issue where there is delay before displaying the continue button in Site Picker screen [https://github.com/woocommerce/woocommerce-android/pull/14611]
 
 23.2
 -----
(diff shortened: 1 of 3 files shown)


AdamGrzybkowski (commented):
The button is now visible 🎉 

>Good to test: add a Jetpack CP site to your account and confirm that the button is hidden when we open the screen, and the skeleton is shown.

Regarding this. That's not how it worked for me. I didn't see a skeleton - the update happened in the background without any skeleton/loading indicator, I only saw a new site appear after it was loaded. 

In fact, I don't know how to get that skeleton to appear. I tried a fresh login, but before I could go to that page, the cache was already there. 

Is that expected? If not, could you provide more detailed steps on how to reproduce?

AdamGrzybkowski (approved):
>Regarding this. That's not how it worked for me. I didn't see a skeleton - the update happened in the background without any skeleton/loading indicator, I only saw a new site appear after it was loaded.
>In fact, I don't know how to get that skeleton to appear. I tried a fresh login, but before I could go to that page, the cache was already there.
>Is that expected? If not, could you provide more detailed steps on how to reproduce?

Retested with a new site with Jetpack CP and the skeleton was visible, the continue button was not.

PR #14614, opened 2025-09-15
[POS Orders] Pagination

toupper:
Part of WOOMOB-1149

### Description

With this PR we add pagination to the POS Orders list, loading more pages when the user scrolls down to the bottom.

 This PR will be followed by:
- Pagination Error
- Search Pagination

### Steps to reproduce

With a site with enough POS Orders (more than the page size 25)

1. Open the app
2. Go to POS
3. Go to POS Orders
4. Scroll the Orders list to the bottom
5. See the loading shimmering row
6. See how it loads a new page

See above.

### Images/gif

https://github.com/user-attachments/assets/8b147ba2-80d1-49c1-8155-2e574587a5e8


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -148,10 +159,17 @@ private fun OrdersList(
...
+                    val isLoadingMore = state.paginationState == WooPosPaginationState.Loading
+                    val canLoadMore = state.paginationState == WooPosPaginationState.None &&
+                        state.pullToRefreshState != WooPosPullToRefreshState.Refreshing
+
                     WooPosOrdersListPaneScreen(
                         items = state.items,
                         selectedOrderId = state.selectedOrderId,
                         onOrderSelected = onOrderSelected,
+                        onEndOfOrdersListReached = onEndOfOrdersListReached,
+                        canLoadMore = canLoadMore && state.pullToRefreshState != WooPosPullToRefreshState.Refreshing,
```
kidinov: It looks like that this is equal to

```
canLoadMore = state.paginationState == WooPosPaginationState.None &&
                        state.pullToRefreshState != WooPosPullToRefreshState.Refreshing && state.pullToRefreshState != WooPosPullToRefreshState.Refreshing
```

comparing the same twice

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -54,8 +58,44 @@ class WooPosOrdersViewModel @Inject constructor(
...
 
+    @Suppress("ReturnCount")
+    fun onEndOfOrdersListReached() {
+        val currentState = _state.value
+        if (currentState !is WooPosOrdersState.Content) {
+            return
+        }
+
+        if (loadOrdersJob?.isActive == true) {
+            loadMoreAfterLoadCompletes = true
```
kidinov: In my opinion, queuing another load of requests overcomplicates things and isn’t worth the added complexity. I don’t believe users expects this behavior

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -54,8 +58,44 @@ class WooPosOrdersViewModel @Inject constructor(
...
+            _state.value = currentState.copy(paginationState = WooPosPaginationState.Loading)
+            return
+        }
+
+        if (loadMoreOrdersJob?.isActive == true) {
```
kidinov: You just checked this above

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -54,8 +58,44 @@ class WooPosOrdersViewModel @Inject constructor(
...
+                updateContentState(result.getOrThrow())
+            } else {
+                _state.value = currentState.copy(paginationState = WooPosPaginationState.Error)
+            }
+        }
+    }
+
     private fun loadOrders() {
-        viewModelScope.launch {
+        loadOrdersJob?.cancel()
```
kidinov: I think this is done already in the search PR so can be reused after taking the changes from there. Also, keep in mind that pagination should also work when the search is on

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Good catch! Done in 61e72f3afb3226de2f0198b9d44ebe8368f23d0b

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: I agree, I removed that logic in a62305a35e56ed2e18eaab1d67f74aa51436cd9e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Above I just checked the `loadOrdersJob`, not the `loadMoreOrdersJob` 

```
if (loadOrdersJob?.isActive == true) {
            _state.value = currentState.copy(paginationState = WooPosPaginationState.Loading)
            return
        }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Yeah, I will manage that when merging the work on for the Search

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -194,10 +206,17 @@ private fun OrdersList(
                 }
 
                 is WooPosOrdersState.Content -> {
+                    val isLoadingMore = state.paginationState == WooPosPaginationState.Loading
```
kidinov: Sorry for not commenting on this before. I think the filtering of when to request another page and when to ignore the "end of list" event should be done in the VM

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -80,6 +81,39 @@ class WooPosOrdersViewModel @Inject constructor(
...
 
+    @Suppress("ReturnCount")
+    fun onEndOfOrdersListReached() {
+        val currentState = _state.value
+        if (currentState !is WooPosOrdersState.Content) {
+            return
+        }
+
+        if (loadingJob?.isActive == true) {
+            _state.value = currentState.copy(paginationState = WooPosPaginationState.Loading)
```
kidinov: So, if the loading job is active, which is likely due to the PTR, then when you scroll down, I don’t think we should show the loading item at the bottom without loading the next page. We shouldn’t load the next page, but we shouldn’t show the loading indicator either. So probably

```
        if (loadingJob?.isActive == true || loadingMoreOrdersJob?.isActive == true) {
            return
        }
```

is better

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -80,6 +81,39 @@ class WooPosOrdersViewModel @Inject constructor(
...
+
+        _state.value = currentState.copy(paginationState = WooPosPaginationState.Loading)
+
+        loadingMoreOrdersJob?.cancel()
+        loadingMoreOrdersJob = viewModelScope.launch {
+            val result = ordersDataSource.loadMore()
+            if (result.isSuccess) {
+                appendOrders(result.getOrThrow())
+            } else {
+                _state.value = currentState.copy(paginationState = WooPosPaginationState.Error)
```
kidinov, line 110: You don't show a loading next page error - is this expected for now? Another PR?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -58,6 +72,20 @@ class WooPosOrdersDataSource @Inject constructor(
         }
     }
 
+    suspend fun loadMore(): Result<List<Order>> = withContext(Dispatchers.IO) {
+        val result = fetchOrdersFromRemote(searchQuery = null, page = page.get())
```
kidinov, line 76: You always load the next page without taking into account the search query value

kidinov (commented):
Thanks for your here!

Overall, everything looks good. The only thing I think needs to be fixed surely is that pagination doesn’t currently work for search cases

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Good point, done in f84c0a88fdc95ea27b38339912640cafbcb437f4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: That makes more sense indeed, changed in 0c433d5f913dc3cb506e7b5f2979a24b89794f1a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Yes, that will be part of the next PR

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Thanks for pointing that out! I would like to tackle that on its own future PR for Search pagination

toupper:
Thanks for the review @kidinov! As mentioned in the comments, this PR will be followed by:
- Pagination Error
- Search Pagination

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: It’s not quite what I meant. The VM should not expose anything beyond its state.  Here is an example - I didn't test it much, but it looks correct

[Refactor_pagination_logic_in_WooPosOrdersScreen_and_WooPosOrdersViewModel.patch](https://github.com/user-attachments/files/22406930/Refactor_pagination_logic_in_WooPosOrdersScreen_and_WooPosOrdersViewModel.patch)

kidinov (approved):
LGTM! I left one refactoring proposal which I think would be nice to implement, please take a look

PR #14616, opened 2025-09-15
[CIAB] Integrate the garden arch API and add a helper class for deciding features to be disabled for CIAB sites

hichamboushaba:
Closes WOOMOB-1274

### Description
This is the first PR for updating the app to allow adjusting the experience for CIAB sites, the PR introduces the following changes:
- It updates the `SiteModel` DB table to include the new fields added for the garden architecture.
- It updates the SiteRestClient to make sure we read the new fields from the API (Check https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/)
- Adds the `CIABSiteGateKeeper` class for controlling which features we'll disable for CIAB sites.

**Note:** The `CIABSiteGateKeeper` and `CIABAffectedFeature` architecture is modeled after what the iOS team [did](https://github.com/woocommerce/woocommerce-ios/pull/16095), just with different names. I think the architecture makes sense, even though we'll disable all features of the enum `CIABAffectedFeature` initially. Also, even though I'm not sure we'll ever need this, the flexibility could be helpful in the future if we need to offer different features to the different partners.
For some context on the Garden Architecture, check this PdpAdu-2ai-p2

#### For the sites API
1. Apply the following patch
<details>
<summary> Patch </summary>

```patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt	(revision 0acce60e28a92556603847c978a3940811087867)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt	(date 1757956052483)
@@ -117,6 +117,7 @@
 import org.wordpress.android.login.LoginAnalyticsListener
 import org.wordpress.android.login.LoginMode
 import org.wordpress.android.util.NetworkUtils
+import org.wordpress.android.util.ToastUtils
 import java.lang.ref.WeakReference
 import java.math.BigDecimal
 import java.util.Locale
@@ -363,6 +364,12 @@
             viewModel.handleShortcutAction(intent?.action?.lowercase(Locale.ROOT))
             handleIncomingImages()
         }
+
+        selectedSite.get().let {
+            if (it.isGardenSite) {
+                ToastUtils.showToast(this, "Detected a Garden site:\nGarden Name: ${it.gardenName}\nGarden Partner: ${it.gardenPartner}")
+            }
+        }
     }
 
     private fun handleIncomingImages() {
```

</details>

2. Create a CIAB site using this guide and tool pdpAdu-29A-p2
3. Open the app and sign in to your account.
4. Confirm a toast is shown with the garden name and partner when opening the app.

### Testing information
- Test the API integration.
- Review the unit tests and logic.

The above.

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
@@ -1,0 +1,44 @@
+package com.woocommerce.android.ciab
+
+import com.woocommerce.android.tools.SelectedSite
+import com.woocommerce.android.viewmodel.BaseUnitTest
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.mockito.Mockito.mock
+import org.mockito.kotlin.given
+import org.wordpress.android.fluxc.model.SiteModel
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class CIABSiteGateKeeperTest : BaseUnitTest() {
+    private val selectedSite: SelectedSite = mock()
+    private val ciabSiteGateKeeper = CIABSiteGateKeeper(selectedSite)
+
+    @Test
+    fun `given current site is CIAB, when checking feature support, then feature is unsupported`() {
+        val site = createSite(isCIAB = true)
+        given(selectedSite.getOrNull()).willReturn(site)
+
+        CIABAffectedFeature.entries.forEach {
+            assertFalse(ciabSiteGateKeeper.isFeatureSupported(it))
+        }
+    }
+
+    @Test
+    fun `given current site is not CIAB, when checking feature support, then feature is supported`() {
+        val site = createSite(isCIAB = false)
+        given(selectedSite.getOrNull()).willReturn(site)
+
+        CIABAffectedFeature.entries.forEach {
+            assertTrue(ciabSiteGateKeeper.isFeatureSupported(it))
+        }
+    }
+
+    private fun createSite(isCIAB: Boolean): SiteModel {
+        return SiteModel().apply {
+            setIsGardenSite(isCIAB)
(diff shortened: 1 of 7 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    private val selectedSite: SelectedSite = mock()
+    private val ciabSiteGateKeeper = CIABSiteGateKeeper(selectedSite)
+
+    @Test
+    fun `given current site is CIAB, when checking feature support, then feature is unsupported`() {
+        val site = createSite(isCIAB = true)
+        given(selectedSite.getOrNull()).willReturn(site)
+
+        CIABAffectedFeature.entries.forEach {
+            assertFalse(ciabSiteGateKeeper.isFeatureSupported(it))
```
AdamGrzybkowski, line 24: This test fails because `selectedSite.getIfExists()` returns null in `CIABSiteGateKeeper.isCurrentSiteCIAB()`

AdamGrzybkowski (approved):
Looks good! Tested with the provided patch, but also checked the network responses and database table rows to confirm the values were set correctly. 

I left a comment about the UI test failing. Approving because the test is missing a mock and the rest of the code works well.

WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
hichamboushaba: 😅 
I was using `SelectedSite#getOrNull()`, but after some refactoring, Github copilot decided that I need `getIfExists` instead 😄 

Fixed in the updated commit e19da62

PR #14636, opened 2025-09-19
[POS][Local Catalog] Connect product list to POS database

samiuelson:
WOOMOB-1072

### Description

The goal of this PR is to load products from the DB instead of the existing in-memory cache / remote call hybrid store. Currently, the logic that decides on which data source to use is based on the feature flag WOO_POS_LOCAL_CATALOG_M1 state.

⚠️ Known issue: WOOMOB-1370 - variations have wrong names. This will be solved in a separate PR.

### Steps to reproduce

—

### Testing information

1. Verify that in the release build (FF disabled), the previous data source implementation is used.
2. Verify that PTR works when FF is disabled.
3. When the FF is enabled, verify that products load correctly - full list. They should load significantly faster without pagination.

Above

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
```diff
@@ -61,6 +68,9 @@ class WooPosProductsViewModel @Inject constructor(
 
     init {
         listenEventsFromParent()
+        viewModelScope.launch {
```
kidinov: Do we need this or should the data source object be created every time when the VM is created?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -339,13 +342,37 @@ class WooPosCartViewModel @Inject constructor(
     }
 
     private suspend fun handleSimpleProductClicked(productId: Long): WooPosCartItemViewState {
-        val product = getProductById(productId)!!
+        val product = when {
+            wooPosLocalCatalogM1Enabled() -> {
```
kidinov: I think it'll be better to have this logic in `WooPosGetProductById` as this reusable component that used in many places. wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -339,13 +342,37 @@ class WooPosCartViewModel @Inject constructor(
...
+        }
+
         val itemNumber = getItemNumber()
         return product.toCartListItem(itemNumber)
     }
 
     private suspend fun handleVariationClicked(productId: Long, variationId: Long): WooPosCartItemViewState {
-        val product = getProductById(productId)!!
+        val product = when {
+            wooPosLocalCatalogM1Enabled() -> {
```
kidinov: Even in this VM you have the same code twice, so imo should be in `getProductById`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsRepository.kt
```diff
@@ -115,11 +134,37 @@ class WooPosTotalsRepository @Inject constructor(
...
+                    remoteProductId = LocalOrRemoteId.RemoteId(itemData.productId)
+                )
+                result.getOrNull()?.let { entity ->
+                    productMapper.fromEntity(entity)
+                } ?: error("Product not found in local catalog: ${itemData.productId}")
+            }
+            else -> getProductById(itemData.productId)!!
+        }
+        val variationResult = when {
+            wooPosLocalCatalogM1Enabled() -> {
```
kidinov: Same here for `getVariationById`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
```diff
@@ -44,6 +45,12 @@ class WooPosProductsViewModel @Inject constructor(
     private val selectedSite: SelectedSite,
     private val resourceProvider: ResourceProvider,
 ) : ViewModel() {
+
+    private val currentDataSource: WooPosProductsDataSourceInterface = when (wooPosLocalCatalogM1Enabled()) {
```
kidinov: Have you considered moving this to the place where the instance is created, e.g. in dagger module? So the right implementation will be injected everywhere where needed?

kidinov (approved):
LGTM! I left a couple of comments, please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
samiuelson: Yeah, good idea. Done in b1e60e4110c51eb1c8382af2d1fb93949db20923

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
samiuelson: Refactored in b1e60e4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsRepository.kt
samiuelson: Done in b1e60e4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
samiuelson: Good point. Moved this to the module: 57b2078ee050ab8b121305ea76fec6129bd05817

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
samiuelson: It's not needed since fetchFirstPage() already handles the state reset internally. Removed: 0f3456a114eca615d1ae3792e7deacaafb671dda

PR #14640, opened 2025-09-19
Issue/woomob 1351 add logic to show or hide the booking tab dynamically

JorgeMucientes:
Closes: WOOMOB-1351

### Description

This is a follow-up PR to handle the visibility of the Booking tab dynamically based on the following criteria: 
- Site is CIAB
- Site has at least 1 bookable product published
- Feature flag is enabled locally

I've added a Flow to keep any changes around those 3 conditions reactive.

### Testing information
To make tests quicker I recommed logging into the app with a WP.com account that has at least access to:
- A WP.com/Jetpack connected site non CIAB
- A CIAB site with Bookings plugin installed

1. Log into a CIAB site with Bookings plugin enabled. Have at least one Bookable product created. You can create such sites from matticspace `tools/garden/index.php` 
2. Check the new bookings tab is shown. 
3. Edit the bookable product and make it a draft. Notice how the bookings tab is gone
4. Publish the bookable product again and see how tab is visible again
5. Delete the product completely and check tab is gone. 
6. Change to a non CIAB site with bookable products published and check the Bookings tab is still gone. 

The above ☝🏼 

### Images/gif

https://github.com/user-attachments/assets/45a0f6b6-2142-4193-bc08-f2369f7836e8


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABExtensions.kt
```diff
@@ -0,0 +1,6 @@
+package com.woocommerce.android.ciab
+
+import com.woocommerce.android.ciab.CIABSiteGateKeeper.Companion.CIAB_GARDEN_NAME
+import org.wordpress.android.fluxc.model.SiteModel
+
+fun SiteModel.isCIABSite() = isGardenSite && gardenName == CIAB_GARDEN_NAME
```
JorgeMucientes: Unsure this is the best place for such extension. Open to suggestions

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -34,12 +37,13 @@ class BookingsTabController @Inject constructor(
...
         activity.lifecycleScope.launch {
-            showBookingsTab()
-                .onSuccess {
-                    binding.bottomNav.menu.findItem(R.id.bookings)?.isVisible = it
-                }
-                .onFailure {
-                    // TODO log error or track errors?
+            selectedSite.observe()
+                .filter { it != null }
+                .collect { siteModel ->
```
JorgeMucientes: We start observing bookable products only when we know for sure a site has been selected. Otherwise the app will crash. For example when loggin into the app for the first time, MainActivity is resumed before the app has finished selecting a site, leading to a crash.

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABExtensions.kt
hichamboushaba: We already have `SiteModelExt` for similar extensions, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -34,12 +37,13 @@ class BookingsTabController @Inject constructor(
...
     private fun checkBookingsTabVisibility() {
         activity.lifecycleScope.launch {
-            showBookingsTab()
-                .onSuccess {
-                    binding.bottomNav.menu.findItem(R.id.bookings)?.isVisible = it
-                }
-                .onFailure {
-                    // TODO log error or track errors?
+            selectedSite.observe()
+                .filter { it != null }
```
hichamboushaba: I suggest using `filterNotNull` here, it will automatically cast the variable and allow you to get rid of the double-bang operator below.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsTabVisibility.kt
```diff
@@ -0,0 +1,45 @@
...
+    ).onStart {
+        productListRepository.fetchProductList(
+            productFilterOptions = mapOf(ProductFilterOption.TYPE to BOOKING_PRODUCT_TYPE)
+        ).onFailure {
+            WooLog.w(WooLog.T.BOOKINGS, "Failed to fetch bookable products")
+        }
+    }
+        .map { productsCount ->
+            productsCount > 0 &&
+                siteModel?.isCIABSite() == true &&
```
hichamboushaba: `siteModel` is non-nullable here

```suggestion
                siteModel.isCIABSite() == true &&
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsTabVisibility.kt
```diff
@@ -0,0 +1,45 @@
...
+) {
+
+    operator fun invoke(siteModel: SiteModel): Flow<Boolean> = productListRepository.observeProductsCount(
+        filterOptions = mapOf(
+            ProductFilterOption.STATUS to ProductStatus.PUBLISH.value,
+            ProductFilterOption.TYPE to BOOKING_PRODUCT_TYPE
+        ),
+        excludeSampleProducts = true
+    ).onStart {
+        productListRepository.fetchProductList(
```
hichamboushaba: With this logic, we'll make an API request to fetch bookable products for all sites, WDYT about changing the order of things slightly to make sure we return an early `false` when the site is not CIAB, soemthing like the following:

```kotlin
    operator fun invoke(siteModel: SiteModel): Flow<Boolean> = flow {
        val isCIABSite = FeatureFlag.BOOKINGS_MVP.isEnabled() && siteModel.isCIABSite()
        if (!isCIABSite) {
            emit(false)
        } else {
            emitAll(
                productListRepository.observeProductsCount(
                    filterOptions = mapOf(
                        ProductFilterOption.STATUS to ProductStatus.PUBLISH.value,
                        ProductFilterOption.TYPE to BOOKING_PRODUCT_TYPE
                    ),
                    excludeSampleProducts = true
                ).map { count ->
                    count > 0
                }.onStart {
                    productListRepository.fetchProductList(
                        productFilterOptions = mapOf(ProductFilterOption.TYPE to BOOKING_PRODUCT_TYPE)
                    ).onFailure {
                        WooLog.w(WooLog.T.BOOKINGS, "Failed to fetch bookable products")
                    }
                }.distinctUntilChanged()
            )
        }
    }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -34,12 +37,13 @@ class BookingsTabController @Inject constructor(
 
     private fun checkBookingsTabVisibility() {
```
hichamboushaba, line 25: Not directly related to this PR, but just sharing an observation here, I think that given we use `activity.lifecycleScope.launch` in this function, it's safe to call it directly from `init` without making this class a `LifecycleObserver`, `lifecycleScope` already takes care of this, and won't start the Coroutine until the `Activity` is started.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsTabVisibility.kt
hichamboushaba: Another point related to this, with the fetch happening in `onStart`, we don't emit any value until it's done, and this results in showing the bookings tab before hiding it when switching to any site that doesn't have the requirements:

https://github.com/user-attachments/assets/a6e08591-32af-4324-8e22-14e7f14a940f

This issue affects both your version and the above suggested version (for CIAB sites that don't have bookable products), we can avoid this by making the fetch async:

```kotlin
    .onStart {
        appCoroutineScope.launch {
            productListRepository.fetchProductList(
                productFilterOptions = mapOf(ProductFilterOption.TYPE to BOOKING_PRODUCT_TYPE)
            ).onFailure {
                WooLog.w(WooLog.T.BOOKINGS, "Failed to fetch bookable products")
            }
        }
    }
```

hichamboushaba (commented):
Nice work @JorgeMucientes, the logic is working as expected, but I think we can make some improvements to fix some issues that I shared below, please check the comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -1,17 +1,18 @@
...
 import com.woocommerce.android.ui.main.MainActivity
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 class BookingsTabController @Inject constructor(
-    private val showBookingsTab: ShowBookingsTab
-) : DefaultLifecycleObserver {
+    private val observeBookingsTabVisibility: ObserveBookingsTabVisibility,
+    private val selectedSite: SelectedSite
```
AdamGrzybkowski: ❓ Is there a reason, we can't inject the `selectedSite` in `ObserveBookingsTabVisibility` and keep all the logic there?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -21,25 +22,18 @@ class BookingsTabController @Inject constructor(
...
-                }
-                .onFailure {
-                    // TODO log error or track errors?
+            selectedSite.observe()
+                .filterNotNull()
+                .collect { siteModel ->
+                    observeBookingsTabVisibility(siteModel)
+                        .collect {
+                            binding.bottomNav.menu.findItem(R.id.bookings)?.isVisible = it
+                        }
```
AdamGrzybkowski: ⚠️ This code might not work as expected. The inner `observeBookingsTabVisibility(siteModel).collect{}` will suspend the coroutine and block the outer `collect {}` from receiving new items. 

That means, if the `observeBookingsTabVisibility(siteModel)` takes a lot of time and in the meantime the selected site changes, the new site won't be collected. 

An alternative would be:

```kotlin
selectedSite.observe()
            .filterNotNull()
            .flatMapLatest { siteModel ->
                observeBookingsTabVisibility(siteModel)
            }
            .collect { isVisible ->
                binding.bottomNav.menu.findItem(R.id.bookings)?.isVisible = isVisible
            }
```

hichamboushaba: Good eye, I missed this during my review. It works now becasue the `MainActivity` is re-created when we switch sites, so we always cancel the collection of the outer flow and restart it. But I agree it should be fixed regardless of how MainActivity behaves when switching sites.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
JorgeMucientes: I didn't gave much though to it tbh. Any advantages I'm not seeing from having everything in `ObserveBookingsTabVisibility`?
I feel the current implementation is readable/clean enough, but I'm open to move the selected site to `ObserveBookingsTabVisibility`.

JorgeMucientes:
Excellent feedback @hichamboushaba @AdamGrzybkowski thank you. Great eye for detail! I've addressed all your suggested feedback. Ready for another round :)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
AdamGrzybkowski: Two things I see: 
1. Keeping the whole tab visibility logic in one place. 
2. `ObserveBookingsTabVisibilityTest` - the `BookingsTabController` doesn't have tests, and because of Android dependencies, it's a bit trickier to add them. Keeping everything in `ObserveBookingsTabVisibility` makes it possible to test the whole flow, including the suspending issue mentioned with the previous code.

hichamboushaba (approved):
Nice work @JorgeMucientes, thanks for taking care of this.

I won't merge this to give you a chance to check Adam's last remark about moving `SelectedSite` to `ObserveBookingsTabVisibility`.

AdamGrzybkowski (approved):
Tested and looks good 👍 

The other comment is non-blocking so approving!

JorgeMucientes:
Thanks for all the feedback folks. I've moved the logic around observing selected site to `ObserveBookingsTabVisibility` and kept all the logic there as suggested. The provided reasons where good enough to apply the refactor.

PR #14646, opened 2025-09-22
Develocity 360 trial configuration update

wzieba:
### Description

These changes update `develocity` Gradle Plugin configuration to point to our instance of Develocity (we're currently in trial).

This will update build data to our servers instead of Develocity.


ParaskP7 (approved):
👋 @wzieba and thanks for this update, excited we can try Develocity 360! 🎉

I tested it locally and submitted a build (see [build scan](https://develocity.a8c-ci.services/s/hdi76wdkuvkz6)), it seems everything is working as expected locally. I also checked the instance online (https://develocity.a8c-ci.services) an it seems to be reporting CI builds as expected too, with the correct tags and all. ✅ 🚀

PR #14649, opened 2025-09-23
[WOOMOB-1342] Booking details header/list item component

AdamGrzybkowski:
WOOMOB-1342

### Description

This PR creates the top/header Composable component for the Booking details screen. It's also the same component that can be reused on the Booking list screen. 

I will leave some extra comments below.

### Steps to reproduce

1. Log in
2. Go to Settings -> Developer Options
3. Tap "Show booking details"
4. Verify the UI matches the designs

### Testing information

Test with different font/display scaling and dark/light mode.

The above

### Images/gif

| Light | Dark |
|---|---|
| Screenshot 2025-09-23 at 09 48 50 | Screenshot 2025-09-23 at 09 49 06 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
@@ -1,0 +1,59 @@
+package com.woocommerce.android.ui.bookings.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import com.woocommerce.android.R
+import com.woocommerce.android.ui.compose.component.WCTag
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+
+@Composable
+fun AttendanceStatusTag(
+    state: AttendanceStatus,
+    modifier: Modifier = Modifier,
+) {
+    WCTag(
+        text = state.text(),
+        backgroundColor = colorResource(R.color.tagView_bg),
+        textColor = colorResource(R.color.tagView_text),
+        fontWeight = FontWeight.Normal,
+        modifier = modifier
+    )
+}
+
+enum class AttendanceStatus {
+    BOOKED, CHECKED_IN, CANCELLED, NO_SHOW
+}
+
+@Composable
+private fun AttendanceStatus.text(): String {
+    return when (this) {
+        AttendanceStatus.BOOKED -> R.string.booking_attendance_status_booked
+        AttendanceStatus.CHECKED_IN -> R.string.booking_attendance_status_checked_in
+        AttendanceStatus.CANCELLED -> R.string.booking_attendance_status_cancelled
+        AttendanceStatus.NO_SHOW -> R.string.booking_attendance_status_no_show
+    }.let { stringResource(it) }
+}
+
+@Preview
+@Composable
+private fun AttendanceStatusTagPreview() {
+    WooThemeWithBackground {
+        AttendanceStatusTag(
+            state = AttendanceStatus.BOOKED
+        )
+    }
+}
+
+@Preview(uiMode = android.content.res.Configuration.UI_MODE_NIGHT_YES)
+@Composable
+private fun AttendanceStatusTagDarkPreview() {
+    WooThemeWithBackground {
+        AttendanceStatusTag(
+            state = AttendanceStatus.CHECKED_IN
+        )
+    }
+}
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
@@ -44,7 +44,7 @@ fun BookingSummary(
                 style = MaterialTheme.typography.labelMedium.copy(fontSize = 13.sp),
             )
             Text(
-                text = model.staff,
+                text = model.customerName,
                 color = colorResource(id = R.color.color_surface_variant),
                 style = MaterialTheme.typography.labelMedium.copy(fontSize = 13.sp),
             )
@@ -67,7 +67,7 @@ fun BookingSummary(
 data class BookingSummary(
     val date: String,
     val name: String,
-    val staff: String,
+    val customerName: String,
     val attendanceStatus: AttendanceStatus,
     val paymentStatus: BookingPaymentStatus,
 )
@@ -80,7 +80,7 @@ private fun BookingSummaryPreview() {
             model = BookingSummary(
                 date = "05/07/2025, 11:00 AM",
                 name = "Women’s Haircut",
-                staff = "Margarita Nikolaevna",
+                customerName = "Margarita Nikolaevna",
                 attendanceStatus = AttendanceStatus.CHECKED_IN,
                 paymentStatus = BookingPaymentStatus.PAID
             ),
@@ -97,7 +97,7 @@ private fun BookingSummaryDarkPreview() {
             model = BookingSummary(
                 date = "05/07/2025, 11:00 AM",
                 name = "Women’s Haircut",
-                staff = "Margarita Nikolaevna",
+                customerName = "Margarita Nikolaevna",
                 attendanceStatus = AttendanceStatus.BOOKED,
                 paymentStatus = BookingPaymentStatus.PENDING_CONFIRMATION
             ),
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
@@ -1,0 +1,107 @@
+package com.woocommerce.android.ui.bookings.compose
+
+import android.content.res.Configuration
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.FlowRow
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.woocommerce.android.R
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+
+@Composable
+fun BookingSummary(
+    model: BookingSummary,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        horizontalAlignment = Alignment.Start,
+        modifier = modifier
+            .padding(16.dp)
+    ) {
+        Text(
+            text = model.date,
+            color = MaterialTheme.colorScheme.onSurface,
+            style = MaterialTheme.typography.bodyLarge.copy(fontWeight = FontWeight.Medium),
+        )
+        FlowRow(
+            modifier = Modifier.padding(top = 2.dp),
+        ) {
+            Text(
+                text = "${model.name} • ",
+                color = colorResource(id = R.color.color_surface_variant),
+                style = MaterialTheme.typography.labelMedium.copy(fontSize = 13.sp),
+            )
+            Text(
+                text = model.staff,
+                color = colorResource(id = R.color.color_surface_variant),
+                style = MaterialTheme.typography.labelMedium.copy(fontSize = 13.sp),
+            )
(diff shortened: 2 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+
+@Composable
+fun AttendanceStatusTag(
+    state: AttendanceStatus,
+    modifier: Modifier = Modifier,
+) {
+    WCTag(
+        text = state.text(),
+        backgroundColor = colorResource(R.color.tagView_bg),
```
AdamGrzybkowski, line 20: This color doesn't match the designs, but it's the one used for the same components in the app. I would assume the consistency is more important here, but let me know if I should change it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
```diff
@@ -0,0 +1,59 @@
...
+        textColor = colorResource(R.color.tagView_text),
+        fontWeight = FontWeight.Normal,
+        modifier = modifier
+    )
+}
+
+enum class AttendanceStatus {
+    BOOKED, CHECKED_IN, CANCELLED, NO_SHOW
+}
```
AdamGrzybkowski, line 29: I've decided to keep the models for composable in the same file - would you prefer to put them somwhere else?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
```diff
@@ -0,0 +1,107 @@
...
+            .padding(16.dp)
+    ) {
+        Text(
+            text = model.date,
+            color = MaterialTheme.colorScheme.onSurface,
+            style = MaterialTheme.typography.bodyLarge.copy(fontWeight = FontWeight.Medium),
+        )
+        FlowRow(
+            modifier = Modifier.padding(top = 2.dp),
+        ) {
```
AdamGrzybkowski, line 40: `FlowRow` because it's hard to predict the length of those two fields: booking name and the staff name. Each one could be quite long, so it's highly likely to exceed one line. 

Alternatively, we could use the auto-size feature to decrease font size to fit in one line, not sure what the preference is. Any thoughts?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
```diff
@@ -0,0 +1,61 @@
...
+        backgroundColor = colorResource(R.color.tagView_bg),
+        textColor = colorResource(R.color.tagView_text),
+        fontWeight = FontWeight.Normal,
+        modifier = modifier
+    )
+}
+
+enum class BookingPaymentStatus {
+    UNPAID, PENDING_CONFIRMATION, CONFIRMED, PAID, CANCELLED, COMPLETE
+}
```
AdamGrzybkowski, line 29: These are taken from this post pe5sF9-4Bq-p2. I believe, this is not a final list of possible states.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+
+@Composable
+fun BookingPaymentStatusTag(
+    state: BookingPaymentStatus,
+    modifier: Modifier = Modifier,
+) {
+    WCTag(
+        text = state.text(),
+        backgroundColor = colorResource(R.color.tagView_bg),
```
AdamGrzybkowski, line 20: This should change depending on the state, but not all the states are defined yet.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
```diff
@@ -1,5 +1,16 @@
...
+
 data class BookingDetailsViewState(
     val toolbarTitle: String = "",
+    val bookingSummary: BookingSummary = BookingSummary(
+        date = "05/07/2025, 11:00 AM",
+        name = "Women’s Haircut",
+        staff = "Margarita Nikolaevna",
+        attendanceStatus = AttendanceStatus.CHECKED_IN,
+        paymentStatus = BookingPaymentStatus.PAID
+    ),
```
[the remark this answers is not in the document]
AdamGrzybkowski: This is temporary

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
irfano: Yes, good call! I would do the same.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
irfano: We’ll determine the tag’s status in the view model since it'll come from the backend. So I think these models could live there as well. What do you think about moving `AttendanceStatus` and its `text()` logic into the view model? That way, the UI only needs the tag’s string resource, not the types.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
```diff
@@ -0,0 +1,107 @@
...
+            Text(
+                text = "${model.name} • ",
```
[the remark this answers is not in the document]
irfano: I guess this is a valid comment. Maybe we can use `<string name="booking_name_and_staff">1%s • 2%s</string>`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
irfano: I’m not a fan of auto-size, as it causes inconsistent UI. I think FlowRow is a good solution.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
```diff
@@ -0,0 +1,61 @@
...
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import com.woocommerce.android.R
+import com.woocommerce.android.ui.compose.component.WCTag
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+
+@Composable
+fun BookingPaymentStatusTag(
```
irfano, line 14: Do we really need two separate composables for the attendance and payment status tags? If the only difference is the data, we could just use `WCTag` instead of creating new files (`AttendanceStatusTag` and `PaymentStatusTag`) and determine the text in the view model. Are you aware of any advantages to keeping those dedicated composables?

irfano (commented):
Good job! It looks great across various accessibility settings. I’ve added a few comments and would like to hear your thoughts on them before approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
AdamGrzybkowski: Good questions. I've separated them because they will have different background color logic. The color will change depending on the state, and we can encapsulate the logic for the proper text resource locally rather than passing string resource IDs from the ViewModel. 

We could, of course, keep one more generic component that takes all the properties to achieve the same customization (text and background color). 

So to answer your question - no, we don't, but I prefer readability over a smaller number of files.

| More components | One generic |
|---|---|
| Screenshot 2025-09-23 at 14 07 06 | Screenshot 2025-09-23 at 14 09 09 |

Not related here, because we don't have screenshot tests, but when you have those small, individual components, you can have a pretty good UI test coverage, by testing each one in all possible states with a tool like [Roborazzi](https://github.com/takahirom/roborazzi).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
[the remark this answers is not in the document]
AdamGrzybkowski: That would make it impossible to use `FlowRow`. I can extract the `•` to `strings.xml` though - I didn't do it like that because when I looked, I noticed it would usually be defined like that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
```diff
@@ -0,0 +1,107 @@
...
+                state = model.paymentStatus
+            )
+        }
+    }
+}
+
+data class BookingSummary(
+    val date: String,
+    val name: String,
+    val staff: String,
```
irfano: I also noticed this is not staff's name. It's customer's name.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
```diff
@@ -0,0 +1,107 @@
...
+    Column(
+        horizontalAlignment = Alignment.Start,
+        modifier = modifier
+            .padding(16.dp)
+    ) {
+        Text(
+            text = model.date,
+            color = MaterialTheme.colorScheme.onSurface,
+            style = MaterialTheme.typography.bodyLarge.copy(fontWeight = FontWeight.Medium),
+        )
```
AdamGrzybkowski, line 37: FYI: I've left a comment in Figma to ask about the styles. The ones in the design don't match the typography defined in the app, so it's a bit tricky to make it exactly the same without overriding individual values.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
AdamGrzybkowski: That's correct, Thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
AdamGrzybkowski: >That way, the UI only needs the tag’s string resource, not the types.

Not just that, we will also need a background color. I'm waiting for a confirmation whether the `AttendanceStatus` tag will change color for some states, but the payment one will for sure.

>What do you think about moving AttendanceStatus and its text() logic into the view model?

If the Booking tags were to be reused in other parts of the app (dashboard cards, for example) we would have to extract or copy that logic. With the way it is now, those small UI components can be directly plugged in with a simple state passed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
AdamGrzybkowski: Fixed in https://github.com/woocommerce/woocommerce-android/pull/14649/commits/0695ecc7602128099eae48fc278d39f33628c849

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/AttendanceStatusTag.kt
AdamGrzybkowski: To add to the above: I have intentionally put all the components in this PR outside the `details` package because they will be shared between the list and the details screen.

irfano: That makes sense then. Thanks for the explanation.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingSummary.kt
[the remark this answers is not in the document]
irfano: > I can extract the • to strings.xml though - I didn't do it like that because

I agree. extracting only `•` doesn’t add any value, so there’s no need.

I was just concerned it might cause issues in RTL languages, but in my testing I didn’t see any problems, as shown below.

image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
irfano: I hadn’t considered the color changes. I thought the only difference was the text, which is why I suggested using `WCTag`. So LGTM, thanks for the explanation!

irfano (approved):
LGTM! 👍🏻 I’m approving, but not merging since you mentioned you’re still waiting on design feedback.

AdamGrzybkowski:
Thanks @irfano - I will merge it and apply the changes (if any) when I get back the answers.

PR #14655, opened 2025-09-24
[WOOMOB-1344] Create BookingCustomerDetails composable

AdamGrzybkowski:
Closes WOOMOB-1344

### Description

This PR implements the `BookingCustomerDetailsComposable`. The actions will be implemented separately when I get the confirmation for the behavior.

### Steps to reproduce

1. Log in.

### Testing information

Test light/dark theme and different font/display scaling

The above

### Images/gif

|  |  |
|---|---|
| Screenshot_20250924_131034 | Screenshot_20250924_131042 |
| Screenshot_20250924_131058 | Screenshot_20250924_131026 |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
```diff
@@ -0,0 +1,155 @@
...
+            Column(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(start = 16.dp, end = 16.dp, bottom = 11.dp)
+            ) {
+                model.billingAddressLines.forEach { line ->
+                    Text(
+                        text = line,
+                        style = MaterialTheme.typography.labelLarge,
+                        color = MaterialTheme.colorScheme.surfaceVariant,
```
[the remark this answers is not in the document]
AdamGrzybkowski: I agree with this comment, but I followed Irfano's previous PR. The `onSurfaceVariant` is not a good fit when compared with Figma.

The `onSurface` is the same as `onSurfaceVariant`:

Screenshot 2025-09-24 at 13 55 10

Thoughts? Should we update the color in the theme?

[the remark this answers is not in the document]
hichamboushaba: Yes I agree, we shouldn't use `surfaceVariant` here. Our M3 color scheme was just a migration of the M2 based one to try to achieve a consistent look accross the app while we use both.

Looking at the migration [document](https://developer.android.com/develop/ui/compose/designsystems/material2-material3#m3_16), they say that `ContentAlpha.medium` from M2 should be replaced by `onSurfaceVariant`, so I think we need to update our color scheme, I suggest we test the following change:

```patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt	(revision f5f77c7a8bb9fec200609d2b11f954d1800cd87f)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/theme/WooColors.kt	(date 1758792968338)
@@ -68,7 +68,7 @@
         surface = colorResource(R.color.color_surface),
         onSurface = colorResource(R.color.color_on_surface),
         surfaceVariant = colorResource(R.color.color_surface_variant),
-        onSurfaceVariant = colorResource(R.color.color_on_surface),
+        onSurfaceVariant = colorResource(R.color.color_on_surface_medium),
         outline = colorResource(R.color.divider_color),
         inverseOnSurface = colorResource(R.color.color_surface),
         inverseSurface = colorResource(R.color.color_on_surface),
@@ -105,7 +105,7 @@
         surface = colorResource(R.color.color_surface),
         onSurface = colorResource(R.color.color_on_surface),
         surfaceVariant = colorResource(R.color.color_surface_variant),
-        onSurfaceVariant = colorResource(R.color.color_on_surface),
+        onSurfaceVariant = colorResource(R.color.color_on_surface_medium),
         outline = colorResource(R.color.divider_color),
         inverseOnSurface = colorResource(R.color.color_surface),
         inverseSurface = colorResource(R.color.color_on_surface),
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
```diff
@@ -0,0 +1,155 @@
...
+                    )
+                }
+            }
+            HorizontalDivider(thickness = 0.5.dp)
+        }
+    }
```
hichamboushaba: np, I suggest grouping this area in a parent `Column`, this will just make it clear it's a single section, and aslo you'll get rid of the non-symmetric paddings (`top` for one view, and `bottom` for the other will be replaced with a `vertical` padding), WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
```diff
@@ -0,0 +1,155 @@
...
+    trailingIcon: (@Composable () -> Unit)? = null
+) {
+    Column(
+        modifier = modifier
+    ) {
+        Row(
+            verticalAlignment = Alignment.CenterVertically,
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(horizontal = 16.dp, vertical = 11.dp)
```
hichamboushaba: I don't think we use `11` as padding anywhere in the app, even if Figma says `11`, we round it to the closest even number, and the most common dimensions are all multiples of `4`, so in this case I suggest replacing this with `12`, also for the billing address section.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
```diff
@@ -0,0 +1,155 @@
...
+        )
+    }
+}
+
+data class BookingCustomerDetailsModel(
+    val name: String,
+    val email: String,
+    val phone: String,
+    val billingAddressLines: List<String>,
+)
```
hichamboushaba, line 131: We already have a customer model `Order.Customer`, I think most probably we'll use that other model, as we'll have the `order_id` from the booking that we'll use to fetch all data about the booking.

What do you think? Should we use it directly here instead of adding a new one?

hichamboushaba (commented):
Nice work @AdamGrzybkowski 👍, It works well, I left have some comments for discussion.

> The actions will be implemented separately when I get the confirmation for the behavior.

For the actions, I think we confirmed on one of the calls that we'll use the same actions we use in Order Details, check the `OrderDetailCustomerInfoView` and `OrderCustomerHelper` classes, maybe with some small refactoring we can extract all actions to `OrderCustomerHelper` and use it in both places 🤔

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
AdamGrzybkowski: TIL - thanks I wasn't aware of that, and I got used to following the designs.

AdamGrzybkowski:
>For the actions, I think we confirmed on one of the calls that we'll use the same actions we use in Order Details, check the OrderDetailCustomerInfoView and OrderCustomerHelper classes, maybe with some small refactoring we can extract all actions to OrderCustomerHelper and use it in both places 🤔l

I wasn't aware of that. I've got a different answer when I asked here lxixW6aBBNkCd9ooBK70Z1-fi-362_15368#1439851961.

I'm okay with both, but I would do them in a separate PR anyway to keep the PRs small.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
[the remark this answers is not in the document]
AdamGrzybkowski: Thanks, it looks good:

| light | dark |
|---|---|
| Screenshot_20250925_153108 | Screenshot_20250925_153100 |

Changes applied in https://github.com/woocommerce/woocommerce-android/pull/14655/commits/b4b5a1e1a113e35459b430646089159ee132d48c

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
AdamGrzybkowski: Makes sense, done in https://github.com/woocommerce/woocommerce-android/pull/14655/commits/0295d81c960a17f077f131996c39bb41b0c0ae14

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
AdamGrzybkowski: I've updated `BookingAppointmentDetails` from the other PR as well - https://github.com/woocommerce/woocommerce-android/pull/14655/commits/70b92e21cb69ad232f1a873028f750b8e1623743

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
AdamGrzybkowski: The idea for those was to create models that are 1:1 match with the Composable components. Having more data than needed could cause unnecessary recompositions. In this case, the data is not editable, so shouldn't be a problem.

Are you saying that we will use the `Order` for the booking? Or would we only use the `Order.Customer`?

hichamboushaba: I understand, but the maintenance burden will be higher IMO, as this will also require additional mappers to map from network/DB models to UI models.

> Having more data than needed could cause unnecessary recompositions

Why? Can you explain further? Even if the data was editable, why having more unused fields would cause recompositions?

> Are you saying that we will use the `Order` for the booking? Or would we only use the `Order.Customer`?

I am just speculating for now, as the API schema is not very clear now, but my understanding from the discussions with the team is that the `Order` will be needed for many details about the booking, so we will most probably get the order (or fetch it) when opening the booking details, and with the order, we'll have the `Order.Customer` object.
Just an early confirmation for this can be seen in the next-admin code.

hichamboushaba (approved):
Nice work @AdamGrzybkowski, I'm pre-approving, and I'll leave to you to decide whether to use `Order.Customer` directly or to add an additional model.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
hichamboushaba: Actually thinking more about this, and looking at the other composables you added, I'm beginning to better understand your perspective.

So I'd say let's keep this model, and even if we use `Order.Customer` behind the scenes, we'll map it to the new UI model somewhere.

PR #14656, opened 2025-09-24
[CIAB Bookings] Add pagination logic

hichamboushaba:
Closes WOOMOB-1359

### Description
This PR adds logic to handle the pagination for bookings, to keep things simple, I followed the same approach we used for the other Compose screens, so I added a new component `BookingListHandler` that encapsulates the pagination logic, for now it supports basic fetching, with a stub function for search, and when supporting sorting and filtering, we'll update it as needed.

### Steps to reproduce
1. Use a CIAB site.
4. Create at least 30 bookings (Sorry, but I'm not aware of any way to generate them, so you need to do it manually, the only tip I can share to make this easier, is that when adding a booking from `wp-admin`, on the second step where we choose the date and duration, you can command + click the "Add booking" button to offload the creation to a new tab, and stay on the screen to create another one, then repeat)
5. Optionally: modify the code to reduce the `BookingListHandler#PAGE_SIZE` to a smaller number, just make sure the number is big enough to fill the screen to make sure the basic Compose logic we use in `InfiniteListHandler` still works.
6. Open the bookings tab.

### Testing information
- Confirm the bookings are fetched, and pagination works.
- Confirm that we show the first page from the DB cache when re-opening the screen, while refreshing the data in background.

The above.

### Images/gif

https://github.com/user-attachments/assets/01e8de00-8612-4aa4-8682-2ad481ac1d13

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListHandler.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.ui.bookings
 
+import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -15,7 +16,8 @@ class BookingListHandler @Inject constructor(
     private val bookingsRepository: BookingsRepository
 ) {
     companion object {
-        private const val PAGE_SIZE = 25
+        @VisibleForTesting
+        const val PAGE_SIZE = 25
     }
 
     private val mutex = Mutex()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListHandler.kt
@@ -1,0 +1,88 @@
+package com.woocommerce.android.ui.bookings
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.inject.Inject
+
+class BookingListHandler @Inject constructor(
+    private val bookingsRepository: BookingsRepository
+) {
+    companion object {
+        private const val PAGE_SIZE = 25
+    }
+
+    private val mutex = Mutex()
+    private var page = MutableStateFlow(1)
+    private val canLoadMore = AtomicBoolean(false)
+
+    private val searchQuery = MutableStateFlow<String?>(null)
+    private val searchResults = MutableStateFlow(emptyList<Booking>())
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val bookingsFlow: Flow<List<Booking>> = combine(searchQuery, page) { query, page ->
+        if (query == null) {
+            bookingsRepository.observeBookings(limit = page * PAGE_SIZE)
+        } else {
+            searchResults
+        }
+    }.flatMapLatest { it }
+
+    suspend fun loadBookings(
+        searchQuery: String? = null,
+        forceRefresh: Boolean = false,
+    ): Result<Unit> = mutex.withLock {
+        // Reset pagination attributes
+        page.value = 1
+        canLoadMore.set(true)
+
+        this.searchQuery.value = searchQuery
+        return if (searchQuery == null) {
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -19,30 +20,40 @@ import javax.inject.Singleton
...
     private val bookingsDao: BookingsDao,
+    private val headersParser: HeadersParser,
     private val coroutineEngine: CoroutineEngine,
 ) {
     suspend fun fetchBookings(
         site: SiteModel,
         perPage: Int = BookingsRestClient.DEFAULT_PER_PAGE,
         page: Int = 1
-    ): WooResult<List<BookingEntity>> {
+    ): WooResult<Boolean> {
```
hichamboushaba, line 30: I modified this to return just a Boolean indicating whether we have more pages or not, I though about having a separate class that encapsulates both the list of items and this info, but given we don't need, and we will use the DB as the source of truth, I went for the simpler solution.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+import androidx.compose.material3.pulltorefresh.PullToRefreshBox
+import androidx.compose.material3.pulltorefresh.rememberPullToRefreshState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.livedata.observeAsState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.woocommerce.android.ui.compose.component.InfiniteListHandler
+
+@Composable
+fun BookingListScreen(viewModel: BookingListViewModel) {
```
hichamboushaba, line 26: Just a stub screen to test the functionality for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
```diff
@@ -0,0 +1,33 @@
...
+        } else {
+            Result.success(result.model!!)
+        }
+    }
+
+    fun observeBookings(limit: Int? = null): Flow<List<Booking>> =
+        bookingsStore.observeBookings(selectedSite.get(), limit)
+}
+
+typealias Booking = org.wordpress.android.fluxc.persistence.entity.BookingEntity
```
hichamboushaba, line 33: We will probably need a domain model for the feature, so given the DB entity fields are not final (depending on the final schema that the bookings plugin will eventually support p1756479361447279-slack-C09224W4HAA), I just went for a temporary `typealias` here, let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingListHandler.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    private val searchQuery = MutableStateFlow<String?>(null)
+    private val searchResults = MutableStateFlow(emptyList<Booking>())
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val bookingsFlow: Flow<List<Booking>> = combine(searchQuery, page) { query, page ->
+        if (query == null) {
+            bookingsRepository.observeBookings(limit = page * PAGE_SIZE)
+        } else {
+            searchResults
```
hichamboushaba, line 35: Similar to the other screens, the plan here is to avoid using the cache when searching by text, as it will be tricky to support pagination with it.

Post MVP we can decide whether we want to support searching in the cache or not, but IMO given CIAB sites will eventually be fast enough, this won't be needed.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooResult.kt
```diff
@@ -6,7 +6,10 @@ import org.wordpress.android.fluxc.network.rest.Header
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType.GENERIC_ERROR
 import org.wordpress.android.fluxc.store.Store
 
-data class WooPayload<T>(val result: T? = null) : Payload<WooError>() {
+data class WooPayload<T>(
+    val result: T? = null,
+    val headers: List<Header> = emptyList()
+) : Payload<WooError>() {
```
JorgeMucientes, line 12: Thanks for simplifying this adding the header.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
JorgeMucientes: Sounds good to me!

JorgeMucientes (approved):
Excellent work @hichamboushaba code looks good and I didn't find any issues while testing. 🥇

PR #14657, opened 2025-09-25
[WOOMOB-1378] "Trash Product" option invisible after creating the product

AdamGrzybkowski:
Closes WOOMOB-1378

### Description

From the issue description:

>After creating a new product, the “Trash product” button isn’t visible. It only appears after navigating back and reopening the product.

This is cause becase the `trashEnabled` destination argument defaulted to false when the fragment was opened from the bottom sheet. We can't simply change it to `true` because the same bottom sheet is used in other places, and we can only enable that option when the product details screen is opened from the product list screen.
Adding the same value as a bottom sheet destination argument would be an option, but that would create a long chain of passing arguments, because the above-mentioned bottom sheet is opened from yet another bottom sheet.  It would be a pain to write and maintain.

As proposed by @hichamboushaba this PR removed the destination argument completely, and instead checked whether the product list is in the backstack. 

### Steps to reproduce

1. Open the product list screen
2. Tap on the FAB to create a new product
3. Tap `Add manually`
4. Pick a product type
5. Tap publish
6. Tap on the `...` menu icon 
7. Verify `Trash product` option visibility

### Testing information
The above steps are good to reproduce the bug and verify it's fixed. 

Additionally, it's good to test other entry points for the Product details screen. One example is shown below, where the product is created from the onboarding checklist. 

1. New product from product list.
2. New product from onboarding checklist.
3. Opening a product from the product list and verifying the option is visible.

### Images/gif

Bug

| Before | After |
|---|---|
| product_list_flow_before | product_list_flow_after |

Regression test (should be the same):

| Before | After |
|---|---|
| onboarding_flow_before | onboarding_flow_after |
| existing_product_before | existing_product_after |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailFragment.kt
@@ -132,8 +132,7 @@ class ProductDetailFragment :
     override val activityAppBarStatus: AppBarStatus
         get() = AppBarStatus.Hidden
 
-    @Inject
-    lateinit var crashLogging: CrashLogging
+    @Inject lateinit var crashLogging: CrashLogging
 
     private val productsCommunicationViewModel: ProductsCommunicationViewModel by activityViewModels()
 
@@ -400,7 +399,6 @@ class ProductDetailFragment :
                 is LaunchUrlInChromeTab -> ChromeCustomTabUtils.launchUrl(requireContext(), event.url)
                 is Event.LaunchUrlInAuthenticatedWebView -> findNavController(R.id.nav_host_fragment_main)
                     .navigateSafely(NavGraphMainDirections.actionGlobalAuthenticatedWebViewFragment(event.url))
-
                 is RefreshMenu -> toolbarHelper.setupToolbar()
 
                 is TrashProduct -> {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailFragment.kt
@@ -132,7 +132,8 @@ class ProductDetailFragment :
     override val activityAppBarStatus: AppBarStatus
         get() = AppBarStatus.Hidden
 
-    @Inject lateinit var crashLogging: CrashLogging
+    @Inject
+    lateinit var crashLogging: CrashLogging
 
     private val productsCommunicationViewModel: ProductsCommunicationViewModel by activityViewModels()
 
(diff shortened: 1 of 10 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailFragment.kt
```diff
@@ -154,6 +155,9 @@ class ProductDetailFragment :
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
+        val isTrashEnabled = findNavController().previousBackStackEntry?.destination?.id == R.id.products
```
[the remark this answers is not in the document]
hichamboushaba: I wonder if using something like `BottomNavigationPosition.PRODUCTS.id` won't be better here 🤔

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/products/details/ProductDetailViewModelTest.kt
```diff
@@ -1434,4 +1427,63 @@ class ProductDetailViewModelTest : BaseUnitTest() {
 
     private val productsDraft
         get() = viewModel.productDetailViewStateData.liveData.value?.productDraft
+
+    @Test
+    fun `given add new product flow, when trash action becomes possible, then trashOption remains hidden`() =
```
hichamboushaba, line 1432: not a blocker, but I think we can make this test name clearer, the trash button can be available in the product creation flow, the important bit that controls whether the trash button is hidden not, is whether the product was created on the server or not yet. The current name makes it look like the trash button will always be hidden in the creation flow.

hichamboushaba (approved):
Nice work @AdamGrzybkowski 💯, pre-approving, I left two minor comments, but nothing blocking.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/products/details/ProductDetailViewModelTest.kt
AdamGrzybkowski: This test doesn't verify that flow, but rather simply checks based on the initial conditions. In this case, for the `AddNewProduct` flow, the trash should be invisible. 

That said, I think it's good to test the flow mentioned when the product is stored so I've added it here - https://github.com/woocommerce/woocommerce-android/pull/14657/commits/f848f65192878df77d14e2fe29385f935ba08f5d

hichamboushaba: > This test doesn't verify that flow, but rather simply checks based on the initial conditions. In this case, for the AddNewProduct flow, the trash should be invisible.

Yes I'm aware, I suggested just modifying the test name, not the test itself 🙂

AdamGrzybkowski: Got it, what about: `given add new product flow when product not yet created on server then trashOption is hidden`?

hichamboushaba: I like it, it's clear. Thanks @AdamGrzybkowski

PR #14658, opened 2025-09-25
[CIAB Bookings] List UI

hichamboushaba:
Part of WOOMOB-1358

> [!TIP]
> For an easier review, exclude the commit af89c46, this commit just moves the list specific files to a new package, and this caused Github to mark some files as new files.

### Description
This PR adds basic UI for the bookings list, it includes the following changes:
- Introduces a new ViewState class, that uses the `BookingSummaryModel` for the items.
- Handles basic mapping from the `Booking` class to the view state.
- Updates the UI in the BookingListScreen to use the `BookingSummary` composable.

The navigation to booking details is not included yet, it will be added in another PR.

### Steps to reproduce
1. Use a CIAB site.
4. Create some bookings.
5. Open the Bookings tab in the app.
6. Check the UI.

### Testing information
- Confirm the UI works as expected.
- Confirm the booking `id`, `date` and `payment status` are shown correctly, the other details are mocked now.

The above.

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20250926_095810 | Screenshot_20250926_095905 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
@@ -24,8 +24,19 @@ fun BookingPaymentStatusTag(
     )
 }
 
-enum class BookingPaymentStatus {
-    UNPAID, PENDING_CONFIRMATION, CONFIRMED, PAID, CANCELLED, COMPLETE
+enum class BookingPaymentStatus(val key: String) {
+    UNPAID("unpaid"),
+    PENDING_CONFIRMATION("pending-confirmation"),
+    CONFIRMED("confirmed"),
+    PAID("paid"),
+    CANCELLED("cancelled"),
+    COMPLETE("complete");
+
+    companion object {
+        fun fromKey(key: String): BookingPaymentStatus {
+            return entries.firstOrNull { it.key == key } ?: UNPAID
+        }
+    }
 }
 
 @Composable
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
@@ -10,6 +10,7 @@ import com.woocommerce.android.ui.base.TopLevelFragment
 import com.woocommerce.android.ui.compose.composeView
 import com.woocommerce.android.ui.main.AppBarStatus
 import dagger.hilt.android.AndroidEntryPoint
+import org.wordpress.android.util.ToastUtils
 
 @AndroidEntryPoint
 class BookingListFragment : TopLevelFragment() {
@@ -27,6 +28,21 @@ class BookingListFragment : TopLevelFragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         return composeView {
             BookingListScreen(viewModel)
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        handleEvents()
+    }
+
+    private fun handleEvents() {
+        viewModel.event.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is BookingListViewModel.NavigateToBookingDetails -> {
+                    // TODO navigate to booking details screen
+                    ToastUtils.showToast(requireContext(), "Navigate to booking ${event.bookingId}")
+                }
+            }
         }
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
@@ -1,4 +1,4 @@
-package com.woocommerce.android.ui.bookings
+package com.woocommerce.android.ui.bookings.list
 
 import android.os.Bundle
 import android.view.LayoutInflater
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
@@ -7,23 +7,29 @@ import android.view.ViewGroup
 import androidx.fragment.app.viewModels
 import com.woocommerce.android.R
 import com.woocommerce.android.ui.base.TopLevelFragment
+import com.woocommerce.android.ui.base.UIMessageResolver
 import com.woocommerce.android.ui.compose.composeView
 import com.woocommerce.android.ui.main.AppBarStatus
+import com.woocommerce.android.viewmodel.MultiLiveEvent
 import dagger.hilt.android.AndroidEntryPoint
 import org.wordpress.android.util.ToastUtils
+import javax.inject.Inject
 
 @AndroidEntryPoint
 class BookingListFragment : TopLevelFragment() {
     override val activityAppBarStatus: AppBarStatus
         get() = AppBarStatus.Hidden
 
+    private val viewModel: BookingListViewModel by viewModels()
+
+    @Inject
+    lateinit var uiMessageResolver: UIMessageResolver
+
     override fun getFragmentTitle() = getString(R.string.bookings_tab_title)
     override fun shouldExpandToolbar(): Boolean = false
     override fun scrollToTop() {
         return
     }
-
-    private val viewModel: BookingListViewModel by viewModels()
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         return composeView {
@@ -42,6 +48,7 @@ class BookingListFragment : TopLevelFragment() {
                     // TODO navigate to booking details screen
                     ToastUtils.showToast(requireContext(), "Navigate to booking ${event.bookingId}")
                 }
+                is MultiLiveEvent.Event.ShowSnackbar -> uiMessageResolver.showSnack(event.message)
             }
         }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
@@ -1,6 +1,5 @@
 package com.woocommerce.android.ui.bookings.list
 
-import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -13,7 +12,6 @@ import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
-import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.material3.pulltorefresh.rememberPullToRefreshState
@@ -51,7 +49,9 @@ fun BookingListScreen(state: BookingListViewState) {
             state.bookings.isNotEmpty() -> {
                 BookingList(
                     state = state,
-                    modifier = Modifier.padding(paddingValues)
+                    modifier = Modifier
+                        .padding(paddingValues)
+                        .fillMaxSize()
                 )
             }
 
@@ -94,7 +94,7 @@ private fun BookingList(
         LazyColumn(
             state = listState,
             modifier = Modifier
-                .background(color = MaterialTheme.colorScheme.surface)
+                .fillMaxSize()
         ) {
             itemsIndexed(state.bookings) { _, booking ->
                 BookingSummary(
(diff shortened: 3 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>BookingEntityConverters</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -1,31 +1,43 @@
...
     primaryKeys = ["id", "localSiteId"]
 )
+@TypeConverters(BookingEntityConverters::class)
 data class BookingEntity(
     val id: RemoteId,
     val localSiteId: LocalId,
-    val start: Long,
-    val end: Long,
+    val start: Instant,
+    val end: Instant,
```
hichamboushaba, line 19: This wasn't really necessary for this PR, but since I needed to format dates and it wasn't clear what they referred to when shown as Long, I decided to convert them to `java.time.Instant`. Please share your thoughts on this change.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
```diff
@@ -7,26 +7,49 @@ import android.view.ViewGroup
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        handleEvents()
+    }
+
+    private fun handleEvents() {
+        viewModel.event.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is BookingListViewModel.NavigateToBookingDetails -> {
+                    // TODO navigate to booking details screen
+                    ToastUtils.showToast(requireContext(), "Navigate to booking ${event.bookingId}")
```
hichamboushaba, line 49: The PR is already big enough, so I'll handle this on a separate PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -16,35 +18,36 @@ class BookingListViewModel @Inject constructor(
...
 
-    private fun fetchBookings(initialLoadingState: LoadingState) {
+    private fun fetchBookings(initialLoadingState: BookingListViewState.LoadingState) {
         bookingsFetchJob?.cancel()
         bookingsFetchJob = launch {
             loadingState.value = initialLoadingState
             bookingListHandler.loadBookings(forceRefresh = true)
                 .onFailure {
-                    // Show error message
+                    triggerEvent(MultiLiveEvent.Event.ShowSnackbar(R.string.bookings_fetch_error))
```
hichamboushaba, line 48: We might need some better handling here depending on whether we have local cache or not, but this can wait for future PRs.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
```diff
@@ -0,0 +1,43 @@
...
+        Idle, Loading, Refreshing, Appending
+    }
+}
+
+data class BookingListItem(
+    val id: Long,
+    val summary: BookingSummaryModel
+)
+
+fun Booking.toUiModel(): BookingListItem {
```
hichamboushaba, line 28: I have the mapping just as an extension for now, we can, if needed, move this to a Mapper class.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
```diff
@@ -0,0 +1,44 @@
...
+    val dateFormatter = DateTimeFormatter.ofLocalizedDateTime(
+        FormatStyle.MEDIUM,
+        FormatStyle.SHORT
+    ).withZone(ZoneId.systemDefault())
```
hichamboushaba, line 32: For the date time formatting, there is a descrepancy between web and the app now, the web uses `UTC` as you can see from the commit 63b5349 in 942-ghe-Automattic/next-admin, while the app will use the local device timezone.

I'm not sure what the web is doing offers the best user experience, as a user, if I'm checking a list of bookings, I will most probably want to know when they will start according to my timezone, and that's also the same pattern we [use](https://github.com/woocommerce/woocommerce-android/blob/0caeb4712346016b0dfd37fbc16929a69c6888f7/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailOrderStatusView.kt#L67) in order details, WDYT? If agreed, we can maybe raise this to the web team, otherwise I can update our logic to match theirs.

AdamGrzybkowski: I would be very confused if an app used a non-local timezone, so yeah, I think what you did makes the most sense.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -0,0 +1,150 @@
...
+                .background(color = MaterialTheme.colorScheme.surface)
+        ) {
+            itemsIndexed(state.bookings) { _, booking ->
+                BookingSummary(
+                    model = booking.summary,
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .clickable(onClick = { state.onBookingClick(booking.id) })
+                )
+                HorizontalDivider()
```
AdamGrzybkowski: Based on the Figma, this one should have a start padding.
Screenshot 2025-09-26 at 12 54 19

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
AdamGrzybkowski: I think that makes sense! Especially without a domain model at this point.

AdamGrzybkowski (commented):
The code changes look good to me but at the moment I can't test it due to issues with CIAB sites. I will approve once I fix the issues.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Good eye, fixed in d894522

hichamboushaba:
> The code changes look good to me but at the moment I can't test it due to issues with CIAB sites. I will approve once I fix the issues.

@AdamGrzybkowski just for information, I'm having issues with CIAB sites too, so what I do now for the development: I installed the bookings plugin on a regular site, and I modified the `SiteModel#isCiabSite` to return `true` in the app.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    // TODO replace the mocked details with real data when available from the API
+    return BookingListItem(
+        id = id.value,
+        summary = BookingSummaryModel(
+            date = dateFormatter.format(start),
+            name = "Women’s Haircut",
+            customerName = "Margarita Nikolaevna",
+            attendanceStatus = AttendanceStatus.BOOKED,
+            paymentStatus = BookingPaymentStatus.valueOf(status.uppercase())
```
AdamGrzybkowski: ⚠️ The app can crash here - some of the status are not availalble in `BookingPaymentStatus`. I was able to crash the app with `Pending confirmation`. 

I'm not sure if we have a final list of states? But if not maybe we can do a safer mapping here?

AdamGrzybkowski (commented):
The UI looks good - one thing I noticed, when I only had one booking, is that it's hard to pull to refresh because it only works on the list items - the PTR box doesn't use the whole available space. 

bookings_PTR

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
hichamboushaba: Yes I thought about this when I added this logic, but decided to go for the unsafe approach to catch the issues early, with a safer approach, we'll need some sort of `Unknown` status, which I'd like to avoid unless necessary.

Can you share what the API returned when the app crashed? We have a status for Pending Confirmation, but maybe the actual API response is different from what we have.

hichamboushaba: Nevermind, I just found out I can set the status manually in wp-admin 😅. I'll update the key to match the API one.

AdamGrzybkowski: It's `pending-confirmation`. Here's the error message:
```
java.lang.IllegalArgumentException: No enum constant com.woocommerce.android.ui.bookings.compose.BookingPaymentStatus.PENDING-CONFIRMATION
```

I agree it's good to catch early! I was worried we would get a lot of those crashes, and because it's in the DB it's hard to restore it without modifying the data manually or clearing the app.

hichamboushaba: The commit [e9bbc61](https://github.com/woocommerce/woocommerce-android/pull/14658/commits/e9bbc6137d31ba1a33ff6c8369c1a930ddb816da) should fix this, I updated the keys to match the ones in the plugin code woocommerce/woocommerce-bookings/blob/d2bca88b26687eee7505a49eac730d08cd86bd2f/includes/wc-bookings-functions.php#L131-L136

hichamboushaba:
Thanks @AdamGrzybkowski for the review, ready for another round, I update the pull to refresh box to use `fillMaxSize` which should fix the issue about the pull to refresh.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
```diff
@@ -24,8 +24,19 @@ fun BookingPaymentStatusTag(
...
 
-enum class BookingPaymentStatus {
-    UNPAID, PENDING_CONFIRMATION, CONFIRMED, PAID, CANCELLED, COMPLETE
+enum class BookingPaymentStatus(val key: String) {
+    UNPAID("unpaid"),
+    PENDING_CONFIRMATION("pending-confirmation"),
+    CONFIRMED("confirmed"),
+    PAID("paid"),
+    CANCELLED("cancelled"),
+    COMPLETE("complete");
```
AdamGrzybkowski, line 33: I think that's fine for now, but we are leaking server API design to view-level classes. I think it would be better to handle this elsewhere. 

An alternative without a domain object would be to have a custom type mapper, like you did with an Instant type and returning an enum instead of a raw string. WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -51,7 +51,9 @@ fun BookingListScreen(state: BookingListViewState) {
             state.bookings.isNotEmpty() -> {
                 BookingList(
                     state = state,
-                    modifier = Modifier.padding(paddingValues)
+                    modifier = Modifier
+                        .padding(paddingValues)
+                        .fillMaxSize()
```
AdamGrzybkowski, line 54: This is not enough - the child of the PTR box needs to take the whole space as well. Something like this added to `LazyColumn` works:
```kotlin
            modifier = Modifier
                .fillMaxSize()
                .background(color = MaterialTheme.colorScheme.background)
```

hichamboushaba: Why? it works fine for me without this:

https://github.com/user-attachments/assets/8db4243f-5202-4617-b193-408b737fcd74

Your above suggested change doesn't have any effect.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
hichamboushaba: > An alternative without a domain object would be to have a custom type mapper, like you did with an Instant type and returning an enum instead of a raw string. WDYT?

Yes, good suggestion, I'll add this in a separate PR if that's OK for you.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
AdamGrzybkowski: This only works because you start your gesture at the very top, where the list items are. If you start it on the "gray" area, below PTR won't work.


https://github.com/user-attachments/assets/c3ff4145-f0e5-410f-9b02-79ea242d370a

AdamGrzybkowski (approved):
Thanks for the changes, approving because the crash is fixed :shipit: 

I would make the PTR work on the whole screen, but if you think it's enough if it works only when the gesture starts on one of the list elements, feel free to skip the comment.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Oh, sorry I didn't understand your remark before, I amended the last commit to address this.

PR #14660, opened 2025-09-26
[WOOMOB-1345] - Attendance status section in booking details

AdamGrzybkowski:
Closes WOOMOB-1345

### Description

This PR implements the clickable attendance section. On row click, a bottom sheet is presented with all possible states to update the booking. For now, the state is only updated in the VM's in-memory view state.

### Steps to reproduce

1. Log in.
3. Scroll down to the attendance section
4. Verify the UI
5. Tap on the row
6. Confirm the bottom sheet is visible
7. Verify the UI of the bottom sheet
8. Select a new state
9. Verify the new state is selected

### Testing information

Standard stuff: light/dark theme, different font/display scaling. 

The above

### Images/gif

https://github.com/user-attachments/assets/bec96775-eb19-4569-b47c-ca4c41dbe156


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusBottomSheet.kt
```diff
@@ -0,0 +1,152 @@
...
+
+@Composable
+private fun BookingAttendanceStatusSelection(
+    onSelect: (AttendanceStatus) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier
+            .fillMaxWidth()
+            .verticalScroll(rememberScrollState())
```
AdamGrzybkowski, line 64: With font scaling the content won't fit so we need to make it scrollable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusBottomSheet.kt
```diff
@@ -0,0 +1,152 @@
...
+            )
+        }
+    }
+}
+
+@Composable
+private fun AttendanceStatus.description(): String = when (this) {
+    AttendanceStatus.BOOKED -> R.string.booking_attendance_status_booked_desc
+    AttendanceStatus.CHECKED_IN -> R.string.booking_attendance_status_checked_in_desc
+    AttendanceStatus.CANCELLED -> R.string.booking_attendance_status_cancelled_desc
```
AdamGrzybkowski, line 121: The `cancelled` option is missing in Figma. For now, I've kept the temporary string resource and the icon.

WooCommerce/src/main/res/drawable/ic_attendance_checked_in.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="26dp"
+    android:height="20dp"
```
AdamGrzybkowski, line 3: This one has a different size after importing to AS. I've asked about it in Figma.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusBottomSheet.kt
```diff
@@ -0,0 +1,152 @@
...
+@Composable
+fun BookingAttendanceStatusBottomSheet(
+    onSelect: (AttendanceStatus) -> Unit,
+    onDismiss: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
+    val scope = rememberCoroutineScope()
+
+    WCModalBottomSheet(
```
AdamGrzybkowski, line 41: The drag handle has a lot of vertical padding, significantly more than in Figma, but keeping it for consistency.

WooCommerce/src/main/res/drawable/ic_attendance_checked_in.xml
hichamboushaba: I think that's expected if you look at the icons closely, the two other icons expand a little below the calendar, while this one doesn't, as you can see in Figma when you try to export them:

| | |
| - | - |
| Image | Image |

So I think we can ignore this, we just set all icons to the same width (In Figma we have 20px, so maybe 20dp), then the icons will have a slightly different height, but it would be OK given they are centered.

hichamboushaba: If you want, you can also try to add some padding to this icon in Figma before exporting it, but I don't think it would bring any advantage for this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceSection.kt
```diff
@@ -0,0 +1,98 @@
...
+    status: AttendanceStatus,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(modifier = modifier) {
+        BookingSectionHeader(R.string.booking_attendance_header)
+        Column(
+            modifier = Modifier.background(color = MaterialTheme.colorScheme.surfaceContainer)
+        ) {
+            HorizontalDivider(thickness = 0.5.dp)
```
hichamboushaba, line 37: Just a remark, I'm not sure about using dividers for the sections here (I know we used them for the other sections, but I just noticed it), Figma designs we have are iOS ones, if you look at order details screen in Android, we don't use dividers, we use `Cards` with some elevation to differentiate the sections from the screen background.

Do you think we should ask Wagner about this? Or accept the difference from order details?

hichamboushaba (approved):
Nice work @AdamGrzybkowski 💯

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceSection.kt
AdamGrzybkowski: Good question - I haven't noticed the difference. I can ask Wagner. 

>we use Cards with some elevation to differentiate the sections from the screen background.

Now that you mention I see the difference, but it looks like the elevation is not used consistently on the order screen 🤔 Is this on purpose? or simply a bug? 
Screenshot_20250929_091357

AdamGrzybkowski: I've left a comment in Figma and once we get the answer we can fix the whole screen in a separate PR.

hichamboushaba: > Now that you mention I see the difference, but it looks like the elevation is not used consistently on the order screen 🤔 Is this on purpose? or simply a bug?

I'd say it's just a bug.

WooCommerce/src/main/res/drawable/ic_attendance_checked_in.xml
hichamboushaba: > So I think we can ignore this, we just set all icons to the same width (In Figma we have 20px, so maybe 20dp), then the icons will have a slightly different height, but it would be OK given they are centered.

@AdamGrzybkowski a minor point, I think that this remark has been missed before merging the PR 🙂, we should have set all the icons to the same width, now this one is at 20dp, while the two other ones have 21dp. I think we can fix this on a a future PR that touches this screen.

AdamGrzybkowski: I did see this, but I knew we would have to update these anyway (the cancelled icon was not available) so I wanted to fix it later. 

>you can also try to add some padding to this icon in Figma before exporting it

To be honest, I'm not sure I can do it. I. I thought you needed a certain permission level to modify Figma files.

PR #14663, opened 2025-09-29
[CIAB] Bookings tabs - UI

hichamboushaba:
Part of WOOMOB-1360

### Description
This PR adds UI for the bookings tabs, the logic will be handled in separated PR.

### Steps to reproduce
1. Use a CIAB site.
4. Create some bookings.
5. Notice the tabs.

### Testing information
Confirm the UI looks good in both light and dark theme.

The above

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20250929_191727 | Screenshot_20250929_215941 |

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -75,6 +75,10 @@
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
+    <inspection_tool class="PreviewParameterProviderOnFirstParameter" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
     <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -75,10 +75,6 @@
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
-    <inspection_tool class="PreviewParameterProviderOnFirstParameter" enabled="true" level="ERROR" enabled_by_default="true">
-      <option name="composableFile" value="true" />
-      <option name="previewFile" value="true" />
-    </inspection_tool>
     <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
@@ -172,7 +172,7 @@ private fun BookingListPreview() {
         BookingListScreen(
             state = BookingListViewState(
                 contentState = BookingListContentState(
-                    bookings = List(3) {
+                    bookings = List(20) {
                         BookingListItem(
                             id = it.toLong(),
                             summary = com.woocommerce.android.ui.bookings.compose.BookingSummaryModel(
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
@@ -15,22 +15,19 @@ import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.PrimaryTabRow
 import androidx.compose.material3.Scaffold
-import androidx.compose.material3.Tab
 import androidx.compose.material3.Text
 import androidx.compose.material3.pulltorefresh.rememberPullToRefreshState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.res.colorResource
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import com.woocommerce.android.R
 import com.woocommerce.android.ui.bookings.compose.BookingSummary
 import com.woocommerce.android.ui.compose.component.InfiniteListHandler
 import com.woocommerce.android.ui.compose.component.Toolbar
+import com.woocommerce.android.ui.compose.component.WCPrimaryTabRow
 import com.woocommerce.android.ui.compose.component.WCPullToRefreshBox
 import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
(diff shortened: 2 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
```diff
@@ -9,22 +9,38 @@ import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
 
 data class BookingListViewState(
+    val contentState: BookingListContentState,
+    val tabState: BookingListTabState
+)
```
hichamboushaba, line 14: Feel free to share any thoughts on the refactoring of the `ViewState` here, my goal is to avoid having all the callbacks on the main view state, with this, we split the state to the various components of the screen and keep the callbacks in each one, and in the future, we can have a `filtersState` property too.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -94,6 +105,7 @@ private fun BookingList(
         LazyColumn(
             state = listState,
             modifier = Modifier
+                .background(MaterialTheme.colorScheme.surfaceContainer)
```
hichamboushaba: I noticed an issue with background color in dark theme, the window background color was used in the dividers, this fixes it.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -75,10 +75,14 @@
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
+    <inspection_tool class="PreviewParameterProviderOnFirstParameter" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
```
irfano: I guess you didn’t add this intentionally. If that’s the case, let’s revert the change.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -124,6 +133,34 @@ private fun BookingList(
     }
 }
 
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun BookingListTabs(
```
irfano: I have two notes about the design: we can set the unselected tab text to black, and make the selected tab’s bottom line longer to match the design.

|PR|Figma|
|-|-|
|image|image|

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -124,28 +137,62 @@ private fun BookingList(
...
+        containerColor = colorResource(id = R.color.color_toolbar),
+        modifier = modifier
+    ) {
+        BookingListTab.entries.forEach { tab ->
+            Tab(
+                selected = tab == tabState.selectedTab,
+                onClick = { tabState.onTabChanged(tab) },
+                text = {
+                    Text(
+                        text = tab.name,
```
irfano: `tab.name` is displaying the enum names. Let’s add string resources for each tab.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
irfano: I think something is wrong here. There are two `.background(` settings, and the divider colors are still an issue. 

image

I'm sharing a patch, and I believe this is how it should be. 

[Remove_redundant_background_color_in_BookingListScreen__The_`BookingListScreen`_LazyColumn.patch](https://github.com/user-attachments/files/22617857/Remove_redundant_background_color_in_BookingListScreen__The_.BookingListScreen._LazyColumn.patch)

#### After patch:

image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
irfano: I agree with your approach. 👍

irfano (approved):
Good job! It works as expected. I have a few minor design suggestions, please review them before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Yes, sorry I missed this, this was supposed just for the development 🤦

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Thanks for the patch @irfano, I missed that the issue was still present with my fix.
We can't use your patch because we need to keep the `fillMaxSize` to solve the issue discussed [here](https://github.com/woocommerce/woocommerce-android/pull/14658#discussion_r2387630114)

I did another attempt in the commit dde8c93f5b8b589bca9b46e1bafd8a1d8d34ef76 that fixes the background issue and still supports starting the pull to refresh gesture from anywhere in the screen:

https://github.com/user-attachments/assets/4b2b1df5-52f0-446b-aa21-5fc45a77f655

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Initially I thought we could benefit from staying consistent with M3 here, as we don't have many tabs in the app anyway.
But I think you are right, we need to stay consistent with the designs, I fixed this in the last commit aaaebd3ec61675ce2c0f69568ab39d07b6b14138, I introduce a new component `WCPrimaryTabRow` that achieves the wanted design to help keeping things consistent in other screens too:

Screenshot_20250930_141329 Screenshot_20250930_141343

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Fixed in 2aa6ce3ad62c74ac2400e85ea1deba6fde438839

hichamboushaba:
Thanks @irfano for the review, given I did multiple changes, if you can check the last changes before we merge this it would be great 🙇

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
irfano: 👍🏻  Works perfectly.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
irfano: Thanks for the improvement. 👍🏻 LGTM!

irfano:
Looks great! Thanks for the changes! 👍🏻

PR #14664, opened 2025-09-29
[CIAB] Bookings tabs - logic

hichamboushaba:
Closes WOOMOB-1360

### Description
This PR introduces logic for filtering bookings according to the selected tabs. I focused on developing the filtering system overall to prepare for future filter features, with tabs being just a particular example.

### Steps to reproduce
1. Use a CIAB site.
4. Create some bookings.
5. Switch between the bookings tabs.

### Testing information
- Confirm switching correctly filters the shown tabs.
- Confirm that cached tabs are shown correctly when switching tabs before the bookings are refreshed.

The above.

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -52,15 +55,25 @@ fun BookingListScreen(state: BookingListViewState) {
...
+
         Column(modifier = Modifier.padding(paddingValues)) {
             BookingListTabs(
                 tabState = state.tabState,
+                onTabChanged = {
+                    // Scroll to top when tab changes
+                    coroutineScope.launch {
+                        lazyListState.scrollToItem(0)
+                    }
+                },
```
hichamboushaba: I think this offers a better user experience, if you think otherwise please let me know.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import java.time.Instant
+
+sealed interface BookingsFilterOption {
+    data class DateRange(
+        val before: Instant?,
+        val after: Instant?
+    ) : BookingsFilterOption
+
+    data class Customer(val customerId: Long) : BookingsFilterOption
```
hichamboushaba, line 11: I just added this as an example of future filter types.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -52,23 +56,35 @@ class BookingListViewModel @Inject constructor(
...
-            bookingListHandler.loadBookings(forceRefresh = true)
-                .onFailure {
-                    triggerEvent(MultiLiveEvent.Event.ShowSnackbar(R.string.bookings_fetch_error))
-                }
-            loadingState.value = BookingListLoadingState.Idle
+    private fun monitorFilterChanges() {
+        launch {
+            selectedTab.collectLatest {
+                // Cancel any ongoing fetch or load more operations
+                bookingsFetchJob?.cancel()
```
hichamboushaba, line 66: `collectLatest` will cancel the previous fetch job, so technically we don't need this line, but I'm keeping it even if it's redundant as it's just clearer.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFiltersBuilder.kt
```diff
@@ -0,0 +1,28 @@
...
+        return when (this) {
+            BookingListTab.Today -> BookingsFilterOption.DateRange(
+                before = ZonedDateTime.now(clock).with(LocalTime.MAX).toInstant(),
+                after = ZonedDateTime.now(clock).with(LocalTime.MIN).toInstant()
+            )
+
+            BookingListTab.Upcoming -> BookingsFilterOption.DateRange(
+                before = null,
+                after = Instant.now(clock)
+            )
```
hichamboushaba, line 23: For `Upcoming` tab, the behavior here is different from next-admin, they set the date to "today at midnight", which is weird.
I'm planning to confirm the behavior of this with the product and update if needed.

Additionally, we could add logic to refresh bookings during `onStart`, ensuring the date updates each time the screen opens. This is important because the `filters` property is cached in `BookingListHandler` and could become outdated if the user has been away for a while.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
AdamGrzybkowski: I think that makes sense with the setup we have here, but I wonder if the "old" way with a fragment per tab wasn't better. This way, we don't have to refetch the data on each tab switch, and the user doesn't lose the scroll position when switching between tabs. 

Not saying we should go that way, just thinking out loud 😅

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFiltersBuilder.kt
AdamGrzybkowski: >Additionally, we could add logic to refresh bookings during onStart, ensuring the date updates each time the screen opens. This is important because the filters property is cached in BookingListHandler and could become outdated if the user has been away for a while.

That makes sense! `filters` will be cached as long as the app is active (not killed), right?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
```diff
@@ -0,0 +1,12 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings
+
+import java.time.Instant
+
+sealed interface BookingsFilterOption {
```
AdamGrzybkowski, line 5: ❓ I don't have a better suggestion, but I wonder if there's a better place for this, given it's used across many layers (rest, db, presentation)?

AdamGrzybkowski (approved):
Great work! 💯 

I've noticed one thing that is not related to this PR. All the tabs use the primary color. I would expect only the selected one to use it. 
Screenshot 2025-09-30 at 13 40 56

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: > I wonder if the "old" way with a fragment per tab wasn't better. This way, we don't have to refetch the data on each tab switch, and the user doesn't lose the scroll position when switching between tabs.

If we want to support all of this with the current setup we can, but I'm not sure it's worth the effort.
Personally, the better reusability of the code with the current setup outweighs the benefits of using multiple fragments here, and also, if we want to use multiple fragments, we'll have to have a `View` based top fragment.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFiltersBuilder.kt
hichamboushaba: > That makes sense! filters will be cached as long as the app is active (not killed), right?

Yes, they are cached in memory in `BookingListHandler`, which means they are cached as long as the ViewModel instance is active, so if the Activity gets recreated in low-memory occasions, it will be reset. But given we save the `BookingListTab` in the saved state, we'll correctly restore them even in this scenario.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
hichamboushaba: I thought about this, but I think we don't need to overengineer this by having multiple implementations per layer, and if we want to use the same one accross layers, then the only place we can define it is next to the `Store` to be available to all callers.
We can use a `typealias` if needed, but I'm not sure it brings any advantages, we'll just have a hard time figuring out a different name 😅.

For the current usage, I consider the `BookingListTab` as the UI model, and we convert it to `DateRange` using the `BookingListFiltersBuilder`, I'm not sure if it makes a lot of sense 😅.

hichamboushaba:
Thanks @AdamGrzybkowski for the review.

> I've noticed one thing that is not related to this PR. All the tabs use the primary color. I would expect only the selected one to use it.

Yes, I missed this in the UI PR, I'm working on fixing it there as Irfan already notified me about it.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
AdamGrzybkowski: Yes, agree on the overengineering side 👍

PR #14667, opened 2025-09-30
[CIAB] A minor refactoring to avoid leaking the booking status to UI layer

hichamboushaba:
### Description
As per the previous [discussion](https://github.com/woocommerce/woocommerce-android/pull/14658#discussion_r2387588859), this PR introduces a model in the DB layer that maps to the API status keys, which avoids leaking this information to the UI layer.
Please feel free to share any thoughts or suggestions for improvement here.

### Testing information
Code review should be enough, but small test to confirm there is no regression would be fine.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -32,12 +32,62 @@ data class BookingEntity(
...
+
+        data object Cancelled : Status {
+            override val key = "cancelled"
+        }
+
+        data object Complete : Status {
+            override val key = "complete"
+        }
+
+        data class Unknown(override val key: String) : Status
```
hichamboushaba, line 63: I know I mentioned I don't like having an `Unknown` value, but after thinking more about this and knowing how flexible WordPress is, I decided to go for the safest approach here.
We can revisit when next-admin implements logic for displaying badges, and see if we need to modify something to align, WDYT?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -32,12 +32,62 @@ data class BookingEntity(
     val parentId: Long,
     val personCounts: List<Long>?,
     val localTimezone: String
-)
+) {
+    sealed interface Status {
```
hichamboushaba, line 36: I named this `Status` and not `PaymentStatus` as this seems related more to the booking status and not just payment (for example `cancelled` is not really a payment status).

I wonder if we should rename the UI model too or not, WDYT?

hichamboushaba:
@AdamGrzybkowski while working on this, I noticed `AttendanceStatusTag` doesn't have the `Booking` prefix, should we add it for the sake of consistency?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
AdamGrzybkowski: At the time, I thought this was the order payment status, but it looks like it's not so yeah the name change makes sense.

It's still a bit confusing to me, because we can have two `cancelled` status tags next to each other (attendance and this one) 🤔

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
AdamGrzybkowski: Sounds good 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
```diff
@@ -24,39 +24,35 @@ fun BookingPaymentStatusTag(
...
-        BookingPaymentStatus.CANCELLED -> R.string.booking_payment_status_cancelled
-        BookingPaymentStatus.COMPLETE -> R.string.booking_payment_status_complete
-    }.let { stringResource(it) }
+        BookingPaymentStatus.Unpaid -> stringResource(R.string.booking_payment_status_unpaid)
+        BookingPaymentStatus.PendingConfirmation -> stringResource(R.string.booking_payment_status_pending_confirmation)
+        BookingPaymentStatus.Confirmed -> stringResource(R.string.booking_payment_status_confirmed)
+        BookingPaymentStatus.Paid -> stringResource(R.string.booking_payment_status_paid)
+        BookingPaymentStatus.Cancelled -> stringResource(R.string.booking_payment_status_cancelled)
+        BookingPaymentStatus.Complete -> stringResource(R.string.booking_payment_status_complete)
+        is BookingPaymentStatus.Unknown -> key
```
AdamGrzybkowski: Not sure if there's a better alternative, but just leaving a comment that this caught my attention. Displaying the raw values to the user could be risky, but it would be very handy at spotting the issue for us. One alternative I can think of would be to hide the badge completely.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
```diff
@@ -55,7 +56,17 @@ fun Booking.toUiModel(): BookingListItem {
             name = "Women’s Haircut",
             customerName = "Margarita Nikolaevna",
             attendanceStatus = AttendanceStatus.BOOKED,
-            paymentStatus = BookingPaymentStatus.fromKey(status)
+            paymentStatus = status.toUiModel()
         )
     )
 }
+
+private fun BookingEntity.Status.toUiModel(): BookingPaymentStatus = when (this) {
```
AdamGrzybkowski: You don't have to do this now, but we will need the same mappings in the details screen, so we could start putting those in a separate file, so they are available for reuse.

AdamGrzybkowski (approved):

AdamGrzybkowski:
>@AdamGrzybkowski while working on this, I noticed AttendanceStatusTag doesn't have the Booking prefix, should we add it for the sake of consistency?


Makes sense, I guess. Feel free to rename in this PR, or we can do it later.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
hichamboushaba: > It's still a bit confusing to me, because we can have two cancelled status tags next to each other (attendance and this one) 🤔

As discussed yesterday in the call, the `cancelled` status for the attendance is just a presentation status, it happens only when the booking itself is also cancelled.

> but it looks like it's not so yeah the name change makes sense.

Sounds good, I'll rename the UI model too.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/PaymentStatusTag.kt
hichamboushaba: > Displaying the raw values to the user could be risky

Not necessarly risky, for example for orders, WordPress has plugins that adds custom statuses, and for those we display the raw keys when there is no label, as this is what wp-admin does, and that's what the users are used too.

So I said, I think we need to revisit this when next-admin adds the logic for the badges.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt
hichamboushaba: Yes, this is something we need, a proper mapper class would be useful for reusability.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/269">23.4</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

PR #14673, opened 2025-10-01
[CIAB] Bookings search

hichamboushaba:
Closes WOOMOB-1402

### Description
This PR adds UI and logic for the search feature, and some notes about the implementation:
- Like what we do in orders and products, the selected filters are also applied during the search.
- To align with the design, and to follow what we do with products and orders, the tabs bar is hidden when search is active.

I have some concerns with the above approach as it can be confusing as to why the search is not returning results, I started a discussion about this in Slack p1759333433022899-slack-C09FHQNQERG, and I'll update the implementation when we make a final decision.

### Steps to reproduce
1. Use a CIAB site.
4. Create some bookings.
5. Test searching.

### Testing information
- From my testing, searching will check against the product name and the customer, so check this. And note that the data displayed is just mocked, so check the actual product name on your store.

The above.

### Images/gif

For the video, ignore the wrong tint color of the search icon, and the wrong paddings, I recorded this before finishing the final UI, the screenshots show the final UI.

https://github.com/user-attachments/assets/bf0b25f1-ce71-42fa-b139-9e107fcb6fa6

| Light | Dark |
| --- | --- |
| Screenshot_20251001_225752 | Screenshot_20251001_225736 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -67,34 +67,31 @@ class BookingListHandler @Inject constructor(
                 canLoadMore.set(false)
                 Result.success(Unit)
             } else {
-                searchBookings()
+                fetchBookings()
             }
         }
     }
 
     suspend fun loadMore(): Result<Unit> = mutex.withLock {
         if (!canLoadMore.get()) return@withLock Result.success(Unit)
-        return if (searchQuery.value == null) {
-            fetchBookings()
-        } else {
-            searchBookings()
-        }
+        return fetchBookings()
     }
 
     private suspend fun fetchBookings(): Result<Unit> {
+        val isSearching = !searchQuery.value.isNullOrEmpty()
         return bookingsRepository.fetchBookings(
             page = page.value,
             perPage = PAGE_SIZE,
+            query = searchQuery.value,
             filters = filters.value
-        ).onSuccess { hasMorePages ->
-            canLoadMore.set(hasMorePages)
-            if (hasMorePages) {
+        ).onSuccess { result ->
+            canLoadMore.set(result.hasMorePages)
+            if (result.hasMorePages) {
                 page.update { it + 1 }
             }
+            if (isSearching) {
+                searchResults.update { it + result.bookings }
+            }
         }.map { }
-    }
-
-    private suspend fun searchBookings(): Result<Unit> {
-        TODO("Not yet implemented")
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -43,7 +43,6 @@ class BookingListHandler @Inject constructor(
 
     suspend fun loadBookings(
         searchQuery: String? = null,
-        forceRefresh: Boolean = false,
         filters: List<BookingsFilterOption> = emptyList()
     ): Result<Unit> = mutex.withLock {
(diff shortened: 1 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
```diff
@@ -43,7 +43,6 @@ class BookingListHandler @Inject constructor(
 
     suspend fun loadBookings(
         searchQuery: String? = null,
-        forceRefresh: Boolean = false,
```
hichamboushaba, line 46: I added this argument thinking it might be needed (copied from the other `ListHandler` implementations), but now with search added, we still always passed `true` for the value, so I removed it to simplify the code.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -104,6 +125,61 @@ fun BookingListScreen(state: BookingListViewState) {
...
+        } else {
+            IconButton(onClick = {
+                searchState.onQueryChanged(null)
+            }) {
+                Icon(
+                    imageVector = Icons.AutoMirrored.Default.ArrowBack,
+                    contentDescription = stringResource(R.string.back),
+                    tint = MaterialTheme.colorScheme.onSurface
+                )
+            }
```
hichamboushaba, line 161: I wanted to use `navigationIcon` instead of this, but the M3 TopAppBar doesn't work well when a view in the `actions` spot uses `fillMaxWidth`, this causes an overlap:

Image

I don't know whether this is a bug or not, the M2 implementation doesn't have the same issue, but anyway, I think this approach works too.

AdamGrzybkowski: Hmm I would say this may be cleaner anyway. With the `navigationIcon`, you would have the logic spread between more places, whereas here you can handle it within one composable.

AdamGrzybkowski (approved):
Looks good! 

One small thing I noticed:

When rotating the device with search active (or doing any other action to cause a configuration change) the search input field cursor will reset its position to the start even when the search query is not empty.

Screenshot 2025-10-02 at 11 29 40

AdamGrzybkowski:
>From my testing, searching will check against the product name and the customer, so check this.

FYI: It also looks like the booking ID is matched.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Yes, that's a good point.

hichamboushaba:
> One small thing I noticed:
>
> When rotating the device with search active (or doing any other action to cause a configuration change) the search input field cursor will reset its position to the start even when the search query is not empty.

Thanks @AdamGrzybkowski for the review, that's a great catch, and impacts also other screens where `WCSearchView` is used and when focus state is restored on config changes (for example the `ProductSelectorScreen`), I attempted to fix the issue in the last commit 49bab4e, can you please review this change? I think it's safe, and from my testing it works well, but if we are worried about it, I can revert it, and try to fix it on a separate PR.

malinajirka:
👋 Hi there, the codefreeze for 23.4 is completed, so I'm moving this to 23.5.

AdamGrzybkowski (approved):
Tested the cursor and looks good to me - I focused on testing the position after config change, but also on being able to change the cursor position manually. :shipit:

PR #14675, opened 2025-10-01
[POS][Local Catalog] Run incremental sync on splash screen

samiuelson:
WOOMOB-1405

### Description

This PR performs incremental sync on the POS splash screen.

### Steps to reproduce

On the POS splash screen, the incremental sync should start. Sync should be non-blocking.

### Testing information

1. Go to POS
2. Notice in logs that incremental sync was attempted
3. Notice the sync does not block the splash screen from moving to POS

Above

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/splash/WooPosSplashViewModel.kt
```diff
@@ -23,13 +25,17 @@ class WooPosSplashViewModel @Inject constructor(
...
+    private val ordersCache: WooPosOrdersInMemoryCache,
+    performIncrementalSyncUseCase: WooPosPerformLocalCatalogIncrementalSync
 ) : ViewModel() {
     private val _state = MutableStateFlow<WooPosSplashState>(WooPosSplashState.Loading)
     val state: StateFlow<WooPosSplashState> = _state
 
     init {
         val splashScreenStartTime = System.currentTimeMillis()
+
+        performIncrementalSyncUseCase.execute(WooPosIncrementalSyncReason.ON_SPLASH_SCREEN)
```
kidinov, line 37: Shouldn’t it be blocking? I mean, shouldn’t we display a splash screen until the process is complete? Or what’s the plan here? Currently, every time I open the POS for the first time, it shows an empty list of products.

kidinov (approved):
LGTM! I left one non-blocking question

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/splash/WooPosSplashViewModel.kt
samiuelson: Good question. The plan is to make the incremental sync non-blocking in general. However, in the next PR, I'm going to add a check if the full sync is required, which will handle cases like POS opened for the first time (empty local catalog). In case the full sync was never performed before, or the catalog is empty, the full sync will be required and will block an access to POS.

PR #14678, opened 2025-10-02
[Woo POS][Local Catalog] Catalog full sync required 

samiuelson:
WOOMOB-1412

### Description

The goal of PR is to improve the reliability of the full sync of POS local catalog. 
The full sync is scheduled to be performed every 24h, at night. However, in case the device was turned off at night, the sync may not succeed. To handle such case, this PR adds:
1. Keeping track of the last full sync timestamp
2. Checking if the last full sync is overdue (7 days)
3. Scheduling one-time, instant full sync in case the full sync was never done before, or data is missing (WooPosProductEntity is empty)

⚠️ Do not merge until the base branch is `trunk`.

### Steps to reproduce

The full sync should be performed in a blocking way before allowing access to the POS for the first time (or in case product data is missing)

### Testing information

* Wipe the PosProductsEntity table and then verify that the full sync occurs when entering POS.

Above.

### Images/gif

https://github.com/user-attachments/assets/83c61e84-c1d0-40f8-8622-840840919ed0


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInitialCatalogFullSync.kt
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+    }
+
+    private suspend fun FlowCollector<WooPosFullSyncStatus>.monitorWorkerProgress() {
+        wooPosLogWrapper.d("One-time full sync worker already running, monitoring progress")
+        emit(WooPosFullSyncStatus.InProgress)
+
+        var workerStillRunning = true
+        while (workerStillRunning) {
```
samiuelson: 💡 I'll investigate if it's possible to replace while with [getWorkInfoById](https://developer.android.com/develop/background-work/background-tasks/persistent/how-to/observe#observe_progress).

samiuelson: Done

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampRepository.kt
```diff
@@ -72,15 +72,35 @@ class WooPosSyncTimestampRepository @Inject constructor(
...
+            dataStore.edit { preferences ->
+                preferences[key] = timestamp.toString()
+            }
+        }
+    }
+
+    suspend fun getFullSyncLastCompletedTimestamp(): Long? {
+        val key = buildSiteSpecificKey(FULL_SYNC_TIMESTAMP_KEY)
+        return if (key != null) {
+            dataStore.data.first()[key]?.toLong()
```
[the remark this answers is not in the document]
samiuelson: Done: b5b0dd2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -31,6 +37,9 @@ class WooPosHomeViewModel @Inject constructor(
     private val parentToChildrenEventSender: WooPosParentToChildrenEventSender,
     private val analyticsTracker: WooPosAnalyticsTracker,
     private val soundHelper: WooPosSoundHelper,
+    private val syncStatusChecker: WooPosFullSyncStatusChecker,
+    private val performInitialFullSync: WooPosPerformInstantCatalogFullSync,
+    private val incrementalSync: WooPosPerformLocalCatalogIncrementalSync,
```
malinajirka, line 36: 🔍 We might want to add `perform` prefix for consistency.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        val site = selectedSite.getOrNull()
+        if (site == null) {
+            wooPosLogWrapper.e("Full sync check failed: No site selected")
+            return WooPosFullSyncRequirement.Error("No site selected")
+        }
+
+        if (!networkStatus.isConnected()) {
+            wooPosLogWrapper.d("Full sync check skipped: No network connection")
```
malinajirka: ❓  Won't this prevent the POS from loading when there is no internet connection even when we have data in the database?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+                0
+            }
+        val catalogIsEmpty = productCount == 0
+
+        return when {
+            lastFullSyncTimestamp == null -> {
+                wooPosLogWrapper.d("Full sync required: Never synced before")
+                WooPosFullSyncRequirement.BlockingRequired
+            }
+            catalogIsEmpty -> {
```
malinajirka: ❓ I'm honestly not sure, but I think if the sync was successfully performed and the catalog is empty, we might still want to let users into the POS. Their catalog is up to date, they just don't have any products. Wdyt?

Moreover, it might bite use, if we keep this logic and introduce custom amounts - we might forget to test the custom amounts on an empty store.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+            }
+            else -> {
+                wooPosLogWrapper.d("Full sync not required: Recent sync found at $lastFullSyncTimestamp")
+                WooPosFullSyncRequirement.NotRequired
+            }
+        }
+    }
+
+    private fun isFullSyncOverdue(lastSyncTimestamp: Long): Boolean {
+        val currentTime = System.currentTimeMillis()
```
malinajirka, line 73: 🔍 This code doesn't seem testable, we might want to inject time provider.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private suspend fun FlowCollector<WooPosFullSyncState>.handleWorkerCompletion(
+        workInfo: WorkInfo?,
+        workerType: String
+    ) {
+        when (workInfo?.state) {
+            WorkInfo.State.SUCCEEDED -> {
+                val completedTimestamp = syncTimestampManager.getFullSyncLastCompletedTimestamp()
+                if (completedTimestamp != null) {
```
malinajirka: ❓ How can this scenario happen?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
```diff
@@ -0,0 +1,125 @@
...
+            else -> {
+                performBlockingSync()
+            }
+        }
+    }
+
+    private suspend fun FlowCollector<WooPosFullSyncState>.monitorOneTimeWorkerProgress() {
+        emit(WooPosFullSyncState.InProgress)
+
+        val finalWorkInfo = syncScheduler.observeOneTimeWorkInfo()
```
malinajirka: ❓ It seems the worker as well as this code runs in a coroutine (not main thread) => are we guaranteed that between 

`val isOneTimeRunning = syncScheduler.observeOneTimeWorkStatus().first()`  

and 

`val finalWorkInfo = syncScheduler.observeOneTimeWorkInfo().filter { workInfo ->
                workInfo?.state?.isFinished == true
            }
            .first()` 

the worker doesn't finish? I'm not sure, but I have a gut feeling this code might lead to a deadlock.

------------------

I think we could fix it by completing this observer also when `workInfo` is null (aka if no work is in progress anymore). Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -54,6 +63,49 @@ class WooPosHomeViewModel @Inject constructor(
         viewModelScope.launch {
             soundHelper.preloadChaChing()
         }
+        startCatalogSyncIfNeeded()
+    }
+
+    private fun startCatalogSyncIfNeeded() {
```
malinajirka: 💡 I'd consider extracting this into a separate class - feels like an implementation detail the VM shouldn't know about, otherwise, it might grow a lot overtime.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
+import com.woocommerce.android.ui.woopos.featureflags.WooPosLocalCatalogM1Enabled
+import com.woocommerce.android.ui.woopos.util.WooPosNetworkStatus
+import com.woocommerce.android.ui.woopos.util.datastore.WooPosSyncTimestampManager
+import org.wordpress.android.fluxc.model.LocalOrRemoteId
+import org.wordpress.android.fluxc.store.pos.localcatalog.WooPosLocalCatalogStore
+import javax.inject.Inject
+import kotlin.time.Duration.Companion.days
+
+class WooPosFullSyncStatusChecker @Inject constructor(
```
malinajirka, line 13: 📓 This class seems to be missing unit tests.

malinajirka (changes requested):
Thanks for working on this @samiuelson!

I'm not entirely sure what is in scope of this PR and what is out of scope. I'll list issues I encountered, feel free to ignore those that are out of scope.

1. When we are doing an initial full sync, the blocking screen should be full screen (part of splash screen):
POS (51)

instead of the current:
Screenshot_20251008_082416


2. I noticed that currently during the full sync, the blocking layer is semi-transparent, so I can see products already loaded below the loading state:

Screenshot_20251008_081437

3. I can see loading skeletons during an incremental sync - the incremental sync shouldn't be visible to the user as it goes against the speed we are trying to achieve.

https://github.com/user-attachments/assets/fa410b08-454b-4c8c-9d6d-eb43147a76c2

4. Not sure if it's an issue, but I can see the loading skeletons for 0.5seconds even when I turn off the feature flag even though we load the items during the splashscreen.

5. The POS is not accessible when there is no internet connection even though I have data in the database + I can see the data behind the error:

Screenshot_20251008_081814

Let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
samiuelson: Moved to splash screen VM and renamed: afed690

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
samiuelson: Moved to splash screen VM and extracted into `WooPosPerformLocalCatalogInitialFullSync` class: afed69091c9d3e06bf74838b40f39ce562cdc518, 88fdeab6bfa364964e9b5cdac26936f36e69c9da

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
samiuelson: Fixed in 1bed6b0bc224ae5cce61059c651b1dc2dd6d6888

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
samiuelson: Fixed in 1bed6b0bc224ae5cce61059c651b1dc2dd6d6888

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
samiuelson: @malinajirka 👋  Most common case I can think of - `completedTimestamp` can be null here, for example, in case the site was switched during full sync. The `completedTimestamp` would be null for the newly selected site, even though the worker would finish full sync with success for the previous site. `WooPosSyncTimestampRepository` handles values specific to the currently selected site.

samiuelson: This makes me think - maybe we should schedule sync workers with names unique to the currently selected site? This way we could make sure at POS splash screen, that the active worker is syncing data of the currently selected site. Wdyt, @malinajirka ?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
samiuelson: Good catch. I tried improving it in 6190d970c66112e712dd8bd58ef689c84283856e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
samiuelson: Done: ec3a521d6f9513819a8e61ce33d5d4d406b3bbbb

samiuelson:
> 1. When we are doing an initial full sync, the blocking screen should be full screen (part of splash screen):
> 2. I noticed that currently during the full sync, the blocking layer is semi-transparent, so I can see products already loaded below the loading state:

That was fixed 👍 

> 3. I can see loading skeletons during an incremental sync - the incremental sync shouldn't be visible to the user as it goes against the speed we are trying to achieve.
> 4. Not sure if it's an issue, but I can see the loading skeletons for 0.5seconds even when I turn off the feature flag even though we load the items during the splashscreen.

Let me improve 3. and 4. in a separate PR, as this one grew too big—sorry for that, BTW! I created task for this: WOOMOB-1525

> 5. The POS is not accessible when there is no internet connection even though I have data in the database + I can see the data behind the error:

Good catch. This was fixed 👍 

Thank you for the review, @malinajirka! I addressed your comments. Also, sorry for the long PR I failed to keep it within a limit.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
malinajirka: @samiuelson I'm up for it, but it might not be trivial I think - the worker should likely be scheduled in app startup and the user might not be logged in yet. I'd be a bit worried we might end up in a situation where we have a bunch of workers waking up the device too often.

samiuelson: Makes sense. In that case, relying on `getFullSyncLastCompletedTimestamp` to verify if the completed sync was related to the currently selected store should be good enough, IMO.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/splash/WooPosSplashViewModel.kt
```diff
@@ -49,15 +48,44 @@ class WooPosSplashViewModel @Inject constructor(
...
+            _state.value = WooPosSplashState.Syncing
+            performInitialFullSync().collect(syncStateCollector(retryStartTime))
+        }
+    }
+
+    private fun syncStateCollector(
+        startTime: Long
+    ) = FlowCollector<WooPosLocalCatalogInitialFullSyncState> { syncState ->
+        when (syncState) {
+            is WooPosLocalCatalogInitialFullSyncState.Ready -> {
```
malinajirka: 🔍 np: I'd consider calling this NotRequired instead of Ready. At first I thought `Ready` means that the sync is about to start.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/pos/WooPosProductsDao.kt
```diff
@@ -16,6 +16,9 @@ abstract class WooPosProductsDao {
     @Query("SELECT * FROM PosProductEntity WHERE localSiteId = :localSiteId AND remoteId = :remoteId")
     abstract suspend fun getProduct(localSiteId: LocalId, remoteId: RemoteId): WooPosProductEntity?
 
+    @Query("SELECT COUNT(*) FROM PosProductEntity WHERE localSiteId = :localSiteId")
+    abstract suspend fun getProductCount(localSiteId: LocalId): Int
```
malinajirka, line 34: 📓 I think this is fine for the current use case, but we need to be careful, since `getProductCount` returns count of all products including those which are not supported by the POS (hidden in the POS).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -0,0 +1,89 @@
...
+            wooPosLogWrapper.d("Full sync check skipped: Local catalog feature not enabled")
+            return WooPosFullSyncRequirement.NotRequired
+        }
+
+        val site = selectedSite.getOrNull()
+        if (site == null) {
+            wooPosLogWrapper.e("Full sync check failed: No site selected")
+            return WooPosFullSyncRequirement.Error("No site selected")
+        }
+
```
malinajirka, line 33: 💡 I'm wondering if we need to check `isPeriodicSyncEnabledForSite` here as well. Otherwise, this class might require sync while the worker always finishes with `failure` on line WooPosLocalCatalogSyncWorker::54. Wdyt? 

P.S. Feel free to create a separate ticket.

malinajirka (approved):
Thanks @samiuelson for iterating on the PR! I've reviewed the code and it all looks good to me. However, I've encountered a couple issues when testing the PR. I'd consider merging it as is and fixing the bugs in a separate PR, wdyt?


#### Infinite splash screen
1. Downgrade your site to WooCommerce 10.2 or older
2. Clean apps data
3. Login
4. Enter POS
5. Notice catalog is syncing -> notice `/variations` requests fail since the endpoint isn't available in v10.2 => the logs mention `Worker result RETRY` which isn't properly handled (it goes again into `ENQUEUED` state on Retry).

Screenshot 2025-10-18 at 19 13 12


#### POS Opened with empty catalog
1. Clean apps data
2. Login
3. Turn off wifi/data
4. Open POS
5. Notice expected error
6. Tap on retry
7. Turn on wifi/data
8. Tap on retry
9. Notice you are allowed to enter POS before the **initial** sync completes.

https://github.com/user-attachments/assets/63fbec00-a5f1-49f9-ac65-eb74a63593e2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/splash/WooPosSplashViewModel.kt
samiuelson: Done: 56bee469a4b2cf1893d0ea1b894af2e806d2c8cc

samiuelson:
> Thanks @samiuelson for iterating on the PR! I've reviewed the code and it all looks good to me. However, I've encountered a couple issues when testing the PR. I'd consider merging it as is and fixing the bugs in a separate PR, wdyt?
> 
> #### Infinite splash screen
> 1. Downgrade your site to WooCommerce 10.2 or older
> 2. Clean apps data
> 3. Login
> 4. Enter POS
> 5. Notice catalog is syncing -> notice `/variations` requests fail since the endpoint isn't available in v10.2 => the logs mention `Worker result RETRY` which isn't properly handled (it goes again into `ENQUEUED` state on Retry).
> 

Task created: WOOMOB-1536

> #### POS Opened with empty catalog
> 1. Clean apps data
> 2. Login
> 3. Turn off wifi/data
> 4. Open POS
> 5. Notice expected error
> 6. Tap on retry
> 7. Turn on wifi/data
> 8. Tap on retry
> 9. Notice you are allowed to enter POS before the **initial** sync completes.
> 

Task created: WOOMOB-1537

I appreciate the thorough review, Jirka! I added follow-up tasks in linear: WOOMOB-1538, WOOMOB-1537, WOOMOB-1536.

PR #14682, opened 2025-10-02
[WOOMOB-1400] - Observe booking object from DB on the details screen

AdamGrzybkowski:
WOOMOB-1400

### Description

The main goal of this PR is to implement `BookingRepository` in the `BookingDetailsViewModel` to observe the booking object in the DB.

This only observes the local object. We will have to fetch more information anyway, so that's where I think we can add refreshing of the booking model as well (if we think it's necessary).

Additionally, since we have to handle object mapping, I went ahead and created a mapper class `BookingMapper`, for all booking-related models. In one of the commits, I have added this mapper to the `BookingListViewModel` as well. 

### Steps to reproduce

1. Use a CIAB site.
3. You'll need to enable bookings v2 on your CIAB testing site. You can do that by installing the Code [Snippets plugin](https://wordpress.org/plugins/code-snippets/) to your site and adding the following PHP snippet: define( 'WC_BOOKINGS_NEXT_ENABLED', true );
4. Create some bookings.
5. Open the app and go to the booking tab
6. Notice the data/time info on one of the booking item
7. Tap it
8. Confirm the same information is visible in the details screen
 
### Testing information

No UI changes. The above should be enough.

The above

### Images/gif

https://github.com/user-attachments/assets/0e5196a4-ac0c-4b6f-942b-45eb5ed6f08c


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
```diff
@@ -4,43 +4,19 @@ import com.woocommerce.android.ui.bookings.compose.BookingAppointmentDetailsMode
...
     val onCancelBooking: () -> Unit = {},
     val onAttendanceStatusSelected: (BookingAttendanceStatus) -> Unit = { _ -> }
 )
+
+data class BookingUiState(
+    val bookingSummary: BookingSummaryModel,
+    val bookingsAppointmentDetails: BookingAppointmentDetailsModel,
+    val bookingCustomerDetails: BookingCustomerDetailsModel,
+    val bookingPaymentDetails: BookingPaymentDetailsModel,
+)
```
AdamGrzybkowski, line 22: All those models would be nullable after this PR, so it's handy to wrap them in a single class to avoid multiple null checks.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -0,0 +1,69 @@
...
+import java.time.format.FormatStyle
+import javax.inject.Inject
+
+class BookingMapper @Inject constructor(
+    private val currencyFormatter: CurrencyFormatter,
+) {
+    private val summaryDateFormatter: DateTimeFormatter = DateTimeFormatter.ofLocalizedDateTime(
+        FormatStyle.MEDIUM,
+        FormatStyle.SHORT
+    ).withZone(ZoneOffset.UTC)
```
AdamGrzybkowski, line 22: cc @hichamboushaba  I've used the UTC based on your discussion.

AdamGrzybkowski:
@hichamboushaba I pushed one extra commit with a one-liner that I forgot to do after rebasing with trunk. https://github.com/woocommerce/woocommerce-android/pull/14682/commits/509794d7b5731f41a1b4ea3847de4599bcf8a3f5

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -35,17 +44,59 @@ class BookingDetailsViewModel @Inject constructor(
                 toolbarTitle = resourceProvider.getString(R.string.booking_details_title, navArgs.bookingId),
             )
         }
+        observeBooking(navArgs.bookingId)
```
hichamboushaba: Before approving this, I just want to discuss one problem with this approach and decide whether we are OK with it or not.
With this approach, we observe the DB regardless of whether the app is in the foreground or not (this is part of what we discussed before on whether we should expose flows or LiveData).

I think a better approach (IMO 🙂) would be to have a cold flow exposed as `LiveData`, so the cold flow won't start until the screen reaches the `Started` lifecycle state, and to achieve it, we can think of something like the following patch:

```Patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt	(revision 509794d7b5731f41a1b4ea3847de4599bcf8a3f5)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt	(date 1759502430662)
@@ -4,7 +4,6 @@
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.asLiveData
-import com.woocommerce.android.ui.bookings.Booking
 import com.woocommerce.android.ui.bookings.BookingMapper
 import com.woocommerce.android.ui.bookings.BookingsRepository
 import com.woocommerce.android.ui.bookings.compose.BookingAttendanceStatus
@@ -15,69 +14,41 @@
 import com.woocommerce.android.viewmodel.navArgs
-import kotlinx.coroutines.flow.launchIn
-import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.filterNotNull
 import javax.inject.Inject
 
 @HiltViewModel
 class BookingDetailsViewModel @Inject constructor(
     savedState: SavedStateHandle,
     resourceProvider: ResourceProvider,
-    private val bookingsRepository: BookingsRepository,
+    bookingsRepository: BookingsRepository,
     private val bookingMapper: BookingMapper,
 
-    private val _state = MutableStateFlow(
-        BookingDetailsViewState(
-            onCancelBooking = ::onCancelBooking,
-            onAttendanceStatusSelected = ::onAttendanceStatusSelected,
-        )
-    )
-    val state: LiveData<BookingDetailsViewState> = _state.asLiveData()
-
-    init {
-        _state.update {
-            it.copy(
-                toolbarTitle = resourceProvider.getString(R.string.booking_details_title, navArgs.bookingId),
-            )
-        }
-        observeBooking(navArgs.bookingId)
-    }
-
-    private fun onAttendanceStatusSelected(status: BookingAttendanceStatus) {
-        val bookingState = _state.value.bookingUiState
-        if (bookingState != null) {
-            _state.update { current ->
-                current.copy(
-                    bookingUiState = bookingState.copy(
-                        bookingSummary = bookingState.bookingSummary.copy(attendanceStatus = status)
-                    )
-                )
-            }
-        }
-    }
-
-    private fun onCancelBooking() {
-        // TODO Add logic to Cancel booking
-    }
-
-    private fun observeBooking(bookingId: Long) {
-        bookingsRepository.observeBooking(bookingId)
-            .onEach { booking ->
-                booking?.let { updateStateWithBooking(it) }
-            }
-            .launchIn(this)
-    }
-
-    private fun updateStateWithBooking(booking: Booking) = with(bookingMapper) {
-        _state.update { current ->
-            current.copy(
+    private val booking = bookingsRepository.observeBooking(navArgs.bookingId)
+
+    // If we need this, then using `savedState.getStateFlow` would be better here to preserve state across process death,
+    private val bookingAttendanceStatus = MutableStateFlow<BookingAttendanceStatus?>(null)
+
+    val state: LiveData<BookingDetailsViewState> = combine(
+        booking.filterNotNull(),
+        bookingAttendanceStatus
+    ) { booking, attendanceStatus ->
+        with(bookingMapper) {
+            BookingDetailsViewState(
+                toolbarTitle = resourceProvider.getString(R.string.booking_details_title, booking.id),
                 orderId = booking.orderId,
                 bookingUiState = BookingUiState(
-                    bookingSummary = booking.toBookingSummaryModel(),
+                    bookingSummary = booking.toBookingSummaryModel().let {
+                        if (attendanceStatus != null) {
+                            it.copy(attendanceStatus = attendanceStatus)
+                        } else {
+                            it
+                        }
+                    },
                     bookingsAppointmentDetails = booking.toAppointmentDetailsModel(),
                     bookingCustomerDetails = BookingCustomerDetailsModel(
                         name = "Margarita Nikolaevna",
@@ -96,7 +67,24 @@
                         total = "$59.50"
                     )
                 ),
+                onCancelBooking = ::onCancelBooking,
+                onAttendanceStatusSelected = ::onAttendanceStatusSelected
             )
         }
+    }.asLiveData()
+
+    private fun onAttendanceStatusSelected(status: BookingAttendanceStatus) {
+        // Whether we'll need the `bookingAttendanceStatus` property or not depending
+        // on whether we'll have a "save" button for the whole screen or we'll update the status
+        // as soon as the user selects it from the bottom sheet.
+        // Personally I think we should update it immediately, and if we do, then we don't need the property
+        // we just need to manage the loading state, and when the API call is done, the booking will be updated accordingly
+        // via the `booking` flow.
+        // But this is a product decision. And this patch just preserves the current behavior.
+        bookingAttendanceStatus.value = status
+    }
+
+    private fun onCancelBooking() {
+        // TODO Add logic to Cancel booking
     }
 }
```

The patch also makes the whole screen reactive, so if during the implementation we find that we need the Order (which most probably will happen), then we can chain the Flows to observe also the Order from the DB, and then the DB will be the single source of truth, and drives the screen.

Please check it out, and share your thoughts. While I think observing the DB when the app is in the background should be avoided, if you don't agree with the above approach, I'm OK with accepting the PR as it's.

AdamGrzybkowski: >Please check it out, and share your thoughts. While I think observing the DB when the app is in the background should be avoided, if you don't agree with the above approach, I'm OK with accepting the PR as it's.

Thanks @hichamboushaba! You're obviously correct. I think I was working on an autopilot, not being used to LiveData 🤦 

I will update the PR on Monday!

AdamGrzybkowski: Done in https://github.com/woocommerce/woocommerce-android/pull/14682/commits/b1687573feeffc7fa691ac4b22fe4159ee915a67

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    private val detailsDateFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("EEEE, dd MMM yyyy")
```
hichamboushaba: IMO we should avoid hardcording formatting pattens, I'd use a localized format instead. I know it's used in the app is some spots, but IMO it's not something we should follow.


```suggestion
    private val detailsDateFormatter: DateTimeFormatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.FULL)
```

The above will update the text from: `Monday, 06 Oct 2025` to `Monday, 6 October 2025` for UK english set as the locale.

Another approach if we are sure we want to keep the shorter month name is to use something like:

```kotlin
    private val detailsDateFormatter: DateTimeFormatter = DateTimeFormatterBuilder()
        .appendPattern("EEEE, ")
        .append(DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM))
        .toFormatter()
        .withZone(ZoneOffset.UTC)
```

But personally I think using `FormatStyle.FULL` is better in this context, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -0,0 +1,69 @@
...
+            date = summaryDateFormatter.format(start),
+            // TODO replace mocked values when product and customer data are available
+            name = "Women’s Haircut",
+            customerName = "Margarita Nikolaevna",
+            attendanceStatus = BookingAttendanceStatus.BOOKED,
+            status = status.toUiModel()
+        )
+    }
+
+    fun Booking.toUiModel(): BookingListItem {
```
hichamboushaba: np, just to make the role a bit clearer, I'd rename this to `toListItem()`, WDYT?

hichamboushaba (approved):
Nice work @AdamGrzybkowski 💯, works well, I left two remarks, but they are not blockers.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
AdamGrzybkowski: Makes sense, I did focus on the format I saw in the Figma, but this will work better across different locales. Thanks!

PR #14684, opened 2025-10-02
[POS Orders] Pagination: Search and Error

toupper:
### Description

This PR finalizes pagination for the POS Orders screen by adding support for search pagination and improving error handling. In case of errors, an error message is shown and users can retry.

Other than that, I refactored `WooPosOrdersViewModel` and `WooPosOrdersDataSource` for simplicity and readability.

#### Search

	1.	Go to the POS tab.
	2.	Open the Orders screen.
	3.	Perform a search with a query that returns more than one page of results.
	4.	Verify that pagination works correctly across multiple pages.

#### Error

	1.	Go to the POS tab.
	2.	Open the Orders screen.
	3.	Wait for the first page of orders to load, then go offline.
	4.	Scroll down until you reach the end of the list — an error row should be displayed.
	5.	Go back online.
	6.	Tap the error row and confirm that the next page of orders loads properly.

### Images/gif

https://github.com/user-attachments/assets/2d5e928b-0dc2-4cfe-83f4-16058170d083


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
kidinov: Should the default value be 1, not 0?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -40,49 +40,49 @@ class WooPosOrdersDataSource @Inject constructor(
...
+            onFailure = { SearchOrdersResult.Error(it.message ?: UNKNOWN_ERROR) }
+        )
     }
 
-    suspend fun loadMore(): Result<List<Order>> = withContext(Dispatchers.IO) {
-        val result = fetchOrdersFromRemote(searchQuery = null, page = page.get())
+    suspend fun loadMore(searchQuery: String? = null): Result<List<Order>> =
+        withContext(Dispatchers.IO) { fetchAndMap(searchQuery) }
+
+    private suspend fun fetchAndMap(
```
kidinov: While default params often are useful tool, I think in this case there are a couple of issues with that code wise:

  * Boolean trap: reset: Boolean = false makes call sites unclear. When you see fetchAndMap(searchQuery, true), it's not immediately obvious what true means without checking the signature.
  * Coupling with nullable parameter: The combination of searchQuery: String? = null and reset: Boolean = false creates implicit dependencies between parameters. When resetting, you often want to start fresh, but the default doesn't enforce this relationship.

So maybe having methods with clear names could make the code better
[Refactor_order_loading_methods_for_pagination_and_search_functionality.patch](https://github.com/user-attachments/files/22676399/Refactor_order_loading_methods_for_pagination_and_search_functionality.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -385,6 +403,57 @@ private fun OrdersPaginationLoadingRow() {
...
+private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
+    WooPosPaginationErrorIndicatorContent(
+        message = stringResource(id = R.string.woopos_orders_pagination_error_title),
+        description = stringResource(id = R.string.woopos_orders_pagination_error_content_description),
+        onRetry = onPaginationErrorTryAgain
+    )
+}
+
+@Composable
+private fun WooPosPaginationErrorIndicatorContent(
```
kidinov: I think we need to reuse `WooPosPaginationErrorIndicator`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: Unused

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -250,21 +260,21 @@ class WooPosOrdersViewModel @Inject constructor(
...
         val current = _state.value as? WooPosOrdersState.Content
         val existingItems = current?.items.orEmpty()
-        val currentSelectedId = current?.selectedOrderId
-            ?: existingItems.firstOrNull()?.id
-            ?: orders.firstOrNull()?.id
+        val selectedId = current?.selectedOrderId ?: existingItems.firstOrNull()?.id ?: orders.firstOrNull()?.id
+        val newItems = mapOrders(orders, selectedId)
 
-        val newItems = mapOrders(orders, currentSelectedId)
+        // Let's remove duplicates as it would crash when showing them
```
kidinov: Won't it be just hiding up the potential issue with implementation? Or are there valid cases when the duplicates can be there?

In any case, there is a more efficient way of removing duplicates:

```
      val existingIds = existingItems.mapTo(HashSet()) { it.id }
      val mergedItems = existingItems + newItems.filterNot { it.id in existingIds }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: I tried using WooPosPaginationErrorIndicator, but it didn’t fit our case: it renders an icon and a button, which isn’t ideal for the narrow width we have. In addition, the component doesn’t adapt to the available width, causing text to be cut off. For these reasons, I created a dedicated component tailored to our scenario.

kidinov (approved):
LGTM! I left a few suggestions. Please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: In general, we shouldn’t encounter duplicated orders, but I think it’s still worth being defensive to avoid potential crashes. I’m not strongly opinionated here and happy to remove it if that’s the consensus. I also improved the duplicate-removal logic in 420433bca95b15b630856ccc574244fd6a08b33b.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: I'd say then we need to adapt the generic component to be able to not render the icon instead of creating a new one. This component should have a button or something actionable anyway—it cannot be just  text

toupper: > This component should have a button or something actionable anyway—it cannot be just text

With this implementation, the user can tap the row to retry—mirroring the behavior in Coupons. It’s a straightforward way to declutter the UI and still provide a quick, obvious action. What do you think?

kidinov: Tbh, I doubt that a text can be obvious as a clickable element unless it's highlighted as a link.

Coupons use the same component for load next page retry with the button.

image

iOS implementation reuses the same component (at least it looks like the same) without an icon, too

Orders

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Thanks for the comment and suggested code! I applied the patch in 90bd7528837a65e6d323670ef6da5a6b9dbf14fb

Thanks for pointing this out 🙏 As I’m still fairly new to Kotlin, I’d like to understand this better. I see the concern about the “boolean trap,” but I’m not sure I fully get why this is considered a bigger issue here compared with other cases in our code.

For example, we already have functions like:

```
private suspend fun fetchOrdersFromRemote(
    page: Int,
    searchQuery: String?
)
```

At the call site, if we just write `fetchOrdersFromRemote(2, "test")`, it’s not obvious that 2 represents a page number unless we use a named argument or pass a meaningful named variable. Wouldn’t we have the same issue in many other cases in our code?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Thanks! Done in 94b38e8d358d12e7a8bb825961c9eea97fe71fc8

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Good catch, [page number is 1-based ](https://woocommerce.github.io/woocommerce-rest-api-docs) Done in 81e1c91727de09b7adf75e11517cd1b6e1acf030

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
kidinov: Great question! You're right that `fetchOrdersFromRemote(2, "test")` has similar readability issues. The key difference is **how often this happens in practice**.

**With parameters like `page` and `searchQuery`:**
- These are usually passed as variables with meaningful names: `fetchOrdersFromRemote(currentPage, query)`
- This naturally documents itself at the call site

**With boolean parameters:**
- They're almost always literals: `fetchAndMap(null, true)` instead of `val shouldReset = true; fetchAndMap(null, shouldReset)`
- This makes them unclear much more frequently

**The fix is also simpler for booleans:**
```kotlin
fetchAndMap(searchQuery)           // normal fetch
fetchAndMapWithReset(searchQuery)  // clear intent
```

So you're not wrong about the broader issue — boolean traps are just a more common subset of it, with an easier fix

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: As I see it, there is a trade off:
* We do not crash but potentially do not show some orders, as due to the issue in the implementation on our side or backend, we fetch the next page that does not contain one or more orders that should be shown. We probably won't notice this, and it's almost impossible to debug.
* We crash when this happens -> notice that as a report in sentry -> we can fix that.

I think #2 is better. 

But again, unless you see the valid case when duplicates are possible. If the only way that duplicates are possible is due to implementation issues, I believe we need to go with #2.

toupper: Awesome, let's go with #2. Done in 88b2c709b630bd32fcc1ab4e5aa240ebd942da70

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Awesome, thanks for adding more context 👍

PR #14686, opened 2025-10-02
[CIAB Bookings] Update logic of tabs filtering

hichamboushaba:
Closes WOOMOB-1404

### Description
This PR updates the logic of filtering to match what the API expects, because the API stores dates without timezone information (dates stored as strings), it ignores any timezone information when filtering, so to get the correct results for date filtering, we need to send the dates as UTC.

### Steps to reproduce
1. Prepare a CIAB site with bookings (check previous PRs for steps if needed).
2. Create some bookings.
3. Confirm Today and Upcoming behave as expected.

### Testing information
- Check the bookings times listed on the server, and confirm that Today lists all bookings for today, and Upcoming lists all future bookings (from tomorrow midnight).

- The above.
- Also I created a booking at `11:30pm` on the server, and confirmed that the logic now works well, as this booking wasn't shown before for today (my local timezone is UTC+1).

### Images/gif
| Before (missing one booking) | After |
| --- | --- |
| Screenshot_20251002_181303 | Screenshot_20251002_182207 |


AdamGrzybkowski (approved):
The filtering now works correctly!

PR #14693, opened 2025-10-03
Fix Room crash in 23.3

hichamboushaba:
### Description
This PR fixes the crash discussed here p1759469754623059-slack-C6H8C3G23, the PR includes the following changes:

- Removes the `@Transaction` annotation from the `FeatureFlagConfigDao` functions, the Dao functions involve a single table, so the transaction usage wasn't needed anyway, and it's most probably what's causing the crash here (more details in the above thread).
- Just for more safety, we cache the feature flag in memory, and we ensure we call the DB just once.

### Steps to reproduce
1. Add a `print` statement in this [line](https://github.com/woocommerce/woocommerce-android/blob/c787f89e663a2b2545e6350819947679606f15fe/WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/WooApplicationPasswordsConfiguration.kt#L25)
2. Sign in using WordPress.com
3. Open the app, and check that the print statement is invoked only once.
4. Confirm the app uses app passwords (by inspecting network).

The above.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/WooApplicationPasswordsConfiguration.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/WooApplicationPasswordsConfiguration.kt
@@ -2,21 +2,34 @@ package com.woocommerce.android.applicationpasswords
 
 import com.woocommerce.android.AppPrefsWrapper
 import com.woocommerce.android.BuildConfig
+import com.woocommerce.android.di.AppCoroutineScope
 import com.woocommerce.android.util.DeviceInfo
 import com.woocommerce.android.util.IsRemoteFeatureFlagEnabled
 import com.woocommerce.android.util.RemoteFeatureFlag
 import jakarta.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/WooApplicationPasswordsConfiguration.kt
```diff
@@ -2,21 +2,34 @@ package com.woocommerce.android.applicationpasswords
...
-    private val isRemoteFeatureFlagEnabled: IsRemoteFeatureFlagEnabled
+    private val isRemoteFeatureFlagEnabled: IsRemoteFeatureFlagEnabled,
+    @AppCoroutineScope private val appCoroutineScope: CoroutineScope
 ) : ApplicationPasswordsConfiguration {
+    private val remoteFeatureFlagDeferred = appCoroutineScope.async(
+        start = CoroutineStart.LAZY,
+        context = Dispatchers.IO
+    ) {
+        isRemoteFeatureFlagEnabled(RemoteFeatureFlag.APP_PASSWORDS_FOR_JETPACK_SITES)
+    }
```
hichamboushaba, line 26: This uses the same approach we use in other places in the app, and it's inspired by the suggestions in the issue 706 in the coroutines repo.

irfano (approved):
LGTM! 👍🏻

PR #14695, opened 2025-10-03
[POS Orders] Empty and Error states

toupper:
Part of WOOMOB-1149

### Description

With this PR we handle the error and empty cases when loading POS Orders. The CTA for the empty state is still to be defined.

#### Empty State

With a store without POS orders

1. Go to POS
2. Go to Orders
3. See the empty state.
4. Tap on the back arrow. You should navigate to the POS homescreen.

#### Error State

1. Go to POS
2. Turn offline
3. Go to Orders
4. See the error state.
5. Go Online
6. Tap on Retry
7. Orders should be loaded properly
8. Check also that tapping on the back arrow sends you to the POS Homescreen when the error state is shown. 

### Images/gif

https://github.com/user-attachments/assets/386ae265-7770-443e-8c99-01cb9ae5a400


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
```diff
@@ -78,7 +79,7 @@ fun WooPosErrorScreen(
                     text = it.text,
                     onClick = it.click,
                     modifier = Modifier
-                        .fillMaxWidth(.5f)
```
toupper, line 81: @kidinov In this case (and in the empty state), the button has a narrower width in the design. Let me know if you have a better idea for how to adapt it.

kidinov: That changes the state on all 11 places where this component is used. If this is what Wagner's idea was, then it's fine. Overall, I think it looks ok

kidinov: We might want to extract that constant so that changing the width will only require one change. This is because both buttons must have the same width

WooCommerce/src/main/res/drawable/ic_woo_pos_orders_empty.xml
```diff
@@ -0,0 +1,25 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
kidinov, line 1: This icon is not adapted for the dark mode. Notice in Figma that design system colors are used and you need to use them here as well

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -74,6 +77,22 @@ fun WooPosOrdersScreen(
     val onBackClicked = { onNavigationEvent(WooPosNavigationEvent.GoBack) }
     BackHandler { onNavigationEvent(WooPosNavigationEvent.GoBack) }
 
+    if (state is WooPosOrdersState.Error) {
```
kidinov: Generally, sealed class -> when statement without else -> compilation time check that all cases are handled.

And I think there’s some issue with structure. Your top-level state goes down to what is "Content" state, while you still treat as top-level `WooPosOrdersState` state

The toolbar should be visible all the time, not rerendered on the state change. Also there are some naming issues. I tried to address those in the patch, please let me know wdyt?

[Refactor_WooPosOrdersScreen_to_improve_state_handling_and_simplify_composables.patch](https://github.com/user-attachments/files/22718971/Refactor_WooPosOrdersScreen_to_improve_state_handling_and_simplify_composables.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -97,6 +97,15 @@ class WooPosOrdersViewModel @Inject constructor(
         loadMoreIfPossible()
     }
 
+    fun onOrdersEmptyActionClicked() {
+        // Action to be defined
```
kidinov, line 101: Just sharing my perspective - I think the only CTA needed on an empty state is "create a new something" one, and here it doesn’t apply. I think only what we need is to have pull to refresh enabled on this state, and that's it. There is nothing to "learn more" here.

kidinov (commented):
Thanks for your work here! I left a few comments; please take a look and let me know what do you think?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
toupper: Good idea, I created a shared modifier here 3c3b5f40359ecef6499c7ed53c19cbb82504a341

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Thanks for your input @kidinov! Sharing with our project lead @staskus

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
toupper: Note, make it more specific by not calling it just Button

WooCommerce/src/main/res/drawable/ic_woo_pos_orders_empty.xml
toupper: As we talked, we will tackle all the dark mode POS assets in one Linear task

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Great, thanks for the patch and the explanation in our call, I've added it in 2aec62a907479523b7d36aba137f14c876d5817c

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
staskus: My first move was to show "Refresh" as well, but the design change was to go from "Refresh" to "Learn More". pdfdoF-7Av-p2#comment-9460

kidinov (approved):
LGTM!

The only thing I think we should do is to extract the empty screen state to the components, as currently it's part of the "items list." And name `WooPosButtonModifier` more specifically

Something like that 


[Add_WooPosEmptyScreen_component_and_refactor_empty_list_handling.patch](https://github.com/user-attachments/files/22747032/Add_WooPosEmptyScreen_component_and_refactor_empty_list_handling.patch)

toupper:
Thanks for the review @kidinov!

> The only thing I think we should do is to extract the empty screen state to the components, as currently it's part of the "items list." And name WooPosButtonModifier more specifically

Done in 189e0a46e5ad052490cf379ab445f23d4f53754e

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #14702, opened 2025-10-06
Update developer.properties-example

malinajirka:
Update developer.properties-example

### DIFF
--- a/developer.properties-example
+++ b/developer.properties-example
@@ -1,11 +1,12 @@
-# The type of in app update FLEXIBLE = 0; IMMEDIATE = 1
-# Changing these values will affect performance of remote build cache on local builds.
+# Build settings
+use_remote_build_cache_locally=false
+
+# Changing values below  will affect performance of remote build cache on local builds.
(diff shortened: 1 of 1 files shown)


developer.properties-example
```diff
@@ -1,11 +1,12 @@
...
+
+# Changing values below  will affect performance of remote build cache on local builds.
 # Consider using the default values for local builds, if it's not necessary to change them.
 # See details on internal P2: https://wp.me/paaHJt-82c
+
+# The type of in app update FLEXIBLE = 0; IMMEDIATE = 1
 wc.in_app_update_type=0
 wc.jitm_testing_json_file_name=
 wc.force_password_login=false
 enable_leak_canary=true
```
wzieba: WDYT about disabling this by default?

```suggestion
enable_leak_canary=false
```

malinajirka: Done in 6f3f4e1d20b0c9efb303d88a9e78c22fad8886b5

wzieba (approved):

PR #14712, opened 2025-10-07
[WOOMOB-1433][Woo POS][Surveys] Schedule current user survey notification on app start

kidinov:
WOOMOB-1433

### Description
This PR adds functionality to schedule a current user survey notification when the app starts. The notification is scheduled with a 5-minute delay for users who have already used Woo POS at least once.

The implementation:
- Injects `WooPosSurveysNotificationSchedular` into `AppInitializer`
- Calls `schedularCurrentUserSurveyNotification()` during app initialization in `onFirstActivityResumed()`
- Schedules notification only when:
  - User has opened Woo POS at least once
  - Notification hasn't been shown before
  - Country is US or GB
  - Feature flag is enabled

### Steps to reproduce
1. Open the app with a fresh install
2. Open Woo POS at least once
3. Close and reopen the app
4. Wait 5 minutes (or change the code to wait less)
5. A survey notification should appear

### Testing information
- The notification scheduling logic checks multiple conditions before scheduling
- The notification has a 5-minute delay after app start
- Only users in US or GB will receive the notification
- Users who have never opened Woo POS will not receive this notification

- Unit tests verify all notification scheduling conditions
- Tests cover scenarios: notification scheduled when conditions met, not scheduled when already shown, not scheduled when Woo POS not opened before, not scheduled for non-allowed countries
- Manual testing: app start behavior, notification scheduling with delay

image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
@@ -44,7 +44,7 @@ import com.woocommerce.android.ui.jitm.JitmStoreInMemoryCache
 import com.woocommerce.android.ui.login.AccountRepository
 import com.woocommerce.android.ui.main.MainActivity
 import com.woocommerce.android.ui.payments.cardreader.onboarding.CardReaderOnboardingChecker
-import com.woocommerce.android.ui.woopos.common.util.WooPosSurveysNotificationSchedular
+import com.woocommerce.android.ui.woopos.common.util.WooPosSurveysNotificationScheduler
 import com.woocommerce.android.ui.woopos.localcatalog.WooPosLocalCatalogSyncScheduler
 import com.woocommerce.android.util.AppThemeUtils
 import com.woocommerce.android.util.ApplicationEdgeToEdgeEnabler
@@ -166,7 +166,7 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
 
     @Inject lateinit var posLocalCatalogScheduler: WooPosLocalCatalogSyncScheduler
 
-    @Inject lateinit var wooPosSurveysNotificationSchedular: Lazy<WooPosSurveysNotificationSchedular>
+    @Inject lateinit var wooPosSurveysNotificationScheduler: Lazy<WooPosSurveysNotificationScheduler>
 
     private var connectionReceiverRegistered = false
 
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
```diff
@@ -165,6 +166,8 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
 
     @Inject lateinit var posLocalCatalogScheduler: WooPosLocalCatalogSyncScheduler
 
+    @Inject lateinit var wooPosSurveysNotificationSchedular: Lazy<WooPosSurveysNotificationSchedular>
```
malinajirka: 🔍 For consistency, I'd consider sticking with the 'schedulEr' spelling.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/util/WooPosSurveysNotificationSchedular.kt
```diff
@@ -19,10 +20,14 @@ class WooPosSurveysNotificationSchedular @Inject constructor(
 ) {
     companion object {
         private val ALLOWED_COUNTRIES = setOf("us", "gb")
+        private const val CURRENT_USER_NOTIFICATION_DELAY_MINUTES = 5L
     }
 
     suspend fun schedularPotentialUserSurveyNotification() {
```
malinajirka: 🔍 Potential typo in the name of these two methods -> `schedular` instead of a verb `schedule`?

malinajirka (approved):
Thanks @kidinov! I've tested various scenarios and it all seems to work as expected.

One note - I don't see a feature flag and the app crashes upon tapping on the notification. I assume you are aware of this and plan to implement it before the code-freeze.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/util/WooPosSurveysNotificationSchedular.kt
kidinov: Fixed here - 45b9506ffe297f79962c4438e03b3dc4eb51adfc

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
kidinov: Fixed here - 45b9506ffe297f79962c4438e03b3dc4eb51adfc

kidinov:
> One note - I don't see a feature flag and the app crashes upon tapping on the notification. I assume you are aware of this and plan to implement it before the code-freeze.

There is a feature flag on the notification scheduling, so it should  never have been shown in the first place. Or am I missing something?

PR #14713, opened 2025-10-07
[CIAB] Bookings empty state

hichamboushaba:
Closes WOOMOB-1361

> [!IMPORTANT]
> Please review the  commit [1e900c6](https://github.com/woocommerce/woocommerce-android/pull/14713/commits/1e900c6e0fdfecfd22de3076f5de16cfa4ba3c74) independently, it's not related to the changes, but it's an overdue improvement.

### Description
This PR adds the empty states of the bookings list. We have three versions of empty states:
- Search empty state.
- Filters empty state
- Default empty state

For the third empty view, we have a different title and description based on the selected tab.

### Steps to reproduce
1. Enable API faker with the importing the following configuration (API Faker is available in Developer Options):
<details>
<summary> Config </summary>

```json
[
  {
    "request": {
      "httpMethod": "GET",
      "id": 0,
      "path": "/wc-bookings/v2/bookings",
      "queryParameters": [],
      "type": {
        "type": "wp-api"
      }
    },
    "response": {
      "body": "[]",
      "endpointId": 0,
      "statusCode": 200
    }
  }
]
```
</details>

2. Open the bookings tab
3.  For filtering: you need to update the `areFiltersEnabled` manually to `true` [here](https://github.com/woocommerce/woocommerce-android/blob/6ddc5a899437a0c00e22ddeb07c95d780942e840/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewState.kt#L21)

### Testing information
- Check and confirm that the empty view looks as expected.

The above.

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20251007_152811 | Screenshot_20251007_152802 |
| Screenshot_20251007_141323 | Screenshot_20251007_141336 |
| Screenshot_20251007_140933 | Screenshot_20251007_140946 |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/res/drawable-night/img_empty_search.xml
```diff
@@ -1,6 +1,6 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-    android:width="104dp"
-    android:height="104dp"
+    android:width="80dp"
+    android:height="80dp"
```
hichamboushaba: I updated the size to match the designs, we should be using this for all screens.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/creation/targets/BlazeCampaignTargetSelectionScreen.kt
```diff
@@ -226,7 +226,7 @@ private fun SearchList(
         }
 
         is SearchState.NoResults -> {
-            val image = painterResource(id = R.drawable.search_failed_illustration)
+            val image = painterResource(id = R.drawable.img_empty_search)
```
hichamboushaba, line 229: I removed the duplicated image, although they had a slightly different size, the new one still looks good and more consistent with the other screens:

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -123,13 +128,17 @@ fun BookingListScreen(state: BookingListViewState) {
...
-                        text = "No bookings found",
-                        modifier = Modifier
-                            .padding(paddingValues)
-                            .wrapContentSize()
-                    )
+                    val modifier = Modifier
+                        .background(MaterialTheme.colorScheme.surface)
+                        .fillMaxSize()
+                        .padding(paddingValues)
+                        .padding(32.dp)
```
AdamGrzybkowski: This padding takes some of the precious screen estate when the device has a font/display scaling enabled. I think we might be able to keep it once we enable scrolling on the empty views. (I will mention that elsewhere)

Screenshot_20251008_093101

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -123,13 +131,23 @@ fun BookingListScreen(state: BookingListViewState) {
...
+                            query = state.searchState.query.orEmpty(),
+                            modifier = modifier
+                        )
+                    } else {
+                        EmptyView(
+                            selectedTab = state.tabState.selectedTab,
+                            areFiltersActive = state.areFiltersActive,
+                            modifier = modifier
+                        )
+                    }
```
AdamGrzybkowski: ⚠️ We need to make those scrollable. You can't see the whole UI in both landscape and portrait with display/font scaling enabled. 

Screenshot_20251008_092922

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
```diff
@@ -123,13 +131,23 @@ fun BookingListScreen(state: BookingListViewState) {
...
-                        modifier = Modifier
-                            .padding(paddingValues)
-                            .wrapContentSize()
-                    )
+                    val modifier = Modifier
+                        .background(MaterialTheme.colorScheme.surface)
+                        .fillMaxSize()
+                        .padding(32.dp)
+
+                    if (state.searchState.isSearchActive) {
```
AdamGrzybkowski: I'm not sure this is enough - I found it weird when the empty search view with an empty string was shown. This can happen if you tap on the search bar from an already empty state. I know it's a weird thing to do, to try search an empty list, but since it's possible, I think we should handle it. 


https://github.com/user-attachments/assets/264bfb69-8de7-4926-a8f5-5ce39cd7837e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
AdamGrzybkowski: Also, this might not be directly related to this PR, but it is related to scrollable content. With the keyboard shown in landscape, the sort/filter chips are compressed into tiny elements. We should consider making them part of the scrollable area.

Screenshot_20251008_095606

AdamGrzybkowski (changes requested):
Looks good in general, but I found some issues in landscape and/or with font/display scaling that I think have to be fixed before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: @AdamGrzybkowski I made some adjustments to improve the situation in the commit 6d97fb5, please check it out.

For the list UI, I opted to keep the filters tab header sticky to match the other screens, let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: Good catch, I didn't catch this because I didn't use the ApiFaker approach for testing search 😅, fixed in af8a145

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListScreen.kt
hichamboushaba: @AdamGrzybkowski I made some adjustments to improve the situation in the commit 73c1c74eba8436fc56336a4dbccb255b206f4923, please check it out.

For the list UI, I opted to keep the filters tab header sticky to match the other screens, let me know what you think.

hichamboushaba: I deleted my last comment, I just noticed an issue with my improvements, please hold a bit from reviewing them 🙏

hichamboushaba: This should be good now, please check the commit e33fcbb, I adjusted the screen layout to ensure the empty screen is scrollable, and also that the filters bar is scrollable too.

https://github.com/user-attachments/assets/53c6d2f1-f947-4e09-93b0-50ba72d0d463

For the list state, I opted to keep the filters bar sticky to match the behavior of the other screens, please let me know what you think about this.

AdamGrzybkowski (approved):
🔥

PR #14717, opened 2025-10-07
[WOOMOB-1068] Add initial implementation of Local Catalog Settings

malinajirka:
#WOOMOB-1068

### Description

This PR implements a first version of the Local Catalog settings. 
- The designs aren't ready yet, so the UI layer can be mostly ignored.
- There are still some TBD comments, since the PR size reached the upper limit.
     - [Errors during catalog refresh are ignored.](https://linear.app/a8c/issue/WOOMOB-1447/woo-poslocal-catalog-handle-manual-sync-errors-in-pos-settings)
     - [Catalog size is hardcoded.](https://linear.app/a8c/issue/WOOMOB-1444/woo-poslocal-catalog-calculate-catalog-size-in-pos-settings)
     - [Both Last update and Last full update use the same date, since we are currently storing just one date.](https://linear.app/a8c/issue/WOOMOB-1445/woo-poslocal-catalog-use-full-sync-timestamp-in-pos-settings)
     - [Save 'Download over cellular' toggle state](https://linear.app/a8c/issue/WOOMOB-1446/woo-poslocal-catalog-save-downloadovercellular-preference-in-pos)
 
Unit tests - I skipped unit tests for the VM, since the logic is just dummy. I'll add them in the follow up PR.

Question - currently, the full refresh is performed within VM scope. I'm wondering whether we want to keep it like this or if we want the button to schedule an immediate background sync instead. Wdyt?

### Testing information

1. Open POS
2. Tap on the more icon
3. Select Settings
4. Tap on Catalog
5. Notice the Local Catalog settings
-----
Verify the Catalog section is not visible when the Local Catalog FF is turned off.

The above

### Images/gif

Screenshot_1759855996


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogScreen.kt
```diff
@@ -0,0 +1,290 @@
+package com.woocommerce.android.ui.woopos.settings.details.localcatalog
```
malinajirka, line 1: 📓  Feel free to ignore this file, since we don't have final designs yet.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosSettingsLocalCatalogViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogState.kt
```diff
@@ -0,0 +1,14 @@
+package com.woocommerce.android.ui.woopos.settings.details.localcatalog
+
+data class WooPosSettingsLocalCatalogState(
```
kidinov, line 3: I think this state declaration is a bit ambiguous:

  1. Nullable catalogStatus - when it's `null`, does that mean loading or error - not clear
  2. Overlapping flags - you could have `isLoading` and `isRefreshing` both `true` at the same time, which is can be correct, but I am not sure. When you hit refresh I'd assume we may want to show catalog status "loading" too?

Maybe something like that?

```
data class WooPosSettingsLocalCatalogState(
    val catalogStatus: CatalogStatus = CatalogStatus.Loading,
    val allowCellularDataUpdate: Boolean = false,
) {
    sealed class CatalogStatus {
        data class Available(
            val catalogSize: String,
            val lastUpdate: String,
            val lastFullUpdate: String
        ) : CatalogStatus()

        object Loading : CatalogStatus()
    }
}
```

And when it's "loading" maybe to not show "refresh button" at all?

kidinov (approved):
LGTM!

I'd just suggest thinking about the state declaration, making sure that it cannot be in an ambiguous state, and checking this in compile time. So ideally, to avoid flags and the nullables there.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogState.kt
malinajirka: Good callout! Fixed in https://github.com/woocommerce/woocommerce-android/pull/14717/commits/541fce315a37258b9fc7ea2ddaa367ce4338602f and https://github.com/woocommerce/woocommerce-android/pull/14717/commits/be52b507814785b50f7cdbc2a688a5a19bc65f79.

PR #14720, opened 2025-10-08
[WOOMOB-1421] List-Details layout for bookings on tablet

AdamGrzybkowski:
Part of WOOMOB-1421

### Description

This PR adds support for split pane layout on the bookings tab for bigger screens. The implementation mostly follows the Products tab logic. I highly recommend reviewing commit by commit, where I tried to fix issues one by one.

**About the `View Order` navigation logic.** 

The `OrderDetailFragment` has some logic that pretty much makes it impossible to show this Fragment outside the main graph with the Order list in the backstack. I've added an option to ignore it https://github.com/woocommerce/woocommerce-android/pull/14720/commits/dc55081626eff082eacb159c7158b0555d55bb35. Additionally, I've decided to show the Order details in full screen when in Split Pane. While this may not be the best, it's something we already do when opening a Product from Order.

### Testing

1. Launch the app on a tablet or a resizable emulator
2. Go to the Bookings tab
3. Play with it

### Images/gif

https://github.com/user-attachments/assets/03842b1a-b4ec-4406-be12-763251e8bc9e

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
@@ -14,6 +14,8 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
@@ -28,12 +30,14 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.shadow
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.dimensionResource
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.times
 import com.woocommerce.android.R
+import com.woocommerce.android.extensions.isTwoPanesShouldBeUsed
 import com.woocommerce.android.ui.bookings.compose.BookingAppointmentDetails
 import com.woocommerce.android.ui.bookings.compose.BookingAppointmentDetailsModel
 import com.woocommerce.android.ui.bookings.compose.BookingAttendanceSection
@@ -85,11 +89,13 @@ fun BookingDetailsScreen(
     onViewOrder: (Long) -> Unit,
     onViewNotes: () -> Unit,
 ) {
+    val context = LocalContext.current
     val showAttendanceSheet = remember { mutableStateOf(false) }
     Scaffold(
         topBar = {
             Toolbar(
                 title = viewState.toolbarTitle,
+                navigationIcon = if (context.isTwoPanesShouldBeUsed) null else Icons.AutoMirrored.Filled.ArrowBack,
                 onNavigationButtonClick = onBack,
                 modifier = Modifier.shadow(4.dp)
             )
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
@@ -4,6 +4,7 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
@@ -101,34 +102,37 @@ fun BookingDetailsScreen(
             )
         }
     ) { innerPadding ->
-        WCPullToRefreshBox(
-            isRefreshing = viewState.loadingState == BookingDetailsLoadingState.Refreshing,
-            onRefresh = viewState.onRefresh,
-            state = rememberPullToRefreshState(),
-            modifier = Modifier
-                .fillMaxSize()
-                .padding(innerPadding)
-                .detailsPanePadding(),
-        ) {
-            Column(
+        if (viewState.shouldShowEmptyState) {
+            BookingDetailsEmptyScreen()
+        } else {
+            WCPullToRefreshBox(
+                isRefreshing = viewState.loadingState == BookingDetailsLoadingState.Refreshing,
+                onRefresh = viewState.onRefresh,
+                state = rememberPullToRefreshState(),
                 modifier = Modifier
-                    .verticalScroll(rememberScrollState())
+                    .fillMaxSize()
+                    .padding(innerPadding)
+                    .detailsPanePadding(),
             ) {
-                when {
-                    viewState.shouldShowSkeleton -> BookingDetailsLoading()
-                    viewState.bookingUiState != null -> {
-                        BookingDetailsContent(
-                            booking = viewState.bookingUiState,
-                            onCancelBooking = viewState.onCancelBooking,
-                            onViewOrder = onViewOrder,
-                            onAttendanceStatusClicked = { showAttendanceSheet.value = true },
-                            onViewNotes = onViewNotes,
-                            onMarkAsPaid = viewState.onMarkAsPaid,
-                            markAsPaidInProgress = viewState.paymentUpdateStatus == PaymentUpdateStatus.InProgress,
-                        )
+                Column(
+                    modifier = Modifier
+                        .fillMaxSize()
+                        .verticalScroll(rememberScrollState())
+                ) {
+                    when {
+                        viewState.shouldShowSkeleton -> BookingDetailsLoading()
+                        viewState.bookingUiState != null -> {
+                            BookingDetailsContent(
+                                booking = viewState.bookingUiState,
+                                onCancelBooking = viewState.onCancelBooking,
+                                onViewOrder = onViewOrder,
+                                onAttendanceStatusClicked = { showAttendanceSheet.value = true },
+                                onViewNotes = onViewNotes,
+                                onMarkAsPaid = viewState.onMarkAsPaid,
+                                markAsPaidInProgress = viewState.paymentUpdateStatus == PaymentUpdateStatus.InProgress,
+                            )
+                        }
                     }
-
-                    else -> BookingDetailsEmptyScreen()
                 }
             }
         }
@@ -235,18 +239,24 @@ private fun BookingDetailsLoading() {
 }
 
 @Composable
-fun BookingDetailsEmptyScreen() {
-    Column(
-        modifier = Modifier.fillMaxSize(),
-        horizontalAlignment = Alignment.CenterHorizontally,
-        verticalArrangement = Arrangement.Center
+fun BookingDetailsEmptyScreen(
+    modifier: Modifier = Modifier
+) {
+    Box(
+        modifier = modifier.fillMaxSize(),
     ) {
-        Image(
-            painter = painterResource(id = R.drawable.img_woo_generic_error),
-            contentDescription = null
-        )
-        Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.major_200)))
-        Text(text = stringResource(R.string.booking_not_selected))
+        Column(
+            modifier = Modifier.fillMaxSize(),
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center
+        ) {
+            Image(
+                painter = painterResource(id = R.drawable.img_woo_generic_error),
+                contentDescription = null
+            )
+            Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.major_200)))
+            Text(text = stringResource(R.string.booking_not_selected))
+        }
     }
 }
 
@@ -312,6 +322,7 @@ private fun BookingDetailsLoadingPreview() {
             viewState = BookingDetailsViewState(
                 toolbarTitle = "",
                 bookingUiState = null,
+                loadingState = BookingDetailsLoadingState.Refreshing,
             ),
             onBack = {},
             onViewOrder = {},
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
@@ -46,6 +46,7 @@ import com.woocommerce.android.ui.compose.Render
 import com.woocommerce.android.ui.compose.animations.SkeletonView
 import com.woocommerce.android.ui.compose.component.Toolbar
 import com.woocommerce.android.ui.compose.component.WCPullToRefreshBox
+import com.woocommerce.android.ui.compose.modifier.detailsPanePadding
 import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
 import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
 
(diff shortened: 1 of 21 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>BookingsCommunicationViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/util/TabletLayoutSetupHelper.kt
```diff
@@ -161,7 +161,7 @@ class TabletLayoutSetupHelper @Inject constructor(private val context: Context)
     }
 
     private fun adjustLayoutForNonTablet(screen: Screen) {
-        if (screen.twoPanesWereShownBeforeConfigChange) {
+        if (screen.twoPanesWereShownBeforeConfigChange && screen.automaticallyAdjustLayoutAfterConfigChange) {
```
AdamGrzybkowski, line 164: I don't like this logic when switching from split pane to single pane. It shows the detail pane, and we end up with a different navigation stack compared to if we opened the booking details from the list in a single pane from the beginning.  This causes some issues with back navigation. 

I wanted an option to disable this and handle it separately in the BookingListFragment.

WooCommerce/src/main/res/navigation/nav_graph_bookings_details.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<navigation xmlns:android="http://schemas.android.com/apk/res/android"
```
AdamGrzybkowski, line 2: This was separated from the `nav_graph_bookings` so that I can use this for the details pane.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/modifier/DetailsPanePadding.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.woocommerce.android.extensions.isTwoPanesShouldBeUsed
+import com.woocommerce.android.extensions.windowWidthSizeClass
+import org.wordpress.android.util.DisplayUtils
+
+/**
+ * Adds horizontal padding to detail screens when shown in a split-pane layout (tablets).
+ * The padding mirrors the margin calculations used in TabletLayoutSetupHelper.setDetailsMargins.
+ */
+@Composable
+fun Modifier.detailsPanePadding(): Modifier {
```
AdamGrzybkowski, line 21: `TabletLayoutSetupHelper.setDetailsMargins` doesn't work with `ComposeView` and even if we wrap the `ComposeView` in a ViewGroup we end up with the whole screen paddes horizontally, including the compose Toolbar. This is not how the details screen look on other tabs, so I've added this modifier to set directly on the booking details content composable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -46,8 +46,13 @@ class BookingDetailsViewModel @Inject constructor(
 
     private val navArgs: BookingDetailsFragmentArgs by savedState.navArgs()
 
-    private val booking = bookingsRepository.observeBooking(navArgs.bookingId)
-        .shareIn(scope = viewModelScope, started = SharingStarted.WhileSubscribed(), replay = 1)
+    private val bookingId: Long? = (navArgs.mode as? BookingDetailsFragment.Mode.ShowBooking)?.bookingId
```
AdamGrzybkowski, line 55: I'm not a big fan of this. I would like to rework how we model the VM state, but since this PR grew quite a bit, I would rather do it in a separate PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -131,19 +136,19 @@ class BookingDetailsViewModel @Inject constructor(
...
             if (!networkStatus.isConnected()) {
                 triggerEvent(MultiLiveEvent.Event.ShowSnackbar(R.string.offline_error))
                 return@launch
             }
-
             loadingState.value = initialLoadingState
 
             val bookingTask = async {
-                bookingsRepository.fetchBooking(navArgs.bookingId)
+                bookingsRepository.fetchBooking(bookingId ?: 0L)
```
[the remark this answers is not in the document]
AdamGrzybkowski: Yes, yes - that's what I want to fix as part of this https://github.com/woocommerce/woocommerce-android/pull/14720#discussion_r2478360297

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
AdamGrzybkowski: A follow up PR is here https://github.com/woocommerce/woocommerce-android/pull/14881

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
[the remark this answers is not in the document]
AdamGrzybkowski: Fixed here https://github.com/woocommerce/woocommerce-android/pull/14881

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
```diff
@@ -219,6 +228,22 @@ private fun BookingDetailsLoading() {
     }
 }
 
+@Composable
+fun BookingDetailsEmptyScreen() {
```
irfano: Have we discussed this empty state before? On the Orders and Products screens, we always have an item selected, but this screen behaves differently. If I change the sorting and the new list doesn’t include the previously selected booking, we show an empty screen. Alternatively, we could just show the first booking from the new list.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListFragment.kt
```diff
@@ -135,6 +135,7 @@ class ProductListFragment :
     override val detailPaneContainer: View
         get() = binding.detailNavContainer
     override var twoPanesWereShownBeforeConfigChange: Boolean = false
+    override val automaticallyAdjustLayoutAfterConfigChange: Boolean = true
```
irfano, line 138: I couldn’t quite understand why we’re handling configuration changes in Bookings but not in Products. Could you please elaborate on this? Also, how can I test this? I tried it on a tablet, and it always shows two panes in both orientations.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -55,13 +64,30 @@ class BookingListViewModel @Inject constructor(
...
+                }
+            }
+        }
+    }
+
     private val contentState = combine(
         bookingListHandler.bookingsFlow.map { bookings ->
+            openFirstLoadedBookingOnTablet(bookings)
             with(bookingMapper) { bookings.map { it.toListItem() } }
         },
```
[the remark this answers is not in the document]
irfano: I guess we’re already checking whether `selectedBookingIdOnBigScreen == null` to prevent duplicate navigation events. _(Not a suggestion, just confirming my understanding.)_

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -55,13 +66,30 @@ class BookingListViewModel @Inject constructor(
...
+
     private val isSortSheetVisible = MutableStateFlow(false)
 
     private var bookingsFetchJob: Job? = null
     private var bookingsLoadMoreJob: Job? = null
 
+    private fun openFirstLoadedBookingOnTablet(bookings: List<BookingEntity>) {
+        if (isWindowClassLargeThanCompact()) {
+            if (bookings.isNotEmpty()) {
+                if (selectedBookingIdOnBigScreen == null) {
```
irfano: I guess it would be nicer to combine these three if conditions into one.

irfano (changes requested):
Nice job! This wasn’t an easy task. I asked a few questions to understand it better, noticed some minor issues, and made a few suggestions.

#### Padding issue
Screenshot_20251101_002320
After tapping "Add note" and navigating back to the booking details, an unwanted bottom gap appears.

#### "View customer orders" button issue
Screenshot_20251101_002506
This button doesn’t do anything when the order detail screen is opened from the booking details.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
irfano: Perhaps we can improve this empty screen. I’m adding the empty order details screen for comparison. We can use the “No booking selected” text as the title, just like we do on the empty order details screen. I’m fine with leaving out the description. Just the title above and the illustration below should be fine.

|Empty order details|Empty booking details|
|-|-|
|||

AdamGrzybkowski:
>After tapping "Add note" and navigating back to the booking details, an unwanted bottom gap appears.

Yeah, this seems to be a general problem with how we handle insets with edge-to-edge in the app and the combination with Jetpack Compose. This can be reproduced on phones as well, especially when you use the `back button` from the toolbar. I think it's separate from this PR, so I created an issue for this - WOOMOB-1630

AdamGrzybkowski:
>"View customer orders" button issue

Ohh, that's a good catch. In fact, I'm not sure what would be the best way to handle it. This button applies the filter to only match this customer and shows the order list screen... I could hide that button, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
AdamGrzybkowski: I tried to match the design with the Products tabs - that how it looks:

Screenshot_20251103_110025

I have to change the style of the text to match it better. 

But the Order one is different. I can match only one of them though. Would you prefer the Order empty screen style?

AdamGrzybkowski: >Have we discussed this empty state before? On the Orders and Products screens, we always have an item selected, but this screen behaves differently. If I change the sorting and the new list doesn’t include the previously selected booking, we show an empty screen. Alternatively, we could just show the first booking from the new list.

Not really, I was just trying to replicate the behavior we have. I will look into that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
AdamGrzybkowski: Done https://github.com/woocommerce/woocommerce-android/pull/14720/commits/4c51d87bd7e5974d4adaa7d92013db557aa5e7c5

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListFragment.kt
AdamGrzybkowski: I've made a quick comment mentioning that here https://github.com/woocommerce/woocommerce-android/pull/14720#discussion_r2413822787. 

We do handle it in the Products tab, but it's fully done in `TabletLayoutSetupHelper`. There are some problems with this logic, so that's why I wanted to disable that in the Bookings tab.

In short, when you transition from a two-pane to a single-pane layout:

**Products**
The list pane is hidden, and the details pane gets full width. 

**Bookings**
The list pane gets full width, and we show the Booking details fragment properly by adding it to the main nav graph.

The main difference is that the single-pane navigation stack is fully restored. In addition, the bottom bar visibility is correct. 

Here's the recording from the Products tab:


https://github.com/user-attachments/assets/7383b8e5-e8c9-4ca2-bf41-202d93910223

You can observe the following:
1. At 0:16 second mark you can see that the first two pane to single pane transition failed. The product details screen wasn't shown. 
2. Next it worked, but you can see that the bottom bar is visible (it shouldn't)
3. Then I opened the same details again - notice the bottom bar visibility is different.
4. Then I navigated back - the previous details pane is visible !?

Here's the Bookings tab for a comparison:


https://github.com/user-attachments/assets/0032b020-20a0-4d95-9ad5-24e14ee3c436

To test, you need a resizable emulator or a foldable device big enough to show two panes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
AdamGrzybkowski: >On the Orders and Products screens, we always have an item selected, but this screen behaves differently. If I change the sorting and the new list doesn’t include the previously selected booking, we show an empty screen.

I was confused about this behavior you described because it shouldn't be happening. The problem is that we do clear the DB when we change the sort order, and this wipes out the Booking that is shown in the details pane... I will check how to prevent that.

AdamGrzybkowski: @irfano I've pushed the changes here https://github.com/woocommerce/woocommerce-android/pull/14720/commits/8ee6908a58bc71996b88a173f39b376f5de1c4be. The selected Booking should now be kept when you switch the order. 
As mentioned above, the problem was DB clearing not the selection logic.

irfano:
Another note: We’re adding a color to the selected order item in the list. For consistency, it would be nice to follow the same behavior for bookings as well, but feel free to handle this in a separate PR.

Screenshot_20251103_172329

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
irfano: Looks good! 👍🏻

irfano:
> Ohh, that's a good catch. In fact, I'm not sure what would be the best way to handle it. This button applies the filter to only match this customer and shows the order list screen... I could hide that button, WDYT?

I think hiding the button is the best approach in this case.

AdamGrzybkowski:
>Another note: We’re adding a color to the selected order item in the list. For consistency, it would be nice to follow the same behavior for bookings as well, but feel free to handle this in a separate PR.

Good call, I've added an issue for that - WOOMOB-1632

AdamGrzybkowski:
>I think hiding the button is the best approach in this case.

Done in https://github.com/woocommerce/woocommerce-android/pull/14720/commits/f7963a1b29a7050ac4d5a7c49128c59d1b2a5523

I've decided to check whether there's an order list in the backstack or we are in the order detail pane. Otherwise, the button is hidden. This should also work in case the Order screen is shown from other parts of the app that don't support that logic or where it doesn't make sense.  

Alternatively, I could use the nav argument `ignoreTwoPaneLayoutLogic`. This would limit the changes to the scenario you tested, but I think the above is better as it will handle more scenarios. 

FYI. This button doesn't work in the current `trunk` either, so it's great that you found it!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
irfano: > Would you prefer the Order empty screen style?

It doesn’t matter. I hadn’t seen the empty Products screen before, so it’s fine, let’s keep it as it is.

irfano: > @irfano I've pushed the changes here https://github.com/woocommerce/woocommerce-android/commit/8ee6908a58bc71996b88a173f39b376f5de1c4be. The selected Booking should now be kept when you switch the order.

Great! The issue has been resolved.

irfano (approved):
> Done in https://github.com/woocommerce/woocommerce-android/commit/f7963a1b29a7050ac4d5a7c49128c59d1b2a5523

Nice! It's now hidden when navigating from Bookings.

All good now, thanks for addressing my feedback and for the thorough explanations to my questions. LGTM! 👍🏻

PR #14722, opened 2025-10-08
[CIAB] Fetch bookings' additional details from the orders endpoint

hichamboushaba:
Closes WOOMOB-1439

### Description
This PR implements the fetching logic discussed in pe5sF9-4Fm-p2, which does the following for each page:
- Call `GET /wc-bookings/v2/bookings`
- Call `GET /wc/v3/orders, passing include=[ids]`, where `[ids]` is the array of order IDs returned from the bookings API.

The PR updates the `Bookings` table to include more data around the order using `Room`'s `@Embedded` annotation, the resulting table schema can be found in this [line](https://github.com/woocommerce/woocommerce-android/blob/8c51d42637658db912887ba4281891581274c540/libs/fluxc-plugin/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/68.json#L3954).

### Steps to reproduce
1. Prepare a CIAB site with bookings plugin.
2. Prepare some bookings:
    - Bookings attached to an order.
    - Create an order with multiple bookings.
    - Create a booking that's not attached to any order (Just to confirm the app behaves correctly, this is not needed for CIAB)
3. Open the bookings tab.

### Testing information
- Confirm the order information is shown correctly for the bookings.
- Confirm that when there is no customer information, we correctly show `Guest`.

The above.

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -57,6 +58,22 @@ class BookingMapper @Inject constructor(
...
+            email = billingEmail,
+            phone = billingPhone,
+            billingAddressLines = listOfNotNull(
+                billingAddress1,
+                billingAddress2,
+                listOfNotNull(billingCity, billingState, billingPostcode).takeIf { it.isNotEmpty() }?.joinToString(" "),
+                billingCountry
+            ).ifEmpty { listOf("-") }
+        )
+    }
```
hichamboushaba, line 75: This and the other mapping changes are not final, I'll revise and add payments mapping in a separate PR, I just added these to test the implementation.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -576,10 +586,8 @@ class WCOrderStore @Inject internal constructor(
         )
     }
 
-    private fun fetchOrdersByIds(payload: FetchOrdersByIdsPayload) {
-        payload.orderIds.chunked(NUM_ORDERS_PER_FETCH).forEach { idsToFetch ->
```
hichamboushaba, line 580: I removed the usage of `chunked` here, with coroutines, it complicates the logic, and I don't think it's needed, it's up to the consumer to dictate how many orders they want to fetch.

Also, this doesn't have any impact on the current usage, the only two places where we call this is [here](https://github.com/woocommerce/woocommerce-android/blob/5df0cb7a1a8b1154158d5c2f310bcdc48937b07b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderFetcher.kt#L57) and [here](https://github.com/woocommerce/woocommerce-android/blob/c4ade750ee770f222e1c4f99ee69967072681a46/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/FetchOrdersRepository.kt#L52), both are used as part of the order list handling, and the list of orders we pass will never exceed a single page.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingDtoMapper.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+import org.wordpress.android.fluxc.persistence.dao.ProductsDao
+import org.wordpress.android.fluxc.persistence.entity.BookingEntity
+import org.wordpress.android.fluxc.persistence.entity.OrderEntity
+import java.math.BigDecimal
+import java.time.Instant
+import javax.inject.Inject
+
+internal class BookingDtoMapper @Inject constructor(
+    private val productsDao: ProductsDao
```
AdamGrzybkowski, line 13: I have mixed feelings about this. This is a mapper after all, and its responsibility should be limited to mapping between objects and not querying the DB. 

I believe the idea was not to provide the backup product name as a parameter in the `.toEntity()` extension function to avoid calling the DB for each item. What if we did something like this:


```kotlin
// In BookingsStore
val entities = response.result.map {
    with(bookingDtoMapper) {
        val orderEntity = ordersResult.model?.get(it.orderId)
        val productEntity = if (orderEntity == null) {
            productsDao.getProduct(site.localId().value, it.productId)
        } else null
        it.toEntity(
            localSiteId = site.localId(),
            orderEntity = orderEntity,
            productEntity = productEntity,
        )
    }
}
                    
// In BookingDtoMapper
order = orderEntity?.toBookingOrderInfo(orderItemId) ?: BookingOrderInfo(
    productInfo = productEntity?.let {
        BookingProductInfo(name = it.name)
    }
)
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -72,25 +88,17 @@ class BookingsStore @Inject constructor(
...
-        personCounts = personCounts?.map { it.toLong() },
-        localTimezone = localTimezone
-    )
+    private suspend fun fetchOrders(
+        site: SiteModel,
+        orderIds: List<Long>
+    ): WooResult<Map<Long, OrderEntity>> {
+        if (orderIds.isEmpty()) { return WooResult(emptyMap()) }
+
+        val result = orderStore.fetchOrdersByIds(WCOrderStore.FetchOrdersByIdsPayload(site, orderIds))
```
AdamGrzybkowski, line 97: ❓ From a quick look, it doesn't look like this saves the order in the DB. I also checked the flow in the app, and I saw the loading shimmer on each Order screen when entered for the first time.

AdamGrzybkowski (commented):
Great work! I left some comments for verification/discussion.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
hichamboushaba: You are right, I didn't handle the migration to coroutines well here 🤦, this should be fixed by the commit 5746aae.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingDtoMapper.kt
hichamboushaba: I understand the concern, but I prefer to stay pragmatic rather than idealistic. Given this approach, the current mapping method is simpler to follow and extend, for example, in the future, we could add logic to retrieve the customer from the database if it's missing, and it would be easy to add in the mapper too.

And just as a note, this is not a first in the app; we have other mappers that need to check the DB for additional information, like the `OrderMapper`, which needs the DB to get location details. And now looking at this, I think we might need something similar for the `BookingMapper` as well to print the correct country name 😅.

hichamboushaba: I sent the comment before finishing it 😅 

I just wanted to add: happy to adjust it if you have a strong opinion about it 🙂, let me know what you think.

AdamGrzybkowski: Got it, thanks for the explanation. I would probably try to stick to the single responsibility principle, but I won't block the PR because of it!

AdamGrzybkowski (approved):
:shipit:

PR #14725, opened 2025-10-09
[WOOMOB-1446] Implement cellular data preference for WooPos local catalog sync

malinajirka:
Fixes WOOMOB-1446

### Description
Implements cellular data preference functionality for WooPos local catalog synchronization. When users disable "Download over cellular" in the local catalog settings, background sync operations will only run when connected to WiFi, preventing unexpected data usage charges.

The implementation uses WorkManager's built-in network constraints at the OS level for optimal efficiency - the system won't even start the worker when network conditions don't match the user's preferences.

### Testing information
- Verify the cellular data toggle appears in WooPos local catalog settings
- Test that toggling the preference saves the setting persistently
- Confirm worker is rescheduled with updated constraints when you update the toggle

The above


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncScheduler.kt
```diff
@@ -33,45 +39,51 @@ class WooPosLocalCatalogSyncScheduler @Inject constructor(
...
+                .setBackoffCriteria(
+                    BackoffPolicy.EXPONENTIAL,
+                    1,
+                    TimeUnit.MINUTES
+                )
+                .build()
+
+            workManager.enqueueUniquePeriodicWork(
+                WooPosLocalCatalogSyncWorker.WORK_NAME,
+                ExistingPeriodicWorkPolicy.UPDATE,
```
malinajirka, line 59: 📓 I replaced `Keep` with `Update` and wrapped the block in coroutine.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogViewModel.kt
```diff
@@ -47,18 +52,25 @@ class WooPosSettingsLocalCatalogViewModel @Inject constructor(
                 lastFullUpdate = formattedTimestamp // TBD local catalog: Replace with full sync timestamp
             )
 
+            val allowCellularDataUpdate = preferencesRepository.allowCellularDataUpdate.first()
```
kidinov: Consider making the cellular preference fully reactive by continuously listening to the repository instead of reading it once at startup. This way the UI state always stays in sync with the stored
  preference, and the Switch can pass its new value directly rather than the ViewModel toggling the current state.

Something like that. Wdyt?

[Refactor_cellular_data_update_handling_to_pass_boolean_value_and_improve_state_management.patch](https://github.com/user-attachments/files/22879211/Refactor_cellular_data_update_handling_to_pass_boolean_value_and_improve_state_management.patch)

kidinov (approved):
LGTM! I left one code improvement suggestion, please take a look

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogViewModel.kt
malinajirka: Thanks for the suggestion @kidinov! I'm wondering what is the benefit of observing, since the settings screen is the only place how this setting can be changed 🤔. The VM would be observing a repository that only the VM itself can change -> so the VM would update the setting and the VM would receive an update from the repo. I might be missing something, but I'm not sure what is the benefit.

Could you please elaborate on it a bit?

kidinov: 1. Unidirectional data flow: Data flows in one direction: UI → VM → Repo → VM → UI. This makes it easy to trace what's happening. No shortcuts where the VM updates itself directly.
  2. Less code: No need to flip !_state.value.allowCellularDataUpdate - the Switch already has the new value, just save it
  3. Can't get out of sync: Right now, if something goes wrong between updating the VM state and saving to repo, they could be different. With observing, that's impossible.
  4. Works if anything else changes it: currently there is assumption that it can be changed only from 1 place and in sync way, but if it changes asynchronously, it'll be out of sync with the view

  Bottom line: We don't maintain two copies of the same data. The repo is the source of the truth

malinajirka: Good points, I've tested and applied the patch 🙇.

e54cf5c63f1af75376e2669f2f8766edd9e8007c

PR #14729, opened 2025-10-09
[POS Historical Orders] Orders List UI

toupper:
Closes WOOMOB-1156

### Description

With this PR we apply the design to the POS Orders list, plus adding the data we miss.

### Steps to reproduce

1. Go to POS
2. Tap on Menu
3. Tap on Orders
4. Wait until they are loaded. See that they match the design.

### Images/gif

https://github.com/user-attachments/assets/903f50b5-0ba2-4f83-85f5-8b7b7f48fb04


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -304,7 +304,7 @@ class WooPosOrdersViewModel @Inject constructor(
                 applyDecimalFormatting = true
             )
 
-            val formattedStatus = order.status.toString().replaceFirstChar {
+            val formattedStatus = order.status.value.replaceFirstChar {
```
kidinov: Can be just 

```
            val formattedStatus = order.status.value
                .replaceFirstChar { it.titlecase(locale) }
                .replace("-", " ")
```

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
```diff
@@ -88,6 +88,15 @@ fun Date.formatToEEEEMMMddhha(locale: Locale): String {
     return dateFormat.format(this)
 }
 
+fun Date.formatToMMMddYYYYAtHHmm(
+    locale: Locale = Locale.getDefault(),
+    atWord: String
+): String {
+    val datePart = SimpleDateFormat("MMM d, yyyy", locale).format(this)
```
kidinov: Can be directly

```
fun Date.formatToMMMddYYYYAtHHmm(
    locale: Locale = Locale.getDefault(),
    atWord: String
): String = SimpleDateFormat("MMM d, yyyy '$atWord' h:mm a", locale).format(this)
```

Also, I'd say we need to have this in the woopos package to no create extra dependency. It'll be easier to modularize this eventually

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -390,3 +438,24 @@ private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
 fun WooPosOrdersScreenPreview() {
     WooPosTheme { WooPosOrdersScreen(onNavigationEvent = {}) }
 }
+
+@Composable
+fun OrderStatusBadge(status: PosOrderStatus) {
+    val bgColor = when (status.colorKey) {
```
kidinov, line 475: I think we should use compose colors. Check `WooPosTheme.kt`. If something is missing then we have to add it there

Also, keep in mind dark mode - now it's all one color in both.

Also, I think it should be in the VM, not in the view. View should just receive the color to use

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
```diff
@@ -1,17 +1,43 @@
 package com.woocommerce.android.ui.woopos.orders
 
 import androidx.compose.runtime.Immutable
+import com.woocommerce.android.model.Order.Status
 import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearchInputState
 import com.woocommerce.android.ui.woopos.home.items.WooPosPaginationState
 import com.woocommerce.android.ui.woopos.home.items.WooPosPullToRefreshState
 
+enum class OrderStatusColorKey {
```
kidinov: Why is this above the actual class in the file? Usually if the file is `WooPosOrdersState.kt` then the class that it contains goes first

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
```diff
@@ -1,17 +1,43 @@
...
+    COMPLETED,
+    FAILED,
+    PROCESSING,
+    ON_HOLD,
+    OTHER;
```
kidinov: Why are all other statuses ignored/defaulted to other?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: Also, It won't be localized - it's that you wanted here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
toupper: Because these other statuses all have the same color. Check iOS [here](https://github.com/woocommerce/woocommerce-ios/pull/16184/files#diff-1d4e2343c6e96da4a48b6471c100bfeea0b2aa3b33a52212d4979e79824667e2) `POSOrderBadgeView`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: > I think we should use compose colors. Check WooPosTheme.kt. If something is missing then we have to add it there

With the current implementation we have one source of truth for the order badge colors in the Orders tab and here. If we use compose colors here, wouldn't we be declaring the colors for the badge twice?

> Also, I think it should be in the VM, not in the view. View should just receive the color to use

Good question. I’ve thought about that for a while, and I came to the conclusion that colors belong in the UI layer. I also confirmed this with AI, which agreed with that approach:

> ✅ In general: colors should not live in the ViewModel — they belong in the UI layer (View / Composables / UI state classes).
> 
> Here’s the reasoning:
> 	•	ViewModel’s job: hold and expose state and data, not visual details. It should provide a status (like Order.Status.Pending) or a colorKey (OrderStatusColorKey.PENDING) — but not actual colors.
> 	•	View/UI’s job: interpret that state and decide how to represent it visually — e.g., which Color or ColorRes to use for a PENDING status.


With that in mind, I created `OrderStatusColorKey` in the state, to hold that color key to be interpreted by the view.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -222,47 +232,85 @@ private fun OrdersList(
...
+                                    color = Color.Black,
+                                    shape = MaterialTheme.shapes.medium
+                                )
+                            } else {
+                                Modifier
+                            }
+                        )
+                        .clickable { onOrderSelected(item.id) }
+                        .semantics { selected = item.isSelected }
+                        .padding(
```
kidinov: Can be just

```
.padding(WooPosSpacing.Medium.value)
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: > Also, It won't be localized - it's that you wanted here?

Good point. Do you know the best way to localize the status? I took this implementation from another part of the code — specifically the Wear app — but I see that it might not be localized there either.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -222,47 +232,85 @@ private fun OrdersList(
...
+                            color = MaterialTheme.colorScheme.onSurface,
+                            fontWeight = FontWeight.Bold
+                        )
+
+                        Spacer(modifier = Modifier.width(WooPosSpacing.Medium.value))
+
+                        WooPosText(
+                            item.date,
+                            style = WooPosTypography.BodySmall,
+                            color = MaterialTheme.colorScheme.onSurfaceVariant
```
kidinov: I think this supposed to be (check Figma)

```
color = WooPosTheme.colors.onSurfaceVariantHighest,
```

The same for the email

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -390,3 +438,24 @@ private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
...
+fun OrderStatusBadge(status: PosOrderStatus) {
+    val bgColor = when (status.colorKey) {
+        OrderStatusColorKey.COMPLETED -> R.color.tag_bg_completed
+        OrderStatusColorKey.FAILED -> R.color.tag_bg_failed
+        OrderStatusColorKey.PROCESSING -> R.color.tag_bg_processing
+        OrderStatusColorKey.ON_HOLD -> R.color.tag_bg_on_hold
+        OrderStatusColorKey.OTHER -> R.color.tag_bg_other
+    }
+
+    WooPosText(
```
kidinov, line 491: Usually the previews are at the bottom of the file, not the actual views

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: > With the current implementation we have one source of truth for the order badge colors in the Orders tab and here. If we use compose colors here, wouldn't we be declaring the colors for the badge twice?

Correct. And it's totally fine, as there are 2 different themes used for the main app and for the POS. And we don't want to mix them.

Also, more importantly, those are different colors at all. Check the Figma design. It's not `tag_bg_completed` but 

Image

While they may (may not) have the same value, they semantically different from the theme's point of view.

And as I said, you don't support dark mode here at all

> Good question. I’ve thought about that for a while, and I came to the conclusion that colors belong in the UI layer. I also confirmed this with AI, which agreed with that approach:

I guess it also fine, the main reason is that this `when` you have here is not testable, while it could be in the VM. Up to you!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -390,3 +438,24 @@ private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
 fun WooPosOrdersScreenPreview() {
```
kidinov: It would be nice to have working preview for easier testing/debugging of the UI

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -222,47 +232,85 @@ private fun OrdersList(
...
+                        )
+                        .clickable { onOrderSelected(item.id) }
+                        .semantics { selected = item.isSelected }
+                        .padding(
+                            horizontal = WooPosSpacing.Medium.value,
+                            vertical = WooPosSpacing.Medium.value
+                        ),
+                    verticalAlignment = Alignment.Top
+                ) {
+                    Column(verticalArrangement = Arrangement.spacedBy(WooPosSpacing.XSmall.value)) {
```
kidinov: The spacing between Order number, data, email, and the status does not match the design here

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -222,47 +232,85 @@ private fun OrdersList(
...
-                    .clip(MaterialTheme.shapes.medium)
-                    .background(background)
-                    .clickable { onOrderSelected(item.id) }
-                    .semantics { selected = isSelected }
-                    .padding(
-                        horizontal = WooPosSpacing.Medium.value,
-                        vertical = WooPosSpacing.Medium.value
-                    ),
-                verticalAlignment = Alignment.Top
+            WooPosCard(
```
kidinov, line 275: We'd need to relayout it to support smaller screens and larger fonts. Currently, it looks like that on the smaller screens

Image

We'd probably need to use constraint layout. We need to make sure that we show the price in full, and the rest can be "ellipsized" if there is not enough horizontal space.

kidinov (commented):
Great job! I left a few comments; please take a look!

There is also the loading state of the order that doesn't match a bit, but I'll take care of this when I work on the full loading state.

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: > Do you know the best way to localize the status? 

Sure! Can be as simple as that:
```

            val formattedStatus = when (order.status) {
                Order.Status.Cancelled -> resourceProvider.getString(R.string.woopos_orders_status_cancelled)
                Order.Status.Completed -> TODO()
                is Order.Status.Custom -> TODO()
                Order.Status.Failed -> TODO()
                Order.Status.OnHold -> TODO()
                Order.Status.Pending -> TODO()
                Order.Status.Processing -> TODO()
                Order.Status.Refunded -> TODO()
            }
```

There is another question if it should be localised - I don't know. I tried to change the language of wp-admin but it either doesn't work or I am doing something wrong

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -390,3 +438,24 @@ private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
...
+        OrderStatusColorKey.FAILED -> R.color.tag_bg_failed
+        OrderStatusColorKey.PROCESSING -> R.color.tag_bg_processing
+        OrderStatusColorKey.ON_HOLD -> R.color.tag_bg_on_hold
+        OrderStatusColorKey.OTHER -> R.color.tag_bg_other
+    }
+
+    WooPosText(
+        text = status.text,
+        style = WooPosTypography.BodySmall,
+        color = MaterialTheme.colorScheme.onSurface,
```
kidinov: Notice that the text color is not `onSurface` but depends on the status

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
toupper: Good points! Done in 246ac469d49140e76b6ec1618e8d300bce45d5a4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Nice, done in 9c9c4d41e444e609f8e3b9ded5e188bcdd385b49

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: That makes total sense, done in e44ea89035e5486ad044614b2c0f235b23e851c1

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
toupper: Good to know, done in c37ba594b0b8005d24de72875237d69097ee742f

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, done in c1ffed15eae64039872bb56a9467ed4cf5904763:

Screenshot_20251010_110309_Gallery

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, done in 063fa7141efed2b104f15ebab964082f9e54a34a and e1c136363eb26f7b39e579f5705294c667e406e6. I tried to adjust it according to the design label inner padding

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Oh nice, TIL about these colors available in `WooPosTheme`. Done in 8ce54343e9b998dac6736969f42b89dc5a104f1a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, thanks for adding more context. Done in c9fc39a450df2633be5f508162c279a8c8f654e8 Please note that some colors are still to be defined, hence the else.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Yeah, done in c9fc39a450df2633be5f508162c279a8c8f654e8

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Awesome, thanks. I checked with iOS, and it should be done locally. Done in 78de434e1926cb04abb374db24dedf1edd05ed30

toupper:
Thanks a lot for the thorough review @kidinov! I addressed the comments, please take a look again.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: @toupper I fixed the preview here

4ae7ac337389f1424828e93e4b414d11b19ed45c

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -385,8 +470,78 @@ private fun OrdersPaginationErrorRow(onPaginationErrorTryAgain: () -> Unit) {
     )
 }
 
+@Composable
+fun OrderStatusBadge(status: PosOrderStatus) {
+    val bgColor = when (status.colorKey) {
+        OrderStatusColorKey.COMPLETED -> WooPosTheme.colors.infoLowest
+        OrderStatusColorKey.FAILED -> WooPosTheme.colors.errorLowest
+        else -> WooPosTheme.colors.default
```
kidinov: Better to keep it compile time checked to replace else

```
    val bgColor = when (status.colorKey) {
        COMPLETED -> WooPosTheme.colors.infoLowest
        FAILED -> WooPosTheme.colors.errorLowest
        PROCESSING,
        ON_HOLD,
        OTHER -> WooPosTheme.colors.default
    }
```

and for the second color too

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/designsystem/WooPosTheme.kt
```diff
@@ -28,6 +28,15 @@ data class CustomColors(
     val onSurfaceVariantLowest: Color,
     val onSurfaceVariantHighest: Color,
 
+    val infoLowest: Color,
+    val errorLowest: Color,
+
+    val onInfoLowest: Color,
+    val onErrorLowest: Color,
+
+    val default: Color,
```
kidinov, line 37: I think we should work on the naming here. Or maybe we can just use what we have already in the palette:

```
    val bgColor = when (status.colorKey) {
        COMPLETED -> WooPosTheme.colors.infoLowest
        FAILED -> WooPosTheme.colors.errorLowest
        PROCESSING,
        ON_HOLD,
        OTHER -> MaterialTheme.colorScheme.surfaceContainerHigh
    }

    val textColor = when (status.colorKey) {
        COMPLETED -> WooPosTheme.colors.onInfoLowest
        FAILED -> WooPosTheme.colors.onErrorLowest
        PROCESSING,
        ON_HOLD,
        OTHER -> MaterialTheme.colorScheme.onSurface
    }
```

Image

Also, maybe to follow the other colors for consistency in terms of how it's structured:

```
    val infoLowest: Color,
    val onInfoLowest: Color,
    
    val errorLowest: Color,
    val onErrorLowest: Color,

    val default: Color,
    val onDefault: Color,
```

kidinov (approved):
LGTM! I left a couple, maybe important commens, please take a look!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Awesome, done in 1608fbb50439b4c8ef83521a2d59818a2e2be342

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/designsystem/WooPosTheme.kt
toupper: Thanks for the comment @kidinov! I updated the structure for consistency in 4d65a1f4df778cf0dce2f235db3fee3a448110fb Regarding the naming, `default` and `onDefault` are part of the Design System variable names, see here p1760091329238559-slack-C070SJRA8DP

PR #14734, opened 2025-10-10
Fix trash product button visibility on tablets

AdamGrzybkowski:
WOOMOB-1484

### Description

While fixing a bug related to the `Trash product` button visibility (https://github.com/woocommerce/woocommerce-android/pull/14657) I introduced this bug where the trash button is not visible in the split pane layout. 

This PR adds an additional check to make sure it's enabled in split pane even if the product list is not in the backstack.

### Steps to reproduce

1. Open the app on tablet
2. Go to products
3. Tap on the `...` menu button
4. Verify the `Trash product` is visible

### Testing information

The above plus:

Using a tablet
2. Go to order 
3. Tap on a product
4. Verify the `Trash product` is gone

Using a phone
1. Go to products
2. Tap on a product
3. Tap on the `...` menu button
4. Verify the `Trash product` is visible

The above

### Images/gif

| Before | After |
|---|---|
| trash_unavailable | trash_visible |


hichamboushaba (approved):
Looks good, thanks for working on this @AdamGrzybkowski, and sorry for missing it too in my review of the previous PR 😅

PR #14735, opened 2025-10-10
[WOOMOB-1463] Booking filter persistence

AdamGrzybkowski:
WOOMOB-1463

### Description

This PR adds a persistence layer for booking filters. The persistence is not fully implemented and should evolve as we add more filter options.

Booking filters are persisted in a DataStore and exposed via Flow to react to any changes.

The PR is quite big mostly because I wanted to show you a fully working flow: Booking list -> Filter screen -> Booking list. Whenever the filters are updated on the filter screen the new list should be loaded.

### Steps to reproduce

Filters are empty for now so it's not technically possible. Please apply this patch [filter_test.patch](https://github.com/user-attachments/files/22847137/filter_test.patch) and test with it. 

Note - this will only work once as the date range filter is hardcoded. The above patch adds some logs to verify the flow. You can filter the logcat by `WooCommerce-Bookings`

1. Open the booking list
2. Go to the All tab
3. You should see a full list 
4. Tap on Filters
5. Tap on Show bookings
6. Verify the list is updated and you see the same results as on the Today tab
7. Go to the filters 
8. Tap on Show booking
9. Verify via logs or monitoring network that no request was made to fetch the bookings

The above.

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -187,10 +192,15 @@ class BookingListViewModel @Inject constructor(
...
-            selectedTab.value.asDateRangeFilter()
-        )
+    private suspend fun prepareFilters(): List<BookingsFilterOption> = with(filtersBuilder) {
+        when (selectedTab.value) {
+            BookingListTab.Today,
+            BookingListTab.Upcoming -> listOfNotNull(
+                selectedTab.value.asDateRangeFilter()
+            )
+
+            BookingListTab.All -> bookingFilterRepository.bookingFilterFlow.first().asList()
```
AdamGrzybkowski: I'm not super happy with that. I thought about having a StateFlow to hold the value by doing `bookingFilterRepository.bookingFilterFlow.toStateFlow(BookingFilter())` but this requires to set the initial state by in the VM and we have to drop 2 initial values.

Do you have a better idea how to collect the filter object?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
```diff
@@ -22,3 +22,14 @@ sealed interface BookingsFilterOption {
 
     object ServiceEvent : BookingsFilterOption
 }
+
+data class BookingFilter(
```
AdamGrzybkowski: I've decided to wrap all the possible filters in one class, to make it clearer what filters are available. We could remove it and save/expose a list of filters from the repository, but I found it to be confusing, especially when saving the filters without a clear way to remove the previously set filters.  WDYT?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
```diff
@@ -11,7 +11,7 @@ sealed interface BookingsFilterOption {
 
     object BookingType : BookingsFilterOption
 
-    data class Customer(val customerId: Long?) : BookingsFilterOption
+    data class Customer(val customerId: Long?, val customerName: String?) : BookingsFilterOption
```
AdamGrzybkowski, line 14: 1. `customerName` was added as it will be used instead of `Any` when selected. Otherwise, we would have to fetch the customer name separately.
Screenshot 2025-10-10 at 12 04 37

2. I don't like that we have those values as nullable. If a filter is created, values that can't be null shouldn't be. A `Customer` filter with customerId is not a valid filter. On the other hand, passing `null` to the API is probably fine, so maybe it's okay. WDYT?

hichamboushaba: > customerName was added as it will be used instead of Any when selected. Otherwise, we would have to fetch the customer name separately.

I see, this is an acceptable solution, but the best solution would be to get the customer name from the DB or fetch it, because the customer name could have been modifier after the filter was set. But I think this is an edge case, and the simplicity of this outweighs the benefits of the other solution.

> I don't like that we have those values as nullable. If a filter is created, values that can't be null shouldn't be. A Customer filter with customerId is not a valid filter. On the other hand, passing null to the API is probably fine, so maybe it's okay. WDYT?

TBH I don't know why we have these nullables, when I added the initial class as example I had it non-nullable, because my plan was to have the `Customer` filter added to the list only when it's set, when we choose `Any`, we just remove it from the list.
I see you are already handling the removal in the repository [here](https://github.com/woocommerce/woocommerce-android/blob/b03cf6487499766f056f7593faaddda35d9f211a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/data/BookingFilterRepository.kt#L50-L59), so I think we can make both fields non-nullable, or did I miss something here?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
hichamboushaba: I like, just a nitpick here, maybe we can make the name plural to make it clear this combines multiple filters, so `BookingFilters`.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
AdamGrzybkowski: I think it's because of [this](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListViewModel.kt#L50-L59). I'm currently working on the filter navigation, and if you don't mind, I can tackle it there as I have some classes that make it more straightforward. 

>I see, this is an acceptable solution, but the best solution would be to get the customer name from the DB or fetch it, because the customer name could have been modifier after the filter was set. But I think this is an edge case, and the simplicity of this outweighs the benefits of the other solution.

We could always extend the current approach with a data refresh.

hichamboushaba: > I think it's because of [this](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListViewModel.kt#L50-L59). I'm currently working on the filter navigation, and if you don't mind, I can tackle it there as I have some classes that make it more straightforward.

I see, thanks, I think the default should be an empty list if we agree on the above, just thinking aloud here.
And anyway, no worries about tackling it in a separate PR.

hichamboushaba: >  I think the default should be an empty list if we agree on the above

Forget about this, I see we are using the `BookingFilterOption` on the UI too, so an empty list won't work I think.
I'll leave it to you to handle this in the separate PR, it's a minor point anyway.

AdamGrzybkowski: >I see, this is an acceptable solution, but the best solution would be to get the customer name from the DB

I was not aware of the `CustomerDao`, but looking at the code I'm not sure we can rely on that. The whole table is cleared on the initial page load 🤔

hichamboushaba: You would use the `CustomerStore` instead of the `Dao` directly, but the issue you mentioned still applies.

For Orders, we have this [logic](https://github.com/woocommerce/woocommerce-android/blob/f8590dafacec7a6df7f91e9d5678c874fd24243c/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt#L368-L379) to get the customer name from the Store, and if not available, it will use `Id: <id>`.

I think what you have now would work better in most cases, the possibility of changing the customer name is smaller than clearing the local cache.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
hichamboushaba: I think this is a result of some mistakes I did with the screen when I started working on it, the logic is not fully reactive, to be fully reactive, we should avoid reading the values from the `StateFlow#value`, and instead just react to them like any other `Flow`, doing this will allow us to observe the `DataStore` without requiring to make it `StateFlow`.

<details>
<summary>I tried to experiment with this in this patch:</summary>

```patch
IDEA additional info:
<+>UTF-8
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt	(revision ea1eea127d84d4eeae954b74352063423983fb63)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt	(date 1760107393626)
@@ -14,15 +14,15 @@
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
-import kotlinx.coroutines.flow.drop
-import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.flow.withIndex
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingFilter
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingsFilterOption
 import javax.inject.Inject
 
@@ -42,6 +42,7 @@
         clazz = String::class.java,
         key = "searchQuery"
     )
+    private val refreshTrigger = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
 
     private val sortOption = savedStateHandle.getStateFlow(viewModelScope, BookingListSortOption.NewestToOldest)
 
@@ -59,7 +60,7 @@
         BookingListContentState(
             bookings = bookings,
             loadingState = loadingState,
-            onRefresh = { fetchBookings(BookingListLoadingState.Refreshing) },
+            onRefresh = { refreshTrigger.tryEmit(Unit) },
             onLoadMore = ::loadMore,
             onBookingClick = ::onBookingClick
         )
@@ -115,37 +116,66 @@
     @OptIn(FlowPreview::class)
     private fun monitorSearchAndFilterChanges() {
         launch {
+            var lastFetchParams: FetchParams? = null
             val queryFlow = searchQuery
-                .drop(1) // Skip the initial value to avoid double fetch on init
-                .debounce {
-                    if (it.isNullOrEmpty()) 0L else AppConstants.SEARCH_TYPING_DELAY_MS
+                .withIndex()
+                .debounce { (index, query) ->
+                    // Skip debounce for the initial value or when the query is cleared
+                    if (index == 0 || query.isNullOrEmpty()) 0L else AppConstants.SEARCH_TYPING_DELAY_MS
                 }
-            val sortFlow = sortOption.drop(1) // Skip the initial value to avoid double fetch on init
-            val bookingFilterFlow =
-                bookingFilterRepository.bookingFilterFlow.drop(1) // Skip initial to avoid double fetch on init
+                .map { it.value }
 
-            merge(selectedTab, queryFlow, sortFlow, bookingFilterFlow).collectLatest {
+            combine(
+                selectedTab,
+                queryFlow,
+                sortOption,
+                bookingFilterRepository.bookingFilterFlow
+            ) { tab, query, sort, filters ->
+                FetchParams(
+                    searchQuery = query,
+                    sortOption = sort,
+                    selectedTab = tab,
+                    filters = filters
+                )
+            }.collectLatest { fetchParams ->
                 // Cancel any ongoing fetch or load more operations
                 bookingsFetchJob?.cancel()
                 bookingsLoadMoreJob?.cancel()
 
-                bookingsFetchJob = fetchBookings(
-                    initialLoadingState = if (it is BookingListSortOption) {
+                val initialLoadingState = lastFetchParams.let { lastFetchParams ->
+                    if (lastFetchParams != null && lastFetchParams.sortOption != fetchParams.sortOption) {
                         BookingListLoadingState.Refreshing
                     } else {
                         BookingListLoadingState.Loading
                     }
+                }
+
+                lastFetchParams = fetchParams
+                fetchBookings(
+                    initialLoadingState = initialLoadingState,
+                    fetchParams = fetchParams
                 )
+
+                refreshTrigger.collect {
+                    fetchBookings(
+                        initialLoadingState = BookingListLoadingState.Refreshing,
+                        fetchParams = fetchParams
+                    )
+                }
             }
         }
     }
 
-    private fun fetchBookings(initialLoadingState: BookingListLoadingState) = launch {
+    private suspend fun fetchBookings(
+        initialLoadingState: BookingListLoadingState,
+        fetchParams: FetchParams,
+    ) {
+        println("Fetching bookings with params: $fetchParams")
         bookingListHandler.loadBookings(
-            searchQuery = searchQuery.value,
-            filters = prepareFilters(),
-            sortBy = sortOption.value
+            searchQuery = fetchParams.searchQuery,
+            filters = fetchParams.prepareFilters(),
+            sortBy = fetchParams.sortOption
         ).onFailure {
             triggerEvent(MultiLiveEvent.Event.ShowSnackbar(R.string.bookings_fetch_error))
         }
@@ -192,16 +222,23 @@
         triggerEvent(NavigateToFilters)
     }
 
-    private suspend fun prepareFilters(): List<BookingsFilterOption> = with(filtersBuilder) {
-        when (selectedTab.value) {
+    private fun FetchParams.prepareFilters(): List<BookingsFilterOption> = with(filtersBuilder) {
+        when (selectedTab) {
             BookingListTab.Today,
             BookingListTab.Upcoming -> listOfNotNull(
-                selectedTab.value.asDateRangeFilter()
+                selectedTab.asDateRangeFilter()
             )
 
-            BookingListTab.All -> bookingFilterRepository.bookingFilterFlow.first().asList()
+            BookingListTab.All -> filters.asList()
         }
     }
+
+    private data class FetchParams(
+        val searchQuery: String?,
+        val sortOption: BookingListSortOption,
+        val selectedTab: BookingListTab,
+        val filters: BookingFilter
+    )
 
     data class NavigateToBookingDetails(val bookingId: Long) : MultiLiveEvent.Event()
     object NavigateToFilters : MultiLiveEvent.Event()
```

</details>

I think this is a big change, so probably it needs to go to its own PR, but just wanted to get your opinion on this new approach before investing time in finalizing it.

hichamboushaba (approved):
Nice work @AdamGrzybkowski, I think the pending discussions will be handled in separate PRs, so this is ready to 🚢

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
AdamGrzybkowski: I like it 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
AdamGrzybkowski: Done in https://github.com/woocommerce/woocommerce-android/pull/14735/commits/5df1b51e8a4bf02643c76313e182170a97d66802

PR #14736, opened 2025-10-10
Issue/woomob 1483 woo poslocal catalog remove variations on full sync

malinajirka:
Fixes [WOOMOB-1483](https://linear.app/a8c/issue/WOOMOB-1483)

Remove 'Do not merge' label when the target branch is `trunk`.

### Description
This PR implements functionality to remove all variations before inserting them during a full sync operation. I considered removing only rows that have been deleted in the remote, but the operation to find the ones that were removed is too expensive.

[Note: There is a known bug in the remote so sometimes variations are missing even when they shouldn't.](https://linear.app/a8c/issue/WOOA7S-564/broken-pagination-on-wc-analyticsvariations-endpoint)

### Testing information
Verify that full sync operations properly remove variations that no longer exist in the remote

Unit test + the above


kidinov (approved):
LGTM

PR #14741, opened 2025-10-13
[WOOMOB-1487] - Booking filter internal navigation

AdamGrzybkowski:
WOOMOB-1487

### Description

This PR implements the internal navigation in the BookingFilter screen. It handled going into a specific filer view (for example, date picker) and updates the Toolbar (title and the icon). For now, the content is always the same (Empty date&time picker composable).

Additionally, in the last commit (https://github.com/woocommerce/woocommerce-android/commit/f832e2cf3b7189d79c4d66974d754ac8639d353d), I updated the `BookingFilterOption.Customer` by making the fields non-nullable. This requried some updates to how we built the UI state for the booking filter screen.

### Steps to reproduce

1. Launch the app 
2. Go to the booking tab
3. Tap `All`
4. Tap `Filters`
5. Select any of the options
6. Confirm the toolbar is updated and the content switched
7. Tap back or use the system back button
8. Confirm the list of available filters is visible

The above

### Images/gif

https://github.com/user-attachments/assets/cd47c960-0e5e-43f1-8f62-3445e1b44301

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
@@ -11,9 +11,9 @@ sealed interface BookingsFilterOption {
 
     object BookingType : BookingsFilterOption
 
-    data class Customer(val customerId: Long?, val customerName: String?) : BookingsFilterOption
+    data class Customer(val customerId: Long, val customerName: String) : BookingsFilterOption
 
-    object Category : BookingsFilterOption
+    object Location : BookingsFilterOption
 
     data class DateRange(
         val before: Instant?,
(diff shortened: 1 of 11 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsFilterOption.kt
```diff
@@ -11,9 +11,9 @@ sealed interface BookingsFilterOption {
 
     object BookingType : BookingsFilterOption
 
-    data class Customer(val customerId: Long?, val customerName: String?) : BookingsFilterOption
+    data class Customer(val customerId: Long, val customerName: String) : BookingsFilterOption
 
-    object Category : BookingsFilterOption
+    object Location : BookingsFilterOption
```
AdamGrzybkowski, line 16: FYI: I don't see a `Category` option in Figma, but we were missing Location so I updated the code.

hichamboushaba: The API doesn't currently support `Location` filtering, but I'm OK with keeping it until we know more though.

hichamboushaba (approved):
Nice work @AdamGrzybkowski 🎖️

PR #14743, opened 2025-10-13
Issue/woomob 1499 woo poslocal catalog perform incremental sync after each

malinajirka:
#WOOMOB-1499

### Description

This PR adds an incremental local catalog sync after each full local catalog sync. This ensure that even when the full sync downloads a file that was generated several hours ago, the local catalog remains up to date.

Failure of the incremental sync isn't so vital, since the POS runs incremental sync
- on startup
- on completed order
- on PTR
- every 60 minutes
So failed incremental update is logged but doesn't result in a failure of the whole worker.

### Testing information

1. Clear app data
2. Log in
3. Wait and observe logs
4. Notice both full sync and incremental syncs are performed
5. Open POS and notice data are available

the above
### Images/gif

N/A


kidinov (approved):
LGTM!

PR #14740, opened 2025-10-13
Fix broken skeleton layout

malinajirka:
#WOOMOB-1424

### Description

This PR fixes a minor UI glitch on the skeleton loading view in analytics.

### Steps to reproduce

1. Open dashboard
2. Tap on 'View all store analytics on one of the cards
3. Scroll down and notice a broken (overlapping) skeleton

### Testing information

- Repeat the above steps and verify it's fixed

### The tests that have been performed

The above

### Images/gif

| Before | After |
| -- | -- |
| image | image |

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #14745, opened 2025-10-13
Replace margin right with margin end

malinajirka:
This is a quick follow up for this merged PR - https://github.com/woocommerce/woocommerce-android/pull/14740.

I accidentally didn't push the last commit which replaces `marginRight` with `marginEnd` to ensure RTL support.


kidinov (approved):
Thanks for noticing!

PR #14746, opened 2025-10-13
[CIAB] Implement mapping of payment details for Bookings and improve address formatting

hichamboushaba:
Closes WOOMOB-1503

### Description
This PR adds two main changes:
- Implements the mapping from the `BookingPaymentInfo` to the UI model `BookingPaymentDetailsModel`. For this change I made the property `bookingPaymentDetails` nullable in `BookingUiState`, because if we don't have an order, it will be null, a case that we handle in the DB even though it shouldn't happen in production.
- Improve address formatting to match what we use in other areas in the app, to do so I choose to use the existing `Address` model as an intermediary for the formatting.

I'll leave some comments regarding some decisions below.

### Steps to reproduce
1. Enable taxes calculation for your store.
2. Create a booking with an order.
3. Open the order in wp-admin, and apply a discount (either using a coupon, or by manually modifying the price of the booking in the details screen)
4. Open the booking in the app.

### Testing information
- Confirm the address formatting matches what we use in order details.
- Confirm the payment details are shown correctly.

The above.


WooCommerce/src/main/kotlin/com/woocommerce/android/model/Address.kt
```diff
@@ -63,12 +63,13 @@ data class Address(
         } ?: this.orderAddressToString()
     }
 
-    fun getFullAddress(name: String, address: String, country: String): String {
```
hichamboushaba, line 66: I opted to remove this function, I don't know the history about it, but the fact that the function accepted the details made it look like a utility function that don't belong here in the model.
So instead, I updated the function not to require any arguments, which simplified its usage everywhere. For better context about the change, check the commit f383c3207562c12c2c619ddd0594fdb85761dc20 individually.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingCustomerDetails.kt
```diff
@@ -148,14 +148,14 @@ data class BookingCustomerDetailsModel(
     val name: String?,
     val email: String?,
     val phone: String?,
-    val billingAddressLines: List<String>,
+    val billingAddress: String?,
```
hichamboushaba, line 151: I updated this to be a single `String` over a list, I did this because the existing API for Address formatting returns a single String, so instead of splitting it, we show it in a single `Text`. If you see any downsides with this change, please let me know.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -87,4 +105,24 @@ class BookingMapper @Inject constructor(
     private fun BookingCustomerInfo.fullName(): String? {
         return "${billingFirstName.orEmpty()} ${billingLastName.orEmpty()}".trim().ifEmpty { null }
     }
+
+    private suspend fun BookingCustomerInfo.address(): Address? {
+        val countryCode = billingCountry ?: return null
+        val (country, state) = withContext(Dispatchers.IO) {
+            getLocations(countryCode, billingState.orEmpty())
+        }
```
hichamboushaba, line 113: As we discussed before, this follows what we use in `OrderMapper`. If we want to avoid doing this in the mapper (personally don't see an issue with having this in the mapper), we'll have to do something like the following patch to move the `getLocations` call to the consumer, let me know if you prefer the other approach.

```patch
Subject: [PATCH] Changes
---
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt	(revision 6fb8c209cfc832aeb9f9445bff7415af171deafb)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt	(date 1760374703373)
@@ -2,7 +2,9 @@
 
 import com.woocommerce.android.extensions.isNotEqualTo
 import com.woocommerce.android.model.Address
+import com.woocommerce.android.model.AmbiguousLocation
 import com.woocommerce.android.model.GetLocations
+import com.woocommerce.android.model.Location
 import com.woocommerce.android.ui.bookings.compose.BookingAppointmentDetailsModel
 import com.woocommerce.android.ui.bookings.compose.BookingAttendanceStatus
 import com.woocommerce.android.ui.bookings.compose.BookingCustomerDetailsModel
@@ -11,8 +13,6 @@
 import com.woocommerce.android.ui.bookings.compose.BookingSummaryModel
 import com.woocommerce.android.ui.bookings.list.BookingListItem
 import com.woocommerce.android.util.CurrencyFormatter
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingCustomerInfo
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingPaymentInfo
 import org.wordpress.android.fluxc.persistence.entity.BookingEntity
@@ -67,14 +67,17 @@
         )
     }
 
-    suspend fun BookingCustomerInfo?.toCustomerDetailsModel(): BookingCustomerDetailsModel {
+    fun BookingCustomerInfo?.toCustomerDetailsModel(
+        state: AmbiguousLocation?,
+        country: Location?
+    ): BookingCustomerDetailsModel {
         if (this == null) return BookingCustomerDetailsModel.EMPTY
 
         return BookingCustomerDetailsModel(
             name = fullName(),
             email = billingEmail,
             phone = billingPhone,
-            billingAddress = address()?.getFullAddress()
+            billingAddress = address(state, country)?.getFullAddress()
         )
     }
 
@@ -106,18 +109,19 @@
         return "${billingFirstName.orEmpty()} ${billingLastName.orEmpty()}".trim().ifEmpty { null }
     }
 
-    private suspend fun BookingCustomerInfo.address(): Address? {
-        val countryCode = billingCountry ?: return null
-        val (country, state) = withContext(Dispatchers.IO) {
-            getLocations(countryCode, billingState.orEmpty())
-        }
+    private fun BookingCustomerInfo.address(
+        state: AmbiguousLocation?,
+        country: Location?
+    ): Address? {
+        if (country == null) return null
+
         return Address(
             company = billingCompany.orEmpty(),
             firstName = billingFirstName.orEmpty(),
             lastName = billingLastName.orEmpty(),
             phone = billingPhone.orEmpty(),
             country = country,
-            state = state,
+            state = state ?: AmbiguousLocation.Defined(Location.EMPTY),
             address1 = billingAddress1.orEmpty(),
             address2 = billingAddress2.orEmpty(),
             city = billingCity.orEmpty(),
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt	(revision 6fb8c209cfc832aeb9f9445bff7415af171deafb)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt	(date 1760374630741)
@@ -4,6 +4,7 @@
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.asLiveData
+import com.woocommerce.android.model.GetLocations
 import com.woocommerce.android.ui.bookings.Booking
 import com.woocommerce.android.ui.bookings.BookingMapper
 import com.woocommerce.android.ui.bookings.BookingsRepository
@@ -23,6 +24,7 @@
     resourceProvider: ResourceProvider,
     bookingsRepository: BookingsRepository,
     private val bookingMapper: BookingMapper,
+    private val getLocations: GetLocations
 
@@ -59,16 +61,22 @@
     private suspend fun BookingMapper.buildBookingUiState(
         booking: Booking,
         attendanceStatus: BookingAttendanceStatus?
-    ): BookingUiState = BookingUiState(
-        bookingSummary = booking.toBookingSummaryModel().let {
-            if (attendanceStatus != null) {
-                it.copy(attendanceStatus = attendanceStatus)
-            } else {
-                it
-            }
-        },
-        bookingsAppointmentDetails = booking.toAppointmentDetailsModel(),
-        bookingCustomerDetails = booking.order.customerInfo.toCustomerDetailsModel(),
-        bookingPaymentDetails = booking.order.paymentInfo?.toPaymentDetailsModel(booking.currency)
-    )
+    ): BookingUiState {
+        val (country, state) = booking.order.customerInfo?.let {
+            getLocations(it.billingCountry.orEmpty(), it.billingState.orEmpty())
+        } ?: Pair(null, null)
+
+        return BookingUiState(
+            bookingSummary = booking.toBookingSummaryModel().let {
+                if (attendanceStatus != null) {
+                    it.copy(attendanceStatus = attendanceStatus)
+                } else {
+                    it
+                }
+            },
+            bookingsAppointmentDetails = booking.toAppointmentDetailsModel(),
+            bookingCustomerDetails = booking.order.customerInfo.toCustomerDetailsModel(state, country),
+            bookingPaymentDetails = booking.order.paymentInfo?.toPaymentDetailsModel(booking.currency)
+        )
+    }
 }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -58,19 +67,28 @@ class BookingMapper @Inject constructor(
...
+        val discount = total - subtotal
+        return BookingPaymentDetailsModel(
+            service = currencyFormatter.formatCurrency(subtotal, currency), // Pre-discount subtotal
+            tax = currencyFormatter.formatCurrency(totalTax, currency), // Tax on total after discount
+            discount = if (discount.isNotEqualTo(BigDecimal.ZERO)) {
+                "- ${currencyFormatter.formatCurrency(discount.abs(), currency)}"
+            } else {
+                "-"
+            },
+            total = currencyFormatter.formatCurrency(total + totalTax, currency) // Total including tax
```
hichamboushaba, line 91: Let me know if you have any questions about the mapping here, I can share details on the role of the fields in the API and how Core handles the calculations.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
AdamGrzybkowski: Thanks for the explanation @hichamboushaba - my preference would be to go with the patch, especially considering the thread switching within the mapper. I won't block the PR, though.

AdamGrzybkowski (approved):
Looks good! The biiling adress matches the Order screen. There's one thing though - there seems to be a missing `\n` between city and country. 

| booking | order | web |
|---|---|---|
| Screenshot 2025-10-14 at 11 13 39 | Screenshot 2025-10-14 at 11 14 11 | Screenshot 2025-10-14 at 11 14 00 |

Technically, this is not related to this PR so ✅ but I imagine this is a simple fix so maybe you could add it?

hichamboushaba:
> There's one thing though - there seems to be a missing \n between city and country.

Good eye, I'm not sure how I missed this. And this is a result of this PR, so it's good that we caught this, I didn't know `appendWithIfNotEmpty` adds the separator first and not after, fixed in the commit 35ce78e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
hichamboushaba: Thanks Adam, after all, I decided to stick with what I have now, the nullability handling when moving `getLocations` call to the ViewModel is not easy to follow, and given we follow the same approach in `OrderMapper`, we stay consistent too.

PR #14750, opened 2025-10-14
[WOOMOB-1453] Booking details - refresh API call

AdamGrzybkowski:
WOOMOB-1453

### Description

This PR implements the refresh logic on the Booking details screen. 
It follows the Booking List logic - so first the booking is retrieved and then the corresponding order.

Additionally, I've added a basic error handling to show a snack when the call fails or when there's no network conneciton.

### Steps to reproduce

1. Monitor network calls via Android Studio or any other tool you like
2. Launch the app
3. Go to bookings tab
4. Tap on the booking
5. Confirm two endpoints were made to fetch the booking and the order

### Testing information

Repeat the above with disabled network. 
Repeat the above with network but this time mock the API responses to fail.

The above.

### Images/gif

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -1,6 +1,5 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings
 
-import org.wordpress.android.fluxc.annotations.endpoint.WCWPAPIEndpoint
 import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse.Error
@@ -24,9 +23,7 @@ class BookingsRestClient @Inject constructor(
         site: SiteModel,
         bookingId: Long
     ): WooPayload<BookingDto> {
-        val endpoint = WCWPAPIEndpoint(
-            WOOCOMMERCE.bookings.id(bookingId).endpoint
-        ).pathV2Bookings
+        val endpoint = WOOCOMMERCE.bookings.id(bookingId).pathV2Bookings
         val response = wooNetwork.executeGetGsonRequest(
             site = site,
             path = endpoint,
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -1,5 +1,6 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings
 
+import org.wordpress.android.fluxc.annotations.endpoint.WCWPAPIEndpoint
 import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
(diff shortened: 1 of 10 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -88,6 +88,36 @@ class BookingsStore @Inject internal constructor(
...
+        bookingId: Long
+    ): WooResult<BookingEntity> {
+        return coroutineEngine.withDefaultContext(AppLog.T.API, this, "fetchBooking") {
+            val response = bookingsRestClient.fetchBooking(site, bookingId)
+            when {
+                response.isError -> WooResult(response.error)
+                response.result != null -> {
+                    val dto = response.result
+                    val orderIds = listOf(dto.orderId).filterNot { it == 0L }
+                    val ordersResult = fetchOrders(site, orderIds)
```
hichamboushaba: This will use `/wc/v3/orders?include=<orderId>`, normally we should use `/wc/v3/orders/<orderId>` here, as there may be some performance benefits in using the single item endpoint.
To do this, we can use the function `WCOrderStore#fetchSingleOrderSync`, this returns the fetched order, and handles the caching too.

hichamboushaba (approved):
Great work @AdamGrzybkowski, I left one comment about the endpoint we use for fetching orders.

And a remark here, I think adding support for pull-to-refresh would be beneficial here, as it would match the experience we have in order details, and also in case of failures, it will allow the user to retry, WDYT? This can be added in a separate PR.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
```diff
@@ -19,6 +20,25 @@ class BookingsRestClient @Inject constructor(
         const val DEFAULT_PER_PAGE = 25 // Number of items to fetch in a single request
     }
 
+    suspend fun fetchBooking(
+        site: SiteModel,
+        bookingId: Long
+    ): WooPayload<BookingDto> {
+        val endpoint = WCWPAPIEndpoint(
+            WOOCOMMERCE.bookings.id(bookingId).endpoint
+        ).pathV2Bookings
```
hichamboushaba: I missed this in my review.

```suggestion
        val endpoint = WOOCOMMERCE.bookings.id(bookingId).pathV2Bookings
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
AdamGrzybkowski: Good call, done in https://github.com/woocommerce/woocommerce-android/pull/14750/commits/5d68d5053d7e953a8c788c6089dc30d31d0895a8

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
AdamGrzybkowski: Done in https://github.com/woocommerce/woocommerce-android/pull/14750/commits/5d68d5053d7e953a8c788c6089dc30d31d0895a8

AdamGrzybkowski:
@hichamboushaba Agree about the PTR - done here https://github.com/woocommerce/woocommerce-android/pull/14750/commits/45ff617c3eb4d57583bcf8fc323cd08728a8e8ab

hichamboushaba (approved):
Thanks @AdamGrzybkowski for all the changes, nice work 👏

PR #14752, opened 2025-10-14
[WOOMOB-1511] - Booking details cancel flow

AdamGrzybkowski:
WOOMOB-1511

### Description

This PR implements the booking flow as per the designs lxixW6aBBNkCd9ooBK70Z1-fi-1251_19683.

What's included:
- Removal of the Cancelled state from the bottom sheet
- Cancel booking confirmation dialog
- A loading indicator when the cancel is in progress (a fixed delay is set for now)

What's not included:
- API call
- Cancel button visibility based on the cancelled state

### Steps to reproduce

1. Launch the app
2. Go to the bookings tab
3. Open any booking
4. Tap the `Cancel booking` button

### Testing information

Test Dark/Ligth theme.

The above

### Images/gif

https://github.com/user-attachments/assets/2f39d55e-5e2a-4d4f-817e-3e7813ba19d8


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusBottomSheet.kt
```diff
@@ -118,16 +124,16 @@ private fun AttendanceStatusRow(
 private fun BookingAttendanceStatus.description(): String = when (this) {
     BookingAttendanceStatus.BOOKED -> R.string.booking_attendance_status_booked_desc
     BookingAttendanceStatus.CHECKED_IN -> R.string.booking_attendance_status_checked_in_desc
-    BookingAttendanceStatus.CANCELLED -> R.string.booking_attendance_status_cancelled_desc
     BookingAttendanceStatus.NO_SHOW -> R.string.booking_attendance_status_no_show_desc
-}.let { stringResource(it) }
+    else -> null
+}.let { it?.let { id -> stringResource(id) } ?: "" }
```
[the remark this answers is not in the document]
AdamGrzybkowski: Not sure what's wrong with AS but it complains when I do that 😕 It's wrong though, the code works just fine.

Screenshot 2025-10-14 at 17 07 33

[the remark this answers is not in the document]
AdamGrzybkowski: Tested the code, AS is wrong here so updated the code.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
```diff
@@ -62,9 +66,17 @@ fun BookingAppointmentDetails(
...
             ) {
-                Text(text = stringResource(R.string.booking_details_cancel_booking_button))
+                if (model.cancelInProgressShown) {
+                    CircularProgressIndicator(
+                        color = LocalContentColor.current,
+                        modifier = Modifier.size(24.dp)
+                    )
+                } else {
+                    Text(text = stringResource(R.string.booking_details_cancel_booking_button))
+                }
```
hichamboushaba, line 104: minor remark, WDYT about updating `WCOutlinedButton` to accept a `loading` parameter like what we do with `WCColoredButton` [here](https://github.com/woocommerce/woocommerce-android/blob/c5901b7986bc285e6400f51b96a1f143c714a2e8/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt#L105), and move this logic to the component.

hichamboushaba: If you do, please also make sure to update `WCColoredButton`'s implementation to pass `enabled = enabled && !loading`, to make sure we disable button during loading.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
```diff
@@ -18,6 +18,10 @@ data class BookingDetailsViewState(
     val loadingState: BookingDetailsLoadingState = BookingDetailsLoadingState.Idle,
     val onCancelBooking: () -> Unit = {},
     val onAttendanceStatusSelected: (BookingAttendanceStatus) -> Unit = { _ -> },
+    val showCancelBookingDialog: Boolean = false,
+    val cancelDialogMessage: String = "",
+    val onDismissCancelDialog: () -> Unit = {},
+    val onConfirmCancelBooking: () -> Unit = {},
```
hichamboushaba: What do you think about using the existing common [`DialogState`](https://github.com/woocommerce/woocommerce-android/blob/c5901b7986bc285e6400f51b96a1f143c714a2e8/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/DialogState.kt) here, I mean something like this:

```kotlin
    val bookingUiState: BookingUiState? = null,
    val loadingState: BookingDetailsLoadingState = BookingDetailsLoadingState.Idle,
    val onCancelBooking: () -> Unit = {},
    val onAttendanceStatusSelected: (BookingAttendanceStatus) -> Unit = { _ -> },
    val dialogState: DialogState? = null,
    val onRefresh: () -> Unit = {},
) {
    val shouldShowSkeleton: Boolean = bookingUiState == null && loadingState == BookingDetailsLoadingState.Refreshing
}
```

Then, on the screen, we would use something like the following:

```kotlin
viewState.dialogState?.Render()
```

And we can then remove completely `CancelBookingDialog`.

One additional advantage of the above is that `dialogState` will be generic enough so that it can be used for any flow that requires a modal dialog in the details screen.

Also, `DialogState` accepts `UiString`, so you won't have to use `ResourceProvider`:

```kotlin
    fun buildCancelDialogMessage(booking: Booking, resourceProvider: ResourceProvider): UiString {
        val customerName = booking.order.customerInfo?.fullName()
            ?: resourceProvider.getString(R.string.customer_detail_guest_customer)
        val serviceName = booking.order.productInfo?.name ?: "-"
        val date = detailsDateFormatter.format(booking.start)
        val time = timeRangeFormatter.format(booking.start)
        return UiString.UiStringRes(
            R.string.booking_cancel_dialog_message,
            listOf(
                UiString.UiStringText(customerName),
                UiString.UiStringText(serviceName),
                UiString.UiStringText(date),
                UiString.UiStringText(time)
            )
        )
    }
```

BTW, `ResourceProvider` doesn't play well with configuration changes, as it uses the app context behind the scenes.

hichamboushaba: I just noticed the buttons use custom colors in the dialog, so this changes things a bit, but personally I would still vote to use `DialogState`, we can add a parameter to `DialogButton` to set the text color, or even a whole `ButtonColors`, or we can update the `Render` function to allow customizing the button colors if we want to keep the color logic in the UI.

Also, if we want to match the other dialogs in the app, the dismiss color should use the primary color and not `onSurface`.

WDYT?

hichamboushaba (approved):
Great work @AdamGrzybkowski 👏 

I left one remark about potentially using `DialogState` for managing the dialog state instead of the custom one we have, please check it out. But I won't block the PR because of it, so I'm pre-approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
AdamGrzybkowski: If you don't mind, I would prefer to open a separate PR with it just to isolate the change that goes beyond Booking Details screen.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
AdamGrzybkowski: That's a great idea, I was not aware of the `DialogState`. The migration is here https://github.com/woocommerce/woocommerce-android/pull/14752/commits/d0c3f31f3ec92339187f6a81cc1424a3e308d334

I kept the standard colors for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -54,25 +58,62 @@ class BookingDetailsViewModel @Inject constructor(
...
+        } else {
+            null
+        }
+    }
+
+    val state: LiveData<BookingDetailsViewState> = combine(
+        booking,
+        bookingUiStateFlow,
+        loadingState,
+        cancelBookingDialogState,
```
AdamGrzybkowski, line 105: This was becoming quite big, so I extracted some of those to separate Flows (booking Ui state, and cancel booking dialog state)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
AdamGrzybkowski: PR is here https://github.com/woocommerce/woocommerce-android/pull/14768

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -54,25 +58,62 @@ class BookingDetailsViewModel @Inject constructor(
     // Temporary, the booking status should come from the stored object
     private val bookingAttendanceStatus = MutableStateFlow<BookingAttendanceStatus?>(null)
 
-    val state: LiveData<BookingDetailsViewState> = combine(
+    private val cancelStatusState = MutableStateFlow<CancelStatus>(CancelStatus.Idle)
+    private val showCancelBookingDialog = MutableStateFlow(false)
+
+    val cancelBookingDialogState = combine(
```
hichamboushaba: We can make this and `bookingUiStateFlow` private.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
```diff
@@ -18,6 +19,7 @@ data class BookingDetailsViewState(
     val loadingState: BookingDetailsLoadingState = BookingDetailsLoadingState.Idle,
     val onCancelBooking: () -> Unit = {},
     val onAttendanceStatusSelected: (BookingAttendanceStatus) -> Unit = { _ -> },
+    val cancelBookingDialogState: DialogState? = null,
```
hichamboushaba: Just sharing a thought here, no need to change it now if you prefer to keep it for later: I think using a generic name like `dialogState` here would be more future proof, as if we need any other AlertDialog in the screen, we can then keep the same state, and also we'll ensure we are not trying to show multiple dialogs at the same time.

hichamboushaba (approved):
Thanks @AdamGrzybkowski for the changes, nice work 💯

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewState.kt
AdamGrzybkowski: I like it! Code updated.

PR #14753, opened 2025-10-14
[CIAB] Fetch staff member name in booking details

hichamboushaba:
Closes WOOMOB-1454

### Description
This PR adds logic to fetch the staff member's name on the booking details screen, the logic is as follows:
- If the resource was already cached, we show it right away, otherwise we show a shimmer effect.
- We always fetch the resource on launch, just like the booking itself.
- If the fetching fails, and the resource wasn't cached, then we show `"-"` in the staff member name.
- If the booking didn't specify any resources (`resource_id : 0`), then we hide the row completely.

> [!NOTE]
> The commit a27d1aaee9d4e131158ea72084b3e0a46dbd871a is not related to the PR changes, but given this PR requires a DB migration, I wanted to handle other DB changes in the same PR, I'll share more details in the comment section. 
 
### Steps to reproduce
1. Use a CIAB site.
2. Create a bookable product that has resources (the Hair Salon template uses resources by default)
3. Create a booking.
4. Open the booking in the app.

### Testing information
- Check that the resource is fetched when the booking details screen is opened.
- Remove the cached data, then disable internet, and open the booking details screen again, and confirm that the staff member row shows `"-"`.

The above.

### Images/gif
https://github.com/user-attachments/assets/f73cae40-faed-453a-99a9-ba178e037384

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
@@ -7,6 +7,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingsFilter
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingsOrderOption
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingsStore
 import org.wordpress.android.fluxc.persistence.entity.BookingEntity
+import org.wordpress.android.fluxc.persistence.entity.BookingResourceEntity
 import javax.inject.Inject
 
 class BookingsRepository @Inject constructor(
@@ -74,6 +75,25 @@ class BookingsRepository @Inject constructor(
         }
     }
 
+    suspend fun fetchResource(
+        resourceId: Long
+    ): Result<Unit> {
+        val result = bookingsStore.fetchResource(
+            site = selectedSite.get(),
+            resourceId = resourceId
+        )
+        return if (result.isError) {
+            Result.failure(WooException(result.error))
+        } else {
+            Result.success(Unit)
+        }
+    }
+
+    fun observeResource(resourceId: Long): Flow<BookingResource?> = bookingsStore.observeResource(
+        site = selectedSite.get(),
+        resourceId = resourceId
+    )
+
     data class FetchResult(
         val bookings: List<Booking>,
         val hasMorePages: Boolean
@@ -81,4 +101,4 @@ class BookingsRepository @Inject constructor(
 }
 
 typealias Booking = BookingEntity
-typealias BookingStatus = BookingEntity.Status
+typealias BookingResource = BookingResourceEntity
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
@@ -90,13 +90,13 @@ class BookingsRepository @Inject constructor(
         }
     }
 
-    fun observeResource(booking: Booking): Flow<BookingResource?> {
-        return if (booking.resourceId == 0L) {
+    fun observeResource(resourceId: Long): Flow<BookingResource?> {
+        return if (resourceId == 0L) {
             flowOf(null)
         } else {
             bookingsStore.observeResource(
                 site = selectedSite.get(),
-                resourceId = booking.resourceId
+                resourceId = resourceId
             )
         }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
@@ -3,6 +3,7 @@ package com.woocommerce.android.ui.bookings
 import com.woocommerce.android.WooException
 import com.woocommerce.android.tools.SelectedSite
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.bookings.BookingsFilterOption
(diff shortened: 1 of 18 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingOrderInfo.kt
```diff
@@ -12,24 +11,26 @@ data class BookingOrderInfo(
...
+    val billingState: String? = null,
+    val billingPostcode: String? = null,
+    val billingCountry: String? = null,
+    val billingEmail: String? = null,
+    val billingPhone: String? = null,
 )
 
 data class BookingPaymentInfo(
+    val paymentMethodId: String?,
+    val paymentMethodTitle: String?,
```
hichamboushaba, line 33: The payment method information will be used to decide whether we display the "Pay at location" badge, so we need these two fields.

Side note: When working on this, I'm starting to reconsider my decision to use `@Embedded` for these additional columns, I think that if we saved the various items `BookingProductInfo`, `BookingCustomerInfo` and `BookingPaymentInfo` each in a single Text column, and we stored their json then it would have been easier, we won't then have to deal with migrations for each change, especially during this development time where we don't know all the fields that we'll need.

This is just a thought, I know that the DB migrations should be safer here, but just wanted to share this with you, WDYT? Should we switch to storing the items as json?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -60,17 +80,30 @@ class BookingDetailsViewModel @Inject constructor(
...
-                    triggerEvent(MultiLiveEvent.Event.ShowSnackbar(R.string.bookings_fetch_error))
-                }
+
+            loadingState.value = initialLoadingState
+
+            val bookingTask = async {
+                bookingsRepository.fetchBooking(navArgs.bookingId)
+            }
+            val resourceTask = async {
+                val booking = booking.first() ?: bookingTask.await().getOrNull()
```
hichamboushaba, line 100: In case the booking wasn't cached yet (doesn't happen now, but it could happen if in the future, for example if we support deeplinking to this screen), we wait for the booking fetch to finish before fetching the resource.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingOrderInfo.kt
AdamGrzybkowski: Given that we already have this and have most of the information, I would keep it. 

BTW. What else will we have to add to the DB before alpha? A booking note and the attendance status?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
```diff
@@ -62,23 +64,48 @@ class BookingsRepository @Inject constructor(
...
+            resourceId = resourceId
+        )
         return if (result.isError) {
             Result.failure(WooException(result.error))
         } else {
             Result.success(Unit)
         }
     }
 
+    fun observeResource(booking: Booking): Flow<BookingResource?> {
```
AdamGrzybkowski: ⛏️ Do we need the whole `Booking` as a param here?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/BookingsDao.kt
```diff
@@ -97,4 +98,13 @@ interface BookingsDao {
             order = order
         )
     }
+
+    @Query("SELECT * FROM BookingResources WHERE localSiteId = :localSiteId AND id = :resourceId")
+    suspend fun getResource(localSiteId: LocalId, resourceId: Long): BookingResourceEntity?
```
AdamGrzybkowski: ⛏️ This is unused at the moment. Are you planning to use it in the future?

AdamGrzybkowski (commented):
Great work @hichamboushaba :shipit:

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingOrderInfo.kt
hichamboushaba: > What else will we have to add to the DB before alpha? A booking note and the attendance status?

Yes, that's what we know about for now.

> Given that we already have this and have most of the information, I would keep it.

Sounds good.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
hichamboushaba: Good point, this was just a left over from the early explorations of the logic, updated it in e3b40ef

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/BookingsDao.kt
hichamboushaba: Removed in 7185888

hichamboushaba:
Thanks @AdamGrzybkowski for the review, I've addressed all the comments.

AdamGrzybkowski (approved):
Sorry, I didn't check the `Approve` before.

PR #14760, opened 2025-10-15
[WOOMOB-1515] - Format booking duration following web's logic

AdamGrzybkowski:
WOOMOB-1515

### Description

This PR updates the booking duration formatting to follow the same logic as the web. 

### Steps to reproduce

1. Make sure you have bookings with different durations
2. To test the normalization logic, you can create a room rental booking that lasts full day
3. Launch the booking tab and open different bookings
4. Verify the duration

The above.

### Images/gif

Without the normalization, this would show `1 day 23 hours 59 minutes`
Screenshot_20251015_151451

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
@@ -15,6 +15,8 @@ import com.woocommerce.android.ui.bookings.compose.BookingSummaryModel
 import com.woocommerce.android.ui.bookings.details.CancelStatus
 import com.woocommerce.android.ui.bookings.list.BookingListItem
 import com.woocommerce.android.util.CurrencyFormatter
+import com.woocommerce.android.util.normalizeDuration
+import com.woocommerce.android.util.toHumanReadableFormat
 import com.woocommerce.android.viewmodel.ResourceProvider
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
@@ -65,8 +67,8 @@ class BookingMapper @Inject constructor(
         cancelStatus: CancelStatus,
     ): BookingAppointmentDetailsModel {
         val duration = Duration.between(start, end)
-            .normalizeBookingDuration()
-            .toHumanReadableFormat()
+            .normalizeDuration()
+            .toHumanReadableFormat(resourceProvider)
         return BookingAppointmentDetailsModel(
             date = detailsDateFormatter.format(start),
             time = "${timeRangeFormatter.format(start)} - ${timeRangeFormatter.format(end)}",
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -109,6 +114,110 @@ class BookingMapper @Inject constructor(
...
+                        quantity = hours,
+                        default = R.string.booking_duration_hours,
+                        one = R.string.booking_duration_hour
+                    )
+                } else {
+                    val hoursPart = resourceProvider.getQuantityString(
+                        quantity = hours,
+                        default = R.string.booking_duration_hours,
+                        one = R.string.booking_duration_hour
+                    )
```
AdamGrzybkowski: I'm not sure this is enough for other languages that we support, but from what I've checked, this is how we do plurals here.

For example, that wouldn't work for Polish, but we don't support Polish, so maybe that's fine.

hichamboushaba: Yes, our plurals support is quite limited, a limitation of Glotpress, this won't work well for Arabic too, but I don't think we have much to do here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -131,6 +135,110 @@ class BookingMapper @Inject constructor(
...
+        val hourInSeconds = Duration.ofHours(1).toSeconds()
+        val minuteInSeconds = Duration.ofMinutes(1).toSeconds()
+
+        return when {
+            totalSeconds >= dayInSeconds -> {
+                val days = (totalSeconds / dayInSeconds).toInt()
+                val hours = ((totalSeconds % dayInSeconds) / hourInSeconds).toInt()
+                val minutes = ((totalSeconds % hourInSeconds) / minuteInSeconds).toInt()
+
+                val parts = mutableListOf<String>()
```
hichamboushaba: np, we can probably use `buildString` here instead of a `mutableListOf<String>`, for example for this first formatting logic:

```kotlin
                buildString {
                    append(
                        resourceProvider.getQuantityString(
                            quantity = days,
                            default = R.string.booking_duration_days,
                            one = R.string.booking_duration_day
                        )
                    )
                    if (hours > 0) {
                        append(" ")
                        append(
                            resourceProvider.getQuantityString(
                                quantity = hours,
                                default = R.string.booking_duration_hours,
                                one = R.string.booking_duration_hour
                            )
                        )
                    }
                    if (minutes > 0) {
                        append(" ")
                        append(
                            resourceProvider.getQuantityString(
                                quantity = minutes,
                                default = R.string.booking_duration_minutes,
                                one = R.string.booking_duration_minute
                            )
                        )
                    }
                }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
```diff
@@ -109,6 +114,110 @@ class BookingMapper @Inject constructor(
...
+            val difference = if (remainder == 0L) 0L else boundary - remainder
+            if (difference > 0 && difference <= minuteInSeconds) {
+                durationInSeconds += difference
+            }
+        }
+        return Duration.ofSeconds(durationInSeconds)
+    }
+
+    @Suppress("LongMethod")
+    private fun Duration.toHumanReadableFormat(): String {
```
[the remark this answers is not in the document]
hichamboushaba: I think AI has a point here, the method is quite long, and the multiple branches complicate the code a bit.

I'm suggesting a simpler version below, please check it out:

```kotlin
    private fun Duration.toHumanReadableFormat(): String {
        if (this < Duration.ofMinutes(1)) {
            return resourceProvider.getQuantityString(
                quantity = seconds.toInt(),
                default = R.string.booking_duration_seconds,
                one = R.string.booking_duration_second
            )
        }

        val days = toDays()
        val hours = minusDays(days).toHours()
        val minutes = minusDays(days).minusHours(hours).toMinutes()

        return buildString {
            if (days > 0) {
                append(
                    resourceProvider.getQuantityString(
                        quantity = days.toInt(),
                        default = R.string.booking_duration_days,
                        one = R.string.booking_duration_day
                    )
                )
            }
            if (hours > 0) {
                append(" ")
                append(
                    resourceProvider.getQuantityString(
                        quantity = hours.toInt(),
                        default = R.string.booking_duration_hours,
                        one = R.string.booking_duration_hour
                    )
                )
            }
            if (minutes > 0) {
                append(" ")
                append(
                    resourceProvider.getQuantityString(
                        quantity = minutes.toInt(),
                        default = R.string.booking_duration_minutes,
                        one = R.string.booking_duration_minute
                    )
                )
            }
        }.trim()
    }
```

This gives the same results, and the tests are green.

But please feel free to ignore the suggestion if you prefer the other version 🙂

hichamboushaba (approved):
Really nice work @AdamGrzybkowski.

I left some comments but nothing is blocking.

And just a question: how do you feel about moving the formatting logic to a separate class? I feel like it's something that can be used elsewhere if we ever use `Duration`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingMapper.kt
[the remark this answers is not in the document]
AdamGrzybkowski: I like it - code updated!

AdamGrzybkowski:
>And just a question: how do you feel about moving the formatting logic to a separate class? I feel like it's something that can be used elsewhere if we ever use Duration.

Done!

hichamboushaba (approved):
Nice work 💯

PR #14762, opened 2025-10-15
[WOOMOB-1209] Disable periodic worker for sites with too large catalog

malinajirka:
#WOOMOB-1209

### Description

This PR adds logic that disables periodic Local catalog sync for sites with a catalog that is too large. We simply store a flag when the catalog sync fails with `CatalogTooLarge` exception and skip all future syncs.

I considered cancelling the scheduled job in these cases, but since we use a single worker for all sites, I opted for this approach instead.

### Testing information

1. Clear app data
2. Login with a store that has more than 1000 products.
3. Observe logcat and notice the sync fails
4. Check shared preferences and verify the flag is `false` for the site - alternatively, you can modify how often is the worker run and wait until it runs again. Having said that, keep in mind that usually the OS won't let you run the worker more often than once in 15 minutes.

I have tested the above - using the approach where I reduced the period for the worker from 24 hours to 15 minutes and I waited.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
@@ -8,6 +8,7 @@ import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.login.AccountRepository
 import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
 import com.woocommerce.android.ui.woopos.featureflags.WooPosLocalCatalogM1Enabled
+import com.woocommerce.android.ui.woopos.util.datastore.WooPosPreferencesRepository
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 
@@ -23,6 +24,7 @@ constructor(
     private val syncRepository: WooPosLocalCatalogSyncRepository,
     private val logger: WooPosLogWrapper,
     private val featureFlagM1Enabled: WooPosLocalCatalogM1Enabled,
+    private val preferencesRepository: WooPosPreferencesRepository,
 ) : CoroutineWorker(appContext, workerParams) {
 
     companion object {
@@ -43,6 +45,11 @@ constructor(
         val site = selectedSite.getOrNull()
         if (site == null) {
             logger.e("No selected WooCommerce site found, skipping local catalog sync")
+            return Result.failure()
+        }
+
+        if (!preferencesRepository.isPeriodicSyncEnabledForSite(site.siteId)) {
+            logger.e("Periodic sync permanently disabled for site ${site.url}, skipping local catalog sync.")
             return Result.failure()
         }
 
@@ -72,8 +79,11 @@ constructor(
             }
 
             is PosLocalCatalogSyncResult.Failure.CatalogTooLarge -> {
-                // TBD Local Catalog - stop future syncs for this site if catalog too large
-                logger.e("Local catalog FULL sync failed: ${fullSyncResult.error}.")
+                preferencesRepository.disablePeriodicSyncForSite(site.siteId)
+                logger.e(
+                    "Local catalog FULL sync failed: ${fullSyncResult.error}. Permanently " +
+                        "disabling periodic sync for site ${site.siteId}."
+                )
                 Result.failure()
             }
         }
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -42,7 +44,12 @@ constructor(
 
         val site = selectedSite.getOrNull()
         if (site == null) {
-            logger.e("No selected WooCommerce site found, skipping local catalog sync")
+            logger.w("No selected WooCommerce site found, skipping local catalog sync")
```
malinajirka, line 47: 📓 I changed the logging level to warning.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -72,8 +79,11 @@ constructor(
             }
 
             is PosLocalCatalogSyncResult.Failure.CatalogTooLarge -> {
-                // TBD Local Catalog - stop future syncs for this site if catalog too large
-                logger.e("Local catalog FULL sync failed: ${fullSyncResult.error}.")
+                preferencesRepository.disablePeriodicSyncForSite(site.siteId)
```
samiuelson, line 82: ❓ Just to double check—are we sure we want to disable periodic sync forever, not handling the case when the merchant decreases the number of items in the catalog?

samiuelson (approved):
LGTM 👍 I left one question, feel free to merge.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
malinajirka: Good question @samiuelson - it's up to us. My reasoning was that we want to introduce support for stores of any size in the not-so-distant future, so I opted for keeping things simple. Wdyt?

samiuelson: Makes sense 👍

samiuelson: @malinajirka, I think we overlooked an important thing—we should also update the logic in [WooPosModule::provideWooPosProductsDataSource](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/di/WooPosModule.kt#L29):

```kotlin
@Provides
@Singleton
fun provideWooPosProductsDataSource(
    wooPosLocalCatalogM1Enabled: WooPosLocalCatalogM1Enabled,
    productsDataSource: WooPosProductsDataSource,
    productsInDbDataSource: WooPosProductsInDbDataSource
): WooPosProductsDataSourceInterface {
    return if (wooPosLocalCatalogM1Enabled()) {
        productsInDbDataSource
    } else {
        productsDataSource
    }
}
```
Currently, the binding logic considers only the FF when deciding which instance of `WooPosProductsDataSourceInterface` should be injected into the receiver classes (in case we want to guard the local catalog behind a remote FF in the future, it should also be included here). Now, we can end up in a situation where the local catalog is disabled, but we are not falling back to the in-memory caching. Wdyt?

malinajirka: Nice catch!

Let me create an issue - I think it won't be as simple fix as changing this condition, since we need to change the dataSource dynamically. I'll need to test how hilt behaves here, but since it's marked with `singleton` annotation I think it might call this method just once per app session.

https://linear.app/a8c/issue/WOOMOB-1530/woo-poslocal-catalog-fallback-to-legacy-data-source

PR #14765, opened 2025-10-15
[CIAB] Open bookable products in WebView

hichamboushaba:
Closes WOOMOB-1320

### Description
This PR updates the product navigation to make sure we open bookable products in a WebView.
The URL we use for the products is temporary now.

> [!NOTE]
> - I tried to simplify the navigation logic to product details by combining the various flows in a single function, and removing the extra navigation action `action_global_productDetailFragment_popUpTo_product_list`, please review commit by commit to better follow the changes. 
> - Loading the product details is super slow, I assume it's just because of the test site

### Steps to reproduce
1. Use the test CIAB site shared here p1758030630245249-slack-C03L1NF1EA3
2. Sign in using site credentials from the above site
3. Open a bookable product from the product list
4. Open a regular product from the product list
5. Open a bookable product from the top performers card in the dashboard
6. Open a regular product from the top performers card in the dashboard

### Testing information
- Confirm that bookable products are opened in a WebView.
- Confirm that other products are not affected.

The above.

### Images/gif
https://github.com/user-attachments/assets/d83b080b-fd4e-4844-9a1d-ca1cd44cc256


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/270">23.5</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stock/DashboardProductStockCard.kt
```diff
@@ -76,17 +75,13 @@ fun DashboardProductStockCard(
...
-    DisposableEffect(event, navController, lifecycleOwner) {
+    DisposableEffect(event, context, lifecycleOwner) {
         val observer = Observer { event: Event ->
             when (event) {
-                is OpenProductDetail -> navController.navigateSafely(
-                    NavGraphMainDirections.actionGlobalProductDetailFragment(
-                        mode = ShowProduct(event.productId),
-                    )
-                )
+                is OpenProductDetail -> (context.findActivity() as? MainActivity)?.showProductDetail(event.productId)
```
hichamboushaba: One other approach instead of getting the activity would be to use a custom Hilt `EntryPoint` and injecting `ProductDetailNavigator` here, I don't have a strong opinion about either approach, I just went for the simplest here.

AdamGrzybkowski (commented):
The navigation on the phone works as described, but... it doesn't in the Split Pane layout on the Product List. 

https://github.com/user-attachments/assets/a68f1c72-2246-4588-b99b-b03633945008

I guess, it's something that we want to support as well?

hichamboushaba:
Thanks @AdamGrzybkowski for the test and for catching the issue.

To have better tablet support, I worked on an alternative solution, please check the last commits. The main parts of the new solution are:
- This [logic](https://github.com/woocommerce/woocommerce-android/blob/53840125da5a77dc7fc630904819c3b4e69bc5e0/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailViewModel.kt#L430-L439) here where we identify Bookable products in `ProductDetailViewModel` and then switch to the WebView.
- This [logic](https://github.com/woocommerce/woocommerce-android/blob/53840125da5a77dc7fc630904819c3b4e69bc5e0/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/webview/ProductDetailWebViewFragment.kt#L49-L57) for the navigation back from the WebView.

Given the navigation is handled now inside `ProductDetailFragment`, we don't need to change any other parts of the app, so I reverted the previous commits that introduced `ProductDetailNavigator`.

The result for tablets now is:

https://github.com/user-attachments/assets/f804358e-a8c5-427d-a160-11cbbf2587a2

and for phones:

https://github.com/user-attachments/assets/f3897a71-5fae-40ed-94e4-1723a78edd4b

Ready for another round.

------

There is an issue about displaying the CIAB nav bar inside the WebView in tablets, I'll start a thread about it and decide if it's something that we want to hide, anyway this is outside of the scope of this PR anyway.

hichamboushaba:
I forgot to share, there is one additional advantage of this solution, in case the product wasn't cached before, the solution would still work well, we'll show a shimmer loading effect until the product is loaded, then we'll open the WebView:

https://github.com/user-attachments/assets/ab2994ce-7510-4ffa-92cc-481ffb798462

This wouldn't have worked with the previous solution.

AdamGrzybkowski (approved):
I like the new approach! The shimmer is a great bonus 👍 

:shipit:

AdamGrzybkowski:
One nitpick:

The webView has an extra entry animation in the split pane compared to the non-webView fragment


https://github.com/user-attachments/assets/a4546dd6-09a8-4936-b667-f48952978635

hichamboushaba:
> One nitpick:
> 
> The webView has an extra entry animation in the split pane compared to the non-webView fragment

I took a quick look at this, and couldn't find an easy solution, I tried disabling the animation, but I think it's not better, as this causes the ProdudctDetailFragment content to blink a bit before showing the WebView:

https://github.com/user-attachments/assets/ff43c3eb-bb52-4fc5-9222-863b2fe88863

If you have some ideas we can try please let me know, otherwise I think what we have here is fine, especially given all this sould be a temporary solution until we have a native editor, WDYT?

AdamGrzybkowski:
hmm you're right... The only thing that comes to my mind is to handle switching to WebView within the `ProductDetailFragment` instead of showing a new fragment. That should preserve the animation and the Toolbar. Only the content would switch. 

Not sure if it's worth it.

hichamboushaba:
> hmm you're right... The only thing that comes to my mind is to handle switching to WebView within the ProductDetailFragment instead of showing a new fragment. That should preserve the animation and the Toolbar. Only the content would switch.
>
> Not sure if it's worth it.

Yes, that would be a good alternative, but given how the ProductDetailFragment is already pretty complicated, I think the simplicity and separation of concerns of the current solution is better.
Let's merge this as is, if we have time post Alpha, we can revisit the solution.

PR #14767, opened 2025-10-16
[WOOMOB-1443] Filter out unsupported products and variations in Local Catalog

malinajirka:
#WOOMOB-1443

### Description

This PR filters out unsupported products and variations in the Local Catalog. Variations were filtered out even before, but they are filtered out on the database layer now.

### Testing information

1. Verify downloadable products are not visible in the POS
2. Verify downloadable variation isn't visible in the POS
3. Verify product with non-published status isn't visible in the POS

The above

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosProductsTypesFilterConfig.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosProductsTypesFilterConfig.kt
@@ -6,6 +6,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.DownloadableOptions
 import org.wordpress.android.fluxc.store.WCProductStore.ProductFilterOption
 import javax.inject.Inject
 
+// Note: For the Local Catalog feature, the filtering is done locally in WooPosProductsDao.
 class WooPosProductsTypesFilterConfig @Inject constructor() {
     val filters: Map<ProductFilterOption, String> =
         mapOf(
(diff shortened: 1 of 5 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosProductsTypesFilterConfig.kt
```diff
@@ -6,6 +6,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.DownloadableOptions
 import org.wordpress.android.fluxc.store.WCProductStore.ProductFilterOption
 import javax.inject.Inject
 
+// Note: For the Local Catalog feature, the filtering is done locally in WooPosProductsDao.
```
kidinov, line 9: Seeing the comment, I assume you considered reusing this for the filteriting to have a single source of truth for the whole POS; therefore, just curious why you didn't go with that approach?

kidinov (approved):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosProductsTypesFilterConfig.kt
malinajirka: Good question. I couldn't figure out a way on how to re-use it.
1. We likely want to filter on the DB layer for performance/memory reasons.
2. This config lives in a different module than the DB -> we can't create a circular dependency.
3. Room queries can reference only constants.
Wdyt? Can you think of a way on how to re-use it?

PR #14768, opened 2025-10-16
[Compose component] - Add loading param to WCOutlinedButton

AdamGrzybkowski:
### Description

As per the [request](https://github.com/woocommerce/woocommerce-android/pull/14752#discussion_r2432557675) this PR add the `loading` param to the WCOutlinedButton. 

>If you do, please also make sure to update WCColoredButton's implementation to pass enabled = enabled && !loading, to make sure we disable button during loading.

I didn't do that, because this changes the button appearance in loading state, here's the comparison:

| Current | Disabled |
|---|---|
| Screenshot 2025-10-16 at 11 23 49 | Screenshot 2025-10-16 at 11 24 35 |

I can do it, but this feels like a bigger decision.

### Steps to reproduce

Follow the steps from https://github.com/woocommerce/woocommerce-android/pull/14752 to test the button in the app. You can also launch the Preview with all the buttons.

### Testing information

The default loading param value is false, so this should be a safe change.

The above.

### Images/gif

Screenshot 2025-10-16 at 11 35 14

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
@@ -191,9 +191,9 @@ fun WCOutlinedButton(
     interactionSource: MutableInteractionSource? = null,
 ) {
     WCOutlinedButton(
-        onClick = onClick,
+        onClick = { if (!loading) onClick() },
         modifier = modifier,
-        enabled = enabled && !loading,
+        enabled = enabled,
         shape = shape,
         colors = colors,
         border = border,
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
@@ -123,10 +123,8 @@ fun WCColoredButton(
     ) {
         Box {
             if (loading) {
-                CircularProgressIndicator(
-                    color = LocalContentColor.current,
+                ButtonCircularProgressIndicator(
                     modifier = Modifier
-                        .size(24.dp)
                         .align(Alignment.Center)
                 )
             }
@@ -185,6 +183,7 @@ fun WCOutlinedButton(
     leadingIcon: @Composable (() -> Unit)? = null,
     trailingIcon: @Composable (() -> Unit)? = null,
     enabled: Boolean = true,
+    loading: Boolean = false,
     shape: Shape = MaterialTheme.shapes.small,
     colors: ButtonColors = ButtonDefaults.wcOutlinedButtonColors(),
     border: BorderStroke? = ButtonDefaults.wcOutlinedButtonBorder(enabled),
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
```diff
@@ -194,21 +193,28 @@ fun WCOutlinedButton(
     WCOutlinedButton(
         onClick = onClick,
         modifier = modifier,
-        enabled = enabled,
+        enabled = enabled && !loading,
```
hichamboushaba: Reading the PR description I'm confused with this, I thought you decided to not use this 😕.

I agree with the concern shared in the PR description, so I'm suggesting an alternative in the following patch:

```patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt	(revision ef4a400b1f7c7787e7eb4d471f4adc79b3fb0b27)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt	(date 1760629171929)
@@ -40,6 +40,8 @@
 import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
@@ -190,10 +192,11 @@
     contentPadding: PaddingValues = ButtonDefaults.ContentPadding,
     interactionSource: MutableInteractionSource? = null,
 ) {
+    val loadingState by rememberUpdatedState(loading)
     WCOutlinedButton(
-        onClick = onClick,
+        onClick = { if (!loadingState) onClick() },
         modifier = modifier,
-        enabled = enabled && !loading,
+        enabled = enabled,
         shape = shape,
         colors = colors,
         border = border,
@@ -544,6 +547,7 @@
                     )
                 }
             )
+
         }
     }
 }
```
WDYT?

hichamboushaba: The patch just prevents firing the click event, the click indication (ripple) will still happen, this is not ideal, but I think it's an OK tradeoff if we want to keep the code simple.

AdamGrzybkowski: The description was about `WCColoredButton`. The code above is for `WCOutlinedButton` - I'm not breaking any behavior as this is something new.

hichamboushaba: I see, I misunderstood the remark in the PR description, but still I think the remark applies to `WCOutlinedButton` too, as disabling the button causes using a different color for the progress indicator:

| | |
| - | - |
| Screenshot_20251017_151412 | Screenshot_20251017_151547 |

For bookings, this is not an issue because we override the content color locally, but anywhere where might want to use the same pattern with `WCOutlinedButton`, it will have the gray color, and won't match the app theme.
So I think we should still consider changing this, WDYT?

hichamboushaba (approved):
Thanks for taking care of this @AdamGrzybkowski, I'm pre-approving, but please check my comment about the `enabled = enabled && !loading`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/Buttons.kt
AdamGrzybkowski: Got it, pushed a commit with the above changes. I skipped the `rememberUpdatedState` - I don't think it has much value here.

PR #14769, opened 2025-10-16
[POS Orders] Details State + View Model

toupper:
Part of: WOOMOB-1157

Sorry for the long PR, _dummy_ UI and tests are quite long.

### Description

With this PR, we add the view model and state for the POS order details, plus a non-final UI to show it. Refunded information, final design, and Email receipt action are coming in the following PRs.

### Steps to reproduce

1. Go to POS
2. Tap on the ... Menu
3. Tap on Orders
4. Wait until orders are loaded. Once that happens, check the order details, selecting different orders for better appraisal.

### Images/gif

https://github.com/user-attachments/assets/f761510b-f16d-43e1-a66a-4bec946a785d

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
@@ -23,7 +23,9 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.ExperimentalMaterialApi
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.Inventory2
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -363,30 +379,259 @@ private fun OrdersList(
...
+                                text = it,
+                                style = WooPosTypography.BodySmall,
+                                color = WooPosTheme.colors.onSurfaceVariantHighest
+                            )
+                        }
+                        Spacer(Modifier.height(WooPosSpacing.Small.value))
+                        OrderStatusBadge(details.status)
+                    }
+
+                    androidx.compose.material3.Button(
```
kidinov: Shall we use WooPosButton from the DS system?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -363,30 +379,259 @@ private fun OrdersList(
...
     order: OrderItemViewState?,
+    details: OrderDetailsViewState,
+    onEmailReceiptButtonClicked: (Long) -> Unit = {}
 ) {
     Column(modifier = modifier.fillMaxSize()) {
         WooPosToolbar(
             modifier = Modifier.fillMaxWidth(),
-            titleText = order?.title ?: "--",
-            titleFontWeight = FontWeight.Bold
+            titleText = details?.number ?: order?.title ?: "--",
```
kidinov: What's written here is the same as `titleText = details.number`. You can simplify it. Also, you should remove ` order: OrderItemViewState?,` from the list of the params

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -363,30 +379,259 @@ private fun OrdersList(
...
-            modifier = Modifier
-                .fillMaxSize()
-                .padding(start = WooPosSpacing.Large.value, end = WooPosSpacing.Large.value)
-        ) {
-            WooPosText(
-                text = "Order details goes here",
-                style = WooPosTypography.BodyMedium,
-                color = MaterialTheme.colorScheme.onSurfaceVariant
+
+        WooPosLazyColumn(
```
kidinov: LazyColumn is for the large lists where the views are reused, while here it looks like just simple Column can be used

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -363,30 +379,259 @@ private fun OrdersList(
 }
 
 @Composable
-private fun OrderDetails(
+fun OrderDetails(
```
kidinov: Does it need to be public? Also, I think considering that the view is rather complex, it may make sense to extract this to a separate file calling it `WooPosOrderDetails`

kidinov: I guess this is for the next PR, but just in case. From the preview, notice:

* There is no margin between the text and the image

Image

* Notice that the price supposed to be aligned with the name of the product, not in the middle

Image

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -28,8 +31,9 @@ class WooPosOrdersViewModel @Inject constructor(
     private val wooCommerceStore: WooCommerceStore,
     private val selectedSite: SelectedSite,
     private val resourceProvider: ResourceProvider,
-    private val locale: Locale
-) : ViewModel() {
+    private val locale: Locale,
+    private val productImageMap: ProductImageMap,
```
kidinov: I don't think we need to use it. Coil (library we use for the image) caches the images - just pass the image URL there. Follow the same approach as in the products list

kidinov:
Hey @toupper 👋

I played with this, making it slightly simpler, and came up with the following proposal (AI based summary)

⏺ Changes Summary

  The changes refactor how order items and details are stored and managed in the POS Orders screen:

  Key Changes:

  1. WooPosOrdersState.kt:171-181
    - Changed items from List<OrderItemViewState> to Map<OrderItemViewState, OrderDetailsViewState>
    - Renamed selectedOrderDetails to selectedDetails
    - Removed selectedOrderId field
  2. WooPosOrdersViewModel.kt:34,59
    - Replaced ProductImageMap with WooPosGetProductById
    - Removed ordersById LinkedHashMap and listener callbacks
    - Created buildItemsMap() to pair items with their details upfront
  3. WooPosOrdersScreen.kt:184,286,383
    - Updated to work with map-based structure
    - Removed redundant order parameter from OrderDetails

  Why It's Better:

  Data Coupling: Items and details are now paired together in a single map structure, eliminating the need to search for details by ID and preventing synchronization issues.

  Simpler State: No need to track selected order ID separately - selection is determined by the isSelected flag on items, and details are directly accessible from the map.

  Cleaner Architecture: Removed the observer pattern for product image updates. Product data is now fetched directly when building the details model, making data flow more predictable and easier to reason about.

  Less Redundancy: Eliminates the need for maintaining a separate ordersById cache, reducing memory overhead and potential inconsistencies.

Wdyt?


[Refactor__Update_order_details_handling_and_improve_state_management.patch](https://github.com/user-attachments/files/22965949/Refactor__Update_order_details_handling_and_improve_state_management.patch)

kidinov (commented):
LGTM! I left a few suggestions, please take a look

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Good point. Even if the final design is not part of this PR, it's easy to change it now, so done in 0f6d77022828bff831097224b98d76fab054a068

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: good catch! Done in 636e0a06cfe085af280c1423320015039a4b92d4

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: That's true, done in 9a873098f8ff0948b3591bff2c71551b005e86d2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: That makes sense, done in 6bc2c7cdc3520eef5aeaba062a6dea8ddd8c9fd3

toupper: Awesome, let's keep that in mind for the design PR

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Thanks for the note, @kidinov! Small clarification: `productImageMap` isn’t about image caching (Coil already handles that) — it’s our lookup to get the product entity so we can read its image URL. Order line items don’t include a product image URL, so we need that extra fetch to display images. We follow the same pattern in `OrderDetailViewModel`.

toupper:
Thanks for the thorough review and suggestions to simplify the logic @kidinov! I applied them here e13107cc63684719ce6d9039d79cc142eaa5c149. Please take another look when you have time.

kidinov (approved):
LGTM! Thanks @toupper

PR #14770, opened 2025-10-16
Introduce tests for WooPosVariationMapper

malinajirka:
#WOOMOB-1286

### Description

Adds unit tests for WooPosVariationMapper.

Note: I decided to initialize an actual Gson instead of mocking it. Since it's a well-tested library, it feels better to have an instance of the class in the tests.

### Testing information

Green CI

Green CI

### Images/gif

N/A


samiuelson (approved):
LGTM, thank you!

PR #14773, opened 2025-10-17
[Woo POS][Local Catalog] Sync error dialog in Settings

samiuelson:
WOOMOB-1447

### Description

This PR displays the error dialog in POS Settings > Local Catalog, in case an error occurs during manual catalog sync. 

* Figma: FmV0bBXHmsGYZOgUJb0yzD-fi-10218_6796

Implementation details:
* Added communication system between parent settings screen and local catalog child screen for handling actions like: showing error dialog, retry, and dismiss. 
* Updated sync logic to display error dialog when catalog sync fails

### Steps to reproduce

1. Go to POS > Settings > Local Catalog
2. Run the manual sync and trigger an error (e.g., turn on airplane mode)

### Testing information

1. Verify the dialog's UI is correct
2. Test Retry button
3. Test dialog dismiss

Above

### Images/gif

Screenshot_20251017_102803


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka (approved):
Thanks @samiuelson! Looks and tests well.

The only thing I'm wondering about is whether we want to introduce this layer of complexity to the settings section of the POS. Good thing is that the approach is consistent to the approach in the POS Home However, it's complex and the only reason for that complexity is that we can display a dialog. 

I think it's a good solution for the designs, but I'm thinking it might be worth considering sharing with Wagner the complexity the designs introduce. I'm honestly not a fan of blocking dialogs so I think even from the UX POV, this design isn't great. In combination with the code-complexity, I'd personally just display a red notice below the `progress button` with the failure.

Wdyt? I'd also like to hear your opinion @kidinov since this is a decision which might shape the future of settings related code.

P.S. One note regarding the behavior - it might be worth adding 500ms artificial delay to the `retry` action - right now the dialog just moves few pixels whenever I tap on retry when my connection is still turned off.

samiuelson:
> Thanks @samiuelson! Looks and tests well.
> 
> The only thing I'm wondering about is whether we want to introduce this layer of complexity to the settings section of the POS. Good thing is that the approach is consistent to the approach in the POS Home However, it's complex and the only reason for that complexity is that we can display a dialog.
> 
> I think it's a good solution for the designs, but I'm thinking it might be worth considering sharing with Wagner the complexity the designs introduce. I'm honestly not a fan of blocking dialogs so I think even from the UX POV, this design isn't great. In combination with the code-complexity, I'd personally just display a red notice below the `progress button` with the failure.
> 
> Wdyt? I'd also like to hear your opinion @kidinov since this is a decision which might shape the future of settings related code.
> 
> P.S. One note regarding the behavior - it might be worth adding 500ms artificial delay to the `retry` action - right now the dialog just moves few pixels whenever I tap on retry when my connection is still turned off.

I'm fine with adding additional callbacks to `WooPosSettingsContent` composable: `onShowCatalogSyncErrorDialog: (String)->Unit`, `onDismissCatalogSyncErrorDialog: (String)->Unit`, and this was the initial approach I took. However, I couldn't find a clean way of passing `onRetryCatalogSyncTapped` callback from `WooPosSettingsViewModel` to `WooPosSettingsLocalCatalogViewModel`. I'm curious about your opinion or ideas, @kidinov and @malinajirka.

malinajirka:
> I'm fine with adding additional callbacks to WooPosSettingsContent composable: onShowCatalogSyncErrorDialog: (String)->Unit, onDismissCatalogSyncErrorDialog: (String)->Unit, and this was the initial approach I took. However, I couldn't find a clean way of passing onRetryCatalogSyncTapped callback from WooPosSettingsViewModel to WooPosSettingsLocalCatalogViewModel. 

As I mentioned, I believe the solution in this PR is a good solution. I'm just not sure if it's worth introducing the complexity just so we can display a blocking dialog (which IMO doesn't provide a good UX anyway).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogScreen.kt
```diff
@@ -274,6 +284,95 @@ private fun SectionTitle(title: String) {
     )
 }
 
+@Composable
+fun WooPosSyncErrorDialog(
+    isVisible: Boolean,
+    onRetry: () -> Unit,
+    onDismissRequest: () -> Unit,
+    modifier: Modifier = Modifier
```
kidinov: np: I believe there is a convention to have modifier as the first parameter

kidinov:
@malinajirka @samiuelson 

I have three points here:
* UX-wise, I feel it's not an ideal solution to show the dialog here, as, please correct me if I am wrong, the sync will be running in the background (and it can take quite some time). Now if I leave the settings and the sync fails, even if I come back, nothing will be displayed. Maybe to have just something on the settings local catalog screen like "The last sync failed at 13:44" for this case instead of the dialog, which won't be displayed anyway in many cases
* Even if the dialog is needed, I believe we can reuse the same event-based infrastructure we use on the home - it's all one activity and this has activity scope. We just need to separate events properly to keep it as manageable and simple as possible.
* We should not have both cross and cancel buttons on a dialog

kidinov:
@samiuelson @malinajirka 

Another minor thing I noticed not related to the PR - this should be clickable as a row, not only the toggle to initiate the settings change

image

malinajirka:
> UX-wise, I feel it's not an ideal solution to show the dialog here, as, please correct me if I am wrong, the sync will be running in the background (and it can take quite some time). Now if I leave the settings and the sync fails, even if I come back, nothing will be displayed. Maybe to have just something on the settings local catalog screen like "The last sync failed at 13:44" for this case instead of the dialog, which won't be displayed anyway in many cases

This is a manual sync, which should be used only in rare circumstances, and currently runs in a VM scope. This was an intentional decision. It might take a bit before it finishes, but since the user invoked it, they are likely seeing some issues so I think making it synchronous is desired from the UX perspective. Having said all that, we can update it.

> Even if the dialog is needed, I believe we can reuse the same event-based infrastructure we use on the home - it's all one activity and this has activity scope. We just need to separate events properly to keep it as manageable and simple as possible.


Good point, @samiuelson have you explored this approach?

> Another minor thing I noticed not related to the PR - this should be clickable as a row, not only the toggle to initiate the settings change

Good catch, I've created an issue.

kidinov:
@malinajirka thanks for providing more context!
> This is a manual sync, which should be used only in rare circumstances, and currently runs in a VM scope. This was an intentional decision. It might take a bit before it finishes, but since the user invoked it, they are likely seeing some issues so I think making it synchronous is desired from the UX perspective. Having said all that, we can update it.

Then I think we may need to make it clear that they have to be on this screen to finish the sync. Maybe showing the loading dialog there with a "cancel sync" button could be the solution to communicate this properly. Otherwise, currently it's highly likely people may just leave the screen, canceling the sync, which can be confusing for them

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/localcatalog/WooPosSettingsLocalCatalogScreen.kt
samiuelson: Moved up: 0a56aa7

samiuelson:
> > Even if the dialog is needed, I believe we can reuse the same event-based infrastructure we use on the home - it's all one activity and this has activity scope. We just need to separate events properly to keep it as manageable and simple as possible.
> 
> Good point, @samiuelson have you explored this approach?

Thanks for suggestion @kidinov and @malinajirka. Implemented here: 0c1b2e0

PR #14774, opened 2025-10-17
[POS Orders] Email Receipt 1/2

toupper:
Part of WOOMOB-1154

### Description

This PR wires up the “Email receipt” action from the POS Order Details screen. Tapping the button now triggers the navigation event to the Email Receipt flow for the selected order.

#### Scope
	•	Connects the Email Receipt button to ToEmailReceipt(orderId) via the existing children→parent navigation.
	•	No UI/UX changes beyond enabling the action.
	•	No data mutations in this part.

#### Next step (separate PR)
	•	Refresh the orders list and details after sending the email so the customer’s email is surfaced/updated in the UI.

### Steps to reproduce

1. Go to POS
2. Go to Orders through the ... menu
3. Tap on Email Receipt on the selected order
4. Enter your email address and tap on Email Receipt.
5. Once it succeeds, it goes back to the previous screen.

	•	Unit tests added to verify the navigation event is sent with the correct orderId.
	•	Manually verified from Order Details: the button triggers the Email Receipt flow.

### Images/gif

https://github.com/user-attachments/assets/69c6b5a5-538a-4813-a054-50af773994e2


kidinov (commented):
LGTM, but we need to adjust the back animation to fit the current navigation pattern. It should be going from left to right on both screens (e.g., check the settings screen back navigation to the home screen).

https://github.com/user-attachments/assets/008ef2d6-3d31-414a-9b03-759a3070df16

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModelTest.kt
```diff
@@ -46,6 +48,7 @@ class WooPosOrdersViewModelTest {
     private val resourceProvider: ResourceProvider = org.mockito.kotlin.mock()
     private val getProductById: WooPosGetProductById = org.mockito.kotlin.mock()
     private val providedLocale: Locale = Locale.US
+    private val childrenToParentEventSender: WooPosChildrenToParentEventSender = org.mockito.kotlin.mock()
```
kidinov: NP: We probably should use import for those.

toupper: Good catch, done in 60f870fd5636ea6587a1b32436c046e876e3d5f1

toupper:
> LGTM, but we need to adjust the back animation to fit the current navigation pattern. It should be going from left to right on both screens (e.g., check the settings screen back navigation to the home screen).

Good point, done in b4d30acf9573463aeabf4a081eecb52781c31261

toupper:
Thanks for review @kidinov! Please take another look

kidinov (approved):
LGTM, thanks!

PR #14779, opened 2025-10-17
[Woo POS][Historical Orders] Add refund data to model and screen

kidinov:
WOOMOB-1533

### Description
Adds refund information to the historical orders details screen in Woo POS. The order details now display individual refund line items and calculate net payment (total paid minus refunds).

Changes:
- Added `WooPosGetOrderRefundsByOrderId` utility to fetch refund data for orders
- Updated `OrderDetailsViewState.TotalsBreakdown` to include refunds list and net payment
- Modified `WooPosOrdersViewModel` to fetch and format refund amounts
- Updated `WooPosOrderDetails` UI to display refunds and net payment sections
- Added string resources for "Refunded" and "Net Payment" labels

Note: The UI layout is not final and will be refined in subsequent iterations.

### Steps to reproduce
1. Open Woo POS
2. Navigate to Historical Orders
3. Select an order that has refunds
4. Observe the order details screen showing:
   - Total paid with payment method
   - Individual refund line items (if any)
   - Net payment calculation (if refunds exist)

### Testing information
- Verified orders without refunds display correctly (no refund section shown)
- Verified orders with single refund show refund amount and net payment
- Verified orders with multiple refunds show all refunds separately
- Confirmed refund amounts are formatted with negative signs
- Confirmed net payment calculation is correct (total - sum of refunds)

### Images/gif
image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetOrderRefundsByOrderId.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetOrderRefundsByOrderId.kt
@@ -1,0 +1,21 @@
+package com.woocommerce.android.ui.woopos.common.data
+
+import com.woocommerce.android.model.Refund
+import com.woocommerce.android.model.toAppModel
+import com.woocommerce.android.tools.SelectedSite
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.wordpress.android.fluxc.store.WCRefundStore
+import javax.inject.Inject
+
+class WooPosGetOrderRefundsByOrderId @Inject constructor(
+    private val refundStore: WCRefundStore,
+    private val selectedSite: SelectedSite
+) {
+    suspend operator fun invoke(orderId: Long): List<Refund> {
+        return withContext(Dispatchers.IO) {
+            refundStore.getAllRefunds(selectedSite.get(), orderId)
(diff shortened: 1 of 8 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetOrderRefundsByOrderId.kt
```diff
@@ -0,0 +1,21 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.wordpress.android.fluxc.store.WCRefundStore
+import javax.inject.Inject
+
+class WooPosGetOrderRefundsByOrderId @Inject constructor(
+    private val refundStore: WCRefundStore,
+    private val selectedSite: SelectedSite
+) {
+    suspend operator fun invoke(orderId: Long): List<Refund> {
```
toupper, line 15: Please consider to add tests to this one if necessary

toupper (approved):
Great work @kidinov! LGTM 🚢 :shipit:

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetOrderRefundsByOrderId.kt
kidinov: Done!

8b8420af60e00cd0e9add3a2fb795027b08c3f5b

PR #14780, opened 2025-10-17
[CIAB] BookingListViewModel refactoring and some fixes

hichamboushaba:
### Description
This PR has three main changes:

- Implement the approach [discussed](https://github.com/woocommerce/woocommerce-android/pull/14735#discussion_r2420675966) previously to make the ViewModel more reactive, and avoid reading the fetch parameters from `StateFlow#value`.
- Fix an issue with pull-to-refresh and search (check the comments for more details)
- Fix an issue with the WCSearchView composable (check the comments for more details)

### Steps to reproduce
1. Use a CIAB site that has multiple bookings, multiple pages.
2. Open bookings list.

### Testing information
- Test changing tabs, searching.
- Test changing the sorting order, and confirm it shows the refresh status.
- Check for any regressions.

The above.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.take
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -48,7 +49,7 @@ class BookingListHandler @Inject constructor(
                 order = sortBy.toBookingsOrderOption()
             )
         } else {
-            searchResults
+            searchResults.take(page * PAGE_SIZE)
         }
     }.flatMapLatest { it }
 
@@ -61,6 +62,7 @@ class BookingListHandler @Inject constructor(
         page.value = 1
         canLoadMore.set(true)
 
+        val previousQuery = this.searchQuery.value
         this.searchQuery.value = searchQuery
         this.filters.value = filters
         this.sortBy.value = sortBy
@@ -68,7 +70,9 @@ class BookingListHandler @Inject constructor(
         return@withLock if (searchQuery == null) {
             fetchBookings()
         } else {
-            searchResults.value = emptyList()
+            if (searchQuery != previousQuery) {
+                searchResults.value = emptyList()
+            }
             if (searchQuery.isEmpty()) {
                 // If the query is empty, return cached results directly
                 canLoadMore.set(false)
@@ -85,10 +89,11 @@ class BookingListHandler @Inject constructor(
     }
 
     private suspend fun fetchBookings(): Result<Unit> {
+        val pageToFetch = page.value
         val isSearching = !searchQuery.value.isNullOrEmpty()
         val order = sortBy.value.toBookingsOrderOption()
         return bookingsRepository.fetchBookings(
-            page = page.value,
+            page = pageToFetch,
             perPage = PAGE_SIZE,
             query = searchQuery.value,
             filters = filters.value,
@@ -99,7 +104,11 @@ class BookingListHandler @Inject constructor(
                 page.update { it + 1 }
             }
             if (isSearching) {
-                searchResults.update { it + result.bookings }
+                if (pageToFetch == 1) {
+                    searchResults.value = result.bookings
+                } else {
+                    searchResults.update { it + result.bookings }
+                }
             }
         }.map { }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -4,6 +4,7 @@ import androidx.annotation.VisibleForTesting
 import com.woocommerce.android.ui.bookings.Booking
 import com.woocommerce.android.ui.bookings.BookingsRepository
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
@@ -23,6 +24,7 @@ class BookingListHandler @Inject constructor(
     companion object {
         @VisibleForTesting
         const val PAGE_SIZE = 25
+        private const val MIN_FETCH_DURATION_MS = 100L
     }
 
     private val mutex = Mutex()
@@ -75,7 +77,8 @@ class BookingListHandler @Inject constructor(
             }
             if (searchQuery.isEmpty()) {
                 // If the query is empty, return cached results directly
-                canLoadMore.set(false)
+                // Mimic network delay to allow the UI to show then hide the refreshing indicator
+                delay(MIN_FETCH_DURATION_MS)
                 Result.success(Unit)
             } else {
                 fetchBookings()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -9,7 +9,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.flatMapLatest
-import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
```diff
@@ -61,14 +62,17 @@ class BookingListHandler @Inject constructor(
...
         this.filters.value = filters
         this.sortBy.value = sortBy
 
         return@withLock if (searchQuery == null) {
             fetchBookings()
         } else {
-            searchResults.value = emptyList()
+            if (searchQuery != previousQuery) {
+                searchResults.value = emptyList()
+            }
```
hichamboushaba, line 77: This is the main part of the second change mentioned in the description, it fixes an issue where pull-to-refresh would result in displaying the Empty state during the fetch. I added two unit tests to confirm the behavior too.

Check the below recordings to see the diffference:
| Before | After |
| --- | --- |
| <video src=https://github.com/user-attachments/assets/1fd5ef87-bace-4093-9553-d494f07077b1 /> | <video src=https://github.com/user-attachments/assets/159c6605-f070-42d2-a644-a4bae219b9fd /> |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/TextFields.kt
```diff
@@ -209,10 +209,13 @@ fun WCSearchField(
     var textFieldValueState by rememberSaveable(stateSaver = TextFieldValue.Saver) {
         mutableStateOf(TextFieldValue(text = value))
     }
+    // Keep TextFieldValue in sync with external value
+    val textFieldValue = textFieldValueState.copy(text = value)
```
hichamboushaba, line 213: Without this, when the value is changed externally (for example by tapping on the clear button), the update is not reflected on the text field.
This line is copied from the `BasicTextField` String implementation.

| Before | After |
| --- | --- |
| <video src=https://github.com/user-attachments/assets/ddc46601-95dc-4238-8e04-3a1477b28087 /> | <video src=https://github.com/user-attachments/assets/b30bda06-cd35-4b59-819f-74b9e03aa317 /> |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
```diff
@@ -48,7 +49,7 @@ class BookingListHandler @Inject constructor(
                 order = sortBy.toBookingsOrderOption()
             )
         } else {
-            searchResults
+            searchResults.take(page * PAGE_SIZE)
```
AdamGrzybkowski: Do we need that? 

`searchResults` returns a List of bookings, not single bookings so I'm not sure why we take `page * PAGE_SIZE` 🤔

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -115,37 +116,66 @@ class BookingListViewModel @Inject constructor(
...
+                    initialLoadingState = initialLoadingState,
+                    fetchParams = fetchParams
                 )
+
+                refreshTrigger.collect {
+                    fetchBookings(
+                        initialLoadingState = BookingListLoadingState.Refreshing,
+                        fetchParams = fetchParams
+                    )
+                }
```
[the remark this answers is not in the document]
AdamGrzybkowski: I think we are good here, but I tend to agree, I never like a collect inside another collect.

AdamGrzybkowski (changes requested):
There is an issue when I try to pull to refresh with a search field activated but empty:


https://github.com/user-attachments/assets/b65a75a4-b66a-4d52-9771-19020a647e3f

AdamGrzybkowski:
Hmm the issue above might not be related to this PR 🤔

hichamboushaba:
@AdamGrzybkowski let me take a look, even if unrelated to the PR, it would be good to investigate it and fix it in this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
hichamboushaba: Given we don't clear the results now when after refresh if the query is not modified, in case more than a single page was previously fetched, this will mean that the "load more" login won't work as expected without this, as we won't reach the view that triggers it.
This is similar to what we do with the DB observation too.

Does this clarifies it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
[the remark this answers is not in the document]
hichamboushaba: Yes we are good, the `collectLatest` means that we will cancel the nested `collect` on each new emission.

The above patterns allows us to retrigger the fetching with the same parameters without reverting to reading the values from the `Flows` like how we did before, so I think it fits our need well.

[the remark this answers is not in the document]
hichamboushaba: I ended up getting rid of the nested collection, as it caused a minor issue before: pull-to-refresh action needed to wait for any pending fetch to finish before triggerring, with the last change in the commit 090a1489b23eea2c049e8c5304b5da664df5b1f6, we will cancel pending fetches on pull-to-refresh, which matches the previous behavior.

hichamboushaba:
@AdamGrzybkowski the issue you found should be fixed by the commit 6fe7f69. The cause of the issue was that `StateFlow` would conflate values emitted quickly before the collector had a chance to observe them, so with our logic to return results right away when clearing the search query, we were setting the loading state to `Idle` right after setting it to `Refreshing`, so the Compose code didn't get any update to `isRefreshing`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
AdamGrzybkowski: To be honest, no 😅 

>This is similar to what we do with the DB observation too.

DB uses that value to limit the nr of entries returned. Here we are `taking` a specific number of values (list#1, list#2...) so those seem like different things 🤔 

I'm testing the code and if I remove the `take` operator it it behaves in the same way. I can reach the view that triggers the next page load.

Something like this would make more sense `searchResults.map { it.take(page * PAGE_SIZE) }` but I'm not sure we need that at all. When the pull to refresh is triggered the `page` is reset to 1, so we for the short moment when the refresh is happening we would show less items than before triggering it 🤔

AdamGrzybkowski:
>@AdamGrzybkowski the issue you found should be fixed by the commit https://github.com/woocommerce/woocommerce-android/commit/6fe7f698d4520188876874083770452d614226a2. The cause of the issue was that StateFlow would conflate values emitted quickly before the collector had a chance to observe them, so with our logic to return results right away when clearing the search query, we were setting the loading state to Idle right after setting it to Refreshing, so the Compose code didn't get any update to isRefreshing.


This is fixed, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
hichamboushaba: > `searchResults.map { it.take(page * PAGE_SIZE) }`

That's what I thought I was doing 🤦 😅 

> I'm testing the code and if I remove the take operator it it behaves in the same way. I can reach the view that triggers the next page load.

The question here, is what's the end of the list? If we keep all the items, then the end of the list is further more than our current page, this breaks the pagination logic, because when the first page is fetched, we'll then get to this [line](https://github.com/woocommerce/woocommerce-android/blob/6fe7f698d4520188876874083770452d614226a2/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt#L111) while will remove the items you are seeing at the end of list, and then the list will jump above, and won't be a good experience.
Again, this logic is similar to what we do with DB, because we clear the DB after the fetch, we ask the DB to return only the number of items that fits in our pages count.

Let me know if this still doesn't clarify it 🙂, if you agree with the above, then I'll fix the logic to apply `take` on the list of items and not the `Flow`.

AdamGrzybkowski: >The question here, is what's the end of the list? If we keep all the items, then the end of the list is further more than our current page, this breaks the pagination logic, because when the first page is fetched, we'll then get to this [line](https://github.com/woocommerce/woocommerce-android/blob/6fe7f698d4520188876874083770452d614226a2/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt#L111) while will remove the items you are seeing at the end of list, and then the list will jump above, and won't be a good experience.

When you pull to refresh you are already at the top, so there's no visible jump in the list as there no visible change. 

Anyway

>Let me know if this still doesn't clarify it 🙂, if you agree with the above, then I'll fix the logic to apply take on the list of items and not the Flow.


That sounds good! My main issue was the `.take` on the Flow, which was breaking my brain, because I though I'm missing something obvious.

hichamboushaba: > When you pull to refresh you are already at the top, so there's no visible jump in the list as there no visible change.

But the user can scroll down when the fetch is ongoing, and if the fetch takes long, then this would be noticeable.

Thanks @AdamGrzybkowski for catching the issue, I made the change in the last commit edcbf72

AdamGrzybkowski (approved):
:shipit:

PR #14784, opened 2025-10-20
[WOOMOB-1543] Propagate network errors instead of missing server date

malinajirka:
Fixes WOOMOB-1543

### Description
This PR fixes an issue where network errors were being logged as "missing server date" instead of properly propagating the actual network error. Previously, when network requests failed, the error handling was incorrectly attributing all failures to missing server date, making debugging and error tracking difficult.

The fix ensures that network errors are properly propagated through the error handling chain, allowing for more accurate error reporting and better debugging capabilities in the WooCommerce POS local catalog functionality.

### To test
1. Simulate a network error scenario in WooCommerce POS local catalog
2. Observe that the error was previously logged as "missing server date" instead of the actual network error (in the logs)
3. With this fix, the proper network error should now be reported

### Images/gif
N/A - This is a backend error handling improvement


samiuelson (approved):
LGTM!

PR #14788, opened 2025-10-20
[Woo POS][Local Catalog] Optimize `catalog too large` check

samiuelson:
# Optimize catalog too large check

WOOMOB-1328

### Description

This PR optimizes the catalog sync process by adding an upfront check for catalog size before attempting to sync products and variations. Instead of fetching full product/variation data only to discover the catalog is too large, the implementation now makes lightweight API calls (requesting only 1 item per page) to read the X-WP-Total header and determine total counts.

**Key Changes:**
1. Added header parsing support for X-WP-Total to extract total item counts from API responses
2. Implemented fetchProductsCount() and fetchVariationsCount() methods that make minimal API calls
3. Added pre-sync catalog size validation that fails fast if total items exceed 1000

**⚠️ This PR should be tested on a site running WC version 10.3.0+.**

### Steps to reproduce

—

### Testing information

1. Try accessing POS with a store with 1000+ items in catalog. Observe that the local catalog sync fails.
2. Verify sync works fine on a store with <1000 items.

Above.

### Images/gif

N/A


malinajirka (approved):
Thanks @samiuelson, looks and works as expected.

I noticed that we check against MAX_TOTAL_ITEMS = 1000 for both full and incremental sync. I believe we decided to set the maximum for incremental sync to 3 pages (300 items), perhaps, worth including in this check too, wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
```diff
@@ -83,6 +91,51 @@ class WooPosLocalCatalogSyncRepository @Inject constructor(
         )
     }
 
+    @Suppress("ReturnCount")
+    private suspend fun checkCatalogSize(
+        site: SiteModel,
+        modifiedAfterGmt: String? = null,
+    ): PosLocalCatalogSyncResult? {
```
malinajirka: 🔍 I'd consider extracting this method into a separate `action` class.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>WooPosCheckCatalogSizeAction</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
samiuelson: Done: a9aa2c1

samiuelson:
> I noticed that we check against MAX_TOTAL_ITEMS = 1000 for both full and incremental sync. I believe we decided to set the maximum for incremental sync to 3 pages (300 items), perhaps, worth including in this check too, wdyt?

Good point. I added an incremental sync constraint: 2ac43a9

PR #14793, opened 2025-10-21
[POS Orders] Navigate to the blog when tapping on the empty state CTA

toupper:
WOOMOB-1150

### Description

With this PR we implement the CTA when the POS orders list is empty. After tapping the button, we navigate to the woo blog.

Other than that, we remove invalid references to coupons in the orders screen file and make some functions private.

### Steps to reproduce

With a store without orders:

1. Go to POS
2. Tap on Menu > Orders
3. See the empty state. Tap on Learn More
4. https://woocommerce.com/blog/ should be opened with the proper navigation 

### Images/gif

https://github.com/user-attachments/assets/774ce549-c4a3-41a8-88c5-6ee4bc418b23


kidinov (approved):
@toupper LGTM!

PR #14802, opened 2025-10-22
[Woo POS][Local Catalog] Skip background full sync if POS tab is not available

samiuelson:
WOOMOB-1558

### Description

This PR implements logic to skip background full sync of the local catalog when the POS tab is not available. This optimization prevents unnecessary background work.

### Testing information
Verify that in case POS is not available, we WooPosLocalCatalogSyncWorker skips the sync by returning success, however, it's not syncing the catalog.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
@@ -28,7 +28,7 @@ constructor(
     private val preferencesRepository: WooPosPreferencesRepository,
     private val syncRepository: WooPosLocalCatalogSyncRepository,
     private val logger: WooPosLogWrapper,
-    private val timeProvider: DateTimeProviderInterface,
+    private val timeProvider: DateTimeProvider,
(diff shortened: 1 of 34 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -37,6 +39,11 @@ constructor(
 
     @Suppress("ReturnCount")
     override suspend fun doWork(): Result {
+        val isPosTabAvailable = wooPosTabShouldBeVisible()
+        if (isPosTabAvailable.isSuccess && isPosTabAvailable.getOrNull() == false) {
+            return Result.success()
```
malinajirka, line 45: 📓 I'd consider adding a log why we are skipping just to simplify debugging.

malinajirka (approved):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
samiuelson: Added: 249debf

PR #14805, opened 2025-10-23
[POS - Historical Orders] Refresh orders after the receipt was sent.

toupper:
### Description

With this PR we refresh the selected POS order after the user sends the email recept and they are navigated back. We do so by passing a parameter to the back navigation event that is then read by the POS orders navigation logic.

### Test Steps

1. Go to POS
2. Go Orders
3. On the selected Order details, tap on Email Receipt
4. Enter the email and tap send
5. When that ends succesfully and you are navigated back, see how the selected order is refreshed and the email appears in details and the list

Verify also that when the email is not sent, i.e. by tapping on the back button, the orders aren't refreshed.

### Images/gif

https://github.com/user-attachments/assets/c12139d6-b05f-43cd-9fed-629befd5a263


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptScreen.kt
```diff
@@ -47,7 +49,7 @@ fun WooPosEmailReceiptScreen(onNavigationEvent: (WooPosNavigationEvent) -> Unit)
         onEmailAddressChanged = { viewModel.onUIEvent(WooPosEmailReceiptUIEvent.EmailChanged(it)) },
         onSendReceiptClicked = { viewModel.onUIEvent(WooPosEmailReceiptUIEvent.SendEmailClicked) },
         onBackClicked = { onNavigationEvent(WooPosNavigationEvent.GoBack) },
-        onEmailSent = { onNavigationEvent(WooPosNavigationEvent.GoBack) }
+        onEmailSent = { onNavigationEvent(WooPosNavigationEvent.GoBackWithResult(key = EMAIL_RECEIPT_SENT, true)) }
```
toupper, line 50: I was wondering if this logic needs to be in the view model, but given that we already have the `onEmailSent` callback and we don't add more model knowledge to the screen, I left it there.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/root/navigation/WooPosNavigationEventHandler.kt
```diff
@@ -36,15 +36,30 @@ fun NavHostController.handleNavigationEvent(
             }
             popBackStack()
         }
+
+        is WooPosNavigationEvent.GoBackWithResult -> {
```
toupper, line 25: I considered adding a specific event like `GoBackAfterReceiptSent`, but I’d rather keep this generic so it’s reusable across flows. Using `GoBackWithResult` keeps the navigation API abstract and avoids event proliferation (“one event per case”).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -75,10 +78,24 @@ val WOO_POS_ORDERS_TOOLBAR_HEIGHT = 56.dp
...
 
+    val entry = remember(navController) { navController.currentBackStackEntry }
+    LaunchedEffect(entry) {
+        entry?.savedStateHandle
+            ?.getStateFlow(EMAIL_RECEIPT_REFRESH_ORDERS, false)
+            ?.collect { shouldRefresh ->
+                if (shouldRefresh) {
+                    viewModel.onRefresh()
+                    // clear so it doesn't retrigger on recomposition / process death restore
+                    entry.savedStateHandle[EMAIL_RECEIPT_REFRESH_ORDERS] = false
```
kidinov: Probably we should remove this from the bundel, not set to false?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptScreen.kt
```diff
@@ -37,6 +37,8 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosThe
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
 import com.woocommerce.android.ui.woopos.root.navigation.WooPosNavigationEvent
 
+const val EMAIL_RECEIPT_SENT = "email_receipt_sent"
```
kidinov: I would consider placing this in the navigation because I believe it’s semantically related to that

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -75,10 +78,24 @@ val WOO_POS_ORDERS_TOOLBAR_HEIGHT = 56.dp
 @Composable
 fun WooPosOrdersScreen(
     onNavigationEvent: (WooPosNavigationEvent) -> Unit,
+    navController: NavHostController,
 ) {
     val viewModel: WooPosOrdersViewModel = hiltViewModel()
     val state by viewModel.state.collectAsState()
 
+    val entry = remember(navController) { navController.currentBackStackEntry }
```
kidinov: Have you considered not passing navController inside the screen but passing the value to the screen from navigation composable? Both, I think, are fine, but separating the navigation part from the actual screen may be a bit cleaner.

Example:

```
@Composable
fun MyNavHost() {
    val navController = rememberNavController()
    
    NavHost(navController = navController, startDestination = "screenA") {
        composable("screenA") { backStackEntry ->
            val result = backStackEntry.savedStateHandle
                .getStateFlow<String?>("selected_item", null)
                .collectAsState()
            
            ScreenA(
                selectedItem = result.value,
                onNavigateToScreenB = {
                    navController.navigate("screenB")
                }
            )
        }
        
        composable("screenB") {
            ScreenB(
                onItemSelected = { item ->
                    navController.previousBackStackEntry
                        ?.savedStateHandle
                        ?.set("selected_item", item)
                    navController.popBackStack()
                }
            )
        }
    }
}

@Composable
fun ScreenA(selectedItem: String?, onNavigateToScreenB: () -> Unit) {
    Column {
        Text("Selected: ${selectedItem ?: "None"}")
        Button(onClick = onNavigateToScreenB) {
            Text("Go to Screen B")
        }
    }
}

@Composable
fun ScreenB(onItemSelected: (String) -> Unit) {
    Column {
        Button(onClick = { onItemSelected("Item 1") }) {
            Text("Select Item 1")
        }
        Button(onClick = { onItemSelected("Item 2") }) {
            Text("Select Item 2")
        }
    }
}
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Yeah that's better, done in 02c24c8c620df9f3e7c14e312f4bdfe88fa37303

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptScreen.kt
toupper: That's true, done in ddf806d70a6ecef0274ab2676fdbe8405dce12c7

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Yeah, it's cleaner that way, done in 68e10c313bfca695097730e3b19f994341d5b3d2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptNavigation.kt
```diff
@@ -11,6 +11,7 @@ import com.woocommerce.android.ui.woopos.home.HOME_ROUTE
 import com.woocommerce.android.ui.woopos.root.navigation.WooPosNavigationEvent
 import com.woocommerce.android.ui.woopos.root.navigation.navigateOnce
 
+const val EMAIL_RECEIPT_SENT = "email_receipt_sent"
```
kidinov, line 14: np: I'd recommend naming it `EMAIL_RECEIPT_SENT_KEY` - for clarity

kidinov (commented):
The approach LGTM! But:
* Notice that it doesn't work and the email is not actually appeared neither on the list or in the order details
* The selection of the order is lost
* I doubt that we need to show the PTR indicator. Check the iOS UX for reference.


Also, there is an issue not related to the changes in the PR in`WooPosEmailReceiptRepository::sendReceiptByEmail` there is an assumption that the order is stored in the database, which may not be the case when you open the POS after a clean install; therefore, sending fails. So:
* We either need to take the order from the in-memory cache, but correct me if I am wrong, it contains just 1 page of orders at max
* We need to fetch the order from the remote if it's not in the database.

toupper:
Thanks for the comment @kidinov!

> Notice that it doesn't work and the email is not actually appeared neither on the list or in the order details
> The selection of the order is lost
> I doubt that we need to show the PTR indicator. Check the iOS UX for reference.

After merging from trunk, the order details are now displayed correctly.
Most importantly, after discussing with @staskus, we agreed that we don’t necessarily need to maintain full technical parity with iOS.

On iOS, the selected order is reloaded directly from the Email Receipt screen, but in our case, I’ve kept that logic within the POS Orders screen. This approach has the added benefit of making the email receipt action faster.

In this PR:
	•	We refresh only the selected order.
	•	We don’t show the Pull-to-Refresh indicator (since the user didn’t trigger it manually).
	•	Instead, we show a subtle LinearProgressIndicator to communicate that the order details are being refreshed.

Let me know what you think.

> Also, there is an issue not related to the changes in the PR inWooPosEmailReceiptRepository::sendReceiptByEmail there is an assumption that the order is stored in the database, which may not be the case when you open the POS after a clean install; therefore, sending fails. So:
> 
> We either need to take the order from the in-memory cache, but correct me if I am wrong, it contains just 1 page of orders at max
> We need to fetch the order from the remote if it's not in the database.

Good catch! Let's tackle that in a separate PR 👍

toupper:
> Also, there is an issue not related to the changes in the PR inWooPosEmailReceiptRepository::sendReceiptByEmail there is an assumption that the order is stored in the database, which may not be the case when you open the POS after a clean install; therefore, sending fails. So:
> 
> We either need to take the order from the in-memory cache, but correct me if I am wrong, it contains just 1 page of orders at max
> We need to fetch the order from the remote if it's not in the database.


Done in https://github.com/woocommerce/woocommerce-android/pull/14817

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -67,6 +67,17 @@ class WooPosOrdersDataSource @Inject constructor(
     suspend fun loadMore(searchQuery: String? = null): Result<List<Order>> =
         withContext(Dispatchers.IO) { loadNextPage(searchQuery) }
 
+    suspend fun refreshOrderById(orderId: Long): Result<Order> {
```
kidinov, line 70: You don't update the cache, so if I understand correctly, after returning back to this screen, the order will be displayed without the updated email as cached data firstly used

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -221,6 +221,52 @@ class WooPosOrdersViewModel @Inject constructor(
         }
     }
 
+    fun onBackFromSuccesfullySendingEmailReceipt() {
```
kidinov: There is a typo in `succesfully`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -221,6 +228,14 @@ private fun OrdersListPane(
 
         Spacer(modifier = Modifier.height(WooPosSpacing.Small.value))
 
+        AnimatedVisibility(visible = state.isRefreshingSelectedDetails) {
+            LinearProgressIndicator(
```
kidinov: We don't have a Woo POS-specific linear progress component, and the default one uses colors that do not match the POS colors or style. I'd suggest not having any loading element at all. It'll simplify the code and won't bother merchants with not really clear what and why reloading indicator

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -221,6 +221,52 @@ class WooPosOrdersViewModel @Inject constructor(
...
+    }
+
+    private fun setRefreshingOrderDetails(value: Boolean) {
+        val current = _state.value
+        if (current is WooPosOrdersState.Content) {
+            _state.value = current.copy(isRefreshingSelectedDetails = value)
+        }
+    }
+
+    private suspend fun applyOrderUpdate(updated: Order) {
```
kidinov, line 247: Probably can be simplified to

```
    private suspend fun applyOrderUpdate(updated: Order) {
        val current = _state.value as? WooPosOrdersState.Content ?: return
        val loaded = current.items as? WooPosOrdersState.Content.Items.Loaded ?: return

        val selectedId = loaded.items.keys.firstOrNull { it.isSelected }?.id
        val newItem = mapOrderItem(updated, selectedId)
        val newDetails = mapOrderDetails(updated)

        val newMap = loaded.items.entries.associate { (item, details) ->
            if (item.id == updated.id) newItem to newDetails else item to details
        }

        _state.value = current.copy(
            items = WooPosOrdersState.Content.Items.Loaded(newMap),
            selectedDetails = if (selectedId == updated.id) newDetails else current.selectedDetails
        )
    }
```

kidinov (approved):
It looks good to me; I'd just really consider removing the loading indicator, as this is more of a side effect of the temporary architecture we are currently using, and I think it's fine to not show it at the moment. If you think the loading indicator is needed, I think we need to consider using something else, not the one currently used, as I think it's not good for the current POS design language.

Also, there are a few minor suggestions - please take a look!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Good catch! Done in 7f696e1dc524285582e308c35c351a116ac7055f

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
toupper: Good point, fixed it in 9052df4a5f594a3d5464bbff131b0a9f2ddcdf82

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: Nice, let's remove it then. Done in 7ca6800bfa51cf9536bb3eb44f59024c29ffaf74

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
toupper: Thanks, changed in e829d88b2ddd59da17433d321a95c094e752e402

PR #14807, opened 2025-10-23
[WOOMOB-1562] Fallback to remote variations data source when needed

malinajirka:
Fixes WOOMOB-1562

Do not merge - Merge when the target branch is trunk.

Note: This PR exceeds the 300 lines limit. However, the code is just being moved or renamed, so no need to review the business logic.

### Description

This PR implements data source selection for WooPos variations, allowing the app to choose between remote data source and local DB data source based on sync requirements and availability. 

This is a follow up for https://github.com/woocommerce/woocommerce-android/pull/14798 that implements corresponding changes to the variations data source.

Instead of having two separate data sources for products and variations, I opted for merging them into a single data source with separate methods.

### Testing information

The most important test is that local catalog is not used when the FF is off.

With the FF on, test that local catalog is used (variations load instantly) on small stores. Also verify local catalog is NOT used on large stores (variations send remote request).

### Images/gif

No UI changes - this is an internal architecture improvement.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -38,14 +48,13 @@ class WooPosProductsDataSource @Inject constructor(
     private var activeSource: WooPosProductsDataSourceInterface? = null
 
     fun prepopulateProductsCache(): Flow<WooPosPrepopulatingDataStatus> = flow {
-        resetState()
```
malinajirka, line 41: I believe this is the only business logic change - everything else was just moved. Resetting the state here ins't needed since we reset `offset = 0` in fetchFirstProductsPage. (also notice the the `resetState` method wasn't used at all on trunk).

samiuelson (commented):
Thanks, @malinajirka! 

Switching between the local catalog and the hybrid data source works well in the Items screen. However, we still need to update this logic in other parts, e.g., `WooPosGetVariationById`, which is used in Cart and Totals:

```kotlin
    suspend operator fun invoke(productId: Long, variationId: Long): WooPosVariation? = withContext(IO) {
        when {
            wooPosLocalCatalogM1Enabled() -> {
                val siteModel = site.getOrNull() ?: return@withContext null
                val result = localCatalogStore.getVariation(
                    siteId = siteModel.id,
                    productId = productId,
                    variationId = variationId
                )
                result.getOrNull()?.toWooPosVariation(mapper)
            }
            else -> getFromInMemoryCache(productId, variationId)
        }
    }
```
Please let me know if you'd like to address this in a separate pull request, and I'll approve this one.

malinajirka:
Good catch @samiuelson ! I've created an issue for this task - https://linear.app/a8c/issue/WOOMOB-1574/woo-poslocal-catalog-use-local-db-in-cart-and-totals-screen.

samiuelson (approved):
LGTM!

PR #14808, opened 2025-10-23
[WOOMOB-1561] Booking details - cancel booking action

AdamGrzybkowski:
WOOMOB-1561

### Description

This PR implements the Cancel booking action (REST API part). There were some extra things that had to updated to make the whole Cancel work as expected:
- Update the logic to determine `Pay on site` badge. The previous `order.status != "completed"` didn't work well for `cancelled` bookings/orders`. This also follows the latest changes to the web logic. - p1761223364429439/1761044767.213429-slack-C09FHQNQERG. https://github.com/woocommerce/woocommerce-android/commit/d762e9a58f5a3c6952812a22260f46fdeed6a497
- Cancell button visibility - https://github.com/woocommerce/woocommerce-android/pull/14808/commits/7f131b1bc9a6293bf7a8f490cc1072f327d382b5
- Attendance section visibility - https://github.com/woocommerce/woocommerce-android/pull/14808/commits/5486a2703a6a4b756c441fdac02768c35185a671

Additionally, there's an extra commit that adds an option to refresh the underlying order when updating the booking https://github.com/woocommerce/woocommerce-android/pull/14808/commits/800901648935e63318d5ad601548779ff2de2a81. Up untill now, updating the booking was not affecting the order, but the new cancel booking is different as it also marks the order as cancelled. The order status is used to determine the badge so we need it updated. 

### Test Steps

1. Launch the app
2. Open the bookings tab
3. Find a booking that wasn't cancelled
4. Tap `Cancel` 
5. If successful, make sure that the button and the attendance section were hidden and the badge status changed
6. You can also verify that a request to refresh an order was made

Please, try changing the attendance status, saving a new note and verify network requests to make sure we are not refreshing the order.

### Images/gif

https://github.com/user-attachments/assets/2e82dc97-e56c-4790-b988-82232dd5276b

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -131,4 +131,7 @@ private val BookingUpdatePayload.asMap: Map<String, Any>
         note?.let {
             put("note", it)
         }
+        status?.let {
+            put("status", it.key)
+        }
     }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -8,6 +8,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooNetwork
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
 import org.wordpress.android.fluxc.utils.extensions.filterNotNull
+import org.wordpress.android.fluxc.utils.extensions.putIfNotNull
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -124,14 +125,8 @@ class BookingsRestClient @Inject constructor(
 }
 
 private val BookingUpdatePayload.asMap: Map<String, Any>
-    get() = buildMap {
-        attendanceStatus?.let {
-            put("attendance_status", it.key)
-        }
-        note?.let {
-            put("note", it)
-        }
-        status?.let {
-            put("status", it.key)
-        }
-    }
+    get() = mutableMapOf<String, Any>().putIfNotNull(
+        "attendance_status" to attendanceStatus?.key,
+        "note" to note,
+        "status" to status?.key
+    )
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
@@ -164,18 +164,17 @@ class BookingsStore @Inject internal constructor(
                 response.result != null -> {
                     val bookingDto = response.result
 
-                    val updatedBookingEntityResult = if (refreshOrder) {
+                    val updatedBookingEntity = if (refreshOrder) {
                         getBookingEntityWithRefreshedOrder(site, bookingDto)
+                            ?: getBookingEntityWithLocalOrder(site, bookingDto) // Fallback to local
                     } else {
                         getBookingEntityWithLocalOrder(site, bookingDto)
                     }
-                    if (updatedBookingEntityResult.isError) {
-                        return@withDefaultContext WooResult(updatedBookingEntityResult.error)
+                    if (updatedBookingEntity == null) {
+                        return@withDefaultContext WooResult(WooError(GENERIC_ERROR, UNKNOWN))
                     } else {
-                        updatedBookingEntityResult.model?.let {
-                            bookingsDao.insertOrReplace(it)
-                        }
-                        return@withDefaultContext WooResult(updatedBookingEntityResult.model)
+                        bookingsDao.insertOrReplace(updatedBookingEntity)
+                        return@withDefaultContext WooResult(updatedBookingEntity)
                     }
                 }
 
@@ -203,10 +202,10 @@ class BookingsStore @Inject internal constructor(
     private suspend fun getBookingEntityWithRefreshedOrder(
         site: SiteModel,
         bookingDto: BookingDto
-    ): WooResult<BookingEntity> {
+    ): BookingEntity? {
         val orderResult = orderStore.fetchSingleOrderSync(site, bookingDto.orderId)
         if (orderResult.isError) {
-            return WooResult(orderResult.error)
+            return null
         } else {
             val entity = with(bookingDtoMapper) {
                 bookingDto.toEntity(
@@ -214,20 +213,20 @@ class BookingsStore @Inject internal constructor(
                     orderEntity = orderResult.model,
                 )
             }
-            return WooResult(entity)
+            return entity
         }
     }
 
     private suspend fun getBookingEntityWithLocalOrder(
         site: SiteModel,
         bookingDto: BookingDto
-    ): WooResult<BookingEntity> {
+    ): BookingEntity? {
         val storedBooking = bookingsDao.getBooking(
             localSiteId = site.localId(),
             bookingId = bookingDto.id,
         )
         if (storedBooking == null) {
-            return WooResult(WooError(GENERIC_ERROR, UNKNOWN))
+            return null
         }
         val entity = with(bookingDtoMapper) {
             bookingDto.toEntity(
@@ -238,6 +237,6 @@ class BookingsStore @Inject internal constructor(
                 order = storedBooking.order,
             )
         }
-        return WooResult(entity)
+        return entity
     }
 }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
@@ -155,6 +155,7 @@ class BookingsStore @Inject internal constructor(
         site: SiteModel,
         bookingId: Long,
         bookingUpdatePayload: BookingUpdatePayload,
+        refreshBooking: Boolean = false,
     ): WooResult<BookingEntity> {
         return coroutineEngine.withDefaultContext(AppLog.T.API, this, "updateBooking") {
             val response = bookingsRestClient.updateBooking(site, bookingId, bookingUpdatePayload)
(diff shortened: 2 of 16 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -134,3 +139,13 @@ internal class BookingEntityConverters {
...
+
+val BookingEntity.isCancellable: Boolean
+    get() = status !in listOf(
+        BookingEntity.Status.Cancelled,
+        BookingEntity.Status.InCart,
+        BookingEntity.Status.Complete
+    )
+
+val BookingEntity.isAttendanceStatusEditable: Boolean
+    get() = status != BookingEntity.Status.Cancelled
```
AdamGrzybkowski, line 151: Matches booking-attendance-card.tsx#L67-L69 n CIAB admin

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -134,3 +139,13 @@ internal class BookingEntityConverters {
         return BookingEntity.AttendanceStatus.fromKey(key)
     }
 }
+
+val BookingEntity.isCancellable: Boolean
+    get() = status !in listOf(
+        BookingEntity.Status.Cancelled,
+        BookingEntity.Status.InCart,
+        BookingEntity.Status.Complete
+    )
```
AdamGrzybkowski, line 148: Matches cancel-booking.tsx#L43-L49 in CIAB admin

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -155,6 +155,7 @@ class BookingsStore @Inject internal constructor(
         site: SiteModel,
         bookingId: Long,
         bookingUpdatePayload: BookingUpdatePayload,
+        refreshBooking: Boolean = false,
```
JorgeMucientes: Np. But this param what really does is refresh the associated order, should it be renamed to "refreshOrder" instead?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
```diff
@@ -131,4 +131,7 @@ private val BookingUpdatePayload.asMap: Map<String, Any>
         note?.let {
             put("note", it)
         }
+        status?.let {
+            put("status", it.key)
+        }
     }
```
JorgeMucientes: Minor suggestion you can use `putIfNotNull` to simplify this code: 


```suggestion
private val BookingUpdatePayload.asMap: Map<String, Any>
    get() = mutableMapOf<String, Any>().putIfNotNull(
        "attendance_status" to attendanceStatus?.key,
        "note" to note,
        "status" to status?.key
    )
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
JorgeMucientes, line 148: Wdyt of handling this "cancelable" state like this: 
```kotlin
    enum class Status(
        val key: String,
        val isCancellable: Boolean
    ) {
        Unpaid("unpaid", isCancellable = true),
        PendingConfirmation("pending-confirmation", isCancellable = true),
        Confirmed("confirmed", isCancellable = true),
        Paid("paid", isCancellable = true),
        Cancelled("cancelled", isCancellable = false),
        Complete("complete", isCancellable = false),
        InCart("in-cart", isCancellable = false),
        Unknown("unknown", isCancellable = false);

        companion object {
            fun fromKey(key: String): Status {
                // Use firstOrNull for a safe lookup, defaulting to Unknown
                return entries.firstOrNull { it.key == key } ?: Unknown
            }
        }
    }
```
I expect the different statuses to be something that could change in the future. With the above approach we'll always be forced to think is a new/updated status is cancelable or not.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
JorgeMucientes: PS: In case you decide to rename this param, don't forget to update test names [here](https://github.com/woocommerce/woocommerce-android/pull/14808/files#diff-4f28be37ca09b0d430b2e2523b35982db47fd10d5445eacda922e4c09fe25ecfR54)

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -163,24 +164,18 @@ class BookingsStore @Inject internal constructor(
...
-                        )
-                        bookingsDao.insertOrReplace(updatedBooking)
-                        return@withDefaultContext WooResult(updatedBooking)
+                    if (updatedBookingEntityResult.isError) {
+                        return@withDefaultContext WooResult(updatedBookingEntityResult.error)
+                    } else {
+                        updatedBookingEntityResult.model?.let {
+                            bookingsDao.insertOrReplace(it)
+                        }
+                        return@withDefaultContext WooResult(updatedBookingEntityResult.model)
```
JorgeMucientes: I'm not sure about this logic. 
If cancelling the booking succeeds but then for whatever reason we fail to refresh the order, we'll display an error to the user stating that the booking wasn't cancelled, but it was. I think that can be very confusing. I'm aware is an edge case so I propose one of 2 things:
- Ignore the order refresh error. The most important is that the booking was cancelled successfully and that things will be reflected correctly in `wp-admin` as well as in order details as soon as the app is able to sync successfully again. 
- Display a custom error specific to the failure to refreshing the order. 

I think the first option should be enough. Let me know your thoughts

JorgeMucientes (changes requested):
Great work @AdamGrzybkowski 👏🏼 
I've left a couple non-blocking nitpicks, and one thing that I believe is worth checking before merging. Let me know what you think :)

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
AdamGrzybkowski: I had some similar thoughts and, in the end, went with the above approach. I was mostly worried that not showing the error would be weird if the state on the screen wasn't correct. 

But given that you had the same feeling, I think the solution nr one (at least for now, might be better). I will update it.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
AdamGrzybkowski: >With the above approach we'll always be forced to think is a new/updated status is cancelable or not.

That's tempting but: 
- It couples the state (status) with behavior (cancel action).
- Those enums could get pretty big if we decided to add more properties (isRefundable, isPayable, etc.). That would make them hard to maintain.
- At this point, the logic checks the Booking status only, but that could change. In such case, we would have to redo the logic completely, whereas here, we would only have to update one extension property. 

Given the above, I would prefer to keep the current one. WDYT?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
AdamGrzybkowski: omg, you're correct 🤦 Thanks!

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
AdamGrzybkowski: TIL! Love it, updated!

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
JorgeMucientes: > At this point, the logic checks the Booking status only, but that could change. In such case, we would have to redo the logic completely, whereas here, we would only have to update one extension property.

True, that is a very good point. Let's leave it as is

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
JorgeMucientes: > I was mostly worried that not showing the error would be weird if the state on the screen wasn't correct.

If we kept the original approach, the only place where the inconsistent status would be displayed is if the user opens the order detail right? If that is the case, that seems like a less likely scenario than having an inconsistent state for sure in the booking detail screen itself. 

Ignoring the refresh order error, is a tradeoff too, but I think has way less impact than the above. 

Anyway thanks for addressing the change 🙏🏼

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
JorgeMucientes: Just minor reminder after seeing the changes added here: [fed7b12](https://github.com/woocommerce/woocommerce-android/pull/14808/commits/fed7b12e58fb94a6ee7e0de7752bd875d906cf3f) (ignore this comment if you are still working on it)
I think it would be worth also updating the test names like [here](https://github.com/woocommerce/woocommerce-android/pull/14808/files#diff-4f28be37ca09b0d430b2e2523b35982db47fd10d5445eacda922e4c09fe25ecfR54) as they are referencing `refreshBooking` directly.

AdamGrzybkowski: Thanks, I missed the tests. Should be good now.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
AdamGrzybkowski: I wasn't able to finish this on Friday, that's why I didn't request another review. I went with your suggestion and the order refresh fail won't be returned. Instead, the app will fallback to the local entity.

AdamGrzybkowski:
@JorgeMucientes This is not ready for another round.

JorgeMucientes (approved):
Great work @AdamGrzybkowski 
Thanks for addressing the changes 👍🏼

PR #14814, opened 2025-10-24
[WOOMOB-1574] Use DataSource in Totals and Cart

malinajirka:
Fixes #woomob-1574

### Description

This PR is a follow up for https://github.com/woocommerce/woocommerce-android/pull/14807 that ensures the correct data source is used even in Totals and Cart sections of the POS.

Tbh I'm not excited about the implementation
- We access rest client outside of store - this breaks FluxC architecture
- We don't handle cases when the variation is not in the database - might happen during ongoing sync
- The code isn't tested

Having said that, considering these issues apply just to the Remote data source that will be removed, I decided to focus on migrating the code without improving it.

### Test Steps

Smoke test POS

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetProductById.kt
```diff
@@ -1,65 +1,16 @@
...
-import com.woocommerce.android.ui.woopos.featureflags.WooPosLocalCatalogM1Enabled
+import com.woocommerce.android.ui.woopos.home.items.products.WooPosProductsDataSource
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.model.LocalOrRemoteId
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.ProductRestClient
-import org.wordpress.android.fluxc.store.pos.localcatalog.WooPosLocalCatalogStore
 import javax.inject.Inject
 
 class WooPosGetProductById @Inject constructor(
```
samiuelson, line 10: ❓ Does it make sense to keep these use case classes, given they became simple wrappers over `WooPosProductsDataSource`?

samiuelson (approved):
Looks good 👍 I left non-blocking question.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosGetProductById.kt
malinajirka: I was thinking about this, but I left them with the intention that when we remove the remote data source, we might get rid of the DataSource class and migrate back to the use-cases. Wdyt?

samiuelson: We'll see 👍

PR #14816, opened 2025-10-24
[Woo POS][Local Catalog] Refactor and fix pull-to-refresh logic in products and variations

samiuelson:
WOOMOB-1556

### Description

This PR refactors pull-to-refresh logic for both products and variations in the WooCommerce POS local catalog feature. The changes consolidate different refresh behaviors by introducing new `refreshProducts()` and `refreshVariations()` methods that delegate to either local catalog incremental sync or remote data source refresh based on feature flags and sync requirements.

### Test Steps

* Crucial: Verify that there's no regression to the existing PTR behavior in the hybrid data source.
* Verify that local catalog incremental sync is performed for both variations and products.
* Verify that local catalog incremental sync is not performed if the local catalog is not enabled.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
```diff
@@ -337,32 +320,42 @@ class WooPosProductsViewModel @Inject constructor(
...
-    private fun buildViewStateForSyncResult(syncResult: PosLocalCatalogSyncResult): WooPosProductsViewState =
-        when (syncResult) {
-            is PosLocalCatalogSyncResult.Success -> {
-                hidePTRIndicator()
-            }
+            dataSource.refreshProducts().collect { result ->
+                when (result) {
+                    is WooPosProductsDataSource.ProductsResult.Cached -> {
+                        if (result.products.isNotEmpty()) {
+                            _viewState.value = result.products.toContentState()
```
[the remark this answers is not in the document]
samiuelson: Cached results are intermediate, meaning we're still waiting for Remote results, so if empty Cached result is received, we should keep showing PTR indicator as we're awaiting Remote results.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsViewModel.kt
```diff
@@ -254,32 +232,43 @@ class WooPosVariationsViewModel @Inject constructor(
...
-        when (syncResult) {
-            is PosLocalCatalogSyncResult.Success -> {
-                hidePTRIndicator()
-            }
+            dataSource.refreshVariations(productId).collect { result ->
+                when (result) {
+                    is WooPosProductsDataSource.VariationsResult.Cached -> {
+                        // Handle cached variations
+                        if (result.data.isNotEmpty()) {
+                            _viewState.value = createContentState(result.data, productId)
```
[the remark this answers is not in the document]
samiuelson: Cached results are intermediate, meaning we're still waiting for Remote results, so if empty Cached result is received, we should keep showing PTR indicator as we're awaiting Remote results.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -223,6 +240,68 @@ class WooPosProductsInDbDataSource @Inject constructor(
...
+                VariationsResult.Remote(
+                    Result.failure(Exception("No site selected"))
+                )
+            )
+        } else {
+            fetchFirstVariationsPage(productId, forceRefresh = true).collect { emit(it) }
+        }
+    }.flowOn(Dispatchers.IO)
+
+    private suspend fun shouldUseIncrementalSync(): Boolean {
```
malinajirka: ❓  I'm not sure I follow the code - why do we invoke `fetchFirstVariationsPage` from refreshVariations 😕 ?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -223,6 +240,68 @@ class WooPosProductsInDbDataSource @Inject constructor(
 
     override fun canLoadMoreVariations(numOfVariations: Int): Boolean = false
 
+    override suspend fun refreshProducts(): Flow<WooPosProductsDataSource.ProductsResult> = flow {
```
malinajirka: ❓ It feels a bit off that `refresh` returns a flow, especially since it always emits just one response. Have you considered returning `Result<Unit>` and relying on Room updating `getProductsFromDatabaseFlow` that is being observed? Alternatively, we could return Result<ProductsResult> which isn't optimal but might simplify implementing the remote approach.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CatalogTooLarge</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: Fixed: 801786d76ccca43e62797219c19e43bb25635770

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: @malinajirka, thanks for your question. This is not obvious how to marry the Hybrid and LocalCatalog incremental sync results. Technically, the Hybrid returns a flow of `Cached` (intermediate) and `Remote` (final) results. However, in case `forceRefresh=true`, the cache gets cleared, I believe we can be interested solely in the `last` emitted result and can return it as a result. Could you please take a look at my refactor commit here: a7fd7dbd322dd82de15d997d9e0124849e935a27 and let me know what you think about the approach of updating/consolidating the data sync result models?

malinajirka: > This is not obvious how to marry the Hybrid and LocalCatalog incremental sync results. 

Yeah you are right, it's not trivial. I think ideally both approaches would return data from cache - the only difference should be whether it's persistent db or in-memory cache. The fact that we return the response directly isn't optimal, as it means we have multiple sources of truth. Having said that, I understand that's out of scope. Moreover, the VM logic currently determines whether to hide/show progress/error-screen, based on whether the returned data are cache/remote, so it's not something we could easily refactor.


> Could you please take a look at my refactor commit here: https://github.com/woocommerce/woocommerce-android/commit/a7fd7dbd322dd82de15d997d9e0124849e935a27 and let me know what you think about the approach of updating/consolidating the data sync result models?

I like it!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -232,6 +236,24 @@ class WooPosProductsInDbDataSource @Inject constructor(
         )
         return result.getOrNull()?.toWooPosVariation(variationMapper)
     }
+
+    override suspend fun refreshProducts(): Result<WooPosSyncResult> = withContext(Dispatchers.IO) {
+        selectedSite.getOrNull()?.let { site ->
+            val syncResult = localCatalogSyncRepository.syncLocalCatalogIncremental(site)
+            Result.success(syncResult)
+        } ?: Result.success(
```
malinajirka: ❓ Is this intentionally returning `Success` when site is not selected?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -232,6 +236,24 @@ class WooPosProductsInDbDataSource @Inject constructor(
...
+            PosLocalCatalogSyncResult.Failure.UnexpectedError("No site selected")
+        )
+    }
+
+    override suspend fun refreshVariations(productId: Long): Result<WooPosSyncResult> = withContext(Dispatchers.IO) {
+        selectedSite.getOrNull()?.let { site ->
+            val syncResult = localCatalogSyncRepository.syncLocalCatalogIncremental(site)
+            Result.success(syncResult)
+        } ?: Result.success(
```
malinajirka: Same as above.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -488,6 +510,24 @@ class WooPosProductsRemoteDataSource @Inject constructor(
             ?.toWooPosVariation(variationMapper)
     }
 
+    override suspend fun refreshProducts(): Result<WooPosSyncResult> = withContext(
```
malinajirka: > when forceRefresh = true the in-memory cache gets wiped out

I can't fine the convo right now, but Andrei mentioned we have not been wiping cache on PTR - I remember we used to do it early in the POS but then switched to keeping the in-memory cache until app restart. Do you know at which point we started calling `productsIndex.clearCache()` in the data source implementation? I mean, why we decided to do it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsViewModel.kt
```diff
@@ -254,32 +228,60 @@ class WooPosVariationsViewModel @Inject constructor(
...
-        when (syncResult) {
-            is PosLocalCatalogSyncResult.Success -> {
-                hidePTRIndicator()
-            }
-
-            is PosLocalCatalogSyncResult.Failure -> {
+            val result = dataSource.refreshVariations(productId)
+            result.onSuccess { posSyncResult ->
+                when (posSyncResult) {
+                    is com.woocommerce.android.ui.woopos.localcatalog.PosLocalCatalogSyncResult.Success -> {
```
malinajirka: 🔍 We could remove the full path.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosHybridSyncResult.kt
```diff
@@ -0,0 +1,16 @@
+package com.woocommerce.android.ui.woopos.localcatalog
+
+import com.woocommerce.android.ui.woopos.common.data.WooPosVariation
+import com.woocommerce.android.ui.woopos.common.data.models.WooPosProductModel
+
+sealed class WooPosHybridSyncResult : WooPosSyncResult()
```
malinajirka: 💡 I think I understand the idea here, but I would consider doing it the other way round. The current implementation returns a result which can be either ProductsResult or VariationsResults (for hybrid approach). That is a bit misleading and forces us to ignore variationsResult in product VM and product result in Variations VM. 

Instead, we could have two almost identical copies for local Catalog:
```
sealed class WooPosSyncProductResult
sealed class WooPosSyncVariationResult
```


Local:
```
sealed class PosLocalCatalogProductSyncResult : WooPosSyncProductResult() {
        val value: PosLocalCatalogSyncResult    
}

sealed class PosLocalCatalogSyncResult : WooPosSyncVariationResult() {
    val value: PosLocalCatalogSyncResult
}


sealed class PosLocalCatalogSyncResult {
    data class Success(
        val productsSynced: Int,
        val variationsSynced: Int,
        val syncDurationMs: Long
    ) : PosLocalCatalogSyncResult()

    sealed class Failure(val error: String) : PosLocalCatalogSyncResult() {
        class CatalogTooLarge(error: String, val totalPages: Int, val maxPages: Int) : Failure(error)
        class UnexpectedError(error: String) : Failure(error)
    }
}
```

Hybrid:
```
sealed class WooPosHybridProductSyncResult : WooPosSyncProductResult()
sealed class WooPosHybridVariationSyncResult : WooPosSyncVariationResult()

```

Wdyt?

malinajirka (commented):
Thanks @samiuelson !

As we discussed I haven't tested this PR since I'll test it when we update the behavior of cleaning cache on PTR.

I've left some suggestions/questions but it looks good overall and I believe can be merged as is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: I just found out that the only place the in-memory cache is cleared is in `WooPosProductsDataSource::prepopulateProductsCache` which is called in `WooPosSplashViewModel`. On PTR only the productsIndex is cleared (responsible for pagination) — not the in-memory cache (`WooPosProductsCache`). This means that we are not clearing the cache and Andrey's issue is related to something else.

malinajirka: Ahh, good find! So this PR is ready to be tested, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosHybridSyncResult.kt
malinajirka: No need to update this - the more I think about it the more I feel like perhaps the downside of the current approach aren't that significant to justify this refactoring.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: Yes, ready 👍

malinajirka (approved):
Works as expected!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: Fixed: 699b021

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
samiuelson: Fixed: 699b021

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/variations/WooPosVariationsViewModel.kt
samiuelson: Done: fa6161a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosHybridSyncResult.kt
samiuelson: I like the idea, done: 5cefb66

PR #14817, opened 2025-10-24
[POS Orders] Fetch order when sending receipt if it's not in DB

toupper:
WOOMOB-1154

### Description

From this [comment](https://github.com/woocommerce/woocommerce-android/pull/14805#pullrequestreview-3370419387)

We fetch the order remotely if it's not in the DB

### Test Steps

Open POS after a clean install of the app, and ensure that the receipt is sent.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptRepository.kt
```diff
@@ -65,6 +67,21 @@ class WooPosEmailReceiptRepository @Inject constructor(
             orderMapper.toAppModel(it)
         }
 
+    suspend fun fetchOrderById(orderId: Long): Order? {
```
toupper: @kidinov This code is basically duplicated from [OrderDetailRepository](https://github.com/woocommerce/woocommerce-android/blob/09fa4970ef1759fba18139727e71a21915bee02e/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt#L43) I opted for keeping it duplicated instead of extracting it into a separated helper class so we keep POS decoupled. Let me know what you think

kidinov: Can be private?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptRepository.kt
```diff
@@ -65,6 +67,21 @@ class WooPosEmailReceiptRepository @Inject constructor(
...
+        val result = withTimeoutOrNull(AppConstants.REQUEST_TIMEOUT) {
```
kidinov: I think it can be as simple as 
```
    suspend fun fetchOrderById(orderId: Long): Order? {
        val result = orderStore.fetchSingleOrder(
            selectedSite.get(),
            orderId
        )

        return if (!result.isError) {
            getOrderById(orderId)
        } else {
            null
        }
    }
```

`withTimeoutOrNull` looks like it's redundant. There is already a request on timeout in the network layer

kidinov (approved):
LGTM! Thanks. I left a couple of suggestions, please take a look

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptRepository.kt
toupper: good point, done in 0703601c74f5307170e03d08a1d9d4a8408685d6

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/emailreceipt/WooPosEmailReceiptRepository.kt
toupper: Yes, done in 4a8de225f11e7eb629e5d6f24d7a7c7a654fa771

PR #14821, opened 2025-10-24
[Woo POS][Local Catalog] Consolidate requirements for local catalog support

samiuelson:
WOOMOB-1576

### Description

This PR consolidates the requirements for local catalog support duplicated across different places into a single, centralized check through the new `WooPosIsLocalCatalogSupported` class. This refactoring reduces duplication and simplifies the logic across multiple components.

### Test Steps

This is purely a refactoring PR. It's important to analyze the code to make sure the business logic didn't change and smoke test the POS local catalog syncing behavior.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsLocalCatalogSupported.kt
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        if (!isVariationsEndpointAvailable()) {
+            return false
+        }
+
+        if (!prefsRepo.isPeriodicSyncEnabledForSite(siteId)) {
+            return false
+        }
+
```
malinajirka, line 28: 💡 We could consider having the catalog size check here as well - we could have a parameter, something like `forceRefresh`, that would indicate whether we actually check the catalog size or whether we use a cached value. Either case, it's not in scope of this PR, since it'd be more than just refactoring.

malinajirka (approved):
LGTM!

PR #14825, opened 2025-10-25
[Authenticated WebView] Introduce a helper class to handle the navigation

hichamboushaba:
Closes WOOMOB-1584

### Description
This PR introduces `AuthenticatedWebViewLauncher`: a helper class that allows showing the Authenticated WebView when authentication is supported, and if not supported it falls back to a Chrome custom tab, the class will allow us to update all relevant cases reported in WOOMOB-1585 to support using the AuthenticatedWebView without repeating the logic everywhere.

The launcher is based on the idea that using the AuthenticatedWebView should be preferred whenever authentication is supported, but if not, we should prefer using a Chrome custom tab, as there is a high chance the user is already authenticated in their browser.

> [!NOTE]
> I didn't add tests for  `AuthenticatedWebViewLauncher`, as mocking the extension `findNavController` is not straightforward, I think the value we get from the tests don't justify the effort.

### Test Steps
Our goal here is not to test the `CanAutoAuthenticateInWebView` logic, so please override its return value to test the different cases.
1. Open a Product in product details screen.
2. Tap on "View product on store"
3. Repeat 1-2 in a tablet.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jetpack/benefits/JetpackBenefitsViewModel.kt
```diff
@@ -192,6 +192,5 @@ class JetpackBenefitsViewModel @Inject constructor(
         val jetpackStatus: JetpackStatus
     ) : Event()
 
-    data class OpenWpAdminJetpackActivation(val activationUrl: String) : Event()
```
hichamboushaba, line 195: Not related to the changes of the PR, just noticed this is not used when I was researching the WebView usage in the app.

AdamGrzybkowski (approved):
Looks good :shipit:

PR #14828, opened 2025-10-26
[CIAB] Hide Card Reader and TTP options

hichamboushaba:
Closes WOOMOB-1554

### Description
This PR handles the following changes:
- Disables card reader support for CIAB sites.
- Disables TTP support for CIAB sites.
- Hides the In-Person Payments "learn more" button in the payment method selection screen.

### Test Steps
1. Use a phyiscal device (to confirm TTP logic)
2. Open an existing order or create a new one.
3. Tap on Collect Payment.
4. Confirm that the unsupported options are hidden now.

### Images/gif
| Before | After |
| --- | --- |
| Screenshot_20251026_123227 | Screenshot_20251026_143226 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
@@ -2,7 +2,7 @@ package com.woocommerce.android.ciab
 
 enum class CIABAffectedFeature {
     Blaze,
-    Payments,
+    WooPayments,
     WooShippingSplitShipments,
     GroupedProducts,
     VariableProducts,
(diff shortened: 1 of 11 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/taptopay/TapToPayAvailabilityStatus.kt
```diff
@@ -38,6 +42,7 @@ class TapToPayAvailabilityStatus @Inject constructor(
 
     sealed class Result {
         object Available : Result()
+        object Hidden : Result()
```
hichamboushaba, line 45: I included the `Hidden` option instead of creating a new one under `NotAvailable` because we already have logic in `PaymentsHubTapToPayUnavailableHandler` that displays information on why the feature isn't available. Since this logic is only used in the Payments hub, which we hide for CIAB sites, it was simpler to add the new top-level option, and not deal with adding a new branch in `PaymentsHubTapToPayUnavailableHandler`

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
```diff
@@ -2,7 +2,7 @@ package com.woocommerce.android.ciab
 
 enum class CIABAffectedFeature {
     Blaze,
-    Payments,
+    WooPayments,
```
hichamboushaba, line 5: I renamed this for clarity, Payments as a feature is still supported for CIAB sites, we just don't support features that rely on WooPayments (the Payments hub, IPP and TTP)

AdamGrzybkowski (approved):
Works great :shipit:

PR #14831, opened 2025-10-27
Move Product and Variations sync action into a single class

malinajirka:
Do not merge label - target branch needs to be trunk.

### Description

This PR is just preparation for refactoring the WooPosSyncProductsAction and WooPosSyncVariationsAction into a single action. Since moving and refactoring the classes within a single PR could easily get out of hand and would be difficult to review, this PR simply moves the actions into a single file and a shared class. 

### Test Steps

No need to test anything - it's just a simple move.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosSyncAction.kt
```diff
@@ -6,10 +6,41 @@ import kotlinx.coroutines.coroutineScope
...
+        pageSize: Int,
+        maxPages: Int
+    ) = variationsAction.execute(
+        site = site,
+        modifiedAfterGmt = modifiedAfterGmt,
+        pageSize = pageSize,
+        maxPages = maxPages
+    )
+}
+
```
malinajirka, line 43: 📓 All the code within this file below this line was copy-pasted from the separate files.

samiuelson (approved):
LGTM!

PR #14835, opened 2025-10-28
[Mobile Payments] Enable card payments using Stripe gateway in the UK

samiuelson:
# Enable card payments using Stripe gateway in the UK

WOOMOB-1431

## Description

This PR enables Stripe Payment Gateway extension in the UK for In-Person Payments and POS.

⚠️ Do not merge label - to ensure we release this feature in the same version across Android and iOS
💡 The public-facing documentation will be updated once the feature is released.

## Test Steps

1. Card-present payment collection should be tested using UK-based store, having Stripe Paymen Gateway plugin activated:
- Test using the WisePad reader with Order Creation (IPP)
- Test using the WisePad reader with Point of Sale
- Test using Tap to Pay with Order Creation (IPP) – this is phone-only, so no need to check POS
- Test onboarding with one and both plugins active on the site — plugin selector should appear.
2. Link to purchase the card reader — should open the GB site and offer WisePad 3
3. "Learn more" links — should point to Stripe-specific site rather than WooPayments

## Screenshots

N/A

---


kidinov (approved):
I tested the main flows! LGTM

PR #14837, opened 2025-10-28
Prevent crash when creating or adding logs to file

JorgeMucientes:
Fixes: WOOMOB-1449

### Description
Avoid crashing the app when attempting to save logs into the disk and the device doesn't have space left: 
```
android.system.ErrnoException: write failed: ENOSPC (No space left on device)
    at libcore.io.Linux.writeBytes(Linux.java)
    at libcore.io.Linux.write(Linux.java:288)
    at libcore.io.BlockGuardOs.write(BlockGuardOs.java:345)
    at libcore.io.IoBridge.write(IoBridge.java:553)
    at java.io.FileOutputStream.write(FileOutputStream.java:326)
    at kotlin.io.FilesKt__FileReadWriteKt.writeTextImpl(FileReadWrite.kt:174)
    at kotlin.io.FilesKt__FileReadWriteKt.appendText(FileReadWrite.kt:150)
    at kotlin.io.FilesKt__FileReadWriteKt.appendText$default(FileReadWrite.kt:149)
    at com.woocommerce.android.util.crashlogging.EncryptedLogsFileProvider$provide$2.invokeSuspend(EncryptedLogsFileProvider.kt:18)
 ```

### Test Steps
I've tried replicating the bug as well as trying to see if I could make the logcat display `Could not create log file: ...` error log. But after filling up space on the device the app would start crashing for many issues realted to no space left:
```
Error writing Envelope /data/user/0/com.woocommerce.android.dev/cache/sentry/dd5244304f3c4c51 ...
java.io.IOException: write failed: ENOSPC (No space left on device)
	at libcore.io.IoBridge.write(IoBridge.java:651)


OR 

Error writing entry; local database full: android.database.sqlite.SQLiteFullException: database or disk is full
```

Which also validates the [point made](https://linear.app/a8c/issue/WOOMOB-1449/crash-due-to-encrypted-logs-writing#comment-acc5d466) in Linear Issue about disk space issues.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
@@ -12,10 +12,14 @@ class EncryptedLogsFileProvider @Inject constructor(
 ) {
     suspend fun provide(): File {
         return withContext(dispatchers.io) {
-            val logs = wooLog.getCurrentLogEntries().take(LOG_ENTRIES_LIMIT)
-
-            File.createTempFile("log", "").apply {
-                appendText(logs.joinToString("\n"))
+            runCatching {
+                val logs = wooLog.getCurrentLogEntries().take(LOG_ENTRIES_LIMIT)
+                File.createTempFile("log", "").apply {
+                    appendText(logs.joinToString("\n"))
+                }
+            }.getOrElse { exception ->
+                WooLog.e(WooLog.T.UTILS, "Could not create log file: ${exception.message}")
+                File.createTempFile("log_empty_file_due_to_error", "")
             }
         }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt
@@ -10,17 +10,16 @@ class EncryptedLogsFileProvider @Inject constructor(
     private val wooLog: WooLog,
     private val dispatchers: CoroutineDispatchers
 ) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: WooCommerce/src/main/kotlin/com/woocommerce/android/util/crashlogging/EncryptedLogsFileProvider.kt [hunk 1]
```diff
@@ -12,10 +12,14 @@ class EncryptedLogsFileProvider @Inject constructor(
...
-            File.createTempFile("log", "").apply {
-                appendText(logs.joinToString("\n"))
+            runCatching {
+                val logs = wooLog.getCurrentLogEntries().take(LOG_ENTRIES_LIMIT)
+                File.createTempFile("log", "").apply {
+                    appendText(logs.joinToString("\n"))
+                }
+            }.getOrElse { exception ->
+                WooLog.e(WooLog.T.UTILS, "Could not create log file: ${exception.message}")
+                File.createTempFile("log_empty_file_due_to_error", "")
```
hichamboushaba: This might still crash, if you check some reports like this [one](https://a8c.sentry.io/issues/6870361751/events/2e70d238dfac4c51b45d0a56a5a3e368/?project=1459556&referrer=previous-event), you'll see that the crash happens in File.createTempFile call.

I think the best approach here is to make the return type nullable, and ignore the encrypted logs upload when the file is null, WDYT?

JorgeMucientes: Good catch, was unaware that `createTempFile` was also causing crashes. Makes total sense to return null then. Change applied: [b6e83ab](https://github.com/woocommerce/woocommerce-android/pull/14837/commits/b6e83aba0cbc5321d6931f482449875d901606a7)

hichamboushaba (approved):
Looks good, thanks @JorgeMucientes for working on this.

PR #14840, opened 2025-10-28
[Woo POS][Bug] Incorrect `activeSource` non-null checks

samiuelson:
### Description

This PR fixes incorrect checks for `activeSource` being null in `WooPosProductsDataSource`. Before, the code was not really checking if `activeSource` is null, but the result returned by `activeSource.getProductById()` and `activeSource.getVariationById()`.

### Test Steps

Check if the code is correct.

### Images/gif

N/A


kidinov (approved):
Thanks! LGTM

PR #14845, opened 2025-10-28
[CIAB] Fix case where the bookings tab is deleted when filtering products

hichamboushaba:
Closes WOOMOB-1603

### Description
This PR fixes an issue that caused the bookings tab to be deleted when filtering products, the cause is that we were missing some filter options in `WCProductStore` when deciding whether the products table should be cleared or not.

### Test Steps
1. Use a CIAB site.
2. Open the app and confirm the bookings tab is shown.
3. Switch to the products tab.
4. Apply the "Status" filter to an option that doesn't include any bookable product.
5. Confirm the bookings tab is still shown.
6. Apply the "Product Type = Simple" filter.
7. Confirm the bookings tab is still shown.


AdamGrzybkowski (approved):
Good catch! Works great :shipit:

PR #14849, opened 2025-10-29
[WOOMOB-1547] Fix card reader dialog images not showing on tablets

kidinov:
WOOMOB-1547

### Description
On tablets, images were not showing in the card reader payment flow. The issue was caused by the `isDisplaySizeScaleIncreased()` function incorrectly identifying tablets as having increased display size for accessibility purposes.

The function was checking if `densityDpi > DENSITY_DEVICE_STABLE`, which on tablets (densityDpi=419 vs DENSITY_DEVICE_STABLE=400) would return true even when the user hadn't changed any accessibility settings. This caused images to be hidden via `getIllustrationVisibilityForAccessibility()`.

The fix changes the check to require a significant increase (>15%) in density rather than any increase, so it only hides images when users have actually increased display size for accessibility.

Changes:
- Modified `isDisplaySizeScaleIncreased()` in `UiHelpers.kt` to calculate percentage increase and require >15% threshold
- This preserves the accessibility feature while fixing the false positive on tablets

### Test Steps
This needs to be tested on both phones and tablets:

1. Open card reader payment flow on a phone → images should be visible
2. Open card reader payment flow on a tablet → images should be visible (this is the fix)
3. Increase display size in accessibility settings → images should be hidden (accessibility feature should still work)
4. Verify all affected screens show images correctly: onboarding, connection, payment, tutorial, and update dialogs

### Images/gif

https://github.com/user-attachments/assets/a796dad9-4f32-4864-8afb-dabcb35ac2ca


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/util/UiHelpers.kt
```diff
@@ -87,15 +87,13 @@ object UiHelpers {
...
 
-    fun setImageOrHideInLandscapeOnCompactScreenHeightSizeClass(
+    fun setImageOrHideOnCompactScreenHeightSizeClass(
         imageView: ImageView,
         @DrawableRes resId: Int?,
         setInvisible: Boolean = false
     ) {
-        val isLandscape = DisplayUtils.isLandscape(imageView.context)
-        val isNotCompact = imageView.context.windowHeightSizeClass >= WindowSizeClass.Medium
-        val shouldShowBasedOnOrientationAndSize = !isLandscape || isNotCompact
```
malinajirka, line 97: ❓ @kidinov  One question before I dive into the testing. I noticed the description mentions:
```
On tablets, images were not showing in the card reader connection flow. The issue was in the setImageOrHideInLandscapeOnCompactScreenHeightSizeClass function which was hiding images based on device orientation, even when there was enough vertical space available.
```
However, if I understand the original code correctly, the item will NOT be shown only when the device is in landscape mode AND has a compact size. So I think the issue here isn't that we are checking `landscape` but that the device is considered compact but could still fit the image so we are hiding it unnecessarily. (in other words I would understand if we showed it when we are not supposed to show it [portrait but compact height] with this code)

The part I'm confused with, is that the fix you implemented even works - because, the device should still be evaluated as `compact` and therefore the image should be hidden 🤔 .

kidinov: @malinajirka good catch! The issue actually comes not from here, but from the `isDisplaySizeScaleIncreased` method which checks if there are non-standard display size settings set and hides the images if it's bigger than the standard. But on my device, while being standard it detected it incorrectly, therefore I made it percentages based to hide the image when there is a noticeable difference.

> The part I'm confused with, is that the fix you implemented even works - because, the device should still be evaluated as compact and therefore the image should be hidden 🤔 .

That's because I tested it on an emulator where it worked even before 🫤  not the real tablet where the issue was reproduced

Please review again!

malinajirka (approved):
LGTM, thanks!

PR #14855, opened 2025-10-29
[WOOMOB-1604] Update settings store screen design to I2 specifications

kidinov:
### Description

Updated the WooPos Store Settings screen to match I2 design specifications:

- Replaced icon-based menu items with card-based layout using WooPosCard components
- Added horizontal dividers between setting items for cleaner visual separation
- Updated section titles and styling (Store Information → General)
- Restored shimmer loading states for both Store Information and Receipt sections that follow the new card design pattern
- Adjusted spacing to use Large instead of Medium for bottom padding

### Test Steps

1. Navigate to Woo POS Settings → Store
2. Verify the Store section displays with a card containing store name and address separated by a horizontal divider
3. Verify the Receipt section displays with a card containing 5 fields (name, address, phone, email, refund policy) separated by horizontal dividers
4. Force a loading state and verify shimmer effects appear in both sections with proper card layout and dividers

https://github.com/user-attachments/assets/a27828a3-b604-49b4-975d-d23dd40bd164

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailPaneScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailPaneScreen.kt
@@ -8,8 +8,10 @@ import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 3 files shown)


malinajirka (commented):
Thanks @kidinov!

It seems the content font color is off - at least in dark mode. Seems like [we are using color `outline`](https://github.com/woocommerce/woocommerce-android/blob/5db5be2bf22b5b34bea6d4e7bc3930eb5d077ac4/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailsMenuItem.kt#L69) instead of of `on surface`. I also noticed the value is brighter then the title in the implementation but it is the other way round in the designs. Could you please double check?

Screenshot_20251030_074351

Designs:

Screenshot 2025-10-30 at 7 51 23

kidinov:
@malinajirka 👋 Thank you for the review!

I am a bit confused with your comment. `WooPosSettingsDetailsMenuItem` - is not used in `WooPosSettingsStoreScreen`

The color of the content is `onSurfaceVariantLowest` and it seems to be correct 

image

The screenshot of the design seems like not the design we are now reimplementing. The i2 comes from 

pdfdoF-8el-p2
qKAWGmvUsvfnW0Z4CssJHe-fi-6415_430

malinajirka:
Hmm, this is indeed confusing :D. It seems your screenshot from Figma is from the `experiments` tab, which I assumed aren't final designs. However, the link you shared links to the non-experimental(final?) designs. 😕  Looking at the P2 post, the designs look like this:

Screenshot 2025-10-30 at 10 18 19

> I am a bit confused with your comment. WooPosSettingsDetailsMenuItem - is not used in WooPosSettingsStoreScreen

My bad, I quickly tried to find the code and I apparently found a wrong one.

kidinov:
@malinajirka thank you for catching this! Indeed I was following "experimental" tab 😮‍💨 

Fixed here - c2808c0375c825103d714e869109cc00f29f645e

malinajirka (approved):
LGTM, thanks!

PR #14857, opened 2025-10-29
[WOOMOB-1334] Execute Product and Variation sync in single transaction

malinajirka:
#WOOMOB-1334

Do not merge label - ensure the target branch is trunk before merging.

### Description

This is the final PR in the chain of PRs that aim to ensure that both product and variation sync is done within a single transaction.

This PR is quite big - but the changes are so related that I couldnt' find a way how to split it further. Essentially we are removing WooPosSyncProductAction and WooPosSyncVariationAction and all the logic is done within WooPosSyncAction (since we removed the classes, we also need to remove all their tests and introduce new tests for WooPosSyncAction => that's the main reason for the large PR). 

Considering this feature is still behind a FF and this PR doesn't touch any code outside of Local Catalog, I believe the change is safe to make.

### Test Steps

Verify both initial local catalog sync as well as incremental sync work.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
```diff
@@ -92,63 +92,34 @@ class WooPosLocalCatalogSyncRepository @Inject constructor(
...
-                }
+        successResult.variationsServerDate?.let { serverDate ->
+            syncTimestampManager.parseTimestampFromApi(serverDate)?.let { timestamp ->
+                syncTimestampManager.storeVariationsLastSyncTimestamp(timestamp)
+                logger.d("Stored variations sync timestamp: $serverDate")
             }
         }
 
-        return result
+        val syncDuration = System.currentTimeMillis() - startTime
```
samiuelson: 💡 n.p.: I'm wondering if it would be better to inject the current time via `DateTimeProvider` and replace all 3 `System::currentTimeMillis` calls with `DateTimeProvider::now` already. Currently, it's not a big deal since we are not even asserting the `syncDurationMs` value in tests, but it can avoid making tests non-deterministic by accident in the future.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosSyncAction.kt
```diff
@@ -12,131 +12,107 @@ import org.wordpress.android.fluxc.store.pos.localcatalog.WooPosPaginatedFetchRe
...
+        FetchResults(products, trashProducts, productsServerDate, variations, variationsServerDate)
     }
-}
 
-class WooPosSyncProductsAction @Inject constructor(
-    private val posLocalCatalogStore: WooPosLocalCatalogStore,
-    private val logger: WooPosLogWrapper,
-) {
-    suspend fun execute(
+    private suspend fun executeDatabaseTransaction(
```
samiuelson: 💡 n.p.: The `executeDatabaseTransaction` name is slightly misleading, because it suggests it's some general-use DB transaction executor. Could it be something like `updateDatabaseWithFetchedItems`?

samiuelson (approved):
LGTM! I left a couple of nonblocking comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
malinajirka: Fixed 5c95269e1f00d3251e7abe44ec49151c2ed53c32

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosSyncAction.kt
malinajirka: Good point, fixed in 71ea3aaa96eccdabe11f32a21cdf696f82731596.

PR #14859, opened 2025-10-29
[CIAB Bookings] Update filter screen navigation to use Compose Navigation

hichamboushaba:
Closes WOOMOB-1608

### Description
As discussed and agreed, this PR updates the filters internal navigation to use Compose Navigation, the navigation is still driven by the state of the `ViewModel`, the added `NavHost` and its graph will just reflect the selected page.
As stated, the benefit of this solution is that we'll have a correct lifecycle for the sub-ViewModels we use for the filter subscreens, as the navigation component will use a proper `ViewModelStore` tied to the composable destination.

To better test the solution, I rebased the draft customer filter [PR](https://github.com/woocommerce/woocommerce-android/pull/14834) on this branch, and this shows how we can instantiate the sub-ViewModels, and how we can navigate back.
For communication between the parent ViewModel (`BookingFilterListViewModel`) and other ViewModels, we'll use simple callbacks passed to the subscreen Composables, which then pass them back to their ViewModels. Assisted injection improves this solution, as shown in this [commit](https://github.com/woocommerce/woocommerce-android/pull/14834/commits/8a548c9d5125d8c3da68a03d4f16ffe49bc21ec2). This approach eliminates the need to rely on events for communication.

Please share your thoughts on the suggested approach.

### Test Steps
1. Test the changes in the PR https://github.com/woocommerce/woocommerce-android/pull/14834
2. Open the booking filters screen
3. Test navigation to the date time and customer filter screens (the other items will crash the app)
4. Test selecting a customer and confirm the result is returned

**Important:** For step 4, make sure that you select an actual customer, and not a guest, tapping on guests won't have any effect.

### Images/gif

https://github.com/user-attachments/assets/45ad5763-8935-4fd4-b30a-c89695ebbdaf

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
@@ -137,7 +137,7 @@ private fun FiltersNavHost(
 }
 
 private val BookingFilterPage.route: String
-    get() = this::class.java.simpleName
+    get() = name
 
 private fun slideIn(popNavigation: Boolean): EnterTransition {
     return slideInHorizontally(animationSpec = tween(durationMillis = TRANSITION_DURATION)) { fullWidth ->
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
@@ -1,6 +1,13 @@
 package com.woocommerce.android.ui.bookings.filter
 
 import androidx.activity.compose.BackHandler
+import androidx.compose.animation.EnterTransition
+import androidx.compose.animation.ExitTransition
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
+import androidx.compose.animation.slideInHorizontally
+import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
@@ -92,6 +99,10 @@ private fun FiltersNavHost(
     NavHost(
         navController = navController,
         startDestination = BookingFilterPage.List.route,
+        enterTransition = { slideIn(popNavigation = true) },
+        exitTransition = { slideOut(popNavigation = true) },
+        popEnterTransition = { slideIn(popNavigation = false) },
+        popExitTransition = { slideOut(popNavigation = false) },
         modifier = modifier
     ) {
         composable(BookingFilterPage.List.route) {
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
```diff
@@ -66,62 +65,94 @@ fun BookingFilterListScreen(state: BookingFilterListUiState) {
...
+        startDestination = BookingFilterPage.List.route,
+        enterTransition = { slideIn(popNavigation = true) },
+        exitTransition = { slideOut(popNavigation = true) },
+        popEnterTransition = { slideIn(popNavigation = false) },
+        popExitTransition = { slideOut(popNavigation = false) },
+        modifier = modifier
+    ) {
+        composable(BookingFilterPage.List.route) {
+            BookingFilterRootPage(state.items)
+        }
```
AdamGrzybkowski, line 111:
```suggestion
        composable<BookingFilterPage.List> {
            BookingFilterRootPage(state.items)
        }
```

np: We could use the type-safe destinations to build the graph. We need to make those serializable, but I don't think it's a problem here.

AdamGrzybkowski (approved):
Great work! I like it!

>For communication between the parent ViewModel (BookingFilterListViewModel) and other ViewModels, we'll use simple callbacks passed to the subscreen Composables, which then pass them back to their ViewModels. Assisted injection improves this solution, as shown in this [commit](https://github.com/woocommerce/woocommerce-android/pull/14834/commits/8a548c9d5125d8c3da68a03d4f16ffe49bc21ec2). This approach eliminates the need to rely on events for communication.


I think that's a good idea, I'm okay with events too. One thing to keep in mind is that we will likely need to update the Root ViewModel multiple times without closing the screen, with the multiselect options.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
hichamboushaba: Yes, I thought about it, but given we don't use the serialization plugin on the app, I didn't use, and given we most likely won't need it as the state already exposes all the information that we need, I decided to just use String routes, and they will be static routes.

hichamboushaba:
> One thing to keep in mind is that we will likely need to update the Root ViewModel multiple times without closing the screen, with the multiselect options.

Good point, but I think it's something to discuss later: If we need to follow what you suggested, then we can use two callbacks. But we may not need it if we follow the same approach we use in Order filters today, we save the changes only when the user either taps on "Show orders" or back button, which means we'll close the screen anyway when updating the root ViewModel.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
hichamboushaba: Related to this, I thought about converting `BookingFilterPage` to an enum, but I'm not sure if we'll need to pass data for the items in the future, I think given the root ViewModel state is available, we probably won't need any addiitonal data on the `currentPage` property, WDYT?
If we use an item, we can then just use the `name` property for route, but it's a minor point.

AdamGrzybkowski: Good call. At the moment, I can't think of any data we would pass so I think we should be okay with an enum!

AdamGrzybkowski:
Correct, though the `Show orders` button click is handled in the `BookingFilterListViewModel` so whatever was selected on any of the child pages (while one of those still being visible) has to be already in that VM state as a temp filter (not stored yet).

hichamboushaba:
> Correct, though the Show orders button click is handled in the BookingFilterListViewModel so whatever was selected on any of the child pages (while one of those still being visible) has to be already in that VM state as a temp filter (not stored yet).

Yes, later after I sent the comment I thought about this and figured my mistake, you are right, we'll need two callbacks then. Thanks @AdamGrzybkowski

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/BookingFilterListScreen.kt
hichamboushaba: Done in 6474dc8, I kept using Camel Case for the item names, and kept the `route` extension, this way even if we need to use a `sealed inteface`, it will be an easy change.

PR #14861, opened 2025-10-30
Issue/woomob 1490 predictive back support for android 36

JorgeMucientes:
Closes: WOOMOB-1490

### Description
Last change needed to update project to `targetSdk=36`. Full review here: https://wp.me/p91TBi-dyf

When upgrading the codebase to support the upcoming version of Android 16 (`targetSdk = 36`), among other things the function `override fun onBackPressed()` from any [Activity isn't invoked anymore](https://developer.android.com/about/versions/16/behavior-changes-16#predictive-back). That means that the custom back navigation we heavily rely on in the app is broken. This PR adds the necessary changes for the app to behave the same in new Android version 36 than in prior versions. 

An example of the current problem when `targetSdk=36` with custom back navigation: 

| targetSdk = 35  | targetSdk=36 |
| <video src="https://github.com/user-attachments/assets/0508670b-f8d9-4e69-9e7f-c652b46dfc0a">  | <video src="https://github.com/user-attachments/assets/4a9cfce5-b4f2-44e2-9435-081a27fb44f9">  |

**Context on Custom Back Navigation in the App**
The app overrides the default onBackPressed behavior in multiple places by implementing custom `MainActivity.BackPressListener`: 
```kotlin
        interface BackPressListener {
            fun onRequestAllowBackPress(): Boolean
        }
```
If a Fragment implements `BackPressListener` then when the callback `onBackPressed()` from `MainActivity` (which hosts ~90% of the application screens/fragments) is invoked by the system, then `super.onBackPressed()` is never called because we have an early return: 
```kotlin
    override fun onBackPressed() {
        WooLog.d(WooLog.T.UTILS, "Predictive back: Legacy OnBackPressed called")
        AnalyticsTracker.trackBackPressed(this)

        getActiveChildFragment()?.let { fragment ->
            if (fragment is BackPressListener && !(fragment as BackPressListener).onRequestAllowBackPress()) {
                return
            }
        }

        supportFragmentManager.primaryNavigationFragment?.let { fragment ->
            updateAppBarVisibility(fragment)
        }
        super.onBackPressed()
    }
```

So this PR, migrates the above logic to a `OnBackPressedCallback`always. 

### Test Steps
Have an Android device with Android API 36 and another with API < 36 and go through the following steps: 
1. Open the app
2. Navigate to product details
3. Tap price
4. Edit the value
5. Swipe back to trigger back navigation and see how keyboard is hidden
6. Swipe back to trigger back navigation and see how you land on product detail screen with the price updated
7. Swipe back to trigger back navigation and see how a confirmation dialog to discard changes is shown. 
8. Open product detail again. 
9. Swipe back to navigate back and see it works as expected
10. Exit product detail screen (no need to save changes). 
11. Swipe back from product list and see how you are taken to dashboard tab
12. Swipe back one last time to close the app. 

In both versions the app should have the exact same behavior.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
@@ -378,7 +378,7 @@ class MainActivity :
     }
 
     private fun handleOnBackNavigationCompat() {
-        if (VERSION.SDK_INT >= 33) {
+        if (VERSION.SDK_INT >= VERSION_CODES.BAKLAVA) {
             onBackPressedDispatcher
                 .addCallback(this) {
                     AnalyticsTracker.trackBackPressed(this@MainActivity)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
@@ -391,7 +391,7 @@ class MainActivity :
                         updateAppBarVisibility(fragment)
                     }
                     if (isAtNavigationRoot()) {
-                        remove()
+                        remove() //Remove callback to avoid infinite recursion
                         onBackPressedDispatcher.onBackPressed()
                     }
                 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
@@ -307,7 +307,7 @@ class MainActivity :
 
     override fun onCreate(savedInstanceState: Bundle?) {
         installSplashScreen()
-        handleOnBackNavigationCompat()
+        setOnBackNavigationCallback()
 
         super.onCreate(savedInstanceState)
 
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes:
Moving this to draft as I need to double check a new thing I found. Well get  it back to "ready" shortly

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -306,6 +307,7 @@ class MainActivity :
 
     override fun onCreate(savedInstanceState: Bundle?) {
         installSplashScreen()
+        setOnBackNavigationCallback()
```
AdamGrzybkowski: In the [docs](https://developer.android.com/guide/navigation/navigation-custom-back) the callback is added after `super.onCreate`. Is there any particular reason you've added it before?

AdamGrzybkowski (commented):
Are there other critical parts of the app that should be tested given the targetSdk update here? Just trying to be extra safe.

Also, the targetSdk change could impact POS. Should someone from the Peacock team be tagged as well?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
JorgeMucientes: Good eye!
No particular reason tbh. But to stay on the safe side I'll do it after the super.onCreate() as the sample code does 👍🏼. 
In theory either way would be safe though: 
> Callbacks are added using the addCallback methods. We recommend using the [addCallback()](https://developer.android.com/reference/androidx/activity/OnBackPressedDispatcher#addCallback(androidx.activity.OnBackPressedCallback)) method, which takes a [LifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/LifecycleOwner). This ensures that the OnBackPressedCallback is only added when the LifecycleOwner is [Lifecycle.State.STARTED](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.State#STARTED).

JorgeMucientes:
> Are there other critical parts of the app that should be tested given the targetSdk update here? Just trying to be extra safe.

The goal of this PR is to replicate the logic we previously had inside `onbackPressed()` but using the `onBackPressedDispatcher` instead. So any Fragment that currently implements `BackPressListener` will be impacted by this change. However the logic applied to all `BackPressListener` implementations is the same, so technically testing a feature that covers all possible cases for the `OnBackPressedCallback` added should be fine. That's why I described in the testing instructions 4 different scenarios: 
- On back pressed triggered on a screen implementing `BackPressListener` where `onRequestAllowBackPress` returns `true`
- On back pressed triggered on a screen implementing `BackPressListener` where `onRequestAllowBackPress` returns `false`
- On back pressed triggered from a top level screen that doesn't implement `BackPressListener`
- On back pressed triggered from applications root fragment.

Afaik those 4 cases should cover all posible scenarios of the custom back handling. Hopefully I'm not missing anything 🤞🏼 

> Also, the targetSdk change could impact POS. Should someone from the Peacock team be tagged as well?

Yes that's a good point. Friendly ping to @malinajirka is you could take a look at this PR sometime next week 🙏🏼.

malinajirka:
The code LGTM and I also tested the POS and it seems to work as expected. 🥳

AdamGrzybkowski (approved):
:shipit:

PR #14863, opened 2025-10-30
[Woo POS][Historical Orders] Order Details: Analytics

toupper:
WOOMOB-1155

## Description

Analytics events: pdfdoF-7XF-p2

- `orders_menu_item_tapped` - Orders menu item tapped (no properties)
- `orders_list_loaded` - Orders view loaded (no properties)
- `orders_list_pull_to_refresh` - Pull to refresh triggered (no properties)
- `orders_list_fetched` - Orders fetched completion with `milliseconds_since_request_sent`
- `orders_list_next_page_loaded` - Next page loaded with `page_number`
- `orders_list_row_tapped` - Order row tapped with `order_id`, `order_status`, `list_position` (0 for topmost), `days_since_created` (0 for today, 1 for yesterday)
- `orders_list_search_button_tapped` - Search button tapped (no properties)
- `orders_list_search_results_fetched` - Search results fetched completion with `milliseconds_since_request_sent`
- `order_details_loaded` - Order details loaded with `order_id`, `order_status`, `days_since_created` (0 for today, 1 for yesterday)
- `order_details_email_receipt_tapped` - Email receipt tapped (no properties)

Events registering are coming.

## Steps to reproduce

1. Launch the app and POS
2. Select Menu -> Orders
3. Navigate the flow in various ways to confirm the behavior:
   - View orders list
   - Pull to refresh
   - Paginate
   - Search for orders and paginate
   - Tap on order rows
   - View order details
   - Tap the email receipt button

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

kidinov (commented):
Thanks @toupper ! Overal looks good to me, but I'd propose a few changes with the main goal to keep it simplier


[Implement_WooPosOrdersAnalyticsTracker_for_enhanced_order_analytics_tracking.patch](https://github.com/user-attachments/files/23240745/Implement_WooPosOrdersAnalyticsTracker_for_enhanced_order_analytics_tracking.patch)


Proposed Changes: Extract Tracking Logic from WooPosOrdersViewModel

  1. Add tracking fields to OrderItemViewState, to avoid separate "caching"

```
  // WooPosOrdersState.kt
  data class OrderItemViewState(
      ...
      val statusSlug: String,
      val createdAtMillis: Long
  )
```

  2. Create WooPosOrdersAnalyticsTracker class

  New file with all tracking methods as suspend functions:
  - trackOrdersListFetched(elapsedMs)
  - trackOrdersListRowTapped(orderId, orderStatus, listPosition, createdAtMillis)
  - trackOrderDetailsLoaded(orderId, orderStatus, createdAtMillis)
  - etc.

  3. Remove from ViewModel:

  - OrderTrackingMeta data class
  - trackingMeta mutable map
  - cacheOrderTrackingMeta() method
  - retrieveOrderTrackingData() helper
  - trackOrderDetailsShown() and trackOrdersListRowTapped() private methods
  - 9 analytics event imports + MILLIS_PER_DAY

  4. Update ViewModel:

  - Inject WooPosOrdersAnalyticsTracker instead of WooPosAnalyticsTracker
  - Replace all direct analyticsTracker.track() calls with ordersAnalyticsTracker.track*() in viewModelScope.launch
  - Track OrderDetailsLoaded directly in onOrderSelected() (no separate UI callback needed)

  Why This Is Better:

  - Separation of concerns: Tracking logic separated from state management
  - No mutable cache: Tracking data lives naturally in the view state
  - Cleaner: Removes ~60 lines from ViewModel
  - Testable: Analytics logic can be tested in isolation
  - Maintainable: Changes to tracking don't touch ViewModel and don't make it even bigger

Wdyt?

toupper:
Thanks for the patch @kidinov! I added it here a8dec0332f8d11580c0a7ed375a8a4b8c9ab151d

One question — as in the patch, I initially added the tracking data into the state, but after some research, I found that it’s generally better to keep only what’s strictly needed for the UI there. In this case, memory usage would be the same since we store it in another structure, but keeping it out of the state avoids unnecessary recompositions and keeps the state cleaner and more focused on UI data. What do you think — do we have a convention for handling this kind of case?

toupper:
Also for this:

`Track OrderDetailsLoaded directly in onOrderSelected() (no separate UI callback needed)`

toupper:
@kidinov Also about this one

> Track OrderDetailsLoaded directly in onOrderSelected() (no separate UI callback needed)

If we couple the `OrderDetailsLoaded` event to `onOrderSelected`, we’ll miss the initial auto-selected case (first item on load/refresh), and we’re not matching the event’s intent. Selection is an intention; loading details is an outcome. A bug could select an order without actually rendering/loading its details. I know that second case shouldn’t happen with our current flow, but it’s still a leaky proxy. What do you think?

kidinov (approved):
@toupper 

> In this case, memory usage would be the same since we store it in another structure, but keeping it out of the state avoids unnecessary recompositions and keeps the state cleaner and more focused on UI data. What do you think — do we have a convention for handling this kind of case?

That's all correct, except in this case you do the changes only when the state that changes the UI changes as well, so no extra recompositions unless I am missing something. All this should be handled by having all the orders we show in cache, and then we would take the order data lazily, when it's needed to be tracked, from the cache. I hope we'll implement this in I2 of this project alongside the replacement of the refetching of the order after email sent and/or refund made.

> If we couple the OrderDetailsLoaded event to onOrderSelected, we’ll miss the initial auto-selected case (first item on load/refresh), and we’re not matching the event’s intent. Selection is an intention; loading details is an outcome. A bug could select an order without actually rendering/loading its details. I know that second case shouldn’t happen with our current flow, but it’s still a leaky proxy. What do you think?

Yep, you are right. My intention was to simplify it by not passing it via the UI layer, but indeed it's not ideal. The workaround, which I don't like but avoids having this round trip, is here. So apparently no ideal solution here - up to you
[Track_order_details_loading_in_WooPosOrdersViewModel.patch](https://github.com/user-attachments/files/23256916/Track_order_details_loading_in_WooPosOrdersViewModel.patch)

Also, I noticed that we now track `pos_orders_list_fetched` and `pos_orders_list_search_results_fetched` on success, but as peer description it's about request completion - check the patch out

[Track_elapsed_time_for_orders_list_search_and_fetch_operations.patch](https://github.com/user-attachments/files/23256967/Track_elapsed_time_for_orders_list_search_and_fetch_operations.patch)


Other than that, LGTM!

PR #14864, opened 2025-10-30
[Woo POS][Local Catalog] Skip manual one-time sync on site switch if it's not needed

samiuelson:
WOOMOB-1327

### Description

This PR refactors the triggerManualFullCatalogSync() method to check sync requirements before triggering a full catalog sync, preventing unnecessary sync operations when the catalog is already up to date or when the local catalog is disabled.

### Test Steps

The feature can be tested by switching back and forth between sites and observing if the full sync is performed.

### Images/gif

N/A


malinajirka (commented):
Thanks for working on this @samiuelson!

I'd like to propose an alternative approach and hear your thoughts about it. I'm thinking it might be easier to follow the code, if we had these checks within the worker instead of having them in the scheduler for two reasons:
1. Single source of truth - skipping sync because the data are up to date is similar to skipping it because POS is inactive, and this logic currently lives in the worker itself.
2. Skipping sync when the check sync requirements fails feels off to me - I believe we should retry it in such case. The worker handles this scenario already, since we run `retry()` on an unexpected error. However, this PR doesn't handle this scenario I believe.


No matter which approach we decide to go with, I have a couple more thoughts:
1. We skip when sync is `NotRequired` => however, `NotRequired` is returned for any site that hasn't been synced in 0-6.999 days. Is this intentional or should we run it even if it has been run more than 24 hours ago?
2. `logger.d("Manual full sync skipped: Catalog is up to date")` feels slightly misleading, I'd suggesting updating the log to `Catalog was updated on xyz.`. Wdyt?

samiuelson:
Thank you for thorough review, @malinajirka!

> I'd like to propose an alternative approach and hear your thoughts about it. I'm thinking it might be easier to follow the code, if we had these checks within the worker instead of having them in the scheduler for two reasons:
> 
> 1. Single source of truth - skipping sync because the data are up to date is similar to skipping it because POS is inactive, and this logic currently lives in the worker itself.
> 2. Skipping sync when the check sync requirements fails feels off to me - I believe we should retry it in such case. The worker handles this scenario already, since we run `retry()` on an unexpected error. However, this PR doesn't handle this scenario I believe.

I agree. I addressed this in: 9f2a174

> No matter which approach we decide to go with, I have a couple more thoughts:
> 
> 1. We skip when sync is `NotRequired` => however, `NotRequired` is returned for any site that hasn't been synced in 0-6.999 days. Is this intentional or should we run it even if it has been run more than 24 hours ago?

IMO this is a bug. I addressed it in a separate PR on top of this PR's feature branch: https://github.com/woocommerce/woocommerce-android/pull/14894. Please let me know if you agree.

> 2. `logger.d("Manual full sync skipped: Catalog is up to date")` feels slightly misleading, I'd suggesting updating the log to `Catalog was updated on xyz.`. Wdyt?

I improved this within 9f2a174 as well.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -47,6 +48,8 @@ constructor(
             return Result.success()
         }
 
+        checkSyncRequirement()?.let { return it }
+
         val site = isCatalogSyncSupported() ?: return Result.failure()
```
malinajirka: Is this check still needed when we are using the checker? Seems like there is some duplication in the status checks, wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
```diff
@@ -84,6 +87,33 @@ constructor(
         }
     }
 
+    @Suppress("ReturnCount")
+    private suspend fun checkSyncRequirement(): Result? {
```
malinajirka: It'd be nice to add unit tests for this logic, wdyt?

malinajirka (approved):
Thanks @samiuelson! It works as expected, I've left a couple non-blocking comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
samiuelson: Added: 12d1651

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWorker.kt
samiuelson: Refactored in f0726ab and f95953e

PR #14874, opened 2025-10-31
Fix Interac Refund Flow

malinajirka:
#WOOMOB-1623

### Description

This is a tricky one. The issue is that when the navigation component r[eturns result from RefundConfirmationDialog](https://github.com/woocommerce/woocommerce-android/blob/b4cd467f5d401fa42ad9cc6189d66ab7d1d44ed6/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundConfirmationDialog.kt#L16) to RefundSummaryFragment, it [first returns the result ](https://github.com/woocommerce/woocommerce-android/blob/0758f76139f12a340e0689268aab01c412541675/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/FragmentExt.kt#L49)and later on [performs the actual navigation to the RefundSummaryFragment](https://github.com/woocommerce/woocommerce-android/blob/0758f76139f12a340e0689268aab01c412541675/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/FragmentExt.kt#L51). However, as [the app processes the result it attempts to navigate to `action_refundSummaryFragment_to_cardReaderFlow`](https://github.com/woocommerce/woocommerce-android/blob/3ed5374398cfecfbbab76c4b312efd95df05236c/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/RefundSummaryViewModel.kt#L136) but since the navigation component is still on RefundConfirmationDialog the navigation doesn't happen since action [`action_refundSummaryFragment_to_cardReaderFlow` is not found for RefundConfirmationDialog](https://github.com/woocommerce/woocommerce-android/blob/da5b5f69135facfbabc242a318d0a03e434e541c/WooCommerce/src/main/res/navigation/nav_graph_refunds.xml#L15).

The proper fix would be to move [this line ](https://github.com/woocommerce/woocommerce-android/blob/0758f76139f12a340e0689268aab01c412541675/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/FragmentExt.kt#L49) to the bottom of the method. However, that's quite a risky change since it changes order throughout the app and we can't be sure it doesn't break anything.

Therefore, I opted for a suboptimal solution that feels significantly safer.

P.S. This issue was likely introduced here when we moved the action from dialog to summary fragment - https://github.com/woocommerce/woocommerce-android/pull/14083.

### Test Steps

1. Open an order that was paid with Interac card (canadian store)
2. Tap on Refund Order
3. Select some items in the order
4. Continue
5. Confirm Refund and notice the card reader flow starts (You won't be able to complete the flow without interac card, but just checking it starts is enough to verify the fix works)

### Images/gif

https://github.com/user-attachments/assets/c39e7fbd-8a7b-462e-b8bc-1e105d7790e7


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

kidinov (approved):
I can not test it as I don't have the card, but the explanation and the code make sense to me

kidinov:
@malinajirka I tested it with a simulator - works well!

PR #14880, opened 2025-10-31
[WOOMOB-1575] Fix race condition and indefinite caching in WooPosGetTotalProductCount

kidinov:
WOOMOB-1575

### Description

Fixed two issues in `WooPosGetTotalProductCount`:

**1. Race condition causing duplicate network calls**

The mutex was positioned incorrectly - it wrapped only the fetch operation, not the cache check. This meant that multiple concurrent calls could all check the cache (finding it null), then queue up at the mutex and execute the fetch sequentially. Result: multiple unnecessary network requests for the same data.

**What was happening:**
- Thread 1: checks cache (null) → waits for mutex → fetches → updates cache
- Thread 2: checks cache (null) → waits for mutex → fetches again (duplicate!)
- Thread 3: checks cache (null) → waits for mutex → fetches again (duplicate!)

**The fix:**
Moved the mutex to wrap both the cache validation AND the fetch operation. Now only the first thread fetches, and subsequent threads get the cached value immediately.

**2. Indefinite cache with no expiration**

The cached product count persisted forever (until app restart), which could show stale data to users if product inventory changed.

**The fix:**
- Implemented 1-hour cache expiration using timestamp tracking
- Cache automatically refreshes after expiration
- Introduced `CurrentTimeProvider` interface for testable time handling

### Test Steps

1. Open Woo POS and navigate to product search
2. Verify product count displays correctly
3. Monitor logs to confirm only one network call is made on initial load
4. Test cache behavior remains consistent across searches

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosGetTotalProductCount.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosGetTotalProductCount.kt
@@ -14,25 +14,39 @@ import javax.inject.Singleton
 class WooPosGetTotalProductCount @Inject constructor(
     private val productStore: WCProductStore,
     private val selectedSite: SelectedSite,
+    private val currentTime: CurrentTimeProvider,
     private val dispatchers: CoroutineDispatchers
 ) {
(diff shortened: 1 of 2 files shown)


malinajirka (approved):
Nice catch! Thanks for fixing this 🙇

PR #14886, opened 2025-11-03
[WOOMOB-1622] Handle Bluetooth pairing removed disconnect reason

kidinov:
WOOMOB-1622

### Description
In Stripe SDK v4.7.0, a new disconnect reason `BLUETOOTH_PEER_REMOVED_PAIRING_INFORMATION` was added to provide better visibility into mobile reader disconnects. This PR adds handling for this disconnect reason by:

- Adding a new error code `BLUETOOTH_PEER_REMOVED_PAIRING` to `CardReaderStatus.NotConnected.ErrorCode`
- Mapping the SDK disconnect reason to our error code in `BluetoothReaderListenerImpl`
- Displaying the message "Forget the reader in Android Bluetooth Settings" to users when this error occurs, as recommended by Stripe's documentation

### Test Steps
I don't know how to reproduce this scenario, as it requires the Bluetooth pairing information to be removed by the peer device. Therefore, only code validation is possible for this change.

The AI says:

```
  This is likely a rare edge case that happens when:
  - The card reader's firmware encounters an error and clears pairings
  - The reader runs out of memory for storing pairings
  - A reader firmware bug causes pairing loss
  - Someone accidentally triggers a pairing reset on the reader
  - The reader battery dies completely and loses volatile pairing memory
```

To validate the code:
1. Review that the new error code is properly added to the enum
2. Verify the mapping in `BluetoothReaderListenerImpl.onDisconnect()`
3. Confirm the error message mapping in `CardReaderConnectViewModel`
4. Check that the user-facing message matches Stripe's recommendation

### Images/gif
N/A - Error handling only

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,7 +3,7 @@
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 23.7
 -----
-
+- [Internal] Handle BLUETOOTH_PEER_REMOVED_PAIRING_INFORMATION disconnect reason from Stripe SDK [https://github.com/woocommerce/woocommerce-android/pull/14886]
 
 23.6
 -----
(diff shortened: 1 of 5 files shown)


samiuelson (approved):
LGTM!

PR #14891, opened 2025-11-04
[WOOMOB-1630] Fix bottom padding on the booking details screen

AdamGrzybkowski:
WOOMOB-1630

### Description

When navigating back from the Booking Note fragment via the Toolbar back button, the `innerInsets` from the Scaffold would have non-zero bottom padding. This is wrong because the bottom padding should be 0. Because we don't have a bottom bar in the Scaffold, we can ignore the bottom padding and only apply the top. This fixes the issue locally, but the general problem remains. 

The general issue (in my opinion) is caused by [MainActivityEdgeToEdgeHelper](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityEdgeToEdgeHelper.kt). 

There's a timing issue in consuming the insets there. On top of that, we can't properly handle edge-to-edge. The above code applies the padding to the root, which means each screen doesn't have access to the part of the screen. That is a separate issue, though. 

### Test Steps

1. Go to booking details
2. Scroll down
3. Open the note
4. Tap on the back button in the Toolbar 
5. Confirm the padding at the bottom is correct

### Images/gif

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/4c459b1b-1517-485b-bebe-026eac18c8a3"/> | <video src="https://github.com/user-attachments/assets/01455d77-d5b9-4efc-bbab-081aa98573b5"/> |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
@@ -96,46 +96,49 @@ fun BookingDetailsScreen(
             )
         }
     ) { innerPadding ->
-        when (viewState) {
-            BookingDetailsViewState.Empty -> BookingDetailsEmptyScreen()
-            is BookingDetailsViewState.ShowBooking -> {
-                val state = rememberPullToRefreshState()
-                WCPullToRefreshBox(
-                    isRefreshing = viewState.loadingState == BookingDetailsLoadingState.Refreshing,
-                    onRefresh = viewState.onRefresh,
-                    state = state,
-                    modifier = Modifier
-                        .fillMaxSize()
-                        .padding(innerPadding)
-                        .detailsPanePadding(),
-                ) {
-                    Column(
+        Box(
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Thanks for addressing this @AdamGrzybkowski. Changes look good and work well. 

As per the general edge to edge issue I understand it is not ideal, but fixing it would require a significant refactor that so far we haven't been able to prioritise. I'm not sure how we should proceed here, as I believe removing `MainActivityEdgeToEdgeHelper` implies refactoring all screens at once to delegate the edge to edge handling. So basically, we'd have a branch open until all screens have been migrated before we can merge?

AdamGrzybkowski:
>So basically, we'd have a branch open until all screens have been migrated before we can merge?

Yeah, that's how it looks at the moment, which is not ideal.

PR #14892, opened 2025-11-04
[POS Orders] Performance Improvements Discussion

toupper:
### ⚙️ Context

This **Draft PR** is not intended to be merged — it serves as a space for collaboration to find the best long-term solution.

During testing, we noticed that the **orders cache was not displaying results immediately**. After investigation, we found that mapping the order models from the backend to the corresponding view models in the UI was taking a considerable amount of time:

`replaceOrders: buildItemsMap() took 5955ms for 25 orders
`

### 🔍 Findings

The main bottleneck was identified in `WooPosFormatPrice`, where site settings were being fetched on **every single call**.  
After introducing currency-code caching, the mapping time improved significantly:

`replaceOrders: buildItemsMap() took 2568ms for 25 orders
`
### 💡 Discussion

Caching the currency code introduces a potential downside — if the site settings change (for example, switching to another site), the cached value could become outdated.  

However, since `WooPosFormatPrice` is **not a singleton**, it will be deallocated together with its parent object, so the cache should remain consistent within the object’s lifecycle.  

An alternative could be to **inject the currency code directly**, though that comes with its own disadvantages (e.g., tighter coupling and less flexibility).

### ❓ Questions

@kidinov — what do you think about this caching approach?  
Do you see other opportunities to reduce the mapping time based on our measurements?

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.ui.woopos.orders
 
+import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.woocommerce.android.AppUrls
@@ -29,6 +30,7 @@ import kotlinx.coroutines.launch
 import java.math.BigDecimal
 import java.util.Locale
 import javax.inject.Inject
+import kotlin.system.measureTimeMillis
 import kotlin.time.TimeSource.Monotonic
 
 @HiltViewModel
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kidinov:
@toupper Thank you for your invistigation!

I think the fundamental issue here is that we calculate all the order details even if we may never show them. So maybe the simplest solution for now will be to perform the lazy calculation. Having said that, in-memory cache of the currency is something that we should do as well, so perhaps a ticket to a backlog? Another thing to remember is that with the local catalog, we are going to hit the DB for every order, so lazy calculation is needed anyway.

I also added mapping in parallel. 

Wdyt?

[Performance_optimisations.patch](https://github.com/user-attachments/files/23338220/Performance_optimisations.patch)

PR #14893, opened 2025-11-04
[POS Orders] Filter draft orders out

toupper:
### Description

With this PR, we filter orders with draft statuses from the orders list, as they don't need to be shown to the users. 

### Test Steps

- Open POS
- Add a product to the cart
- Tap on checkout
- Tap on back
- Tap on checkout
- Complete the order
- Open Order List – Before: there is one order in auto-draft and one completed. Now: Just one completed

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
@@ -122,7 +122,7 @@ class WooPosOrdersDataSource @Inject constructor(
         page = page,
         orderBy = OrderBy.DATE,
         sortOrder = OrderRestClient.SortOrder.DESCENDING,
-        statusFilter = null,
+        statusFilter = "pending,processing,on-hold,completed,cancelled,refunded,failed",
         createdVia = "pos-rest-api",
         searchQuery = searchQuery
     )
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -122,7 +122,7 @@ class WooPosOrdersDataSource @Inject constructor(
         page = page,
         orderBy = OrderBy.DATE,
         sortOrder = OrderRestClient.SortOrder.DESCENDING,
-        statusFilter = null,
+        statusFilter = "pending,processing,on-hold,completed,cancelled,refunded,failed",
```
kidinov, line 125: Keep in mind that this approach won’t work if there are any custom statuses or if a new status is introduced. Because of that reason, Povilas decided not to filter out this way the checkout-draft status.

Also, can it be that auto-draft not returned by default in the new versions of Woo? I don't see them. I checked the source code with Ai and it says:


```
Copilot said: Auto-draft orders are excluded because WooCommerce registers
Auto-draft orders are excluded because WooCommerce registers that post‑status as non‑public (not exposed to REST) and the DraftOrders code/hooks explicitly remove draft/auto‑draft from order queries, so the orders list API omits them by default.
```

malinajirka: You are onto something, notice [this comment](https://github.com/woocommerce/woocommerce-android/blob/39acb49a0ec66289e282dc06716536a0138dc5a2/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDaoDecorator.kt#L48):
```
 // Draft orders are not returned from the API. We need to re-fetch order list from the API
// when the order is new or its status changed from draft to another status.
```

malinajirka: Okay. I just tested on WooCommerce 10.2.0 and auto-drafts are not returned there. So this is apparently a change in the behavior of the API! I think we'll need to track whether it's intentional or not and either fix it or update the apps (not just POS, but even Order list).

toupper:
This is fixed in the backend, closing this PR

PR #14894, opened 2025-11-04
[Woo POS][Local Catalog] Refine sync status checker states

samiuelson:
💡 Don't merge until parent branch is `trunk`

### Description

This pull request refactors the sync requirement logic to introduce a more granular distinction between "overdue" and "required but not yet overdue" sync states. It replaces the previous `Overdue` state with a new `NonBlockingRequired` state that includes an `isOverdue` boolean flag, allowing the system to differentiate between syncs that are simply due (between 2-7 days) versus those that are overdue (>7 days).

**Key changes:**
* Introduced a DateTimeProvider dependency and fixed timestamp in tests to eliminate flaky time-dependent test behavior
* Replaced `WooPosFullSyncRequirement.Overdue` with `NonBlockingRequired(lastSyncTimestamp, isOverdue)` to support intermediate sync states
* Added a new `FULL_SYNC_NOT_REQUIRED_THRESHOLD` (2 days) to complement the existing `FULL_SYNC_OVERDUE_THRESHOLD` (7 days)

### Test Steps

N/A

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
@@ -18,7 +18,8 @@ class WooPosFullSyncStatusChecker @Inject constructor(
     private val prefsRepo: WooPosPreferencesRepository,
     private val checkCatalogSizeAction: WooPosCheckCatalogSizeAction,
     private val isLocalCatalogSupported: WooPosIsLocalCatalogSupported,
-    private val wooPosLogWrapper: WooPosLogWrapper
+    private val wooPosLogWrapper: WooPosLogWrapper,
+    private val time: DateTimeProvider,
 ) {
     @Suppress("ReturnCount")
     suspend fun checkSyncRequirement(): WooPosFullSyncRequirement {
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
```diff
@@ -54,43 +55,45 @@ class WooPosFullSyncStatusChecker @Inject constructor(
...
-        val currentTime = System.currentTimeMillis()
-        val timeSinceLastSync = currentTime - lastSyncTimestamp
-        val overdueThreshold = FULL_SYNC_OVERDUE_THRESHOLD.inWholeMilliseconds
-        return timeSinceLastSync >= overdueThreshold
-    }
-
     companion object {
-        private val FULL_SYNC_OVERDUE_THRESHOLD = 7.days
+        private val FULL_SYNC_OVERDUE_THRESHOLD = 7.days.inWholeMilliseconds
+        private val FULL_SYNC_NOT_REQUIRED_THRESHOLD = 2.days.inWholeMilliseconds
```
malinajirka: 💬  Two notes here
1. According to the test plan, we should run periodic full sync every 24 hours. So we should either update the test plan and sync with iOS folks or change this to one-ish day.
2. If we set it exactly to 1 day, it won't work either -> the worker might be run some time before the exact 24 hours, that's up to to the system to determine when to run the worker => I'd set it to 23 hours.

malinajirka (approved):
Thanks @samiuelson ! The approach looks solid, I've left a note about the threshold.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFullSyncStatusChecker.kt
samiuelson: Good catch, I changed `FULL_SYNC_NOT_REQUIRED_THRESHOLD` to 23 hours: 309cbbc

PR #14899, opened 2025-11-04
Issue/woomob 1075 woo poslocal catalog implement local search for products v2 step 1

malinajirka:
Fixes WOOMOB-1075

### Description
This PR introduces the foundational search infrastructure for WooPos local catalog functionality. It adds the ability to search products stored locally by name, SKU, or global unique ID, providing the groundwork for faster product lookup in POS scenarios.

**Key Changes:**
- Add `searchProducts` method to `WooPosProductsDao` with SQL query supporting name, SKU, and global unique ID search
- Add `searchProducts` method to `WooPosLocalCatalogStore` with pagination support
- Introduce `WooPosLocalCatalogSearchDataSource` - dedicated data source for local product search operations
- Comprehensive test coverage for the new search data source

This is part 1 of a 2-part implementation. Part 2 will integrate this infrastructure with existing UI components.

### Test Steps
Green CI is enough since the code isn't used

### Images/gif
N/A - Infrastructure changes only

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/search/WooPosProductsSearchInDbDataSource.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/search/WooPosProductsSearchInDbDataSource.kt
@@ -62,7 +62,7 @@ class WooPosProductsSearchInDbDataSource @Inject constructor(
             onSuccess = { products ->
                 val mappedProducts = products.map { entity ->
                     productMapper.fromEntity(entity)
-                }.sortedBy { it.name.lowercase() }
+                }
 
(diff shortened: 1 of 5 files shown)


samiuelson (approved):
LGTM, @malinajirka! The only thing I'm not sure if is necessary is the pagination. I think it would be fine to show just ~20 results that match search query. Wdyt?

malinajirka:
@samiuelson Thanks for the review!

> The only thing I'm not sure if is necessary is the pagination. I think it would be fine to show just ~20 results that match search query. Wdyt?

I was thinking the same, but then I tested by typing "Bronze" on my store and found out I have 45 items called Bronze xyz => so I ended up keeping the pagination. We can use the same approach for the products list which should significantly speed up the performance.

PR #14903, opened 2025-11-05
[WOOMOB-1645] Add WooPosShimmerText component with automatic text measurement

kidinov:
WOOMOB-1645

@malinajirka @samiuelson @toupper FYI, it's better to use this, newly added, component when we need to display loading text

### Description
This PR introduces a new `WooPosShimmerText` component that automatically measures text dimensions to create accurately sized shimmer loading placeholders. Previously, developers had to manually calculate text dimensions using `WooPosShimmerBox`, which was error-prone and led to inconsistent shimmer sizes.

The component:
- Uses `rememberTextMeasurer()` to automatically calculate the exact width and height of text based on style and font weight
- Automatically determines appropriate corner radius based on the calculated height
- Accepts an optional `fontWeight` parameter to ensure accurate measurement for bold text
- Reduces code duplication by replacing manual shimmer box calculations across the codebase

Updated the following screens to use `WooPosShimmerText`:
- Settings screens (Store and Local Catalog)
- Items list
- Cart screen
- Orders list and details

### Test Steps
1. Open Woo POS and navigate to various screens with loading states
2. Observe that shimmer placeholders accurately match the size of the actual text that appears after loading
3. Verify shimmer sizes for both regular and bold text (e.g., product names, order titles)
4. Check that corner radius is appropriate for different text sizes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka (approved):
Thanks @kidinov!

I have never seen this approach. TBH I'm not entirely sure the goal is to match the shimmer size with the loading size 1:1, I always assumed the idea is for the shimmer effect to clearly suggest it's loading and roughly meet the structure while keeping a visual appealing look. So I'd suggest consulting this with Wagner.

Overall, it works as expected. The only screen which felt a bit off is the Local Catalog screen:
image
image
Wdyt?

kidinov:
@malinajirka Thanks for the review!

> I always assumed the idea is for the shimmer effect to clearly suggest it's loading and roughly meet the structure while keeping a visual appealing look. So I'd suggest consulting this with Wagner.

Could you please elaborate on this? There are no design changes - the idea is that we won't try to manually mimic the size of the shimmer boxes (keep in mind that the current approach doesn't work when the font changes) to avoid jumping of the UI between "loading" and "content" states in different parts of the app. How is this related to Wagner?

> The only screen which felt a bit off is the Local Catalog screen:

What specifically is off on this one? Width smaller than you would expect - I used "size" as a reference for the top line and I guess "just now" for the bottom

malinajirka:
> There are no design changes - the idea is that we won't try to manually mimic the size of the shimmer boxes

The shimmer boxes have a different size than before. Example:

| Before | After |
| -- | -- |
| image | image |



> What specifically is off on this one? Width smaller than you would expect - I used "size" as a reference for the top line and I guess "just now" for the bottom

Yep, the size of the boxes feels quite small - I think wider boxes look visually better even if they are bigger than the text that replaces them.

> the idea is that we won't try to manually mimic the size of the shimmer boxes

I'm not sure if we did that before - I think we usually mimick the overall structure of the screen and choose width of the boxes that is visually appealing (even if it has a different size than the final text).

kidinov:
> The shimmer boxes have a different size than before. Example:
> Yep, the size of the boxes feels quite small - I think wider boxes look visually better even if they are bigger than the text that replaces them.



Used another texts here

b04a159f9fd4093d69e089844a36658b3356c384

image

> I'm not sure if we did that before - I think we usually mimick the overall structure of the screen and choose width of the boxes that is visually appealing (even if it has a different size than the final text).

The problem is largely not in the width, but in the height. The issue is that before we were doing it manually, and if we did a bad job matching the sizes, then the whole UI structure would jump, and there is no nice transition from loading to content state

PR #14905, opened 2025-11-05
[POS Orders] Improve loading by doing it lazily

toupper:
### Description

With this PR we improve the loading times of the POS Orders screen by loading them lazily. See more about it here https://github.com/woocommerce/woocommerce-android/pull/14892

### Test Steps

1. Go to POS
2. Open Orders
3. First time, it takes a while to load because it is fetching remotely
4. Go back to POS Homescreen and back to Orders
5. This time, we had the orders cached, so it should show almost immediately.
6. Select different orders to check that they are properly loaded.

### Images/gif

https://github.com/user-attachments/assets/8f1ff8c5-b6c6-4d67-a34b-5fad076290a6


kidinov (approved):
LGTM!

PR #14913, opened 2025-11-06
[Woo POS][Surveys] Improve back navigation in feedback survey

kidinov:
WOOMOB-1474

### Description
Previously, pressing the system back button while viewing a feedback survey would immediately exit the survey, even if the user had navigated through multiple survey pages. This caused frustration when users accidentally pressed back and lost their progress.

This change adds intelligent back navigation that:
- Navigates through webview history when the back button is pressed
- Only exits the survey when the webview is at the initial page
- Keeps the toolbar close button (X) behavior unchanged - it still exits immediately

The implementation uses OnBackPressedCallback to intercept back button presses and checks webView.canGoBack() to determine whether to navigate within the webview or exit the fragment.

### Test Steps
1. Open a feedback survey from any source (e.g., Woo POS surveys)
2. Navigate through multiple pages in the survey
3. Press the system back button
4. Verify it navigates back through survey pages
5. Continue pressing back until at the first page
6. Press back once more and verify it exits the survey
7. Test the toolbar X button - verify it still exits immediately regardless of webview history

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/feedback/FeedbackSurveyFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/feedback/FeedbackSurveyFragment.kt
@@ -7,6 +7,7 @@ import android.webkit.WebResourceRequest
 import android.webkit.WebSettings
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.AppCompatActivity
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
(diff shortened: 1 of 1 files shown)


malinajirka (approved):
Thanks @kidinov!

I found a bug in the back behavior, but I'm not sure how much important it is. I think the PR still fixes the critical issue of users leaving the survey unintentionally.

Notice in the video that I go back, but when I go forth again, it skips the intermediate steps. I guess this might be a limitation of the survey, I'm not sure.

https://github.com/user-attachments/assets/1ce0c56b-57ae-4577-8ce8-81f81bda968c

kidinov:
> Notice in the video that I go back, but when I go forth again, it skips the intermediate steps. I guess this might be a limitation of the survey, I'm not sure.

Yep, looks like they remember that you saw the page and don't show it again

PR #14914, opened 2025-11-06
[Woo POS][Local Catalog] Hide `Local Catalog` setting category if local catalog is not supported

samiuelson:
WOOMOB-1624

### Description

This PR hides "Local catalog" setting category if local catalog not supported.

### Test Steps

Test `POS>Menu>Settings` screen with stores eligible for POS and not eligible (e.g. 1000+ items in catalog).

### Images/gif

—


kidinov (approved):
Looks and tests good!

Just keep in mind if in `WooPosIsLocalCatalogSupported` we'll have any long-running operation, then we need to have a loading state added to the settings, which is a bit off. So maybe consider showing the initial state right away while doing calculations if local catalog is supported, and if it is, then adding it?

samiuelson:
Thanks for review, @kidinov!

> Just keep in mind if in `WooPosIsLocalCatalogSupported` we'll have any long-running operation, then we need to have a loading state added to the settings, which is a bit off. So maybe consider showing the initial state right away while doing calculations if local catalog is supported, and if it is, then adding it?

This is already what we are doing, see the video below with artificial delay added to `WooPosIsLocalCatalogSupported`:

https://github.com/user-attachments/assets/6e55f41f-fd62-4c4b-bb6c-fe6df3b6a4f3

For this specific behavior, we're removing local catalog instantly from the initial state:
```kotlin
private fun createInitialState(): WooPosSettingsCategoriesState {
        return WooPosSettingsCategoriesState(
            categories = WooPosSettingsCategory.entries.filter { it != WooPosSettingsCategory.LOCAL_CATALOG }
        )
    }
```

PR #14918, opened 2025-11-06
[Woo POS] Fix variation lookup in `WooPosProductsDataSource`

samiuelson:
WOOMOB-1573

### Description

This PR removes dependency on `WCProductStore` from `WooPosProductsRemoteDataSource::getVariationById`. Instead of looking for variation in `WCProductStore`, it switches to in-memory variations cache used in POS, and falls back to the rest client in case the cache doesn't hold the variation.

Before, the `getVariationById` was using `WCProductStore` (used by store management) to get variation instance in  POS Totals screen. This was incorrect, because the POS, in case local catalog is disabled, uses in-memory cache to render items, and `WCProductStore` is not really a good fit there because it is not synced in any way with the in-memory variations cache in POS.

### Test Steps

1. Open a store not eligible for local catalog (e.g. 1000+ items)
2. Add variation items to Cart
3. Check out and collect payment

Smoke test POS, focusing on the Totals and Checkout screens.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
```diff
@@ -538,10 +538,30 @@ class WooPosProductsRemoteDataSource @Inject constructor(
             }
```
kidinov, line 538: Is it intentional that we have 2 classes in the same file now? `WooPosProductsInDbDataSource` and `WooPosProductsDataSource`?

Upd: Ah, actually, 3 in 1 file. It’s a bit too big now and hard to navigate. E.g. I search the name of the method and there are 3 and it’s not clear which belongs to which file. So maybe extracting to separate files would make it a bit cleaner.

kidinov: Also, interestingly, `WooPosProductsDataSource` does not extend `WooPosProductsDataSourceInterface`

It would be nice to work with the naming here—I cannot understand what is what now 😃

malinajirka: > Is it intentional that we have 2 classes in the same file now? WooPosProductsInDbDataSource and WooPosProductsDataSource?

Yes, that's intentional, since only the WooPosProductDataSource should be referenced from the outside - the other classes are "private" to the file.

WooPosProductDataSource is the parent - class that decided which approach/strategy the app should be using => via `activeSource` => activeSource is one of the classes extending WooPosProductsDataSourceInterface (either local catalog or in-memory/remote).


> Also, interestingly, WooPosProductsDataSource does not extend WooPosProductsDataSourceInterface

👍 Agree, it originally extended that interface, but after the changes we removed it but didn't rename the class after that.

kidinov: > the other classes are "private" to the file.

Gotcha! They are not private to the file, though, so if that was the idea, maybe to make it private then

Interestingly, I asked Claude regarding this file, and it suggested splitting it as well 

```
Give feedback on this file

This code would benefit significantly from being split into multiple files. Here's my feedback:

Structural Issues
Single File Problem: Having ~600 lines with 3 major classes and interfaces in one file makes it hard to navigate and maintain. Each class has distinct responsibilities and should live in its own file.

Recommended Split:

products/
  ├── WooPosProductsDataSource.kt
  ├── WooPosProductsInDbDataSource.kt
  ├── WooPosProductsRemoteDataSource.kt
  └── WooPosProductsDataSourceInterface.kt
```

kidinov (approved):
The change LGTM! I just didn’t understand how that worked before 🤔  We didn't put in the DB what's fetched, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsDataSource.kt
malinajirka: They are not private since it wouldn't be possible to have tests for them. I mean, we can split it into multiple files - but IMO it will make it more difficult to get oriented in rather than easier. Right now, it's clear what is the parent and what are the "children", if we put each class into a separate file, it won't be clear. Having said that, if you both believe it's better to split them, I don't mind.

samiuelson:
> The change LGTM! I just didn’t understand how that worked before 🤔 We didn't put in the DB what's fetched, right?

Previously, we were relying on the database cache used in store management which was a **bug**. In a happy path scenario, it was not visible (all variations fetched in a db), however, this approach was potentially prone to errors, e.g. if the user triggers PTR in store management; and is not reliable because POS and store management data cache is not synced.

PR #14922, opened 2025-11-07
[WOOMOB-1470] Booking Filtering - Date&Time picker

AdamGrzybkowski:
Part of WOOMOB-1470

### Description

This PR introduces the UI for selecting dates and times in the booking filter. There's a small change to the Figma. Both time and date pickers are shown in a dialog, not in the layout directly as this is what Material guidelines suggest to do. 

I wasn't sure what to do when the user decides to pick time first (before the date), so at the moment we will use `Clock.now()`, if the date is null, but the time is selected. 

This PR is getting big already, so it **doesn't**:
- Emit the selected date to the root screen in any way - meaning you can filter the list yet.
- Do any validation on the dates - it's a free form for the user.
- Allows for clearing the selection. There is no UI for it, so it's under discussion.

### Test Steps

1. Launch the app
2. Go to bookings tab
3. Go to All tab
4. Tap Filters
5. Tap Date & time
6. Pick a date, confirm the selected value is shown in a later reused in the dialog when changing. 

### Images/gif

https://github.com/user-attachments/assets/75fe6b30-8e9c-4a37-aaeb-9d328006dfa6


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/TimePickerDialog.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    AlertDialog(
+        onDismissRequest = onDismissRequest,
+        dismissButton = {
+            TextButton(onClick = { onDismissRequest() }) {
+                Text(text = stringResource(id = android.R.string.cancel))
+            }
+        },
+        confirmButton = {
+            TextButton(
```
irfano: In `DatePickerDialog`, we use `WCTextButton` for both `dismissButton` and `confirmButton`, which is why the button labels appear in all caps. For consistency, we can apply the same styling here as well. The patch shows my suggestion:
[Use_WCTextButton_in_TimePickerDialog__Replaces_the_standard_Material_`TextButton`_with_the.patch](https://github.com/user-attachments/files/23417659/Use_WCTextButton_in_TimePickerDialog__Replaces_the_standard_Material_.TextButton._with_the.patch)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/datetime/DateTimeFilterPage.kt
```diff
@@ -0,0 +1,212 @@
...
+                showTimeDialog = false
+            },
+            onDismissRequest = { showTimeDialog = false },
+        )
+    }
+}
+
+@Composable
+private fun DateTimeSection(
+    label: String,
```
irfano: np: We’ve named this `header` on other screens. If it makes sense here as well, consider renaming this to `header`. To me, `FROM` and `TO` are headers, `Date` and `Time` are labels.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2808,36 +2807,14 @@
...
-
-
-
-
-
-
-
-
 
     <!-- Z -->
```
irfano: It looks like `// C` and `<!-- Z -->` were left over from a cleaning PR. Let's remove them here as well.

irfano (changes requested):
Screenshot_20251107_154329

I’m not sure why, but the time picker UI looks broken on the Pixel 9 Pro (API 36.1) emulator. I checked the Calendar app on the same emulator and it looks fine.
np: Also, can we update the theme to change pink to purple.

---
On Date & time page, both the headers and rows share the same black background color. In contrast, on the booking details screen, they have different background colors, which creates a better visual distinction. I’ve attached screenshots from both screens for comparison. Let’s update the Date & time screen to match other screens.

Additionally, we should align the text styling. For headers, we can use `BookingSectionHeader`. We can also rename `BookingDetailsLabel` to `BookingLabel` and reuse it for the `Date` and `Time` texts.

|Light|Dark|
|-|-|
|||
|||

AdamGrzybkowski:
Thanks for the review @irfano!

>I’m not sure why, but the time picker UI looks broken on the Pixel 9 Pro (API 36.1) emulator. I checked the Calendar app on the same emulator and it looks fine.

I'm looking into this, but so far I haven't been able to fix it. The issue appears to be project-specific. When I create a new project, the TimePicker behaves completely fine...

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/filter/datetime/DateTimeFilterPage.kt
AdamGrzybkowski: Fixed in https://github.com/woocommerce/woocommerce-android/pull/14922/commits/0fe24b1a85d24a515ea33e7d02aa7322ced7b23a

WooCommerce/src/main/res/values/strings.xml
AdamGrzybkowski: Done in https://github.com/woocommerce/woocommerce-android/pull/14922/commits/e5d3b30533aaae689bf9c9f1a4cc183c99e76348

AdamGrzybkowski:
Ok, found the issue @irfano 

This commit fixed the issue https://github.com/woocommerce/woocommerce-android/pull/14922/commits/50379a1d3d250e87f5c80d34dea60689d8445cfe

The culprit was typography definition in `WooTheme`. The DisplayLarge font style was significantly bigger compared to the default and the TimePicker couldn't handle it. For obvious reasons, I didn't want to change the Theme, so I fixed this issue locally only. 

Additionally, I moved to the `TimePickerDialog` component internally, but kept the one I created for the API.

AdamGrzybkowski:
>On Date & time page, both the headers and rows share the same black background color. In contrast, on the booking details screen, they have different background colors, which creates a better visual distinction. I’ve attached screenshots from both screens for comparison. Let’s update the Date & time screen to match other screens.
>
>Additionally, we should align the text styling. For headers, we can use BookingSectionHeader. We can also rename BookingDetailsLabel to BookingLabel and reuse it for the Date and Time texts.


This should also be fixed now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/TimePickerDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+        MaterialTheme(
+            colorScheme = MaterialTheme.colorScheme,
+            shapes = MaterialTheme.shapes,
+            typography = reducedDisplayTypography
+        ) {
+            TimePicker(
+                state = timeState,
+                colors = TimePickerDefaults.colors(
+                    clockDialColor = MaterialTheme.colorScheme.surfaceContainerHighest,
+                    periodSelectorSelectedContainerColor = MaterialTheme.colorScheme.primaryContainer,
```
AdamGrzybkowski: >np: Also, can we update the theme to change pink to purple?

I did that. Material guidelines suggest using `tertiaryContainer` for this, but I think our theme is not set up correctly so it doesn't look good with it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/TimePickerDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+        // WooTheme sets very big (compared to defaults) font sizes. Those are not handled well in the TimePicker,
+        // so we have to override those locally.
+        val reducedDisplayTypography = MaterialTheme.typography.copy(
+            displayLarge = MaterialTheme.typography.displayLarge.copy(fontSize = 57.sp),
+        )
+        MaterialTheme(
+            colorScheme = MaterialTheme.colorScheme,
+            shapes = MaterialTheme.shapes,
+            typography = reducedDisplayTypography
+        ) {
```
AdamGrzybkowski: This is where the Theme override happens for this component only.

irfano (approved):
Good job on finding the theme issue. All issues have been fixed now.
I’m approving, but one last point: the background color is okay on the date picker, but it still seems broken to me on the time picker. Can we match these background colors?

|Date|Time|
|-|-|
|||
|||

AdamGrzybkowski:
@irfano I did what I could, but unfortunately there's a [bug](https://issuetracker.google.com/issues/403183883) that was fixed in 1.5.0 (this is not stable yet)

For now, I have disabled the tonal elevation (that was causing the slight tint). This fixed the background for light theme, but the dark one is still different.

| Dark | Light |
|---|---|
| Screenshot_20251114_110748 | Screenshot_20251114_110741 |

irfano:
I thought it would be simple. I wasn’t aware of the library bug. It looks much better now, @AdamGrzybkowski, great job! 👍🏻

PR #14926, opened 2025-11-09
Issue/woomob 1313 use remote feature flag for local catalog

malinajirka:
WOOMOB-1313

### Description

This PR replaces local FF with a remote FF that is currently enabled on debug builds and on Automattic accounts.

### Test Steps

1. Verify POS is not accessible on non-automattic account on release build

### Images/gif

N/A


kidinov (approved):
LGTM!

Just for historical reasons maybe to fix the PR description

> Verify POS is not accessible on non-automattic account on release build


Not POS but POS local catalog

PR #14928, opened 2025-11-09
Improve performance of loading POS product list

malinajirka:
#WOOMOB-1670

### Description

This PR adds in-memory caching of sites currency and moves mapping of models to a bg thread to speed up performance of loading the POS product list. The initial load after opening the app takes a bit longer then follow up loadings. However, this PR improves performance of both initial load as well as follow up loads.

### Test Steps

Smoke test the POS

### Images/gif

Before

https://github.com/user-attachments/assets/c923ecc7-740e-4a9f-a145-0eded6f8e263

After

https://github.com/user-attachments/assets/b1bcffe3-19e7-433a-a7bd-0306b52fac7c

cc @samiuelson since we discussed the performance and seems like this PR fixes the issue.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
@@ -148,7 +148,7 @@ class WooPosProductsViewModel @Inject constructor(
     private fun loadProducts(forceRefreshProducts: Boolean) {
         loadProductsJob?.cancel()
         loadMoreProductsJob?.cancel()
-        loadProductsJob = viewModelScope.launch(dispatchers.computation) {
+        loadProductsJob = viewModelScope.launch(dispatchers.io) {
             _viewState.value = WooPosProductsViewState.Loading()
 
             dataSource.fetchFirstPage(forceRefresh = forceRefreshProducts).collect { result ->
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
@@ -21,6 +21,7 @@ import com.woocommerce.android.ui.woopos.util.analytics.WooPosAnalyticsEvent.Eve
 import com.woocommerce.android.ui.woopos.util.analytics.WooPosAnalyticsEventConstant
 import com.woocommerce.android.ui.woopos.util.analytics.WooPosAnalyticsTracker
 import com.woocommerce.android.ui.woopos.util.format.WooPosFormatPrice
+import com.woocommerce.android.util.CoroutineDispatchers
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/format/WooPosFormatPrice.kt
```diff
@@ -1,23 +1,16 @@
...
-    private val wooCommerceStore: WooCommerceStore,
-    private val selectedSite: SelectedSite,
+    private val getCachedStoreCurrency: WooPosGetCachedStoreCurrency,
 ) {
     suspend operator fun invoke(price: BigDecimal?): String {
-        val currencyCode = withContext(Dispatchers.IO) {
-            wooCommerceStore.getSiteSettings(selectedSite.get())?.currencyCode
-        }
-        return PriceUtils.formatCurrency(price, currencyCode, currencyFormatter)
+        return PriceUtils.formatCurrency(price, getCachedStoreCurrency(), currencyFormatter)
```
kidinov, line 14: @toupper jfyi

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
```diff
@@ -146,7 +148,7 @@ class WooPosProductsViewModel @Inject constructor(
     private fun loadProducts(forceRefreshProducts: Boolean) {
         loadProductsJob?.cancel()
         loadMoreProductsJob?.cancel()
-        loadProductsJob = viewModelScope.launch {
+        loadProductsJob = viewModelScope.launch(dispatchers.computation) {
```
kidinov: Do you think computation makes sense here? I don't see any CPU-intensive operations here. If you do, then maybe we need to do the same for the fetching of the next pages as well

kidinov (approved):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/products/WooPosProductsViewModel.kt
malinajirka: You are right, I switched it to IO - bb143c989bf05dc6e964f870c8fae6b11c24dd9a

kidinov: IO may also not be necessary as all the actual IO work is already on this dispatcher inside of the data sources.

malinajirka: Ohh, I switched it to a non-main dispatcher because there was an obvious lag on the UI. 

I don't think it's necessary for the next page, since we load only 25 items there, however, here we load all products - this will be improved as soon as we add pagination to the DB layer. I attempted to do that on Wednesday, but it's not straightforward since the existing implementation assumes we fetch items from cache + remote, so making it work with observable database is trickier than it seemed at first sight.

PR #14934, opened 2025-11-11
[WOOMOB-1675] Remove page limit for incremental local catalog sync

malinajirka:
Fixes WOOMOB-1675
Fixes WOOMOB-1681 

### Description
This PR removes the artificial page limit for incremental local catalog sync in the WooCommerce POS feature. Previously, incremental syncs were limited to 3 pages (300 items), which could cause issues when there were more modified products. Now the incremental sync can fetch all modified products without any page limit.

Related slack convo p1762784153019609-slack-C070SJRA8DP

The changes also move the incremental sync trigger from POS Home screen to the Product List screen and rename the sync reason from `ON_POS_HOME` to `ON_PRODUCTS_TAB` to better reflect where the sync is initiated. This is needed since HOME VM is initialized before the Splash screen, and we don't want to run incremental sync before the Splash screen verifies whether a full sync is needed or not.

Last but not least, this PR also modifies the check for availability of /variations endpoint - we fetch woo version if it's not already cached, to ensure the catalog sync is performed after login.

P.S. @samiuelson I usually try to avoid assigning PRs to folks who just came from AFK, however, it changes the core syncing logic and I might be missing some context.

### Test Steps
1. Clean apps data
2. Login
3. Notice the catalog sync is scheduled and starts
4. When it starts open the POS tab
5. Notice the splash screen "connects" to the ongoing background sync and listens for progress. Notice a standalone incremental sync is NOT performed (it's performed only as part of the full sync)
6. Notice when splash screen disappears and product list is displayed the incremental sync is performed

### Images/gif
N/A - Backend sync logic change


samiuelson (approved):
Thanks, LGTM!

PR #14935, opened 2025-11-11
Fix overlapping texts by pinning to the text bottom

toupper:
### Description

With this quick PR we fix a glitch happening when the product has a long title in POS Orders. With two lines, it overlaps the bottom text of quantity. We fix it by pinning the quantity text to the title, instead of to the image.

### Test Steps

With a long product title and a POS Order containing that product:

- Go to POS Orders
- Check that the title doesn't overlap the quantity text

#### Before
Screenshot_20251111_162453_Woo (Dev)

#### After
Screenshot_20251111_162244_Woo (Dev)


kidinov (approved):
LGTM! I'd suggest to add one long name to `WooPosOrderDetailsPreview` so it's clear in the preview that this case is handled

toupper:
> LGTM! I'd suggest to add one long name to WooPosOrderDetailsPreview so it's clear in the preview that this case is handled


Sure! Done in 65318c06d386221cf344bdacb9e4c9555e4136bf

PR #14940, opened 2025-11-12
[POS Orders] Hide lines with zero value

toupper:
### Description

With this PR we hide those lines from the order lines which value is zero. The function used to compare `BigDecimal` wasn't giving our expected behavior, so fixing that solved the issue.

Other than that, we align with iOS on the Discount texts, now it's "Discount total".

### Test Steps

With an order that doesn't have any shipping or discount:

1. Go to POS
2. Open Orders
3. Check that Order
4. See that shipping or discount are hidden

#### Before

image

#### After

Screenshot_20251112_123830_Woo (Dev)


kidinov (approved):
Thanks! LGTM!

PR #14942, opened 2025-11-13
Support basic auth when fetching nonce with site credentials

JorgeMucientes:
Closes: WOOMOB-1702

### Description

Adds support for basic auth when fetching site nonce using site credentials

### Images/gif

### DIFF
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.network.rest.wpapi
 
 import com.android.volley.NoConnectionError
 import com.android.volley.RequestQueue
+import okhttp3.Credentials
 import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

wzieba:
👋 @JorgeMucientes I know it's only a draft, but just in case if it'll be helpful I'd like to share a few comments:
- HTTP Basic Auth _can_ be authorized with **different** credentials than WP. They are indeed sometimes the same (like in Pressable) but that's not actually a rule/requirement.
- We have a dedicated flow for HTTP Basic Auth with a different screen etc. - maybe we can use it. See:

https://github.com/woocommerce/woocommerce-android/blob/131ff2f5963f784fe8d740ca6842ea0619571081/libs/login/src/main/java/org/wordpress/android/login/LoginSiteAddressFragment.java#L346

and the remaining flow that starts there.

JorgeMucientes:
I'm closing this, as there are more details to be measured and investigated before we move forward with a solution for this: pe5sF9-4LP-p2

PR #14960, opened 2025-11-14
[WOOMOB-1689] Add Exit POS button to catalog sync error screen

malinajirka:
Fixes WOOMOB-1689

**I'm targeting 23.8 - if this PR is merged before the codefreeze, please update the milestone.**

### Description
Added an "Exit POS" button with outline style below the retry button on the "Unable to sync catalog" error screen. This provides users with an additional way to exit POS when catalog sync fails, making the user experience consistent with iOS.

### Test Steps
1. Launch the WooCommerce app
2. Turn on airplane mode
3. Navigate to POS mode
5. Verify that the error screen shows both:
   - A primary "Retry" button
   - A secondary "Exit POS" button with outline style below it
6. Tap the "Exit POS" button
7. Verify that you are taken back out of POS mode

### Images/gif
Screenshot_1763109134


kidinov (approved):
LGTM! But it looks like we show the same error screen even when the "remote" is used - not related to this PR though.

PR #14961, opened 2025-11-14
[WOOMOB-1676] Pass modified_after when fetching trashed products in POS sync

malinajirka:
Fixes WOOMOB-1676

**I'm targeting 23.8, if this is merged before the codefreeze please update the milestone.**

### Description
Updated the POS local catalog sync to pass the `modified_after` parameter when fetching trashed products during incremental syncs. Previously, trashed products were always fetched without a time filter, which was inefficient and inconsistent with how we handle published products.

### Test Steps
2. Open POS
3. Pull to refresh
5. Verify that only recently trashed products are fetched (e.g. using App inspection) 
6. Verify the sync completes successfully

### Images/gif
N/A - Backend change only


samiuelson (approved):
LGTM!

PR #14964, opened 2025-11-14
Run FluxC instrumentation tests on CI

wzieba:
Closes: [AINFRA-1460: Run fluxc-plugin tests as part of the Woo CI pipeline](https://linear.app/a8c/issue/AINFRA-1460/run-fluxc-plugin-tests-as-part-of-the-woo-ci-pipeline)

### Description

This PR introduces running FluxC Plugin (WooCommerce FluxC) instrumented tests on CI builds. Currently, these tests are only about database migrations.

### Test Steps

Green CI is completely fine.

### Images/gif

Screenshot 2025-11-14 at 11 53 14


libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
```diff
@@ -34,21 +34,9 @@ class MigrationTests {
...
-    fun testRoomDBSchemasUnchanged() {
-        val context = InstrumentationRegistry.getInstrumentation().context
-        for (i in 3..WC_DATABASE_VERSION) {
-            val currentHash = MigrationSchemasUtils.getIdentityHash(i, context)
-            val expectedHash = MigrationSchemasUtils.DB_HASHES.getValue(
-                MigrationSchemasUtils.getDBHashKey(i)
-            )
-            assertThat(currentHash).isEqualTo(expectedHash)
-        }
-    }
```
wzieba, line 50: Please see the commit note. If you think this test if actually needed, I'm okay with reverting it and fixing. But as I couldn't find a decent justification myself, I've decided to remove it.

AdamGrzybkowski: I agree, although I don't have the historical context, so I might be missing something.

AdamGrzybkowski (approved):
Thanks!

libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
wzieba: Yeah, I understand. But because this test was not actually ever executed on CI, and the project was fine, I think we can proceed without it. If we'll see a need for it in the future, we can always revert 👍

PR #14969, opened 2025-11-14
[Woo POS][Local Catalog] Overdue sync banner UI update

samiuelson:
### Description

This PR updates UI of Items screen:
1. Removes weird internal, horizontal spacing in `WooPosItemList` hardcoded to 2.dp
2. Aligns padding of overdue banner with items list
3. Increases clickable area around "x" buttonin banner
4. Fixes bug in `ItemListPreview` preview

### Test Steps

1. Open POS and trigger displaying sync overdue banner (e.g. by modifying `WooPosFullSyncStatusChecker` logic)
2. Verify banner's UI

### Images/gif

Screenshot_20251114_125130


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsList.kt
```diff
@@ -583,7 +582,7 @@ fun ItemListPreview() {
                         expiredState = Coupon.ExpiredState.NotExpired
                     ),
                     Coupon(
-                        id = 4,
+                        id = 5,
```
samiuelson, line 586: 💡 This is a fix for Composable preview rendering issue, not related to this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsList.kt
```diff
@@ -77,7 +77,6 @@ fun WooPosItemList(
     WooPosLazyColumn(
         modifier = modifier,
         verticalArrangement = Arrangement.spacedBy(WooPosSpacing.Small.value),
-        contentPadding = PaddingValues(2.dp),
```
kidinov, line 80: It's not horizontal but vertical spacing, isn't? And I think we need it, as without that the cards shadow overlaps with the next card and generally spacing between the cards is too small

kidinov (commented):
I think the banner still looks off.

* There is still inconsistency with the color and the size of the cross compared to other "crosses" in the POS. 

Check, for instance,

`WooPosExitConfirmationDialog`

```
            IconButton(
                onClick = { onDismissRequest() },
                modifier = modifier
                    .align(Alignment.TopEnd)
            ) {
                Icon(
                    Icons.Default.Close,
                    contentDescription = stringResource(
                        id = R.string.woopos_exit_dialog_confirmation_close_content_description
                    ),
                    modifier = modifier
                        .size(40.dp),
                    tint = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.6f)
                )
            }
```

* Also, all the titles in the are bold, while here it is not (in Figma it's also bold)
* The title is not aligned with the cross


I'd suggest following Figma design 

FmV0bBXHmsGYZOgUJb0yzD-fi-10012_3962

samiuelson:
> I think the banner still looks off.
> 
> * There is still inconsistency with the color and the size of the cross compared to other "crosses" in the POS.
> * Also, all the titles in the are bold, while here it is not (in Figma it's also bold)
> * The title is not aligned with the cross

Thanks for review, I addressed your comments, and made the layout match Figma specs:
image

Also, instead of removing item list content padding completely, I changed it to vertical (removed removed horizontal padding which is not necessary): d95ca7e, so that banner and items are the same width.

kidinov (changes requested):
@samiuelson  👋

* The cross is too small now (the icon itself has padding, and you set it's size as it's without padding). It's 32 with paddings on the design.
* Distance between the title and the description is more than 8dp, as in the design
* There are double margins for most of the elements, e.g. the cross is 32 to the top and to the left, while it should be 16.

samiuelson:
> @samiuelson 👋
> 
> * The cross is too small now (the icon itself has padding, and you set it's size as it's without padding). It's 32 with paddings on the design.
> * Distance between the title and the description is more than 8dp, as in the design
> * There are double margins for most of the elements, e.g. the cross is 32 to the top and to the left, while it should be 16.

Thanks, @kidinov. I fixed these in abec1b8. Let me know wdyt.

|Before|Now|
|---|---|
|Screenshot_20251118_142900|Screenshot_20251118_142224|

kidinov (approved):
LGTM!

PR #14970, opened 2025-11-14
[WOOMOB-1699] Fix search progress indicator appearing briefly for local catalog

malinajirka:
Fixes WOOMOB-1699

### Description
Fixed an issue where the search input field's progress indicator would briefly appear for 0.1-0.2 seconds when performing searches in POS local catalog mode. Since local catalog searches are instantaneous (no network requests), the progress indicator should never be shown.

The issue was in `WooPosItemsSearchHelper.kt` where the loading state logic always showed the progress indicator when `isLoading` was true, regardless of whether we were using local catalog or remote data source. 

### Test Steps

Local catalog mode:
2. Go to POS → Products tab  
3. Open search and type a query
4. No progress indicator appears in search input field (search results shimmer still works as expected)

Also test with remote mode:
1. Disable local catalog in POS settings
2. Repeat search test - progress indicator should still appear in search input field for remote searches

### Images/gif

https://github.com/user-attachments/assets/58088ac9-c96c-408f-beac-35c76c025e35


samiuelson (approved):
LGTM!

PR #14972, opened 2025-11-14
Introduce `pos` module

wzieba:
### Description

This PR introduces `pos` module. The goal of introducing one is to improve developer experience of the POS by shortening feedback loops.

Code that is required to be used by both `WooCommerce` and `pos` should be placed in `common` module. You can read more about it internally: p91TBi-dCL-p2

### Test Steps

Smoke test the app, review code - especially the part with `WooFileLogger` injection I commented below.

### Images/gif

image


libs/commons/src/main/java/com/woocommerce/android/util/WooLog.kt
```diff
@@ -64,13 +59,13 @@ object WooLog {
 
     private lateinit var fileLogger: WooFileLogger
 
-    fun init(context: Context) {
+    fun init(fileLogger: WooFileLogger) {
```
wzieba, line 62: Extra caution: I'm not sure why `EntryPoints.get(context, FileLoggerEntryPoint::class.java).fileLogger()` was introduced instead of directly passing `WooFileLogger`. It was added in 41e9dc0462c240ffed3bcddd99c1cd845a1c7bf6 but I couldn't find the reason there on in PR. So I assume, this was just a cosmetic choice to use `context` + `EntryPoints`. 

So here I refactored it to make it simpler. This also moves injection logic from the `commons` module, leaving only annotations here.

samiuelson (approved):
LGTM, thanks!

PR #14979, opened 2025-11-14
[AINFRA-1533] Adopt `git-conceal` in this repo

AliSoftware:
Closes: [AINFRA-1533](https://linear.app/a8c/issue/AINFRA-1533/adopt-git-crypt-in-woocommerce-android)

## What

This PR migrates the repo to use [`git-conceal`](https://github.com/Automattic/git-conceal) instead of `configure_apply`, so that the usage of secrets in the repo gets easier and more transparent for developers.

## How

 - 🗑️  Delete `.configure` and `.configure-files/*.enc` files
 - Add files that were listed in `.configure` as encrypted files in the repo instead, by also adding their path as entries in the `.gitattributes` with `filter=git-conceal diff=git-conceal`—so they get encrypted by `git-conceal` on push
 - Update `.gradle` files and `Fastfile` to account for new file paths
 - Update `.gradle` code to account for the fact that now the files are present (but encrypted) after cloning the repo—as opposed to being absent when we didn't run `configure_apply` in the old way.
 - Replace calls to `bundle exec fastlane run configure_apply` in `.buildkite/*` commands—and calls to `configure_apply(force: true)` in `Fastfile`—with calls to `git-conceal-unlock`, and bump `ci-toolkit` plugin to point to https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195 where this helper lives
 - 📚  Updated documentation

> [!NOTE]
> <details><summary>ℹ️ About the <tt>git-conceal-helper</tt> helper from <tt>ci-toolkit</tt></summary>
>
> The way this currently works is that:
> - The `git-conceal` GitHub repo builds new binaries for macOS, Linux and Windows when we do a new release, and attach them as assets to the GitHub Release
> - The `git-conceal` GitHub repo also provides an `install.sh` script to make it easy to install `git-conceal` for your current platform (the script takes care of detecting if you're on a Linux, macOS or Windows agent, downloads the right asset from the latest GitHub release for that platform, `chmod +x` it and install it)
> - The `git-conceal-helper` wrapper from our `a8c-ci-toolkit` plugin is then just a small bootstrap/helper script that checks if `git-conceal` is installed in the CI agent, if so calls `git-conceal unlock "env:${1:-GIT_CONCEAL_SECRET_KEY}"` and exit; otherwise calls the `install.sh` script of `git-conceal` to install it first before calling `git-conceal unlock`
>
> The way this might work in the future:
>  - We will probably ship `git-conceal` in most of our self-hosted CI agents and custom AMIs to avoid having to download it on each job… even if [in practice it takes less than 1s to download _and_ unlock the repo](https://buildkite.com/automattic/woocommerce-android/builds/34200/steps/canvas?sid=019ae633-6050-4b3c-8a44-b71718dd7313#019ae633-60ac-4156-bf46-1a4c815e12ba/316)
>  - We will still keep that wrapper script in `ci-toolkit` that does the "tests if installed already and install it if not" logic, because not all our CI agents are custom agents (e.g. we sometimes use the `default` queue which uses EC2 instances with the official Buildkite AMI, not customized by us), so that would still be useful for those.
</details>

## Merge timing

> [!IMPORTANT]
> Do not merge this PR until (1) we have all the documentation / FAQ ready for all devs to follow in FG and (2) we have validated similar PRs in other repos—especially ones using macOS/Windows and Tumblr CI agents—work too..

### General check

 - Verify CI is green
 - Checkout this branch and validate that **all** the secret files added to the repo in this PR are encrypted and unreadable:
   - `WooCommerce/google-services.json`
   - `WooCommerce/upload.jks`
   - `debug.keystore`
   - `firebase.secrets.json`
   - `google-upload-credentials.json`
   - `secrets.properties`
   - `sentry.properties`
 - Review the documentation updates
 - Review and ensure all the changes necessary in `*.gradle` files to reflect the new setup are covered by this PR (path updates to `.properties` or keystore files, code updates around logic that was based on if a secret file exists or not, …)

### Follow the `README.md` instructions **as an Automattician**

 - Clone the repo from scratch in a **new directory** on your Mac, then checkout this PR's branch
 - Copy the git-conceal encryption key [from our Secret Store](https://mc.a8c.com/secret-store/?secret_id=13697) in your clipboard, then run `pbpaste | base64 -d | git-conceal unlock -` at the root of your new clone's directory
 - Validate that after this step, you are able to read the content of secret files like `secrets.properties` in clear now
 - Build and Run the project, validating everything works as expected (including e.g. Google login to validate `google-services.json` is taken into account, etc)

> [!NOTE]
> <details><summary>ℹ️ The real migration won't require devs to clone the repo from scratch; this is just for the testing steps of this PR</summary>
>
> Above I suggest to test this scenario in a fresh clone of the repo rather than your everyday working copy, especially to avoid risking to leave your git repo in a state that is setup for `git-conceal` while you were just testing and the PR is not merged yet.
>
> In practice, once the PR is merged in `trunk`, devs won't need to clone the repo fresh to migrate to `git-conceal` though. 
They will just have to run `pbpaste | base64 -d | git-conceal unlock -` on their existing working copy once, and probably never run a `git-conceal` command again afterwards (except maybe `git-conceal status` if they want to validate a new secret file they're about to add is properly gonna be encrypted as they expect).
>
> At some point I'd still highly recommend for them also get rid of the old files that were installed by their previous setup with `configure` (`rm -rf ~/.configure/woocommerce-android` and `git clean -dxi -e .DS_Store -e .idea -e local.properties`), to avoid confusion and risking relying on obsolete files.
</details>

### Follow the `README.md` instructions **as an external contributor**

 - Clone the repo from scratch in a **new directory** on your Mac, then checkout this PR's branch
 - Validate files like `secrets.properties` show as encrypted garbage
 - Edit `defaults.properties` to add the relevant values for `wp.oauth.*`
 - Overwrite `WooCommerce/google-services.json` encrypted file with `WooCommerce/google-services.json-example`
 - Build and Run the project, validating everything works as expected (except Google login)


.buildkite/commands/diff-merged-manifest.sh
```diff
@@ -14,9 +14,6 @@ BUILD_VARIANT=$1
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
-echo "--- :closed_lock_with_key: Installing Secrets"
-bundle exec fastlane run configure_apply
-
```
AliSoftware, line 19: Having the repo `git-crypt`-unlocked during the branch dance that is done internally by `comment_with_manifest_diff` [caused `Your local changes to the following files would be overwritten by checkout`issues](https://buildkite.com/automattic/woocommerce-android/builds/33809/steps/canvas?sid=019a89db-dcc4-4129-b6d3-2a6a0d874ae0#019a89db-dd22-4e1a-a12b-3dcf7fd4dbf1/257-4114), especially if the `git-crypt`'d files listed in `.gitattributes` on the HEAD branch are not the same as the ones in the BASE branch[^1]

Since we don't need any secret in practice to generate the manifest and call `process{variant}Manifest`, the solution is simple: just don't bother unlocking the repo's secrets for that task.

---

A better long-term solution to make `comment_with_manifest_diff` more resilient to situations like this would be to make it use `git worktree` instead of switching branches in-place:
1. Generate base manifest: `git worktree add $TMP_DIR_FOR_BASE $BASE_BRANCH && cd $TMP_DIR_FOR_BASE` then run `./gradlew process{variant}Manifest` there
2. Generate head manifest: `cd $CHECKOUT_DIR && rm $TMP_DIR_FOR_BASE && git worktree prune` then run `./gradlew process{variant}Manifest` there

That way each checkout is done in independent folders, eliminating the risk of conflicts during the branch dance.

[^1]: like will be the case during that transition to `git-crypt`, or when we'll add a new secret file, especially if that secret file previously existed unencrypted in the BASE branch as an example file for external contributors I think?

.buildkite/shared-pipeline-vars
```diff
@@ -3,6 +3,7 @@
 # This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
 # to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
 
-export CI_TOOLKIT="automattic/a8c-ci-toolkit#5.4.0"
+# "git-crypt-unlock" branch / https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195
+export CI_TOOLKIT="automattic/a8c-ci-toolkit#0a3f10921096cee57c18ac5667fc64c1aaad4a7d"
```
AliSoftware: 🎗️ TODO: Revert back to pointing to a tag version once https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195 is merged and we have an official new version of the `ci-toolkit`

WooCommerce/build.gradle
```diff
@@ -496,12 +496,13 @@ android.buildTypes.all { buildType ->
     }
 
     // If Google services file doesn't exist, copy example file.
-    if (!file("google-services.json").exists()) {
+    def googleServicesFile = file("google-services.json")
+    if (!googleServicesFile.exists() || isFileEncrypted(googleServicesFile)) {
         tasks.copyGoogleServicesExampleFile.copy()
     }
```
wzieba, line 504: _If_ `google-services.json` is encrypted, we'll copy the example file. Then why do we require to

>    1. `cp WooCommerce/google-services.json-example WooCommerce/google-services.json` (to replace that encrypted file with placeholder content)

in the docs?

.buildkite/commands/gradle-cache-build.sh
```diff
@@ -13,8 +13,8 @@ fi
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
-echo "--- :closed_lock_with_key: Installing Secrets"
-bundle exec fastlane run configure_apply
+echo "--- :closed_lock_with_key: Decrypting Secrets"
+.buildkite/commands/git-crypt-unlock.sh
```
wzieba: If you wish, we can add this command to the pre-build scripts in Gradle (only on CI and with a flag to disable it for `diff-merged-manifest.sh‎`). This way we could remove some these `Decrypting Secrets` steps (excluding cases like `download_release_translations`).

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

docs/project-overview.md
```diff
@@ -35,7 +32,15 @@ Read more about [OAuth2][oauth] and the [WordPress.com REST endpoint][wp-api].
 
 #### `secrets.properties`
 
-The `secrets.properties` file is used to store sensitive information that should not be checked into version control. This file is located at `~/.configure/woocommerce-android/secrets/secrets.properties`.
+The `secrets.properties` file is used to store sensitive information that should not be checked into version control in clear text.
+This file is encrypted (using `git-crypt`), and only developers working at Automattic have the decryption key.
+
+If you are a developer working at Automattic, ensure you followed those instructions once after cloning the repo:
+  1. Make sure you have `git-crypt` installed (`brew install git-crypt`)
+  1. Search for "WooCommerce Android git-crypt encryption key" in our Secret Store, and copy the Base64 value in your clipboard
```
wzieba: nit: I wonder, is it okay to provide a direct link to Secret Store? 🤔

wzieba (commented):
>  ℹ️ The real migration won't require devs to clone the repo from scratch; this is just for the testing steps of this PR

Hm, I don't know why exactly, but  `~/woocommerce-android/.configure-files/firebase.secrets.json` is not respected in `.gitignore` on this branch. 

So, as a result, I could by mistake include the `firebase.secrets.json` in my git commit. To reproduce:

- `gco trunk`
- `bundle exec fastlane run configure_apply`
- `gco AINFRA-882-adopt-git-crypt`
- `git status -u`

AliSoftware:
> > ℹ️ The real migration won't require devs to clone the repo from scratch; this is just for the testing steps of this PR
> 
> Hm, I don't know why exactly, but `~/woocommerce-android/.configure-files/firebase.secrets.json` is not respected in `.gitignore` on this branch.
> 
> So, as a result, I could by mistake include the `firebase.secrets.json` in my git commit. To reproduce:
> 
> * `gco trunk`
> * `bundle exec fastlane run configure_apply`
> * `gco AINFRA-882-adopt-git-crypt`
> * `git status -u`

This is a very good and important point, thanks for raising it!

This PR indeed removed those files from `.gitignore` to clean things up, but in retrospect, we should keep them around especially to avoid legacy files (since deleted in this branch, but that could have been leftovers from switching to old branches) from being accidentally committed indeed 👍

WooCommerce/build.gradle
AliSoftware: Because I updated the docs before first, then later realized I had to update the `.gradle` file to make CI work… and didn't get back to review the docs again after doing the `.gradle` change 😅 

Good point then, we should indeed remove that extra step from the docs.

.buildkite/commands/gradle-cache-build.sh
AliSoftware: Mmmh interesting, that could be a good idea… but I'm also torn becauce I like that it's clear and explicit in the `.buildkite/*` files that we do it and when; while if it's in the pre-build scripts in Gradle, it will be a bit hidden away from the rest of what we do in `.buildkite/commands/*.sh` scripts that is specific to setting up CI.
Especially if we have to add an exception for `diff-merged-manifest.sh`… in the end I fear this is going to make the maintenance of the whole CI setup less clear and more convoluted, especially for when one of us (who might not know—or had forgotten—about it being in the gradle pre-build scripts) might need to do maintenance on the setup?

docs/project-overview.md
AliSoftware: Yeah I hesistated on this too. It's probably OK I guess because it's just a link that won't be accessible by anyone outside Automattic, and those links just contain `?id=<number>` so it doesn't really reveal anything in the link URL either…

.buildkite/commands/gradle-cache-build.sh
wzieba: yeah, there are benefits to being explicit here 👍 let's leave it like this

docs/project-overview.md
wzieba: Yes, I also had second thoughts before suggesting this, but actually our security doesn't rely on obscurity of a url. I think it's fine to share the link here, MC requires multpile security measures anyway to get anything from it.

README.md
```diff
@@ -38,8 +38,14 @@
...
-1. Generate the developer oauth2 tokens. These values get copied into the `~/.configure/woocommerce-android/secrets.properties` file in the next step. See the [OAuth2 Authentication](docs/project-overview.md#oauth2-authentication) section for details.
+1. If you are a developer at Automattic:
+   1. Make sure you have `git-crypt` installed (`brew install git-crypt`)
+   1. Search for "WooCommerce Android git-crypt encryption key" in our Secret Store, and copy the Base64 value in your clipboard
+   1. Run `pbpaste | base64 -d | git-crypt unlock -` to decrypt the encrypted files (including `secrets.properties` and `WooCommerce/google-services.json`)
+1. If you are an external contributor:
+   1. Generate developer OAuth2 tokens. See the [OAuth2 Authentication](docs/project-overview.md#oauth2-authentication) section for details.
+   1. Edit `defaults.properties` and adjust the values as needed—especiallyincluding `wp.oauth.*` ones. See the [Configuration Files](docs/project-overview.md#configuration-files) section for a breakdown of the properties.
```
iangmaia:
```suggestion
   1. Edit `defaults.properties` and adjust the values as needed — especially including `wp.oauth.*` ones. See the [Configuration Files](docs/project-overview.md#configuration-files) section for a breakdown of the properties.
```

.buildkite/commands/prototype-build.sh
```diff
@@ -12,8 +12,8 @@ APP_TO_BUILD="${1?You need to specify the app to build, WooCommerce or WooCommer
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
-echo "--- :closed_lock_with_key: Installing Secrets"
-bundle exec fastlane run configure_apply
+echo "--- :closed_lock_with_key: Decrypting Secrets"
```
iangmaia, line 12: Underrated improvement: now we won't need to start Ruby, initialize the Fastlane runtime to start a lane, etc just for decrypting secrets (admittedly not that slow, but still... 😄) and leave that only once for the main lane.

💭 Which also makes me think we could call `.buildkite/commands/git-crypt-unlock.sh` before `install_gems`. I don't see any big advantages, though, other than code organization and potentially early failure on secrets. Anyway, just wanted to mention it while thinking out loud 🙂

WooCommerce/upload.jks
iangmaia, line 1: 💭  I was wondering if there could be a way to be sure, in reviews, that this has in fact been encrypted or not specially given files like this are binary files. Then I've noticed all `git-crypt` encrypted files start with `GITCRYPT` so perhaps this could be a simple way to check for that in an automated way?

AliSoftware: You mean [like I do here](https://github.com/woocommerce/woocommerce-android/pull/14979/files#diff-ec0b8224e9624f6e960cf80011b3fe65616b30f77b0a1e4ecde4f76c87609793R518-R524)? 😛 

Indeed maybe we can write a Dangermattic plugin to detect which file in the PR are encrypted and add an inline comment on the file if so as an extra information? Is that what you meant?

As for manually testing locally if a file is properly encrypted _before_ pushing a commit to the remote, one can use `git-crypt status -e <file>` directly to check that (or, alternatively, print the raw content of the file with `git show :<file>`—e.g. git show :secrets.properties`, with leading `:` colon—and confirm that row content is some binary garbage starting with `\0GITCRYPT\0`.

iangmaia: > You mean [like I do here](https://github.com/woocommerce/woocommerce-android/pull/14979/files#diff-ec0b8224e9624f6e960cf80011b3fe65616b30f77b0a1e4ecde4f76c87609793R518-R524)? 😛

Ha, yes, I realized that function was doing that after I posted the comment 😂 

> Indeed maybe we can write a Dangermattic plugin to detect which file in the PR are encrypted and add an inline comment on the file if so as an extra information? Is that what you meant?

Yeah, though at the same time I find it a bit difficult to do that in a systemic way that will be useful (as we don't add secrets that often)...
And without a clear pattern on the secret files, it's also not clear how to generalize a check without knowing all files in advance (defeating the purpose of warning when that new file has been added without encryption, there's still some value for updates...) 🤔 

> As for manually testing locally if a file is properly encrypted _before_ pushing a commit to the remote, one can use `git-crypt status -e <file>` directly to check that (or, alternatively, print the raw content of the file with `git show :<file>`—e.g. git show :secrets.properties`, with leading `:`colon—and confirm that row content is some binary garbage starting with`\0GITCRYPT\0`.

👍

WooCommerce/google-services.json
iangmaia, line 1: 💭 Again thinking out loud, not sure if it's worth the trouble: what do you think of grouping encrypted / secret files in the same folder and making a convention out of this for all projects (well, kinda similar to what we had before but making it more obvious)?
The main advantage is added clarity and creating a pattern where we can build things on top (e.g. validation to make sure everything under that folder is encrypted). Of course, it wouldn't completely prevent  mistakes, but it would be a way to keep things clear.

AliSoftware: Yeah, I thought about that.

I think that for some particular secret files, they do need to be at specific locations in the project structure for the compilation to work (e.g. `WooCommerce/google-services.json`, `secrets.properties`). So we won't have much choice on those, it's enforced by Gradle expecting them there.

For others (`upload.jks`, `debug.keystore`, `google-upload-credentials.json`…), their location might be arbitrary and ok to move somewhere else. In that case maybe we can move them to some `.secret-files/` folder just to group them indeed.

The only trick is that if they were already existing in the repo a particular location on disk before that PR, and we move them elsewhere as part of this PR, we'd then need to add their old location back to `.gitignore` to avoid the case that @wzieba [found about above](https://github.com/woocommerce/woocommerce-android/pull/14979#pullrequestreview-3472276727) of someone switching to a branch in which those were in the old location, then switching to this branch while having the leftover of the decrypted file from old branch at the old location if they didn't run `git clean` in-between to remove untracked files… and then risk of commiting those old files and thus leaking secrets accidentally.

I still see the appeal of having secret files grouped nicely in a dedicated folder rather than keeping them at the root of the repo still. But I'm not sure moving them is worth the potential risk (or having to keep legacy entries in `.gitignore` just for that risk) 🤔  WDYT?

AliSoftware:
> > ℹ️ The real migration won't require devs to clone the repo from scratch; this is just for the testing steps of this PR
> 
> Hm, I don't know why exactly, but `~/woocommerce-android/.configure-files/firebase.secrets.json` is not respected in `.gitignore` on this branch.
> 
> So, as a result, I could by mistake include the `firebase.secrets.json` in my git commit. To reproduce:
> 
> * `gco trunk`
> * `bundle exec fastlane run configure_apply`
> * `gco AINFRA-882-adopt-git-crypt`
> * `git status -u`

Addressed in 1754ce8ec2f1425bff6637da3199096b55078097 by re-adding `.configure-files` to `.gitignore` to cover that potential case. @wzieba can you confirm this solves the issue with the previous scenario you tried above?

WooCommerce/build.gradle
AliSoftware: Removed in cf9139caa1915122b4a279716319c482f6d7a450

docs/project-overview.md
AliSoftware: Added link to SecretStore in 63f3291781be86cd26ffdda29100ee32ec21eaa7

.buildkite/commands/prototype-build.sh
AliSoftware: Updated in bc1d18d6c8faafdbfcccd624908cf3c6fc4afd16 🙂

WooCommerce/google-services.json
iangmaia: > But I'm not sure moving them is worth the potential risk 

👍  I'm also leaning towards that. It's also annoying that they can't be in the same place, which would be the ideal scenario and perhaps would make the change worth it.

wzieba:
Fixed, thanks @AliSoftware !

.buildkite/shared-pipeline-vars
AliSoftware, line 1: 🎗️  TODO: Point back to stable version once https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195 has been merged and released

AliSoftware:
@wzieba I have now updated the PR to use [my new `git-conceal` Rust tool](https://github.com/Automattic/git-conceal) instead of `git-crypt` after all.

Can you thus give it another review?

.buildkite/shared-pipeline-vars
```diff
@@ -3,6 +3,6 @@
 # This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
 # to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
 
-export CI_TOOLKIT="automattic/a8c-ci-toolkit#5.4.0"
+export CI_TOOLKIT="automattic/a8c-ci-toolkit#9ab2426db86e5a096938340800de26ecd5f7acfb" # https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195
```
mokagio:
```suggestion
export CI_TOOLKIT="automattic/a8c-ci-toolkit#4f8cad67accd2e5b4286c3def4babbc3c3f0b269" # https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195
```

.buildkite/shared-pipeline-vars
```diff
@@ -3,6 +3,6 @@
 # This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
 # to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
 
-export CI_TOOLKIT="automattic/a8c-ci-toolkit#5.4.0"
+export CI_TOOLKIT="automattic/a8c-ci-toolkit#4f8cad67accd2e5b4286c3def4babbc3c3f0b269" # https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195
```
AliSoftware: Once https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/198 lands and we create the `6.0.0` tag:
```suggestion
export CI_TOOLKIT="automattic/a8c-ci-toolkit#6.0.0
```

wzieba:
Sorry @AliSoftware , I didn't make it to review this today. I'll prioritize it first on Monday.

AliSoftware:
> Sorry @AliSoftware , I didn't make it to review this today. I'll prioritize it first on Monday.

No worries.

Btw, in practice after discussing the timing during my 1:1 with Tanner and given the EOY rush and my upcoming long AFK, I am not even sure I'm going to merge this before January. i.e.
 - I could aim to merge this soon so that devs can start playing with it and be beta testers, reporting any issues or questions they might encounter with the tool… but since I'm not gonna be around past next week, that might be a tough spot to let the teams in
 - So instead I think I'll prepare all the PRs for all the repos before my AFK but not merge any until when I get back in January

I am still interested in your review of course, especially since I've started to work on draft PRs to do similar changes to other Android repos so if you have feedback on the code changes / on the approach in the Gradle code to fit that change, it'd be easier to get them before I start applying the same approach to my other draft PRs… but it's also not critically urgent because none will be merged before January anyway.

---

PS: I'm also interested in having WCAndroid devs playing with this PR and experimenting with the tool—while staying in this PR—so that they can validate, as users, that it's easy to use and works well for them. Though that won't allow them to test it in real situations (like validate it still works when switching branches) until I actually merge it in January—when I'll be more available and around for any needed support—but at least it'd give early user-side feedback too.

settings.gradle
```diff
@@ -23,7 +23,7 @@ ext {
 
 gradle.ext {
     isCi = System.getenv('CI')?.toBoolean() ?: false
-    secretsPath = "${System.getProperty("user.home")}/.configure/woocommerce-android/secrets/secrets.properties"
+    secretsPath = "secrets.properties"
```
wzieba, line 26: This makes secrets visible for LLM agents. I wonder how we could address this. I understand moving secrets out of the repository is not an option as we use git filters. 

We could use something like

```
{
  permissions: {
    deny: [
      "Read(secrets.properties)"
    ]
  }
}
```

for Claude, but our developers use various differnt agents.

On the other hand, maybe it's fine - after all, we don't keep here any secrets that leaking would be a danger to users.

README.md
```diff
@@ -38,8 +38,13 @@
     $ cd woocommerce-android
     ```
 
-1. Copy `defaults.properties` to the secrets directory: `cp defaults.properties ~/.configure/woocommerce-android/secrets/secrets.properties`. See the [Configuration Files](docs/project-overview.md#configuration-files) section for a breakdown of the properties.
-1. Generate the developer oauth2 tokens. These values get copied into the `~/.configure/woocommerce-android/secrets.properties` file in the next step. See the [OAuth2 Authentication](docs/project-overview.md#oauth2-authentication) section for details.
+1. If you are a developer at Automattic:
+   1. Make sure [`git-conceal` is installed](https://github.com/Automattic/git-conceal?tab=readme-ov-file#installation) on your machine
+   1. Open [the "WooCommerce Android git-conceal encryption key" entry in our Secret Store](https://mc.a8c.com/secret-store/?secret_id=13697), and copy the Base64 value in your clipboard
+   1. Run `pbpaste | base64 -d | git-conceal unlock -` to decrypt the encrypted files (including `secrets.properties` and `WooCommerce/google-services.json`)
```
wzieba, line 44: Because our engineers are used to a different process (run `configure_apply` to update secrets) I'd suggest mentioning that `unclock` is a one-time action.


```suggestion
   1. Run `pbpaste | base64 -d | git-conceal unlock -` to decrypt the encrypted files (including `secrets.properties` and `WooCommerce/google-services.json`). Unlike before, you only need to do this **once per clone** - after unlocking, you can read or update secrets as often as needed without repeating this command.
```

wzieba (approved):
Works really nice 👏 !

settings.gradle
AliSoftware: Good point.

I wonder if in a future iteration of `git-conceal` we could maybe make it auto-generate/update ignore files for various agents automatically. We could start by supporting Claude—making `git-conceal` it automatically open `.claude/settings.json` file and add secret files to the `permissions.deny` array (creating that array if it doesn't exist) if they aren't listed there already. And then iterate to support more agents and their corresponding way to declare what files to ignore.

README.md
AliSoftware: While I get the "Unlike before", this will not age well especially for people reading this `README.md` in a few months/years after `configure_apply` will be long forgotten and that "before" won't mean much 😉 

```suggestion
   1. Run `pbpaste | base64 -d | git-conceal unlock -` to decrypt the encrypted files (including `secrets.properties` and `WooCommerce/google-services.json`)
   1. Note: you only need to do this **once per clone**. After unlocking, you can read and update secret files as often as you need, manipulating them as any other non-secret files, without repeating this command—git filters will take care of encrypting/decrypting them on the fly during commit/checkout transparently for you.
```

AliSoftware:
Closing this because:
1. the PR has become stale (and secrets might have changed in mobile-secrets since I opened this PR)
2. And after consulting with SecOps, we ended up concluding that `git-conceal` wouldn't be a good enough solution for various reasons, and we'll need to use a different approach.

PR #14981, opened 2025-11-17
[Woo POS][Local Catalog] When reconciling items in cart, delete products from Local Catalog DB as well

samiuelson:
When a product is deleted from the cart because it no longer exists, it is now also removed from the local catalog database (WooPosLocalCatalogStore) in addition to the WooPosProductsCache.

[WOOMOB-1692](https://linear.app/a8c/issue/WOOMOB-1692)

### Description

This pull request extends the cart item reconciliation functionality to delete products from the Local Catalog database when they no longer exist in the cart. Previously, products were only removed from WooPosProductsCache; now they are also removed from WooPosLocalCatalogStore to maintain consistency across data sources.

* Added `deleteProducts()` and `deleteVariations()` methods to `WooPosLocalCatalogStore` for removing products and variations from the local database
* Refactored `deleteProductFromCache()` in `WooPosCartItemsUpdater` to handle both Simple and Variation products, deleting them from both the cache and Local Catalog DB

### Test Steps

1. Open POS and add simple product to cart
2. Remove that product in wp-admin
3. Go to Checkout in POS
4. Verify the error message is shown informing that the product doesn't exist; tap "Remove product" button
5. Go back to Home and verify that product is no longer in the Catalog

### Images/gif

—


kidinov (approved):
LGTM

PR #14983, opened 2025-11-18
New `bookable-service` product type

AdamGrzybkowski:
WOOMOB-1736

### Description

We need to update the bookable product type from `booking` to `bookable-service`. This PR:
- updates the value we pass to the server when fetching products of type `Booking`
- uses the new value to determine the booking type in Product details
- add another check to determine the booking tab visibility - bookingsCount > 0

### Test Steps

1. Launch the app and login to this test site p1763423349396019-slack-C03L1NF1EA3
2. Confirm you see the bookings tab
3. Go to products and tap on a bookable one
4. Confirm that it's shown in a web view

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductType.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductType.kt
@@ -15,7 +15,7 @@ enum class ProductType(@StringRes val stringResource: Int = 0, val value: String
     BUNDLE(R.string.product_type_bundle, CoreProductType.BUNDLE.value),
     COMPOSITE(R.string.product_type_composite, "composite"),
     VARIATION(R.string.product_type_variation, "variation"),
-    BOOKING(R.string.product_type_booking, "booking"),
+    BOOKING(R.string.product_type_booking, "bookable-service"),
     OTHER;
 
     fun isVariableProduct() = this == VARIABLE || this == VARIABLE_SUBSCRIPTION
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductType.kt
```diff
@@ -15,7 +15,7 @@ enum class ProductType(@StringRes val stringResource: Int = 0, val value: String
     BUNDLE(R.string.product_type_bundle, CoreProductType.BUNDLE.value),
     COMPOSITE(R.string.product_type_composite, "composite"),
     VARIATION(R.string.product_type_variation, "variation"),
-    BOOKING(R.string.product_type_booking, "booking"),
```
JorgeMucientes, line 18: I'm not sure we want to replace this. I think we likely still want to support "booking" type for non-CIAB sites. 
I'd suggest adding a new bookable service type like I did for service/events filter here -> https://github.com/woocommerce/woocommerce-android/commit/64091b268fe284d7ca85480ff4339c1d3a1bb7c6

JorgeMucientes (approved):
Thanks for addressing this so promptly @AdamGrzybkowski. Code looks good and works as expected. I just left some concern that we should agree on.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductType.kt
AdamGrzybkowski: [This type was added very recently, specifically to handle CIAB sites](https://github.com/woocommerce/woocommerce-android/commit/f4adca43ec93ac2659813b3ebaa1967f2e0cf4f9). I'm not sure we need both at the moment, because the `booking` wouldn't be used at all. WDYT?

PR #14990, opened 2025-11-19
[WOOMOB-1712] Auto-scroll to newly added orders in POS Orders screen

kidinov:
WOOMOB-1712

### Description
When users create a new order in POS and return to the Orders screen, the newly added order appears at the top of the list but remains hidden because the list preserves its scroll position. This fix automatically scrolls the list to the top with a smooth animation when new orders are detected, ensuring users can immediately see their newly created order.

The implementation detects when new orders are added by comparing the first order ID before and after the list updates. When a change is detected, a scroll event is emitted and the list animates to position 0 after a brief delay to allow the composition to settle.

### Test Steps
1. Open the POS Orders screen
2. Navigate back and create a new order
3. Return to the Orders screen
4. Verify the list smoothly scrolls to the top, showing the newly created order

https://github.com/user-attachments/assets/39e0416e-e298-4648-86bf-f22ee28aba14

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 23.8
 -----
 - [Internal][Woo POS] Improve detection and handling of deleted products during checkout [https://github.com/woocommerce/woocommerce-android/pull/14981]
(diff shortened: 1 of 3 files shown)


toupper (approved):
LGTM! Thanks for working on this @kidinov!

PR #14993, opened 2025-11-19
[WOOMOB-1731] Fix extra vertical space in the Bookings top bar

AdamGrzybkowski:
WOOMOB-1731

### Description

This issue was likely caused by one of the recent compose library updates, because our code was not modified in the last two weeks at least. 

**What's happening?** 

With the search active, the `SearchSection` component takes the full width of the Toolbar, leaving no space for the Toolbar title. The problem is that this title is still in the UI tree, and because it doesn't fit in one line, it gets squized into multiple lines. This makes the whole Toolbar taller than it should be. 

I've decided to set the Toolbar title to an empty string as this is the most straightforward fix. Alternatively, we could replace the whole Toolbar when the search is active, but that would require a few more changes. 

### Test Steps

1. Go to bookings tab
2. Tap on the search icon

### Images/gif

| Before | After |
|---|---|
| Screenshot_20251119_133345 | Screenshot_20251119_133910 |


JorgeMucientes (approved):
Thanks for the quickfix @AdamGrzybkowski. 
Code looks good and works as expected. Let ship this.

PR #14995, opened 2025-11-19
[POS - Empty copies] Remove em dashes after feedback and update copy.

toupper:
### Description

After the feedback received here, pdfdoF-8lY-p2 we proceed to remove the em dashes from the empty search copy. Other than that, we adjust the copy to the latest designs. NFWCJetWSgHq2jPLKjpKYw-fi-8160_5862

### Test Steps

Go to POS and enter terms that don't provide a result in Products, Coupons, and Orders. Check that the empty search copy doesn't show em dashes.

#### Products
Screenshot_20251119_142028_Woo (Dev)

#### Coupons
Screenshot_20251119_142056_Woo (Dev)

#### Orders
Screenshot_20251119_142129_Woo (Dev)


kidinov (approved):
LGTM!

PR #14996, opened 2025-11-19
[WOOMOB-1750] - Fix zone issues in Date&Time picker 

AdamGrzybkowski:
WOOMOB-1750

### Description

I've noticed that the Date&Time picker had time zones messed up. I thought we should operate in the device's local zone, but that's not true and due to the booking plugin we have to force UTC always. 

This PR fixes that.

Additionally, there's one small fix related to filters - https://github.com/woocommerce/woocommerce-android/pull/14996/commits/2db08d5390aa20133dad2e0b4e24a9c94c2249e4
The All tab always shows an empty view, because we pass the wrong info to the server.

### Test Steps

1. Launch the app
2. Go to bookings tab -> All
3. Open the filters -> Date & time
4. Play with it, especially at around midnight where a different time zone could change the day (see video)
5. Confirm the filters are properly applied and the list shows all results that should be shown

### Images/gif

| Before | After |
|---|---|
|  <video src="https://github.com/user-attachments/assets/6872fcf8-c299-4222-8f0b-fe55dfc3e48b"/> | <video src="https://github.com/user-attachments/assets/abccee6a-7c64-4a9f-8875-9cb938c51b2b"/> |


JorgeMucientes (approved):
Changes look good to me. This not only fixed issues with date/time filter, it also fixes the issue where past bookings are not displayed even if not date time filter is applied.

PR #14997, opened 2025-11-19
Migrate `WCOrderShipmentTrackingModel`

wzieba:
Closes: [AINFRA-556: Migrate `WCOrderShipmentTrackingModel`](https://linear.app/a8c/issue/AINFRA-556/migrate-wcordershipmenttrackingmodel)

### Description

This PR migrates `WCOrderShipmentTrackingModel` from WellSql to Room.

The migration, like always, focuses on actually migrating database entities and _not_ fixing threading issues. All database calls on main thread are now more visible though, because of the `runBlocking` usage.

### Test Steps

1. Check out this PR
2. Install the app, log in to our legacy shipping label flow test site (details internally: p1763551085820349/1763487138.230489-slack-C03L1NF1EA3)
3. Open orders (e.g. id `2174`) and verify you can see tracking numbers under packages and can interact with them (copy, track)
4. Inspect `OrderShipmentTrackingEntity` database table and make sure it all looks correct

#### Before (WellSql)

Screenshot 2025-11-20 at 13 18 40

#### After (Room)

Screenshot 2025-11-20 at 13 00 11


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/model/OrderShipmentTracking.kt
```diff
@@ -16,26 +18,26 @@ data class OrderShipmentTracking(
...
     return OrderShipmentTracking(
-        id,
-        localSiteId,
-        orderId,
-        remoteTrackingId,
-        trackingNumber,
-        trackingProvider,
-        trackingLink,
-        dateShipped
+        id = 0, // No longer using synthetic id
```
AdamGrzybkowski: ❓ It looks like this is always `0`. Can't we remove the `id` property?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
```diff
@@ -259,14 +259,17 @@ class OrderDetailRepository @Inject constructor(
...
+    ): OrderShipmentTracking? = runBlocking {
+        orderStore.getShipmentTrackingByTrackingNumber(
+            selectedSite.get(),
+            orderId,
+            trackingNumber
+        )?.toAppModel()
+    }
 
-    fun getOrderShipmentTrackings(orderId: Long) =
+    fun getOrderShipmentTrackings(orderId: Long) = runBlocking {
```
AdamGrzybkowski, line 270: Me wanting to leave a comment, because it's only a few `suspend` keywords here and there
restraint
I know, I know.

AdamGrzybkowski (commented):
Hey @wzieba 

Code changes look good, but I'm a bit unclear on how to test that properly. I could verify steps suggested in the description but I also noticed that the same table holds tracking numbers for different orders (with different tracking UI). I thought this only affects the old flow, but that doesn't seem to be the case. 🤔 

I'm not familiar with either flow, so maybe I'm confusing something. 

I will try to find some info about this before approving.

AdamGrzybkowski:
Hi @wzieba 

~I was testing this with @irfano and we noticed the `Creating shipping label` button is gone when using the old flow. Notice the missing purple button.~

| Trunk | This branch |
|---|---|
| Screenshot 2025-11-26 at 16 10 18 | Screenshot 2025-11-26 at 16 06 03 |

~Given that this PR is on top of some others, it's not clear where the issue is originating from, but it requires some investigation to at least point to the problematic PR.~

I checked `trunk` (button visible), then the base branch for this PR (button visible) and back to this PR and the button was visible. Not sure what happened, but it could be an issue with me activating/deactivating the plugin? IDK 😞 

Anyway, the above doesn't seem to be an issue.

AdamGrzybkowski (approved):
LGTM, but if you want to feel extra safe I would encorage to ask for one extra review from someone more familiar with the flow that uses those models.

WooCommerce/src/main/kotlin/com/woocommerce/android/model/OrderShipmentTracking.kt
wzieba: Good point! Fixed in 154de99d0cf2a2ea946214f490f29baf73ef0d46

wzieba:
Thanks for the review @AdamGrzybkowski !

> LGTM, but if you want to feel extra safe I would encorage to ask for one extra review from someone more familiar with the flow that uses those models.

I think we're ok with the change: we both tested it, and the migration was quite straightforward, without modifying logic  of the flow (except of upserting instead of ignoring, but it's expected for cache). I think it's pretty safe to merge.

PR #14999, opened 2025-11-20
Fix serviceEvents filter condition

AdamGrzybkowski:
### Description

All tab shows an empty page without filters set. This is because we pass an empty array of products due to wrong check. This PR fixes that. 

### Test Steps

1. Launch the app
2. Go to booking tab -> All
3. Clear filters
4. Confirm you see results

### Images/gif


JorgeMucientes (approved):
LGTM Thanks!

PR #15000, opened 2025-11-20
Issue/woomob 1732 customer note missing in booking details

JorgeMucientes:
Fixes https://github.com/woocommerce/woocommerce-android/issues/WOOMOB-1732

### Description
This PR adds the missing customer note field to the booking details screen. Previously, when viewing booking details, the customer note was not displayed even though it was available in the booking data. This fix ensures that customer notes are now visible in the customer details section of the booking screen.

**Changes made:**
- Added `customerNote: String?` field to `BookingCustomerDetailsUiModel` data class
- Updated `BookingMapper` to map the customer note from the booking data model
- Modified `BookingCustomerDetails` composable to display the customer note in the UI with appropriate styling and spacing
- The customer note is shown below the billing address with a label and value layout

### Test Steps
1. Open the WooCommerce app and navigate to a booking with a customer note
2. View the booking details
3. Scroll to the customer details section
4. Verify that the customer note is now displayed below the billing address
5. Test with bookings that have no customer note to ensure the field is hidden completely
6. Verify the UI spacing and styling matches the existing design patterns

### Images/gif
 

- [ ] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/BookingEntity.kt
```diff
@@ -35,6 +35,7 @@ data class BookingEntity(
     val parentId: Long,
     val personCounts: List<Long>?,
     val localTimezone: String,
+    val customerNote: String?,
```
JorgeMucientes, line 38: This customer note is currently coming from `OrderEntity.customerNote`, however, I'm adding this directly to the booking entity because the backend is planning to add this customer note field directly to the Booking entity

AdamGrzybkowski (approved):
Works great! 

There's one thing we could consider - the customer note doesn't have any character limit (at least when I set it from the admin) so it can take a lot of space. Maybe we could set the maxLines to 4 and expand on click? 

Screenshot_20251121_101624

JorgeMucientes:
Thanks for the suggestion @AdamGrzybkowski 

My only concern about about setting a maxLine restriction is that we currently don't provide an alternative way to show the full note content as we do in order detail screen. So, if merchant is unaware that they can find the full note in the order detail screen they may be really annoyed by not being able to read the full note. 

Considering both tradeoffs, I believe is less disruptive to have a long note displayed (which I expect will be very rare) than running into the risk of cutting off the note's content when is more that 4 lines, which might not be that rare.

Wdyt? 

In the future this might get solved if we end up adding the "edit" functionality as we have in the order details, but I'm not sure will ever do that. Tbh, I don't think it makes sense for a merchant to be able to edit the customer note. But that's different topic for discussion :)

AdamGrzybkowski:
@JorgeMucientes My suggestion was to set the initial limit to 4 and add the logic to expand on tap. 

>Maybe we could set the maxLines to 4 **and expand on click**?

JorgeMucientes:
Oh dammit I read your suggestion too fast @AdamGrzybkowski. 

> was to set the initial limit to 4 and add the logic to expand on tap.

Yeah that makes total sense. Applied in the last commit:

https://github.com/user-attachments/assets/360c9e54-6cb1-4288-a304-ebdbb3a2163f

PR #15009, opened 2025-11-21
[WOOMOB-1744] - Date&Time booking picker improvement

AdamGrzybkowski:
WOOMOB-1744

### Description

This PR adds two improvements to the date&time picker in the Booking filter:
1. Sets the time when the initial date was selected. Start of the day for `FROM` and end of day for `TO`. This only happens if the date hasn't been selected yet.
2. Limits the possible date the user can choose. From the commit message: 

>- When the user already selected FROM date, the TO date will have a min date equal to the FROM date
>- When the user already selected TO date, the FROM date will have a max date equal to the TO date

### Test Steps

1. Launch the app
2. Open the Bookings tab - All
3. Tap filters
4. Clear filters (if any selected)
5. Tap Date & Time
6. Pick a FROM date
7. Confirm a time was shown as well (start of day)
8. Pick a TO date
9. Confirm a time was shown as well (end of day)
10. Change the TO time
11. Change the TO date
12. Confirm the time from step 10 stays 
13. While changing dates verify the min/max date restrictions

### Images/gif

https://github.com/user-attachments/assets/8f0e8a36-aeb8-4e9f-b4a9-59cccf34f323


JorgeMucientes (approved):
Nicely done @AdamGrzybkowskim, works well and code looks good. 

I just noticed that when the same day is selected for start and end, we don't ensure the user doesn't pick a `to` time earlier than `from` time. 

Image

it's sort of a corner case so I don't think we need to support this.

PR #15016, opened 2025-11-24
[WOOMOB-1767] Fix system back press not working in OrderCreateEditFormFragment

kidinov:
WOOMOB-1767

### Description

The system back button/gesture was not working in `OrderCreateEditFormFragment` while it worked correctly in other fragments. The issue was caused by the nested `NavHostFragment` (used for product selector) registering its own `OnBackPressedCallback` with higher priority, consuming the back press before the fragment's `BackPressListener` could handle it.

Fixed by adding a custom `OnBackPressedCallback` registered in `onResume()` to ensure it has higher priority (LIFO order) than the nested NavHostFragment's callback. This allows the fragment to properly intercept back press and show the discard changes dialog when needed.

### Test Steps

1. Open the app and navigate to Order Creation (Menu → Orders → Create Order (FAB))
2. Add a product or make any change to the order
3. Press the system back button or use the back gesture
4. Verify the discard changes dialog appears
5. **Test on tablet**: Repeat steps 1-4 on a tablet device to verify the fix works with the two-pane layout

### Images/gif

https://github.com/user-attachments/assets/f0e6a3bf-07ae-479d-99af-92d2bc6f51ae

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 23.8
 -----
+- [*] Fixed system back button not working on Order Creation screen [https://github.com/woocommerce/woocommerce-android/pull/15016]
 - [*] Optimized POS Local Catalog Settings screen [https://github.com/woocommerce/woocommerce-android/pull/14971]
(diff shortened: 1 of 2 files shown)


samiuelson (approved):
LGTM!

PR #15020, opened 2025-11-25
[Tooling] Update hotfix pipeline to to get version code directly from the release tool

iangmaia:
Fixes AINFRA-1524
See https://github.a8c.com/Automattic/wpcom/pull/196874

### Description

We're currently using a [Buildkite block step](https://github.com/woocommerce/woocommerce-android/blob/09080efeabdcf86a3250c147d165d23464ff3d09/.buildkite/release-pipelines/new-hotfix-release.yml#L5) to get the version code.

This PR changes the automation code so that it accepts the version code directly from ReleasesV2.

⚠️ We need to be careful when merging this PR to make sure it aligns with the changes done on ReleasesV2 ⚠️

### DIFF
--- a/.buildkite/release-pipelines/new-hotfix-release.yml
+++ b/.buildkite/release-pipelines/new-hotfix-release.yml
@@ -2,11 +2,6 @@
 ---
 
 steps:
-  - block: "What version code do you want to use for the new hotfix release?"
-    fields:
-      - text: "Version Code"
-        key: "version_code"
-        format: "[0-9]+" # Only allow numbers
   - label: "New Hotfix Release"
     plugins: [$CI_TOOLKIT]
     command: |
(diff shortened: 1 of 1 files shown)


AliSoftware (approved):
I checked that the `$${VERSION}` env var—which was removed from being passed by ReleasesV2 as part of 196874-ghe-Automattic/wpcom—was not used anywhere anymore (i.e. the one you replaced on line L24R15 of this PR was the only occurrence of it in `.buildkite/**` 👍 

❔ 🤔   I wonder how things were working before that change regarding the `$${VERSION}` being double-escaped btw… because in this case we actually DO want that env var to be resolved as `pipeline upload` time, since that env var is provided by ReleasesV2 when it triggers the step, while I don't think it's exposed to the job that actually runs that `command` (as that step doesn't include `env: { VERSION: $VERSION }` to forward it from the pipeline to the job's own environment)…

But since this PR fixes that by not double-escaping those env vars anymore so that they're interpreted at `pipeline upload` time as expected, that's better now anyway 👍

iangmaia:
> ❔ 🤔 I wonder how things were working before that change regarding the `$${VERSION}` being double-escaped btw… b

tbh I was also confused 😓 but based on what I see in a hotfix build [here](https://buildkite.com/automattic/woocommerce-android/builds/30624#019812a6-a8f8-4cd9-a977-c85f394207d1), it seems it still worked?

AliSoftware:
Oh right, I think because those are variables ReleasesV2 pass to the API call itself (similar to as if we provided them ourselves via the "Options" field when doing a manual build with the "New Build" button in the Buildkite web UI), those are propagated to all the jobs:
image
So that's how they are _also_ exposed at runtime to the child jobs that are created by the `pipeline upload`

That would be a different story if those were env vars set via `export VAR=value` during the pipeline-upload job for example.

iangmaia:
> That would be a different story if those were env vars set via `export VAR=value` during the pipeline-upload job for example.

Ah, exactly 🎯 👍

PR #15023, opened 2025-11-26
[Woo POS] Fix status bar height handling in Orders Loading and Order Details screen

samiuelson:
WOOMOB-1809

### Description

This pull request fixes status bar height handling in the WooCommerce POS Orders Loading and Order Details screens by applying proper status bar padding to ensure UI elements are not obscured by the system status bar.

- Renamed WooPosOrdersLoadingState to WooPosOrdersLoadingScreen for consistency
- Applied statusBarsPadding() modifier to both loading and detail screens to account for status bar height
- Adjusted padding and alignment in order detail views to maintain proper spacing with the new status bar handling

### Test Steps

Verify POS Orders Loading and Content state have correct UI.

#### Loading
| before | after |
|---|---|
|loading-before|loading-after|

#### Content
| before | after |
|---|---|
|loaded-before|loaded-after|


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
```diff
@@ -40,8 +41,11 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosThe
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
 
 @Composable
-fun WooPosOrdersLoadingState() {
```
samiuelson, line 43: `WooPosOrdersLoadingState` was a bit misleading, because it fits more of a name for a UI state model.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
```diff
@@ -40,8 +41,12 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosThe
...
-    Row(modifier = Modifier.fillMaxSize()) {
+fun WooPosOrdersLoadingScreen() {
+    Row(
+        modifier = Modifier
+            .fillMaxSize()
+            .statusBarsPadding()
```
kidinov: Shell we have it on the top level `WooPosOrdersScreen`? Why only 2 states not the rest then?

kidinov (commented):
@samiuelson I think it's not the best handling of this.

* Empty and error states not handled
* Loading state has the layout broken - notice the "list" part has padding at the top when loading.

Image

samiuelson:
> @samiuelson I think it's not the best handling of this.
> 
> * Empty and error states not handled
> * Loading state has the layout broken - notice the "list" part has padding at the top when loading.

Thanks for the review and suggestions, @kidinov. I applied padding to the root Box and updated the paddings in the children. Take a look:

**Empty state**
Screenshot_20251127_133740

**Loading**
Screenshot_20251127_124901

**Content**
Screenshot_20251127_124912

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
kidinov: @samueljseay 

Screen is also misleading, as this is not a screen but one of the "states" (loading) of the OrdersScreen

kidinov (changes requested):
@samiuelson 

Now the issue I mentioned in the previous review both on the content and loading states

Image

samiuelson:
> @samiuelson
> 
> Now the issue I mentioned in the previous review both on the content and loading states
> 

Ah, I didn't realize you meant the backgrounds of the panes are not shown behind the status bar, and I didn't notice that either. To accomplish that, we can't use a single padding at the root Box element, but we need to adjust the paddings in all the children. This is fixed in: 92be6f7

Screenshot_20251127_181420
Screenshot_20251127_181337
Screenshot_20251127_181304

kidinov (changes requested):
@samiuelson Looks close, but there is still an issue - the loading is not aligned with content and "jumping"

https://github.com/user-attachments/assets/ca7ad5d2-96b0-4720-a718-ef6839b718cf

Image

Image

Its not the case on the trunk

samiuelson:
> @samiuelson Looks close, but there is still an issue - the loading is not aligned with content and "jumping"
> 

Thanks @kidinov. I adjusted spaces and typography to match Order details and Order list layouts: 9a935d008d69a694f2045773870bfad0647dba37. No more jumping should happen:

https://github.com/user-attachments/assets/90a5743b-0207-4c87-a893-9f9d72d5238a

kidinov (changes requested):
@samiuelson 👋

Now the loading list has way to big margin at the top


https://github.com/user-attachments/assets/3fd75b9e-ce4c-4d32-8162-d58dc5941b29

image

samiuelson:
> Now the loading list has way to big margin at the top

Good catch, @kidinov! Fixed: fea4682

Screenshot_20251203_114554

kidinov (changes requested):
@samiuelson 👋

Now we are back to padding at the top of the orders section on the loading state

https://github.com/user-attachments/assets/92500903-d9df-4699-9e6e-06df414198da

Image

samiuelson:
> Now we are back to padding at the top of the orders section on the loading state

Fixed! 91fc260924da87b73898d3013db6ebf17f4a8473

kidinov (approved):
LGTM!

PR #15024, opened 2025-11-26
[Woo POS][Refunds] Add Issue Refund button

samiuelson:
WOOMOB-1801

### Description

This PR adds an "Issue Refund" button to the WooCommerce POS orders screen with a placeholder dialog implementation for the refund flow.

💡 The `WooPosIssueRefundDialog` layout is temporary. This PR aims to just open a dialog; its layout will be implemented in next PR.

💡 Feel free to remove `do not merge` label once the parent branch becomes `trunk`.

### Test Steps

1. Verify that the `Issue refund` button is not showing up when the `POS_REFUNDS` is `false`.
2. Verify that `Email Receipt` button works
3. Verify that `Issue Refund` button opens a dialog with a temporary layout.

### Images/gif

Screenshot_20251126_184347
Screenshot_20251126_184354


kidinov (commented):
👋 @samiuelson 

* The button is visible even when FF is off - so not finished functionality will be released?
* I am not sure about the design of the button—in Figma there is no more overflow menu without an email button at all 🤔 

Image

samiuelson:
Thanks for the review, @kidinov!

> * The button is visible even when FF is off - so not finished functionality will be released?

I was convinced I've committed that 🤦 Missing commit pushed: f41a4ac942fe3b3a9eb231ebae1ee223a8485065

> * I am not sure about the design of the button—in Figma there is no more overflow menu without an email button at all 🤔

This UX pattern proposal comes from a designer:
* Discussion: p1764073551626429/1763399395.564619-slack-C070SJRA8DP. 
* Docs: pdfdoF-7Av-p2

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -46,8 +47,13 @@ class WooPosOrdersViewModel @Inject constructor(
     private val ordersAnalyticsTracker: WooPosOrdersAnalyticsTracker
 ) : ViewModel() {
 
+    private val isRefundsEnabled = FeatureFlag.POS_REFUNDS.isEnabled()
```
kidinov: The refund button probably should be hidden also when selected order is fully refunded, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDetails.kt
```diff
@@ -69,10 +80,24 @@ fun WooPosOrderDetails(
 
             Spacer(Modifier.weight(1f))
 
-            WooPosButtonSmall(
-                text = stringResource(R.string.woopos_orders_email_receipt),
-                onClick = { onEmailReceiptButtonClicked(details.id) },
-            )
+            if (isRefundsEnabled) {
```
kidinov: I think it's not a good state model. It looks like we need to expose a list of buttons to the view, where the first one will be visible while the rest will be hidden in the overflow menu. Currently we kinda hardcode the priority in the view, while it's highly likely it will depend on the state of the specific order

kidinov: With the list of buttons, we can pass events that each specific button emits when clicked, just like we do on the home page. Eventually, we’ll have numerous actions that can be performed on the orders.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
samiuelson: Yes, good point. I plan to implement that later: WOOMOB-1773

kidinov (approved):
I left a couple of comments but they can be implemented as a seprate PR

Also:
Image

The floating menu button doesn't match the design

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDetails.kt
```diff
@@ -43,7 +54,9 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTyp
 fun WooPosOrderDetails(
     modifier: Modifier = Modifier,
     details: OrderDetailsViewState.Computed.Details,
-    onEmailReceiptButtonClicked: (Long) -> Unit
+    actions: List<OrderAction>,
+    onEmailReceiptButtonClicked: (Long) -> Unit,
```
kidinov: Not sure if it's the final state or still WIP, but this is not something that is driven by the VM as you have to pass hardcoded click listeners here and everywhere instead of calling one `onUIEvent` that accepts events that are part of the “actions”. Also, why is `OrderAction` not part of the whole state but a separate class?

samiuelson: Good idea, done here: d6443ab

kidinov: Looks nice, thanks!

PR #15027, opened 2025-11-26
[Tooling] Firebase App Distribution `trunk` prototype builds

iangmaia:
Fixes AINFRA-1579

### Description

This PR introduces Firebase App Distribution prototype builds, to facilitate testing across internal (or eventual external) testing groups.
I've decided to take an approach similar to what we have in [PocketCasts Android](https://github.com/Automattic/pocket-casts-android), where `trunk` builds will have a step requiring a manual trigger to create a build that will be distributed in Firebase (see screenshot below). We can use other approaches (like an automated nightly build) if needed.

I've initially implemented to select either Woo app (Wear / Regular app) but I could also tweak it to have both, if that's an interesting feature.

### Test Steps

To test the setup, I've temporarily (7d3ee6c734b8aae32944a3c551a516ca4e6f2723) enabled running the FAD job in this branch and already uploaded a couple of builds.

To install the builds via the Firebase: 

1. Go to MC's **Mobile Apps Tools**
2. Follow the instructions and add yourself to the **Firebase App Distribution** testing group 
3. You should be able to see the Woo Android builds built based on this branch.


AliSoftware (approved):
Tested installation in FAD on my test Android device, was able to install and launch the app 👍

PR #15048, opened 2025-12-02
[WOOMOB-1647] Migrate refunds to processRefund() API

kidinov:
WOOMOB-1647

### Description

This PR migrates the refund flow from the deprecated two-step API (`collectRefundPaymentMethod()` + `confirmRefund()`) to the new single-call `processRefund()` API in Stripe Terminal SDK 5.0.0.

**Changes:**
- Added `ProcessRefundAction` that uses the new `processRefund()` API
- Updated `TerminalWrapper` to expose the new `processRefund()` method
- Simplified `InteracRefundManager` to use the single-step flow
- Removed deprecated `CollectInteracRefundAction` and `ProcessInteracRefundAction`
- Updated and added unit tests for the new implementation

This is PR 3 of 5 in the Stripe SDK 5.0.0 migration series.

### Test Steps

1. Connect a card reader
2. Process a payment with an Interac card
3. Initiate a refund for the payment
4. Verify the refund completes successfully
5. Verify error handling works when refund fails

@malinajirka I could not test it as I don't have an interac card with me. Could you please test it?

### Images/gif

N/A - No UI changes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka (approved):
Hmm, there is something wrong with the refunds, but I'm able to replicate it even on trunk, so not related to this PR.

The refund always succeeds on Stripe's end, but fails on the site with `Invalid refund amount`. It almost seems like the hook from Stripe to the site is faster than our app -> so the order is already refunded at the time we send the refund request => we can't refund an already refunded order. I've created an issue for this - https://linear.app/a8c/issue/WOOMOB-1839/interac-refunds-shows-error-despite-being-successful.

PR #15056, opened 2025-12-03
[Woo POS][Refunds] Refundable items calculation

samiuelson:
## Description

WOOMOB-1802 part 1/2

This PR introduces foundational components for handling refunds in the Woo POS. It adds business logic to calculate refundable items from orders by accounting for previous refunds, along with the necessary ViewModel and state management for the refund UI.

**Key changes:**
* Implements WooPosGetRefundableItems use case that expands order line items into individual refundable units and calculates remaining quantities after refunds
* Adds WooPosRefundViewModel for managing refund dialog state with loading, error, and success states
* Introduces getOrderById method in WooPosOrdersDataSource with cache-first retrieval strategy

## Test Steps
- Verify unit tests pass - business logic correctly calculates refundable quantities
- To see these changes in the UI you can test the follow-up PR: https://github.com/woocommerce/woocommerce-android/pull/15057

## Images/gif
N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            item.total / item.quantity.toBigDecimal()
+        }
+    }
+
+    private fun calculateUnitTax(item: Order.Item): BigDecimal {
+        return if (item.quantity == 0f) {
+            item.totalTax
+        } else {
+            item.totalTax / item.quantity.toBigDecimal()
```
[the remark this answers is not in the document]
samiuelson: Done: 6539be6

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
```diff
@@ -0,0 +1,102 @@
...
+                null
+            }
+        }.toMap()
+    }
+
+    private fun calculateUnitPrice(item: Order.Item): BigDecimal {
+        return if (item.quantity == 0f) {
+            item.total
+        } else {
+            item.total / item.quantity.toBigDecimal()
```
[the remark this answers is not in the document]
samiuelson: Done: 6539be6

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItemsTest.kt
```diff
@@ -0,0 +1,1098 @@
...
+        val refunds = emptyList<Refund>()
+
+        // WHEN
+        val result = sut.invoke(order, refunds)
+
+        // THEN
+        assertThat(result).hasSize(5) // 2 + 3
+        assertThat(result.filter { it.variationId == 101L }).hasSize(2)
+        assertThat(result.filter { it.variationId == 102L }).hasSize(3)
+    }
```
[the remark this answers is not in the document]
samiuelson: Added: 3ee268f

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItemsTest.kt
```diff
@@ -0,0 +1,1098 @@
...
+
+        // THEN
+        assertThat(result).hasSize(5)
+        result.forEach { item ->
+            assertThat(item.unitTax).isEqualTo(BigDecimal("2.00"))
+        }
+    }
+
+    @Test
+    fun `given order item with quantity zero, when invoke called, then unit tax equals total tax`() {
```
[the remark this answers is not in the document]
samiuelson: Renamed: 3ee268f

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItemsTest.kt
```diff
@@ -0,0 +1,1098 @@
...
+
+        // THEN
+        assertThat(result).hasSize(4)
+        result.forEach { item ->
+            assertThat(item.unitPrice).isEqualTo(BigDecimal("20.00"))
+        }
+    }
+
+    @Test
+    fun `given order item with quantity zero, when invoke called, then unit price equals total`() {
```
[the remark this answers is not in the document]
samiuelson: Renamed: 3ee268f

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
```diff
@@ -0,0 +1,102 @@
...
+        refunds: List<Refund>,
+        productItems: List<Order.Item>
+    ): Map<Long, Float> {
+        val allRefundedItems = refunds.flatMap { it.items }
+
+        return productItems.mapNotNull { orderItem ->
+            val refundedQuantity = allRefundedItems
+                .filter { refundedItem ->
+                    refundedItem.productId == orderItem.productId &&
+                        refundedItem.variationId == orderItem.variationId
```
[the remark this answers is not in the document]
samiuelson: Fixed: ebbe003

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson:
I'm adding a "do not merge" label to prevent merging until the parent branch is trunk.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Calculate remaining refundable quantities for each order item.
+     */
+    private fun calculateMaxRefundQuantities(
+        refunds: List<Refund>,
+        productItems: List<Order.Item>
+    ): Map<Long, Float> {
+        val allRefundedItems = refunds.flatMap { it.items }
```
malinajirka: 💡 It seems we are iterating over allRefundedItems for each item. I guess performance is not an issue here, but we can consider simplifying this to something like:
```
private fun calculateMaxRefundQuantities(
    refunds: List<Refund>,
    productItems: List<Order.Item>
): Map<Long, Float> {
    val refundedByItemId = refunds
        .flatMap { it.items }
        .groupingBy { it.orderItemId }
        .fold(0f) { acc, item -> acc + item.quantity }

    return productItems
        .associate { it.itemId to (it.quantity - (refundedByItemId[it.itemId] ?: 0f)) }
        .filterValues { it > 0 }
}
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
```diff
@@ -0,0 +1,102 @@
...
+                null
+            }
+        }.toMap()
+    }
+
+    private fun calculateUnitPrice(item: Order.Item): BigDecimal {
+        return if (item.quantity == 0f) {
+            item.total
+        } else {
+            item.total.divide(item.quantity.toBigDecimal(), 2, RoundingMode.HALF_UP)
```
malinajirka: ⚠️ I realize this might not have an easy solution, but this feels extremely risky. WooCommerce uses 2 decimal places as default for USD, but not necessarily for all languages. Moreover, it's a setting that can be changed. 

Is there perhaps a solution in which we would avoid doing local calculations like this? 

If not, I think we might still want to explore if we could fetch the decimal-places-setting and use that instead of hardcoding 2. We also need to be careful to do the rounding the same way backend does - by row vs by item vs by total => each approach yields different results.

P.S. We have had several rounding issues in the app. I remember at least one case in which it was impossible to refund an order, since the app calculated the total with 0.01 difference compared to backend and the backend refused the refund. So I believe this is not just a theoretical issue.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -89,6 +89,15 @@ class WooPosOrdersDataSource @Inject constructor(
             }
         }
 
+    suspend fun getOrderById(orderId: Long): Result<Order> {
+        val cached = ordersCache.getAll().firstOrNull { it.id == orderId }
+        if (cached != null) {
+            return Result.success(cached)
+        }
+
+        return refreshOrderById(orderId)
```
malinajirka, line 98: ❓ If I'm reading the content of refreshOrderById correctly, it won't update the cache with the order we fetch, right? There is `updateCachedOrderIfPresent` method, but this line is unreachable if the order is already in cache, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModel.kt
```diff
@@ -0,0 +1,93 @@
...
+                refundableItems = refundableItems
+            )
+        }
+    }
+
+    private fun buildContentState(
+        order: Order,
+        refundableItems: List<WooPosRefundableItem>
+    ): WooPosRefundState.Content {
+        val subtotal = refundableItems.sumOf { it.lineTotal }
```
malinajirka, line 78: 📓 Similar as above - this is technically a local calculation, if there is some solution how we could avoid it that would be best. However, if there isn't, I think we might want to do explicit rounding here (ideally, again based on the setting from Woo core). Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
```diff
@@ -0,0 +1,282 @@
...
+        advanceUntilIdle()
+
+        // THEN
+        val state = viewModel.state.value
+        assertThat(state).isInstanceOf(WooPosRefundState.Content::class.java)
+        assertThat((state as WooPosRefundState.Content).refundableItems).hasSize(1)
+    }
+
+    @Test
+    fun `given order with refunds, when init, then calculates remaining refundable items correctly`() = runTest {
```
malinajirka: ❓ It feels this test mocks refunds and then verifies whether they are as expected - in order words it tests the mock, wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItemsTest.kt
```diff
@@ -0,0 +1,1173 @@
...
+import com.woocommerce.android.ui.orders.OrderTestUtils
+import com.woocommerce.android.util.CurrencyFormatter
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.mockito.kotlin.mock
+import java.math.BigDecimal
+import java.util.Date
+import kotlin.test.Test
+
+class WooPosGetRefundableItemsTest {
```
malinajirka, line 14: 💡 I would consider cleaning up this class. Ideally, each test should set only the data that it needs in order to pass. We could for example introduce a generateOrderItem method with default values - if the test for example needs the total to be `$10` it would pass just `10` to the method. This way, it would be clearer what each test is testing and each test would be significantly shorter and easier to maintain. Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
```diff
@@ -0,0 +1,282 @@
...
+                    variationId = 0,
+                    name = "Product 1",
+                    unitPrice = BigDecimal("10.00"),
+                    unitTax = BigDecimal("1.00"),
+                    formattedUnitPrice = "$10.00",
+                    formattedUnitTax = "$1.00",
+                    rowIndex = 0
+                ),
+                WooPosRefundableItem(
+                    orderItemId = 1L,
```
malinajirka, line 268: 🔍 It seems this orderItemId is identical to the first one, is this intentional? 

P.S. We might want to consider creating a helper generate class here as well. Wdyt?

malinajirka (approved):
Thanks @samiuelson ! 

I haven't found anything blocking, but I'm personally a bit worried about the local calculations. I think we might want to consider updating/introducing APIs that would give us all data we need, so we can avoid any local calculations. 
Doing local calculations has several risks:
- Rounding issues - we have been dealing with these as long as I can remember in Woo.
- Extensions/Hooks - there might be hooks that modify/change what can or can't be refunded (I'm not sure if they exist, but even if they don't, anyone can introduce them at any point).
- API-non-breaking changes in core might break the apps.

Having said that, I don't want to block progress of this project. If this is the approach we decided for, then lets go with it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
samiuelson: Good call, thanks 👍  I updated this to use `orderItem.price`, which matches the logic in store management: 5092da9c704a56c315b607f5c2c9ca38f63457b1

Now, only the tax calculation relies on local calculation (the same as in store management), but this is limited by the current API design.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
samiuelson: Yes. The `refreshOrderById` returns the fetched order, but it doesn't do an upsert. This is a pattern used in `WooPosOrdersViewModel` too, and I didn't want to interfere with the existing caching logic in this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModel.kt
samiuelson: This function matches the logic in store management refunds (`fun List<ProductRefundListItem>.calculateTotals(): Pair<BigDecimal, BigDecimal>`), so I don't think it's critical. I logged an issue to improve this code: https://linear.app/a8c/issue/WOOMOB-1852/woo-posrefunds-use-decimal-places-rounding-mode-mode-from-api, and will tackle it in a separate PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
malinajirka: Might be easier to discuss on a sync call, but I'd personally consider updating the API. Not doing any local calculations was (and maybe still is) one of the stones we have been building the POS around, so if it's solvable by updating the API it might be worth considering. (If we calculate taxes just for display, but the API uses value that it calculates itself, it's not worth it - but if the API uses the value the app provides, I think it might be worth considering to update the API).

malinajirka: > (the same as in store management) 

FWIW I personally believe that the fact it's used in store management shouldn't impact our decision on how to go about it in the POS - I mean it's good to know, but the fact we do local calculations in store management shouldn't be a justification for using it in POS. You were technically not saying that it is a justification, but want to be explicit about this because I honestly believe it's important to keep the quality of the POS high - even if it means slower development, changes on backend, and limitations on out-dated versions of Woo core.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
samiuelson: I renamed this method to `given order with refunds, when init, then builds content state with refundable items`: 036565e3e2701efaae7e526f73c82487228d8d5e

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
samiuelson: Yes, it's explicit to test the scenario of the same orderItemId and different `rowIndex`. I cleaned up the code to make it more explicit here: 0bebfe6110b68695dc31d39bd2a69f1fdee52521

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
samiuelson: > I personally believe that the fact it's used in store management shouldn't impact our decision on how to go about it in the POS - I mean it's good to know, but the fact we do local calculations in store management shouldn't be a justification for using it in POS. 

Good call. II think it's a good idea to raise a discussion internally on how to approach the implementation. Thanks for questioning the current logic.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetRefundableItems.kt
samiuelson: Great idea. Applied: 8d8ce37

PR #15059, opened 2025-12-04
[WOOMOB-1848] Fix coupon sharing message to display percentage correctly

malinajirka:
Fixes WOOMOB-1848

### Description

Fixed an issue where coupon sharing messages always displayed discount amounts with a currency symbol (e.g., $10.00) even when the coupon was percentage-based. Now percentage coupons correctly display as percentages (e.g., 10%) while fixed amount coupons continue to show currency symbols.

**Changes made:**
- Updated `formatSharingMessage` method in `CouponUtils` to use the existing `formatDiscount` method which properly handles both percentage and currency formatting based on coupon type
- Refactored method signature to accept a `Coupon` object instead of individual parameters to reduce parameter count and fix detekt issues
- Updated the call in `CouponDetailsViewModel` to pass the coupon object and type

### Test Steps

1. Create a percentage-based coupon (e.g., 10% off)
2. Go to coupon details and tap the share button
3. Verify the sharing message shows "Apply 10% off..." instead of "Apply $10.00 off..."
4. Create a fixed amount coupon (e.g., $15 off)
5. Go to coupon details and tap the share button  
6. Verify the sharing message shows "Apply $15.00 off..." with currency symbol

### Images/gif
| Before | After |
| -- | -- |
| Screenshot_20251204_092803 | Screenshot_20251204_093041 |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/coupons/details/CouponDetailsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/coupons/details/CouponDetailsViewModel.kt
@@ -184,11 +184,8 @@ class CouponDetailsViewModel @Inject constructor(
     fun onShareButtonClick() {
         coupon.value?.let { coupon ->
             couponUtils.formatSharingMessage(
-                amount = coupon.amount,
-                currencyCode = currencyCode,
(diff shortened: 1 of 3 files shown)


kidinov (approved):
LGTM! Thanks!

PR #15066, opened 2025-12-08
[HACK][POS] Prepopulate orders cache after opening POS

samiuelson:
### Description

This PR introduces a cache prepopulation mechanism for POS orders by fetching a small subset of orders (5) after a 5-second delay when the POS home screen is initialized. 

### Test Steps

1. Go to POS
2. After 5 seconds, open Orders screen
3. Notice the 5 first orders are available instantly

### Images/gif

|before|after|
|---|---|
|<video src=https://github.com/user-attachments/assets/330a4960-6db3-4600-9fd4-a8792de2e99c/>|<video src=https://github.com/user-attachments/assets/c3315751-5117-49bf-97b6-22ad9cf3a068/>|


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -255,6 +263,15 @@ class WooPosHomeViewModel @Inject constructor(
         }
     }
 
+    private fun prepopulateOrdersCache() {
+        viewModelScope.launch {
+            delay(ORDERS_CACHE_WARMUP_DELAY_MS)
```
samiuelson: 💡 The delay is added to relieve resources on POS start-up, because there are other network calls started immediately after POS launches (loading products in case the hybrid storage is used).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -255,6 +263,15 @@ class WooPosHomeViewModel @Inject constructor(
...
+            ordersDataSource.loadOrders(
```
kidinov: Orders prepopulation feels like something that completely has nothing to do with the home screen and its view model. Configuration (delay) also feels not related to the home screen. Have you considered calling it from from WooPosSplashViewModel, but on global coroutine scope?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -51,17 +51,18 @@ class WooPosOrdersDataSource @Inject constructor(
 
     companion object {
         const val POS_ORDERS_PAGE_SIZE = 25
+        const val POS_ORDERS_CACHE_PREPOPULATION_SIZE = 5
         private const val UNKNOWN_ERROR = "Unknown error"
     }
 
-    fun loadOrders(): Flow<LoadOrdersResult> = flow {
+    fun loadOrders(pageSize: Int = POS_ORDERS_PAGE_SIZE): Flow<LoadOrdersResult> = flow {
```
kidinov: All these changes related to loading 5 instead of 25, in my opinion, are not worth the effect - we are "lazy" loading 5kb

Image

kidinov (commented):
LGTM! But I think we should make it much simpler - just by calling `loadOrders` from the global context from `init` of `WooPosSplashViewModel`

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
samiuelson: The shorter page makes the request itself faster, which was my main motivation. But that's probably not worth complicating the codebase. Simplified in be96ae1

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
samiuelson: Moved to splash and also removed delay - simply calling prepopulation after all the calls in splash complete: be96ae1a990cfb001a9bb3ab997a219fa585a27a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSource.kt
```diff
@@ -133,8 +135,11 @@ class WooPosOrdersDataSource @Inject constructor(
         return fetchAndMap(searchQuery)
     }
 
-    private suspend fun fetchAndMap(searchQuery: String? = null): Result<List<Order>> {
-        val result = fetchOrdersFromRemote(page.get(), searchQuery)
+    private suspend fun fetchAndMap(
```
kidinov: Do we need any of the changes in this file at all?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -445,7 +445,10 @@ class WooPosOrdersViewModel @Inject constructor(
                                 searchInputState = WooPosSearchInputState.Closed
                             )
                         } else {
-                            replaceOrders(result.ordersWithRefunds)
+                            replaceOrders(
```
kidinov: Do we need this change?

kidinov (approved):
LGTM! I just think we could revert all the unnecessary changes before merging.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson:
Addressed the last 2 review comments - removed the unnecessary changes in WooPosOrdersDataSource and WooPosOrdersViewModel. The prepopulation works correctly with the default implementation. Changes in commit [350dc8a](https://github.com/woocommerce/woocommerce-android/pull/15066/commits/350dc8a1b69673aa94b762def3576c78214d68fe).

PR #15067, opened 2025-12-08
[Tooling] Expand Claude analysis to any failing build job

iangmaia:
### Description
This PR follows-up https://github.com/woocommerce/woocommerce-android/pull/14669 by running Claude in the entire build and generating a report considering all jobs that failed.

### Test Steps
You can create PRs on top of the present PR to check how the analysis report is going to look like.
I've run some tests on https://github.com/woocommerce/woocommerce-android/pull/15068, and you can check a failing build for it [here](https://buildkite.com/automattic/woocommerce-android/builds/34325).

### Images/gif
Screenshot 2025-12-09 at 12 49 07

### DIFF
--- a/.buildkite/claude-analysis.yml
+++ b/.buildkite/claude-analysis.yml
@@ -1,0 +1,28 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+# This pipeline is dynamically uploaded at the end of the build after a wait step
+
+agents:
+  queue: default
+
+steps:
+  - label: ":claude: Claude Build Analysis"
+    key: claude-analysis
+    if: build.state == "failing"
+    command: "exit 0"
+    plugins:
+      - $CLAUDE_PLUGIN:
+          api_key: "$ANTHROPIC_API_KEY"
+          buildkite_api_token: "$BUILDKITE_TOKEN_FOR_CLAUDE"
+          analysis_level: "build"
+          trigger: "always"
+          model: "claude-sonnet-4-5"
+
+  - label: ":claude: 💬 Comment Claude Analysis"
+    command: .buildkite/commands/comment-claude-analysis.sh
+    depends_on: claude-analysis
+    allow_dependency_failure: true
+    if: build.pull_request.id != null
+    plugins:
+      - $CI_TOOLKIT
--- a/.buildkite/claude-analysis.yml
+++ b/.buildkite/claude-analysis.yml
@@ -10,13 +10,15 @@ steps:
   - label: ":claude: Claude Build Analysis"
     key: claude-analysis
     if: build.state == "failing"
-    command: "exit 0"
+    soft_fail: true
+    command: "exit 1"
     plugins:
       - $CLAUDE_PLUGIN:
           api_key: "$ANTHROPIC_API_KEY"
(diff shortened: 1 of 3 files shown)


.buildkite/commands/comment-claude-analysis.sh
```diff
@@ -1,11 +1,14 @@
...
-  comment_on_pr --id claude-test-analysis "## 🤖 Test Failure Analysis
+# The claude-analysis step only runs when there are build failures,
+# so if it ran (and soft_failed), it means there were failures that Claude analyzed.
+CLAUDE_OUTCOME=$(buildkite-agent step get outcome --step claude-analysis 2>/dev/null || echo "not_run")
 
-Your tests failed. Claude has analyzed the failures - <a href=\"${BUILDKITE_BUILD_URL}/annotations#annotation-claude-analysis-${BUILDKITE_BUILD_ID}\" target=\"_blank\">check the annotation</a> for details."
+if [[ "${CLAUDE_OUTCOME}" == "soft_failed" ]]; then
+  comment_on_pr --id claude-build-analysis "## 🤖 Build Failure Analysis
+
+This build has failures. Claude has analyzed them - <a href=\"${BUILDKITE_BUILD_URL}\" target=\"_blank\">check the build annotations</a> for details."
```
AliSoftware: 💡 
```suggestion
This build has failures. Claude has analyzed them - <a href=\"${BUILDKITE_BUILD_URL}/annotations\" target=\"_blank\">check the build annotations</a> for details."
```

.buildkite/claude-analysis.yml
```diff
@@ -0,0 +1,30 @@
...
+          analysis_level: "build"
+          trigger: "always"
+          custom_prompt: "Do not mention successful points, focus on failures and recovery to keep the analysis succint. Only add the `Best Practices` section if the changes in this PR are directly relevant to it. Only add a `Root Cause` section when you're sure about the issues' causes."
+          model: "claude-sonnet-4-5"
+
+  - label: ":claude: 💬 Comment Claude Analysis"
+    command: .buildkite/commands/comment-claude-analysis.sh
+    depends_on: claude-analysis
+    allow_dependency_failure: true
+    if: build.pull_request.id != null
```
AliSoftware, line 28: @iangmaia I wonder if this would still work if we added `if: build.pull_request.id != null && build.state == "failing"` on that step too, so that it doesn't even get added to the pipeline if the build is successful (no job failure)?

This would prevent this step to be added at all when the Claude Build Analysis didn't even run (like in [this build](https://buildkite.com/automattic/woocommerce-android/builds/34324/steps/canvas)—instead of having the `comment-claude-analysis.sh` script having to check the `step get outcome` of that `claude-analysis` step manually?

iangmaia: I thought about that, but in principle we still want the comment job to run so that it can remove PR comments from eventual previous runs, otherwise we'd get scenarios such as:
1. Build runs and fails, comment is added on the PR
2. Dev fixes the build, pushes a fix, build runs again
3. Build succeeds, but since the comment job doesn't run, previous failure PR comment remains in the PR

AliSoftware: Ohhh good point 👍

AliSoftware (approved):

PR #15080, opened 2025-12-10
Fix/invalidate whats new cache on version change

wzieba:
Closes: [AINFRA-1635: Feature announcements are never refreshed when app version changes](https://linear.app/a8c/issue/AINFRA-1635/feature-announcements-are-never-refreshed-when-app-version-changes)

### Description

This PR fixes a bug in which, after initial fetch, the app never again requestes "what's new" announcements. 

### Test Steps

<details><summary>Clean Shared Preferences <code>LAST_VERSION_WITH_ANNOUNCEMENT</code></summary>

```shell
adb shell "run-as com.woocommerce.android.dev sed -i '/<string name=\"LAST_VERSION_WITH_ANNOUNCEMENT\"/d' /data/data/com.woocommerce.android.dev/shared_prefs/com.woocommerce.android.dev_preferences.xml" && adb shell am force-stop com.woocommerce.android.dev
```

</details> 

<details><summary>API Faker payload</summary>

```json
[
  {
    "request": {
      "httpMethod": "GET",
      "id": 0,
      "path": "/wpcom/v2/mobile/feature-announcements",
      "queryParameters": [],
      "type": {
        "type": "wp-com"
      }
    },
    "response": {
      "body": "{\n    \"announcements\": [\n      {\n        \"appVersionName\": \"19.8\",\n        \"announcementVersion\": 1,\n        \"minimumAppVersion\": \"1.0\",\n        \"maximumAppVersion\": \"99.0\",\n        \"appVersionTargets\": [\"19.0\"],\n        \"detailsUrl\": \"\",\n        \"isLocalized\": true,\n        \"responseLocale\": \"en\",\n        \"features\": [\n          {\n            \"title\": \"New Feature Title\",\n            \"subtitle\": \"Description of the new feature\",\n            \"iconBase64\": \"\",\n            \"iconUrl\": \"https://s0.wordpress.com/i/store/mobile/plans-premium.png\"\n          },\n          {\n            \"title\": \"Another Feature\",\n            \"subtitle\": \"Another cool feature description\",\n            \"iconBase64\": \"\",\n            \"iconUrl\": \"https://s0.wordpress.com/i/store/mobile/plans-premium.png\"\n          }\n        ]\n      }\n    ]\n  }",
      "endpointId": 0,
      "statusCode": 200
    }
  }
]
```

</details> 

#### Reproduction

1. In `com.woocommerce.android.util.BuildConfigWrapper` set `versionName` to any constant value, e.g. `19.0`
1. Install app from `trunk`
1. Import the above payload to API Faker
2. Close the app
3. Run `Clean Shared Preferences` from the snippet above (just to have a clean state before tests)
4. Open the app
  - ✅ Assert that you can see "What's new" announcement
5. Go to API Faker settings
6. Update the payload e.g. rename "New Feature Title" to "Test"
7. Save the updated payload
8. In Android Studio, increase `BuildConfigWrapper#versionName` (to simulate new app installation)
9. Install the app
  - ❌ See that "What's New in WooCommerce" shows OLD values

#### Test fix

Repeat above steps but on this branch and see that at the end you see the updated values.

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
@@ -66,11 +66,6 @@ class MainActivityViewModel @Inject constructor(
     unseenReviewsCountHandler: UnseenReviewsCountHandler,
     determineTrialStatusBarState: DetermineTrialStatusBarState,
 ) : ScopedViewModel(savedState) {
-    init {
-        launch {
-            featureAnnouncementRepository.getFeatureAnnouncements(fromCache = false)
-        }
-    }
 
     val startDestination = if (selectedSite.exists()) R.id.dashboard else R.id.nav_graph_site_picker
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
@@ -232,7 +232,6 @@ class MainActivityViewModel @Inject constructor(
             }
 
             val announcement = featureAnnouncementRepository.getLatestFeatureAnnouncement(fromCache = false)
-            prefs.setLastVersionWithAnnouncement(buildConfigWrapper.versionName)
 
             announcement?.let {
                 if (announcement.canBeDisplayedOnAppUpgrade(buildConfigWrapper.versionName)) {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
@@ -227,15 +227,15 @@ class MainActivityViewModel @Inject constructor(
 
     fun showFeatureAnnouncementIfNeeded() {
         launch {
-            val cachedAnnouncement = featureAnnouncementRepository.getLatestFeatureAnnouncement(fromCache = true)
+            if (prefs.getLastVersionWithAnnouncement() == buildConfigWrapper.versionName) {
+                return@launch
+            }
 
-            // Feature Announcement dialog can be shown on app resume, if these criteria are filled:
-            // 1. Current version is different from the last version where announcement was shown
-            // 2. Announcement content is valid and can be displayed
-            cachedAnnouncement?.let {
-                if (prefs.getLastVersionWithAnnouncement() != buildConfigWrapper.versionName &&
-                    cachedAnnouncement.canBeDisplayedOnAppUpgrade(buildConfigWrapper.versionName)
-                ) {
+            val announcement = featureAnnouncementRepository.getLatestFeatureAnnouncement(fromCache = false)
+            prefs.setLastVersionWithAnnouncement(buildConfigWrapper.versionName)
+
+            announcement?.let {
+                if (announcement.canBeDisplayedOnAppUpgrade(buildConfigWrapper.versionName)) {
                     WooLog.i(T.DEVICE, "Displaying Feature Announcement on main activity")
                     analyticsTrackerWrapper.track(
                         AnalyticsEvent.FEATURE_ANNOUNCEMENT_SHOWN,
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainActivityViewModelTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainActivityViewModelTest.kt
@@ -326,7 +326,7 @@ class MainActivityViewModelTest : BaseUnitTest() {
     }
 
     @Test
-    fun `given existing announcement cache, when app is upgraded and announcement is valid, then show announcement`() =
+    fun `given announcement is available on remote, when app is upgraded and announcement is valid, then show announcement`() =
         testBlocking {
             doReturn(
                 testAnnouncement
@@ -339,7 +339,7 @@ class MainActivityViewModelTest : BaseUnitTest() {
         }
 
     @Test
-    fun `given existing announcement cache, when app is upgraded and announcement is valid, track event is tracked`() =
+    fun `given announcement is available on remote, when app is upgraded and announcement is valid, track event is tracked`() =
         testBlocking {
             doReturn(
                 testAnnouncement
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainActivityViewModelTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainActivityViewModelTest.kt
@@ -328,7 +328,7 @@ class MainActivityViewModelTest : BaseUnitTest() {
     @Test
     fun `given existing announcement cache, when app is upgraded and announcement is valid, then show announcement`() =
         testBlocking {
-            doReturn(testAnnouncement).whenever(featureAnnouncementRepository).getLatestFeatureAnnouncement(true)
+            doReturn(testAnnouncement).whenever(featureAnnouncementRepository).getLatestFeatureAnnouncement(fromCache = true)
             doReturn("14.0").whenever(prefs).getLastVersionWithAnnouncement()
             doReturn("14.2").whenever(buildConfigWrapper).versionName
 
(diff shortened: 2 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/277">23.9</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

wzieba:
hi @malinajirka 👋 Do you mind reviewing this PR? I’m not sure whose bucket it should fall into, so please reassign it to a different team or member if they’re a better fit. Thanks! 🙌

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
```diff
@@ -227,15 +227,15 @@ class MainActivityViewModel @Inject constructor(
...
-            cachedAnnouncement?.let {
-                if (prefs.getLastVersionWithAnnouncement() != buildConfigWrapper.versionName &&
-                    cachedAnnouncement.canBeDisplayedOnAppUpgrade(buildConfigWrapper.versionName)
-                ) {
+            if (prefs.getLastVersionWithAnnouncement() == buildConfigWrapper.versionName) {
+                return@launch
+            }
+
+            val announcement = featureAnnouncementRepository.getLatestFeatureAnnouncement(fromCache = false)
+            prefs.setLastVersionWithAnnouncement(buildConfigWrapper.versionName)
```
malinajirka: 🧇 This seems a bit off to me, are you sure about this change?
1. We invoke setLastVersionWithAnnouncement even if the request to fetch the announcement fails.
2. We invoke setLastVersionWithAnnouncement also in the FeatureAnnouncementViewModel when it's shown, which feels more appropriate.

Wdyt?

malinajirka (commented):
Thanks @wzieba for working on this!

I have very little context about how the announcement works, and tbh I'm a bit confused. 

I thought each announcement is displayed just once for each user. I also thought that each announcement has its version and it'd be displayed again only if we updated the version of the announcement.

But if I understand the code correctly, we display the announcement anytime the app is updated, given the min/max announcement versions match? It also seems that the announcement version is completely ignored?

What am I missing?

wzieba:
thanks for taking a look @malinajirka !

> I thought each announcement is displayed just once for each user. 

Yes, this is my understanding as well. But currently on `trunk`, announcement will be displayed **once per app lifetime**. If user sees an announcement for version `19.0`, they won't see any announcement for version `20.0`, `21.0` etc. in the future. Maybe the reproduction instruction could bring more light

> I also thought that each announcement has its version

It has, but it's not used in the app anywhere. I remove it in a different PR: https://github.com/woocommerce/woocommerce-android/pull/15075/changes/1c41d3d18760ca62047b09cb5cc528b0988a1148#diff-fea7c31ec2218d75212182e1ceb6aa44d5d54bb54e0578b4cb2fbede134b9e37L10

> it'd be displayed again only if we updated the version of the announcement.

The announcement is shown whener the version of the app is between `minimumAppVersion` and `maximumAppVersion` + a few more cases (see [`FeatureAnnouncement#canBeDisplayedOnAppUpgrade`](https://github.com/woocommerce/woocommerce-android/blob/e29e56f60989dd35f225cc0ae99b33449950e4fc/WooCommerce/src/main/kotlin/com/woocommerce/android/model/FeatureAnnouncement.kt#L27)).

> But if I understand the code correctly, we display the announcement anytime the app is updated, given the min/max announcement versions match? It also seems that the announcement version is completely ignored?

Yes, you're correct*: we display the announcement when app is updated and the announcement is meant to be shown for this app version. (edit: I clarified this response, previous one missed the point I think :) )

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
wzieba: Thanks! Sure, we can remove `prefs.setLastVersionWithAnnouncement(buildConfigWrapper.versionName)` from here to address the case when `getLatestFeatureAnnouncement` would fail, good idea 👍

wzieba:
Do you think we should also track version of the announcement to not risk showing it again if there are multiple versions between `minimumAppVersion` and `maximumAppVersion`? 

As I'm thinking about it, maybe we should. I mean, it all depends on how we work with `announcementVersion`, `minimumAppVersion` and `maximumAppVersion` when creating a new announcement on the backend.

malinajirka:
> Yes, you're correct*: we display the announcement when app is updated and the announcement is meant to be shown for this app version. 

I believe this was never the intention though. I think the idea for having announcements available in multiple versions was simply to show it to users who skip certain app updates.

> Do you think we should also track version of the announcement to not risk showing it again if there are multiple versions between minimumAppVersion and maximumAppVersion?
> As I'm thinking about it, maybe we should. I mean, it all depends on how we work with announcementVersion, minimumAppVersion and maximumAppVersion when creating a new announcement on the backend.

Tbh, I'm not sure how to go about it given the announcement doesn't have a unique key - I don't see how tracking the version helps here. Perhaps, we could store hash of the announcement and make sure to never show it again as long as the hash is identical?

wzieba:
> Tbh, I'm not sure how to go about it given the announcement doesn't have a unique key

Hm, I think it does: https://github.com/woocommerce/woocommerce-android/blob/e29e56f60989dd35f225cc0ae99b33449950e4fc/libs/fluxc/src/main/java/org/wordpress/android/fluxc/model/whatsnew/WhatsNewAnnouncementModel.kt#L11, we simply don't use it now (at least I'd expect that `announcementVersion` is unique)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
wzieba: > We invoke setLastVersionWithAnnouncement also in the FeatureAnnouncementViewModel when it's shown, which feels more appropriate.

The problem with this is that if the backend doesn't offer announcements for the specific version of the app, then we'll never call `setLastVersionWithAnnouncement` (because `FeatureAnnouncementViewModel` won't ever start).

This means that each time the app would make an unnecessary HTTP request on start.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
```diff
@@ -227,15 +227,15 @@ class MainActivityViewModel @Inject constructor(
...
-            // Feature Announcement dialog can be shown on app resume, if these criteria are filled:
-            // 1. Current version is different from the last version where announcement was shown
-            // 2. Announcement content is valid and can be displayed
-            cachedAnnouncement?.let {
-                if (prefs.getLastVersionWithAnnouncement() != buildConfigWrapper.versionName &&
-                    cachedAnnouncement.canBeDisplayedOnAppUpgrade(buildConfigWrapper.versionName)
-                ) {
+            if (prefs.getLastVersionWithAnnouncement() == buildConfigWrapper.versionName) {
+                return@launch
+            }
```
wzieba, line 227: In case it'll help, it's worth noting that this condition _is not new_. It's just moved from the lines below. This is a slight optimization to not fetch announcements if they won't be shown anyway.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
wzieba: Anyway, I've removed this line. My intention here was to optimize so the app doesn't call backend if there are no new announcements for this version of the app, but covering the case of a failed http request that you mentioned would require a refactor to `Result` from repository (because now we get `null` if there are no announcements OR when there's an error).

wzieba:
hi @malinajirka 👋 after putting more thought into the discovery of this, I've added two more commits. 

Regarding parameters like `announcementVersion`, `minimumAppVersion` or `maximumAppVersion` - I'd like to highlight, that this PR has not much to do with any conditions about showing or not showing announcements. 

It focuses on making sure that the cache is up to date **before** we show the announcement and actually fixes a bug of potentially showing **old** announcements whenever our users updated the app (as it's described in Reproduction).

I limit myself with any other improvements I could spot to minimize efforts here. I'm curious what you think about the current version.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainActivityViewModelTest.kt
```diff
@@ -337,9 +339,11 @@ class MainActivityViewModelTest : BaseUnitTest() {
         }
 
     @Test
-    fun `given existing announcement cache, when app is upgraded and announcement is valid, track event is tracked`() =
+    fun `given announcement is available on remote, when app is upgraded and announcement is valid, track event is tracked`() =
```
malinajirka: 🔍 The last part should be `when track ....` - otherwise detekt will complain.

malinajirka (approved):
Thanks @wzieba! I've tested it and it works as described. Thanks for the fix 🙇.

PR #15085, opened 2025-12-11
Issue/woomob 1859 declared age range apis for app stores

JorgeMucientes:
Part of WOOMOB-1859

### Description
This PR integrates [Play Age Signals API](https://developer.android.com/google/play/age-signals/use-age-signals-api) in Woo Android. 

The added logic works as follows:
- Every time the app is opened (after being previously killed) we'll check the Google Age Signal API values to verify if the user is still eligible to use the app. 
- If the user is ineligible (more on when this could happen) the we'll log the user out and show a dialog that the access to the app is restricted for them based on their age. 

When is a user ineligible to use WooCommerce app? Discussion p2y3YZ-anO-p2#comment-24048. 
According to WooCommerce TOS our app usage is expected for 13+ years old users: 
> All of our apps have a required age of 13+ and so we would need the app stores to return a result showing the user is at least that age. Parents can also revoke consent for a minor in their family account; the API should be capable of informing us of such revocations and we’ll need to figure out how to honor those. 

With the TOS in mind the logic to determine if a user is eligible will be based on the 4 different user statuses the API provides: 

Screenshot 2025-12-12 at 18 11 47

- Verified users will be granted access bu default
- Supervised users above +13 plus we'll be granted access by default. 
- Supervised users whose access was denied will be log out and access restricted via a blocking dialog. See screen recording below. 
- Any other case were we fail to retrieve the `userStatus`, we'll grant access by default. 

Here's a diagram to help better understand the different flows:

Screenshot 2025-12-22 at 17 37 32

### Test Steps
We currently can't test in prod, because we app needs to be downloaded from Google Play in order for the API to work. For that Google provides a `FakeAgeSignalsManager()` that enables us to test the different scenarios. 

**Apply the following patch** in order to simulate the different API responses we'll get from the Play Signals API:

[Use_FakeAgeSignalsManager_to_simulate_supervised_users.patch](https://github.com/user-attachments/files/24125755/Use_FakeAgeSignalsManager_to_simulate_supervised_users.patch)

For each of the following tests you'll need to open `AgeSignalsClient.kt.` and update the mocked `userStatus`

**Test Case 1: User 18+** (Verified)
- Set userStatus to `VERIFIED` 
- Open the app 
- Log in as usual

**Test Case 2: Approved 13+ user** (Supervised)
- Set `userStatus` to SUPERVISED
- Set `ageUpper ` to 14 (use setAgeUpper(14))
- Open the app 
- Log in as usual

**Test Case 3: Supervised user under 13**
- Set `userStatus` to SUPERVISED_APPROVAL_PENDING
- Set `ageUpper ` to 12
- Open the app 
- Logged out and restricted access for being below 13+ years old

**Test Case 4: Supervised user approval denied**
- Set `userStatus` to SUPERVISED_APPROVAL_DENIED
- No matter what the age is. 
- Open the app 
- You'll be logged out and a dialog displayed saying you Google account is restricted from using the app.  

### Images/gif

https://github.com/user-attachments/assets/0c819619-c0bd-457f-bd51-5f86d80637d3


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>GoogleAgeSignalsClient</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes:
This is a fairly critical change I'd also would like to ask if someone from Android in Peacock could give a look to this. Would really appreciate if one of you can take a look into this 🙏🏼  @malinajirka @kidinov @samiuelson.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
```diff
@@ -206,6 +207,9 @@ class LoginActivity :
             unifiedLoginTracker.setFlow(ss.getString(KEY_UNIFIED_TRACKER_FLOW))
             connectSiteInfo = ss.parcelable(KEY_CONNECT_SITE_INFO)
         }
+        if(appPrefsWrapper.isUserAgeEligibleForAppUse.not()){
```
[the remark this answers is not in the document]
irfano: Detekt didn’t flag this, but let’s just fix it.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityCheckerTest.kt
```diff
@@ -0,0 +1,157 @@
...
+    @Test
+    fun `given user is supervised approval pending and age is 12, when checkAge called, then user is not eligible`() =
+        testBlocking {
+            client.setExpectedValues(AgeSignalsVerificationStatus.SUPERVISED_APPROVAL_PENDING, 12)
+
+            ageEligibilityChecker.checkAge()
+
+            assertEquals(false, ageEligibilityChecker.isUserAgeRangeEligible.value)
+            verify(prefsWrapper).isUserAgeEligibleForAppUse = false
+        }
```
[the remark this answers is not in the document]
irfano: This is valid feedback.  
- `given user is supervised approval pending and age is under 13, when checkAge called, then user is NOT eligible`
- `given user is supervised approval pending and age is 12, when checkAge called, then user is not eligible`

These are the same scenarios.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
```diff
@@ -1043,6 +1047,18 @@ class LoginActivity :
...
 
+    private fun showAgeRestrictionDialog() {
+        val dialog = AlertDialog.Builder(this)
+            .setTitle(R.string.age_restriction_dialog_title)
+            .setMessage(R.string.age_restriction_dialog_message)
+            .setCancelable(false)
+            .setPositiveButton(android.R.string.ok) { _, _ -> finishAffinity() }
+            .create()
+        dialog.setCancelable(false)
+        dialog.setCanceledOnTouchOutside(false)
```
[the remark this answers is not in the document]
irfano: This also seems valid.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4348,4 +4348,6 @@
     <string name="about_automattic_logo_description">Automattic logo</string>
     <string name="about_automattic_back_icon_description">Back icon</string>
     <string name="about_automattic_app_icon_description">App icon</string>
+    <string name="age_restriction_dialog_title">App Access Restricted</string>
+    <string name="age_restriction_dialog_message">Your Google Account lost access because you’re underage. Contact your parent or guardian to get permission again.</string>
```
irfano: Small copy suggestion ✍️  

The current message uses **“lost access” + “again”**, which implies the user had permission before and then lost it.  
What about this?

“Your Google Account can't access the app due to age restrictions. Contact your parent or guardian to get permission.”

gradle/libs.versions.toml
```diff
@@ -110,6 +111,7 @@ wordpress-utils = '3.15.0'
 zendesk = '5.5.1'
 
 [libraries]
+age-signals = { module = "com.google.android.play:age-signals", version.ref = "ageSignals" }
```
irfano: We’ve defined other Google Play libraries in this way:

```
google-play-app-update = { group = "com.google.android.play", name = "app-update", version.ref = "google-play-app-update" }
google-play-review = { group = "com.google.android.play", name = "review", version.ref = "google-play-review" }
```

It’s not a major, but for consistency, we can follow the same style and also place the Age Signals library above the other Google Play libraries.

Suggestion:
```
google-play-age-signals = '0.0.2'
google-play-age-signals = { group = "com.google.android.play", name = "age-signals", version.ref = "google-play-age-signals" }
```

irfano (changes requested):
- It's worth adding RELEASE-NOTES for this update.
- When I change the upper age limit from 12 to 13 and open the app for the first time, the app still applies the previous behavior (age 12) and blocks access.
Similarly, when the upper age limit is changed from 14 to 12 and the app is restarted, on the first launch it behaves as if the limit were still 14 and allows access, but on the second launch it correctly treats it as 12 and blocks the app.

https://github.com/user-attachments/assets/a957df15-1c8d-45bb-af4f-3bfbdecd1285

All other parts look good. 👍🏻 Thanks for adding this feature.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
```diff
@@ -0,0 +1,67 @@
...
+) {
+
+    private val _isUserAgeRangeEligible = MutableStateFlow(prefsWrapper.isUserAgeEligibleForAppUse)
+    val isUserAgeRangeEligible: StateFlow<Boolean> = _isUserAgeRangeEligible.asStateFlow()
+
+    suspend fun checkAge() {
+        try {
+            val result = client.checkAge()
+            processAgeCheck(result.userStatus, result.ageUpper)
+        } catch (exception: ApiException) {
```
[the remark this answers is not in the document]
JorgeMucientes: >  use a more general Exception type to ensure all error cases are properly handled.


This is exactly what we are doing. Use a more generic exception to stay on the sage side. 
`ApiException` is the parent class for `AgeSignalsException`. Google doesn't document all the different `ApiException` that are thrown by `AgeSignalsManager` so this approach is the safer one without being too generic.

WooCommerce/src/main/res/values/strings.xml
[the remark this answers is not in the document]
JorgeMucientes: Doesn't apply. A user can only be denied access if its under age anyway.

JorgeMucientes: Good suggestion, thanks. I actually updated both texts (title and message). Let me know what your thoughs are on the new values.

JorgeMucientes:
I'm moving this to draft as I need to add bigger changes than I expected. I'll let you know once everything is ready again @irfano. And thanks for the first round 👍🏼

JorgeMucientes:
Hey @irfano this is ready for another round 🙏🏼 
- Applied all of your small suggested changes
- Addressed the main bug you reported by making the age eligibility status observable
- Added tracking so we can figure out the results the API returns and how often this dialog is shown \
- Some other minor refactor in AgeEligibilityChecker to avoid unnecessary checks

> It's worth adding RELEASE-NOTES for this update.

I haven't added release notes yet as I'm still not sure where these changes are going to land (23.8.1 or 23.9)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    private fun processAgeCheck(userStatus: Int?, ageUpper: Int?) {
+        val isUserAgeEligible = when (userStatus) {
+            AgeSignalsVerificationStatus.VERIFIED -> true
+            AgeSignalsVerificationStatus.SUPERVISED,
+            AgeSignalsVerificationStatus.SUPERVISED_APPROVAL_PENDING -> {
+                if (ageUpper == null) {
+                    true // If we can't determine the age return true
+                } else {
+                    ageUpper >= WOOCOMMERCE_TOS_MINIMUM_AGE_FOR_APP_USE
```
malinajirka: I mentioned this on slack (p1766407567026729-slack-C03L1NF1EA3). From what I understand, if the state is SUPERVISED (approved by parent) we shouldn't check their age against WOOCOMMERCE_TOS_MINIMUM_AGE_FOR_APP_USE. For example, 12 years old user whose parents approved the app should be able to use it, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.woocommerce.android.analytics.AnalyticsTrackerWrapper
+import com.woocommerce.android.ui.login.AccountRepository
+import com.woocommerce.android.util.WooLog
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
malinajirka, line 20: ❓ Is the singleton annotation needed here? From what I understand we inject the checker just once in the initializer. If we declare it as singleton, it'll never be garbage collected. It's a minor memory optimization, but unless we need it I'd personally replace it with `@Reusable`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeSignalsClient.kt
```diff
@@ -0,0 +1,29 @@
...
+interface AgeSignalsClient {
+    suspend fun checkAge(): AgeCheckResult
+}
+
+data class AgeCheckResult(
+    val userStatus: Int?,
+    val ageUpper: Int?
+)
+
+@Singleton
```
malinajirka, line 20: ❓Same as above, is the Singleton annotation needed here or could we replace it with @Reusable?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
```diff
@@ -1043,6 +1050,25 @@ class LoginActivity :
         }
     }
 
+    private fun keepTrackOfAgeEligibility() {
+        val dialog = AlertDialog.Builder(this)
```
malinajirka: 🔍 The dialog is leaking after configuration change, but considering how rare this dialog will be I think it's not worth fixing.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityCheckerTest.kt
```diff
@@ -0,0 +1,259 @@
...
+            mapOf(
+                "retrieved_age" to DEFAULT_USER_AGE_UPPER,
+                "user_status" to "VERIFIED",
+                "access_restricted" to true
+            )
+        )
+    }
+
+    @Test
+    fun `given user is supervised and age is under 13, when checkAge called, then user is NOT eligible`() =
```
malinajirka, line 58: This test is related to the behavior I mentioned above - I believe the user should be allowed to use the app, since they have been approved by their parent. Wdyt?

malinajirka (commented):
Thanks @JorgeMucientes! Overall looks good to me. I've left some comments with suggestions + questions + one potential bug.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
irfano: > According to WooCommerce TOS our app usage is expected for 13+ years old users:

Given that, the current behavior seems expected: we shouldn’t allow 12-year-old users to use the WooCommerce platform.

> **Test Case 3: Supervised user under 13**
> 
> - Set `userStatus` to SUPERVISED_APPROVAL_PENDING
> - Set `ageUpper`  to 12
> - Open the app
> - Log in as usual

I think the last step is misleading for this case. If the user is under 13, they shouldn’t be able to “log in as usual”; they should be logged out.
@JorgeMucientes, can you confirm my understanding is correct?

JorgeMucientes: > According to WooCommerce TOS our app usage is expected for 13+ years old users:
> > Given that, the current behavior seems expected: we shouldn’t allow 12-year-old users to use the WooCommerce platform.


Yes, that's it. Replied on Slack as well and shared the diagram that might help (will ad it to the PR desc): 

Image

> I think the last step is misleading for this case. If the user is under 13, they shouldn’t be able to “log in as usual”; they should be logged out.

Well, spotted those instructions are wrong. The user should be logged out and access restricted. Updated the PR desc with the correct steps.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
```diff
@@ -37,6 +43,8 @@ class AgeEligibilityChecker @Inject constructor(
         if (isUserAgeRangeEligible.value.not()) {
             accountRepository.logout()
         }
+        trackingProperties["access_restricted"] = _isUserAgeRangeEligible.value
```
irfano: `!` is needed here.
`_isUserAgeRangeEligible.value` → `!_isUserAgeRangeEligible.value`

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4348,6 +4348,6 @@
     <string name="about_automattic_logo_description">Automattic logo</string>
     <string name="about_automattic_back_icon_description">Back icon</string>
     <string name="about_automattic_app_icon_description">App icon</string>
-    <string name="age_restriction_dialog_title">App Access Restricted</string>
-    <string name="age_restriction_dialog_message">Your Google Account lost access because you’re underage. Contact your parent or guardian to get permission again.</string>
+    <string name="age_restriction_dialog_title">Account Access Required</string>
+    <string name="age_restriction_dialog_message">Your Google Account settings indicate that you need a parent or guardian's permission to continue. They can grant access using their Google account to get you back online.</string>
```
irfano: Should we show the same message to supervised users who are under 13? In that case, the restriction isn’t related to their Google account. So I think we should either show a different message or make this one more generic.

irfano (changes requested):
The incorrect “previous app state” issue is fixed now. 👍🏻 Thanks for the change. I added more comments that needs your attention. Also some reminders
- RELEASE-NOTES for the change,
- Register new tracks

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
JorgeMucientes: Good point @malinajirka. However, AgeEligibilityChecker.kt is injected in several places were we subscribe to its state like LoginActivity.kt or MainActivityViewModel.kt. Removing the `@Singleton` would mean I'll need to save and expose the state in an observable DataStore. This is not a big deal but I believe it's overkill for this feature. 

> It's a minor memory optimization, but unless we need it I'd personally replace it with @Reusable.

Curious about that. Given AgeEligibilityChecker is injected in `AppInitializer` class which afaik will survive as long as the app is not killed, setting `AgeEligibilityChecker` as a singleton will prevent the app from creating multiple instances of the same class, were as `@Resuable` wouldn't guarantee that. So technically, in terms of memory usage, @Reusable could be worse?

WooCommerce/src/main/res/values/strings.xml
JorgeMucientes: Totally! Well spotted I've refactored the logic from `AgeEligibilityChecker.kt` to be more flexible and display a different error message depending on why the access was restricted [3788e0d](https://github.com/woocommerce/woocommerce-android/pull/15085/commits/3788e0da436aa4e1e1cd6ceb7abeb29ed2b5310b)

JorgeMucientes:
Hey @irfano thanks for your thorough review 🙏🏼 
I've addressed all of your feedback in the last couple commits. Hopefully this will leave the implementation ready to merge once we decide (after January 1) wether this is released as a hotfix or as regular release.

> RELEASE-NOTES for the change,

As mentioned above the reason I haven't added release notes yet is because I'm unsure which app version will be the one including these changes in the end. 
 
> Register new tracks

I'll proceed and register those events now. Great reminder!

WooCommerce/src/main/res/values/strings.xml
irfano: Thanks for the improvement. 🚀

irfano (approved):
> As mentioned above the reason I haven't added release notes yet is because I'm unsure which app version will be the one including these changes in the end.

Sorry for unnecessary reminder. I missed your previous comment. 

LGTM! 👍🏻  Thanks for staying on top of this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/ageeligibility/AgeEligibilityChecker.kt
malinajirka: > Curious about that. Given AgeEligibilityChecker is injected in AppInitializer class which afaik will survive as long as the app is not killed

Good point, I didn't realize we hold onto the instance in AppInitializer. So it doesn't really matter which one we use.

> setting AgeEligibilityChecker as a singleton will prevent the app from creating multiple instances of the same class, were as @Resuable wouldn't guarantee that.

Reusable annotation also guarantees maximum one instance at a time, but it can be garbage collected and re-created when needed later. However, as you correctly pointed out, we reference it so we never let the GC to collect it anyway.

JorgeMucientes:
So, the code is ready to be merged. However, given the Court has paused the Texas ruling, I'll merge this with a feature flag disabled in prod until we know more about next steps on the law and whether Google will update anything related to the age eligiblity APIs. 
Let me know if you want to double check the changes added in this last [commit](https://github.com/woocommerce/woocommerce-android/pull/15085/commits/b9f9b1896a26d6096a75d3c6b2aeeb1696a3ddf0) @irfano if not let's proceed and merge this PR.

irfano:
LGTM! Thanks for handling this,

PR #15086, opened 2025-12-11
[HACK week] Modularization of SelectedSite

samiuelson:
### Description

This PR modularizes the `SelectedSite` class by moving it (along with supporting utilities and extensions) from the main WooCommerce module to the libs/commons module. `SelectedSite` is one of the core classes used by a great number of classes, moving it to `libs/commons` enables better code reuse and reduces coupling, unlocking further modularization possibilities 👉  https://github.com/woocommerce/woocommerce-android/pull/15094

### Test Steps

- Verify CI is green
- Smoke test the app

### Images/gif

N/A


WooCommerce/build.gradle
```diff
@@ -257,8 +257,6 @@ dependencies {
     implementation(libs.automattic.tracks.android)
     implementation(libs.automattic.tracks.crashlogging)
 
-    implementation(project(":libs:fluxc"))
```
samiuelson, line 267: 💡 I'm looking for opinions on whether it's better to declare `fluxc` as a transitive dependency in `libs/commons` and remove it from the `WooCommerce` module (this is what I did here since think it's cleaner) or keep it as a regular dependency in both of them. cc @wzieba

wzieba: I'm leaning into declaring `fluxc` as regular dependency. I can imagine a scenario in which we add a new module that will use `common` but won't have to use anything from `fluxc`. Not having `fluxc` as transitive dependency would be an advantage in such a case, to not share unnecessary classes, recuding dependencies etc.

malinajirka: Unless I'm missing something I believe Fluxc will be part of `common` either case @wzieba. The difference is just in the WooCommerce module - whether it's declared explicitly there or its a transitive dependency.

I mean, ideally, I would also prefer not having fluxc in `common`, but I don't think that's what's being discussed here.

libs/commons/build.gradle
```diff
@@ -19,17 +21,41 @@ android {
...
 
+    buildFeatures {
+        buildConfig = true
+    }
+
+    lintOptions {
+        // Workaround for kotlinx-metadata version incompatibility:
+        // Kotlin 2.2.21 generates metadata v2.2.0, but AGP 8.13.2's Compose lint detector
+        // only supports up to v2.0.0, causing crashes when analyzing Kotlin files.
+        // This can be removed when AGP updates its bundled kotlinx-metadata-jvm dependency.
```
malinajirka, line 34: 💡 I think we might want to create a linear task for this to ensure we don't forget about it.

malinajirka (approved):
Thanks, LGTM

WooCommerce/build.gradle
wzieba: Ok so I understand the issue as @samiuelson described it is

#### Option A: `fluxc transitive`

```groovy
// libs/commonts/build.gradle
dependencies {
    api(project(":libs:fluxc"))
}

// WooCommerce/build.gradle
dependencies {
    // nothing
}
```

#### Option B: `fluxc non-transitive`

```groovy
// libs/commonts/build.gradle
dependencies {
    implementation(project(":libs:fluxc"))
}

// WooCommerce/build.gradle
dependencies {
    implementation(project(":libs:fluxc"))
}
```

My opinion is that I think `Option B` is better, because in the future, if we introduce a new module `NewModule` and we'll want to use `common` there **but** we don't **need** to use `fluxc` , then with Option B we don't unnecessarily offer `fluxc`. This way we reduce the dependency chain.

wzieba: > I mean, ideally, I would also prefer not having fluxc in common, but I don't think that's what's being discussed here.

That's also an interesting point to discuss: it depends on what `common` is for us. I guess we might need both: a `common` that includes `fluxc` and a `common` that doesn't have to and shares different types of things. But this split is still doable incrementally later, so I _think_ I would proceed with this PR as it's a step in right direction.

malinajirka: Ahh, I see - makes sense, thanks for elaborating on it! I agree that option B sounds better.

PR #15093, opened 2025-12-12
Issue/woomob 1867 woo pos implement catalog feed polling mechanism

malinajirka:
#WOOMOB-1867

### Description

This PR introduces a new action and a feature flag. The code isn't used yet - this will be done in follow up PRs.

The action poll the server (/catalog) endpoint, until 
- the server returns status that the job is completed 
or 
- the maximum number of polling attempts is exceeded
or
- three consecutive requests return a failure (e.g. network down/server down/...)

### Test Steps

Green CI should be enough since the code isn't used yet.

### Images/gif

N/A


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/pos/localcatalog/WooPosGenerateCatalogResult.kt
```diff
@@ -3,11 +3,27 @@ package org.wordpress.android.fluxc.store.pos.localcatalog
...
+    IN_PROGRESS("in_progress"),
+    COMPLETED("completed"),
+    UNKNOWN("");
+
+    companion object {
+        fun from(value: String?): WooPosGenerateCatalogState = when (value) {
+            "scheduled" -> SCHEDULED
+            "in_progress" -> IN_PROGRESS
+            "completed" -> COMPLETED
+            else -> UNKNOWN
```
samiuelson, line 26: Out of curiosity: does the backend also return "error" for the catalog generation successful response, or rather an error response type?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFileBasedSyncAction.kt
```diff
@@ -0,0 +1,116 @@
...
+class WooPosFileBasedSyncAction @Inject constructor(
+    private val posLocalCatalogStore: WooPosLocalCatalogStore,
+    private val logger: WooPosLogWrapper,
+) {
+    companion object {
+        private const val INITIAL_POLL_INTERVAL_MS = 3000L
+        private const val MAX_POLL_ATTEMPTS = 20
+        private const val MAX_CONSECUTIVE_FAILED_ATTEMPTS = 3
+
+        // Backoff settings
```
samiuelson:
```suggestion
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFileBasedSyncAction.kt
```diff
@@ -0,0 +1,116 @@
...
+    @Suppress("ReturnCount")
+    suspend fun syncCatalog(
+        site: SiteModel
+    ): Result<FileBasedSyncResult> {
+        logger.d("Starting file-based catalog generation for site ${site.id}")
+
+        var attemptCount = 0
+        var failedConsecutiveAttempts = 0
+
+        while (attemptCount < MAX_POLL_ATTEMPTS) {
```
samiuelson: 💡 n.p.: `repeat(MAX_POLL_ATTEMPTS) { attemptIndex -> ... }` would be more idiomatic, a bit easier to understand, and maintain the code. However, keep in mind that this suggestion may just be my personal taste, and feel free to ignore it.

samiuelson (approved):
LGTM!

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/pos/localcatalog/WooPosGenerateCatalogResult.kt
malinajirka: If it returns 200 it's always scheduled/success/in-progress state.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFileBasedSyncAction.kt
malinajirka: Good call, replaced in 60307fc982f0e3d6e620af246cb31b5d0cce86d1

PR #15106, opened 2025-12-15
Stop using custom Develocity instance

mokagio:
### Description

Our Develocity 360 trial has expired and we decided to not use the product. 

Internal ref paaHJt-99X-p2

Fixes https://linear.app/a8c/issue/AINFRA-1641

### Test Steps

See CI in the test PR #15107 which touches a file to trigger the lint:

image

(_The failure in the screenshot is from `detkt` because I had two consecutive empty line in the code I added._)

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes. — N.A.

### DIFF
--- a/config/gradle/gradle_build_scan.gradle
+++ b/config/gradle/gradle_build_scan.gradle
@@ -1,15 +1,12 @@
 develocity {
     buildScan {
-        server = "https://develocity.a8c-ci.services/"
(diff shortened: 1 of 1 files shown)


wzieba (approved):
Thanks!

PR #15123, opened 2025-12-17
WOOMOB-1626 Fix WooPos preferences not updating when switching sites

malinajirka:
WOOMOB-1626

### Description
Fixed a theoretical issue that could cause WooPos preferences (recent product and coupon searches) showing data from the previous site after switching sites - it's theoretical since the WooPosPreferencesRepository is not injected into a singleton and is always re-created upon site switch. However, there is no guarantee someone won't inject it into a singleton leading to hard-to-debug/reproduce issues.

The problem was that site-specific preference keys were initialized once at class instantiation and never updated when the site changed. The fix changes the site-specific keys from eagerly-initialized `val` properties to lazily-evaluated `get()` properties, ensuring they always use the current site ID.

### Test Steps
- The unit tests are enough.

### Images/gif
N/A - Backend data fix with no UI changes


samiuelson (approved):
LGTM!

PR #15138, opened 2025-12-23
Fix `siteComponent!!` crash on the dashboard

irfano:
Closes WOOMOB-1843

### Description

This PR resolves a long-standing `NullPointerException` crash in `DashboardDataStore` and `DashboardRepository` occurring during app initialization or site switching.  

`DashboardDataStore` and `DashboardRepository` were attempting to access `selectedSite.siteComponent` using the non-null assertion operator (!!) during their initialization.

If the application started in a state where no site was strictly selected (e.g., fresh install, after logout, or race conditions during startup. I'm not sure  because I couldn't reproduce it), `siteComponent` would be null, causing an immediate crash.

Instead of assuming a site exists at initialization time, I refactored both classes to be Reactive:

Also I updated `SelectedSite` to support this reactive approach.

### Test Steps

Since I can't reproduce the crash, it's not possible to test this PR on running app. But I wrote unit tests to reproduce crash cases.

- [DashboardDataStoreTest](https://github.com/woocommerce/woocommerce-android/commit/77b6bd1611b613a8aa80d1e9d463fe52d807dceb): If you check out this commit and run the test, you can reproduce a `NullPointerException` crash at  `selectedSite.siteComponent!!,` line. Then check out the latest commit and confirm the crash was fixed.
- [Add DashboardRepositoryTest for initialization without crash](https://github.com/woocommerce/woocommerce-android/commit/26c709fa0a97c39e7090ead415b0e2182b66c9b2): If you check out this commit and run the test, you can reproduce a `NullPointerException` crash at  `selectedSite.siteComponent!!,` line. Then checkout the latest commit and confirm the crash was fixed.

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/tools/SelectedSite.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/tools/SelectedSite.kt
@@ -87,18 +87,18 @@ class SelectedSite @Inject constructor(
     @Suppress("DEPRECATION")
     @Synchronized
     fun set(siteModel: SiteModel) {
+        // Create a new site component tied to the lifecycle of the selected site
+        siteComponent = siteComponentProvider.get()
+            .setSite(siteModel)
+            .setCoroutineScope(createSiteCoroutineScope())
+            .build()
+
         wasReset = false
         state.value = siteModel
         PreferenceUtils.setInt(getPreferences(), SELECTED_SITE_LOCAL_ID, siteModel.id)
 
         // Notify listeners
         getEventBus().post(SelectedSiteChangedEvent(siteModel))
-
-        // Create a new site component tied to the lifecycle of the selected site
-        siteComponent = siteComponentProvider.get()
-            .setSite(get())
-            .setCoroutineScope(createSiteCoroutineScope())
-            .build()
     }
 
     @Synchronized
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/DashboardDataStore.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/DashboardDataStore.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.ui.dashboard.data
 
+import androidx.annotation.VisibleForTesting
 import androidx.datastore.core.DataStore
 import com.woocommerce.android.di.SiteComponentEntryPoint
 import com.woocommerce.android.model.DashboardWidget
@@ -53,7 +54,8 @@ class DashboardDataStore @Inject constructor(
         }
     }
 
-    private fun getDefaultWidgets(): List<DashboardWidgetDataModel> {
+    @VisibleForTesting
+    internal fun getDefaultWidgets(): List<DashboardWidgetDataModel> {
         fun DashboardWidget.Type.shouldBeEnabledByDefault() =
             this == DashboardWidget.Type.STATS ||
                 this == DashboardWidget.Type.POPULAR_PRODUCTS ||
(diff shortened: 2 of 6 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/tools/SelectedSite.kt
```diff
@@ -87,18 +87,18 @@ class SelectedSite @Inject constructor(
...
         PreferenceUtils.setInt(getPreferences(), SELECTED_SITE_LOCAL_ID, siteModel.id)
 
         // Notify listeners
         getEventBus().post(SelectedSiteChangedEvent(siteModel))
-
-        // Create a new site component tied to the lifecycle of the selected site
-        siteComponent = siteComponentProvider.get()
-            .setSite(get())
-            .setCoroutineScope(createSiteCoroutineScope())
-            .build()
```
irfano, line 101: I made this change because `state.value = siteModel` triggers observers, and it was causing them to receive the old `siteComponent`. Now, I update `siteComponent` before updating the state so observers get the correct value.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/DashboardDataStore.kt
```diff
@@ -9,31 +10,37 @@ import com.woocommerce.android.ui.mystore.data.DashboardWidgetDataModel
...
+    private val dataStoreFlow: Flow<DataStore<DashboardDataModel>?> = selectedSite.observe().map {
+        selectedSite.siteComponent?.let { component ->
+            EntryPoints.get(component, SiteComponentEntryPoint::class.java).dashboardDataStore()
+        }
+    }
 
-    val widgets: Flow<List<DashboardWidgetDataModel>> = dataStore.data
-        .catch { exception ->
+    val widgets: Flow<List<DashboardWidgetDataModel>> = dataStoreFlow.flatMapLatest { dataStore ->
+        val flow = dataStore?.data ?: flow { throw IOException("Site component is null") }
```
irfano: Now we pass null site component case to the existing `catch` block.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/DashboardDataStore.kt
```diff
@@ -9,31 +10,37 @@ import com.woocommerce.android.ui.mystore.data.DashboardWidgetDataModel
...
+        flow.catch { exception ->
             // dataStore.data throws an IOException when an error is encountered when reading data
             if (exception is IOException) {
                 WooLog.e(T.DASHBOARD, "Error reading dashboard data.", exception)
                 emit(DashboardDataModel.getDefaultInstance())
             } else {
                 throw exception
             }
```
JorgeMucientes, line 45: I'll confess it was a bit hard to follow this piece of code. 
I think we can achieve the similar behavior with some slight changes. Something like: 
```kotlin
    val widgets: Flow<List<DashboardWidgetDataModel>> = dataStoreFlow.flatMapLatest { dataStore ->
        // If dataStore is null, immediately return a flow with default widgets.
        // This is clearer than throwing and then catching an exception.
        if (dataStore == null) {
            WooLog.e(T.DASHBOARD, "DashboardDataStore: Site component is null, providing default widgets.")
            return@flatMapLatest flow { emit(DashboardDataModel.getDefaultInstance()) }
        }
        dataStore.data
} .map { ...
```

Not sure if I missed something there. Wdyt?

JorgeMucientes (approved):
Nicely done @irfano. 
I can't think of a situation that would lead to the crash tbh. There must be some sort of rare race condition upon switching sites. 
Anyway the new implementation is safer while being reactive. Works well I just left a minor suggestion that I believe simplifies the code a bit. Lmk your thoughts.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/DashboardDataStore.kt
irfano: I noticed it was a bit confusing, which is why I added the GitHub comment above.
I applied your suggestion (af71d2d7c173ae2c3cbdff361997852a0b828215) with a small change: I removed the `return@flatMapLatest` since the flow must continues with the `.map` calls below to convert `DashboardDataModel.getDefaultInstance()` into default widgets.

PR #15141, opened 2025-12-25
Persist push notification token id

irfano:
Closes WOOMOB-1903

### Description

We need to store the push token IDs we receive from the `wc-push-notifications/push-tokens` endpoint. This PR handles storing and removing the push token IDs.

### Test Steps

`PushNotificationsStore` isn't usable in the app for now, so just run `PushNotificationsStoreTest`.

### Images/gif

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
@@ -1,44 +1,96 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.pushnotifications
 
+import androidx.core.content.edit
 import org.wordpress.android.fluxc.BuildConfig
 import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.BaseRequest
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
 import org.wordpress.android.fluxc.tools.CoroutineEngine
-import org.wordpress.android.util.AppLog
+import org.wordpress.android.fluxc.utils.PreferenceUtils
+import org.wordpress.android.util.AppLog.T
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class PushNotificationsStore @Inject internal constructor(
     private val pushNotificationsRestClient: PushNotificationsRestClient,
+    private val prefsWrapper: PreferenceUtils.PreferenceUtilsWrapper,
     private val coroutineEngine: CoroutineEngine,
 ) {
+    private val preferences by lazy { prefsWrapper.getFluxCPreferences() }
+
     suspend fun registerPushToken(
         site: SiteModel,
         token: String,
         deviceUuid: String
-    ): WooResult<Unit> = coroutineEngine.withDefaultContext(AppLog.T.API, this, "deletePushToken") {
+    ): WooResult<Unit> = coroutineEngine.withDefaultContext(T.API, this, "registerPushToken") {
         val origin = if (BuildConfig.DEBUG) ORIGIN_DEV else ORIGIN
-        val result = pushNotificationsRestClient.registerPushToken(site, token, origin, deviceUuid)
+        val payload = pushNotificationsRestClient.registerPushToken(site, token, origin, deviceUuid)
 
-        if (result.isError) {
-            WooResult(result.error)
+        if (payload.isError || payload.result == null) {
+            WooResult(payload.error)
         } else {
-            // TODO persist push token id
+            persistPushToken(site, payload.result.id)
             WooResult(Unit)
         }
     }
 
     suspend fun deletePushToken(
         site: SiteModel
-    ): WooResult<Unit> = coroutineEngine.withDefaultContext(AppLog.T.API, this, "deletePushToken") {
-        // TODO Retrieve persisted id
-        val dummyId = 0
-        pushNotificationsRestClient.deletePushToken(site, dummyId).asWooResult()
+    ): WooResult<Unit> = coroutineEngine.withDefaultContext(T.API, this, "deletePushToken") {
+        getPersistedPushToken(site)?.let { pushTokenId ->
+            val result = pushNotificationsRestClient.deletePushToken(site, pushTokenId).asWooResult()
+            if (!result.isError) {
+                clearPersistedPushToken(site)
+            }
+            result
+        } ?: WooResult(
+            WooError(
+                WooErrorType.GENERIC_ERROR,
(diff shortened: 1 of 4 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
```diff
@@ -1,44 +1,96 @@
...
     private val coroutineEngine: CoroutineEngine,
 ) {
+    private val preferences by lazy { prefsWrapper.getFluxCPreferences() }
+
     suspend fun registerPushToken(
         site: SiteModel,
         token: String,
         deviceUuid: String
-    ): WooResult<Unit> = coroutineEngine.withDefaultContext(AppLog.T.API, this, "deletePushToken") {
+    ): WooResult<Unit> = coroutineEngine.withDefaultContext(T.API, this, "registerPushToken") {
```
irfano, line 28: `loggedMessage` was incorrect in the previous PR. I fixed it here.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
```diff
@@ -1,44 +1,96 @@
...
         site: SiteModel
-    ): WooResult<Unit> = coroutineEngine.withDefaultContext(AppLog.T.API, this, "deletePushToken") {
-        // TODO Retrieve persisted id
-        val dummyId = 0
-        pushNotificationsRestClient.deletePushToken(site, dummyId).asWooResult()
+    ): WooResult<Unit> = coroutineEngine.withDefaultContext(T.API, this, "deletePushToken") {
+        getPersistedPushToken(site)?.let { pushTokenId ->
+            val result = pushNotificationsRestClient.deletePushToken(site, pushTokenId).asWooResult()
+            if (!result.isError) {
+                clearPersistedPushToken(site)
```
irfano: I decided not to clear the persisted push token if the network call fails. Since we only use this token to call the delete endpoint, keeping it in persistence poses no risk.
Ideally, we would handle specific API errors and act accordingly, but that is not strictly necessary here. If we were to remove the cached token on every API error, we might lose it permanently (e.g., during a temporary network issue), which would prevent us from ever deleting it from Woo Core.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
```diff
@@ -1,44 +1,96 @@
...
+        val updatedPushTokens = persistedPushTokens.filterNot { it.startsWith(getSiteTokenPrefix(site)) }.toSet()
+
+        if (persistedPushTokens != updatedPushTokens) {
+            preferences.edit { putStringSet(PUSH_TOKENS, updatedPushTokens) }
+        }
+    }
+
+    private fun getPersistedTokenSet() = preferences.getStringSet(PUSH_TOKENS, null) ?: setOf()
+
+    private fun getSiteTokenPrefix(site: SiteModel) = "${site.id}:"
```
irfano: Tokens are persisted in `SharedPreferences` as a `Set<String>` using the `site.id:token` format. This pattern is similar to the one used in `JetpackApplicationPasswordsSupport`.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
```diff
@@ -1,44 +1,96 @@
...
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class PushNotificationsStore @Inject internal constructor(
     private val pushNotificationsRestClient: PushNotificationsRestClient,
+    private val prefsWrapper: PreferenceUtils.PreferenceUtilsWrapper,
     private val coroutineEngine: CoroutineEngine,
 ) {
+    private val preferences by lazy { prefsWrapper.getFluxCPreferences() }
```
JorgeMucientes: May I ask why use FluxC Preferences and not regular WooCommerce preferences? 

Moreover I'm thinking this could benefit from using Android DataStore and expose an observable flow with the sert of registered tokens. Similar to what we do for BookingFilters for example. 

Wdyt?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
JorgeMucientes: I wonder what would happen if we keep the PN token persisted after log out (considering this request fails on log out). Users would keep receiving push notifications for sites they are not logged in anymore? This might be a corner case, but maybe we could have away to keep track of persisted but "disabled" PN tokens, so that we can filter out PN that belong to those sites that should be unregistered but failed to do so. 
Is this overkill? I'm aware we don't have such logic in the current implementation with WP.com tokens. o this might not be necessary at all in the new system.

JorgeMucientes (approved):
LGTM! :shipit:

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
irfano: We can filter out notifications, but the underlying issue would still remain. So I think we should focus on the root cause.

1. As a simple solution, we could add the following safeguard:
- On app startup, if a `push-token` exists and the user is logged out, try deleting it one more time.
- If it fails again, keep the token persisted (so we can retry later).
- If the user logs in while a previous `push-token` is still stored, just overwrite it.
This at least adds a retry attempt and keeps the solution simple.

2. I know there’s a risk with option 1: notifications might still arrive while the user is logged out. There are more robust approaches, like a proper retry mechanism or even a background task, but that may be overkill.

3. Another option is to wait for the `deletePushToken()` response during logout and block logout if it fails. But this would likely hurt the UX because it adds latency to the logout flow. Also, there are some cases where logout is mandatory.

4. We could also ignore this case, since we already have this risk today and we haven’t received any complaints about notifications arriving after logout.

I’m leaning toward option 4. What do you think?

irfano: If we’re going to choose option 4, then I think we should also remove the cached `push-token` after logout, which is missing in this PR. I can add it.

irfano:
I noticed I made a mistake in this PR. I assumed push token IDs should be site-specific, but they aren’t. We only send the device token and receive the push token ID from the backend. So I added a few more commits to fix this.

PR #15143, opened 2025-12-26
When shipping rates loaded, preselect firs option

JorgeMucientes:
Closes WOOMOB-1476

### Description
This change updates WooShippingLabelCreationViewModel so that, after shipping rates are fetched and sorted, the first available rate for a shipment is preselected when no rate has been chosen yet. The previous behavior left selectedRates[index] as null, forcing merchants to manually select a rate option even when only one reasonable default existed. By preselecting the first sorted rate, the shipping label purchase flow becomes faster and less error-prone for merchants.
​
### Test Steps
1. Open an order eligible for Woo Shipping Labels and start the shipping label purchase flow on Android.
2. Proceed to the step where shipping rates are loaded for a shipment. Verify that once rates finish loading, the first rate option is already selected instead of showing no selection.
3. Change the selected rate manually and confirm the UI and state update correctly without overriding the user’s choice.
​
### Images/gif

https://github.com/user-attachments/assets/04255c48-e561-4d6f-b78c-2a8cea3b6ee5


irfano (approved):
LGTM! 👍🏻 
We can merge after fixing detekt and unit tests.

PR #15147, opened 2025-12-29
Hide the Jetpack benefits banner if the push token is registered

irfano:
### Description

This hides the Jetpack benefits banner if the push notification token is registered using the new endpoint. The new endpoint hasn’t been implemented yet, but I’m using the `push_token_id` key in SharedPreferences to determine whether the push token is registered.

This won't affect production since the banner will still be visible because there won't be persisted push token id.

### Test Steps

It’s not possible to test this yet. Once we implement the new endpoint, we’ll retest and verify the behavior. I’m not closing the issue until we confirm it works as expected.

### Images/gif

|Before|After|
|-|-|
|before|after|


JorgeMucientes (approved):
LGTM! :shipit:

PR #15154, opened 2026-01-02
Pass screen dimensions in JITM query params

kidinov:
WOOMOB-1933

### Description

Add screen width and height (in dp) to the JITM query parameters. This allows the server to make decisions based on device screen size.

New parameters added:
- `screen_width` - screen width in density-independent pixels
- `screen_height` - screen height in density-independent pixels

### Test Steps

1. Open the app on a device or emulator
2. Navigate to any screen that triggers JITM requests (e.g., My Store)
3. Verify in network logs that JITM requests include `screen_width` and `screen_height` parameters

### Images/gif

N/A - No UI changes


malinajirka (approved):
Thanks @kidinov! 

Just to double check the expected behavior - `Resources.getSystem().configuration.screenWidthDp` returns width no matter whether it's the longer/shorter dimensions (e.g. in portrait it's the shorter dimension). This is likely desired, but I wanted to double check to be sure.

kidinov:
Thanks @malinajirka !

> Just to double check the expected behavior - Resources.getSystem().configuration.screenWidthDp returns width no matter whether it's the longer/shorter dimensions (e.g. in portrait it's the shorter dimension). This is likely desired, but I wanted to double check to be sure.

Yes, it should be fine - the backend should have logic similar to what we have on the client regarding deciding if the POS is possible to show. We use the short/long side to do so.

PR #15157, opened 2026-01-05
Issue/woomob 1909 add new local feature flag for woo notifications disabled by

JorgeMucientes:
Closes: WOOMOB-1909
CLoses: WOOMOB-1910

### Description
This PR introduces a new feature flag for the new Woo core PN system and wires it into the existing registration logic so that, when enabled, the app uses the new PN registration endpoint instead of the WP.com one.

This PR doesn't include the changes required to display any PN sent through the new system. Neither includes any fallback strategy for cases where PN toke registration fails. That is the reason why the feature flag is disabled by default for now. 
NOT included in this PRs: 
- Add fall back strategy
- Unregister WP.com PN token if successfully registered in the new Woo core system
- Remove PN token using the new endpoint on user log out. 

### Test Steps

1. Install the .zip file version of WooCoomerce containing the new PN registration endpoint. 
2. Make a clean install of the app with the feature flag set to true. 
4. Log in using WP.com credentials
5. Kill the app
6. Restart it with the network inspector enabled
7. Using network inspector verify that the new PN successfully registers


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val selectedSite: SelectedSite,
+    private val appPrefsWrapper: AppPrefsWrapper
+) {
+    suspend fun registerPushToken(token: String) {
+        WooLog.d(
+            tag = WooLog.T.NOTIFS,
+            message = "Registering FCM token in Woo Core instance${if (BuildConfig.DEBUG) ": $token" else ""}"
+        )
+        selectedSite.getIfExists()?.let {
+            val uuid = appPrefsWrapper.wooCorePushDeviceUUID.orNullIfEmpty() ?: generateAndStoreUUID()
```
JorgeMucientes, line 23: For now we are reusing the existing logic to create the device UUID. Asked Hannah here: https://wp.me/pe5sF9-4OM#comment-5081 if the current strategy is good enough.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -26,7 +27,8 @@ enum class FeatureFlag {
             BOOKINGS_MVP,
             POS_REFUNDS -> PackageUtils.isDebugBuild()
 
-            WOO_POS_LOCAL_CATALOG_FILE_APPROACH -> false
+            WOO_POS_LOCAL_CATALOG_FILE_APPROACH,
+            WOO_PUSH_NOTIFICATIONS_SYSTEM -> false
```
JorgeMucientes, line 35: Disabled by default for now to avoid breaking PN for all devs until the endpoint starts working

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -33,7 +36,9 @@ class RegisterDevice @Inject constructor(
 
     private suspend fun sendToken() {
         val token = appPrefsWrapper.getFCMToken()
-        if (accountStore.hasAccessToken() && token.isNotEmpty()) {
+        if (FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM.isEnabled()) {
+            pushNotificationRepository.registerPushToken(token)
+        } else if (accountStore.hasAccessToken() && token.isNotEmpty()) {
```
irfano: We can use the existing `token.isNotEmpty()` check for the new `registerPushToken(token)` call as well.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+            pushNotificationsStore.registerPushToken(
+                site = it,
+                token = token,
+                deviceUuid = uuid
+            )
+            // TODO ensure any WP.com token already registered is removed
+        }
```
irfano: I know it’s not possible to reach this point without a selected site, but we can still add a log here.
```suggestion
        } ?: run { WooLog.w(WooLog.T.NOTIFS, "No selected site, skipping PN registration") }
```

irfano (commented):
LGTM! I just have one question before approving: why do we need to kill and restart the app to trigger the `push-tokens` endpoint? Shouldn’t it be triggered right after the first login?

JorgeMucientes:
> LGTM! I just have one question before approving: why do we need to kill and restart the app to trigger the push-tokens endpoint? Shouldn’t it be triggered right after the first login?


So the problem is that this new enpoint requires a site to be selected first in order to register the PN token. So, on log in, by the time `RegisterDevice.invoke()` is called, there's no selected site yet. But this is not a big deal given PN token registration will be attempted the next time you resume the app (you don't actually have to kill the app, just send it to the background).

irfano (approved):
> So, on log in, by the time RegisterDevice.invoke() is called, there's no selected site yet. But this is not a big deal given PN token registration will be attempted the next time you resume the app (you don't actually have to kill the app, just send it to the background).

This is also the current behavior for the wpcom push registration, right? If so, we can keep improving it outside this project’s scope.

JorgeMucientes:
> This is also the current behavior for the wpcom push registration, right? If so, we can keep improving it outside this project’s scope.

Not really @irfano. Because registering for WP.com push notifications doesn't require a site to be selected. So the device registration for WP.com system will happen right away. In any case agree we can improve this in subsequent PRs and I don't think is a big deal tbh, as soon as they resume the app from background the next time they'll get registered.

PR #15162, opened 2026-01-06
Fix minor bug in JitmStoreInMemoryCache.kt

malinajirka:
I spotted this minor bug in the JitmStoreInMemoryCache so here is a fix. No tests needed, just double-check the code.


kidinov (approved):
Good catch! It has been there since 2023 =)

PR #15165, opened 2026-01-06
Fix notification group summary behavior (double cha-ching) in WooNoti…

AlexKlimaj:
Attempt to fix double cha-ching for every order.

Summary
The problem was in WooNotificationBuilder.kt. When the first order notification arrives:

An individual notification is created and displayed (plays sound ✓)
A group summary notification is also created for non-wearable devices (was playing sound again ✗)
The fix was to call setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN) before setting the group summary. This tells Android to only alert (play sound) for the child notifications, not the summary.

The key change was moving the line from after setGroupSummary(true) to before it, which ensures the alert behavior is set before the summary is created. This prevents the duplicate sound while still maintaining proper notification grouping functionality.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
@@ -117,8 +117,8 @@ class WooNotificationBuilder @Inject constructor(
             // Also add a group summary notification, which is required for non-wearable devices
             // Do not need to play the sound again. We've already played it in the individual builder.
             if (!isGroupNotification) {
-                setGroupSummary(true)
                 setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
+                setGroupSummary(true)
                 showNotification(notification.getGroupPushId(), notification, this)
             }
         }


JorgeMucientes:
Hey @AlexKlimaj thanks for surfacing this issue and attempting to provide a fix for it. 

I was not able to reproduce the double "cha-ching" issue, so I suspect it is likely an issue with how different OEM/Android devices handle some push notification code. 

The fix attempt you provided though, I don't think will make any difference. Have you tested that it fixed the bug? Because changing the order on how some values are applied in `builder` pattern shouldn't really matter, because it is not until `showNotification()` is called that the values actually take effect. 

That being said this is what I suspect is happening is the following:

1. handleWooNotification() is called with isGroupNotification = false
2. buildAndDisplayWooNotification() is called:
  - Sets setDefaults(NotificationCompat.DEFAULT_ALL) → enables sound
  - showNotification() → First cha-ching (from channel)
  - Since isGroupNotification = false:
      - Sets setGroupSummary(true) and setGroupAlertBehavior(GROUP_ALERT_CHILDREN)
      - showNotification() again → Should NOT play sound due to GROUP_ALERT_CHILDREN

**Why GROUP_ALERT_CHILDREN Might Not Work:**
The issue is that GROUP_ALERT_CHILDREN is set on the same builder that was already used to show the first notification. The builder still has:
- Segfaults(NotificationCompat.DEFAULT_ALL)
- The notification channel with cha-ching sound
 
Possible causes:

1. Builder reuse issue: The builder is reused and .build() is called twice. The second call might still inherit properties that cause sound.
2. Channel-level sound override: On some Android versions/OEMs, the channel's sound setting might play regardless of GROUP_ALERT_CHILDREN.
3. setDefaults(DEFAULT_ALL) conflict: This might override or conflict with GROUP_ALERT_CHILDREN on some devices.

I think adding the following code: 
```kotlin
            if (!isGroupNotification) {
                setGroupSummary(true)
                setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
                setDefaults(0)
                setSound(null)
                setVibrate(null)
                showNotification(notification.getGroupPushId(), notification, this)
            }
```
Will fix the issue for all OEMs. I've opened [a new PR with the above changes. ](https://github.com/woocommerce/woocommerce-android/pull/15208)

I'll proceed to close this PR though. Let me know if you have any other feedback. And once again thanks for contributing to making the Woo app better.

PR #15170, opened 2026-01-08
[WOOMOB-1953] Show JITMs regardless of onboarding status

kidinov:
WOOMOB-1953

### Description

Previously, JITMs (Just-In-Time Messages) were only displayed on the Dashboard when the onboarding widget was not visible. This change removes that condition so JITMs are always requested and shown regardless of whether onboarding is in progress.

### Test Steps

1. Open the app with a fresh install or reset onboarding
2. Go to the Dashboard (My Store tab)
3. Verify that JITMs appear even when the onboarding checklist is visible

image


samiuelson (approved):
LGTM

PR #15172, opened 2026-01-08
Disable WP.com Push Notification When Successfully Registered for Woo Self-Serve PN

JorgeMucientes:
Closes: WOOMOB-1954

⚠️ Don't panic about the many file changes. Some of them are simple class renames. I'll provide more comments and context in the most relevant changes. 
The implementation is based on the P2 and its summarized comment: pe5sF9-4ZT-p2
I suggest starting the review from `PushNotificationRepository.kt` class which is the central piece and where most of the new logic concentrates. 

### Description

This PR refactors the push notification registration system to support the new Woo Core self-driven push notifications:

- **New registration flow**: When `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag is enabled, push tokens are registered via the new Woo Core endpoint instead of WPCom
- **Token storage migration**: Push tokens are now stored per-site in DataStore instead of SharedPreferences in `PushNotificationsStore`
- **Automatic WPCom notification disabling**: After successful Woo Core registration, WPCom notifications are disabled for that site to prevent duplicates
- **Proper logout cleanup**: `unregisterDeviceFromAllPushes()` now calls the API to delete tokens from the server for all sites before clearing local storage
- **Simplified `PushNotificationsStore`**: Removed SharedPreferences-based token persistence; `deletePushToken()` now accepts `pushTokenId` as parameter
- **New `PushNotificationRegistrationStatus` class**: Centralizes the logic to determine if a device is registered for push notifications (checks both legacy WPCom and new Woo Core systems) 

#### Prerequisites
- Clean install of the app (or clear app data)
- A site with Woo self-serve push notifications available (use the zip file provided from Hannah)
- Network inspector tool (e.g. AS App Inspector should suffice) configured

#### 1. Feature flag disabled (default behavior - legacy WPCom registration)

1. Ensure `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag is **disabled** (default)
2. Fresh install the app or clear app data
3. Log in with a WPCom account
5. **Verify**: Check Logcat for "Registering FCM token in WPCOM instance" log message
6. **Verify**: Push notifications should work as before
7. Log out from the app
8. **Verify**: Check Logcat for "WPCom push token unregistered" log message
9. **Verify**: No crashes occur during logout

#### 2. Feature flag enabled (new Woo Core registration)

1. Enable `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag
2. Fresh install the app or clear app data
3. Log in with a WPCom account
4. Attempt to register PN will fail given there's no selected site yet
5. Send app to the background and resume it
6. Verify the following logcat sequence: 
```
Registering FCM token in Woo Core instance: fXLX3AfJR26M...
No selected site, skipping PN registration
Push notifications registration status: UNREGISTERED
Registering FCM token in Woo Core instance: fXLX3AfJR26M...
Failed to disable WPCom notifications for site 251754054
Push notifications registration status: REGISTERED
```

#### 3. Logout with multiple sites (token cleanup)

1. Log in the app using WP.com credentials and the feature flag disabled.
2. Enable `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag and re run the app
3. Switch between 2 different WooCommerce sites with the DEV Woo Core version of the plugin to register at least 1 PN token in WooCore system. 
5. Log out from the app
6. **Verify**: Check Logcat for "Woo Core push token deleted for site [siteId]" messages for each site
10. **Verify**: Check Logcat for "Woo Core push notification tokens cleared from DataStore" message
11. **Verify**: Check Logcat for "WPCom push token unregistered" message
12. **Verify**: No crashes occur during logout

---

### Other Notes

- This PR doesn't address the logic need to trigger Woo Core PN registration for users logged in with site credentials (even if their site is Jetpack CP connected) . It will only trigger registration for users logging. Linear task: WOOMOB-2034
- Notification settings API is still missing so we can't handle disabling PN in Woo core system for sites hidden in the site picker like we do for WP.com. Linear task: WOOMOB-1956
- When switching sites, PN registration logic is not triggered automatically. The app has to be sent to the background and then reopened. This is something that we need to fix as it happens for the WP.com token registration as well. But is out of scope for this PR


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsStore.kt
```diff
@@ -1,75 +1,42 @@
...
-    }
-
-    suspend fun deletePushToken(
-        site: SiteModel
-    ): WooResult<Unit> = coroutineEngine.withDefaultContext(T.API, this, "deletePushToken") {
-        getPersistedPushTokenId()?.let { pushTokenId ->
-            val result = pushNotificationsRestClient.deletePushToken(site, pushTokenId).asWooResult()
-            if (!result.isError) {
-                preferences.edit { remove(PUSH_TOKEN_ID) }
+    ): WooResult<String> =
```
JorgeMucientes, line 21: I've moved the saving of the token out of this class into PushNotificationRepository.kt to have a central place where tokens are managed and exposed.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -180,6 +180,22 @@ class NotificationRestClient @Inject constructor(
             })
         add(request)
     }
+
+    suspend fun unregisterDevice(deviceId: String): UnregisterDeviceResponsePayload {
```
JorgeMucientes: This is similar to the existing function `unregisterDeviceForPushNotifications` just above, but I've refactored it into coroutine and suspendable function to stay away from the event but, both at the rest client and the store levels.

irfano:
> 7. Verify: Check Logcat for "WPCom push token unregistered" log message

The log message in the test steps seems incorrect. I'm seeing "Unregister device action succeeded" instead of "WPCom push token unregistered."

> 5. Verify: Check Logcat for "Woo Core push token deleted for site [siteId]" messages for each site

I couldn't trigger this log either. The rest api returned a 404 error: `message=Push token could not be found., apiErrorCode=rest_invalid_push_token`. I guess this is because the Woo endpoints are still a work in progress.

I'll move on to the code review.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRegistrationStatus.kt
```diff
@@ -0,0 +1,34 @@
...
+            } else {
+                Status.UNREGISTERED
+            }
+        } else {
+            val deviceId = prefsWrapper.getFluxCPreferences()
+                .getString(NotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
+            when (deviceId.isNullOrEmpty()) {
+                true -> Status.UNREGISTERED
+                false -> Status.REGISTERED
+            }
```
irfano: I'd suggest switching this to if/else to align with [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#if-versus-when)

> Prefer using if for binary conditions instead of when.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -1,35 +1,134 @@
...
 
+    private suspend fun disableWpComNotificationsForSite(siteId: Long) {
+        val setting = SiteNotificationSetting(
+            siteId = siteId,
+            newCommentEnabled = false,
+            storeOrderEnabled = false
+        )
+        val result = notificationStore.updateNotificationSettingsFor(listOf(setting))
+        if (result.isFailure) {
+            // TODO we may want to add tracking to check how often this happens and if a fallback strategy is needed
```
irfano: 👍🏻  We're already planning to track the `*_wpcom_device_disable_push_notifications_error` and `_wpcom_device_disable_push_notifications_success` events here.

irfano (approved):
LGTM! Solid refactoring work! 🎉 Left a few minor comments, but approving.

JorgeMucientes:
Thanks for the review @irfano I've applied both of your suggested code changes.

About the following: 
> The log message in the test steps seems incorrect. I'm seeing "Unregister device action succeeded" instead of "WPCom push token unregistered."

Yeah I had initially added new logs for unregistering from WP.com but then I noticed that there were already logs in place. In any case in my last commit I've updated the existing logs for unregistering device to make it explicit that we are unregistering from WP.com 

> I couldn't trigger this log either. The rest api returned a 404 error: message=Push token could not be found., apiErrorCode=rest_invalid_push_token. I guess this is because the Woo endpoints are still a work in progress.

I've reported this to Hannah here: p91TBi-dyW-p2#comment-14594. However, I've seen  cases where this endpoint worked well for me and some other where I got a 404 error. I'll do some more testing. 

For now, I'll proceed with merging this

PR #15174, opened 2026-01-08
[WOOMOB-1949] Wire file-based sync action and use shared result type

malinajirka:
Fixes WOOMOB-1949

### Description
Wires `WooPosFileBasedSyncAction` into the repository when `WOO_POS_LOCAL_CATALOG_FILE_APPROACH` feature flag is enabled, and refactors it to return wrapper around PosLocalCatalogSyncResult instead of `Result<FileBasedSyncResult>`.

This eliminates the need for error mapping in the repository layer - the action now handles all error cases and returns the appropriate `PosLocalCatalogSyncResult.Failure` types directly (NetworkError, InvalidResponse, DatabaseError, CatalogGenerationTimeout).

Note: There are still a lot of things that need to be done before this code is ready for production. For example:
- The 1000 products limit must be removed
- Background worker needs to properly handle scenario in which polling runs out of maximum attempts
- Checks for catalog size must be skipped when file based approach is enabled
- The flag stored in data store for skipping local catalog must be reseted on store with catalog large then 1000 items
- ....

### Test Steps

IMPORTANT - Make sure to use supported version of Woo on your site. You can use Jetpack Beta plugin and select `issue/woomob-1894-add-pos-visibility-taxonomy-to-core` branch of Woo core.

**Test file-based sync (new behavior):**
1. Enable `WOO_POS_LOCAL_CATALOG_FILE_APPROACH` feature flag
2. Clear app data
3. Log in
4. Observer network requests and verify the file is downloaded and incremental update fetches just first page with 0 results.
5. Open POS and verify products are visible

**Test for regressions (paginated sync):**
1. Disable `WOO_POS_LOCAL_CATALOG_FILE_APPROACH` feature flag
2. Clear app data
3. Log in and open POS
4. Verify products load successfully

### Images/gif
N/A - Internal sync logic changes


samiuelson (approved):
LGTM!

PR #15180, opened 2026-01-09
Replace Jetpack banner with a WP dashboard card UI

irfano:
Closes WOOMOB-1921

### Description

Adds a new "Never miss a new order" dashboard card that prompts users to connect their store to WordPress.com for push notifications. The card is implemented as a configurable widget that can be hidden and reordered via the Customize screen.

The card is currently gated behind FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM and will be shown when the flag is enabled. The onClick action is currently a no-op and will be implemented in a follow-up PR.

### TC1: Feature flag disabled, authenticated with app passwords to site without support of new PN.
1. Disable the feature flag `WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` and build the app.
2. Log in to a site without support of new PN with site credentials.
3. Confirm that the new dashboard card is **not available** in the Dashboard, and Jetpack banner is displayed.
4. Confirm that the card is not visible in the Customize screen.

### TC2: Feature flag enabled, authenticated with app passwords to site without support of new PN.
1. Enable the feature flag `WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` and build the app.
2. Log in to a site without support of new PN with site credentials.
3. Confirm that the new dashboard card **is displayed**, and Jetpack banner is hidden.
5. Tap the overflow menu on the card and select "Hide 'Enable push notifications'".
6. Confirm that the card is hidden and will not display again after refreshing or relaunching.
7. Confirm the card can be re-enabled from the Customize screen.

### TC3: Feature flag enabled, authenticated with app passwords to site with support of new PN.
1. Enable the feature flag `WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` and build the app.
2. Log in to a site **with** support of new PN with site credentials.
3. Confirm that the new dashboard card is **not available** (site is already registered), and Jetpack banner is not displayed.

### TC4: Feature flag enabled, authenticated with WPCom to site without support of new PN.
1. Enable the feature flag `WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` and build the app.
2. Log in to a site without support of new PN with **WPCom credentials**.
3. Confirm that the new dashboard card is **not available**, and Jetpack banner is not displayed.

### Images/gif

Screenshot_20260109_153453
Screenshot_20260109_153457
Screenshot_20260128_154506
Screenshot_20260128_154410


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.woocommerce.android.tools.connectionType
+import com.woocommerce.android.util.FeatureFlag
+import javax.inject.Inject
+
+/**
+ * Checks whether the "Enable Push Notifications" UI should be shown.
+ *
+ * This is part of the Woo Core push notifications system for app-password authenticated sites.
+ */
+class ShouldShowEnablePushNotificationsUi @Inject constructor(
```
JorgeMucientes, line 21: Np, it be nice to have unit tests added for this class.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
```diff
@@ -0,0 +1,28 @@
...
+    private val selectedSite: SelectedSite,
+    private val pushNotificationRegistrationStatus: PushNotificationRegistrationStatus
+) {
+    suspend operator fun invoke(): Boolean {
+        if (!FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM_M2.isEnabled()) return false
+
+        val site = selectedSite.getIfExists() ?: return false
+        if (site.connectionType != SiteConnectionType.ApplicationPasswords) return false
+
+        return pushNotificationRegistrationStatus(site.siteId) != Status.WOO_REGISTERED ||
```
JorgeMucientes: Np, `pushNotificationRegistrationStatus()` function access disk several times via app prefs and datastore. It be nice to call this function only once and keep the returned value in a function variable.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
```diff
@@ -0,0 +1,28 @@
...
+            pushNotificationRegistrationStatus(site.siteId) != Status.REGISTERED_IN_BOTH
```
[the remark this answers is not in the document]
JorgeMucientes: This is a great catch and major issue that needs to be fixed before merging.  TC3 doesn't work with this logic.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/pushnotifications/DashboardPushNotificationsCard.kt
```diff
@@ -0,0 +1,88 @@
...
+import com.woocommerce.android.ui.compose.component.getText
+import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
+import com.woocommerce.android.ui.dashboard.DashboardViewModel.DashboardWidgetMenu
+import com.woocommerce.android.ui.dashboard.defaultHideMenuEntry
+
+@Composable
+fun DashboardPushNotificationsCard(
+    modifier: Modifier = Modifier,
+    onHideClicked: () -> Unit,
```
JorgeMucientes: Minor NP, but modifier should be placed last as it is the [first optional parameter](https://github.com/woocommerce/woocommerce-android/blob/trunk/docs/compose.md#composable-functions-best-practices--): 
> We will adopt [official guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter) on how to pass the Modifier as parameter: "MUST be named "modifier" and MUST appear as the first optional parameter in the element function's parameter list".

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/ObservePushNotificationsWidgetStatus.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.woocommerce.android.model.DashboardWidget
+import com.woocommerce.android.notifications.push.ShouldShowEnablePushNotificationsUi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import javax.inject.Inject
+
+class ObservePushNotificationsWidgetStatus @Inject constructor(
+    private val shouldShowEnablePushNotificationsUi: ShouldShowEnablePushNotificationsUi
+) {
+    operator fun invoke(): Flow<DashboardWidget.Status> = flow {
```
JorgeMucientes: This function is called from `DashboardRepository,kt` which has `@ActivityRetainedScoped`. This means that when we log in with site credentials into a site that supports Woo PN notifications we'll always display the new "connect to WP.com" banner even after registering in Woo core PN system. Basically TC3 doesn't work with the current implementation even after fixing the issue reported [above](https://github.com/woocommerce/woocommerce-android/pull/15180/changes#r2757819770). We probably need to refresh the PNWidgetStatus after successfully registering in Woo system. 

https://github.com/user-attachments/assets/3e340b90-92d2-46f4-a2ce-341ce08dd111

JorgeMucientes (changes requested):
Hey @irfano I've noticed a few issues that are worth checking particularly around TC3 not working.  Some are NPs but some others are blocking issues.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/pushnotifications/DashboardPushNotificationsCard.kt
irfano: 👍🏻  Done: 194efdd7cdddc8e96a17ce543caabd9f78d35720

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/ObservePushNotificationsWidgetStatus.kt
irfano: Good catch @JorgeMucientes! This isn't actually reproducible in https://github.com/woocommerce/woocommerce-android/pull/15283/ but I still think it's better to apply your suggestion. Done here: c61cd28668e8a0b2a15d361cecadfc77ba31181d

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
irfano: This logic has changed in my latest commits.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
irfano: Since `FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` is a local feature flag that defaults to false and cannot be mocked in tests, we cannot write meaningful unit tests for this class when the flag is disabled.

We could add a wrapper similar to `IsRemoteFeatureFlagEnabled` for local feature flags, but given that a feature flag refactor (WOOMOB-2129) is already planned in the near future, I'd prefer not to introduce this change at this point.

irfano:
Thanks for your review, @JorgeMucientes! Ready for another round.
While addressing your feedback, I noticed another issue: we weren't clearing the Woo push token when the user was logged in with site credentials. I fixed it here: 0b88f0fe6a1143d119ca4cefc7869e8d07e6e2f3

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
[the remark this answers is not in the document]
irfano: Applied Copilot's suggestion.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/ObservePushNotificationsWidgetStatus.kt
JorgeMucientes: Nice refactoring into reactive approach. Works much better 👏🏼

JorgeMucientes (approved):
Great job @irfano code changes look good and everything works as expected now. Thanks for applying the changes. 
:shipit:

PR #15185, opened 2026-01-12
Issue/woomob 1737 toolbar not hiding after material 1130 upgrade

JorgeMucientes:
Fixes: WOOMOB-1737

### Description
This PR fixes toolbar visibility issues that appeared after upgrading Material library to version `1.13.0`. See examples of the issue: 

https://github.com/user-attachments/assets/2efa5573-2148-49db-a781-20d684a3791e

Possible explanation for the problem (AI based):
After the Material library upgrade, the CollapsingToolbarLayout was not hiding/showing properly when navigating between fragments. The previous implementation used MainAnimatorHelper.animateToolbarHeight() with ValueAnimator to manually animate the toolbar height via updateLayoutParams, but this approach became unreliable with the new Material version.

**Solution:**
Fixing the above issue was straightforward. However depending on the approach adopted some other glitches showed up. One that was super tricky to solve was this flickering effect  on collapsing toolbar when navigating back to Dashboard from order or product tab:

https://github.com/user-attachments/assets/db42c90d-5558-4909-8232-3d338aa9b9fe

It looks like the collapsing toolbar is redrawn twice. I suspect there's a problem with too many animations applied over the toolbar at once (expand toolbar + expand Appbar + expand subtitle + expand subtitle margin). So, after testing dozens of options I opted for simplifying some code and remove some of the animations as well as ensuring some other animations weren't triggered more than once while still running. This is the resulting effect: 

https://github.com/user-attachments/assets/a7234666-9072-41e6-971f-3c7c2892ebd9

Might not be perfect but honestly I already spent way too much time on this, and Claude Code wasn't that helpful coming up with a solution that would keep all the existing animation and resolve the flickering UI effects. 

### Test Steps
1. Open the app and navigate to the Dashboard tab
2. Tap on edit dashboard widgets and see toolbar is shown correctly
3. Open order detail and see toolbar is shown as expected


libs/login/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
```diff
@@ -275,7 +274,7 @@ protected void setupBottomButton(Button button) {
     @NonNull
     private Spanned formatTosText(int stringResId) {
         final int primaryColorResId = ContextExtensionsKt.getColorResIdFromAttribute(getContext(),
-                com.google.android.material.R.attr.colorPrimary);
+                androidx.appcompat.R.attr.colorPrimary);
```
JorgeMucientes, line 277: Unrelated to the fix. Just a compile issue after upgrading material library to 1.13.0

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -620,12 +594,23 @@ class MainActivity :
...
         animatorHelper.animateCollapsingToolbarMarginBottom(show = true) {
             binding.collapsingToolbar.expandedTitleMarginBottom = it
+            if (binding.collapsingToolbar.expandedTitleMarginBottom ==
+                resources.getDimensionPixelSize(R.dimen.expanded_toolbar_bottom_margin_with_subtitle)
+            ) {
+                binding.toolbarSubtitle.show()
+                binding.toolbarSubtitle.animate()
+                    .scaleY(1f)
+                    .setDuration(EXPAND_COLLAPSE_ANIMATION_DURATION_MILLIS)
+                    .start()
```
JorgeMucientes, line 612: I'm using this custom "expand" animation instead of the existing "expand()" function, because the existing one will expand the subtitle from the bottom to the top, instead of top to bottom which is what we want here.

kidinov (approved):
Thanks @JorgeMucientes !

Maybe not ideal, but it was not ideal before, so I think it's good enough. Maybe someone can dedicate some time during hack weeks to tune this even better.

PR #15187, opened 2026-01-12
[WOOMOB-1963] Display variation attributes on POS order detail screen

malinajirka:
Fixes WOOMOB-1963

### Description
Display variation attributes (e.g., size, color) for line items on the POS order detail screen. Previously, only the parent product name was shown, while the checkout/cart view correctly displayed variation attributes. Also centers the product image vertically to align with the expanded content.

### Test Steps
1. Add a variable product to cart in POS
2. Complete checkout
3. Open Historical Orders
4. Verify variation attributes appear on order detail screen below product name
5. Verify simple products without variations still display correctly

### Images/gif

| Before | After |
|--------|-------|
| image | Screenshot 2026-01-12 at 18 25 10 |


toupper (approved):
LGTM! Thanks for working on this @malinajirka!

PR #15189, opened 2026-01-13
Enable remote build cache for local builds by default

malinajirka:
### Description
Changes the default value of `use_remote_build_cache_locally` from `false` to `true` in the example properties file. This improves build times for new developer setups out of the box.

### Images/gif
N/A - configuration change only


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kidinov (approved):
Thanks! 

Maybe update the .java-version file as well as you proposed? I think this is a good idea

.java-version
```diff
@@ -1 +1 @@
-21.0
+21.0.6-corretto
```
malinajirka: @wzieba I tried to also update the .java-version file, but it seems this version is not installed on CI 🤔. I'm not sure how is that possible, since that's the version enforced for the remote build cache.

Is the reason that the version has a different name on CI or is there a version discrepancy?

wzieba: I just checked and it's available on CI but named differently: `corretto64-21.0.6` https://buildkite.com/automattic/woocommerce-android/builds/34893/steps/canvas?jid=019bbbdc-df76-4d88-9dfb-6b10ab6ce0b7#019bbbdc-df76-4d88-9dfb-6b10ab6ce0b7/L237

I'm guessing it might be a mismatch between Linux and macOS JDK metadata - it's not the first time jEnv interprets these differently. I'll add the other name to `.java-version` to see if it'll fix the problem (I hope it supports more than 1 version)

wzieba: It didn't work: [c11096e](https://github.com/woocommerce/woocommerce-android/pull/15189/commits/c11096e622a8376ed7ffe0e134b7f063bb34c21e) - I think we have to stick to a name without `vendor` in it 😕. 

In theory, we could probably rename things on CI/Linux, but it's not trivial and releasing a new AMI is quite time-consuming. 

I'd propose 
```suggestion
21.0.6
```

And the check in `settings.gradle` will fail the build if a user wants to run build with Remote Build Cache but with JDK of different vendor than `amazon`.

How does it sound to you @malinajirka ?

malinajirka: Sounds good - let's set it to 21.0.6 as you are suggesting - the vendor is checked in gradle, so folks won't be able to build anyway unless they disable cache or install the correct jdk. Thanks for investigating this 🙇

PR #15195, opened 2026-01-14
Refresh selected order after refund dialog is dismissed

samiuelson:
WOOMOB-1955

### Description

After creating a refund in WooPos Orders, the selected order was not being refreshed when the refund dialog was dismissed. This meant the order details shown to the user didn't reflect the newly created refund, requiring manual refresh or reselection to see the updated state.

This PR adds a call to `refreshSelectedOrder()` when the refund dialog is dismissed, ensuring the order details are automatically updated to reflect the refund that was just created.

### Test Steps

  1. Open WooPos and navigate to Orders
  2. Select an order that can be refunded
  3. Tap "Issue Refund" button
  4. Complete the refund creation process
  5. Dismiss the refund dialog
  6. Verify that the order details automatically refresh and display the newly created refund

### Images/gif

—

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
@@ -227,6 +227,7 @@ class WooPosOrdersViewModel @Inject constructor(
         _state.value = currentState.copy(
             dialogState = WooPosOrdersState.Content.DialogState.Hidden
         )
+        refreshSelectedOrder()
     }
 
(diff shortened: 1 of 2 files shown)


kidinov (approved):
LGTM! 

But, just to make sure that it's expected, you do the network call even when there is no refund done and the dialog was just dismissed. I think it can be fine for the simplicity of the code. Wdyt?

samiuelson:
> LGTM!
> 
> But, just to make sure that it's expected, you do the network call even when there is no refund done and the dialog was just dismissed. I think it can be fine for the simplicity of the code. Wdyt?

Thanks for the review and suggestion. I decided to go for the simple approach for now, because it's quite likely we will redesign the UX and change the way of passing events between VMs.

PR #15197, opened 2026-01-15
Distribute PR prototype builds via FAD and on-demand

mokagio:
https://linear.app/a8c/issue/AINFRA-1742

---

### Description

Unifies the prototype builds automation for `main` and PRs so that:

- We can trigger builds from `main` and PRs
- Build for mobile are distributed via Firebase
- Build for wear are stored in S3 with install link in PRs

### Test Steps

image

image

### Images/gif

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes. — N.A.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
@woocommerce/android-developers can I get some of you to try and install the Wear app following the Firebase link?

The installation fails on my test device, but I wonder if that's because I don't have an Android Wear device connected (_do they pair like iOS-watchOS? I don't even know_)

Truth be told, I don't expect it to work, because my understanding of Firebase App Distribution is that it's for mobile apps only. But it would be great to rule that one out.

I also wonder if there's a way to get the APK out of FAD without attempting the installation. Then one could use `adb` to install manually. However at that point maybe a download from S3 would be more straightforward.

Thanks for your help!

JorgeMucientes:
Hey @mokagio I have never tried this tbh. I don't even have an Android watch (I use watch emulators to test the app), but digging around this I've learned: 
- It’s not directly supported to install a Wear OS (Android Wear) app on the watch via Firebase App Distribution; Firebase only targets phone/tablet (and iOS) clients, not Wear OS devices. [There's an ignored feature request around this. ](https://firebase.uservoice.com/forums/948424-general/suggestions/48812501-wearos-support-for-app-distribution)

> I also wonder if there's a way to get the APK out of FAD without attempting the installation. Then one could use adb to install manually. However at that point maybe a download from S3 would be more straightforward.

Yes I think so. 

From the phone, transfer and install the Wear APK to the watch using:

- A tool like Wear Installer, which can push APKs to the watch over ADB from the phone.

- Or classic ADB sideload from a PC: enable Developer options and wireless ADB on the watch, then run `adb connect <watch-ip>` and `adb install wear-app.apk`. 

- Or what I've tried in the past: From a phone device, download the apk from FAD as if you were to install it. Then from Files app, navigate to the downloaded .apk file and send it to you PC via Slack/Mail/etc whatever means to share the file you want to use. Then drag and drop the file into the adb connected device. Hope this helps. 



https://github.com/user-attachments/assets/062f730c-c50b-463e-9c39-dc4fc3c8a064



​

hichamboushaba:
While (as Jorge mentioned) the official status is that Firebase App Distribution doesn't support Wear OS, we can actually trick it to install the tester app manually, which gives us access to the test builds:

https://github.com/user-attachments/assets/683118a9-0e30-481d-8c35-40a1720b4261

This is not a straightforward process though, and the main issue is that given the watch can't scan QR codes, it's not gonna be easy to find the correct build.

> I also wonder if there's a way to get the APK out of FAD without attempting the installation. Then one could use adb to install manually. However at that point maybe a download from S3 would be more straightforward.

I think providing the APK link directly would be easier for us, as we can just drag it to the watch screen (in Android Studio) and install the APK, without dealing with any adb commands.

mokagio:
@hichamboushaba 

> I think providing the APK link directly would be easier for us, as we can just drag it to the watch screen (in Android Studio) and install the APK, without dealing with any adb commands.

Noted. Implemented via [1ff9c89](https://github.com/woocommerce/woocommerce-android/pull/15197/commits/1ff9c89374add1c86880d2b1dc30886d9372e8dd)

.buildkite/pipeline.yml
```diff
@@ -65,18 +65,31 @@ steps:
           - "**/build/reports/diff_manifest/**/**/*"
 
   ########################################
-  - group: "🛠 Prototype Builds"
+  - group: ":rocket: Prototype Build"
```
mokagio:
```suggestion
  - group: "🚀 Prototype Build"
```

fastlane/Fastfile
```diff
@@ -82,14 +81,19 @@ VERSION_FILE = Fastlane::Wpmreleasetoolkit::Versioning::AndroidVersionFile.new(v
 FIREBASE_APP_ID = '1:124902176124:android:4046dc320bf1825b1620f9'
 FIREBASE_TESTERS_GROUP = 'woocommerce-android---prototype-builds'
 
+########################################################################
+# S3 Prototype Builds
+########################################################################
```
mokagio:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -65,18 +65,31 @@ steps:
...
+            hint: "Which WooCommerce app should be built? 📱"
+            required: true
+            default: "WooCommerce"
+            options:
+              - label: "📱 WooCommerce Mobile"
+                value: "WooCommerce"
+              - label: "⌚️ WooCommerce Wear"
+                value: "WooCommerce-Wear"
+              - label: "📱⌚️ Both"
+                value: "Both"
```
AliSoftware, line 85: ❤️  Nice inclusion of an option to build both 👍

fastlane/Fastfile
```diff
@@ -82,14 +81,16 @@ VERSION_FILE = Fastlane::Wpmreleasetoolkit::Versioning::AndroidVersionFile.new(v
 FIREBASE_APP_ID = '1:124902176124:android:4046dc320bf1825b1620f9'
 FIREBASE_TESTERS_GROUP = 'woocommerce-android---prototype-builds'
 
+PROTOTYPE_BUILD_DOMAIN = 'https://cdn.a8c-ci.services'
```
AliSoftware, line 86: 💄 
```suggestion
# NOTE: Only Wear apps are uploaded to S3 and exposed via CDN.
# FAD doesn't support Android Wear apps (yet?)
PROTOTYPE_BUILD_DOMAIN = 'https://cdn.a8c-ci.services'
```

fastlane/Fastfile
```diff
@@ -82,14 +81,16 @@ VERSION_FILE = Fastlane::Wpmreleasetoolkit::Versioning::AndroidVersionFile.new(v
...
+
 ########################################################################
 # Environment
 ########################################################################
 Dotenv.load('~/.wcandroid-env.default')
-GITHUB_REPO = 'woocommerce/woocommerce-android'
 DEFAULT_BRANCH = 'trunk'
 REPOSITORY_NAME = 'woocommerce-android'
 GH_ORG_NAME = 'woocommerce'
+GITHUB_REPO = "#{GH_ORG_NAME}/#{REPOSITORY_NAME}".freeze
```
AliSoftware, line 95: ❤️

fastlane/Fastfile
```diff
@@ -1349,89 +1350,93 @@ platform :android do
   #####################################################################################
   # build_and_upload_prototype_build
   #
-  # This lane builds a prototype build and uploads it to S3
+  # This lane builds a prototype build and uploads it.
+  # - Mobile app is uploaded to Firebase App Distribution (FAD)
+  # - Wear app is uploaded to S3
```
AliSoftware, line 1357: ❤️

fastlane/Fastfile
```diff
@@ -1349,89 +1350,93 @@ platform :android do
...
       app_display_name: "#{app} Android",
+      app_icon: app == WEAR_APP ? ':amazon-s3:' : nil,
       download_url: install_url,
       metadata: {
         Platform: platform_name,
         Flavor: prototype_flavor,
         'Build Type': prototype_build_type
       },
-      fold: app == WEAR_APP # Keep comment for Mobile app always expanded, but collapse comment for Wear app by default, as it's modified less often
+      fold: app == WEAR_APP,
```
AliSoftware, line 1395: 💄  Any reason for removing the justification comment?

```suggestion
      fold: app == WEAR_APP, # Keep comment for Mobile app always expanded, but collapse comment for Wear app by default, as it's modified less often
```

fastlane/Fastfile
```diff
@@ -1722,3 +1727,45 @@ def report_milestone_error(error_title:)
...
+  pr_number = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
+  return unless pr_number && pr_number != 'false'
+
+  comment_args = {
+    app_display_name: app_display_name,
+    download_url: download_url,
+    metadata: metadata,
+    fold: fold
+  }
+  comment_args[:app_icon] = app_icon if app_icon
```
AliSoftware: 💄 Suggested alternative:

```suggestion
  comment_args = {
    app_icon: app_icon,
    app_display_name: app_display_name,
    download_url: download_url,
    metadata: metadata,
    fold: fold
  }.compact
```

AliSoftware (approved):
I meant to Approve not just Comment 😅 

Regarding the Wear app, it's a shame FAD doesn't support it indeed so I guess we don't have much choice… I'm just hoping that Google won't flag our CDN as malicious once again (like it did multiple times in the past due to finding it suspicious that it hosts many `.apk` files, apparently—and which is the main reason why we're moving those to FAD to begin with) in the future again just because we still have `.apk` for Wear there… 🤞

fastlane/Fastfile
mokagio: The comment read to me like mostly an explanation of what the check does.

Updating to track _the rationale_ only

```suggestion
      # Fold for Wear app, as it's modified less often
      fold: app == WEAR_APP,
```

fastlane/Fastfile
mokagio: Love it!

I might take a second longer to parse the code ("why is that `compact` there?") but it's much more elegant and compact than `...app_icon = app_icon if app_icon`.

mokagio:
Notice I triggered new Prototype Builds for the latest commit, just in case.

fastlane/Fastfile
mokagio: Follow up https://github.com/woocommerce/woocommerce-android/pull/15322

Issue #4525, opened 2021-07-30
Notifications: Support new order & review notifications for all stores


anitaa1990:
List of tasks to be completed in order to support push notifications for all stores.

Ref: `P91TBi-5If`

- [x] Clean up FCMRegistrationService to make it unit testable (#4526).
- [x] Clean up NotificationHandler to make it unit testable.
- [x] Register a device for push notifications without including the store information. 

#### New Order notifications
- [x] Change the text of the notification to include the store name and the order amount.
- [x] Multiple new order notifications from different stores should be grouped together. Currently all notifications are grouped under a single bundle.
- [x] Clicking on a new order notification should open the order detail screen for that store and order.

#### New Review notifications
- [x] Change the text of the notification to include the store name, reviewer name and the review.
- [x] Multiple new review notifications from different stores should be grouped together.
- [x] Clicking on a new review notification should open the order review screen for that store.

PR #15205, opened 2026-01-16
Fix notification displaying issues

irfano:
### Description

**Problem:** The notification system was using an in-memory cache (`ACTIVE_NOTIFICATIONS_MAP`) to track active notifications, which caused synchronization issues with Android's system notification tray. When notifications were dismissed externally or when the app was restarted, the cache state could become stale, leading to incorrect notification grouping and display issues.
**⚠️ The main issue is that when the app is killed by the user or OS memory management, we lose track of tapped notifications.**

I removed the `and {count} more` text from group notifications. It’s redundant since the total count is already visible, and the OS often truncates the summary to two lines, making the "more" count inaccurate (e.g., showing "2 more" when 5 are actually hidden). This logic dates back five years ([https://github.com/woocommerce/woocommerce-android/issues/4525](https://github.com/woocommerce/woocommerce-android/issues/4525)) and is no longer necessary.

**Solution:** This PR replaces the in-memory cache with direct queries to Android's `NotificationManager.getActiveNotifications()`. Key changes include:

### Test Steps

**Example tapped event:**
```
push_notification_tapped, Properties: {"notification_note_id":9554402367,"notification_type":"PRODUCT_REVIEW","push_notification_token":"XXX","is_from_selected_site":true,"blog_id":111,"is_wpcom_store":true,"was_ecommerce_trial":false,"plan_product_slug":"ecommerce-bundle","store_id":"bfxxxxxx","is_debug":true,"site_url":"https:\/\/irfanwoo.wpcomstaging.com","cached_woo_core_version":"10.4.3"}
```

1. ⚠️ Receive a single push notification, kill the app, tap notif → Should track analytics.
3. Receive a single push notification → Should display as individual notification.
4. Receive 2+ notifications → Should display as grouped notification with correct count
5. Tap on individual notification → Should track analytics and navigate correctly.
6. Tap on grouped notification → Should track analytics for all child notifications.
7. Dismiss notifications from system tray → App should reflect correct state on next notification.

### Images/gif

||Before|After|
|-|-|-|
|Multiple reviews|before-multple|after-review|
|Different notif types|Screenshot_20260116_040413|after-order-review|
|Different stores (nothing changed)|Screenshot_20260116_040626|after-different-stores|

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
@@ -10,7 +10,6 @@ import android.content.Intent
 import android.graphics.Bitmap
 import android.os.Bundle
 import android.os.RemoteException
-import android.service.notification.StatusBarNotification
 import androidx.annotation.RequiresPermission
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
@@ -42,9 +41,19 @@ class WooNotificationBuilder @Inject constructor(
 
     fun cancelAllNotifications() = NotificationManagerCompat.from(context).cancelAll()
 
-    fun getActiveNotifications(): List<StatusBarNotification> {
+    fun getActiveNotifications(): List<ActiveNotificationData> {
         val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
-        return notificationManager.activeNotifications.toList()
+        return notificationManager.activeNotifications.map { sbn ->
+            ActiveNotificationData(
+                id = sbn.id,
+                remoteNoteId = sbn.notification.extras.getLong(EXTRA_REMOTE_NOTE_ID),
+                remoteSiteId = sbn.notification.extras.getLong(EXTRA_REMOTE_SITE_ID),
+                channelType = sbn.notification.extras.getString(EXTRA_CHANNEL_TYPE),
+                noteMessage = sbn.notification.extras.getString(EXTRA_NOTE_MESSAGE),
+                noteTypeTrackingValue = sbn.notification.extras.getString(EXTRA_NOTE_TYPE),
+                isGroupSummary = (sbn.notification.flags and FLAG_GROUP_SUMMARY) != 0
+            )
+        }
     }
 
     private fun getNotificationBuilder(
@@ -250,21 +259,6 @@ class WooNotificationBuilder @Inject constructor(
         }
     }
 
-    val StatusBarNotification.remoteNoteId: Long
-        get() = notification.extras.getLong(EXTRA_REMOTE_NOTE_ID)
-
-    val StatusBarNotification.remoteSiteId: Long
-        get() = notification.extras.getLong(EXTRA_REMOTE_SITE_ID)
-
-    val StatusBarNotification.channelType: String?
-        get() = notification.extras.getString(EXTRA_CHANNEL_TYPE)
-
-    val StatusBarNotification.noteMessage: String?
-        get() = notification.extras.getString(EXTRA_NOTE_MESSAGE)
-
-    val StatusBarNotification.noteTypeTrackingValue: String?
-        get() = notification.extras.getString(EXTRA_NOTE_TYPE)
-
     companion object {
         private const val EXTRA_REMOTE_NOTE_ID = "remote_note_id"
         private const val EXTRA_REMOTE_SITE_ID = "remote_site_id"
@@ -273,3 +267,13 @@ class WooNotificationBuilder @Inject constructor(
         private const val EXTRA_NOTE_TYPE = "note_type"
     }
 }
+
+data class ActiveNotificationData(
+    val id: Int,
+    val remoteNoteId: Long,
+    val remoteSiteId: Long,
+    val channelType: String?,
+    val noteMessage: String?,
+    val noteTypeTrackingValue: String?,
+    val isGroupSummary: Boolean = false
+)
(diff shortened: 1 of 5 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -48,8 +46,6 @@ class NotificationMessageHandler @Inject constructor(
 
         @VisibleForTesting
         const val MAX_INBOX_ITEMS = 5
-
-        private val ACTIVE_NOTIFICATIONS_MAP = mutableMapOf<Int, Notification>()
```
irfano, line 52: The main change is that we’re now using real UI notifications for tracking and grouping logic instead of an in-memory map.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -130,42 +126,33 @@ class NotificationMessageHandler @Inject constructor(
     }
 
     private fun handleWooNotification(notification: Notification) {
-        val randomNumber = if (notification.noteType == NewOrder) Random.nextInt() else 0
```
irfano: I removed the random integer logic for new orders since localPushId is already unique for each notification.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
```diff
@@ -47,7 +65,6 @@ class WooNotificationBuilder @Inject constructor(
             .setSmallIcon(R.drawable.ic_woo_w_notification)
             .setColor(ContextCompat.getColor(context, R.color.color_primary))
             .setOnlyAlertOnce(true)
-            .setAutoCancel(true)
```
irfano, line 50: I disabled `autoCancel` because it was automatically dismissing notifications, causing `notificationManager.activeNotifications` to return an empty list and preventing us from tracking taps. Since we already handle notification cancellation manually after processing the tap event, `autoCancel` isn't necessary.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationAnalyticsTracker.kt
```diff
@@ -15,10 +15,24 @@ class NotificationAnalyticsTracker @Inject constructor(
...
+        noteTypeTrackingValue: String
+    ) {
+        val isFromSelectedSite = selectedSite.getIfExists()?.siteId == remoteSiteId
         val properties = mutableMapOf<String, Any>()
-        properties["notification_note_id"] = notification.remoteNoteId
-        properties["notification_type"] = notification.noteType.trackingValue
+        properties["notification_note_id"] = remoteNoteId
+        properties["notification_type"] = noteTypeTrackingValue
         properties["push_notification_token"] = appPrefsWrapper.getFCMToken()
         properties["is_from_selected_site"] = isFromSelectedSite == true
```
JorgeMucientes: Minor np, is the == true needed here? I think `isFromSelectedSite` is already a non nullable boolean

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
JorgeMucientes: Good catch!

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -174,108 +162,89 @@ class NotificationMessageHandler @Inject constructor(
...
-                analyticsTracker.trackNotificationAnalytics(PUSH_NOTIFICATION_TAPPED, row.value)
+        notificationBuilder.getActiveNotifications()
+            .filter { it.channelType == type.name && !it.isGroupSummary }
+            .forEach {
+                analyticsTracker.trackNotificationAnalytics(
+                    stat = PUSH_NOTIFICATION_TAPPED,
+                    remoteNoteId = it.remoteNoteId,
+                    remoteSiteId = it.remoteSiteId,
+                    noteTypeTrackingValue = it.noteTypeTrackingValue.orEmpty()
+                )
```
JorgeMucientes, line 219: I'm aware this is how it worked before, but I'm not sure we want to track as "notification tapped" all notifications in a group, when user has only tapped in the group notification. No need to address that change now, and we should check how iOS tracks this before doing any chages. But just sharing some food for thought

JorgeMucientes (approved):
Nicely done @irfano , good catch fixing this. I left a few comments but nothing blocking. Let me know your thoughts.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationAnalyticsTracker.kt
irfano: Yes, it can be simplified. Updated: 67aa178b4ffbb34caaddc065286c536e75e33b5a

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: I thought about this during development. Since iOS doesn't support group notifications yet, it’s probably best to make a decision now. We could add a new group notification event as an alternative, but I’d prefer to stick with the current implementation for now.

JorgeMucientes: Would you mind discussing this with iOS folks to see if there's any plans to support group notifications? And if that were the case define a new event for taps on group notifications. 
For now let's merge this :shipit:

PR #15208, opened 2026-01-16
Attempt to fix a double "cha-ching" issue when creating group notific…

JorgeMucientes:
Coming from this external contributor [PR fix attempt](https://github.com/woocommerce/woocommerce-android/pull/15165)

### Description
Will copy paste my [comment](https://github.com/woocommerce/woocommerce-android/pull/15165#issuecomment-3759765934) in the external contributor PR: 

I was not able to reproduce the double "cha-ching" issue, so I suspect it is likely an issue with how different OEM/Android devices handle some push notification code. 

The fix attempt you provided though, I don't think will make any difference. Have you tested that it fixed the bug? Because changing the order on how some values are applied in `builder` pattern shouldn't really matter, because it is not until `showNotification()` is called that the values actually take effect. 

That being said this is what I suspect is happening is the following:

1. handleWooNotification() is called with isGroupNotification = false
2. buildAndDisplayWooNotification() is called:
  - Sets setDefaults(NotificationCompat.DEFAULT_ALL) → enables sound
  - showNotification() → First cha-ching (from channel)
  - Since isGroupNotification = false:
      - Sets setGroupSummary(true) and setGroupAlertBehavior(GROUP_ALERT_CHILDREN)
      - showNotification() again → Should NOT play sound due to GROUP_ALERT_CHILDREN

**Why GROUP_ALERT_CHILDREN Might Not Work:**
The issue is that GROUP_ALERT_CHILDREN is set on the same builder that was already used to show the first notification. The builder still has:
- Segfaults(NotificationCompat.DEFAULT_ALL)
- The notification channel with cha-ching sound
 
Possible causes:

1. Builder reuse issue: The builder is reused and .build() is called twice. The second call might still inherit properties that cause sound.
2. Channel-level sound override: On some Android versions/OEMs, the channel's sound setting might play regardless of GROUP_ALERT_CHILDREN.
3. setDefaults(DEFAULT_ALL) conflict: This might override or conflict with GROUP_ALERT_CHILDREN on some devices.

I think adding the following code: 
```kotlin
            if (!isGroupNotification) {
                setGroupSummary(true)
                setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
                setDefaults(0)
                setSound(null)
                setVibrate(null)
                showNotification(notification.getGroupPushId(), notification, this)
            }
```
Will fix the issue for all OEMs. I've opened a new PR with those changes. 

### Test Steps
I haven't been able to reproduce the issue neither in emulator nor in Pixel 10 device. However the way to  have the changes code called is: 
1. From a clean install of the app, log in and accept notification permissions
2. Purchase an item using the storefront
3. Wait for the notification to come in. A single cha-ching should sound. 
4. Tap on the received notification to clear any "active" notification so that `isGroupNotification` is set to false in subsequent pushes
5. Repeat steps 1-3 if need to re-test

### Images/gif

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -9,6 +9,7 @@
 - [Internal][Woo POS] Apply POS product visibility filtering to remote/non-local-catalog sync [https://github.com/woocommerce/woocommerce-android/pull/15191]
 - [Internal][Woo POS] Remove products hidden from POS during local catalog sync [https://github.com/woocommerce/woocommerce-android/pull/15190]
 - [Internal] Show JITMs on Dashboard regardless of onboarding status [https://github.com/woocommerce/woocommerce-android/pull/15170]
+- [Internal] Fix push notification double "cha-ching" issue [https://github.com/woocommerce/woocommerce-android/pull/15208]
 
 23.9
 -----
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
@@ -119,6 +119,9 @@ class WooNotificationBuilder @Inject constructor(
             if (!isGroupNotification) {
                 setGroupSummary(true)
                 setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
+                setDefaults(0)
+                setSound(null)
+                setVibrate(null)
                 showNotification(notification.getGroupPushId(), notification, this)
             }
         }


irfano (commented):
I couldn't reproduce this on my emulator or physical device. I noticed a slight echo during the "cha-ching" sound, but it persists even with grouped notifications disabled, so it’s likely just my speakers. Since we can’t reliably reproduce the issue, I suggest not merging this. We can reconsider if the contributor responds to your last comment. WDYT?

JorgeMucientes:
Hey @irfano thanks for the review. 
Given the known issues for different OEMs behaving different in regards to how PN are treated I'd vote to merge this. I think the changes are safe, moreover, the changes I'm adding in this PR already exist in our codebase. In the same `NotificationBuilder.kt` around line 140 you'll find this code: 
```kotlin
      NotificationCompat.Builder(context, channelId)
            .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_CHILDREN)
            .setSmallIcon(R.drawable.ic_woo_w_notification)
            .setColor(ContextCompat.getColor(context, R.color.color_primary))
            .setGroup(notification.getGroup())
            .setGroupSummary(true)
            .setAutoCancel(true)
            .setTicker(notification.noteMessage)
            .setContentTitle(notification.noteTitle)
            .setContentText(subject)
            .setStyle(inboxStyle)
            .setSound(null)
            .setVibrate(null)
```
which does the same thing. I think there's a high chance the changes I introduced will fix the problem while being safe enough of not breaking anything.

irfano (approved):
Good point. Since it's a safe change that doesn't break anything and just silences group notifications, let's go ahead and merge it.

AlexKlimaj:
I've had a double cha-ching issue for the past few years on Pixel devices. Just installed this build and I'll let you know if it fixes it.

AlexKlimaj:
With this binary installed, I still get a double cha-ching for every order. Pixel 9 Pro. Any logs I can provide that could help debug?

irfano:
Hey @AlexKlimaj! Thanks for reporting and following up on this.

Regarding the double sound: are both playing simultaneously, or does the second one start right after the first?

Also, have you tried reproducing this on an emulator, or is it only happening on your Pixel?

AlexKlimaj:
> Hey @AlexKlimaj! Thanks for reporting and following up on this.
> 
> Regarding the double sound: are both playing simultaneously, or does the second one start right after the first?
> 
> Also, have you tried reproducing this on an emulator, or is it only happening on your Pixel?

This is happening on my pixel hardware. 

The second sound comes sometimes seconds after to many minutes.

irfano:
A few more questions to help me understand your environment, @AlexKlimaj: When the sound triggers twice, do you see one notification in the status bar or two? Also, are you sure there's only one instance of the app installed? If you have both the Play Store version and this PR build, you might have duplicate apps on your phone.

AlexKlimaj:
I see two notifications in the status bar.

This has been happening for years with the app store release. I uninstalled that when I loaded this version.

AlexKlimaj:
Here are two notifications events in the app log that appear in close succession. 

```
[Jan-22 03:19:50:287 UTILS i] 🔵 Tracked: woocommerceandroid_new_order_push_notification_sound, Properties: {"sound_status":"DEFAULT","blog_id":xxxxxx,"is_wpcom_store":false,"was_ecommerce_trial":false,"plan_product_slug":"jetpack_free","store_id":"xxxxxxxxxxx","is_debug":true,"site_url":"https:\/\/xxxxxxxx.com","cached_woo_core_version":"10.4.3"}
```

```
[Jan-22 03:19:50:586 UTILS i] 🔵 Tracked: woocommerceandroid_push_notification_received, Properties: {"notification_note_id":xxxxx,"notification_type":"NEW_ORDER","push_notification_token":"xxxxxxxxxxxx,"is_from_selected_site":true,"blog_id":xxxxxxxxx,"is_wpcom_store":false,"was_ecommerce_trial":false,"plan_product_slug":"jetpack_free","store_id":"xxxxxxxxxxx","is_debug":true,"site_url":"https:\/\/xxxxxxx.com","cached_woo_core_version":"10.4.3"}
```

JorgeMucientes:
Hey @AlexKlimaj thanks for getting back to us 🙏🏼. 
We've looked into this and came to the conclusion that it doesn't seem to be an issue with the app but in our backend. It seems we are triggering the same push several times. This may be due to your WP.com account having multiple push notification tokens registered in our system. 
At this point I'd suggest if you could please reach out to use via woo app > settings > Help & Support > Contact support. Share you issue and even link to this comment for extra context. From there you can check how many push tokens are linked tto your WordPress.com account and try to troubleshoot those duplicated pushes.

PR #15210, opened 2026-01-16
[WOOMOB-1960] Use order currency on POS order list and detail

malinajirka:
Fixes WOOMOB-1960

### Description
The POS order list and detail screens were displaying prices using the store's default currency instead of the order's actual currency. This caused incorrect currency symbols to appear when orders were placed in a different currency than the store's base currency.

Added a new overload to `WooPosFormatPrice` that accepts a currency code parameter, and updated all price formatting calls in `WooPosOrdersViewModel` to use the order's currency.

### Test Steps
1. Configure a store with a base currency (e.g., USD)
2. Create orders in different currencies (e.g., EUR, GBP)
3. Open the POS Orders screen
4. Verify the order list shows prices with the correct currency symbol for each order
5. Tap an order and verify the detail view shows all prices (total, line items, refunds, breakdown) with the correct currency

### Images/gif

| Before | After |
|--------|-------|
| image | Screenshot_1768573921 |


samiuelson (approved):
Good catch, LGTM!

PR #15212, opened 2026-01-16
[Woo POS][Refunds] Keep historical refunds fresh 

samiuelson:
WOOMOB-2004

### Description

This PR fixes an issue in WooCommerce POS where historical refunds were not being refreshed when opening order details, or pulling to refresh the orders list, causing order actions (like "Issue Refund" button) to remain in a stale state indefinitely.

### Test Steps

1. Open POS orders list with orders containing refunds
2. Select an order - verify actions appear with shimmer loading state initially
3. Pull to refresh - verify refunds are updated and actions reflect current state
4. Rapidly switch between orders - verify no stale data appears (race condition fix)
5. Open refund modal - verify historical refunds are always fresh

### Images/gif

https://github.com/user-attachments/assets/4554b88e-792b-4913-a651-0f58cc27878a


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -35,6 +35,7 @@ import java.util.Locale
 import javax.inject.Inject
 import kotlin.time.TimeSource.Monotonic
 
+@Suppress("LargeClass")
```
samiuelson, line 38: I addressed this and other detekt issues inside this class in a follow-up refactoring PR: https://github.com/woocommerce/woocommerce-android/pull/15218

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDetails.kt
```diff
@@ -78,10 +79,44 @@ fun WooPosOrderDetails(
...
+    details: WooPosOrdersState.OrderDetailsViewState.Computed.Details,
+    onUIEvent: (WooPosOrdersUIEvent) -> Unit
+) {
+    when (val actionsState = details.actionsState) {
+        is WooPosOrdersState.OrderActionsState.Loading -> {
+            WooPosShimmerBox(
+                modifier = Modifier
+                    .height(40.dp)
+                    .width(160.dp)
+                    .clip(RoundedCornerShape(WooPosCornerRadius.Small.value))
```
kidinov: np: I'd suggest using a larger radius here as we kinda loading button which has large corner radious

kidinov (commented):
LGTM, but 2 questions

* There is something weird between the POS orders - some of them cannot be refunded in the main app, and those that are refunded do not change the status to refunded. Check the video

https://github.com/user-attachments/assets/c73ec3ee-61a6-4023-be58-5207821d5ade

* Do we need to display "loading" if we already have the status of the order in "refunded"? Can we show it this way, and in background check if it's correct (if this even needed to be checked)?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -113,10 +115,17 @@ class WooPosOrdersViewModel @Inject constructor(
         return getRefundableItems(order, refunds).isNotEmpty()
     }
 
+    @Suppress("LongMethod", "CyclomaticComplexMethod")
     fun onOrderSelected(orderId: Long) {
         val current = _state.value as? WooPosOrdersState.Content ?: return
         val loadedItems = current.items as? WooPosOrdersState.Content.Items.Loaded ?: return
 
+        if (current.selectedDetails?.id == orderId &&
+            current.selectedDetails.actionsState is WooPosOrdersState.OrderActionsState.Loaded
```
kidinov: Cannot understand this one. So if we selected the currently selected order but it's `Loading` then we continue and start loading again?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosRetrieveOrderRefunds.kt
```diff
@@ -14,23 +14,24 @@ class WooPosRetrieveOrderRefunds @Inject constructor(
...
         withContext(Dispatchers.IO) {
             if (order.refundTotal.compareTo(BigDecimal.ZERO) == 0) {
                 return@withContext Result.success(emptyList())
             }
 
             val site = selectedSite.get()
 
-            var refundModels = refundStore.getAllRefunds(site, order.id)
-            if (refundModels.isEmpty()) {
+            val refundModels = if (forceRefresh) {
```
kidinov, line 25: If I got this right - the old behavior was "try cache first, if empty fetch from network." The new behavior with forceRefresh = false always returns cache (even if empty). This means orders with refunds not yet cached will show no refunds in the list until selected. Expected?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDetails.kt
samiuelson: Updated to medium (the same as Button): [29f4231](https://github.com/woocommerce/woocommerce-android/pull/15212/commits/29f4231929510ce76e74e9f99b614ec3cd90a8eb)

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
samiuelson: Nope—if we select the currently selected order, **we do not load** again in case refunds (actions) have been **loaded** already (OrderActionsState.Loaded).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosRetrieveOrderRefunds.kt
samiuelson: This is a mistake. I added the fallback to fetch refunds in case the cache is empty: de8b00770aa776384c0c488a1c682c84cbd83247, in case `forceRefresh` is `false`.

samiuelson:
Thanks for review, @kidinov! I addressed your comments. 

> There is something weird between the POS orders - some of them cannot be refunded in the main app, and those that are refunded do not change the status to refunded. Check the video

This is a bug in store management—tax total is calculated incorrectly. Notice that order total is $2982.00 and we attempt to request a refund for all of the items with refund amount of $2981.56. Backend prevents from creating refund with incorrect amount. In the POS refund flow, the calculations should be correct.

> Do we need to display "loading" if we already have the status of the order in "refunded"? Can we show it this way, and in background check if it's correct (if this even needed to be checked)?

I think it's better to hide "Issue refund" button by default and offer it only if we're sure the refund can be done. Otherwise, it can be frustrating.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
kidinov: However, the tone of the other things happening below in this method, including tracking and more, and we just return from here based on the loaded refunds? It feels off. Even if nothing is broken right now (which I am not sure), it definitely may be broken later because this behavior is not really clear why?

kidinov:
> This is a bug in store management—tax total is calculated incorrectly. Notice that order total is $2982.00 and we attempt to request a refund for all of the items with refund amount of $2981.56. Backend prevents from creating refund with incorrect amount. In the POS refund flow, the calculations should be correct.

Is there a ticket on that? Sounds like quite a serious bug, right? in this case probably would make sense to extract calculation in one place and reuse it to make sure that if something is found, we fix it everywhere

> I think it's better to hide "Issue refund" button by default and offer it only if we're sure the refund can be done. Otherwise, it can be frustrating.

Yes, but that's not what I telling. There is order status "refunded" and we still display "loading" of the issue refund button, while we know it's been "refunded" already. Or am I missing something here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
samiuelson: I refactored code in 9e3cf78d7cc93e2856ed75690afb31ba766f0346. Let me know what you think.

samiuelson:
> > This is a bug in store management—tax total is calculated incorrectly. Notice that order total is $2982.00 and we attempt to request a refund for all of the items with refund amount of $2981.56. Backend prevents from creating refund with incorrect amount. In the POS refund flow, the calculations should be correct.
> 
> Is there a ticket on that? Sounds like quite a serious bug, right? in this case probably would make sense to extract calculation in one place and reuse it to make sure that if something is found, we fix it everywhere
> 

I created the ticket: WOOMOB-2011. We will examine data and decide whether to wait for backend improvements (getting rid of local calculation) or fixing the store management.

> > I think it's better to hide "Issue refund" button by default and offer it only if we're sure the refund can be done. Otherwise, it can be frustrating.
> 
> Yes, but that's not what I telling. There is order status "refunded" and we still display "loading" of the issue refund button, while we know it's been "refunded" already. Or am I missing something here?

Hm. Sounds like good optimization idea. Implemented: 1456480c8ec68f83f7c332bf0e63387dbd303ff1

Thsnks for review, @kidinov !

kidinov (approved):
LGTM!

> FYI, even when it's cash payment it shows a card refund

Yeah, it's a hardcoded string at the moment.

PR #15214, opened 2026-01-19
[WOOMOB-2008] Add local feature flag for client-side POS banner

kidinov:
WOOMOB-2008

### Description

Add a local feature flag `WOO_POS_CLIENT_SIDE_BANNER` that controls the visibility of the client-side POS promo banner. The banner now requires both the local flag and the remote flag (`WOO_POS_TABLET_PROMO_BANNER`) to be enabled before showing.

This needed to make sure that when the remote FF enabled we won't show text that not A/B tested on the older versions of the app

### Test Steps

1. Verify the banner shows on a phone device (not tablet) in an eligible country (US or GB)
2. Set `WOO_POS_CLIENT_SIDE_BANNER` to `false` in `FeatureFlag.kt`
3. Verify the banner no longer shows

### Images/gif

N/A - No UI changes


malinajirka (approved):
LGTM, thanks!

PR #15215, opened 2026-01-19
Introduce in-memory `MediaModel` cache

wzieba:
Closes: [AINFRA-542: Migrate `MediaModel`](https://linear.app/a8c/issue/AINFRA-542/migrate-mediamodel)

### Description

This PR migrates `MediaModel` from being an ORM model to a stored in-memory POJO. This reduces configuration overhead, while not degrading user experience. See: p91TBi-dHl-p2 

It also removes a decent portion of unused properties and classes, that happened to not be used anymore.

The goal of this PR was not to improve every aspect of handling `MediaModel` as it'd simply require too many changes for a single PR. The list of remaining improvements, while probably not exhousted, can be found in [AINFRA-1764: Address remaining `Media*` improvements in Woo Android](https://linear.app/a8c/issue/AINFRA-1764/address-remaining-media-improvements-in-woo-android)

### Test Steps

We need to test media upload for two connection types: WPCOM and Application Password.

- **WPCOM** is when users log via WPCOM account
- **Application Passwords** can be easily tested by creating a Jurassic page with Woo but without Jetpack and log in via site credentials

#### Scenarios to test
- Product image
  - Upload
  - Remove
  - Reorder
  - Add image from WordPress media library
  - Failed upload (e.g. try to upload a `.tif` file https://svs.gsfc.nasa.gov/vis/a030000/a030800/a030877/frames/5760x3240_16x9_01p/)
- Add "downloadable file" in Virtual product
  - Upload image
  - Upload document
  - Select a file from WordPress Media Library

### Images/gif

There are no UI changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    fun getSiteAudio(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "audio")
+    }
+
+    fun getSiteDocuments(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "application")
+    }
+
+    fun searchSiteImages(siteId: Int, searchTerm: String): List<MediaModel> {
```
wzieba, line 26: We can't test `search*` methods because the app doesn't enable searching 

https://github.com/woocommerce/woocommerce-android/blob/9aac6dd71c854d8a8f3fd6d9d71d2981aaef3111/WooCommerce/src/main/kotlin/com/woocommerce/android/mediapicker/medialibrary/MediaLibraryPickerSetup.kt#L16

As it's like this for years, we should probably update contract with `MediaPicker` library to not being forced to provide search methods if searching is hidden.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    fun cacheMediaList(localSiteId: Int, mediaList: List<MediaModel>) {
+        cache[localSiteId] = mediaList
+    }
+
+    fun addOrUpdate(localSiteId: Int, media: MediaModel) {
+        val currentList = cache[localSiteId] ?: emptyList()
+        val mutableList = currentList.toMutableList()
+        val existingIndex = mutableList.indexOfFirst { it.mediaId == media.mediaId }
```
wzieba: This class operates on `mediaId` (remote id) because we only cache `MediaModel` that we fetched from remote.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -488,67 +483,56 @@ public void onRegister() {
...
-        }
-
-        return insertedMedia;
-    }
-
-    @Nullable
-    public MediaModel getSiteMediaWithId(@NonNull SiteModel siteModel, long mediaId) {
-        List<MediaModel> media = MediaSqlUtils.getSiteMediaWithId(siteModel, mediaId);
-        return media.size() > 0 ? media.get(0) : null;
+        media.setId(mMediaIdGenerator.generate(media.getFilePath()).getValue());
```
wzieba, line 486: `instantiateMediaModel` is only used for upload, hence a generated, local-only id. We should probably remove `instantiateMediaModel` method - there's no point to inject a heavy `MediaStore` only to call this constructor. I've added it to the list of improvements.

libs/fluxc/src/testFixtures/kotlin/org/wordpress/android/fluxc/media/MediaTestUtils.kt
```diff
@@ -0,0 +1,135 @@
...
+                this.mediaId,
+                this.postId,
+                this.uploadDate,
+                this.url,
+                this.fileName,
+                this.mimeType,
+                this.title,
+                this.caption,
+                this.description,
+            )
```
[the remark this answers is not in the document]
wzieba: True and this is I think the only case when I'm okay with leaving unused code 😛 . It's a builder to prepare a fake used for unit tests - utility as its finest. But if it's prefered to remove elements here, I'm also okay with this (we can always add them when/if needed).

libs/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -3,22 +3,13 @@
...
 import org.wordpress.android.util.StringUtils;
 
 import java.io.Serializable;
 
-// WARN: This class is used within WordPress-MediaPicker-Android, do not remove!
-@Table
-public class MediaModel extends Payload<BaseNetworkError> implements Identifiable, Serializable {
-    private static final long serialVersionUID = -1396457338496002846L;
-
+public class MediaModel extends Payload<BaseNetworkError> implements Serializable {
```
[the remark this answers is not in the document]
wzieba: `serialVersionUID` will be autogenerated 👍 . We don't persist `MediaModel` anywhere, so we don't have to care about backward-compatibility (e.g. we don't restore it from a local file). It only works in-memory, so it's safe to remove this property.

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
```diff
@@ -0,0 +1,155 @@
...
+            id.toLong(), // mediaId
+            0L, // postId
+            null, // uploadDate
+            "https://example.com/$fileName", // url
+            fileName, // fileName
+            "image/jpeg", // mimeType
+            fileName, // title
+            "", // caption
+            "" // description
+            // uploadState
```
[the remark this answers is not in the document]
wzieba: 👍 2bac569c535bc62469c0e0ca10f94d358c02ab6

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import javax.inject.Inject
+import kotlin.time.Clock
+import kotlin.time.ExperimentalTime
+
+@OptIn(ExperimentalTime::class)
+class TimestampMediaIdGenerator @Inject constructor(private val clock: Clock) : MediaIdGenerator {
+    override fun generate(filePath: String): LocalId {
+        val combined = "$filePath:${clock.now().toEpochMilliseconds()}"
+        return LocalId(combined.hashCode())
```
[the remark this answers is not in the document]
wzieba: 1. System doesn't care if id is possitive or negative
2. `hashCode` clashing could be theoretically a problem but I also think is acceptable? I think this capsulates my thinking:

>When is this acceptable?
Local, in‑memory identifiers, short‑lived, where a rare collision would at worst cause a minor glitch and can be detected/handled.

but if it's somehow important for us, I can apply changes (e.g. make `id` a String and simply use value of `combined`, instead of `combined.hashCode()`)

libs/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -242,29 +106,22 @@ public boolean equals(@Nullable Object other) {
...
                 && StringUtils.equals(getFileName(), otherMedia.getFileName())
                 && StringUtils.equals(getFilePath(), otherMedia.getFilePath())
                 && StringUtils.equals(getMimeType(), otherMedia.getMimeType())
                 && StringUtils.equals(getTitle(), otherMedia.getTitle())
                 && StringUtils.equals(getDescription(), otherMedia.getDescription())
-                && StringUtils.equals(getCaption(), otherMedia.getCaption())
-                && StringUtils.equals(getAlt(), otherMedia.getAlt())
-                && StringUtils.equals(getUploadState(), otherMedia.getUploadState());
+                && StringUtils.equals(getCaption(), otherMedia.getCaption());
     }
```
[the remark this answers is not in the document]
wzieba: Good point, especially that we use this model in `ConcurentHashMap`. bc5c92311837df2e67c61292ba0945ac603a1b4

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    @Test
+    fun `when removing media by local id, then it filters out media`() {
+        val mediaToRemove = testMedia1
+        cache.cacheMediaList(1, listOf(mediaToRemove, testMedia2))
+
+        cache.remove(1, mediaToRemove.mediaId)
+
+        val result = cache.getMediaList(1)
+        assertThat(result).containsExactly(testMedia2)
+    }
```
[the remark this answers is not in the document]
wzieba: Good catch! I initially used a local id here :). 25bf86ff6541190dffc38d06c68cceed879b7ff

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -654,10 +626,7 @@ private void performFetchMediaList(@NonNull FetchMediaListPayload payload) {
     private void performCancelUpload(@NonNull CancelMediaPayload payload) {
         MediaModel media = payload.media;
         if (payload.delete) {
-            MediaSqlUtils.deleteMedia(media);
-        } else {
-            media.setUploadState(MediaUploadState.FAILED);
-            MediaSqlUtils.insertOrUpdateMedia(media);
+            mMediaLibraryCache.remove(payload.site.getId(), media.getMediaId());
```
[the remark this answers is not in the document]
wzieba: Correct me if I'm wrong, but cache is only filled with `MediaModel` that has **already been fetched** (`MediaStore#updateFetchedMediaList`) - so 

> When canceling an upload with delete=true for media that hasn't been uploaded yet

should never happen?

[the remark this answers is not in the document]
wzieba: Claude Code confirms: 

> No, that comment doesn't make sense based on the actual code flow. Here's why:
  The MediaLibraryCache only stores remote media that have been fetched from the server, not local media pending upload.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,40 @@
...
+        if (existingIndex != -1) {
+            mutableList[existingIndex] = media
+        } else {
+            mutableList.add(media)
+        }
+        cache[localSiteId] = mutableList
+    }
```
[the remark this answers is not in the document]
wzieba: Actually, the same comment as above: https://github.com/woocommerce/woocommerce-android/pull/15215#discussion_r2713586896

`MediaLibraryCache` is only for remotely fetched media - they don't even have local ids. Confusion like this would be addressed by introducing two different models, like `Local*` and `Remote*` variants of `MediaModel`. It's in the list of future improvements.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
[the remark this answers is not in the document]
wzieba: I'm aware of this. This should never happen, because when initialising local `MediaModel` we should always provide a path. I wouldn't be overly protective here.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    fun remove(localSiteId: Int, mediaId: Long) {
+        val currentList = cache[localSiteId] ?: return
+        cache[localSiteId] = currentList.filter { it.mediaId != mediaId }
+    }
+
+    fun clear() {
+        cache.clear()
+    }
+}
```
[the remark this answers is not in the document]
wzieba: Interesting point - but it's hard for me to believe that anyone has a media library **that** big (important note: we have pagination, so they not only have to have extremly huge media library, but also scroll for a few.. hours 🤷 ). 

Important note: `MediaLibraryCache` is for caching `MediaModel` POJO - not the actuall image/content.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -488,67 +483,56 @@ public void onRegister() {
...
-        return insertedMedia;
-    }
-
-    @Nullable
-    public MediaModel getSiteMediaWithId(@NonNull SiteModel siteModel, long mediaId) {
-        List<MediaModel> media = MediaSqlUtils.getSiteMediaWithId(siteModel, mediaId);
-        return media.size() > 0 ? media.get(0) : null;
+        media.setId(mMediaIdGenerator.generate(media.getFilePath()).getValue());
+        return media;
     }
```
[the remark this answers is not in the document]
wzieba: That's a good point 👍 fb896cbe4d84f7eb190ea165dbf04baf97d5154f

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,40 @@
...
+            mutableList[existingIndex] = media
+        } else {
+            mutableList.add(media)
+        }
+        cache[localSiteId] = mutableList
+    }
+
+    fun remove(localSiteId: Int, mediaId: Long) {
+        val currentList = cache[localSiteId] ?: return
+        cache[localSiteId] = currentList.filter { it.mediaId != mediaId }
```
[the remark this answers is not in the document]
wzieba: Yeah, good point. Addressed in 
- 307f316acc08efb6b3bf5d5d49eb3ec22576918b
- 8ba74459073db6a48f03f48166a894bec8f71f80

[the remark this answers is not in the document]
wzieba: https://github.com/woocommerce/woocommerce-android/pull/15215#discussion_r2713821674

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
```diff
@@ -0,0 +1,78 @@
+package org.wordpress.android.fluxc.store
+
+import org.wordpress.android.fluxc.model.MediaModel
+import java.util.Locale
+import javax.inject.Inject
+
+class MediaCacheOperations @Inject constructor(
```
ParaskP7: Question (❓): Should we (maybe) make all these classes `internal` to this `fluxc` module? 🤔

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    fun searchSiteAudio(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "audio", searchTerm)
+    }
+
+    fun searchSiteDocuments(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "application", searchTerm)
+    }
+
+    fun getCacheSize(siteId: Int): Int {
```
ParaskP7: Minor (🔍): This function could be made `private`.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    fun remove(localSiteId: Int, mediaId: Long) {
+        cache.computeIfPresent(localSiteId) { _, currentList ->
+            currentList.filter { it.mediaId != mediaId }
+        }
+    }
+
+    fun clear() {
```
ParaskP7: Minor (🔍): Maybe we should add the `@TestOnly` annotation to this function. 🤔

libs/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -242,29 +92,27 @@ public boolean equals(@Nullable Object other) {
...
 
     @Override
+    public int hashCode() {
+        return Objects.hash(mId, mLocalSiteId, mMediaId, mPostId, mUploadDate, mUrl, mFileName, mFilePath, mMimeType, mTitle, mCaption, mDescription);
```
ParaskP7: Minor (🔍): Although this rule is active, this is not reported by Detekt as `MaximumLineLength` warning, because it excluded the `**/fluxc/** module, but it nevertheless its length is bigger than the norm.

`Suggestion`
```java
    @Override
    public int hashCode() {
        return Objects.hash(
                mId,
                mLocalSiteId,
                mMediaId,
                mPostId,
                mUploadDate,
                mUrl,
                mFileName,
                mFilePath,
                mMimeType,
                mTitle,
                mCaption,
                mDescription
        );
    }
```

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        startLatch.countDown()
+        latch.await()
+
+        assertThat(cache.getMediaList(siteId))
+            .hasSize(threadCount)
+            .containsExactlyInAnyOrderElementsOf(mediaItems)
+    }
+
+    private fun createTestMedia(id: Int, fileName: String) =
+        MediaTestUtils.createRemoteTestMedia().mediaId(id.toLong()).fileName(fileName).build()
```
ParaskP7: Suggestion (💡): I find it easier to read such builder code, and know it's a builder in the first place, if split into multiple lines:

`Suggestion`
```kotlin
    private fun createTestMedia(id: Int, fileName: String) =
        MediaTestUtils.createRemoteTestMedia()
            .mediaId(id.toLong())
            .fileName(fileName)
            .build()
```

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
```diff
@@ -0,0 +1,42 @@
+package org.wordpress.android.fluxc.store
+
+import org.wordpress.android.fluxc.model.MediaModel
+import java.util.concurrent.ConcurrentHashMap
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class MediaLibraryCache @Inject constructor() {
```
ParaskP7: 1. Suggestion (💡): Why the extra `Library` word in it and not having it simply be called `MediaCache`, similar to how `MediaCacheOperations` is named?
2. Suggestion (💡): What not combining all these 3 `MediaLibraryCache`, `MediaCacheOperations` and `MediaIdGenerator` classes into 1 `MediaCache` class. I asked AI to do that for me and here's how it looks (I find `MediaStore.java` looking simpler for it (see `AI Patch` below):

<details>
    <summary>AI Patch</summary>

`AI Prompt`
```
I am currently reviewing this 'MediaStore.java' file, a change based on this 
https://github.com/woocommerce/woocommerce-android/pull/15215 PR.
I think 'MediaLibraryCache', 'MediaCacheOperations' and 'MediaIdGenerator'
can be all combined into 1 class, and this way we avoid using 3 different
classes and fields, all of which are for media cache purposes. Can you help me?  
```

`AI Patch`
```diff
Subject: [PATCH] Introduce in-memory MediaModel cache #15215
---
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
deleted file mode 100644
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ /dev/null	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
@@ -1,78 +0,0 @@
-package org.wordpress.android.fluxc.store
-
-import org.wordpress.android.fluxc.model.MediaModel
-import java.util.Locale
-import javax.inject.Inject
-
-class MediaCacheOperations @Inject constructor(
-    private val cache: MediaLibraryCache
-) {
-    fun getSiteImages(siteId: Int): List<MediaModel> {
-        return filterByMimeType(siteId, "image")
-    }
-
-    fun getSiteVideos(siteId: Int): List<MediaModel> {
-        return filterByMimeType(siteId, "video")
-    }
-
-    fun getSiteAudio(siteId: Int): List<MediaModel> {
-        return filterByMimeType(siteId, "audio")
-    }
-
-    fun getSiteDocuments(siteId: Int): List<MediaModel> {
-        return filterByMimeType(siteId, "application")
-    }
-
-    fun searchSiteImages(siteId: Int, searchTerm: String): List<MediaModel> {
-        return searchByMimeTypeAndTerm(siteId, "image", searchTerm)
-    }
-
-    fun searchSiteVideos(siteId: Int, searchTerm: String): List<MediaModel> {
-        return searchByMimeTypeAndTerm(siteId, "video", searchTerm)
-    }
-
-    fun searchSiteAudio(siteId: Int, searchTerm: String): List<MediaModel> {
-        return searchByMimeTypeAndTerm(siteId, "audio", searchTerm)
-    }
-
-    fun searchSiteDocuments(siteId: Int, searchTerm: String): List<MediaModel> {
-        return searchByMimeTypeAndTerm(siteId, "application", searchTerm)
-    }
-
-    fun getCacheSize(siteId: Int): Int {
-        return cache.getMediaList(siteId)?.size ?: 0
-    }
-
-    fun getUploadedMediaCount(siteId: Int, mimeTypePrefix: String?): Int {
-        if (mimeTypePrefix == null) {
-            return getCacheSize(siteId)
-        }
-        return filterByMimeType(siteId, mimeTypePrefix).size
-    }
-
-    private fun filterByMimeType(siteId: Int, mimeTypePrefix: String): List<MediaModel> {
-        val allMedia = cache.getMediaList(siteId) ?: return emptyList()
-        return allMedia.filter { media ->
-            media.mimeType?.startsWith(mimeTypePrefix) == true
-        }
-    }
-
-    private fun searchByMimeTypeAndTerm(
-        siteId: Int,
-        mimeTypePrefix: String,
-        searchTerm: String
-    ): List<MediaModel> {
-        val allMedia = cache.getMediaList(siteId) ?: return emptyList()
-        val lowerSearchTerm = searchTerm.lowercase(Locale.ROOT)
-        return allMedia.filter { media ->
-            media.mimeType?.startsWith(mimeTypePrefix) == true &&
-                    matchesSearchTerm(media, lowerSearchTerm)
-        }
-    }
-
-    private fun matchesSearchTerm(media: MediaModel, lowerSearchTerm: String): Boolean {
-        return (media.title?.contains(lowerSearchTerm, ignoreCase = true) == true) ||
-            media.caption.contains(lowerSearchTerm, ignoreCase = true) ||
-            media.description.contains(lowerSearchTerm, ignoreCase = true)
-    }
-}
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaIdGenerator.kt
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaIdGenerator.kt b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaIdGenerator.kt
deleted file mode 100644
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaIdGenerator.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ /dev/null	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
@@ -1,7 +0,0 @@
-package org.wordpress.android.fluxc.store
-
-import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
-
-interface MediaIdGenerator {
-    fun generate(filePath: String): LocalId
-}
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt
deleted file mode 100644
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ /dev/null	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
@@ -1,14 +0,0 @@
-package org.wordpress.android.fluxc.store
-
-import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
-import javax.inject.Inject
-import kotlin.time.Clock
-import kotlin.time.ExperimentalTime
-
-@OptIn(ExperimentalTime::class)
-class TimestampMediaIdGenerator @Inject constructor(private val clock: Clock) : MediaIdGenerator {
-    override fun generate(filePath: String): LocalId {
-        val combined = "$filePath:${clock.now().toEpochMilliseconds()}"
-        return LocalId(combined.hashCode())
-    }
-}
Index: libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/TimestampMediaIdGeneratorTest.kt
===================================================================
diff --git a/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/TimestampMediaIdGeneratorTest.kt b/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/TimestampMediaIdGeneratorTest.kt
deleted file mode 100644
--- a/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/TimestampMediaIdGeneratorTest.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ /dev/null	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
@@ -1,52 +0,0 @@
-package org.wordpress.android.fluxc.store
-
-import org.assertj.core.api.Assertions
-import org.junit.Before
-import org.junit.Test
-import kotlin.time.Clock
-import kotlin.time.ExperimentalTime
-import kotlin.time.Instant
-
-@OptIn(ExperimentalTime::class)
-class TimestampMediaIdGeneratorTest {
-
-    lateinit var sut: TimestampMediaIdGenerator
-
-    var currentMillis = 123L
-    private val fakeClock = object : Clock {
-        override fun now(): Instant {
-            return Instant.fromEpochMilliseconds(currentMillis)
-        }
-    }
-
-    @Before
-    fun setUp() {
-        sut = TimestampMediaIdGenerator(fakeClock)
-    }
-
-    @Test
-    fun `when generating with same inputs, then it returns consistent ID`() {
-        val id1 = sut.generate("/path/to/file.jpg")
-        val id2 = sut.generate("/path/to/file.jpg")
-
-        Assertions.assertThat(id1).isEqualTo(id2)
-    }
-
-    @Test
-    fun `when generating with different file paths, then it returns different IDs`() {
-        val id1 = sut.generate("/path/to/file1.jpg")
-        val id2 = sut.generate("/path/to/file2.jpg")
-
-        Assertions.assertThat(id1).isNotEqualTo(id2)
-    }
-
-    @Test
-    fun `when generating with different timestamps, then it returns different IDs`() {
-        currentMillis = 1000L
-        val id1 = sut.generate("/path/to/file.jpg")
-        currentMillis = 2000L
-        val id2 = sut.generate("/path/to/file.jpg")
-
-        Assertions.assertThat(id1).isNotEqualTo(id2)
-    }
-}
Index: libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt b/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
--- a/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ b/libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt	(date 1769085781074)
@@ -6,15 +6,24 @@
 import org.wordpress.android.fluxc.media.MediaTestUtils
 import org.wordpress.android.fluxc.model.MediaModel
 import java.util.concurrent.CountDownLatch
+import kotlin.time.Clock
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
+@OptIn(ExperimentalTime::class)
 class MediaLibraryCacheTest {
     private lateinit var cache: MediaLibraryCache
     private lateinit var testMedia1: MediaModel
     private lateinit var testMedia2: MediaModel
+    private var currentTime = 1000L
+
+    private val testClock = object : Clock {
+        override fun now(): Instant = Instant.fromEpochMilliseconds(currentTime)
+    }
 
     @Before
     fun setup() {
-        cache = MediaLibraryCache()
+        cache = MediaLibraryCache(testClock)
         testMedia1 = createTestMedia(1, "image1.jpg")
         testMedia2 = createTestMedia(2, "image2.jpg")
     }
@@ -148,7 +157,6 @@
 
         val mediaItems = (1..threadCount).map { createTestMedia(it, "image$it.jpg") }
 
-        // Launch threads that all try to add media at the same time
         mediaItems.map { media ->
             Thread {
                 startLatch.await()
@@ -165,6 +173,139 @@
             .containsExactlyInAnyOrderElementsOf(mediaItems)
     }
 
+    @Test
+    fun `when getting site images, then it returns only images`() {
+        val image = createTestMediaWithMimeType(1, "image/jpeg")
+        val video = createTestMediaWithMimeType(2, "video/mp4")
+        cache.cacheMediaList(1, listOf(image, video))
+
+        val result = cache.getSiteImages(1)
+
+        assertThat(result).containsExactly(image)
+    }
+
+    @Test
+    fun `when getting site videos, then it returns only videos`() {
+        val image = createTestMediaWithMimeType(1, "image/jpeg")
+        val video = createTestMediaWithMimeType(2, "video/mp4")
+        cache.cacheMediaList(1, listOf(image, video))
+
+        val result = cache.getSiteVideos(1)
+
+        assertThat(result).containsExactly(video)
+    }
+
+    @Test
+    fun `when getting site audio, then it returns only audio`() {
+        val image = createTestMediaWithMimeType(1, "image/jpeg")
+        val audio = createTestMediaWithMimeType(2, "audio/mp3")
+        cache.cacheMediaList(1, listOf(image, audio))
+
+        val result = cache.getSiteAudio(1)
+
+        assertThat(result).containsExactly(audio)
+    }
+
+    @Test
+    fun `when getting site documents, then it returns only documents`() {
+        val image = createTestMediaWithMimeType(1, "image/jpeg")
+        val document = createTestMediaWithMimeType(2, "application/pdf")
+        cache.cacheMediaList(1, listOf(image, document))
+
+        val result = cache.getSiteDocuments(1)
+
+        assertThat(result).containsExactly(document)
+    }
+
+    @Test
+    fun `when searching images by title, then it returns matching images`() {
+        val matchingImage = createTestMediaWithMimeTypeAndTitle(1, "image/jpeg", "vacation photo")
+        val nonMatchingImage = createTestMediaWithMimeTypeAndTitle(2, "image/jpeg", "work document")
+        cache.cacheMediaList(1, listOf(matchingImage, nonMatchingImage))
+
+        val result = cache.searchSiteImages(1, "vacation")
+
+        assertThat(result).containsExactly(matchingImage)
+    }
+
+    @Test
+    fun `when searching is case insensitive, then it returns matches regardless of case`() {
+        val image = createTestMediaWithMimeTypeAndTitle(1, "image/jpeg", "VACATION photo")
+        cache.cacheMediaList(1, listOf(image))
+
+        val result = cache.searchSiteImages(1, "vacation")
+
+        assertThat(result).containsExactly(image)
+    }
+
+    @Test
+    fun `when getting cache size, then it returns correct count`() {
+        cache.cacheMediaList(1, listOf(testMedia1, testMedia2))
+
+        val result = cache.getCacheSize(1)
+
+        assertThat(result).isEqualTo(2)
+    }
+
+    @Test
+    fun `when getting cache size for uncached site, then it returns zero`() {
+        val result = cache.getCacheSize(999)
+
+        assertThat(result).isEqualTo(0)
+    }
+
+    @Test
+    fun `when getting uploaded media count with mime type, then it returns filtered count`() {
+        val image1 = createTestMediaWithMimeType(1, "image/jpeg")
+        val image2 = createTestMediaWithMimeType(2, "image/png")
+        val video = createTestMediaWithMimeType(3, "video/mp4")
+        cache.cacheMediaList(1, listOf(image1, image2, video))
+
+        val result = cache.getUploadedMediaCount(1, "image")
+
+        assertThat(result).isEqualTo(2)
+    }
+
+    @Test
+    fun `when getting uploaded media count without mime type, then it returns total count`() {
+        val image = createTestMediaWithMimeType(1, "image/jpeg")
+        val video = createTestMediaWithMimeType(2, "video/mp4")
+        cache.cacheMediaList(1, listOf(image, video))
+
+        val result = cache.getUploadedMediaCount(1, null)
+
+        assertThat(result).isEqualTo(2)
+    }
+
+    @Test
+    fun `when generating local id, then it uses file path and timestamp`() {
+        val filePath = "/path/to/file.jpg"
+
+        val result = cache.generateLocalId(filePath)
+
+        val expectedCombined = "$filePath:$currentTime"
+        assertThat(result.value).isEqualTo(expectedCombined.hashCode())
+    }
+
+    @Test
+    fun `when generating local ids at different times, then they are different`() {
+        val filePath = "/path/to/file.jpg"
+
+        currentTime = 1000L
+        val result1 = cache.generateLocalId(filePath)
+
+        currentTime = 2000L
+        val result2 = cache.generateLocalId(filePath)
+
+        assertThat(result1.value).isNotEqualTo(result2.value)
+    }
+
     private fun createTestMedia(id: Int, fileName: String) =
         MediaTestUtils.createRemoteTestMedia().mediaId(id.toLong()).fileName(fileName).build()
+
+    private fun createTestMediaWithMimeType(id: Int, mimeType: String) =
+        MediaTestUtils.createRemoteTestMedia().mediaId(id.toLong()).mimeType(mimeType).build()
+
+    private fun createTestMediaWithMimeTypeAndTitle(id: Int, mimeType: String, title: String) =
+        MediaTestUtils.createRemoteTestMedia().mediaId(id.toLong()).mimeType(mimeType).title(title).build()
 }
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt	(date 1769085781073)
@@ -1,14 +1,21 @@
 package org.wordpress.android.fluxc.store
 
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.MediaModel
+import java.util.Locale
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlin.time.Clock
+import kotlin.time.ExperimentalTime
 
+@OptIn(ExperimentalTime::class)
 @Singleton
-class MediaLibraryCache @Inject constructor() {
+class MediaLibraryCache @Inject constructor(private val clock: Clock) {
     private val cache = ConcurrentHashMap<Int, List<MediaModel>>()
 
+    // CACHE OPERATIONS
+
     fun getMediaList(localSiteId: Int): List<MediaModel>? {
         return cache[localSiteId]
     }
@@ -39,4 +46,88 @@
     fun clear() {
         cache.clear()
     }
+
+    // ID GENERATION
+
+    fun generateLocalId(filePath: String): LocalId {
+        val combined = "$filePath:${clock.now().toEpochMilliseconds()}"
+        return LocalId(combined.hashCode())
+    }
+
+    // QUERY OPERATIONS
+
+    fun getSiteImages(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "image")
+    }
+
+    fun getSiteVideos(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "video")
+    }
+
+    fun getSiteAudio(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "audio")
+    }
+
+    fun getSiteDocuments(siteId: Int): List<MediaModel> {
+        return filterByMimeType(siteId, "application")
+    }
+
+    // SEARCH OPERATIONS
+
+    fun searchSiteImages(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "image", searchTerm)
+    }
+
+    fun searchSiteVideos(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "video", searchTerm)
+    }
+
+    fun searchSiteAudio(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "audio", searchTerm)
+    }
+
+    fun searchSiteDocuments(siteId: Int, searchTerm: String): List<MediaModel> {
+        return searchByMimeTypeAndTerm(siteId, "application", searchTerm)
+    }
+
+    // COUNT OPERATIONS
+
+    fun getCacheSize(siteId: Int): Int {
+        return cache[siteId]?.size ?: 0
+    }
+
+    fun getUploadedMediaCount(siteId: Int, mimeTypePrefix: String?): Int {
+        if (mimeTypePrefix == null) {
+            return getCacheSize(siteId)
+        }
+        return filterByMimeType(siteId, mimeTypePrefix).size
+    }
+
+    // HELPER METHODS
+
+    private fun filterByMimeType(siteId: Int, mimeTypePrefix: String): List<MediaModel> {
+        val allMedia = cache[siteId] ?: return emptyList()
+        return allMedia.filter { media ->
+            media.mimeType?.startsWith(mimeTypePrefix) == true
+        }
+    }
+
+    private fun searchByMimeTypeAndTerm(
+        siteId: Int,
+        mimeTypePrefix: String,
+        searchTerm: String
+    ): List<MediaModel> {
+        val allMedia = cache[siteId] ?: return emptyList()
+        val lowerSearchTerm = searchTerm.lowercase(Locale.ROOT)
+        return allMedia.filter { media ->
+            media.mimeType?.startsWith(mimeTypePrefix) == true &&
+                matchesSearchTerm(media, lowerSearchTerm)
+        }
+    }
+
+    private fun matchesSearchTerm(media: MediaModel, lowerSearchTerm: String): Boolean {
+        return (media.title?.contains(lowerSearchTerm, ignoreCase = true) == true) ||
+            media.caption.contains(lowerSearchTerm, ignoreCase = true) ||
+            media.description.contains(lowerSearchTerm, ignoreCase = true)
+    }
 }
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/module/MediaModule.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/module/MediaModule.kt b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/module/MediaModule.kt
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/module/MediaModule.kt	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/module/MediaModule.kt	(date 1769085781072)
@@ -1,21 +1,13 @@
 package org.wordpress.android.fluxc.module
 
-import dagger.Binds
 import dagger.Module
 import dagger.Provides
-import org.wordpress.android.fluxc.store.MediaIdGenerator
-import org.wordpress.android.fluxc.store.TimestampMediaIdGenerator
 import kotlin.time.Clock
 import kotlin.time.ExperimentalTime
 
 @OptIn(ExperimentalTime::class)
 @Module
-interface MediaModule {
-    @Binds
-    fun bindMediaIdGenerator(generator: TimestampMediaIdGenerator): MediaIdGenerator
-
-    companion object {
-        @Provides
-        fun provideClock(): Clock = Clock.System
-    }
+object MediaModule {
+    @Provides
+    fun provideClock(): Clock = Clock.System
 }
Index: libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java	(revision 8ba74459073db6a48f03f48166a894bec8f71f80)
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java	(date 1769085781073)
@@ -407,8 +407,6 @@
     private final WPComV2MediaRestClient mWPComV2MediaRestClient;
     private final ApplicationPasswordsMediaRestClient mApplicationPasswordsMediaRestClient;
     @NonNull private final MediaLibraryCache mMediaLibraryCache;
-    @NonNull private final MediaCacheOperations mMediaCacheOperations;
-    @NonNull private final MediaIdGenerator mMediaIdGenerator;
 
     private final ApplicationPasswordsConfiguration mApplicationPasswordsConfiguration;
 
@@ -420,17 +418,13 @@
             ApplicationPasswordsMediaRestClient applicationPasswordsMediaRestClient,
             ApplicationPasswordsConfiguration applicationPasswordsConfiguration,
             @NonNull FluxCCrashLogger crashLogger,
-            @NonNull MediaLibraryCache mediaLibraryCache,
-            @NonNull MediaCacheOperations mediaCacheOperations,
-            @NonNull MediaIdGenerator mediaIdGenerator) {
+            @NonNull MediaLibraryCache mediaLibraryCache) {
         super(dispatcher);
         mWPComV2MediaRestClient = wpv2MediaRestClient;
         mApplicationPasswordsMediaRestClient = applicationPasswordsMediaRestClient;
         mApplicationPasswordsConfiguration = applicationPasswordsConfiguration;
         mCrashLogger = crashLogger;
         mMediaLibraryCache = mediaLibraryCache;
-        mMediaCacheOperations = mediaCacheOperations;
-        mMediaIdGenerator = mediaIdGenerator;
     }
 
     @Subscribe(threadMode = ThreadMode.ASYNC)
@@ -483,56 +477,56 @@
 
     @NonNull
     public MediaModel instantiateMediaModel(@NonNull MediaModel media) {
-        media.setId(mMediaIdGenerator.generate(media.getFilePath()).getValue());
+        media.setId(mMediaLibraryCache.generateLocalId(media.getFilePath()).getValue());
         return media;
     }
 
     @NonNull
     public List<MediaModel> getSiteImages(@NonNull SiteModel siteModel) {
-        return mMediaCacheOperations.getSiteImages(siteModel.getId());
+        return mMediaLibraryCache.getSiteImages(siteModel.getId());
     }
 
     @NonNull
     public List<MediaModel> getSiteVideos(@NonNull SiteModel siteModel) {
-        return mMediaCacheOperations.getSiteVideos(siteModel.getId());
+        return mMediaLibraryCache.getSiteVideos(siteModel.getId());
     }
 
     @NonNull
     public List<MediaModel> getSiteAudio(@NonNull SiteModel siteModel) {
-        return mMediaCacheOperations.getSiteAudio(siteModel.getId());
+        return mMediaLibraryCache.getSiteAudio(siteModel.getId());
     }
 
     @NonNull
     public List<MediaModel> getSiteDocuments(@NonNull SiteModel siteModel) {
-        return mMediaCacheOperations.getSiteDocuments(siteModel.getId());
+        return mMediaLibraryCache.getSiteDocuments(siteModel.getId());
     }
 
     @NonNull
     public List<MediaModel> searchSiteImages(
             @NonNull SiteModel siteModel,
             @NonNull String searchTerm) {
-        return mMediaCacheOperations.searchSiteImages(siteModel.getId(), searchTerm);
+        return mMediaLibraryCache.searchSiteImages(siteModel.getId(), searchTerm);
     }
 
     @NonNull
     public List<MediaModel> searchSiteVideos(
             @NonNull SiteModel siteModel,
             @NonNull String searchTerm) {
-        return mMediaCacheOperations.searchSiteVideos(siteModel.getId(), searchTerm);
+        return mMediaLibraryCache.searchSiteVideos(siteModel.getId(), searchTerm);
     }
 
     @NonNull
     public List<MediaModel> searchSiteAudio(
             @NonNull SiteModel siteModel,
             @NonNull String searchTerm) {
-        return mMediaCacheOperations.searchSiteAudio(siteModel.getId(), searchTerm);
+        return mMediaLibraryCache.searchSiteAudio(siteModel.getId(), searchTerm);
     }
 
     @NonNull
     public List<MediaModel> searchSiteDocuments(
             @NonNull SiteModel siteModel,
             @NonNull String searchTerm) {
-        return mMediaCacheOperations.searchSiteDocuments(siteModel.getId(), searchTerm);
+        return mMediaLibraryCache.searchSiteDocuments(siteModel.getId(), searchTerm);
     }
 
     //
@@ -611,7 +605,7 @@
         int offset = 0;
         if (payload.loadMore) {
             String mimeTypeValue = payload.mimeType != null ? payload.mimeType.getValue() : null;
-            offset = mMediaCacheOperations.getUploadedMediaCount(payload.site.getId(), mimeTypeValue);
+            offset = mMediaLibraryCache.getUploadedMediaCount(payload.site.getId(), mimeTypeValue);
         }
         if (payload.site.getOrigin() == SiteModel.ORIGIN_WPCOM_REST) {
             mWPComV2MediaRestClient.fetchMediaList(payload.site, payload.number, offset, payload.mimeType);
```

</details>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/TimestampMediaIdGenerator.kt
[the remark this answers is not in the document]
ParaskP7: > Additionally, different file paths with timestamps could theoretically produce the same hash.

Thinking (🤔): I think I agree that this theoretical problem might not be easily achievable. I also understand that if that problem happens, it will be nearly impossible to debug. Having said both, I too feel that we could leave this be (as is).

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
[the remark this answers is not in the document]
ParaskP7: Praise (❤️): An awesome job you did here, first setting up the test, making sure it fails, then addressing this with `compute`, kudos! 🙇

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, good job! 🌟 x 🌟 ^ 🌟

-----

I have left few questions (❓), a couple of suggestions (💡) and some minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
wzieba: Good point 👍! Fixed in c92a68dbfc3898ac33b780efae7ddb8f13ae285

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaCacheOperations.kt
wzieba: 👍 38dfa496b55dd6ae24c4b33b17591fce71acdea

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
wzieba: Ah thanks! If it's only used in tests, I decided to remove it: 6ebd22e26837c80370d016dfe6d79336ee45832

libs/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
wzieba: 👍 3f95039484f327680b68931a622099326055ae6

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/MediaLibraryCacheTest.kt
wzieba: 👍 f60ab06f48c51509e0e47a3d85c1fa9db0cf093

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaLibraryCache.kt
wzieba: > Suggestion (💡): Why the extra `Library` word in it and not having it simply be called `MediaCache`, similar to how `MediaCacheOperations` is named?

I used `Library` because this cache is (at least currently?) only used for WordPress Media Library feature, but there's nothing in this class that makes it usable only for Media Library. I renamed to `RemoteMediaCache` because it better suits its purpose: `Remote*` because it operates on `MediaModel#mediaId` which is set by backend only.

> Suggestion (💡): What not combining all these 3 `MediaLibraryCache`, `MediaCacheOperations` and `MediaIdGenerator` classes into 1 `MediaCache` class. I asked AI to do that for me and here's how it looks (I find `MediaStore.java` looking simpler for it (see `AI Patch` below):

I examined this but I'd actually prefer to leave them 3 seperate, as I'm a voter for single responsibility
- I'd like to keep `Cache` as simple as possible: add/remove media models there only. Also, it doesn't have to do anything with `MediaIdGenerator`, so this could add a point of confusion to combine them in the same class.
- `MediaCacheOperations` - it's "just" a "Kotlin extension" of `MediaStore` - I hope that when working on [AINFRA-1764: Address remaining Media* improvements in Woo Android](https://linear.app/a8c/issue/AINFRA-1764/address-remaining-media-improvements-in-woo-android) we'll simply transform `MediaCacheOperations` to `MediaStore.**kt**`

ParaskP7: > I examined this but I'd actually prefer to leave them 3 seperate, as I'm a voter for single responsibility

Yea @wzieba , actually after commenting on that I realized that I had something better in mind, to keep these 3 classes separate (single responsibility), but use them via `RemoteMediaCache`. The main thing I was trying to do is to have `MediaStore` use a single field, the `RemoteMediaCache`, and not 3 of them. 👍 (not a bit deal)

wzieba: Gotcha, thanks for more context! We can always iterate on this later, but for me a big thing was to not make `RemoteMediaCache` know _anything_ about `MediaIdGenerator` as id generator is **only** for local models and cache is **only** for remote.

ParaskP7: Gotcha back, thanks! 💯

PR #15220, opened 2026-01-20
[WOOMOB-2020] Fix tracking events firing incorrectly during POS connection flow

kidinov:
Closes WOOMOB-2020

### Description

Fixes two tracking issues during the card reader connection flow:

1. **`card_present_onboarding_completed` firing multiple times**
   - The event was tracked every time `getOnboardingState()` was called, including when returning cached results
   - Fixed by moving tracking inside the fetch block so it only fires when state is freshly fetched from server

2. **`software_update_failed` firing incorrectly with "Unknown software update status"**
   - The `Unknown` status is the initial/idle state of the `SoftwareUpdateStatus` StateFlow, not an actual failure
   - When `listenToSoftwareUpdateStatus()` started collecting, it immediately received this initial `Unknown` value and tracked it as a failure
   - Removed tracking for `Unknown` status entirely since it only occurs:
     - As the initial StateFlow value (before any update starts)
     - When `resetConnectionState()` is called after reader disconnection (cleanup)
   - Neither case represents an actual update failure. @malinajirka wdyt about that? It changes tracking for the card reader flow as well

### Test Steps

1. Connect to a card reader in POS
2. Verify `card_present_onboarding_completed` fires only once (not on every `getOnboardingState()` call)
3. Verify `software_update_failed` does not fire with "Unknown software update status" during normal connection flow

### Images/gif

N/A - Analytics tracking changes only

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/onboarding/CardReaderOnboardingChecker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/onboarding/CardReaderOnboardingChecker.kt
@@ -107,9 +107,8 @@ class CardReaderOnboardingChecker @Inject constructor(
                         state.preferredPlugin,
                         version
                     )
+                    paymentsFlowTracker.trackOnboardingState(state)
                 }
-        }.also {
-            paymentsFlowTracker.trackOnboardingState(it)
         }
     }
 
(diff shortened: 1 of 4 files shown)


malinajirka (approved):
The changes look good to me, thanks @kidinov. I think skipping tracking for `unknown` update is fine.

Btw I noticed we don't track anything on the `Request location and bluetooth` dialogs. I guess this is how it worked even before, which is surprising, but it might still be worth double checking it's consistent with iOS.

kidinov:
Thanks @malinajirka 

> Btw I noticed we don't track anything on the Request location and bluetooth dialogs. I guess this is how it worked even before, which is surprising, but it might still be worth double checking it's consistent with iOS.

I checked, and the iOS has:

```
- card_reader_location_permission_pre_alert_shown                                                                                                                                                                                                                                                                  
  - card_reader_location_permission_required_shown  
 ```

But seems nothing for BT. Issue - WOOMOB-2035

PR #15222, opened 2026-01-21
Update how the app decides whether to accept a push notification

irfano:
Closes WOOMOB-1957

### Description

This PR adds deduplication logic to `NotificationMessageHandler` to prevent duplicate push notifications. When `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag is enabled, we check if the incoming notification is from WPCOM (note_id > 0) and if the device is already registered with Woo Core (hasPushToken()). If both conditions are true, we skip the WPCOM notification since the same event will also arrive via Woo Core.

### Test Steps

1. Set up a store with the new Woo Core endpoints by following these steps: p91TBi-dyW-p2#comment-14564
2. Install the app and log into your store.
3. Submit a product review from the web interface.
4. **Regression:** Confirm you receive the notification, because the feature flag is disabled.
5. Apply this patch to enable the feature flag: [Enable_WOO_PUSH_NOTIFICATIONS_SYSTEM_feature_flag_for_debug_builds.patch](https://github.com/user-attachments/files/24753586/Enable_.WOO_PUSH_NOTIFICATIONS_SYSTEM._feature_flag_for_debug_builds.patch)
6. Rebuild the app.
7. Logout and log in again the app to trigger new push token registration. (you may need to kill and restart after login)
6. Submit another product review.
8. Verify that you do not receive a notification. (Note: You won't receive it yet because the development plugin is still a work in progress.)

### Images/gif


JorgeMucientes:
Hey @irfano the code changes look good, however, I think we can do this in a cleaner way once we merge the changes from the PR I'll open tomorrow early morning (I'm wrapping up testing changes): https://github.com/woocommerce/woocommerce-android/pull/15238.
With the changes from that PR we can directly use `PushNotificationRegistrationStatus` class to verify if the user is registered in Woo system. That will avoid having to change the code you added here once we remove the feature flag. In fact relying on `PushNotificationRegistrationStatus` use case is feature flag agnostic, 
Let me know what you think.

irfano:
I've rebased onto your branch. Let's merge yours first since this is smaller. I agree that using `PushNotificationRegistrationStatus` in this PR is cleaner.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -80,24 +82,32 @@ class NotificationMessageHandler @Inject constructor(
...
         val notificationModel = notificationsParser.buildNotificationModelFromPayloadMap(messageData)
         if (notificationModel == null) {
             wooLog.e(NOTIFICATIONS, "Notification data is empty!")
             return
         }
 
         val notification = notificationModel.toAppModel(resourceProvider)
-        if (notification.remoteNoteId == 0L) {
+        val isRegisteredForWooPush = runBlocking {
+            registrationStatus(notification.remoteSiteId) == PushNotificationRegistrationStatus.Status.WOO_REGISTERED
```
JorgeMucientes: I think this check should also include the case were user is registered in both systems, right? 


```suggestion
        val isRegisteredForWooPush = runBlocking {
            val registrationStatus = registrationStatus(notification.remoteSiteId)
            registrationStatus == PushNotificationRegistrationStatus.Status.REGISTERED_IN_BOTH 
                || registrationStatus == PushNotificationRegistrationStatus.Status.WOO_REGISTERED
        }
```
The most likely case were we get duplicates is when the user is registered in both systems (even though PN should be turned off in WP.com system)

JorgeMucientes (changes requested):

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: 🤔 You're right. I'm confused about `WOO_REGISTERED` would be true even though it's registered in both. I renamed (82eb6088940d61d56028e254bbcd31faaee612cb) the enums and added `ONLY` suffix to avoid any confusion in the future. Does that sound good?
Also fixed the logic: 0f667f42a1b5c260058a833971919f48d2697e67

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -80,24 +82,32 @@ class NotificationMessageHandler @Inject constructor(
...
         val notificationModel = notificationsParser.buildNotificationModelFromPayloadMap(messageData)
         if (notificationModel == null) {
             wooLog.e(NOTIFICATIONS, "Notification data is empty!")
             return
         }
 
         val notification = notificationModel.toAppModel(resourceProvider)
-        if (notification.remoteNoteId == 0L) {
+        val isRegisteredForWooPush = runBlocking {
+            registrationStatus(notification.remoteSiteId) == PushNotificationRegistrationStatus.Status.REGISTERED_BOTH
```
JorgeMucientes: I think we should still include `PushNotificationRegistrationStatus.Status.REGISTERED_WOO_ONLY` in this check too. 
While it might be a corner case I can see how we can end in a situation were locally, the app thinks the user is only registered in WOO system only, however their site might still be registered in both PN systems in the backend side. For example, if user is registered in WP.com system and fails to unregister device on log out and then logs in again and registers successfully in Woo system but fails to turn off notifications in WP.com. 
That's just one case but I think there could be others. To stay on the safe side I'd check either values `REGISTERED_BOTH || REGISTERED_WOO_ONLY`

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: Please don’t follow commits for the review, I’ve synced with the base branch many times since this PR has been open for a while.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
JorgeMucientes: I'm not sure what happened exactly, but when I added this comment to the PR this morning I would've sworn that the `PushNotificationRegistrationStatus.Status.REGISTERED_WOO_ONLY` was not present in the code. Look like it is now. So all good 👍🏼

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
JorgeMucientes, line 73: Out of scope for this PR but we probably need to update this as this condition will be false for all users loggin in with site credentials. 
Can you create a nwe Linear task and tackle that @irfano 🙏🏼

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: I actually included `PushNotificationRegistrationStatus.Status.REGISTERED_WOO_ONLY` and `PushNotificationRegistrationStatus.Status.REGISTERED_BOTH` after your feedback, but I think I messed it up while resolving conflicts. 
I re-added it today in commit `891dff474a77593f0da8181b34a65cd48905144a`. That's why I [commented](https://github.com/woocommerce/woocommerce-android/pull/15222#discussion_r2745748226) "don't follow my commits" 😅
Anyway, thanks for the review.

JorgeMucientes (approved):
LGTM! Nice work @irfano

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: 👍🏻  Opened: WOOMOB-2113

PR #15236, opened 2026-01-22
[WOOMOB-2022] Add close button to Connecting state in card reader dialog

kidinov:
WOOMOB-2022

### Description

Adds the close (X) button to the Connecting state in the card reader connection dialog. Previously, the Connecting state had no visible dismiss control, but users could still dismiss by tapping outside the dialog. This change makes the dismiss behavior consistent with other dialog states like Scanning and ReaderFound.

### Test Steps

1. Go to WooPOS and initiate card reader connection
2. When the "Connecting to reader" dialog appears, verify the X button is visible in the top-right corner
3. Tap the X button and verify the dialog dismisses


malinajirka (approved):
LGTM, thanks!

PR #15238, opened 2026-01-22
Issue/woomob 2042 add fallback to wpcom pn system

JorgeMucientes:
Closes WOOMOB-2042

## Description

This PR adds a fallback mechanism to the WP.com push notification registration system when the Woo Core push notification registration fails. It also refactors the push notification registration status logic to track both registration systems simultaneously.

### Changes

**Push Notification Registration Status Refactoring:**
- Updated `PushNotificationRegistrationStatus` to return a more granular status instead of a simple binary state:
  - `WOO_REGISTERED` - Only registered in Woo Core system
  - `WPCOM_REGISTERED` - Only registered in WP.com system
  - `REGISTERED_IN_BOTH` - Registered in both systems
  - `UNREGISTERED` - Not registered in either system
- Removed dependency on `FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM` for status checks, now evaluating both systems independently

**Fallback Mechanism:**
- When Woo Core push token registration fails, the app now automatically falls back to the legacy WP.com registration system
- Added logging to track registration failures and fallback behavior

## Test Steps

**Case 1:** Register into WP.com as usual
1. Clean install of the app with `WOO_PUSH_NOTIFICATIONS_SYSTEM -> false`
2. Log in using WP.com credentials .
3. Check use is registered as normal in WP.com system. 
4. Check PN works as expected in WP.com sytem
5. Expected logs: 
```
Current PN registration status: UNREGISTERED
Updated push notification registration status: UNREGISTERED
Current PN registration status: UNREGISTERED
Registering FCM token in WPCOM instance: cvhs71fwRy-X7FPK_KZRq_:APA91bE...
Updated push notification registration status: WPCOM_REGISTERED
```

**Case 2:** 
1. Enable `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag in the code
2. Build and launch the app. See the following logs attempting to register in Woo Core but failing to do so because the Woo Core version of the plugin doesn't support it. After failing to register in Woo core system, we don't fallback to WP.com as the user is already registered there.
```
Current PN registration status: WPCOM_REGISTERED
Registering FCM token in Woo Core instance: cvhs71fwRy-X7FPK_KZRq_:APA91bE-YQUAAh8FZ-jwofbVJPOM
Woo Core push token registration failed:No route was found matching the URL and request method.
Updated push notification registration status: WPCOM_REGISTERED
```
3. Send the app to the background and restart it. You'll see that every time we do this we re-attempt to register the user in Woo Core, which will fail, and because of that re-attempt to register them in WP.com despite being registered already. We are discussion different options to mitigate this over Slack. Whatever solution we come up with for this will be addressed in a different PR. 

**Case 3:** 
1. Reusing the same log in session from Case 2, switch to a site that has Woo Core plugin version supporting new PN
2. See how the site is successfully registered in the new Woo Core system. Expected logs: 
```
Current PN registration status: WPCOM_REGISTERED
Registering FCM token in Woo Core instance: 
WPCom notifications disabled for site 251754054
Updated push notification registration status: REGISTERED_IN_BOTH
```

### Missing tasks to address
- Mitigate attempting to register in Woo PN system every time the app is resumed from background. Solution TBD


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRegistrationStatus.kt
```diff
@@ -12,25 +11,24 @@ class PushNotificationRegistrationStatus @Inject constructor(
...
-            if (deviceId.isNotNullOrEmpty()) {
-                Status.REGISTERED
-            } else {
-                Status.UNREGISTERED
-            }
+    suspend operator fun invoke(): Status {
+        val wpComPushServerId = prefsWrapper.getFluxCPreferences()
+            .getString(WpComPushNotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
+        val siteId = selectedSite.getIfExists()?.siteId
+        val isWooRegistered = siteId != null && pushNotificationRepository.isWooPushTokenRegisteredForSite(siteId)
```
irfano, line 15: We should probably update this. `invoke()` assumes we're checking the selected site, but we might need to check the status of any site (e.g., for notifications from unselected sites). We could add a `siteId` parameter here, or handle it differently if you prefer.

JorgeMucientes: Makes sense!  Applied

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -62,6 +65,14 @@ class PushNotificationRepository @Inject constructor(
...
                     )
                 }
+            } else {
+                WooLog.w(
+                    WooLog.T.NOTIFICATIONS,
+                    "Woo Core push token registration failed:${result.error?.message}, fallback to WPCom"
+                )
+                if (!isWpComPushRegistered()) {
+                    registerPushTokenInWpComSystem(token)
+                }
```
JorgeMucientes: This is something I forgot in my prior PR. If user is already WP.com registered avoid registering again.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -529,6 +531,7 @@ class SitePickerViewModel @Inject constructor(
 
                                     trackLoginEvent(currentStep = UnifiedLoginTracker.Step.SUCCESS)
                                     appPrefsWrapper.removeLoginSiteAddress()
+                                    registerDevice(RegisterDevice.Mode.FORCEFULLY)
```
JorgeMucientes: This fixes several issues: 
- Not registering automatically in Woo Core system when first time loggin in, because the site was not selected yet. 
- Automatically trigger PN token registration when switching sites.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -46,6 +47,11 @@ class RegisterDevice @Inject constructor(
                 pushNotificationRepository.registerPushTokenInWpComSystem(token)
             }
         }
+        WooLog.d(
+            WooLog.T.NOTIFICATIONS,
+            "Updated push notification registration status: " +
+                "${pushNotificationRegistrationStatus(selectedSite.getIfExists()?.siteId)}"
+        )
```
irfano: I think it's better to move this log inside the `token.isNotEmpty()` block.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -163,4 +173,10 @@ class PushNotificationRepository @Inject constructor(
             appPrefsWrapper.wooCorePushDeviceUUID = it
         }
     }
+
+    private fun isWpComPushRegistered(): Boolean {
+        val wpComPushServerId = prefsWrapper.getFluxCPreferences()
+            .getString(WpComPushNotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
+        return wpComPushServerId.isNotNullOrEmpty()
+    }
```
irfano, line 188: Can we use `PushNotificationRegistrationStatus` here instead of duplicating the same logic?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
irfano: I think this caused **duplicate registration**. I noticed sendToken is being triggered twice in **Case 1**, I saw the _Registering FCM token in WPCOM instance:_ log twice. I couldn’t reproduce it on trunk, so it was likely introduced by this PR. Could you please take a look?

irfano (changes requested):

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
JorgeMucientes: `PushNotificationRegistrationStatus` already depends on `PushNotificationRepository`. So adding the suggested dependency will create cyclic dependency preventing Hilt or any DI framework to compile. Moreover, the repository shouldn't depend on the use case IMHO, repository should only depend on stores (or any other data sources).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
JorgeMucientes: Good catch. I accidentally commited this change applying the wrong "policy". We still want to keep this line but call `IF_NEEDED` instead. 
```kotling
registerDevice(RegisterDevice.Mode.IF_NEEDED)
```
That will avoid duplicated registeringf.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -529,6 +531,7 @@ class SitePickerViewModel @Inject constructor(
 
                                     trackLoginEvent(currentStep = UnifiedLoginTracker.Step.SUCCESS)
                                     appPrefsWrapper.removeLoginSiteAddress()
+                                    registerDevice(RegisterDevice.Mode.IF_NEEDED)
```
JorgeMucientes, line 534: Hmm, this might be trickier than I expected. Adding "IF_NEEDED" fixed the duplicated registration issue. However, it brings back one of the problems I was trying to fix in the first place. I've created a Linear issue to address this separatedly: WOOMOB-2101

irfano (approved):
LGTM! 👍🏻 Thanks for creating the switching site issue, we can fix it separately.

PR #15242, opened 2026-01-23
Update .gitignore to exclude .grepai/ directory

malinajirka:
Update .gitignore to exclude .grepai/ directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,4 @@ google-upload-credentials.json
 .claude/
 
 *.mp4
+.grepai/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -125,3 +125,4 @@ google-upload-credentials.json
 .claude/
 
 *.mp4
+.grepai/
```
kidinov, line 128: NP: Maybe logically, keep it above mp4 to group all ai related stuff. Or just move mp4 somewhere above

kidinov (approved):
LGTM!

.gitignore
malinajirka: Fixed in the latest commit

PR #15247, opened 2026-01-26
[WOOMOB-2024] Make URLs clickable in card reader connection error messages

kidinov:
Closes WOOMOB-2024

### Description

Makes URLs in card reader connection error messages clickable. When the Stripe SDK returns an error message containing a URL (e.g., when a reader is not supported in the user's country), the URL is now rendered as a tappable link

### Test Steps

1. Connect a card reader that is not supported in your country (e.g., BBPOS WisePad 3 in the US)
2. Observe the "Connection failed" error dialog
3. Verify the Stripe documentation URL is displayed as a clickable link (underlined, primary color)
4. Tap the link and verify it opens in a Chrome Custom Tab

### Images/gif

https://github.com/user-attachments/assets/4e4a8550-05aa-481a-b695-6a2e1d1ba461


malinajirka (approved):
LGTM, thanks!

PR #15256, opened 2026-01-27
[Woo POS] Fix POS email receipt changing refunded order status to completed

staskus:
WOOMOB-2044

### Description

Fixes an issue where sending an email receipt for a refunded POS order would change the order status from "Refunded" to "Completed".

**Root Cause:**
The previous implementation used `createOrUpdateOrder` to update the billing email, which sent the entire order object to the server. After refund, an order in a database was oudated and still had a "Completed" order status. When we sent it to the backend, it was overwritten from "Refunded" back to "Completed". 

**Solution:**
Added a new `updateOrderBillingEmail` method that only updates the billing email field without sending the entire order. This matches the iOS implementation and prevents unintended status changes.

### Changes
- Added `updateOrderBillingEmail` to `OrderRestClient` and `WCOrderStore`
- Updated `WooPosEmailReceiptRepository` to use the new method
- Removed unused dependencies (`OrderCreateEditRepository`, `OrderMapper`)

### Test Steps

1. Create a POS order and complete payment
2. Issue a refund for the order
3. From the orders list, select the refunded order
4. Send an email receipt
5. Verify the order status remains "Refunded" (not changed to "Completed")

https://github.com/user-attachments/assets/67385d55-feba-4536-b758-1ed0b15757fb

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -1086,6 +1086,28 @@ class OrderRestClient @Inject constructor(
         return response.toWooPayload { it }
     }
 
+    suspend fun updateOrderBillingEmail(
+        site: SiteModel,
+        orderId: Long,
+        email: String
+    ): WooPayload<Unit> {
+        val url = WOOCOMMERCE.orders.id(orderId).pathV3
+        val body = mapOf(
+            "billing" to mapOf(
+                "email" to email
+            )
+        )
+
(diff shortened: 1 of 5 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -1086,6 +1086,28 @@ class OrderRestClient @Inject constructor(
         return response.toWooPayload { it }
     }
 
+    suspend fun updateOrderBillingEmail(
```
staskus, line 1089: This is what we do on iOS. Do you still prefer to pass an Order object on the Android side?

samiuelson (approved):

PR #15258, opened 2026-01-28
[Woo POS] Pass email directly to POS receipt endpoint for WC 10.0.0+

staskus:
WOOMOB-2044

Could be tested together https://github.com/woocommerce/woocommerce-android/pull/15256 if needed

### Description

Enhancement to the POS email receipt fix. For WooCommerce 10.0.0+, this change passes the email directly to the `send_email` endpoint with `force_email_update`, instead of separately updating the billing email first.

- Passing `email` and `force_email_update` directly to `sendOrderPOSSpecificReceipt`
- Skipping the order update step
- Matching the iOS implementation
- Legacy WC versions still use `updateOrderBillingEmail` + `sendOrderReceipt`

### Changes
- Added `email` and `forceEmailUpdate` params to `sendOrderPOSSpecificReceipt` (OrderRestClient, WCOrderStore)
- Updated `WooPosEmailReceiptRepository` to use direct email passing for WC 10.0.0+
- Added tests for the new behavior

### Test Steps

Test with both >=10.0 and <10.0:

1. Create a POS order and complete payment
2. Issue a refund for the order
3. From the orders list, select the refunded order
4. Send an email receipt
5. Verify the order status remains "Refunded" (not changed to "Completed")
6. Verify the email is received correctly

#### For >=10.0

The new email is successfully sent, the order status has not changed

image

#### For <10.0 (Fallback)

This old email is successfully sent, the order status has not changed

image

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -1073,13 +1073,17 @@ class OrderRestClient @Inject constructor(
     suspend fun sendOrderPOSSpecificReceipt(
         site: SiteModel,
         orderId: Long,
+        email: String,
+        forceEmailUpdate: Boolean = true
     ): WooPayload<Unit> {
         val response = wooNetwork.executePostGsonRequest(
             site = site,
             path = WOOCOMMERCE.orders.id(orderId).actions.send_email.pathV3,
             clazz = Unit::class.java,
             body = mapOf(
-                "template_id" to "customer_pos_completed_order"
+                "template_id" to "customer_pos_completed_order",
+                "email" to email,
+                "force_email_update" to forceEmailUpdate
             )
         )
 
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -1073,13 +1073,17 @@ class OrderRestClient @Inject constructor(
     suspend fun sendOrderPOSSpecificReceipt(
         site: SiteModel,
         orderId: Long,
+        email: String,
+        forceEmailUpdate: Boolean = true
```
samiuelson: 💡 n.p.: `sendOrderPOSSpecificReceipt` is used only in one place in prod code (in `WooPosEmailReceiptRepository`). Since we're not trying to provide backward compatibility of some sort here, I suggest removing default value from `forceEmailUpdate`, and passing explicit value.

samiuelson (approved):
Thanks, @staskus! LGTM. I added one suggestion, take a look.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
staskus: Yes, definitely makes sense.

PR #15263, opened 2026-01-29
[Woo POS][Refunds] Improve refund flow state management

samiuelson:
WOOMOB-1951

## Description
This PR improves the refund flow state management. Instead of loading refund data immediately when the ViewModel is created, data is now loaded only when the dialog is actually opened. This ensures proper state reset when the dialog is dismissed.

## Test Steps
Verify that POS refund flow state is reset when refund dialog is being dismissed in different steps. Previously the refund step was not reset in Error state. Now, when Issue refund button is clicked, refund flow should always start by loading refundable items.

### Images/gif

—


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
```diff
@@ -200,7 +200,7 @@ class WooPosRefundViewModelTest {
 
         // WHEN
         viewModel = createViewModel()
-        advanceUntilIdle()
+        viewModel.onUIEvent(WooPosRefundUIEvent.DialogOpened)
```
malinajirka: 🔍 It might be a bit cleaner to trigger this event within the `createViewModel` method, we could perhaps have a boolean parameter that would allow us to disable this behavior. Wdyt?

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModel.kt
```diff
@@ -229,6 +223,7 @@ class WooPosRefundViewModel @AssistedInject constructor(
                 _state.value = currentState.copy(step = WooPosRefundState.Content.RefundStep.ReviewRefund)
             WooPosRefundUIEvent.OnRefundConfirmed -> processRefund(currentState)
             WooPosRefundUIEvent.DialogDismissed -> Unit
+            WooPosRefundUIEvent.DialogOpened -> loadRefundableItems()
```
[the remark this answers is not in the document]
samiuelson: done: [b78e7e8](https://github.com/woocommerce/woocommerce-android/pull/15263/commits/b78e7e809ea177c1641e799bbe63b0200b2b95c5)

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosRefundViewModelTest.kt
samiuelson: Good idea, done in [2444cfc](https://github.com/woocommerce/woocommerce-android/pull/15263/commits/2444cfcca16561fcebf3d7da4e63403ca39d9e2c)

PR #15264, opened 2026-01-29
[WOOMOB-2094] Fix battery level discrepancy between main app and WooPOS

kidinov:
WOOMOB-2094

### Description

The WooPOS settings screen was showing different battery levels compared to the main app for the same card reader. This happened because WooPOS only read the battery level from the CardReader object captured at connection time and never updated it afterward.

The main app subscribes to the `batteryStatus` flow which receives live battery updates from the Stripe Terminal SDK. This PR adds the same subscription pattern to WooPOS so both screens show consistent battery levels.

### Test Steps

1. Enable Developer Options > Simulated Card Reader
2. Set "Simulated reader update" to "Low battery, succeed connect"
3. Connect to the simulated reader in WooPOS
4. Go to WooPOS Settings > Hardware > Card Reader
5. Verify the battery level is 15% (not 80 as used to be)
6. Compare with the main app's card reader details - both should show the same battery level


samiuelson (approved):
LGTM!

PR #15265, opened 2026-01-29
[Woo POS][Refunds] Add payment method text to refund flow

samiuelson:
## Description

This PR replaces the hardcoded payment method placeholder text in the WooPOS refund flow with real payment method information. The implementation retrieves the actual payment gateway and card details used for the order and displays them in the format `"Via {Gateway Title} ({Brand} **** {last4})"` when available.

## Changes
- Created `WooPosGetPaymentMethod` use case to handle payment method retrieval logic
- Updated `WooPosRefundViewModel` to load payment method via the new use case and pass it to the refund state
- Replaced hardcoded payment method text in `WooPosIssueRefundDialog` with dynamic display using `state.paymentMethod`
- Added string resource `woopos_orders_via_payment_method` for consistent payment method formatting

## Test Steps
1. Create an order and process payment with a card payment method
2. Open the order in WooPOS and initiate a refund
3. Navigate to the Review Refund step
4. Verify the payment method is displayed correctly with the card brand and last 4 digits (e.g., "Via Credit Card (Visa **** 4242)")
5. Test with different payment methods (cash, card) and verify appropriate fallback text is shown

## img
image


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetPaymentMethod.kt
```diff
@@ -0,0 +1,77 @@
...
+        )
+    }
+
+    private suspend fun loadPaymentGateway(order: Order): PaymentGateway = withContext(coroutineDispatchers.io) {
+        val paymentGateway = gatewayStore.getGateway(selectedSite.get(), order.paymentMethod)?.toAppModel()
+        return@withContext if (paymentGateway != null && paymentGateway.isEnabled) {
+            paymentGateway
+        } else {
+            PaymentGateway(methodTitle = REFUND_METHOD_MANUAL)
+        }
```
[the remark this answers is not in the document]
samiuelson: Done: [aa9ab69](https://github.com/woocommerce/woocommerce-android/pull/15265/commits/aa9ab69a137ac3188dd2077a03b2a71c5e0841b4)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetPaymentMethod.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        return enrichRefundMethodWithCardDetails(
+            order,
+            paymentGateway.title.ifBlank { paymentGateway.methodTitle }
+        )
+    }
+
+    private suspend fun loadPaymentGateway(order: Order): PaymentGateway = withContext(coroutineDispatchers.io) {
+        gatewayStore.getGateway(selectedSite.get(), order.paymentMethod)?.toAppModel()
+            ?: PaymentGateway(methodTitle = REFUND_METHOD_MANUAL)
```
samiuelson: 💡 `REFUND_METHOD_MANUAL` comes from `RefundSummaryViewModel` (store management). I'm not sure if we should do the same in POS. It's a defensive approach since we expect gateway to be present in the store.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosLoadPaymentMethodTest.kt
```diff
@@ -0,0 +1,267 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.gateways.WCGatewayModel
+import org.wordpress.android.fluxc.store.WCGatewayStore
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class WooPosLoadPaymentMethodTest {
```
malinajirka: 📓 Seems like the test name doesn't match the SUT name. Should we rename it to `WooPosGetPaymentMethodTest`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetPaymentMethod.kt
malinajirka: I personally prefer failing or crashing for unexpected states - if the flow continues with a wrong gateway, we will likely never learn about an issue. We could fail the refund flow without a crash + report the issue to Sentry. Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetPaymentMethodTest.kt
```diff
@@ -0,0 +1,267 @@
...
+            paymentChargeRepository = paymentChargeRepository,
+            gatewayStore = gatewayStore,
+            selectedSite = selectedSite,
+            resourceProvider = resourceProvider,
+            coroutineDispatchers = coroutineDispatchers
+        )
+    }
+
+    @Test
+    fun `given enabled gateway with card details, when invoke, then returns formatted payment method with card info`() =
```
malinajirka, line 66: 📓 The tests are missing // GIVEN // WHEN // THEN comments. I'm personally not a fan of them anymore, but we should likely add them for consistency.

malinajirka (approved):
Thanks, lgtm!

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosGetPaymentMethodTest.kt
samiuelson: Updated: 0bead0c1cb9539eb291b4bf6efd3e4ccb775144e

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosLoadPaymentMethodTest.kt
samiuelson: Updated: 0bead0c1cb9539eb291b4bf6efd3e4ccb775144e

PR #15273, opened 2026-01-30
[WOOMOB-2068] Sync FTS index with local catalog

kidinov:
WOOMOB-2068

### Description

Implements FTS index synchronization with the local product catalog. The FTS (Full-Text Search) index is kept in sync during catalog synchronization:

- **Full sync (paginated)**: FTS index is cleared and rebuilt atomically within the same database transaction
- **Full sync (file-based)**: FTS index is populated after storing catalog data from the downloaded file
- **Incremental sync**: Only changed items are updated in FTS (no full rebuild), including:
  - Updated products and variations are re-indexed
  - Products hidden from POS (`productsToRemove`) are deleted from FTS
  - Trashed products are deleted from FTS
- **On-demand FTS table insertion**: Existing data is migrated to FTS on POS launch if the FTS table is empty but products exist

Key implementation details:
- FTS sync runs inside `executeInTransaction` for data consistency
- File-based sync populates FTS after `storeCatalogData` completes
- `WooPosLocalCatalogSyncWithFts` handles all FTS operations with feature flag checks
- Products and variations are indexed with name, SKU, barcode, and attribute values
- Variation attributes are parsed from JSON and indexed for search
- Added `getProductsByIds` to fetch parent product names for variations during incremental sync
- Added logging for clear execution path visibility

Note: Products that are completely deleted from the backend (not just trashed) are only cleaned up from FTS during full sync, as there's no way to detect them during incremental sync.

### Test Steps

1. Enable the `isPosProductsFtsEnabled` feature flag
2. Open POS and wait for catalog to sync
3. Verify in logs that FTS index is populated (look for "syncFtsForFullSync completed" messages during sync, not after)
4. Force close the app and reopen - verify `ensureFtsPopulated` skips since FTS is already populated
5. Make changes to products on the backend (update name, hide from POS, trash a product, etc.)
6. Trigger incremental sync and verify FTS updates only for changed items (look for "syncFtsForIncrementalSync" messages)

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
@@ -146,8 +146,15 @@ class WooPosLocalCatalogSyncWithFts @Inject constructor(
             parentProducts.forEach { productNamesMap[it.remoteId.value] = it.name }
         }
 
-        val variationFtsEntities = variations.map { variation ->
-            val parentName = productNamesMap[variation.remoteProductId.value] ?: ""
+        val variationFtsEntities = variations.mapNotNull { variation ->
+            val parentName = productNamesMap[variation.remoteProductId.value]
+            if (parentName == null) {
+                logger.w(
+                    "Skipping variation ${variation.remoteVariationId.value}: " +
+                        "parent product ${variation.remoteProductId.value} not found"
+                )
+                return@mapNotNull null
+            }
             variation.toFtsEntity(siteIdString, parentName)
         }
 
@@ -165,8 +172,15 @@ class WooPosLocalCatalogSyncWithFts @Inject constructor(
 
         val productNamesMap = products.associate { it.remoteId.value to it.name }
 
-        val variationFtsEntities = variations.map { variation ->
-            val parentName = productNamesMap[variation.remoteProductId.value] ?: ""
+        val variationFtsEntities = variations.mapNotNull { variation ->
+            val parentName = productNamesMap[variation.remoteProductId.value]
+            if (parentName == null) {
+                logger.w(
+                    "Skipping variation ${variation.remoteVariationId.value}: " +
+                        "parent product ${variation.remoteProductId.value} not found"
+                )
+                return@mapNotNull null
+            }
             variation.toFtsEntity(siteIdString, parentName)
         }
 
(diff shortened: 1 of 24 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+        if (products.isNotEmpty()) {
+            val productIds = products.map { it.remoteId.value.toString() }
+            ftsDao.deleteProducts(siteIdString, productIds)
+        }
+
+        if (productsToRemove.isNotEmpty()) {
+            val productIdsToRemove = productsToRemove.map { it.value.toString() }
+            ftsDao.deleteProducts(siteIdString, productIdsToRemove)
```
malinajirka, line 80: ❓ Do we need to remove products' variations here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
```diff
@@ -0,0 +1,220 @@
...
+            .distinct()
+
+        if (missingParentIds.isNotEmpty() && variations.isNotEmpty()) {
+            val localSiteId = variations.first().localSiteId
+            val parentProducts = productsDao.getProductsByIds(localSiteId, missingParentIds)
+            parentProducts.forEach { productNamesMap[it.remoteId.value] = it.name }
+        }
+
+        val variationFtsEntities = variations.map { variation ->
+            val parentName = productNamesMap[variation.remoteProductId.value] ?: ""
```
malinajirka: ❓ Seems like the parent should always be present, right? Have you considered skipping the variation and logging it instead of setting parent name to ""?

malinajirka (approved):
LGTM, I've left a couple questions.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
kidinov: Good catch — fixed. Now also calling `deleteVariationsByParentProductIds` to remove orphaned variation FTS entries when their parent product is deleted.

5fbcdd7d03364a1852db215c7df20b2315cec5d5

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
kidinov: Fixed. Now skipping the variation and logging a warning instead of using an empty parent name. Applied to both `buildFtsEntities` and `buildFtsEntitiesForIncremental`.

9c99ed64c64ec243a8d7dca7badae60331d88d92

PR #15276, opened 2026-01-30
[Woo POS][Refunds] Remove X button during refund processing

samiuelson:
WOOMOB-1952

### Description

This PR implements conditional display of the close button during refund processing step. Previously the button didn't do anything, and was misleading. It also removes custom X button and uses POS default one from dialog wrapper function.

### Test Steps

1. Try creating refund in POS
2. Verify that during refund creation the refund flow dialog doesn't display X buton

### Images/gif

—


kidinov (approved):
LGTM!

PR #15279, opened 2026-02-02
Allow displaying PN for non WP users

irfano:
### Description

In the Woo Push Notifications project, we allow users who are using self-hosted sites without a WP.com login to receive push notifications. Previously, we had a check that prevented displaying notifications if users were not logged in. Now, for Woo-driven notifications, we skip this check and instead verify that the notification's site is still visible before displaying it.

### Test Steps

Testing isn’t currently possible because the Woo push server isn’t active yet.

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
@@ -25,6 +25,7 @@ import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.NotificationActionBuilder
 import org.wordpress.android.fluxc.model.notification.NotificationModel
 import org.wordpress.android.fluxc.store.AccountStore
+import org.wordpress.android.fluxc.store.SiteStore
 import org.wordpress.android.fluxc.store.WpComPushNotificationStore.FetchNotificationPayload
 import javax.inject.Inject
 import javax.inject.Singleton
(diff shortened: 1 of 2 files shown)


JorgeMucientes (approved):
LGTM! :shipit:

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -101,6 +98,13 @@ class NotificationMessageHandler @Inject constructor(
                 wooLog.d(NOTIFICATIONS, "Skipping WPCOM notification, already registered with Woo Core")
                 return
             }
+            if (siteStore.visibleSites.none { it.siteId == notification.remoteSiteId }) {
+                wooLog.w(NOTIFICATIONS, "Skipping notification, site ${notification.remoteSiteId} is not visible")
+                return
+            }
```
hichamboushaba, line 104: @irfano While investigating a negative review about push notifications (p1772994602271449/1772952726.649509-slack-C013AAPA4G0), I noticed this change.

Was this meant to be about the "edit sites" screen in the site picker? If so, I think we are checking the wrong field. The `visibleSites` property of `SiteStore` depends on the `visible` flag of the WPCom API, which appears to be deprecated since there’s no way to hide sites on Calypso (198778-ghe-Automattic/wpcom). We should use `GetWooVisibleSites` instead.

(This is not related to the PN issue, as it's applied only when the feature flag is enabled, but just wanted to flag it so that we can fix it later).

irfano: Great catch, @hichamboushaba! I’ve opened a new PR to fix this: https://github.com/woocommerce/woocommerce-android/pull/15530/

PR #15281, opened 2026-02-02
[WOOMOB-2093] Show low battery warning in floating toolbar

kidinov:
Closes: WOOMOB-2093

### Description

Display a battery warning icon in the floating toolbar when the connected card reader has low or critical battery. This provides a non-intrusive way to inform merchants that the reader needs charging without interrupting the payment flow.

- Added `BatteryState` enum (NOMINAL, LOW, CRITICAL) to toolbar state
- Combined `readerStatus` and `batteryStatus` flows in ViewModel to map battery state
- Display orange battery alert icon for LOW battery
- Display red empty battery icon for CRITICAL battery
- No icon shown for NOMINAL battery

@joshheald @malinajirka 👋

I’m not sure about this change. I appreciate that it’s not intrusive and the change is minor code-wise. Also, I don’t like showing Toast/Snack bars during the connection flow - that usually feels like an afterthought and especially doesn't look good on tablets. On the other hand, this icon can be missed. iOS doesn't have anything, from what I can tell regarding notifying about battery levels. 

Maybe displaying the battery level in % or with icons all the time to make it more clear where to look?

What are your thoughts?

### Test Steps

1. Connect a card reader with low battery
2. Verify the orange battery icon appears next to "Reader connected" text in the floating toolbar
3. Connect a card reader with critical battery
4. Verify the red battery icon appears
5. Connect a card reader with normal battery
6. Verify no battery icon is shown

### Images/gif

image
image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/toolbar/WooPosHomeFloatingToolbarViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/toolbar/WooPosHomeFloatingToolbarViewModel.kt
@@ -8,6 +8,8 @@ import com.woocommerce.android.cardreader.connection.CardReaderStatus.Connected
 import com.woocommerce.android.cardreader.connection.CardReaderStatus.Connecting
 import com.woocommerce.android.cardreader.connection.CardReaderStatus.NotConnected
 import com.woocommerce.android.cardreader.connection.CardReaderStatus.Reconnecting
+import com.woocommerce.android.cardreader.connection.event.BatteryStatus
+import com.woocommerce.android.cardreader.connection.event.CardReaderBatteryStatus
 import com.woocommerce.android.ciab.CIABSiteGateKeeper
 import com.woocommerce.android.ui.woopos.cardreader.WooPosCardReaderFacade
 import com.woocommerce.android.ui.woopos.cardreader.connection.WooPosCardReaderConnectionController
@@ -27,6 +29,7 @@ import com.woocommerce.android.viewmodel.ResourceProvider
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.launch
 import javax.inject.Inject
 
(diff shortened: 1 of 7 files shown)


malinajirka:
> I’m not sure about this change. I appreciate that it’s not intrusive and the change is minor code-wise. Also, I don’t like showing Toast/Snack bars during the connection flow - that usually feels like an afterthought and especially doesn't look good on tablets. On the other hand, this icon can be missed. iOS doesn't have anything, from what I can tell regarding notifying about battery levels.

I kind of like it to be honest. I think displaying something that is there all the time (%) is more likely to be missed. Users who will use the POS regularly will likely notice anything that is unusual -> so they'll imo notice the red battery.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/toolbar/WooPosHomeFloatingToolbarViewModel.kt
```diff
@@ -52,10 +55,13 @@ class WooPosHomeFloatingToolbarViewModel @Inject constructor(
 
     init {
         viewModelScope.launch {
-            cardReaderFacade.readerStatus.collect {
-                _state.value = _state.value.copy(
-                    cardReaderStatus = mapCardReaderStatusToUiState(it)
-                )
+            combine(
```
malinajirka, line 61: ❓ I'm wondering if combine is the appropriate operator here. Battery status is dependent on reader status => battery is always unknown unless a reader is connected. Have you considered using flatMapLatest here? Something like:

```
  cardReaderFacade.readerStatus.flatMapLatest { readerStatus ->
      when (readerStatus) {
          is Connected -> cardReaderFacade.batteryStatus.map {
              WooPosCardReaderStatus.Connected(batteryState = mapBatteryState(it))
          }
          is NotConnected, Connecting -> flowOf(WooPosCardReaderStatus.NotConnected)
          Reconnecting -> flowOf(WooPosCardReaderStatus.Reconnecting)
      }
  }
```

malinajirka (approved):
Thanks @kidinov!

One thing I noticed is that battery `low` is emitted when battery is below 50% => if that's true, then disregard my previous comment. I expected it to be shown only rarely, when batter is below 20% or so and critical below 10% or so. Since it's shown below 50%, I think users will get use to it and will likely miss the critical batter warning. I'd suggest displaying the battery icon only for the critical warning => 20% should still last for roughly 150 card transactions.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15283, opened 2026-02-02
Register device after logging in with site credentials

irfano:
Closes WOOMOB-2100

### Description

Previously, `registerDevice()` was only invoked when the user logged in via WordPress.com. For users who logged in with site credentials, `registerDevice()` was never called.

Now that we support push notifications for site credentials login (via the WOO self-serve PN system), we need to trigger device registration for this flow as well.

This PR adds `registerDevice()` call to the site credentials login flow.

### Test Steps

1. To enable `WOO_PUSH_NOTIFICATIONS_SYSTEM` feature flag, apply this patch: 
[Enable_WOO_PUSH_NOTIFICATIONS_SYSTEM_feature_flag_for_debug_builds.patch](https://github.com/user-attachments/files/25020347/Enable_WOO_PUSH_NOTIFICATIONS_SYSTEM_feature_flag_for_debug_builds.patch)
2. Log in to a site with support of new PN with site credentials.
3. Expected logs:
```
Registering FCM token in Woo Core instance: cRUfXyhgQnuOZAXUT9UrRC:APA91bHz...
```

### Images/gif


JorgeMucientes (approved):
Nicely done. Changes look good and work as expected. :shipit:

PR #15286, opened 2026-02-03
[WOOMOB-2106] Fix double %% sign in battery low update dialog

kidinov:
WOOMOB-2106

### Description

The battery low update dialog displayed "50%%" instead of "50%". This happened because the string resource used `%%` (the escape for `String.format()`), but Compose's `stringResource()` without format arguments doesn't go through `String.format()`, so `%%` was displayed literally.

Fixed by using a single `%` in the no-args string variant. The `_with_level` variant correctly keeps `%%` since it's called with format arguments.

### Test Steps

1. Connect a card reader with low battery
2. Trigger a software update
3. Verify the dialog shows "50%" (not "50%%")

image


samiuelson (approved):
LGTM!

PR #15288, opened 2026-02-04
[WOOMOB-2061] Add analytics tracking for file-based catalog sync

malinajirka:
Fixes WOOMOB-2061

### Description
Adds analytics tracking for the file-based local catalog sync to help diagnose "stuck" sync issues. When sync times out or fails, we now track additional properties to understand where the sync got stuck:

- **generation_duration_ms**: How long the catalog file generation took on the server
- **poll_attempts**: Total number of poll attempts (persisted across app restarts)
- **last_generation_state**: The last state received from the server when timeout occurs

Poll attempts are accumulated across app restarts so we can see the total attempts when users repeatedly open/close the app during long syncs.

Based on pdfdoF-8GW-p2.

### Test Steps
1. Enable file-based local catalog sync feature flag
2. Trigger a catalog sync (enter POS mode or trigger it in catalog settings)
3. Verify sync completion events include the new properties when using file-based sync
4. Force a timeout scenario and verify failure events include `last_generation_state` and `poll_attempts` (easiest way to force it is to update the code, however, I tested it by scheduling a job and before it started I cancelled it in `WPAdmin -> WooCommerce -> Status -> Scheduled Actions -> Pending -> Search for `feed` -> cancel catalog generation action)

### Images/gif
N/A - Analytics changes only


malinajirka:
📓 This can wait for after the Bookings work.

malinajirka:
@samiuelson This is now ready for review, thanks!

samiuelson (approved):
LGTM!

PR #15290, opened 2026-02-04
[WOOMOB-2067] Wire FTS search to product search UI

kidinov:
Closes WOOMOB-2067

### Description

Wires the FTS (Full Text Search) results — including product variations — to the existing POS product search UI. When the `POS_PRODUCTS_FTS` feature flag is enabled:

- Search results now include variations alongside simple and variable products
- Variation search results display parent product name as the title, variation attributes as subtitle, and price below
- Clicking a variation in search results adds it directly to the cart
- Search placeholder updates to "Search products and variations"

Key implementation details:
- Converted `WooPosProductType` from enum to sealed class so `Variation` can carry `parentProductName` with compile-time safety
- Introduced `VariationSearchResult` view state type to cleanly separate search result variations (with parent name) from variation list variations (without)
- Parent product name flows from FTS entity through the mapper to the UI layer
- `mapProductType()` now crashes on "variation" type string — variations must go through `fromVariationEntity()` which provides the parent product name
- `parentId` uses `requireNotNull` instead of a `?: 0L` fallback since variations always have a parent

### Test Steps

1. Enable `POS_PRODUCTS_FTS` feature flag
2. Open POS and tap the search bar
3. Verify placeholder reads "Search products and variations"
4. Search for a term that matches a variation (e.g. a color or size attribute)
5. Verify variation results show: parent product name (title), attributes (subtitle), price
6. Tap a variation result → verify it's added to cart
7. Disable feature flag → verify search placeholder reverts to "Search products" and no variations appear

### Images/gif

https://github.com/user-attachments/assets/2fbeaad3-a4ae-409e-a4ab-ed934b7f1a5e


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/283">24.3</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModel.kt
```diff
@@ -68,17 +68,26 @@ data class WooPosProductModel(
...
+    sealed class WooPosProductType(val value: String) : Parcelable {
+        @Parcelize data object Simple : WooPosProductType("simple")
+
+        @Parcelize data object Variable : WooPosProductType("variable")
+
+        @Parcelize data object Grouped : WooPosProductType("grouped")
+
+        @Parcelize data object External : WooPosProductType("external")
+
+        @Parcelize data class Variation(val parentProductName: String) : WooPosProductType("variation")
```
malinajirka: 📓 This feels a bit off to me - name of parent product doesn't feel related to ProductType. It makes all `when` and comparisons significantly more complex, for example `Variation("foo") != Variation("bar")`.

I'd consider reverting this and passing the parent info through a custom param.

Alternatively, we could add it as nullable field to the product model, but I assume you wanted to avoid that to prevent creation of nonexisting state. One could argue that changing model, let alone enum, to pass UI data is not any better than allowing creation of a non-existing state though. Wdyt?

malinajirka (approved):
Works like a charm, great job!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/models/WooPosProductModel.kt
kidinov: Good point, fixed — reverted `Variation` back to `data object` and moved `parentProductName` to a nullable field on `WooPosProductModel` itself. Also removed the `error()` in `mapProductType` for `"variation"` since it no longer needs special handling.

PR #15277, opened 2026-01-30
Include origin email in Shipping Label Purchase request

irfano:
Closes WOOMOB-2088

### Description

This PR adds origin email to the shipping label purchase request and implements email validation for the address form.

### Test Steps

1. Log in to a site that is eligible to purchase shipping labels.
2. Go to Orders.
3. Open an order with shippable items.
4. Tap **Create shipping label**.
5. Tap **Shipment details**.
6. Tap ⋯ button next to **Ship from**.
7. Tap the ✎ button on a store address item.
8. Make sure the Email field is required.
9. Enter an invalid email (e.g., test@ or invalid)
10. Enter a valid email (e.g., test@ or invalid)
11. Verify the address.
12. Use a network monitor to debug the endpoint request (I used **Android Studio Network Inspector**).
13. Navigate back, fill in the other required fields, and purchase the label.
14. Verify `"email":"test@email.com"` is included in the request payload for the `wcshipping/v1/label/purchase/` endpoint.

### Images/gif

Screenshot_20260130_225656

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15291, opened 2026-02-04
Require valid phone number in shipping address for shipping labels

JorgeMucientes:
Closes WOOMOB-2130

### Description

This PR adds phone number validation for shipping label purchases. The Woo Shipping server now enforces valid phone numbers on the destination address when purchasing shipping labels (especially for UPS).

### Test Steps

**Setup:**
1. Have a store with WooCommerce Shipping plugin set up
2. Have an order with a destination address that has an empty or invalid phone number (no digits)

**Test Edit Address Phone Validation:**
1. Open an eligible order and tap "Create Shipping Label"
2. Tap on the destination address to edit it
3. Clear the phone number field
4. Verify an error message appears: "This field is required."
5. Verify the "Validate & Save" button is disabled
6. Enter text with no digits (e.g., "abc-def")
7. Verify an error message appears: "Please enter a valid phone number"
8. Enter a valid phone number (e.g., "+1 555-123-4567")
9. Verify the error clears and the "Validate & Save" button becomes enabled

https://github.com/user-attachments/assets/ed9e36ee-3f0d-46b9-94a4-5fc712a9cd38

**Test Notice Banner:**
1. With an address missing a phone number, open the shipping label creation flow
2. Verify a warning banner appears: "A valid phone number is required in the shipping address"
3. Tap the banner to dismiss it

https://github.com/user-attachments/assets/d0e58f49-fe09-46b4-9c3d-e768145c1313

**Test Pre-Purchase Validation:**
1. With an invalid phone number in the destination address, select a package and shipping rate
2. Tap "Purchase Label"
3. Verify a snackbar appears with the error message and an "Edit" button
4. Tap "Edit" to open the address editing screen
5. Add a valid phone number and save
6. Verify you can now purchase the label successfully

https://github.com/user-attachments/assets/ef5b27e2-0604-4536-a01a-1235097d2c3f

**Test API Error Handling:**
1. If the server returns a phone-related error, verify the app shows a snackbar with "Edit" action instead of the generic retry error


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/280">24.1</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/address/AddressValidationHelper.kt
```diff
@@ -57,6 +57,10 @@ class AddressValidationHelper @Inject constructor(
         }
     }
 
+    fun isPhoneValidForShippingLabel(phone: String): Boolean {
+        return phone.isNotBlank() && phone.contains(Regex("\\d"))
```
JorgeMucientes, line 61: Using the same generic regex here as we use for phone number in customs form

irfano:
@JorgeMucientes , I actually wouldn’t add a new notice banner specifically for the phone error because we’re already displaying a general “missing destination” error banner. Phone isn’t the only required field in the destination address form, and we don’t show field-specific banners for the other required fields either.
Similarly, when we added the email requirement for the origin address, I didn’t add a dedicated origin banner either (#15277).
What do you think about keeping the current behavior (general banner only) instead of adding a phone-specific one? It would also keep the PR simpler.

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 
 24.1
 -----
+- [Internal] Providing a phone number in shipping address  [https://github.com/woocommerce/woocommerce-android/pull/15208]
```
irfano: I think the PR title is better for this description. We actually "require" instead of "provide" in this PR. So we can write "Require valid phone number in shipping address for shipping labels" here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
```diff
@@ -986,6 +986,14 @@ class WooShippingLabelCreationViewModel @Inject constructor(
 
         if (selectedPackage == null || selectedAddress == null || shippingRate == null || weight == null) return
 
+        if (!addressValidationHelper.isPhoneValidForShippingLabel(selectedAddress.shipTo.address.phone)) {
+            snackbarData = ShippingLabelsSnackbarData(
+                message = R.string.woo_shipping_labels_purchase_phone_error,
+                actionLabel = R.string.edit,
+            ) { onEditDestinationAddress(selectedAddress.shipTo) }
+            return
+        }
```
irfano, line 995: Image

For consistency, we could use this error state when the phone is invalid.

irfano (approved):
I’m approving since this PR works as expected.
That said, I want to make sure the different error handling for the phone number is intentional, even though it breaks consistency with the other required fields. Maybe I’m missing something and you chose this behavior on purpose.
I’d actually prefer removing `shouldRequirePhone()` and always requiring the phone number (including the US). I haven’t tested it, but I think this would also prevent purchasing a label without a phone number.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/WooShippingLabelCreationViewModel.kt
JorgeMucientes: Thanks for the feedback @irfano, very valid points. 

Email and phone can't be treated equally because email is a mandatory field when checking out in the frontend whereas phone it's optional. This means you can land on the app with an order that has a valid shipping address (even if unverified) with email but lacking phone number. 
The user would be able to attempt purchasing a label, as we don't enforce validating the shipping address  to make the purchase: 

https://github.com/user-attachments/assets/84f06c7e-5d95-4227-ac19-788194224aae

I think it's key to provide proper error handling pointing to the real reason of the failed purchase like in the screen recording above. Which is what this particular code is aiming to. 

As per the red notice snackbar. I'm ok with not being specific about the missing phone number. I've updated it and change the text a little bit to `Destination address missing or invalid` as you can see in the above screen recording. Removed here: [928a715](https://github.com/woocommerce/woocommerce-android/pull/15291/commits/928a7155264a7cb17e52df7d5b9a123080ce6923)

RELEASE-NOTES.txt
JorgeMucientes: Good catch, applied

PR #15200, opened 2026-01-15
[WOOMOB-1935] Replace local feature flag with remote flag for POS tablet promo banner

kidinov:
WOOMOB-1935

### Description

Replace the local `CLIENT_SIDE_POS_BANNER` feature flag with the remote `woo_pos_tablet_promo_banner` feature flag for the POS tablet promotion banner.

Changes:
- Add new `WOO_POS_TABLET_PROMO_BANNER` remote feature flag
- Update `ClientSidePosBanner` to use the remote feature flag
- Remove the local feature flag and targeting percentage logic
- Add banner assets (background image URL)
- Simplify badge icon handling in `JitmViewModel` to match iOS implementation: when no badge image URL is provided from the backend, return `null` instead of falling back to a local "NEW" label

### Test Steps

1. Enable the `woo_pos_tablet_promo_banner` remote feature flag (changing the code on the client)
2. Open the app on a phone device (not tablet)
3. Navigate to My Store
4. Verify the POS promotion banner appears for eligible stores (US, GB)
5. Verify the banner displays correctly without a badge label when none is provided from the backend

### Images/gif

N/A - No UI changes

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15299, opened 2026-02-05
Improve remote and local feature flag handling

irfano:
### Description

Fixes WOOMOB-2129

This PR introduces an injectable `FeatureFlagRepository` and routes runtime feature flag reads through it instead of static `FeatureFlag.isEnabled()` calls and flag-specific wrappers. The goal is to make feature flag resolution testable and consistent while keeping reads synchronous at call sites.

#### Feature Flag Resolution Model
Each flag can now be resolved from three inputs:
- `Override`: developer-set override from the Feature Flags screen
- `Local`: app-defined readiness gate in `FeatureFlag`
- `Remote`: backend value fetched from WPCom and cached locally

```mermaid
flowchart TD
    A["Override set?"] -->|Yes| B["Use Override"]
    A -->|No| C["Local flag enabled?"]
    C -->|No| D["Feature Disabled"]
    C -->|Yes| E["Remote flag exists?"]
    E -->|No| F["Feature Enabled"]
    E -->|Yes, true| F
    E -->|Yes, false| D
```

Equivalent formula:
```
Effective Value = Override ?: (Local && (Remote ?: true))
```

This means:
- `Override` still has the highest priority.
- `Local` acts as a readiness gate.
- `Remote` only controls rollout for features that are already locally enabled.
- If there is no remote value, the local value is used.

#### Developer Overrides UI
The Developer Options → Feature Flags screen now:
- loads flag state from `FeatureFlagRepository`
- exposes tri-state overrides (`Default`, `Disabled`, `Enabled`)
- shows whether the non-override source is currently `Local` or `Remote`

#### Important implementation decisions
- `isEnabled()` remains synchronous.
Instead of making feature flag reads `suspend` everywhere, remote values are observed into an in-memory snapshot and runtime reads stay synchronous.
- Remote flags are still persisted in Room, but `FeatureFlagRepository` is now the single runtime access point.
This keeps call sites simple and makes tests easier to write.
- I intentionally did not introduce separate APIs such as `isEnabledLocally()` / `isEnabledRemotely()`.
That would leak flag source details into consumers and make future migrations easier to miss.

### Review focus
The most important places to review are:

#### Core feature flag model
This is the new resolution model (`Override ?: (Local && (Remote ?: true))`) and the persistence path behind it.
- `FeatureFlag`
- `FeatureFlagRepository`
- `AppPrefs`
- `WPComRemoteFeatureFlagRepository`
- `FeatureFlagsStore` / `FeatureFlagConfigDao`

#### Developer overrides UI
This is where the new model is exposed to developers and where the tri-state override behavior can be validated end to end.
- `DevFeatureFlagsScreen`
- `DevFeatureFlagsViewModel`

#### Consumers where timing matters
These are the places where the switch to the in-memory snapshot model matters most, because they either observe flag changes or wait for the initial cached snapshot before evaluating.

Observation was not a hard requirement, and the previous synchronous access would still have worked, since flags are not refreshed continuously while the app is running anyway. I still added observation in the areas listed below to better cover possible edge cases and leave room for future improvements.

- `ShouldShowEnablePushNotificationsUi`: This path already returns a `Flow`, so it now observes the flag reactively instead of taking a one-time snapshot that could stay stale if the cache hydrates later.
- `JetpackBenefitsViewModel`: The visibility of the push-notifications benefit is now updated from observed flag state instead of being fixed at ViewModel creation time.
- `ObserveBookingsVisibility`: This now combines selected-site observation with reactive flag observation, so the visibility decision can recover correctly after cache hydration.
- `OrderDetailViewModel` / `OrderDetailFragment`: The flag is now resolved in the ViewModel and carried through view state instead of being checked directly in the Fragment.
- `AddressViewModel` / `OrderCreateEditCustomerAddFragment`
- `TaxRateSelectorViewModel` / `TaxRateSelectorScreen`: The bottom bar visibility is now driven by ViewModel state backed by observed flag updates.
- `WooApplicationPasswordsConfiguration`: This is a one-shot gate, so it now waits for the initial cached remote flag snapshot before deciding whether App Passwords can be used.
- `WooPosTabShouldBeVisible`: This is another one-shot startup decision, so it now waits for the initial cached remote flag snapshot before evaluating visibility.
- `ClientSidePosBanner`: This now uses the real `woo_pos_tablet_promo_banner` remote key together with a local readiness gate through `FeatureFlagRepository`.

### Test Steps

I tested some local and remote flags (`woo_pos_local_catalog_m1`, `bookings_mvp`, `woo_app_passwords_for_jetpack_sites`). You can also choose some features, disable them from "Developer Options → Feature Flags", restart the app, and verify that the feature is disabled as expected.

### Images/gif

Screenshot_20260205_224054


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>DevFeatureFlagsViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/config/WPComRemoteFeatureFlagRepository.kt
```diff
@@ -41,7 +41,7 @@ class WPComRemoteFeatureFlagRepository @Inject constructor(
         }
     }
 
-    suspend fun isRemoteFeatureFlagEnabled(key: String): Boolean = withContext(Dispatchers.IO) {
-        featureFlagsStore.getFeatureFlagsByKey(key).firstOrNull()?.value ?: false
+    suspend fun isRemoteFeatureFlagEnabled(key: String): Boolean? = withContext(Dispatchers.IO) {
+        featureFlagsStore.getFeatureFlagsByKey(key).firstOrNull()?.value
```
irfano: This function will now return null if the key can't be found in remote flags.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/ShouldShowEnablePushNotificationsUi.kt
```diff
@@ -20,19 +22,27 @@ import javax.inject.Inject
...
+    private val featureFlagRepository: FeatureFlagRepository
 ) {
-    operator fun invoke(): Flow<Boolean> {
-        if (!FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM_M2.isEnabled()) return flowOf(false)
-        return selectedSite.observe()
-            .flatMapLatest { site ->
+    operator fun invoke(): Flow<Boolean> = selectedSite.observe()
+        .flatMapLatest { site ->
+            flow {
+                if (!featureFlagRepository.isEnabled(FeatureFlag.WOO_PUSH_NOTIFICATIONS_SYSTEM_M2)) {
```
irfano: I made some changes in this file to be able to suspend `isEnabled` function in the flow.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditCustomerAddFragment.kt
```diff
@@ -279,7 +278,7 @@ class OrderCreateEditCustomerAddFragment :
         menu.clear()
 
         if (editingOfAddedCustomer.editingOfAddedCustomer &&
-            FeatureFlag.BETTER_CUSTOMER_SEARCH_M2.isEnabled()
+            addressViewModel.viewStateData.liveData.value?.isBetterCustomerSearchEnabled == true
```
irfano, line 281: I moved this flag to the view model

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailFragment.kt
```diff
@@ -593,9 +592,9 @@ class OrderDetailFragment :
         )
     }
 
-    private fun showInstallWcShippingBanner(isVisible: Boolean) {
+    private fun showInstallWcShippingBanner(isVisible: Boolean, isWcShippingBannerEnabled: Boolean) {
         val banner = binding.orderDetailInstallWcShippingBanner
-        banner.isVisible = isVisible && FeatureFlag.WC_SHIPPING_BANNER.isEnabled()
+        banner.isVisible = isVisible && isWcShippingBannerEnabled
```
irfano, line 607: I moved this flag to the view model.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/developer/DevFeatureFlagsScreen.kt
```diff
@@ -54,20 +80,6 @@ fun DevFeatureFlagsScreen(
...
-    var hasChanges by remember { mutableStateOf(false) }
-
-    fun updateHasChanges(flag: FeatureFlag, newValue: Boolean) {
-        initialValues[flag]?.let { _ ->
-            hasChanges = allFeatureFlags.any { featureFlag ->
-                val currentValue = if (featureFlag == flag) newValue else featureFlag.isEnabled()
-                currentValue != initialValues[featureFlag]
-            }
-        }
-    }
```
irfano, line 69: Since the logic in this screen became larger, I moved this logic to a new ViewModel.

WooCommerce/src/main/kotlin/com/woocommerce/android/WooWellSqlConfig.kt
```diff
@@ -18,7 +18,7 @@ class WooWellSqlConfig(context: Context) : WellSqlConfig(context) {
      * build with a DB downgrade was released, resulting in a lot of crashes.
      */
     override fun onDowngrade(db: SQLiteDatabase?, helper: WellTableManager?, oldVersion: Int, newVersion: Int) {
-        if (FeatureFlag.DB_DOWNGRADE.isEnabled(context)) {
+        if (PackageUtils.isDebugBuild() && context != null && PackageUtils.isBetaBuild(context)) {
```
irfano: I removed `DB_DOWNGRADE` flag since it was not actually a feature flag. We'll never switch this boolean here. So I moved the same logic from the old `FeatureFlag` class to here.
Also since this class runs before Hilt, we can't inject `FeatureFlagRepository` here.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/main/MainPresenterTest.kt
```diff
@@ -116,13 +115,11 @@ class MainPresenterTest : BaseUnitTest() {
 
     @Test
     fun `Handles database downgrade correctly`() = testBlocking {
-        if (FeatureFlag.DB_DOWNGRADE.isEnabled()) {
```
irfano, line 119: `FeatureFlag.DB_DOWNGRADE.isEnabled()` was redundant here because this logic was always `true` in unit tests.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/mobile/FeatureFlagsStore.kt
```diff
@@ -47,32 +48,12 @@ class FeatureFlagsStore @Inject constructor(
...
-        return featureFlagConfigDao.getFeatureFlagList()
-    }
-
     // This returns a list because there can be multiple values for a single key.
     // It will be the client's responsibility to decide which value to use.
     fun getFeatureFlagsByKey(key: String): List<FeatureFlag> {
         return featureFlagConfigDao.getFeatureFlag(key)
     }
 
-    fun insertFeatureFlagValue(key: String, value: Boolean) {
```
JorgeMucientes, line 60: Were these functions unused?

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -1,56 +1,35 @@
...
+    val default: Boolean = PackageUtils.isDebugBuild()
+) {
+    WC_SHIPPING_BANNER("wc_shipping_banner"),
+    BETTER_CUSTOMER_SEARCH_M2("better_customer_search_m2"),
+    ORDER_CREATION_AUTO_TAX_RATE("order_creation_auto_tax_rate"),
+    BOOKINGS_MVP("bookings_mvp"),
+    POS_REFUNDS("pos_refunds"),
+    POS_BOOKINGS("pos_bookings"),
+    POS_PRODUCTS_FTS("pos_products_fts"),
+    WOO_POS_LOCAL_CATALOG_FILE_APPROACH("woo_pos_local_catalog_file_approach"),
```
JorgeMucientes: Heads up that this one is also false right now in trunk:
```kotlin
            WOO_PUSH_NOTIFICATIONS_SYSTEM,
            WOO_PUSH_NOTIFICATIONS_SYSTEM_M2,
            WOO_POS_LOCAL_CATALOG_FILE_APPROACH -> false
```

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -1,56 +1,35 @@
...
-
-            WOO_PUSH_NOTIFICATIONS_SYSTEM,
-            WOO_PUSH_NOTIFICATIONS_SYSTEM_M2,
-            WOO_POS_LOCAL_CATALOG_FILE_APPROACH -> false
-        }
-    }
+enum class FeatureFlag(
+    val key: String,
+    val default: Boolean = PackageUtils.isDebugBuild()
+) {
```
JorgeMucientes, line 16: I wonder if we should be specific here that the `key` value is the one that will be used to fetch the remote feature flag value (if there's any). 
Maybe rename it to `remoteFlagKey` ? WDYT

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.woocommerce.android.AppPrefs
+import com.woocommerce.android.config.WPComRemoteFeatureFlagRepository
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class FeatureFlagRepository @Inject constructor(
+    private val remoteFeatureFlagRepository: WPComRemoteFeatureFlagRepository
+) {
+    suspend fun isEnabled(flag: FeatureFlag) = getFlagState(flag).effectiveValue
```
JorgeMucientes: There's a tradeoff with this approach, where now checking for feature flag value is suspendable and will need to run in a coroutine even for those cases where a remote feature flag was never defined. 
I wonder if it could make sense to have something like 
```kotlin
fun isEnabledLocally(flag: FeatureFlag) = getLocalFlagState(flag).effectiveValue

fun getLocalFlagState(flag: FeatureFlag) = FeatureFlagState(
        flag = flag,
        defaultValue = flag.default,
        remoteValue = null,
        overrideValue = getOverrideValue(flag)
    )
```
This would simplify some checks and avoid using `runBlocking` and DB access in places where the FF value needs to be checked synchronously.

JorgeMucientes (commented):
Nice job @irfano , sorry it took me longer than I expected to review this, but its pretty big change and potentially dangerous. In fact, I'd recommend if someone from Peacock could review this too, to double check we are not accidentally enabling any POS functionality. 
I also let a couple of comments let me know you thoughts :)

libs/fluxc/src/main/java/org/wordpress/android/fluxc/store/mobile/FeatureFlagsStore.kt
irfano: Yes, these are unused in `trunk`.
image

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
irfano: Good eyes! I’ve updated it and synced with trunk.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
irfano: Makes sense. Done: 83caedf565b7d486764be575b0aec5124bc1fe50

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
irfano: I was aware of this trade-off, and thanks for starting this discussion.
Your suggestion was my first attempt to address the issue, but it introduced another trade-off. I started questioning why the consumer side should have to decide between `isEnabledLocally` and `isEnabled`. With that approach, the consumer needs to know whether the flag has a remote value or not. And when we later add a remote value for a flag, we might forget to update places that still use `isEnabledLocally`.

There’s also an alternative: keeping flags in memory. It’s doable, but it adds another layer, and another potential failure point, especially around syncing cached flags.

I’m not sure what the best approach is. WDYT?

irfano:
> In fact, I'd recommend if someone from Peacock could review this too, to double check we are not accidentally enabling any POS functionality.

I agree!

@malinajirka , @samiuelson , @kidinov , this is a large PR that affects the feature flag structure. Could one of you please review it and give approval?

malinajirka (commented):
Thanks @irfano! 

This is quite a big change and feels potentially risky. I'm wondering if you considered deprecating the pre-existing approach, instead of replacing it. That way, we wouldn't need to re-test all the existing FF and would simply start adding the new ones using the new approach. It'd feel less risky - a proper review for a PR with 1500+ lines is quite a challenge.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/editing/address/AddressViewModel.kt
```diff
@@ -45,6 +48,20 @@ class AddressViewModel @Inject constructor(
     val viewStateData = LiveDataDelegate(savedState, ViewState())
     private var viewState by viewStateData
 
+    init {
+        initDefaultViewState()
+    }
+
+    private fun initDefaultViewState() {
```
malinajirka: ⚠️ I haven't looked deeply into this, but feels like a potentially dangerous change.

The coroutine might take a while to complete - when it completes, we override `viewState` ignoring whatever state it is in. We might want to consider initializing the ViewState to `ViewState()` outside of the coroutine, and then using `copy` inside the coroutine.

Something like

    private fun initDefaultViewState() {
        viewState = ViewState()
        launch {
            viewState = viewState.copy(
                isBetterCustomerSearchEnabled = featureFlagRepository.isEnabled(
                    FeatureFlag.BETTER_CUSTOMER_SEARCH_M2
                )
            )
        }
    }

Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
malinajirka: Considering vast majority of our flags are local only and handling the issue Jorge raised isn't trivial in many cases -> I personally also vote for Jorge's suggestion.

malinajirka (commented):
Thanks for working on this @irfano! It's a nice change.

Overall, it looks good to me - the only thing I'm unsure about is the issue Jorge raised regarding all checks being suspendable.

One question regarding the evaluation logic - when remote FF is disabled and locally the FF is enabled => the feature will be disabled. This is a change to the current logic, where local FF always took a precedent, right? I guess it makes sense, I'm just wondering, whether it could backfire during development. Developers will be forced during PR reviews to go to Dev Settings and override the value, right? Likely not a big deal, but feels like a potential friction point. Wdyt?

malinajirka:
> One question regarding the evaluation logic - when remote FF is disabled and locally the FF is enabled => the feature will be disabled. This is a change to the current logic, where local FF always took a precedent, right? I guess it makes sense, I'm just wondering, whether it could backfire during development. Developers will be forced during PR reviews to go to Dev Settings and override the value, right? Likely not a big deal, but feels like a potential friction point. Wdyt?

I realized this has also a side-effect that a feature developer can't force teammates to start using their WIP or finished pre-release work - e.g. as an Alpha test. I personally utilize this quite often, having said that, in 99% of cases I don't have a remote FF initialized, so I guess it might be fine.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsLocalCatalogSupported.kt
```diff
@@ -2,20 +2,19 @@ package com.woocommerce.android.ui.woopos.localcatalog
...
+import com.woocommerce.android.util.FeatureFlagRepository
 import com.woocommerce.android.util.FetchActiveWCPluginVersion
 import com.woocommerce.android.util.GetWooCorePluginCachedVersion
 import org.wordpress.android.fluxc.model.LocalOrRemoteId
 import javax.inject.Inject
 
 class WooPosIsLocalCatalogSupported @Inject constructor(
-    private val wooPosLocalCatalogM1Enabled: WooPosLocalCatalogM1Enabled,
-    private val fileApproachEnabled: WooPosLocalCatalogFileApproachEnabled,
+    private val featureFlagRepository: FeatureFlagRepository,
```
malinajirka, line 19: ⚠️ 
 Heads up — looks like this accidentally drops the AppPrefs.wooPosLocalCatalogEnabled check for local catalog.

On trunk, WooPosIsLocalCatalogSupported goes through WooPosLocalCatalogM1Enabled, which checks both the remote flag and the Beta Features user preference:

  suspend operator fun invoke(): Boolean =
      isRemoteFeatureFlagEnabled(REMOTE_WOO_POS_LOCAL_CATALOG_M1) && AppPrefs.wooPosLocalCatalogEnabled

  In this PR it now calls featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_LOCAL_CATALOG_M1) directly, skipping that user preference entirely. WooPosLocalCatalogM1Enabled still exists with the check, but it's marked @Suppress("unused") and nothing calls it anymore.

So anyone who turned off local catalog from Beta Features would get it re-enabled silently. Since we asked HEs to recommend disabling the feature in Beta Features in case users have some serious troubles with it, I'd suggest updating this code to the previous behavior. Thanks!

JorgeMucientes:
Just a remark about this point @malinajirka 

> One question regarding the evaluation logic - when remote FF is disabled and locally the FF is enabled => the feature will be disabled. This is a change to the current logic, where local FF always took a precedent, right? I guess it makes sense, I'm just wondering, whether it could backfire during development. Developers will be forced during PR reviews to go to Dev Settings and override the value, right? Likely not a big deal, but feels like a potential friction point. Wdyt?

While I understand your point, that friction will only happen when you have both a remote and a local FF. And as you mentioned this is uncommon. The new order of precedence where the remote FF value is prioritised over any other value is what will allow us to enable/disable a FF remotely (without any release intervention) if anything goes sideways, which is one of the main points of this effort.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
JorgeMucientes: > There’s also an alternative: keeping flags in memory. It’s doable, but it adds another layer, and another potential failure point, especially around syncing cached flags.

I think keeping flags in memory would likely be the best approach imho, as: 
- It removes the need of checking 
- Avoid the potential risk you mentioned above "And when we later add a remote value for a flag, we might forget to update places that still use isEnabledLocally." 

If the only tradeoff is the user running the app a few more hours with the outdated remote FF value until it eventually gets the fresh one after app restart, then I think that is acceptable. 

And if we don't want to move on with this solution I'd prefer to have `isEnabledLocally` function vs making all the feature flag checks async.

malinajirka: > If the only tradeoff is the user running the app a few more hours with the outdated remote FF value until it eventually gets the fresh one after app restart, then I think that is acceptable.

Apps on Android might theoretically run for a month. There is also POS mode, where it's quite common merchants leave the device running 24/7. So I'd personally be careful with not having any sort of cache expiration mechanism other than an app restart.

malinajirka: Having said that, we could have an in-memory cache or even persistent cache that syncs once in 24hours in the background. That could solve both problems, wdyt? With periodic jobs and workers, it's relatively straightforward to implement.

malinajirka:
> The new order of precedence where the remote FF value is prioritised over any other value is what will allow us to enable/disable a FF remotely (without any release intervention) if anything goes sideways, which is one of the main points of this effort.

I'm not sure I follow here, I agree we want the ability to remotely disable a feature. But we have had that for years and the evaluation order has been different. Combination of local flag with a remote flag is used for development purposes -> I introduce disabled remote flag, but for development I keep the local flag enabled so I can continue developing and testing the feature before it's ready for release.

When I'm ready for release, I remove the local feature flag and enable the remote FF. If something goes wrong, I disable the remote FF => re-introduce local feature flag until the feature is fixed, and then either re-enable the remote FF (in case fix is backend-side) or introduce a new remote FF (in case it's a client side fix).

If the remote FF always takes precedents, we lose the ability to enable the feature in trunk.

Can you elaborate on the benefits of the new approach?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsLocalCatalogSupported.kt
irfano: Great catch, fixed it: 6941a5d5fcd7f1779573a95a43ff40b9a492a2a2

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
irfano: I don’t think the in-memory approach introduces any additional staleness compared to what we already have today. Remote flags are currently fetched when the initially selected site is observed at app startup, and then again whenever the selected site changes ([SiteObserver](https://github.com/woocommerce/woocommerce-android/blob/76b46dbb0665a643bf6e2a8898e05c49296d90ba/WooCommerce/src/main/kotlin/com/woocommerce/android/SiteObserver.kt#L58)). I’d prefer not to change that refresh policy in this PR.

Also, in the current implementation, the suspend requirement comes from the synchronous DB read being wrapped in `withContext(Dispatchers.IO)` here: [WPComRemoteFeatureFlagRepository.isRemoteFeatureFlagEnabled](https://github.com/woocommerce/woocommerce-android/blob/76b46dbb0665a643bf6e2a8898e05c49296d90ba/WooCommerce/src/main/kotlin/com/woocommerce/android/config/WPComRemoteFeatureFlagRepository.kt#L44-L45). We can keep that.

An alternative I’d suggest is keeping an in-memory snapshot of the flags and observing the DB, so whenever the DB changes the in-memory values are updated automatically as well. In other words: async DB access for persistence, sync in-memory access for runtime reads. I pushed a commit with that approach already. I still want to clean up a couple of follow-up items, so no need for a full review just yet. 0e50860bcb4f62430ace2d37b175208d

WDYT?

irfano:
Thanks for taking the time to review this, @malinajirka. After the latest update, before I start addressing the remaining comments, I want to reopen one question: would you prefer me to split this into smaller PRs, given the valid feedback about the risk of the current scope?

I’m happy to continue either way, but I’d like to align on that first before I proceed further.
 
> a proper review for a PR with 1500+ lines is quite a challenge.

The total diff is actually 1,989 lines, but most of it is mechanical:
- 1,034 lines are formatting/restyling only
- 835 of those come from `WooPosOrdersViewModelTest`; for some reason I had to reformat that file to get `detektAll` passing locally
- 309 lines are test-only changes
- 13 lines are comment-only changes
- 633 lines are the changes that actually need review
633 lines is still a large diff, but I think it is more manageable than the raw total suggests.

> I'm wondering if you considered deprecating the pre-existing approach, instead of replacing it. That way, we wouldn't need to re-test all the existing FF and would simply start adding the new ones using the new approach.

I did consider that. My concern is that if we keep the old approach for existing flags and use the new one only for new flags, we would end up maintaining two paths in parallel unless I follow up with separate PRs to migrate the existing flags one by one. That would make each PR smaller, but I’m not sure it would reduce the overall testing effort.

I also agree with the feedback about introducing new suspend call sites. I changed that in the latest commit, and `isEnabled()` is sync again, which should make the current version safer.

I’m open to splitting this into smaller PRs if you think that is still the better direction. Otherwise, I’d prefer to keep moving with this PR.

@JorgeMucientes, I’d also appreciate your opinion here.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
malinajirka: That sounds like a solid approach, I like it.

malinajirka:
Thanks for the openness @irfano and for all your work on this PR! 

I believe that at this point, proceeding with this PR is the best approach. Having said that, in case we'd be at the beginning, I would have personally preferred `unless I follow up with separate PRs to migrate the existing flags one by one. That would make each PR smaller, but I’m not sure it would reduce the overall testing effort.`. It would have likely slightly increased the testing effort, but it would have IMO significantly reduced the risk that we miss something and the reviews would have required less mental capacity. With all that said, I agree that in the end this PR isn't as complex as it seemed at first sight as lots of the changes are tests or mechanical changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/applicationpasswords/WooApplicationPasswordsConfiguration.kt
```diff
@@ -2,34 +2,23 @@ package com.woocommerce.android.applicationpasswords
...
-import kotlinx.coroutines.async
 import org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords.ApplicationPasswordsConfiguration
 
 class WooApplicationPasswordsConfiguration @Inject constructor(
     private val appPrefs: AppPrefsWrapper,
-    private val isRemoteFeatureFlagEnabled: IsRemoteFeatureFlagEnabled,
-    @AppCoroutineScope private val appCoroutineScope: CoroutineScope
+    private val featureFlagRepository: FeatureFlagRepository,
 ) : ApplicationPasswordsConfiguration {
-    private val remoteFeatureFlagDeferred = appCoroutineScope.async(
```
irfano, line 21: I removed this because the caching concern originally addressed in #14693 is now handled centrally by `FeatureFlagRepository`, so we no longer need a separate per-class deferred cache here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/editing/address/AddressViewModel.kt
irfano: I simplified the changes in this class and added an observed update in `init`, so this should no longer be an issue.

irfano:
> I guess it makes sense, I'm just wondering, whether it could backfire during development. Developers will be forced during PR reviews to go to Dev Settings and override the value, right? Likely not a big deal, but feels like a potential friction point. Wdyt?

Good point, @malinajirka! I did think about whether we need a per-flag merge strategy here, but I ended up preferring the simpler model. Here is my attempt:

image

We have 3 options:
- **A.** `(Current)` Override > Remote > Default
- **B.** `(Ignore Remote in debug builds)` Override > Default
- **C.** Adds a `strategy` field like in the screenshot

I think **C** is too complex and not really necessary. We can always add it later if we feel it’s needed.
**B** addresses your concern, but it introduces another drawback: it adds slightly more logic and we wouldn’t be able to test remote flags in debug builds anymore.  Still, that may not be a big deal. Also, we should change three-state beta feature screen to the two state options.

I believe both **A** and **B** are acceptable, but given your concern I prefer **B**. Do you see any advantage of A over B? Wdyt?

---

Update: I changed the logic, please check my latest [comment](https://github.com/woocommerce/woocommerce-android/pull/15299/#issuecomment-4056464245).

irfano:
> Combination of local flag with a remote flag is used for development purposes -> I introduce disabled remote flag, but for development I keep the local flag enabled so I can continue developing and testing the feature before it's ready for release.

As far as I can tell, in `trunk` we do not combine a local and a remote flag for the same feature. The closest example I found is `ClientSidePosBanner`, but even there the two flags are different (`FeatureFlag.WOO_POS_CLIENT_SIDE_BANNER` and `RemoteFeatureFlag.WOO_POS_TABLET_PROMO_BANNER`), and the merge logic lives in the consumer. That is the main thing this PR is trying to address. 
image

We started this work because we wanted to support both local and remote control for the push notifications work, and the current `ClientSidePosBanner` pattern is not ideal for that.

With this PR, a feature is defined once in `FeatureFlag`. Each flag has a required default value and an optional remote value, and the resolution logic is centralized instead of duplicated across consumers.

> If the remote FF always takes precedents, we lose the ability to enable the feature in trunk.
> 
> Can you elaborate on the benefits of the new approach?

So for me, the main benefit is not just changing the precedence order. The main benefit is having a single flag model and a single evaluation path. For release builds, I still think the remote value should take precedence. But we can be more flexible in debug builds. I evaluated that option in my previous [comment](https://github.com/woocommerce/woocommerce-android/pull/15299#issuecomment-4049959033).

---

Another way to think about it is this: when adding a feature flag, we no longer need to think in terms of “local” vs “remote”. We just add a flag to `FeatureFlag` and use it everywhere that needs it.

If, at some point, we want that feature to be remotely controlled, we can add the corresponding backend key later, because every feature flag now has a `remoteFlagKey`.

This is also very similar to how modern tools such as Firebase Remote Config work:
> When using Remote Config, you create in-app default values that control the behavior and appearance of your application. Then, you can later use the Firebase console or the Remote Config backend APIs to override in-app default values for all Remote Config API consumers - https://firebase.google.com/docs/remote-config

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jitm/clientside/ClientSidePosBanner.kt
```diff
@@ -21,17 +20,17 @@ class ClientSidePosBanner @Inject constructor(
...
 ) {
     val messagePath: String = JitmMessagePathsProvider.MY_STORE
     val bannerId: String = BANNER_ID
 
     @Suppress("ReturnCount")
     suspend fun shouldShow(): Boolean {
-        if (!FeatureFlag.WOO_POS_CLIENT_SIDE_BANNER.isEnabled()) return false
-        if (!isRemoteFeatureFlagEnabled(RemoteFeatureFlag.WOO_POS_TABLET_PROMO_BANNER)) return false
+        if (!featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_CLIENT_SIDE_BANNER)) return false
+        if (!featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_TABLET_PROMO_BANNER)) return false
```
irfano: @kidinov, I guess we don’t need these two flags here. `WOO_POS_TABLET_PROMO_BANNER` used to be a remote flag, and I didn’t remove one of them when combining the local and remote flags.
If you don’t have plans to use one of these flags in another area, I’ll remove one of them.

kidinov: @irfano 👋

From PR [#15214](https://github.com/woocommerce/woocommerce-android/pull/15214) , the description says:

  "This needed to make sure that when the remote FF enabled we won't show text that not A/B tested on the older versions of the app"

The flow was:
  1. First, a local CLIENT_SIDE_POS_BANNER flag was used during development
  2. Then PR #15200 replaced it with the remote woo_pos_tablet_promo_banner flag (controlled from the wpcom backend, also used for the JITM A/B test variations)
  3. Then PR #15214 re-added a local flag WOO_POS_CLIENT_SIDE_BANNER as a guard

The problem they solved: the remote flag gates both server-side JITMs and client-side banners. When they enable the remote flag for the A/B test, older app versions that already have the client-side banner code would start showing banners with copy that hasn't been A/B tested yet. The local flag lets you control which app versions show the client-side banner, independent of the remote flag being turned on for JITMs

JorgeMucientes:
Hey @malinajirka just caught up with you [comment](https://github.com/woocommerce/woocommerce-android/pull/15299#issuecomment-4037531479): 

> If the remote FF always takes precedents, we lose the ability to enable the feature in trunk.
Can you elaborate on the benefits of the new approach?

I think @irfano replied really well to what I meant. I think the key is this: 

>Another way to think about it is this: when adding a feature flag, we no longer need to think in terms of “local” vs “remote”. We just add a flag to FeatureFlag and use it everywhere that needs it.
If, at some point, we want that feature to be remotely controlled, we can add the corresponding backend key later, because every feature flag now has a remoteFlagKey.

And that's why the `Override > Remote > Default` (default being equivalent to what we have now as local FF) makes sense imho. 

About using option B out of this 3: 

> A. (Current) Override > Remote > Default
B. (Ignore Remote in debug builds) Override > Default
C. Adds a strategy field like in the screenshot

Imho I'd stick with A. I think B introduces harder mental model and unnecessary complexity. In any case, I don't have a strong opinion on it. So either options sound good to me. 

> I’m open to splitting this into smaller PRs if you think that is still the better direction. Otherwise, I’d prefer to keep moving with this PR.
@JorgeMucientes, I’d also appreciate your opinion here.

I'd say, now that you have started to work on this here. I'd wrap this up in this same PR. However, one thing I'd suggest doing is writting a P2 recap in Woomobile P2 to inform all Android devs about the new FF system, how it works and what benefits are we aiming for with these changes.

irfano:
@JorgeMucientes , after reading [Andrei's comment](https://github.com/woocommerce/woocommerce-android/pull/15299/#discussion_r2931618188) I realized that `Override > Remote > Default` has a gap for our use case. 

Consider this scenario:
- We start developing a feature.
- The feature is disabled in versions 20, 21, and 22 because it is still incomplete.
- It is completed in version 23.
- We then enable the remote flag on the backend.
- ❗ Versions 20, 21, and 22 could also start showing that incomplete feature.

To avoid that, I updated the resolution logic. For our use case, Jirka’s suggestion makes more sense: the local flag should gate feature readiness, and the remote flag should only control rollout once the local flag is enabled.

I pushed the change here: cf1b6ae473004913856a7ef0c77ac5b94b7204c8.

JorgeMucientes:
> We start developing a feature.
The feature is disabled in versions 20, 21, and 22 because it is still incomplete.
It is completed in version 23.
We then enable the remote flag on the backend.
❗ Versions 20, 21, and 22 could also start showing that incomplete feature.

Hey @irfano good point. Might be missing something, but I'm not sure how that would ever happen given remote FF are enabled on app version basis. For example, a recent remote flag that was added: 
Screenshot 2026-03-13 at 19 47 44

So in the above scenario the remote FF would be enabled starting on version 23 onwards. 

>For our use case, Jirka’s suggestion makes more sense: the local flag should gate feature readiness, and the remote flag should only control rollout once the local flag is enabled.

I'm ok with following that approach but that doesn't have the benefit you stated here: 

> Another way to think about it is this: when adding a feature flag, we no longer need to think in terms of “local” vs “remote”. We just add a flag to FeatureFlag and use it everywhere that needs it.

irfano:
Good point, @JorgeMucientes. You're right: if we use rules in the remote flag configuration, then my latest change is less necessary.
What the latest change gives us is a different trade-off: we would not need to express that version gating in remote config, because older app versions would still remain blocked by the local flag.

#### `Remote > Local` vs `Local > Remote`
I think both approaches can work. In practice, the only place where this changes behavior in this PR is `ClientSidePosBanner`, where we currently use both a local flag and a remote flag for the same feature.
The previous `ClientSidePosBanner` logic effectively behaved as `Local > Remote`, so my latest change preserves that behavior. I'd prefer to keep it that way, at least in this PR.

In other words, the goal of this PR is not to introduce a new product behavior, but to preserve the existing behavior while making the flag resolution model explicit and testable.

Thanks you all for the feedback.
@malinajirka, could you please review the latest version of the logic?

malinajirka (commented):
Thanks again for working on this @irfano!

I'm at a meetup this week, and then another event for 2 weeks. I did my best to review the latest changes myself as well as with the help of AI and it all looks good to me. Having said that, I couldn't find time to do a proper testing of the latest changes. Might be worth asking someone to do a manual test just to be sure.

malinajirka (changes requested):
Actually, I think it doesn't work as expected.

Remote-only flags like WOO_POS default to localValue = isDebugBuild(), which is false in release. Since the formula is local &&  remote, the remote value is ignored and the feature is always off in production. 

Fix: set localValue = true for flags that previously existed only as RemoteFeatureFlag.


Wdyt?

irfano:
Good catch, @malinajirka! My initial logic was `Remote > Local`, so it wasn't a problem. I changed it to `Local > Remote`, but then forgot to update default `localValue`. 
I've added the fix: e5606bb32b5c54a9de6f55f922ed2f2ea9214a2f

malinajirka:
Thanks for the fix @irfano ! LGTM.

A couple more behavioral changes I noticed (with the help of AI). I believe it would be good to double check these are expected and intentional. I think we discussed some of them, I'm not entirely sure if all of them. Following better safe than sorry principle, let's just double check.

**1. Missing remote value: pessimistic → optimistic default**

 | Scenario | Old | New |
  |---|---|---|
  | `WOO_POS`, release, remote=`true` | `true` | `true` |
  | `WOO_POS`, release, remote=`false` | `false` | `false` |
  | `WOO_POS`, release, remote=**absent** | **`false`** | **`true`** |

  "Absent" happens on fresh install before the first remote fetch completes (Room DB is empty). awaitRemoteFlagsLoaded() doesn't help — Room emits an empty list immediately, which satisfies the await, but all keys resolve to null. 

 **2. Debug builds no longer force-enable remote flags**

  Old: IsRemoteFeatureFlagEnabled returned isDebugBuild() || remoteValue, so debug builds always got true regardless of remote.

  New: effectiveValue = localValue && (remoteValue ?: true). If remote explicitly returns false, debug builds now see false.

**3. WOO_POS_CLIENT_SIDE_BANNER change**

Old: two separate flags — WOO_POS_CLIENT_SIDE_BANNER (local, isDebugBuild()) acted as a gate. Release always returned false before checking remote.

New: single flag WOO_POS_TABLET_PROMO_BANNER with localValue = true. In release, the banner can appear if remote is true or absent. It enables the banner in production.

---------------------------
Sorry for not sharing everything in one go, but this is quite complex to reason about.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+                remoteFlagValues.value = remoteFlags.associate { remoteFlag -> remoteFlag.key to remoteFlag.value }
+            }
+        }
+    }
+
+    fun isEnabled(flag: FeatureFlag) = getFlagState(flag).effectiveValue
+
+    fun observeIsEnabled(
+        flag: FeatureFlag
+    ): Flow<Boolean> = remoteFlagValues.map { isEnabled(flag) }.distinctUntilChanged()
```
AdamGrzybkowski: ⚠️ I believe this is a behavioral change compared to what we have now. I'm not sure if we trigger FF refetch when the app is running, but this could potentially trigger weird UI updates for the user while using the app. 

I'm worried this could lead to weird bugs for the users with hard-to-reproduce steps (FF changing while executing steps). I would prefer to have a "known" FF state for a given app launch. So, unless the app is restarted, FF values are constant.

irfano:
> WOO_POS, release, remote=absent | false | true
> -- | -- | --

Yes, this is a behavioral change. When the remote flag is absent, we use localValue.

A remote flag can be absent for two reasons:
 1. Intentionally: We don’t add a remote flag for that key. In this case, localValue is used as expected.
 2. Error: I think this is what your comment refers to. In this case, the new system uses localValue, whereas the old system used false.

The new system does not support the old behavior described in 2. But do we really need it? Consider WOO_POS: if the remote WOO_POS is absent for any reason, such as an error, I think it’s better to fall back to true because the feature is complete.

Do you think we should preserve the old behavior? If we want to keep it, we should change this to Remote > Local and rely on remote Rules for feature-ready versions, or give up on unifying the flags.

> 2. Debug builds no longer force-enable remote flags

I think this is better. There may be multiple features that are remotely disabled, and developers do not need to see them: 59df59fd7eb4045aec7d3d109cdf423c6573e44b

> 3. WOO_POS_CLIENT_SIDE_BANNER change

There were both a local flag and a remote flag used here, and we intentionally unified them into a single flag, but the behavior remains the same.

> Sorry for not sharing everything in one go, but this is quite complex to reason about.

No worries! Thank you for looking into this while you were at the meet-up. You shared valuable feedback.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlagRepository.kt
irfano: This PR doesn't introduce active feature flag observation. Before, local flags were synchronous and remote flags were asynchronous. When I unified them, I didn’t want to keep both `isEnabled()` and `suspend isRemoteEnabled()`, because that would make the unification pointless.
As solution, I introduced in-memory `remoteFlagValues` into `FeatureFlagRepository`. These in-memory `remoteFlagValues` are simply the in-memory equivalent of the remote flag values stored in the DB. Reading from the DB requires suspend functions, while reading the in-memory `remoteFlagValues` is synchronous. I keep the in-memory values updated via a DB observer.

```
init {
        appCoroutineScope.launch {
            featureFlagsStore.observeFeatureFlags().collect { remoteFlags ->
                remoteFlagValues.value = remoteFlags.associate { remoteFlag -> remoteFlag.key to remoteFlag.value }
            }
        }
    }
```

However, this introduced another risk: the in-memory values can be stale after the DB is updated, until the observer receives the change. That is why I originally added `observeIsEnabled()`.

But this is confusing. Having both `observeIsEnabled()` and `isEnabled()` goes against the purpose of the unification goal. After reconsidering it, I concluded that `observeIsEnabled()` is not really addressing a meaningful risk. Let me elaborate:

#### Old way
1. When the site is switched, fetching remote flags is triggered without waiting for the result.
2. After the remote flags are received, they are saved to the DB.
3. When the app needs to read a remote flag, it reads it from the DB.
4. ⚠️ There is a risk between steps 1 and 3. If step 2 takes too long, we may read stale data.

#### New way
1. When the site is switched, fetching remote flags is triggered without waiting for the result.
2. After the remote flags are received, they are saved to the DB.
3. After they are saved to the DB, the in-memory `FeatureFlagRepository.remoteFlagValues` is updated.
4. When the app needs to read a remote flag, it reads it from the in-memory `FeatureFlagRepository.remoteFlagValues`.
5. ⚠️ There is a risk between steps 1 and 4. If steps 2 and 3 take too long, we may read stale data.

`observeIsEnabled()` was intended to solve the **new** risk in step 3. But in practice, the risk introduced by step 3 is negligible compared to the existing risk in step 2. So `observeIsEnabled()` does not really solve anything meaningful and instead adds extra confusion. For that reason, I changed the implementation and removed `observeIsEnabled()`.

I’d appreciate it if you could read my reasoning above and confirm whether it makes sense.

> I would prefer to have a "known" FF state for a given app launch. So, unless the app is restarted, FF values are constant.

This is already the current behavior on `trunk`, and also in this PR, with one slight difference: feature flags are fetched not only on app startup, but also when switching sites.

malinajirka (approved):

malinajirka:
I believe this is not entirely precise:  `Consider WOO_POS: if the remote WOO_POS is absent for any reason, such as an error, I think it’s better to fall back to true because the feature is complete.` => what if the remote FF is attempting to enable a feature in version 10.2+ but since it fails to load, we enable it in 10.1- where the feature is not finished. Having said that, feels like an edge case which is likely not worth worrying about.

Thanks again for working on all this!

irfano:
> what if the remote FF is attempting to enable a feature in version 10.2+ but since it fails to load, we enable it in 10.1- where the feature is not finished.

Actually, the `localValue` on the client works like this:
 - `true`: The feature is completed and ready to be shipped.
 - `false`: The feature is still in progress and not ready.

So, if the remote feature flag is absent, the feature will fall back to `localValue`. In your example, that means it would be false on 10.1 and true on 10.2+, because we would set `localValue` to `true` once the feature is complete.

PR #15302, opened 2026-02-06
[WOOMOB-2133] Copy-paste and rename Orders ViewModel and mappers to Bookings equivalents

malinajirka:
**_### NOTE - I'm thinking about throwing aways this PR and implement the VM based on Kiwi's booking feature._**

Fixes WOOMOB-2133

Do not merge label - https://github.com/woocommerce/woocommerce-android/pull/15297 needs to be merged first.

**Note - this PR exceeds the limit of 300 lines of code. However, since vast majority of the code is copy-pasted => I believe the only review necessary is whether it makes sense from a conceptual point of view.**

### Description
This step copies the Orders ViewModel and its supporting mappers/providers into the Bookings package, renaming all `Order` references to `Booking` equivalents. Specifically:

- Copies `WooPosOrdersViewModel` → `WooPosBookingsViewModel`, replacing order-related types and logic with booking equivalents and removing search functionality
- Copies `WooPosOrderDetailsMapper` → `WooPosBookingDetailsMapper`
- Copies `WooPosOrderItemMapper` → `WooPosBookingItemMapper`
- Copies `WooPosOrderStatusMapper` → `WooPosBookingStatusMapper`
- Copies `WooPosOrderActionsProvider` → `WooPosBookingActionsProvider`
- Replaces `WooPosOrdersState` usage with `WooPosBookingsState`

### Test Steps
1. Open POS
2. Tap on more menu
3. Tap on Bookings
4. Notice a list of orders is loaded and selecting an order works

Next step - replace order model with booking model + replace data source

### Images/gif

Screenshot 2026-02-06 at 8 31 41

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15297, opened 2026-02-05
[WOOMOB-2133] Add WooPosBookings UI components and ViewModel (step 3)

malinajirka:
Fixes WOOMOB-2133

Do not merge label - https://github.com/woocommerce/woocommerce-android/pull/15296 needs to be merged first.

**Note - this PR exceeds the limit of 300 lines of code. However, since vast majority of the code is copy-pasted and there is no behavior to test => I believe the only review necessary is whether it makes sense from a conceptual point of view.**

### Description
This PR introduces the UI layer for POS Bookings by creating dedicated booking components based on the existing Orders screen (I essentially copy-pasted it and replaced order references with booking references). Key additions include:

- `WooPosBookingsScreen` - Main bookings list UI with item selection
- `WooPosBookingsViewModel` - ViewModel with dummy data and loading simulation
- `WooPosBookingsState` - Comprehensive state management for bookings
- `WooPosBookingsStatusBadge` - Status badge component for booking items
- `WooPosBookingsLoadingScreen` - Loading skeleton UI
- `WooPosBookingDetails` - Booking details panel

The implementation uses dummy hardcoded data to demonstrate the UI while the actual data layer is developed separately.

### Test Steps
- Test not needed, this is a WIP behind a FF

### Images/gif

https://github.com/user-attachments/assets/6063fa15-ddea-4571-b7b0-16358bdc7115

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15303, opened 2026-02-06
[WOOMOB-2133] Wire up WooPosBookingsViewModel with BookingListHandler (step 4)

malinajirka:
Fixes WOOMOB-2133

Do not merge label - https://github.com/woocommerce/woocommerce-android/pull/15297 needs to be merged first.

### Description

**Note:** This PR is a first PR in this chain that requires a proper review. Thanks!

Replaces dummy/hardcoded booking data in `WooPosBookingsViewModel` with real data fetched and observed from `BookingListHandler`. 

I first attempted to implement the VM by copy-pasting the POS Order List ViewModel - [in this PR](https://github.com/woocommerce/woocommerce-android/pull/15302). However, it turned out to be way more complex and messy. The BookingListHandler is well written and encapsulated so it's great for our use-case. If we ever decide to split the POS Bookings from Store Management bookings, we can always do that by replacing the handler with a POS specific implementation. 

Implements all ViewModel interactions: booking selection, pull-to-refresh, pagination with load-more, error/empty state retry, UI event handling. The ViewState entity mapping is dummy for now, since we don't have clear designs yet.

**Note: This PR exceeds the limit of 300 lines of business logic by roughly 30 lines of code since I had to remove the mocks, so I hope that's fine.**

https://github.com/user-attachments/assets/0cccb894-09da-4d17-b754-9c2298bb7d7d

### Test Steps

1. First you need to enable POS on CIAB site, you can use the patch below.
3. Open POS and navigate to the bookings screen
4. Verify bookings load from real data and the first booking is auto-selected
5. Tap a different booking and verify the detail pane updates
6. Pull to refresh and verify the list reloads
7. Scroll to the bottom and verify pagination triggers
8. Simulate error/empty states and verify retry actions work

## Patch
```
Subject: [PATCH] ENABLE POS IN CIAB
---
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt	(revision 2f6531899a8bc2c61ded6e4ac886dcc80a45ac53)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisible.kt	(date 1770369072800)
@@ -31,12 +31,12 @@
             return@withContext Result.success(true)
         }
 
-        if (ciabSiteGateKeeper.isFeatureUnsupported(CIABAffectedFeature.POS)) {
-            appPrefs.clearPOSTabVisibilityForSite(site.id)
-            return@withContext Result.success(false).also {
-                wooPosLog.i("POS Tab Not visible reason: Site is CIAB")
-            }
-        }
+//        if (ciabSiteGateKeeper.isFeatureUnsupported(CIABAffectedFeature.POS)) {
+//            appPrefs.clearPOSTabVisibilityForSite(site.id)
+//            return@withContext Result.success(false).also {
+//                wooPosLog.i("POS Tab Not visible reason: Site is CIAB")
+//            }
+//        }
 
         if (!isRemoteFeatureFlagEnabled(WOO_POS)) {
             appPrefs.clearPOSTabVisibilityForSite(site.id)
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt	(revision 2f6531899a8bc2c61ded6e4ac886dcc80a45ac53)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt	(date 1770368930065)
@@ -11,5 +11,4 @@
     CompositeProducts,
     GiftCardEditing,
     ProductsStockDashboardCard,
-    POS
 }
Index: WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt	(revision 2f6531899a8bc2c61ded6e4ac886dcc80a45ac53)
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosTabShouldBeVisibleTest.kt	(date 1770369072805)
@@ -34,9 +34,7 @@
     private val wooCommerceStore: WooCommerceStore = mock()
     private val isScreenSizeAllowed: WooPosIsScreenSizeAllowed = mock()
     private val isRemoteFeatureFlagEnabled: IsRemoteFeatureFlagEnabled = mock()
-    private val ciabSiteGateKeeper: CIABSiteGateKeeper = mock {
-        on { isFeatureUnsupported(CIABAffectedFeature.POS) } doReturn false
-    }
+    private val ciabSiteGateKeeper: CIABSiteGateKeeper = mock()
 
     private lateinit var sut: WooPosTabShouldBeVisible
     private lateinit var siteModel: SiteModel
@@ -131,16 +129,6 @@
         assertTrue(r.isSuccess)
         assertFalse(r.getOrThrow())
     }
-
-    @Test
-    fun `given feature unsupported for CIAB site, when invoked with forceRefresh, then return success false`() = testBlocking {
-        whenever(ciabSiteGateKeeper.isFeatureUnsupported(CIABAffectedFeature.POS)).thenReturn(true)
-
-        val r = sut(forceRefresh = true)
-
-        assertTrue(r.isSuccess)
-        assertFalse(r.getOrThrow())
-    }
 
     @Test
     fun `given cached value is true, when invoked without forceRefresh, then return cached value without remote call`() = testBlocking {
```

### Images/gif


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -2,157 +2,289 @@ package com.woocommerce.android.ui.woopos.bookings
...
-            status = PosBookingStatus("Completed", BookingStatusColorKey.COMPLETED),
-            statusSlug = "completed",
-            createdAtMillis = System.currentTimeMillis()
-        )
+    private fun observeBookings() {
+        viewModelScope.launch {
+            bookingListHandler.bookingsFlow.collectLatest { bookings ->
+                if (bookings.isEmpty() && _state.value is WooPosBookingsState.Loading) {
+                    return@collectLatest
+                }
```
samiuelson, line 75:
```suggestion

```
❓ In case we receive empty `Booking[]` result, we get stuck in `loading` state.

samiuelson (commented):
Thanks, @malinajirka. This looks good overall! One issue I found, though, is that I'm not able to achieve Empty state.

kidinov:
@malinajirka regarding usage of the ListHandler:

```
  What BookingListHandler gives you                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
  It already supports BookingFilters (including DateRange, attendance status, etc.), pagination, and returns BookingEntity which has embedded BookingOrderInfo with customer names, product names, and payment totals. This enrichment happens in BookingDtoMapper.toEntity() which pulls data from the local DB (OrderEntity,
   ProductsDao).                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                              
  Upsides of using it                                                                                                                                                                                                                                                                                                         

  - No new data layer code. Pagination, filtering, DB observation - all there already.
  - Reactive. bookingsFlow emits updates when the DB changes, so if something else updates an order, you get it for free.
  - Tab filtering is doable. You can pass different BookingFilters with DateRange to get today/upcoming bookings.

  Downsides of using it

  - Enrichment depends on local DB state. The customer names, product names, and payment totals come from OrderEntity that must already be fetched and stored locally. In a fresh POS session, many orders likely won't be cached. So booking.order.customerInfo might just be null, and you end up showing "#123" instead of
  a name.
  - The custom data source solves this explicitly. PR #15295 calls CustomerRestClient and ProductRestClient directly to fetch names on demand, then caches them. It also fetches order totals via OrderRestClient when a booking is selected. More API calls, but the data is actually there.
  - Less control over the flow. The handler was built for the store management screen. If POS needs different behavior (e.g. different sort per tab, different page sizes, different error handling), you're constrained by its API.
  - BookingEntity is a DB entity. You're coupling POS to the Room persistence layer even though POS may not need persistence at all.

  Bottom line

  The handler is a good shortcut to get things working fast, but the enrichment gap is real. If the local DB doesn't have the orders cached, you'll have missing customer/product info with no easy way to fill it from inside the handler. The custom data source is more work but gives POS what it actually needs (fresh
  API data with resolved names).
```

The POS doesn't store orders/customers/products and so on in the DB, so we may end up with a lot of unclear and undefined behavior with this (e.g. some entries will be stored in the DB from the main app and some won't, and it may not be that clear to catch this), so I'd suggest considering using remote data only for the MLP

malinajirka:
Thanks for the review @samiuelson and @kidinov !

> - Enrichment depends on local DB state. The customer names, product names, and payment totals come from OrderEntity that must already be fetched and stored locally. In a fresh POS session, many orders likely won't be cached. So booking.order.customerInfo might just be null, and you end up showing "#123" instead of
  a name.
>  - The custom data source solves this explicitly. PR #15295 calls CustomerRestClient and ProductRestClient directly to fetch names on demand, then caches them. It also fetches order totals via OrderRestClient when a booking is selected. More API calls, but the data is actually there.
  
These two contradict each other AFAICT - it already fetches the data and doesn't rely on them being stored in the database.

Having said that, I agree it creates dependencies that we don't want to have in the POS (longterm). However, given the circumstances, I'm thinking we should take all the shortcuts we can and implement the flow end-to-end. When we have it, we can iterate on it and improve it. That way, we'll have a working solution earlier. 

The list handler can be IMO easily replaced with our POS specific solution in a couple weeks if time allows or when we start observing issues - I don't think we should try fixing issues, I think we should replace the listhandler if they come. 

What do you think @samiuelson (Andrei is AFK)? Do you agree with this train of thoughts/approach or do you think it's better to implement a POS specific solution right away?

hichamboushaba:
Hey @malinajirka, please note that I can't check this PR earlier than Wednesday (I'm still doing my support rotation, and also am the release manager this week), so unless you are OK with this, I would suggest assigning the PR to someone else from the team 🙂.

malinajirka:
Thanks @hichamboushaba! Tbh I must have assigned this PR to you by accident - not sure how it happened. I also noticed it contains tons of unrelated changes - not sure what happened here 🤔

samiuelson:
> Thanks for the review @samiuelson and @kidinov !
> 
> > * Enrichment depends on local DB state. The customer names, product names, and payment totals come from OrderEntity that must already be fetched and stored locally. In a fresh POS session, many orders likely won't be cached. So booking.order.customerInfo might just be null, and you end up showing "[Replace default app icon with temporary Woo one #123](https://github.com/woocommerce/woocommerce-android/pull/123)" instead of
> >   a name.
> > * The custom data source solves this explicitly. PR [[Woo POS] Bookings POC #15295](https://github.com/woocommerce/woocommerce-android/pull/15295) calls CustomerRestClient and ProductRestClient directly to fetch names on demand, then caches them. It also fetches order totals via OrderRestClient when a booking is selected. More API calls, but the data is actually there.
> 
> These two contradict each other AFAICT - it already fetches the data and doesn't rely on them being stored in the database.
> 
> Having said that, I agree it creates dependencies that we don't want to have in the POS (longterm). However, given the circumstances, I'm thinking we should take all the shortcuts we can and implement the flow end-to-end. When we have it, we can iterate on it and improve it. That way, we'll have a working solution earlier.
> 
> The list handler can be IMO easily replaced with our POS specific solution in a couple weeks if time allows or when we start observing issues - I don't think we should try fixing issues, I think we should replace the listhandler if they come.
> 
> What do you think @samiuelson (Andrei is AFK)? Do you agree with this train of thoughts/approach or do you think it's better to implement a POS specific solution right away?

👋 @malinajirka. I opt for reusing the `BookingListHandler` at the moment and migrating to POS-specific data source later. We've iterated similar to this in past and I think it's safer approach at the moment 👍

malinajirka:
Thanks @samiuelson !

> One issue I found, though, is that I'm not able to achieve Empty state.

This is fixed now and a unit test is added. (I also renamed some of the composables to avoid accidentaly importing composables for the store management booking feature..

samiuelson (approved):
LGTM!

PR #15313, opened 2026-02-09
Add remote flags for Woo PN project

irfano:
### Description

We added a feature flag to the backend with the key `woo_self_driven_push_notifications_m1`. This PR renames our local flags and adds keys for the remote flags.
Backend PR: 202198-ghe-Automattic/wpcom
The default value (`false`) will be used until the backend returns a value for `woo_self_driven_push_notifications_m1`.

### Test Steps

For now, it’s not testable because the remote flag PR hasn’t been merged on the backend yet.

### Images/gif


WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -20,8 +20,8 @@ enum class FeatureFlag(
     POS_REFUNDS("pos_refunds"),
     POS_PRODUCTS_FTS("pos_products_fts"),
     WOO_POS_LOCAL_CATALOG_FILE_APPROACH("woo_pos_local_catalog_file_approach"),
-    WOO_PUSH_NOTIFICATIONS_SYSTEM("woo_push_notifications_system", default = false),
-    WOO_PUSH_NOTIFICATIONS_SYSTEM_M2("woo_push_notifications_system_m2", default = false),
+    WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1("woo_self_driven_push_notifications_m1", default = false),
```
JorgeMucientes: What about using the local feature flag value as default value when not being able to retrieve the remote FF? I thought that was the idea from what we discussed over Slack

irfano: `default = false` does that. If you want to use `true` when the remote flag isn't available, just update the default value to `true`.

I added the feature in #15299 
> Note: If no remote value exists, the default is used.

JorgeMucientes (approved):
Changes look good @irfano 👍🏼

PR #15316, opened 2026-02-09
[WOOMOB-2181] Introduce booking-specific state model and event types (step 5)

malinajirka:
Fixes WOOMOB-2181

### Description
Replace the order-centric `Computed`/`Lazy` state hierarchy with a flat, booking-specific state model for the POS booking details screen. This is a pure scaffolding change — the ViewModel produces dummy/minimal values and the details pane still shows the placeholder "Booking Detail #xxx" text.

This is not meant to be the final structure - but something we can iterate on.

Key changes:
- Flat `BookingDetailsViewState` data class with `CustomerSection`, `AttendanceSection` (type-safe `AttendanceState?` selection), and `PaymentSection`
- New UI event types: `AttendanceToggled`, `PayByCardClicked`, `PayByCashClicked`, `AddBookingNoteClicked`, `CopyEmailClicked`
- ViewModel wired with empty event handlers and dummy mapping

### Test Steps
1Green CI

### Images/gif
No visual changes — this is a state model refactor only.


samiuelson (approved):
LGTM!

Issue #1758, opened 2019-12-26
Woo: Stats - UX Audit issues


anitaa1990:
#### Issue 1:
When there’s no revenue for the period selected, make the font size for “No revenue this period” and “No activity this period” the same.

<img width="249" alt="screen-shot-2019-11-26-at-12 17 13-pm" src="https://user-images.githubusercontent.com/22608780/71458043-f5426b00-27c6-11ea-97f3-12629f62b535.png">

#### Issue 2
The x axis in the Revenue data graph for “This Week” needs to display all the day of the week and the column width of the daily revenue should stay consistent along the week (currently changes depending on how many days have passed in the week)
<img width="776" alt="screen-shot-2019-11-26-at-12 08 17-pm" src="https://user-images.githubusercontent.com/22608780/71458044-f5db0180-27c6-11ea-960c-a948f23ec0f6.png">

#### Issue 3
Update the top and bottom vertical padding for the “Updated moments ago” text to 16px.
<img width="681" alt="screen-shot-2019-11-26-at-12 03 31-pm" src="https://user-images.githubusercontent.com/22608780/71458047-f70c2e80-27c6-11ea-8bcf-0b35f2393ded.png">

PR #15319, opened 2026-02-09
Show canceled bookings with new canceled label

JorgeMucientes:
## Summary

Closes: WOOMOB-2176

## Changes

- Add info-style blue colors (`#E1ECF8` background, `#001758` text) to the cancelled attendance status tag in the bookings list and detail screens
- Update `cancelled` status to `canceled` with single L, as desktop

## Test Steps
- Log into a site with Bookings enabled
- Cancel a booking
- Open Bookings tab -> `All` and check the canceled bookings is displayed as follows (see links to Figma designs in the Linear issue):


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingStatusTag.kt
```diff
@@ -44,7 +44,7 @@ private fun BookingStatus.text(): String {
         BookingStatus.PendingConfirmation -> stringResource(R.string.booking_payment_status_pending_confirmation)
         BookingStatus.Confirmed -> stringResource(R.string.booking_payment_status_confirmed)
         BookingStatus.Paid -> stringResource(R.string.booking_payment_status_paid)
-        BookingStatus.Cancelled -> stringResource(R.string.booking_payment_status_cancelled)
+        BookingStatus.Cancelled -> stringResource(R.string.booking_attendance_status_cancelled)
```
JorgeMucientes, line 47: We use `cancelled` with double L everywhere in the code so I'm good with leaving that as is, and just updated the user facing values.

irfano (approved):
LGTM! 👍🏻

PR #15322, opened 2026-02-10
Post both Prototype Build comments expanded

mokagio:
Follow up to
https://github.com/woocommerce/woocommerce-android/pull/15197#discussion_r2775383155

I suppose folding the Wear info made sense when Prototype Builds were generated on every build and for both platforms. The Wear app was less likely to be the one people were interested in, so we were better off saving screen real estate and folding the QR code and other info.

In the new setup with optional builds, it's almost guaranteed that if a link is posted, someone needs that information. Therefore, I think it makes more sense to expand the info and make it immediately visible.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1391,8 +1391,7 @@ platform :android do
         Flavor: prototype_flavor,
         'Build Type': prototype_build_type
       },
-      # Fold for Wear app, as it's modified less often
-      fold: app == WEAR_APP,
+      fold: false,
       reuse_identifier: "#{app.downcase}-prototype-build-link"
     )
 


AliSoftware (approved):
Approving the change on a technical point of view and because I think the rationale for doing it makes sense; but maybe devs might want to chime in about if that makes sense for them too

mokagio:
@woocommerce/android-developers I had auto-merge on this so it got merged before I could ping you following up to Olivier's comment.

What do you think? Does it make sense experience wise to have the Wear Prototype Build comment expanded?

> I suppose folding the Wear info made sense when Prototype Builds were generated on every build and for both platforms. The Wear app was less likely to be the one people were interested in, so we were better off saving screen real estate and folding the QR code and other info.
>
> In the new setup with optional builds, it's almost guaranteed that if a link is posted, someone needs that information. Therefore, I think it makes more sense to expand the info and make it immediately visible.

PR #15323, opened 2026-02-10
[CIAB: POS] Collect cash payment for booking

samiuelson:
WOOMOB-2195

## Summary

Enables cash payment collection for bookings by integrating the existing cash payment screen into the bookings flow. When a booking requires payment, users can now complete the transaction with cash directly from the booking details screen.

💡 Don't merge until parent branch is `trunk`

## Changes

**Bookings Module**
- Added `orderId` field to `BookingDetailsViewState` to track the associated order for payment processing
- Updated `WooPosBookingViewStateMapper` to include `orderId` when mapping booking entities to view state
- Added `navigationEvent` SharedFlow to `WooPosBookingsViewModel` to emit navigation events
- Implemented `handlePayByCash()` method in `WooPosBookingsViewModel` to trigger cash payment navigation with booking context
- Enhanced `WooPosBookingsScreen` to collect navigation events from the view model
- Added `savedStateHandle` observation in `WooPosBookingsScreen` to detect successful payment completion via `BOOKING_CASH_PAYMENT_SUCCESS_KEY` and trigger data refresh

**Cash Payment Module**
- Introduced `CashPaymentSource` enum (`CHECKOUT`, `BOOKINGS`) to distinguish payment flow origin
- Added `source` query parameter to the cash payment navigation route
- Modified `WooPosCashPaymentViewModel` to extract and store the `source` from `savedStateHandle`
- Enhanced `WooPosCashPaymentScreen` to conditionally handle completion based on source: `BOOKINGS` returns to bookings screen with success result, `CHECKOUT` navigates to home (existing behavior)

**Navigation**
- Updated `WooPosNavigationEvent.OpenCashPayment` to include `source` field with `CHECKOUT` as default
- Modified `WooPosNavigationEventHandler` to pass the source parameter when handling cash payment navigation
- Updated `WooPosBookingsNavigation` to pass `backStackEntry` to `WooPosBookingsScreen` for savedStateHandle access

## Test plan

- [ ] Verify "Pay by Cash" button on an unpaid booking navigates to the cash payment screen
- [ ] Complete a cash payment from the bookings screen and verify you return to bookings (not home)
- [ ] Verify bookings list refreshes after successful cash payment
- [ ] Verify pressing back from cash payment screen returns to bookings correctly
- [ ] Verify existing cash checkout flow from POS totals screen still works unchanged

## pics

https://github.com/user-attachments/assets/f99765f0-cc78-4368-a16a-b55811d815ca

PR #15326, opened 2026-02-10
[CIAB: POS] Card-present booking checkout. Part-1: Allow WooPayments in CIAB gatekeeper

samiuelson:
WOOMOB-2196 Part 1

## Summary
- This PR enables card payment functionality on CIAB site
- Changes `isFeatureSupported()` in `CIABSiteGateKeeper` from blocking all features on CIAB sites to allowing `WooPayments` specifically

💡 Don't merge until parent branch is `trunk`
👉 Parent PR: [CIAB: POS] Collect cash payment for booking #15323

## Test plan
- [ ] Verify WooPayments works on CIAB sites - enter POS mode on CIAB site and collect payment for regular product

## Images/gif
—

- [x]  I have considered if this change warrants release notes and have added them to RELEASE-NOTES.txt if necessary. Use the "[Internal]" label for non-user-facing changes.


WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
```diff
@@ -9,9 +9,10 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
         @Suppress("unused")
```
malinajirka, line 9: 🔍 We can remove the suppress here.

WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
```diff
@@ -16,13 +16,20 @@ class CIABSiteGateKeeperTest : BaseUnitTest() {
...
+        CIABAffectedFeature.entries
+            .filter { it != CIABAffectedFeature.WooPayments }
+            .forEach {
+                assertFalse(ciabSiteGateKeeper.isFeatureSupported(it))
+            }
+    }
+
+    @Test
+    fun `given current site is CIAB, when checking WooPayments support, then feature is supported`() {
+        assertTrue(ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.WooPayments))
```
malinajirka, line 32: ⚠️ I'm not sure if it's an issue of the test or the behavior of the class or the name of the test, but seems like the test name suggests we are testing CIAB site, however, the test doesn't ensure this is tested on a CIAB site, right? Do we need to add
```
val site = createSite(isCIAB = true)
  given(selectedSite.getOrNull()).willReturn(site)
```
as in the previous test?

malinajirka (approved):
LGTM, I've left one comment I'm not sure about.

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
```diff
@@ -9,9 +9,10 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
...
         feature: CIABAffectedFeature
     ): Boolean {
-        // For now, all affected features are unsupported in CIAB.
-        // If there are exceptions in the future, we can handle them here.
-        return !isCurrentSiteCIAB()
+        return when (feature) {
+            CIABAffectedFeature.WooPayments -> true
```
malinajirka, line 13: ⚠️ Actually, just to be sure we are on the same page here - this will enable WooPayments gated features across the app, including:
- Payments section in the More menu
- Tap to Pay and Card Payments in Order creation
- Payments shortcut

WooCommerce/src/test/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeperTest.kt
samiuelson: This would result in `Unnecessary stubbings detected in test class: CIABSiteGateKeeperTest` error and failed test. This is because the `isCurrentSiteCIAB()` check is not reached in `CIABSiteGateKeeper::isFeatureSupported` in this case.

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
samiuelson: This is intended. Discussion: p1770830594513349-slack-C070SJRA8DP

PR #15328, opened 2026-02-10
Simplify booking attendance status to Attended/Unattended

JorgeMucientes:
Closes: WOOMOB-2117

⚠️ Please don't panic about the many changes. Many are just one liner due to replacing the different legacy bookings statuses with the new ones. 

### Changes

- **Attendance statuses**: Replace Booked/CheckedIn/NoShow/Cancelled with Attended/Unattended across entity model, UI model, mapper, filters, and tests
- **Toggle button**: Remove `BookingAttendanceStatusBottomSheet` and `BookingAttendanceSection`. Add a "Mark as attended" / "Mark as unattended" `WCOutlinedButton` in the appointment details section, above the cancel button. Tapping toggles immediately without a confirmation dialog
- **Cancelled bookings**: Hide attendance tag in both list and detail views. Hide the attendance toggle button. Exclude cancelled bookings from Today and Upcoming tabs (only visible in All tab)
- **Complete status**: Hide the "Complete" booking status tag everywhere (`BookingSummaryModel.status` is now nullable, returns `null` for Complete)
- **Cleanup**: Delete `BookingAttendanceStatusBottomSheet.kt`, `BookingAttendanceSection.kt`, 3 unused drawables, and 7 unused string resources

> **Note**: POS booking code is intentionally unchanged — it uses its own independent `WooPosBookingsState.AttendanceState` model. Only the shared `BookingEntity.AttendanceStatus` was updated, along with the POS test that referenced it.

https://github.com/user-attachments/assets/a1e3ed35-4bb6-41e7-a203-e4d6cab62d02

### Attendance status values
- [ ] Open the bookings list → each booking card shows the correct attendance tag
- [ ] Open the filter screen → attendance filter shows "Any", "Attended", and "Unattended" options
- [ ] Apply an attendance filter → list updates correctly. Note that `unattended` bookings also include canceled ones as those are unattended by default. 

### Attendance toggle button
- [ ] Open a non-cancelled booking detail → "Mark as attended" or "Mark as unattended" button is visible above the cancel button
- [ ] Tap the button → status toggles immediately (e.g., Unattended → Attended), button text updates, tag in summary updates
- [ ] While toggling, button shows a loading spinner and is disabled
- [ ] Toggle back → status reverts correctly
- [ ] Disconnect from network → tap toggle → offline error snackbar appears

### Cancelled bookings
- [ ] Open a **cancelled** booking detail → no attendance tag is shown in the summary, no attendance toggle button is visible
- [ ] In the bookings list, switch to **Today** tab → cancelled bookings are NOT shown
- [ ] Switch to **Upcoming** tab → cancelled bookings are NOT shown
- [ ] Switch to **All** tab → cancelled bookings ARE shown
- [ ] Cancelled bookings in the All tab list → no attendance tag is visible

### Complete status hidden
- [ ] If a booking has "complete" status from the API → no status tag is shown in the summary (neither in list nor detail)

### Cancel button still works
- [ ] Open a non-cancelled booking → cancel button is visible below the attendance toggle button
- [ ] Tap cancel → confirmation dialog appears and works correctly

### POS bookings unaffected
- [ ] Open POS bookings screen → bookings display correctly with their own attendance states


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4341,24 +4341,17 @@
...
     <string name="booking_payment_status_complete">Complete</string>
     <string name="booking_payment_status_pay_on_site">Pay on site</string>
-    <string name="booking_attendance_status_booked">Booked</string>
-    <string name="booking_attendance_status_checked_in">Checked-in</string>
-    <string name="booking_attendance_status_cancelled">Cancelled</string>
-    <string name="booking_attendance_status_no_show">No-show</string>
+    <string name="booking_attendance_status_attended">Attended</string>
+    <string name="booking_attendance_status_unattended">Unattended</string>
+    <string name="booking_mark_as_attended">Mark as attended</string>
+    <string name="booking_mark_as_unattended">Mark as unattended</string>
```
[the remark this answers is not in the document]
JorgeMucientes: Doesn't apply. Afaik this should be removed by CI after removing the strings from main strings file?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
```diff
@@ -165,20 +148,19 @@ private fun BookingDetailsContent(
     BookingAppointmentDetails(
         model = booking.bookingsAppointmentDetails,
         onCancelBooking = onCancelBooking,
+        onAttendanceToggle = {
+            val targetStatus = when (booking.bookingSummary.attendanceStatus) {
+                BookingAttendanceStatus.Attended -> BookingAttendanceStatus.Unattended
+                else -> BookingAttendanceStatus.Attended
+            }
+            booking.onAttendanceStatusSelected(targetStatus)
```
hichamboushaba: minor np, I think this logic is better be in the ViewModel and not here, I think we can rename the callback in BookingUiState to `onAttendanceToggle` to match what we have in the UI, then handle the toggling (identifying the target status) part in the ViewModel.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -79,15 +79,20 @@ class BookingListViewModel @Inject constructor(
...
         selectedBookingIdFlow,
-    ) { bookings, loadingState, selectedBookingId ->
+        selectedTab,
+    ) { bookings, loadingState, selectedBookingId, tab ->
+        val filtered = if (tab != BookingListTab.All) {
+            bookings.filter { it.status != BookingEntity.Status.Cancelled }
+        } else {
+            bookings
+        }
+        openFirstLoadedBookingOnTablet(filtered)
```
JorgeMucientes: After what we recently chat about. Hold on, and don't review this filtering part just yet. I'm looking into filtering this in the API call instead of locally

hichamboushaba: I was almost done about the PR Review before the call, WDYT about merging this PR as is, and update the filtering logic in a separate PR? Or alternatively just revert this part, and handle the filtering in a separate PR.

JorgeMucientes: Applied, ready for full review @hichamboushaba

hichamboushaba (approved):
Nice work @JorgeMucientes, works well.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsScreen.kt
JorgeMucientes: Good suggestion, applied 👍🏼 [1705445](https://github.com/woocommerce/woocommerce-android/pull/15328/commits/170544509b56c7e907c84ff6ddeb9fc35381920f)

PR #15334, opened 2026-02-11
[CIAB MVP] Rearrange bookings and POS entry points

irfano:
WOOMOB-2164

### Description

This PR updates how Bookings is exposed in navigation for large-screen layouts.
- Shows Bookings as a **More menu item** on tablet/large-window layouts.
- Keeps Bookings as a **bottom tab** on phone/compact layouts.
- When opened from More menu, Bookings shows top app bar and hides bottom navigation.
- Uses the existing More menu analytics flow by tracking Bookings with option value `bookings`.
- Sets Bookings button initial state to `Hidden` to avoid loading/flicker behavior.
- Renames `ObserveBookingsTabVisibility` to `ObserveBookingsVisibility`.
- Updates CIAB gatekeeper behavior so POS remains supported while other CIAB-affected features stay gated.

### CIAB site + Tablet
- Login to CIAB site on a tablet
- Make sure the POS is displayed in tab bar
- Make sure there is no "Bookings" tab in tab bar
- Navigate to "More" menu
- Make sure the "Bookings" entry is presented in menu
- Tap on "Bookings" entry. Make sure the bookings screen is presented.

### CIAB site + phone
- Login to CIAB site on iPhone
- Make sure the is no POS is displayed in tab bar
- Make sure there is the "Bookings" tab in tab bar
- Navigate to "More" menu
- Make sure no "Bookings" entry is presented in menu
- Tap on "Bookings" entry. Make sure the bookings screen is presented.

### NON-CIAB site + Tablet
- Login to non-CIAB site on iPad
- Make sure the POS is displayed in tab bar
- Make sure there is no "Bookings" tab in tab bar
- Navigate to "More" menu
- Make sure there is no "Bookings" entry is presented in menu

### NON-CIAB site + phone
- Login to non-CIAB site on iPhone
- Make sure the is no POS is displayed in tab bar
- Make sure there is no "Bookings" tab in tab bar
- Navigate to "More" menu
- Make sure no "Bookings" entry is not presented in menu

### Images/gif

Screenshot_20260211_205142
Screenshot_20260211_205137

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
@@ -56,13 +56,13 @@ class BookingListFragment : TopLevelFragment(), TabletLayoutSetupHelper.Screen {
             ).toBundle()
         )
     override val activityAppBarStatus: AppBarStatus
-        get() = if (navArgs.launchedFromMoreMenu) {
-            AppBarStatus.Visible(hasShadow = false, hasDivider = true)
-        } else {
+        get() = if (navArgs.showBottomNavigation) {
             AppBarStatus.Hidden
+        } else {
+            AppBarStatus.Visible(hasShadow = false, hasDivider = true)
         }
     override val shouldShowBottomNavigation: Boolean
-        get() = !navArgs.launchedFromMoreMenu
+        get() = navArgs.showBottomNavigation
 
     private val viewModel: BookingListViewModel by viewModels()
     private val navArgs: BookingsArgs by navArgs()
@@ -155,7 +155,7 @@ class BookingListFragment : TopLevelFragment(), TabletLayoutSetupHelper.Screen {
 
     private fun handleBottomNavigationVisibility() {
         viewModel.bottomNavigationVisible.observe(viewLifecycleOwner) { isVisible ->
-            if (navArgs.launchedFromMoreMenu || !isVisible) {
+            if (!navArgs.showBottomNavigation || !isVisible) {
                 (activity as? MainActivity)?.hideBottomNav()
             } else {
                 (activity as? MainActivity)?.showBottomNav()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
@@ -10,6 +10,8 @@ import androidx.fragment.app.activityViewModels
 import androidx.fragment.app.viewModels
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.fragment.findNavController
+import androidx.navigation.fragment.navArgs
+import com.woocommerce.android.BookingsArgs
 import com.woocommerce.android.R
 import com.woocommerce.android.databinding.FragmentBookingListBinding
 import com.woocommerce.android.extensions.isTwoPanesShouldBeUsed
@@ -54,9 +56,16 @@ class BookingListFragment : TopLevelFragment(), TabletLayoutSetupHelper.Screen {
             ).toBundle()
         )
     override val activityAppBarStatus: AppBarStatus
-        get() = AppBarStatus.Hidden
+        get() = if (navArgs.launchedFromMoreMenu) {
+            AppBarStatus.Visible(hasShadow = false, hasDivider = true)
+        } else {
+            AppBarStatus.Hidden
+        }
+    override val shouldShowBottomNavigation: Boolean
+        get() = !navArgs.launchedFromMoreMenu
 
     private val viewModel: BookingListViewModel by viewModels()
+    private val navArgs: BookingsArgs by navArgs()
     private val bookingsCommunicationViewModel: BookingsCommunicationViewModel by activityViewModels()
 
     @Inject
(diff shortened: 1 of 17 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
```diff
@@ -72,10 +76,18 @@ class MoreMenuViewModel @Inject constructor(
...
+    isWindowClassLargeThanCompact: IsWindowClassLargeThanCompact,
     private val analyticsTrackerWrapper: AnalyticsTrackerWrapper,
     private val ciabSiteGateKeeper: CIABSiteGateKeeper,
 ) : ScopedViewModel(savedState) {
     private var storeHasGoogleAdsCampaigns = false
+    private val shouldShowBookingsInMenu = isWindowClassLargeThanCompact()
+    private val bookingsVisibilityFlow = if (shouldShowBookingsInMenu) {
+        observeBookingsVisibility()
+    } else {
+        flowOf(false)
```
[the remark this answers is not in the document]
irfano: Valid point! I fixed it in 6f93f2264d4e09d38bd62db43f791e7506cc2d35, but I didn't apply the change suggested here because `isWindowClassLargeThanCompact()` uses the application context, so it wouldn't fix the issue described.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModelTests.kt
```diff
@@ -431,6 +444,42 @@ class MoreMenuViewModelTests : BaseUnitTest() {
...
+        val bookingsButton =
+            states.last().menuSections.flatMap { it.items }.first { it.title == R.string.bookings_tab_title }
+        assertThat(bookingsButton.description).isEqualTo(R.string.more_menu_button_bookings_description)
+    }
+
+    @Test
+    fun `given bookings is visible and not on tablet, when building state, then bookings button is not displayed`() =
+        testBlocking {
+            // GIVEN
+            setup {
```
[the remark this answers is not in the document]
irfano: Fixed: 16fc386bf8be3c482188fff820bd9e37bb805305

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
```diff
@@ -54,9 +56,16 @@ class BookingListFragment : TopLevelFragment(), TabletLayoutSetupHelper.Screen {
...
         )
     override val activityAppBarStatus: AppBarStatus
-        get() = AppBarStatus.Hidden
+        get() = if (navArgs.launchedFromMoreMenu) {
+            AppBarStatus.Visible(hasShadow = false, hasDivider = true)
+        } else {
+            AppBarStatus.Hidden
+        }
+    override val shouldShowBottomNavigation: Boolean
+        get() = !navArgs.launchedFromMoreMenu
```
hichamboushaba: suggestion: WDYT about using `findNavController().previousBackStackEntry?.destination?.id == BottomNavigationPosition.MY_STORE.id` here? This will avoid using an additional launch argument, and make the logic independent from being launched from the More fragment or anywhere else.

hichamboushaba: If you prefer keeping the launch argument, then I'd suggest renaming the argument to something like `showBottomNavigation`, just to make the role clearer, and make it extensible if the fragment were to be launched from elsewhere.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
```diff
@@ -19,12 +20,20 @@ class BookingsTabController @Inject constructor(
...
+        binding.bottomNav.menu.findItem(BottomNavigationPosition.BOOKINGS.id)?.isVisible = false
+
         checkBookingsTabVisibility()
     }
 
     private fun checkBookingsTabVisibility() {
+        // On large screens, Bookings is accessible from the More Menu instead of a bottom tab.
+        if (activity.isTwoPanesShouldBeUsed) {
+            return
+        }
```
hichamboushaba: I'm having some concerns with this, I'll share them in few points:

- POS uses a different class `WooPosIsScreenSizeAllowed` for determining condition to whether it's supported or not. Even though the conditions seem similar, I think we should use the same code in case it was updated.
- POS has other conditions besides the screen size, which means that if those conditions weren't met, we will end up hiding both bookings and POS.

WDYT about injecting and using `WooPosTabShouldBeVisible` directly here? Using it we'll make sure we hide bookings only when needed.

hichamboushaba (commented):
Nice work @irfano, I left two comments, the second one is important IMO.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListFragment.kt
irfano: Thanks for the suggestion.

I kept the explicit nav argument instead of relying on `findNavController().previousBackStackEntry?.destination?.id`, since back stack checks are more fragile with top-level navigation and can break with future graph/back stack changes.

I did apply the naming improvement (dbb0c64e29dd86d5f3b660cc07eed7872c846f44) to make the intent clearer and keeps the behavior explicit.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/tab/BookingsTabController.kt
irfano: Thanks, I applied your suggestion.

I removed the `isTwoPanesShouldBeUsed` condition from `BookingsTabController` and now use `WooPosTabShouldBeVisible` directly. a223a47ded99365723a7b6b39d0b98c292e83129

hichamboushaba (approved):
Nice work :shipit:

PR #15345, opened 2026-02-13
[CIAB: POS] Single collect payment button

samiuelson:
WOOMOB-2215

### Description

This PR updates WooPOS bookings payment UX by replacing separate “Pay by card” / “Pay by cash” actions with a single “Collect payment” CTA that routes into the card payment flow, with an optional switch to cash payment from within the card payment screen.

### Test Steps

Go to POS / Bookings / Booking details screen, and:
1. Verify you can collect card payment for the booking
2. Verify you can collect cash payment fo the booking

### Images/gif

Screenshot_20260213_163044
Screenshot_20260213_163105


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -230,25 +228,14 @@ class WooPosBookingsViewModel @Inject constructor(
...
-    }
-
-    private fun handlePayByCash() {
-        val details = (_state.value as? WooPosBookingsState.Content)?.selectedDetails ?: return
-        viewModelScope.launch {
-            _navigationEvent.emit(
-                WooPosNavigationEvent.OpenCashPayment(
-                    orderId = details.orderId,
-                    source = CashPaymentSource.BOOKINGS,
+                    showPayByCashButton = true,
```
malinajirka: 🔍 Should we rename this to `showCollectPaymentButton`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/root/navigation/WooPosNavigationEventHandler.kt
```diff
@@ -63,5 +64,10 @@ fun NavHostController.handleNavigationEvent(
 
         is WooPosNavigationEvent.OpenBookings ->
             navigateToBookingsScreen()
+
+        is WooPosNavigationEvent.SwitchToCashPayment -> {
+            popBackStack()
+            navigateToCashPaymentScreen(event.orderId, CashPaymentSource.BOOKINGS)
```
malinajirka: 🔍 We might want to consider including the CashPaymentSource in the navigation event - right now, it is a tiny bit confusing that a generic SwitchToCashPayment event hardcodes `Bookings`. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
```diff
@@ -93,13 +98,21 @@ private fun WooPosCardPaymentScreenContent(
 
         StateChangeAnimated(visible = state is WooPosCardPaymentState.Collecting.Preparing) {
             if (state is WooPosCardPaymentState.Collecting.Preparing) {
-                CardPaymentPreparingReader(state)
+                CardPaymentPreparingReader(
```
malinajirka, line 101: ⚠️  I believe we might want to display the cash button even in the `CardPaymentReaderDisconnected/NotConnected` case, wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -306,6 +308,13 @@ class WooPosCardPaymentViewModel @Inject constructor(
         }
     }
 
+    fun onCashPaymentClicked() {
+        cancelPayment()
+        viewModelScope.launch {
+            _navigationEvent.emit(WooPosNavigationEvent.SwitchToCashPayment(orderId))
```
malinajirka: ❓ I'm wondering if `Switch` is appropriate here - AFAIU it's the same pattern as on the totals screen => we prepare the reader for payment just to speed up the flow, not because we assume customer will want to pay with a card => so this `Pay with Cash` path is totally valid and usual, so I'm wondering if we should call it `Switch`. Perhaps, `openCashPayment` is more appropriate?

malinajirka (approved):
Thanks, I've left one important comment regarding the `reader not connected` scenario, other than that it looks great.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
samiuelson: Fixed in 393479945bd — renamed to `showCashPaymentButton` to match the existing ViewModel property and the button label ("Cash payment").

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/root/navigation/WooPosNavigationEventHandler.kt
samiuelson: Fixed in 4a02b0acff1 — added `source: CashPaymentSource` to the event and the handler now uses `event.source` instead of hardcoding `BOOKINGS`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
samiuelson: Fixed in 4971899a80c — cash payment button is now shown in the `ReaderDisconnected` state as well.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
samiuelson: Fixed in 4a02b0acff1 — renamed to `NavigateToCashPayment` which is more neutral. Also maps `CardPaymentSource` to `CashPaymentSource` in the ViewModel so the event carries the correct source.

PR #15346, opened 2026-02-13
Improve phone number validation for US addresses in shipping labels

JorgeMucientes:
## Summary
- Use US-specific phone validation (`validateUSCustomsPhone`) for US addresses in the shipping label address editor, instead of the generic `validatePhoneNumber`
- For non-US addresses, the existing generic phone validation continues to apply
- Fixes an issue where short/invalid US phone numbers (e.g., 5 digits) were accepted locally but failed server-side when purchasing a shipping label
- Add a workaround for saved custom packages that have a height of `0`: coerce the height to `0.25` when fetching shipping rates and purchasing labels, preventing the API from returning `packages[0][height] must be greater than 0`

Closes: WOOMOB-2202

### Phone validation
- [ ] Open an order that supports WooCommerce Shipping labels
- [ ] Start the shipping label purchase flow and go to **destination** address editor
- [ ] Set the country to **United States**
- [ ] Enter a short phone number (e.g., `12345`) in the phone field
- [ ] Verify a validation error appears after a brief delay (phone must be 10 digits for US)
- [ ] Enter a valid 10-digit US phone number (e.g., `1234567890`)
- [ ] Verify the validation error clears
- [ ] Change the country to a **non-US country** (e.g., Argentina)
- [ ] Enter a short phone number (e.g., `12345`)
- [ ] Verify that no US-specific validation error appears (generic validation applies)
- [ ] Complete the shipping label purchase flow and verify it succeeds with a valid phone number

https://github.com/user-attachments/assets/2981b12c-d03b-4c22-b41c-740bfd07a7e8

### Custom package height workaround
- [ ] Create a custom package template with height set to `0` (or use a store that already has one saved with height `0`)
- [ ] Open an order that supports WooCommerce Shipping labels
- [ ] Start the shipping label purchase flow and select the custom package with height `0`
- [ ] Verify that shipping rates load successfully (no `packages[0][height] must be greater than 0` error)
- [ ] Select a rate and complete the label purchase
- [ ] Verify the purchase succeeds without errors
- [ ] Repeat with a custom package that has a valid height (e.g., `5`) and verify rates still load as expected

https://github.com/user-attachments/assets/062ccf56-115c-48ed-91cb-f9624cc04298


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/packages/ui/UIModels.kt
```diff
@@ -39,6 +39,12 @@ data class PackageData(
         height = dimensionList.getOrNull(2).orEmpty().trim()
     }
 
+    val safeHeight: Double
+        get() {
+            val parsed = height.toDoubleOrNull() ?: return DEFAULT_HEIGHT
+            return if (parsed > 0) parsed else MIN_HEIGHT
```
JorgeMucientes, line 45: This is needed to avoid the API error: 
```
{"error":"rest_invalid_param","message":"Invalid parameter(s): packages","data":{"status":400,"params":{"packages":"packages[0][height] must be greater than 0"},"details":{"packages":{"code":"rest_out_of_bounds","message":"packages[0][height] must be greater than 0","data":""}}}}
```
When fetching shipping rates. It is the same workaround iOS is applying. See linear comment for further context around this: https://linear.app/a8c/issue/WOOMOB-2202/improve-phone-number-validation-and-custom-package-selection-error#comment-76f101c2

irfano (approved):
Works as expected! I have a question, though: how can you create a custom package with a height of 0? That isn’t allowed on `trunk` and this PR.

Image

JorgeMucientes:
Thanks for the review @irfano
> Works as expected! I have a question, though: how can you create a custom package with a height of 0? That isn’t allowed on trunk and this PR.

Fair question. You could create them in the web/desktop experience.

PR #15347, opened 2026-02-13
Fix SQLiteDiskIOException crash from disk storage exhaustion

JorgeMucientes:
Fixes: WOOMOB-2170

## Summary

Fixes the `SQLiteDiskIOException: disk I/O error (code 4874 SQLITE_IOERR_SHMSIZE)` crash that occurs when the device runs out of storage. The crash chain is:

1. **Unbounded log file growth** — `LogFileWriter` had no per-file size limit. On days with cascading errors, a single log file could grow indefinitely.
2. **Device storage exhaustion** — Between log files, databases, caches, and other apps, the device hits 0 bytes free. Sentry confirms `free_storage: 0`.
3. **SQLite SHM allocation failure** — Room executes `PRAGMA journal_mode` to configure WAL mode, which requires a `.db-shm` file. With no storage left, this fails with `SQLITE_IOERR_SHMSIZE` and the app crashes on launch.

This changes may also fix other related crashes we started seeing about device running out of memory and potentially related to saving too many logs on disk. 

### Changes

**1. Cap log file size and check disk space** (`LogFileWriter.kt`)
- Adds a **2 MB per-file size cap** — the size check runs inside the mutex to prevent concurrent writes from exceeding the limit.
- Adds a **disk space check** before writing — if available storage drops below 10 MB, log writes are skipped and oldest log files are proactively deleted, keeping at least the 2 most recent files for debugging.
- Disk space checks are **cached for 5 seconds** to avoid calling `StatFs` on every log write.
- `StatFs` fallback returns `0L` (treat as no space) instead of `Long.MAX_VALUE`, and ensures the directory exists before querying.
- `availableDiskBytes` is injected as a lambda to allow testing without real filesystem checks.

**2. Graceful database recovery** (`WCDatabaseModule.kt`)
- Wraps `WCAndroidDatabase.buildDb()` in a try-catch for `SQLiteDiskIOException`.
- On failure, clears log files (filtered by `log_` prefix) under `context.filesDir/logs` to free space, then retries the database open exactly once.
- The retry is also wrapped in try-catch — if it fails, the exception is logged with context before propagating (no silent crash).
- Uses `AppLog` for error logging, consistent with the rest of the fluxc-plugin module.

## Testing instructions

1. Install the app and verify logs are written normally under typical conditions (check `files/logs/` directory via Device File Explorer in Android Studio. View → Tool Windows → Device File Explorer (in some versions it is labeled Device Explorer).

Screenshot 2026-02-23 at 18 58 46

2. Fill device storage to near-zero:
   ```bash
   adb shell fallocate -l 10G /data/local/tmp/fill_disk.bin
   ```
3. Open the app and verify it does not crash on launch.
4. Verify the app writes to logcat: `"Database open failed due to disk I/O error, freeing space and retrying"`.
(I actually never saw this log despite filling the device disk completely). In any case I also wasn't able to make the DB initialization crash or the app crash at all. 
Screenshot 2026-02-23 at 19 10 56

5. Free space and verify logging resumes normally:
   ```bash
   adb shell rm /data/local/tmp/fill_disk.bin
   ```
6. Open the app again and confirm log files are written as expected under normal conditions.

### DIFF
--- a/libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
+++ b/libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
@@ -20,7 +20,12 @@ class LogFileWriter(
     private val logsDirectory: File,
     private val maxLogFiles: Int,
     private val dispatchers: CoroutineDispatchers,
-    private val availableDiskBytes: () -> Long = defaultAvailableDiskBytes(logsDirectory)
+    private val availableDiskBytes: () -> Long = {
+        if (!logsDirectory.exists()) {
+            logsDirectory.mkdirs()
+        }
+        runCatching { StatFs(logsDirectory.absolutePath).availableBytes }.getOrDefault(0L)
+    }
 ) {
     private var lastUsedFile: File? = null
     private val dateFormatter
@@ -44,7 +49,8 @@ class LogFileWriter(
         mutex.withLock {
             withContext(dispatchers.io) {
                 if (logFile.length() >= MAX_LOG_FILE_SIZE_BYTES) {
-                    return@withContext
+                    logFile.delete()
+                    logFile.createNewFile()
                 }
                 logFile.appendText("$logs\n")
             }
@@ -153,12 +159,5 @@ class LogFileWriter(
         private const val MIN_DISK_SPACE_BYTES = 10L * 1024 * 1024
         private const val DISK_SPACE_CHECK_INTERVAL_MS = 5_000L
         private const val MIN_LOG_FILES_TO_KEEP = 2
-
-        fun defaultAvailableDiskBytes(logsDirectory: File): () -> Long = {
-            if (!logsDirectory.exists()) {
-                logsDirectory.mkdirs()
-            }
-            runCatching { StatFs(logsDirectory.absolutePath).availableBytes }.getOrDefault(0L)
-        }
     }
 }
--- a/libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
+++ b/libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.util.logs
 
+import android.os.StatFs
 import com.woocommerce.android.util.CoroutineDispatchers
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -18,7 +19,10 @@ import java.util.Locale
 class LogFileWriter(
     private val logsDirectory: File,
     private val maxLogFiles: Int,
-    private val dispatchers: CoroutineDispatchers
+    private val dispatchers: CoroutineDispatchers,
+    private val availableDiskBytes: () -> Long = {
+        runCatching { StatFs(logsDirectory.absolutePath).availableBytes }.getOrDefault(Long.MAX_VALUE)
+    }
 ) {
     private var lastUsedFile: File? = null
     private val dateFormatter
@@ -27,6 +31,19 @@ class LogFileWriter(
 
     suspend fun writeLogs(logs: String) {
         val logFile = getLogFile()
+
+        val exceedsMaxSize = withContext(dispatchers.io) {
+            logFile.length() >= MAX_LOG_FILE_SIZE_BYTES
+        }
+
+        if (exceedsMaxSize) {
+            return
+        }
+
+        if (!hasEnoughDiskSpace()) {
+            deleteOldestLogFiles()
+            return
+        }
 
         mutex.withLock {
             withContext(dispatchers.io) {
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/commons/src/main/java/com/woocommerce/android/util/logs/WooFileLogger.kt
```diff
@@ -31,7 +31,8 @@ class WooFileLogger(
     private val appCoroutineScope: CoroutineScope,
     private val dispatchers: CoroutineDispatchers,
     private val processLifecycleOwner: LifecycleOwner,
-    private val crashLogging: Provider<CrashLogging>? = null
+    private val crashLogging: Provider<CrashLogging>? = null,
+    availableDiskBytes: (() -> Long)? = null
```
hichamboushaba: Given this is used only for tests of `LogFileWriter`, and we don't override it anywhere, I'd suggest removing the argument completely, and not passing `availableDiskBytes` when constructing `LogFileWriter` below, we pass it only in tests, WDYT?

libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
```diff
@@ -103,8 +119,46 @@ class LogFileWriter(
...
     companion object {
         const val LOG_FILE_NAME_PREFIX = "log_"
         const val DATE_FORMAT_PATTERN = "yyyy-MM-dd"
+
+        private const val MAX_LOG_FILE_SIZE_BYTES = 2L * 1024 * 1024
+        private const val MIN_DISK_SPACE_BYTES = 10L * 1024 * 1024
+        private const val DISK_SPACE_CHECK_INTERVAL_MS = 5_000L
+        private const val MIN_LOG_FILES_TO_KEEP = 2
+
+        fun defaultAvailableDiskBytes(logsDirectory: File): () -> Long = {
```
hichamboushaba: The `default` part was confusing at first, I assumed it's about returning some default size before getting the actual one, WDYT about removing it?

libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
```diff
@@ -18,18 +19,33 @@ import java.util.Locale
...
+            deleteOldestLogFiles()
+            return
+        }
+
         mutex.withLock {
             withContext(dispatchers.io) {
+                if (logFile.length() >= MAX_LOG_FILE_SIZE_BYTES) {
+                    return@withContext
+                }
                 logFile.appendText("$logs\n")
```
hichamboushaba, line 55: With this, we'll skip saving logs when the file is above a certain size, actually I'd prefer removing the file and recreating a new one instead, we care more about recent logs, WDYT?

hichamboushaba (approved):
Thanks @JorgeMucientes for the work on this, I'm pre-approving it 🎉 — I left some comments to check.

Also, I'm not fully convinced about the crash cause as explained in the PR description. In my experience, log files never reached 1MB, and we generally use the app more than merchants (considering we log a lot of stuff during app launch, and we open the app many times during the day). So the idea that disk space is being filled because of logs is still unconfirmed.
That said, the improvements in this PR are useful regardless, and once it's merged we can confirm whether it actually fixes the crash or if it keeps happening 🤞. If it keeps happening, I think we can just ignore it, as it's most like unrelated to the app itself.

libs/commons/src/main/java/com/woocommerce/android/util/logs/LogFileWriter.kt
JorgeMucientes: Great suggestion, applying it!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/commons/src/main/java/com/woocommerce/android/util/logs/WooFileLogger.kt
```diff
@@ -27,11 +27,11 @@ import javax.inject.Singleton
 
 @Singleton
 class WooFileLogger(
-    private val logsDirectory: File,
     private val appCoroutineScope: CoroutineScope,
     private val dispatchers: CoroutineDispatchers,
     private val processLifecycleOwner: LifecycleOwner,
-    private val crashLogging: Provider<CrashLogging>? = null
+    private val crashLogging: Provider<CrashLogging>? = null,
+    private val logFileWriter: LogFileWriter,
```
JorgeMucientes, line 34: Instead of just removing the available `availableDiskBytes` I ended making `LogFileWriter` itself an injectable constructor parameter. I think this DI pattern is cleaner. Also removed the `logsDirectory` param from WooFileLogger's primary constructor since it was only used to construct the `LogFileWriter`.

JorgeMucientes:
Thanks for the review @hichamboushaba 🙏🏼 

> I'm not fully convinced about the crash cause as explained in the PR description. In my experience

I agree, I'm not convinced either. Is sort of a tentative fix. 

> log files never reached 1MB, and we generally use the app more than merchants (considering we log a lot of stuff during app launch, and we open the app many times during the day). 

Tbh, I've never really checked the logs file size before this task. But in my experience as a developer, I tend to wipe out app's data and log in/out constantly, so I don't expect to have big logfiles persisted. 

> So the idea that disk space is being filled because of logs is still unconfirmed.

 Moreover, (and this is just a hunch) I have the feeling that in the last months we came across different crashes related to disk space being full. I don't recall having those crashes before. Of course, can't exactly tell if these crashes started when we released the app logs persistence feature. 
Anyway we'll soon see if this changes have any positive impact.

PR #15352, opened 2026-02-13
[Push Notifications] Make Login screens reusable for the Push Notifications setup flow

hichamboushaba:
Part of WOOMOB-1925

### Description
   This PR makes the WPCom login screens reusable beyond the Jetpack activation flow, so they can also be used for the push notifications setup flow.

Key changes:
- **Extracted `nav_graph_wpcom_login.xml`**: All 5 WPCom login screens (email, password, 2FA, magic link request, magic link handler) are moved into a standalone nav graph, removed from `nav_graph_jetpack_activation.xml`.
- **Introduced `WPComLoginMode` sealed interface**: Replaces the direct `JetpackStatus` nav arg with a mode-aware type (`JetpackSetup` carrying `JetpackStatus`, and `PushNotificationsSetup`), decoupling the login screens from Jetpack-specific data.
- **Dynamic start destination**: The Jetpack dispatcher uses `setStartDestination` on the nested login graph to choose the entry point (email vs password), then navigates with a single action.
- **Global action for post-login**: A global action in the Jetpack activation graph allows login screens to navigate back to `JetpackActivationMainFragment` after login completes.
- **Connected push notifications flow**: `nav_graph_push_notifications` includes the login graph, and the introduction dialog navigates to it with `PushNotificationsSetup` mode on Continue click.

### Test Steps
1. From the Jetpack installation flow (application passwords site), verify the WPCom login screens (email → password → 2FA / magic link) work as before.
2. Enable push notifications feature flag, trigger the push notifications introduction dialog from Dashboard or Settings, tap Continue, and verify it navigates to the WPCom email login screen.

### Images/gif

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes."

PR #15353, opened 2026-02-13
[Push Notifications] WPCom login email screen branding

hichamboushaba:
Part of WOOMOB-1925
Depends on #15352.

### Description
- Branch the email screen UI (header, title, subtitle, button, consent, ~~nav icon~~) based on `WPComLoginMode`.
- Rename `JetpackConsent` to `WPComConsent` with a `forJetpackSetup` parameter to support both Jetpack and generic consent text.
- Create `WordPressWooBadge` composable with overlapping WP + Woo icons, replacing the static `wp_woo.xml` vector drawable across all usages
- To be able to reuse `ic_wordpress.xml` for `WordPressWooBadge`, it's fillColor was updated to allow tinting via `Icon` composable, and add `drawableTint` to affected XML layouts.

> [!TIP]
> Please review c29e0ed39ca21a5d8bf5220ea924057d503498b1 separately, as it's not related to the business changes, it just migrates the screen to M3 components.

### Test Steps
- Make sure the PN M2 feature flag is enabled.
- Start Push Notifications setup (either from the Dashboard or Settings)
- Tap on Continue.
- Verify the branding of the email screen, and confirm the consent section works well.

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20260214_111651 | Screenshot_20260214_111642 |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionScreen.kt
```diff
@@ -54,11 +52,8 @@ fun WooPushNotificationsIntroductionScreen(
...
                 .padding(vertical = 16.dp)
         ) {
-            // Logo
-            Image(
-                painter = painterResource(id = R.drawable.wp_woo),
-                contentDescription = null,
-                modifier = Modifier.size(116.dp, 64.dp)
+            WordPressWooBadge(
+                iconSize = 64.dp
             )
```
hichamboushaba, line 57: I replaced the `wp_woo` drawable with a Composable that combines the two icons, that we already have in the code base.

My goal is to have a better dark theme support without duplicating the resources much, with this change the result is now:

| Before | After |
| --- | --- |
| Image | Image |

Let me know what you think about this? I think the white border in the icon is not correct for dark theme.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -322,6 +321,12 @@
     <string name="login_site_credentials_web_authorization_connection_rejected">Unable to login because application password creation is not approved.</string>
     <string name="login_app_login_malformed_link">We couldn\'t process your app login request</string>
     <string name="login_jetpack_connection_create_account">If you don\'t have an account, we\'ll use this email to create one.</string>
+    <string name="login_wpcom_connect_title">Connect to WordPress.com</string>
+    <string name="login_wpcom_connect_subtitle">Log in with your WordPress.com account to connect your store.</string>
+    <string name="login_wpcom_connect_create_account_hint">If you don\'t have a WordPress.com account, we\'ll use this email to create one.</string>
+    <string name="login_wpcom_connection_consent">%1$s, you agree to our &lt;a href=\'terms\'&gt;Terms of Service&lt;/a&gt; and to &lt;a href=\'sync\'&gt;share details&lt;/a&gt; with WordPress.com.</string>
```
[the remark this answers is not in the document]
hichamboushaba: This will be cleaned during the release process.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/accountmismatch/AccountMismatchErrorScreen.kt
```diff
@@ -94,22 +92,8 @@ private fun AccountMismatchErrorScreen(
...
-            )
-
-            Text(
-                text = consent,
-                style = MaterialTheme.typography.caption,
-                textAlign = TextAlign.Center,
-                color = MaterialTheme.colors.onSurface
+            WPComConsent(
+                forJetpackSetup = true
             )
```
hichamboushaba, line 97: I reused the same WPComConsent component here.

WooCommerce/src/main/res/values/wc_colors_base.xml
```diff
@@ -100,6 +100,7 @@
     <color name="jetpack_black">#0B2621</color>
 
     <color name="wp_blue">#399CE3</color>
+    <color name="wp_blue_50">#3858E9</color>
```
irfano, line 103: The Figma design uses an incorrect color (`#3E57E1`), but you used the latest WP blue here. Looks good! 👍🏻

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/WordPressWooBadge.kt
```diff
@@ -0,0 +1,70 @@
...
+    val overlapStart = iconSize * (38f / 48f)
+
+    Box(modifier = modifier) {
+        Icon(
+            painter = painterResource(id = R.drawable.ic_wordpress),
+            contentDescription = null,
+            tint = colorResource(R.color.wp_blue_50),
+            modifier = Modifier
+                .size(iconSize)
+                .align(Alignment.Companion.CenterStart)
```
irfano: np: `Companion` is redundant here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionScreen.kt
irfano: Looks good! I like the solution, and it’ll eventually mean one less drawable to maintain. (since `wp_woo.xml` was removed)

irfano (approved):
Great work, Hicham! LGTM!
Just added a np feedback for `Companion`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLoginEmailScreen.kt
```diff
@@ -164,15 +155,77 @@ fun WPComLoginEmailScreen(
     }
 }
 
-@Preview(heightDp = 130)
+private data class EmailScreenBranding(
+    @DrawableRes val navIcon: Int,
```
irfano: For consistency, I’d prefer to use the same `navIcon` for both flows. We could keep the `ic_close_24dp` button in both. WDYT?

hichamboushaba: Yes, I agree, I'll update it in all screens.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/WordPressWooBadge.kt
hichamboushaba: Good eye, I renamed the file during the implementation, and it seems the IDE added it.

PR #15354, opened 2026-02-16
[CIAB: POS] Cancel a booking from booking details

kidinov:
WOOMOB-2149

### Description

Figma - 0fzXPGPRTPyjU3ExPLW6To-fi-358_7753

Adds the ability to cancel a booking from the booking details overflow menu in POS.

- "Cancel booking" action appears in the overflow menu for bookings that are in a cancellable state (not already cancelled or completed)
- The menu item is styled in red to indicate a destructive action
- Tapping it shows a confirmation dialog with booking details (number, service, date, time, customer)
- The confirm button shows a loading spinner while the API call is in progress
- The dialog cannot be dismissed during the API call
- On success the dialog closes; on failure an error message is shown inline and the user can retry or dismiss
- Customer label in the dialog falls back to email, then phone, then "unknown customer" when the name is not available

Note: the cancel dialog does not include a close (X) button. This follows the [Material Design guideline on redundant actions](https://m3.material.io/components/dialogs/guidelines#e13b71cb-e809-4b4b-8faf-ac5414e65fa4) since the "No, keep it" button already serves the same purpose.

The loading indicator used the one from the buttons and DS as IMO there is no much profit into going into full screen just to display loading indicator, and then success screen is also redundant. @malinajirka wdyt?

### Test Steps

1. Open POS and navigate to bookings
2. Select a booking that is not cancelled or completed
3. Open the overflow menu (three dots) and tap "Cancel booking"
4. Verify the confirmation dialog shows correct booking info
5. Tap "No, keep it" and verify the dialog closes without cancelling
6. Open the dialog again, tap "Yes, cancel booking" and verify the loading state on the button
7. After the call completes, verify the dialog closes and the booking status updates
8. To test error handling: try cancelling with no network connection and verify the error message appears in the dialog

https://github.com/user-attachments/assets/5f3f3b30-8658-4efa-b878-fa436123e03b


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsUIEvent.kt
malinajirka: 📓 I'm not sure I understand why we use BookingActionClicked with BookingAction for some of the actions and standalone events for other actions. Is there some logic behind this? Should `CancelBookingConfirmed` and `CancelBookingDismissed` be children of BookingAction?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
```diff
@@ -201,58 +202,72 @@ private fun WooPosBookingsContent(
...
+                            contentDescription = stringResource(R.string.woopos_orders_empty_list_image_description)
+                        )
+                    }
                 }
             }
         }
+
+        val cancelDialog =
+            state.dialogState as? WooPosBookingsState.Content.DialogState.CancelBooking
+        WooPosCancelBookingDialog(
```
malinajirka, line 265: 🔍 We usually remove composables from the composition when they are not visible by wrapping them in `if` or `when` - it's not necessary, but would feel more consistent with the rest. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/details/WooPosCancelBookingDialog.kt
```diff
@@ -0,0 +1,117 @@
...
+    onConfirm: () -> Unit,
+    onDismiss: () -> Unit,
+) {
+    WooPosDialogWrapper(
+        modifier = modifier,
+        isVisible = isVisible,
+        dialogBackgroundContentDescription = stringResource(
+            R.string.woopos_bookings_cancel_dialog_background
+        ),
+        onDismissRequest = { if (!isProcessing) onDismiss() },
```
malinajirka: 🔍 I'd consider moving the `if not processing` logic into the VM layer.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsState.kt
```diff
@@ -111,6 +113,27 @@ sealed class WooPosBookingsState {
             data class IssueRefund(
                 val orderId: Long
             ) : DialogState()
+
+            sealed class CancelBooking : DialogState() {
+                abstract val bookingId: Long
+                abstract val message: String
+
+                data class Confirmation(
```
malinajirka: 🔍 "Confirmation" doesn't entirely convey what is happening, I'd consider renaming it to `ConfirmationShown` or `PendingConfirmation` or `Shown` (since it's dialogState) or similar. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -270,6 +277,78 @@ class WooPosBookingsViewModel @Inject constructor(
...
+
+        _state.value = currentState.copy(
+            dialogState = WooPosBookingsState.Content.DialogState.CancelBooking.Processing(
+                bookingId = dialog.bookingId,
+                message = dialog.message,
+            )
+        )
+
+        viewModelScope.launch {
+            val result = bookingsRepository.cancelBooking(bookingId)
```
malinajirka, line 327: 💡 We might want to consider adding an artificial 200ms debounce to avoid double taps and instant-error UI glitches. Wdyt?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -3923,6 +3923,15 @@
...
     <string name="woopos_bookings_details_booking_note_hint">This is a private note. It\'ll not be shared with the customer.</string>
     <string name="woopos_bookings_details_copy_email">Copy email</string>
+    <string name="woopos_bookings_cancel_menu_item">Cancel booking</string>
+    <string name="woopos_bookings_cancel_dialog_title">Cancel this booking?</string>
+    <string name="woopos_bookings_cancel_dialog_message">Booking #%1$s for %2$s on %3$s at %4$s for %5$s will be canceled.</string>
+    <string name="woopos_bookings_cancel_dialog_email_notice">The customer will be notified via email.</string>
+    <string name="woopos_bookings_cancel_dialog_confirm">Yes, cancel booking</string>
+    <string name="woopos_bookings_cancel_dialog_keep">No, keep it</string>
+    <string name="woopos_bookings_cancel_dialog_background">Cancel booking dialog background</string>
+    <string name="woopos_bookings_cancel_dialog_unknown_customer">unknown customer</string>
```
malinajirka, line 3933: Just to double check - is the first letter intentionally in lower case?

malinajirka (approved):
Thanks @kidinov!

It looks good overall, I've left a couple comments to consider. I also noticed that the back button behavior is a bit unexpected - it goes back to cart building, however, I would expect it to dismiss the dialog. Wdyt?

WooCommerce/src/main/res/values/strings.xml
kidinov: Yes! It's part of the another string

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsUIEvent.kt
kidinov: Good point! The cancel confirmed/dismissed events are dialog responses, not menu actions, so I think the separation makes sense. But I agree the naming was confusing - renamed `BookingActionClicked` to `BookingMenuActionClicked` to make it clear that it's specifically for the action menu. The dialog events (`CancelBookingConfirmed`, `CancelBookingDismissed`) stay at the top level since they're a different interaction.

ee899c32ac6

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
kidinov: Good point in general, but in this case `WooPosDialogWrapper` uses `AnimatedVisibility` internally with fade + slide transitions. It needs to stay in the composition when not visible so the exit animation can play. If we wrap in `if` and remove from composition, the dialog would disappear instantly without animation. This is the same pattern other dialogs using `WooPosDialogWrapper` follow (e.g. `WooPosSettingsProductInfoDialog`, `WooPosExitConfirmationDialog`).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/details/WooPosCancelBookingDialog.kt
kidinov: Good catch! The VM already guards this in `handleCancelDismissed()` - removed the duplicate check from the dialog layer.

cfa57cd9b46

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsState.kt
kidinov: Makes sense, renamed to `PendingConfirmation` - it better describes the state (waiting for user decision).

baef73ad44f

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
kidinov: I think the 200ms debounce is not needed here. The VM transitions to `Processing` state right away which sets the confirm button to loading and the dismiss button to disabled, so double taps are already prevented. The instant-error glitch can only happen in artificial settings like airplane mode - in real usage the network request takes enough time for the state transitions to feel natural.

kidinov:
Thanks for the review! Addressed all comments. Regarding the back button - good catch! Added a `BackHandler` that dismisses the dialog on back press, same pattern as `WooPosSearchInput` and `WooPosSettingsDetailPaneScreen`. The VM already ignores dismiss during processing, so back press won't close it mid-cancellation.

0f7a807d993

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
malinajirka: Ahh I see, thanks!

PR #15355, opened 2026-02-16
[Push Notifications] Update branding of password, 2FA, and magic link screens

hichamboushaba:
Part of WOOMOB-1925.
Depends on #15353.

### Description
- Add `wpComLoginMode` to ViewState for password, 2FA, and magic link request screens
- Branch UI elements (header, nav icon, title, subtitle, button text) based on login mode using the same `resolveBranding()` pattern from the email screen
- PushNotificationsSetup mode shows WP+Woo badge, back arrow, "Connect to WordPress.com" branding
- JetpackSetup mode preserves existing Jetpack branding

> [!NOTE]
>  The last commit (3fb3035a8fef7af1fe5fbb577c07812241296b43) is not related to the main branding changes and should be reviewed individually. It migrates the three screens from Material 2 to Material 3, following the same pattern as the email screen migration in the parent PR. 

### Test Steps
1. Enable the Push Notifications M2 feature flag.
2. Start Push Notifications setup from the dashboard or settings.
3. Check the auth flow UI, and confirm the branding matches the expected designs.

### Images/gif
| Light | Dark |
| --- | --- |
| Screenshot_20260216_123427 | Screenshot_20260216_123439 |
| Screenshot_20260216_124103 | Screenshot_20260216_124052 |
| Screenshot_20260216_124425 | Screenshot_20260216_124433 |

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15358, opened 2026-02-16
[Push Notification] Navigate to connection steps after WordPress.com login

hichamboushaba:
Part of WOOMOB-1925.
Depends on #15355.

### Description
- Implement post-login navigation for PushNotificationsSetup mode, the logic follows the same approach we use for Jetpack Setup: WPComLoginPostLoginViewModel#onLoginSuccess is invoked when login is successful, and this triggers the navigation.
- After successful WPCom login, navigate to `WooPushNotificationsConnectionStepsFragment` using a global action in `nav_graph_push_notifications`
- Add unit tests for `WPComLoginPostLoginViewModel#onLoginSuccess` covering both PushNotificationsSetup and JetpackSetup scenarios

> [!NOTE]
> Magic Link support is handled in a separate PR https://github.com/woocommerce/woocommerce-android/pull/15361 

### Test Steps
1. Enable the Push Notifications M2 feature flag.
2. Start Push Notifications setup from the dashboard or settings.
3. Sign in to your WordPress.com account. (Test regular login, Magic login is not supported yet)
4. Confirm the Push Notification steps screen is shown.

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLoginPostLoginViewModelTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLoginPostLoginViewModelTest.kt
@@ -1,0 +1,114 @@
+package com.woocommerce.android.ui.login.wpcom
+
+import androidx.lifecycle.SavedStateHandle
+import com.woocommerce.android.analytics.AnalyticsTrackerWrapper
+import com.woocommerce.android.model.JetpackConnectionStatus
+import com.woocommerce.android.model.JetpackSiteRegistrationStatus
+import com.woocommerce.android.model.JetpackStatus
+import com.woocommerce.android.tools.SelectedSite
+import com.woocommerce.android.ui.login.jetpack.GoToStore
+import com.woocommerce.android.ui.login.jetpack.JetpackActivationRepository
+import com.woocommerce.android.ui.login.wpcom.WPComLoginPostLoginViewModel.ShowJetpackActivationScreen
+import com.woocommerce.android.ui.login.wpcom.WPComLoginPostLoginViewModel.ShowJetpackCPInstallationScreen
+import com.woocommerce.android.ui.login.wpcom.WPComLoginPostLoginViewModel.ShowPushNotificationsConnectionSteps
+import com.woocommerce.android.util.captureValues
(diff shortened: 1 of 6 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLoginPostLoginViewModelTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        val events = viewModel.event.captureValues()
+
+        val result = viewModel.onLoginSuccess(WPComLoginMode.PushNotificationsSetup)
+
+        assertThat(result.isSuccess).isTrue()
+        assertThat(events.last()).isEqualTo(ShowPushNotificationsConnectionSteps)
+    }
+
+    @Test
+    fun `given jetpack setup with user not connected, when login succeeds, then show activation screen`() =
```
JorgeMucientes, line 56: I'm not sure about the overall unit tests added here. They all expect the Jetpack plugin to be installed when setting up the test. But in the new flow Jetpack plugin being installed should not be an expectation anymore. 

Also, the naming of the tests, I believe we should avoid using "Jetpack" and use "Wp.com" or something like that. 

Wdyt?

JorgeMucientes (approved):
Code looks good to me and changes work as expected. Just left some minor suggestions around testing.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLoginPostLoginViewModelTest.kt
hichamboushaba: `WPComPostLoginViewModel` is for the WPCom login flow, the flow is shared between Jetpack Setup and Push Notifications setup, the tests you are seeing are for the `JetpackSetup` flow, and that's why they mention Jetpack.

These are not related to the changes of the PR, but I saw that we didn't have tests for that part and I added them.

Does this clarify the situation?

hichamboushaba: I'm going to merge the PR, we can always change the unit tests if there are any concerns.

PR #15367, opened 2026-02-17
WOOMOB-2229: Send device locale and metadata on PN token registration

JorgeMucientes:
## Summary
- Added `device_locale` (required, `xx_XX` format) and `metadata` (optional object with `app_version`, `device_model`, `os_version`) to the Woo Core push notification token registration endpoint payload
- Introduced `PushTokenRegistrationRequest` data class in `PushNotificationsRestClient` to group registration parameters
- Collected device locale via existing `LocaleProvider` and device metadata via `android.os.Build` / `BuildConfig` in `PushNotificationRepository`
- Updated all unit tests and added a dedicated test verifying locale and metadata are passed correctly

## Test plan
1. Install the [updated plugin](https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/314c5bfe-02da-478e-9f20-ee187d673147/c0883e9d-d639-4ba8-a04b-82b7071b582e) on a test WooCommerce site
2. Build and install the app on a device/emulator with feature flag `WOO_PUSH_NOTIFICATIONS_SYSTEM` enabled
3. Log in into a Jetpack connected site. You can use WP.com creds or site creds doesn't matter, but site must be Jetpack connected. 
4. Monitor the network request to `POST /wp-json/wc-push-notifications/push-tokens`. Use app inspector.
5. Verify the request body now includes:
   - `device_locale` in `xx_XX` format (e.g., `en_US`)
   - `metadata` object with `app_version`, `device_model`, and `os_version`
  Expected request body: 
  ```
  {
  "token": "$token",
  "platform": "android",
  "origin": "com.woocommerce.android:dev",
  "device_uuid": "d1033aad-6cfdss-40d5-9059-898728f9a0f",
  "device_locale": "en_US",
  "metadata": {
    "app_version": "24.1-rc-2",
    "device_model": "Google sdk_gphone64_arm64",
    "os_version": "16"
  }
}
  ``` 
6. Change device locale in settings, re-trigger registration, and verify the locale value updates accordingly
7. Verify existing fields (`token`, `platform`, `origin`, `device_uuid`) remain unchanged

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -204,7 +204,14 @@ class PushNotificationRepository @Inject constructor(
 
     private fun getDeviceLocale(): String {
         val locale = localeProvider.provideLocale() ?: Locale.getDefault()
-        return "${locale.language}_${locale.country}".ifBlank { "en_US" }
+        val language = locale.language
+        val country = locale.country
+
+        return when {
+            language.isEmpty() -> "en_US"
+            country.isEmpty() -> language
+            else -> "${language}_${country}"
+        }
     }
 
     private fun buildDeviceMetadata(): Map<String, String> = mapOf(
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -27,6 +27,7 @@ import org.wordpress.android.fluxc.store.WooCommerceStore
 import org.wordpress.android.fluxc.store.WpComPushNotificationStore
 import org.wordpress.android.fluxc.store.WpComPushNotificationStore.SiteNotificationSetting
 import org.wordpress.android.fluxc.utils.PreferenceUtils
+import java.util.Locale
 import java.util.UUID
 import javax.inject.Inject
 
@@ -202,7 +203,7 @@ class PushNotificationRepository @Inject constructor(
     }
 
     private fun getDeviceLocale(): String {
-        val locale = localeProvider.provideLocale() ?: java.util.Locale.getDefault()
+        val locale = localeProvider.provideLocale() ?: Locale.getDefault()
         return "${locale.language}_${locale.country}".ifBlank { "en_US" }
     }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.notifications.push
 
+import android.os.Build
 import androidx.datastore.core.DataStore
 import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.edit
@@ -12,6 +13,7 @@ import com.woocommerce.android.datastore.DataStoreType.WOO_CORE_PUSH_NOTIFICATIO
 import com.woocommerce.android.extensions.isNotNullOrEmpty
 import com.woocommerce.android.extensions.orNullIfEmpty
 import com.woocommerce.android.util.WooLog
+import com.woocommerce.android.util.locale.LocaleProvider
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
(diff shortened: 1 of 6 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/bookings/note/WooPosBookingNoteViewModelTest.kt
```diff
@@ -56,7 +56,7 @@ class WooPosBookingNoteViewModelTest {
         parentId = 0L,
         personCounts = listOf(1L),
         localTimezone = "",
-        attendanceStatus = BookingEntity.AttendanceStatus.Booked,
+        attendanceStatus = BookingEntity.AttendanceStatus.Attended,
```
JorgeMucientes: Unsure how this ended up in trunk, probably a bad merge. So fixed it.

hichamboushaba: It's already fixed in `trunk` https://github.com/woocommerce/woocommerce-android/commit/13de0c7bdaed9618e638cc4046b912120d003431, if we want a clean git history maybe you can rebase your PR onto trunk, but otherwise it's fine.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsRestClient.kt
```diff
@@ -13,17 +13,19 @@ class PushNotificationsRestClient @Inject constructor(private val wooNetwork: Wo
...
             "platform" to "android",
-            "origin" to origin,
-            "device_uuid" to deviceUuid,
+            "origin" to request.origin,
+            "device_uuid" to request.deviceUuid,
+            "device_locale" to request.deviceLocale,
         )
+        if (request.metadata.isNotEmpty()) {
+            body["metadata"] = request.metadata
+        }
```
hichamboushaba: From my experience with the API (and from looking at the [code](https://github.com/woocommerce/woocommerce/blob/e599d0f7d88fc917a2890979f10cffba2005727c/plugins/woocommerce/src/Internal/PushNotifications/Validators/PushTokenValidator.php#L411-L413)), the API doesn't like receiving null `metadata`, it needs an empty body when we are not sending any data. So I would suggest sending the `metadata` always even when empty.

```suggestion
        body["metadata"] = request.metadata
```

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -194,6 +202,17 @@ class PushNotificationRepository @Inject constructor(
         return wpComPushServerId.isNotNullOrEmpty()
     }
 
+    private fun getDeviceLocale(): String {
+        val locale = localeProvider.provideLocale() ?: Locale.getDefault()
+        return "${locale.language}_${locale.country}".ifBlank { "en_US" }
```
hichamboushaba: Should we validate that `Locale#getCountry` is not empty here? The API requires a locale that exactly follows the format `xx_YY`, so we may need to offer a default if the country is empty in the device locale.
Another option would be to ask Hannah about relaxing the check a bit and accept two letter language code as a valid locale.
WDYT?

hichamboushaba (approved):
Nice work @JorgeMucientes, I left two comments, but I'm pre-approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
JorgeMucientes: Good point @hichamboushaba 
Yeah I think relaxing the requirement makes sense. Bringing this up to @hannahtinkler to get her input 🙏🏼 

In the meantime I'm applying the following change so that we ensure a well formatted locale is always passed: 
```kotlin
    private fun getDeviceLocale(): String {
        val locale = localeProvider.provideLocale() ?: Locale.getDefault()
        val language = locale.language
        val country = locale.country

        return when {
            language.isEmpty() -> "en_US"
            country.isEmpty() -> language
            else -> "${language}_${country}"
        }
    }
```

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/bookings/note/WooPosBookingNoteViewModelTest.kt
JorgeMucientes: Yeah, makes sense. I'll rebase

JorgeMucientes (commented):
Thanks for the review @hichamboushaba I've applied all your suggestion and raised the API Local requirements to Hannah.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
hichamboushaba: Sounds good @JorgeMucientes, I think `language.isEmpty` will always be false, unless the device use `Local.ROOT`, which is not expected I think, but the extra safety doesn't hurt.

PR #15339, opened 2026-02-12
POS Bookings: Add booking note support

samiuelson:
WOOPRD-2293

### Description
Add the ability to add and edit booking notes from the POS bookings detail screen. When a booking is selected, the user can tap "Add note" (or "Edit note" if one already exists) to navigate to a full-screen input where they can type and save a private booking note.

Key changes:
- New `WooPosBookingNoteScreen` with full-screen input layout, auto-focused text field, and loading/error states
- New `WooPosBookingNoteViewModel` that loads existing notes, saves via `BookingsRepository.updateNote()`, and handles success/error feedback
- Navigation wiring: `OpenBookingNote` event, route registration, and `GoBackWithResult` pattern to refresh bookings list after save
- Updated booking details note section to display existing notes inside the card and toggle button text between "Add note" / "Edit note"

### Test Steps
1. Open POS and navigate to Bookings
2. Select a booking without a note — verify the button says "Add note"
3. Tap "Add note" — verify the note input screen opens with keyboard focused
4. Type a note and tap "Send" — verify it saves and returns to the bookings list
5. Select the same booking — verify the note is displayed and the button now says "Edit note"
6. Tap "Edit note" — verify the existing note is pre-filled
7. Edit the note and save — verify the updated note displays correctly
8. Verify the note is saved in wp-admin

### Images/gif
—

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15369, opened 2026-02-17
Fix unit tests after a PR merge with outdated changes

hichamboushaba:
### Description
PR #15339 referenced an outdated value, and this broke unit tests in `trunk`, this PR simply changes the value to fix the issue.

### Test Steps
Green CI is enough.


hichamboushaba:
The issue was fixed in https://github.com/woocommerce/woocommerce-android/pull/15354

PR #15374, opened 2026-02-18
[WOOMOB-2153] Navigate from booking detail to order detail

malinajirka:
Fixes WOOMOB-2153

### Description
Adds a "View Order" action to the booking detail overflow menu that navigates to the associated order. The orders screen now supports a "single order mode" that shows only the order detail pane (without the orders list) when opened via this action.

### Test Steps
1. Open POS and navigate to a booking detail
2. Tap the overflow menu (three dots)
3. Verify "View Order" appears as a menu option
4. Tap "View Order" and confirm the order detail screen opens showing only the order detail (no list pane)
5. Verify the toolbar shows the order number
6. Press back to return to the booking detail
7. Turn on airplane mode, tap "View Order" again, and verify the error state is shown
8. Restore connectivity and tap retry — verify the order loads correctly

### Images/gif

https://github.com/user-attachments/assets/83a46914-0ede-425f-8837-6cf27cc07242


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
```diff
@@ -60,6 +60,7 @@ import com.woocommerce.android.ui.woopos.orders.WooPosOrdersUIEvent
 fun WooPosOrderDetails(
     modifier: Modifier = Modifier,
     details: WooPosOrdersState.OrderDetailsViewState.Computed.Details,
+    showOrderNumber: Boolean = true,
```
kidinov, line 63: np: instead of the showOrderNumber boolean, could we split this into two composables? For example, WooPosOrderDetails (with the order number) and WooPosOrderDetailsContent (without it). Boolean flags
  that toggle UI pieces tend to make composables harder to follow over time.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
```diff
@@ -125,49 +125,56 @@ fun WooPosOrdersListLoadingPane(modifier: Modifier = Modifier) {
 }
 
 @Composable
-fun OrderDetailsLoadingPane(modifier: Modifier = Modifier) {
+fun OrderDetailsLoadingPane(
+    modifier: Modifier = Modifier,
+    showHeader: Boolean = true
```
kidinov: Same here with showHeader - could be two composables instead of a boolean flag

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -101,6 +101,7 @@ fun WooPosOrdersScreen(
 
     WooPosOrdersScreen(
         state = state,
+        isSingleOrderMode = viewModel.isSingleOrderMode,
```
kidinov, line 104: Same suggestion for `isSingleOrderMode` - two composables would be easier to follow than a boolean toggling layout behavior.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersNavigation.kt
```diff
@@ -5,24 +5,38 @@ import androidx.compose.animation.slideOutHorizontally
...
 
 fun NavGraphBuilder.ordersScreen(
     onNavigationEvent: (WooPosNavigationEvent) -> Unit
 ) {
     composable(
         route = ORDERS_ROUTE,
+        arguments = listOf(
+            navArgument(ORDERS_ROUTE_ORDER_ID_KEY) {
+                type = NavType.LongType
+                defaultValue = 0L
```
kidinov: Using 0L as "no order" feels like a workaround - the ViewModel then needs .takeIf { it != 0L } to turn it back to null. Could we use two separate routes instead? We already have navigateToOrdersScreen  
  and navigateToOrderDetailsScreen.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
```diff
@@ -76,7 +83,36 @@ class WooPosOrdersViewModel @Inject constructor(
...
     init {
-        loadOrders()
+        if (singleOrderId != null) {
+            viewModelScope.launch { loadSingleOrderDetail(singleOrderId) }
+        } else {
+            loadOrders()
+        }
+    }
+
+    private suspend fun loadSingleOrderDetail(orderId: Long) {
```
kidinov, line 92: Since single order mode is more like a workaround for now, could we extract it into a separate class? That way, when we don't need it anymore, we just delete one class instead of   
  cleaning up branches all over the ViewModel

kidinov (approved):
LGTM! I left a couple of comments.

Also, I noticed that the loading state looks a bit off. Notice that when we load order details in the historical orders screen, we display shimmer in the titles/buttons at the top


Image

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
malinajirka: While I agree boolean flags in composables can bite us, in this case, splitting it feels like an overkill in this case - we toggle just a small title. I'll leave it as is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
malinajirka: I've removed it completely - feels like the change is not worth the complexity.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersViewModel.kt
malinajirka: I tried to keep it as encapsulated as possible - essentially a single method, but since we are touching internal state, I don't think it can be easily extracted. Having said that, it's all wrapped in `if singleOrderId` so it'll be as simple as removing a feature flag.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
malinajirka: I'm not sure I follow - you mean duplicating the whole layout?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersNavigation.kt
malinajirka: Good call - fixed in 78874a2c5ed55ff22f97575c64b43f20222e4b17

malinajirka:
Thanks for the review! I've implemented most of the suggestions.

> Also, I noticed that the loading state looks a bit off. Notice that when we load order details in the historical orders screen, we display shimmer in the titles/buttons at the top

Good catch, fixed in https://github.com/woocommerce/woocommerce-android/pull/15374/commits/809d9de3903a1df50dd67fc8cc020f83cef84fd8 and  https://github.com/woocommerce/woocommerce-android/pull/15374/commits/61d581cb74eafdcd94de0a7e734ab0c5b07aace4.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
kidinov: No, just excact common part
AI explains the idea:
1. Extract OrderDetailsPane as a standalone composable

  The right side (lines 255-293 in OrdersContent) is already basically its own thing — pull it out:

```
  @Composable
  private fun OrderDetailsPane(
      state: WooPosOrdersState.Content,
      onUIEvent: (WooPosOrdersUIEvent) -> Unit,
      showOrderNumber: Boolean = true,
  ) {
      Box(modifier = modifier.background(MaterialTheme.colorScheme.surface)) {
          when {
              state.selectedDetails != null -> {
                  WooPosOrderDetails(
                      modifier = Modifier.fillMaxHeight(),
                      details = state.selectedDetails,
                      showOrderNumber = showOrderNumber,
                      onUIEvent = onUIEvent
                  )
              }
              state.items is WooPosOrdersState.Content.Items.Searching -> {
                  OrderDetailsLoadingPane(...)
              }
              else -> {
                  WooPosEmptyScreen(...)
              }
          }
      }
  }
```

  2. Replace OrdersContent with two separate composables

  Instead of one OrdersContent with isSingleOrderMode:


```
  @Composable
  private fun OrdersListWithDetails(
      state: WooPosOrdersState.Content,
      scrollToTopEvent: SharedFlow<Unit>,
      onRefresh: () -> Unit,
      onOrderSelected: (Long) -> Unit,
      onEndOfOrdersListReached: () -> Unit,
      onPaginationErrorTryAgain: () -> Unit,
      onSearchEvent: (WooPosSearchUIEvent) -> Unit,
      onSearchErrorRetry: () -> Unit,
      onUIEvent: (WooPosOrdersUIEvent) -> Unit
  ) {
      Row(modifier = Modifier.fillMaxSize()) {
          OrdersListPane(
              ...,
              modifier = Modifier.weight(0.3f).fillMaxHeight()...
          )
          OrderDetailsPane(
              state = state,
              onUIEvent = onUIEvent,
              showOrderNumber = true,
              modifier = Modifier.weight(0.7f)
          )
      }
  }

  @Composable
  private fun SingleOrderDetails(
      state: WooPosOrdersState.Content,
      onUIEvent: (WooPosOrdersUIEvent) -> Unit
  ) {
      OrderDetailsPane(
          state = state,
          onUIEvent = onUIEvent,
          showOrderNumber = false,
          modifier = Modifier.fillMaxSize()
      )
  }

```

  3. Simplify the parent WooPosOrdersScreen

  The when(state) block becomes cleaner too — in the Loading branch you just check isSingleOrderMode directly (that part is already simple). But the Content branch calls either OrdersListWithDetails or SingleOrderDetails instead of forwarding the boolean.

  Why this is better than the boolean flag

  - Each composable does one thing and is easy to read on its own
  - No if (isSingleOrderMode) scattered through the layout code
  - The shared OrderDetailsPane is reused without duplication
  - If the two modes diverge further in the future, they can evolve independently

  The toolbar title logic (lines 191-197) can stay in the parent—that's just a string decision, not layout duplication.

malinajirka: I see - I'll try that in a separate PR, thanks!

PR #15376, opened 2026-02-18
[CIAB: POS] Show collect payment button for cancelled bookings

kidinov:
Closes WOOMOB-2225

### Description

**Cancelled bookings: show collect payment button when unpaid**

Previously, the `buildPaymentSection` method treated cancelled bookings the same as paid/complete bookings — all three statuses were grouped together to determine `isPaid = true`. This meant that the "Collect payment" button was always hidden for cancelled bookings, even when they were never actually paid.

The fix introduces `isBookingPaid()`, which checks the actual `PaymentStatus` (resolved from the order) for cancelled bookings. If the booking was paid before cancellation, `isPaid` remains true and the button stays hidden. If it was never paid (status `FAILED` or `UNPAID`), `isPaid` is false and the "Collect payment" button is shown.

**Card payment screen: standardise button layout**

Refactored the card payment screen composables (`CardPaymentSuccess`, `CardPaymentReaderDisconnected`, `CardPaymentFailed`) to use a consistent layout pattern: content is vertically centered with weighted spacers, and buttons are bottom-aligned with uniform sizing (604dp width, 80dp height). Replaced `ConstraintLayout` in the success state and nested `Box`/`Column` in the disconnected state with simpler `Column`-based layouts. Added compose previews for the main screen states.

### Test Steps

1. Open POS Bookings
2. Select a cancelled booking that was **not paid** -> verify the "Collect payment" button is visible
3. Select a cancelled booking that was **paid** -> verify the "Collect payment" button is hidden
4. Verify the attendance section is still hidden for cancelled bookings
5. Tap "Collect payment" on a cancelled unpaid booking -> verify the card payment screen shows buttons aligned to the bottom with consistent sizing

image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/CardReaderPaymentCollectibilityChecker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/CardReaderPaymentCollectibilityChecker.kt
@@ -21,11 +21,11 @@ class CardReaderPaymentCollectibilityChecker @Inject constructor(
     private val cardReaderPaymentCurrencySupportedChecker: CardReaderPaymentCurrencySupportedChecker,
     private val ciabSiteGateKeeper: CIABSiteGateKeeper
 ) {
-    suspend fun isCollectable(order: Order): Boolean {
+    suspend fun isCollectable(order: Order, allowCancelledStatus: Boolean = false): Boolean {
         return ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.WooPayments) &&
             with(order) {
                 cardReaderPaymentCurrencySupportedChecker.isCurrencySupported(currency) &&
-                    isStatusCollectable() &&
+                    isStatusCollectable(allowCancelledStatus) &&
                     !isOrderPaid &&
                     order.total.compareTo(BigDecimal.ZERO) == 1 &&
                     BigDecimal.ZERO.compareTo(order.refundTotal) == 0 &&
@@ -45,11 +45,11 @@ class CardReaderPaymentCollectibilityChecker @Inject constructor(
             WOOCOMMERCE_BOOKINGS_PAYMENT_TYPE,
         )
 
-    private fun Order.isStatusCollectable() = status in arrayOf(
+    private fun Order.isStatusCollectable(allowCancelledStatus: Boolean) = status in arrayOf(
         Pending,
         Processing,
         OnHold,
         Custom(Order.Status.AUTO_DRAFT),
         Failed,
-    )
+    ) || (allowCancelledStatus && status == Order.Status.Cancelled)
 }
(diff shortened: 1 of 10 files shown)


samiuelson (commented):
Hi @kidinov! My experience is that when a booking is cancelled the corresponding order gets updated with `cancelled` status as well. For such status, card-present payment is always failing, primarily due to our internal conditions in `CardReaderPaymentCollectibilityChecker` that result in payment error being emitted. In such case, showing Collect Payment button doesn't make sense.

I believe we should update `CardReaderPaymentCollectibilityChecker` and test if backend accepts card payment for `cancelled` order, or update the order status before collecting payment. Or hide card payment option and allow cash payment only. 

Let me know what you think.

kidinov:
@samiuelson I enabled canceled status be collectable in the cases of the booking orders. Also tested and the payments go through. Wdyt?

02be253fe4a2a031d79e55223a9101865971b5de

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
```diff
@@ -129,14 +129,22 @@ class WooPosBookingViewStateMapper @Inject constructor(
         return WooPosBookingsState.AttendanceSection(selection = selection)
     }
 
+    private fun isBookingPaid(
+        bookingStatus: BookingEntity.Status,
+        paymentStatus: PaymentStatus,
+    ): Boolean = when (bookingStatus) {
+        BookingEntity.Status.Paid, BookingEntity.Status.Complete -> true
+        BookingEntity.Status.Cancelled -> paymentStatus == PaymentStatus.PAID
```
samiuelson, line 142: ❓ Should we show `Collect payment` button also in case of `BookingEntity.Status.Cancelled &&  PaymentStatus.Refunded`? 🤔

samiuelson (approved):
LGTM! I left one question which I believe is worth testing–in case the backend allows collecting card payment for a booking-order which is cancelled-refunded, let's keep it, if not, let's better hide the collect payment button in such scenario.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
kidinov: I believe displaying it for Cancelled + Refunded doesn’t make sense. If a booking was paid and then refunded, the merchant has already received the money and returned it. From a business perspective, collecting payment again for a Cancelled + Refunded booking would be unusual because it was intentionally refunded. Feels like if will be a new booking in this case if a customer wants to get the service once again. Wdyt?

samiuelson: Current implementation keeps showing the button in case the booking have been refunded. I've created issue to hide it in such case: WOOMOB-2285

PR #15377, opened 2026-02-18
Fix WebView having 0 height in WCWebView composable

hichamboushaba:
Closes: WOOMOB-2251

### Description
The cause of the issue is that the `AndroidView` hosting the `WebView` inside `WCWebView` does end up with 0 height, I don't have an explanation on why this is happening now and not before. The fix I applied is forcing the `AndroidView` to `fillMaxSize` of the `Box` container.

### Test Steps
1. Use a CIAB site.
2. Open a bookable product.
3. Verify the WebView fills the available space and content is visible
4. Test with both authenticated and non-authenticated WebView screens

### Images/gif

https://github.com/user-attachments/assets/2ab8b995-8c80-4b38-af0b-76e216718b53

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/web/WCWebView.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/web/WCWebView.kt
@@ -7,6 +7,7 @@ import android.widget.FrameLayout
 import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.CircularProgressIndicator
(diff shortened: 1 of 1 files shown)


JorgeMucientes (approved):
Thanks for quickly addressing this @hichamboushaba. I think the reason this bug didn't show up before and we noticed about it now is because it's something related to CIAB urls. It's very strange (given what the fix was) but seem like the url issues affect only authenticated urls from CIAB sites. I tested loading several other urls from `trunk` for non CIAB sites and everything worked fine: 

https://github.com/user-attachments/assets/441e4669-6304-4c95-91e1-03da10e9e83d

hichamboushaba:
> Thanks for quickly addressing this @hichamboushaba. I think the reason this bug didn't show up before and we noticed about it now is because it's something related to CIAB urls. It's very strange (given what the fix was) but seem like the url issues affect only authenticated urls from CIAB sites. I tested loading several other urls from `trunk` for non CIAB sites and everything worked fine:

Yes, I agree, what I meant above is I don't understand what changed in the CIAB admin to make it not work now, as it was working before.

PR #15378, opened 2026-02-18
[CIAB: POS] Make collect payment button sticky in POS booking details

kidinov:
Closes: WOOMOB-2223

### Description

Makes the "Collect payment" button in POS booking details sticky at the bottom of the screen. The button now floats on a white panel with a top divider so it's always visible regardless of scroll position. As the user scrolls to the bottom, the panel background and divider fade out smoothly.

### Test Steps

1. Open POS bookings and select an unpaid booking
2. Scroll through the booking details — the collect payment button should stay fixed at the bottom on a white panel with a divider
3. Scroll all the way to the bottom — the white panel and divider should fade out
4. Verify the button works correctly when tapped

https://github.com/user-attachments/assets/0230e526-1fec-4c84-b5aa-22a1a88043fc


samiuelson (approved):
LGTM!

PR #15380, opened 2026-02-18
[CIAB: POS] Payment error event handling

samiuelson:
Closes WOOMOB-2215

### Description

Previously, the `when` block handling `CardReaderPaymentController` events used an `else -> Unit` catch-all. This meant `ShowPaymentErrorMessage` was silently dropped — when the controller emitted this event, the error was lost and the user saw nothing. Other unsupported events (`InteracRefundSuccessful`, `ContactSupportTapped`, etc.) were also silently swallowed, hiding potential integration issues.

This PR:

- **Handles `ShowPaymentErrorMessage`** — now sets state to `PaymentFailed` so the user actually sees the error.
- **Changes `ShowErrorMessage` button** from "Try another payment method" to "Go back", since there is no retry mechanism for these error types.
- **Removes the `else -> Unit` catch-all** — `Exit` and `PlaySuccessfulPaymentSound` are now explicitly listed as no-ops, and unsupported events throw `IllegalArgumentException` to surface integration issues early.
- **Centers the `PaymentFailed` UI** vertically for better presentation.
- Adds `woo_pos_payment_failed_go_back` string resource.

### Test Steps

1. Open POS, add an item to the cart, and tap "Collect payment"
2. Trigger a card payment failure (e.g., use a test card that declines or cancelled booking)
3. Verify the error screen is shown with the error message and a single "Go back" button (no "Try another payment method" or "Go back to checkout" button)
4. Verify the error content is vertically centered
5. Tap "Go back" -> verify you return to the previous screen

### Images/gif

N/A


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModelTest.kt
```diff
@@ -310,20 +310,22 @@ class WooPosCardPaymentViewModelTest {
...
 
-        assertThat(viewModel.state.value)
-            .isInstanceOf(WooPosCardPaymentState.PaymentFailed::class.java)
-    }
+            val state = viewModel.state.value
+            assertThat(state).isInstanceOf(WooPosCardPaymentState.PaymentFailed::class.java)
+            assertThat((state as WooPosCardPaymentState.PaymentFailed).isDismissButtonVisible).isFalse()
+        }
 
     @Test
```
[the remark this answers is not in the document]
samiuelson: added in [00f9498](https://github.com/woocommerce/woocommerce-android/pull/15380/commits/00f9498a4b2a2137ac640338af78efd220dd4960)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -192,13 +192,36 @@ class WooPosCardPaymentViewModel @Inject constructor(
...
+                            title = resourceProvider.getString(
+                                R.string.woopos_success_totals_payment_failed_title
+                            ),
+                            subtitle = resourceProvider.getString(event.message),
+                            retryButtonLabel = resourceProvider.getString(
+                                R.string.woo_pos_payment_failed_go_back
                             ),
                             isDismissButtonVisible = false
                         )
                     }
```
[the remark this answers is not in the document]
samiuelson: cleaned up in [fdeceaf](https://github.com/woocommerce/woocommerce-android/pull/15380/commits/fdeceafc3f2f738339cc643ace2c8673e3a0767c)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -185,20 +185,36 @@ class WooPosCardPaymentViewModel @Inject constructor(
...
+                        }
                         _state.value = WooPosCardPaymentState.PaymentFailed(
                             title = resourceProvider.getString(
                                 R.string.woopos_success_totals_payment_failed_title
                             ),
-                            subtitle = resourceProvider.getString(event.message),
+                            subtitle = resourceProvider.getString(messageRes),
                             retryButtonLabel = resourceProvider.getString(
-                                R.string.woo_pos_payment_failed_try_another_payment_method
+                                R.string.woo_pos_payment_failed_go_back
```
kidinov: Does it need to be called `retryButtonLabel` if it also can be "go back"? Maybe something like "actionButtonLabel"?

kidinov: It's crashing - but there is an issue on this already. Maybe worth fixing here

kidinov (approved):
LGTM! Please take a look into the comment I made

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
samiuelson: good idea; renamed in 4f4dfc7edd24501b4f584b0a2629841e6dc7c759

samiuelson: > It's crashing - but there is an issue on this already. Maybe worth fixing here

Fixed in 0802c7fa04e5f514d08f4ada4f21a8a6b9954f5d

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
```diff
@@ -136,8 +136,7 @@ class WooPosBookingViewStateMapper @Inject constructor(
         val paymentInfo = booking.order.paymentInfo
         val currency = booking.currency
         val isPaid = booking.status == BookingEntity.Status.Paid ||
-            booking.status == BookingEntity.Status.Complete ||
-            booking.status == BookingEntity.Status.Cancelled
+            booking.status == BookingEntity.Status.Complete
```
kidinov: That's against another PR - https://github.com/woocommerce/woocommerce-android/pull/15376. Do we need this change?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
```diff
@@ -420,13 +420,15 @@ private fun CardPaymentFailed(
...
         )
         Spacer(modifier = Modifier.height(WooPosSpacing.XLarge.value))
-        WooPosButton(
-            text = state.retryButtonLabel,
-            modifier = Modifier
-                .height(80.dp)
-                .width(604.dp),
-            onClick = onRetryClicked,
-        )
+        if (state.actionButtonLabel != null) {
```
kidinov, line 441: I think the button is better to be aligned at the bottom + bottom margin to be consistent with other buttons.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
samiuelson: Reverted in c3c7cf6e3e67c60b3db9d22e69d6d23c3d047d07

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
samiuelson: Done in 510b3b5bc3abedb10ad138ba78c21c9da57d2ab5

PR #15381, opened 2026-02-18
[CIAB:POS] Add team member avatar to POS bookings list

kidinov:
Closes WOOMOB-2222
Closes WOOMOB-2228

### Description

Adds a team member avatar to each booking card in the POS bookings list. The avatar is a 24dp circle placed in the top-right corner of the card. When the team member has a photo, it loads the image via Coil. When there's no photo (or while loading), it shows the team member's initials on a gray background. A light border is shown around the avatar only when the image is loaded.

The data comes from the v2 team-members API (`/wc-bookings/v2/resources/team-members/`) which was already being fetched. The `resource_id` on each booking links to a team member, and we now pass the full resource through to the list item mapper.

### Test Steps

1. Open POS with a site that has bookings with v2 team members assigned
2. Navigate to the Bookings tab
3. Verify each booking card shows the team member avatar in the top-right
4. For team members with a photo URL, verify the photo loads inside the circle with a border
5. For team members without a photo, verify initials are shown (first letters of first and last name) without a border

**Note:** Currently you can create team members via the CIAB UI, but it looks like you can not assign them to bookings through the UI. To test this you need to either use the API directly and assign team members to a bookng or I can invite you to my test site.

### Images/gif

https://github.com/user-attachments/assets/375d15a7-1d02-4a5e-a885-5948fa9ce141


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
```diff
@@ -206,4 +209,18 @@ class WooPosBookingViewStateMapper @Inject constructor(
             is BookingEntity.AttendanceStatus.Unknown -> WooPosBookingsState.AttendanceState.UNATTENDED
         }
     }
+
+    private fun mapTeamMember(
```
malinajirka, line 226: 🔍 Might be worth covering with tests.

malinajirka (approved):
LGTM! I've left one minor comment.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
kidinov: a7c32b94298d0d70e9961de5e61eb3873ee82011 Done!

PR #15382, opened 2026-02-18
Configure repo for AI agents + extended Claude Code support

JorgeMucientes:
### Description

Adds comprehensive AI agent configuration for the woocommerce-android repo, encoding architecture boundaries, conventions, and workflows so AI assistants work consistently with how the team builds.

This follows the [Automattic Coding Agents Guidelines](https://fieldguide.automattic.com/coding-agents-guidelines/) and mirrors the iOS initiative ([PR #16682](https://github.com/woocommerce/woocommerce-ios/pull/16682)).

**Agent-agnostic (benefits all AI tools):**
- Expanded `AGENTS.md` (~250 lines) with full repo context: layout, architecture diagram, MVVM patterns with code examples, build/test commands, Kotlin/Compose conventions, analytics, git conventions, common pitfalls, and skills routing table
- `.agents/skills/` with 3 skills: review, pr, verify-on-device
- `.mcp.json` with [mobile-mcp](https://github.com/mobile-next/mobile-mcp) for Android emulator interaction and [context-a8c](https://www.npmjs.com/package/@automattic/mcp-context-a8c) for Automattic internal context (Slack, P2s, Field Guide, etc.)
- `.aiexclude` to keep build artifacts (except reports, outputs, and test-results), secrets, and Gradle cache out of agent context

**Claude Code specific:**
- `.claude/settings.json` — conservative guardrails: read-only git commands + add/commit/push allowed, destructive ops (merge, rebase, force push, push to trunk) denied
- `.claude/agents/` — code-reviewer (opus), debugger (sonnet), test-writer (opus)
- `.claude/skills` — symlink to `.agents/skills/` so Claude Code auto-discovers them

**verify-on-device skill (rewritten based on live testing):**
- Tested the full workflow end-to-end on an Android emulator (build → install → launch → navigate → screenshot)
- Identified and fixed reliability gaps from the mobile-mcp documentation:
  - Added critical rule to always use the accessibility tree (`mobile_list_elements_on_screen`) for tap coordinates — screenshots are scaled down and produce systematically wrong coordinates
  - Documented all 19 mobile-mcp tools (was only 10) with WooCommerce-specific use cases
  - Added WooCommerce navigation reference with stable element identifiers for bottom nav tabs and key screens
  - Added common navigation patterns (go back, scroll, pull to refresh, open list items)
  - Added error recovery table covering 10 real failure modes (SDK not found, wrong taps, app crashes, login screens, missing devices, recording issues)
- Added **ADB screen recording** support: start/stop recording commands, background recording during navigation, chained recordings for flows longer than 3 minutes, and failure mode documentation

**Changes from review feedback:**
- Removed duplicate `pr_template.md` — the PR skill already references `.github/PULL_REQUEST_TEMPLATE.md` directly
- Removed `debug`, `lint`, and `test` skills — redundant with AGENTS.md which already documents build/test/detekt commands
- Tightened `.claude/settings.json` — removed merge/rebase from allow list, added deny for push to trunk
- Expanded code-reviewer agent with WP/Woo API review criteria, Scope section, and detailed review dimensions (correctness, architecture, security, testing, code quality)
- Expanded test-writer agent with core principles (happy path defaults in @Before, one behavior per test, don't mock data classes), mock configuration patterns, good/bad examples, and stricter naming guidance
- PR skill now enforces `[WOOMOB-XYZ]` title prefix and `Fixes WOOMOB-XYZ` in description
- Review skill prioritizes branch diff (pre-PR review) over uncommitted/staged changes
- Softened "no comments" rule to "comments only when they explain business logic not captured in the code"
- `.aiexclude` refined: removed `.buildkite/`, `.github/`, `fastlane/` exclusions; added negation patterns to keep `build/reports/`, `build/outputs/`, and `build/test-results/` accessible
- `.gitignore` uses generic glob patterns for agent local file overrides (Claude, Codex, etc.)

### How to leverage this setup

1. **Any AI tool** reads `AGENTS.md` automatically for full project context.
2. **Claude Code users** get additional workflow automation:
   - `/review` — review changes against project rules (branch diff, staged, or unstaged)
   - `/pr` — create a PR following the template with `[WOOMOB-XYZ]` prefix
   - `/verify-on-device` — build, install, and verify on an Android emulator (with optional screen recording)
3. **mobile-mcp** enables emulator interaction (screenshots, taps, swipes, text input). Prerequisites: Node.js v22+, Android SDK, a running emulator.
4. **context-a8c** enables access to Automattic internal context (Slack, P2s, Field Guide, etc.).
5. **Subagents** are available for delegation: code-reviewer, debugger, test-writer.

### Test Steps

- [ ] Open a new Claude Code session in the repo
- [ ] Verify `AGENTS.md` context is loaded (ask Claude about the architecture)
- [ ] Verify skills appear in the `/` menu (review, pr, verify-on-device)
- [ ] Run `/review` and confirm it diffs against trunk
- [ ] With an emulator running, run `/verify-on-device` and confirm the agent:
  - Discovers the emulator
  - Builds and installs the APK
  - Launches the app and navigates using `mobile_list_elements_on_screen` (not screenshot coordinates)
  - Takes and saves verification screenshots
  - (Optional) Records a screen recording via ADB and pulls the MP4 to the host

### Images/gif

N/A — configuration only, no UI changes.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes. → [Internal]


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>MyClassTest</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>MyFeatureViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes:
### Demo: AI Agent Emulator Verification with Screen Recording

I told Claude Code from this branch: 

> Can you make a screen recording of the WooCommerce app navigating through all the different bottom tabs and opening an order detail and a product detail? Once done update the screen recording to the PR as a comment

So this comment was generated by AI partially. I had to manually upload the .mp4 file given Github doesn't support uploading files via API? AI was extremely slow at navigating through the app though, so we probably want to look into polishing how it interacts with the UI. But it's still pretty nice: 

https://github.com/user-attachments/assets/02c9bdc4-4844-4dfb-aa92-50d07955ca38

PS: given it was taking really long I asked AI to stop. But it was enough to verify the skill works as expected.

.agents/skills/pr/references/pr_template.md
```diff
@@ -0,0 +1,10 @@
+### Description
```
hichamboushaba: I'm not sure this is used anywhere, the SKILL correctly references the `.github/PULL_REQUEST_TEMPLATE.md`, so we don't need to duplicate the template.

.claude/settings.json
```diff
@@ -0,0 +1,39 @@
...
+      "Bash(git diff *)",
+      "Bash(git log *)",
+      "Bash(git show *)",
+      "Bash(git branch *)",
+      "Bash(git checkout *)",
+      "Bash(git switch *)",
+      "Bash(git stash *)",
+      "Bash(git fetch *)",
+      "Bash(git merge *)",
+      "Bash(git rebase *)",
```
hichamboushaba: I'm not sure allowing these commands is always safe, a rebase could be destructive in some scenarios, so I'd prefer to keep these preferences for each of us to handle in their own session, personally I keep many of these commands as `ask`.

I'd keep only the read-only commands here: status, diff, log...

.agents/skills/pr/references/pr_template.md
ParaskP7, line 1: Suggestion (💡): Can't this skill just read from the existing [PULL_REQUEST_TEMPLATE.md](https://github.com/woocommerce/woocommerce-android/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md) from within the `.github` folder itself, just to avoid unnecessary duplication, and then us maintaining both files? 🤔

.claude/settings.json
```diff
@@ -0,0 +1,39 @@
...
+      "Bash(git push *)",
+      "Bash(gh pr *)",
+      "Bash(gh issue *)",
+      "Bash(gh api *)",
+      "Bash(adb *)",
+      "Bash(emulator *)"
+    ],
+    "deny": [
+      "Bash(git push --force *)",
+      "Bash(git push -f *)",
```
ParaskP7, line 28: Suggestion (💡): Maybe a specific `push` to `trunk` should be disallowed as well. 🤔 (we do have such repository rules as well, but maybe worth having that there explicitly as well, just in case)

.aiexclude
```diff
@@ -0,0 +1,12 @@
+.buildkite/
+.configure-files/
+.github/
+.gradle/
+build/
+*/build/
```
ParaskP7, line 5: Question (❓): I wonder if excluding these `build` folders will do us any good, or else, make agents less effective in their (overall) analysis of (at least) build errors, tests, reports, etc... 🤔 (maybe being explicit about files we want to exclude is a better alternative, sensitive files that could potentially be available within these `build` folders)

PS.1: Same of ☝️ for the `.gradle/` folder.
PS.2: I am also skeptical about the `.buildkite/` and `.github/` folders as well. I wouldn't exclude them, why would we, there isn't anything sensitive in there, is it? 🤔
PS.3: Same of ☝️ for `fastlane/`.

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,51 @@
+---
+name: code-reviewer
+description: Reviews code changes for architecture compliance, Kotlin/Compose style, testing, analytics, and PR conventions
+model: opus
+tools: Read, Grep, Glob, Bash
+---
+
+You are a code reviewer for the WooCommerce Android project. Review the current changes against the project's architecture and conventions.
```
hichamboushaba: Overall it's good, I just think it might be useful to mentione knowledge about WordPress and WooCommerce APIs, currently I was experimenting with this:

```
You are an expert code reviewer with deep knowledge of software engineering best practices, design patterns, and the Android ecosystem. You have particular expertise in Kotlin, Jetpack Compose, Hilt, Kotlin Coroutines, and Room, and you have strong knowledge of WordPress and WooCommerce APIs. Your role is to analyze code critically but constructively, helping developers improve their work.
```

And I have this section below:
```
### WooCommerce/WordPress API (when applicable)
- Correct endpoint usage and HTTP methods
- Proper authentication handling
- Pagination handling
- Error response handling and edge cases
```
---- 

Another note, I would suggest adding a `Scope` section just to set some guardrails for the agent:

```
## Scope

You review **recently written or modified code only**, not the entire codebase. Focus exclusively on the specific changes, diffs, or files that were just created or altered.
```

hichamboushaba: I'm sharing below the full agent I use and that I mentioned in Slack:

<details>
<summary>code-reviewer.md</summary>

```markdown
---
name: code-reviewer
description: "Use this agent when code changes have been implemented and need to be reviewed for correctness, style, adherence to project conventions, and potential issues. This agent should be launched after any non-trivial code modification to catch bugs, improve quality, and ensure best practices are followed.\\n\\nExamples:\\n\\n- Example 1:\\n  Context: The user asked to implement a new feature and code was just written.\\n  user: \"Add a product variation bulk editing screen\"\\n  assistant: \"Here is the implementation with the new VariationBulkEditFragment, ViewModel, and Compose UI.\"\\n  <code changes written>\\n  assistant: \"Now let me use the code-reviewer agent to review all the changes I just made.\"\\n  <launches code-reviewer agent via Task tool>\\n\\n- Example 2:\\n  Context: The user asked for a bug fix and the assistant just applied it.\\n  user: \"Fix the crash when navigating back from the order detail screen\"\\n  assistant: \"I've updated the lifecycle handling in the OrderDetailFragment.\"\\n  <code changes written>\\n  assistant: \"Let me launch the code-reviewer agent to verify the fix is correct and doesn't introduce any issues.\"\\n  <launches code-reviewer agent via Task tool>\\n\\n- Example 3:\\n  Context: The user asked to refactor existing code.\\n  user: \"Migrate the product list screen to Jetpack Compose\"\\n  assistant: \"I've migrated the ProductListFragment to use ComposeView and created the Compose UI.\"\\n  <code changes written>\\n  assistant: \"I'll now use the code-reviewer agent to review the migrated code for correctness and consistency.\"\\n  <launches code-reviewer agent via Task tool>\\n\\n- Example 4:\\n  Context: A WooCommerce API integration was just implemented.\\n  user: \"Add WooCommerce order sync functionality\"\\n  assistant: \"Here's the implementation with the OrderSyncRepository and API client.\"\\n  <code changes written>\\n  assistant: \"Let me launch the code-reviewer agent to review the API integration code.\"\\n  <launches code-reviewer agent via Task tool>"
tools: Glob, Grep, Read, WebFetch, WebSearch, ListMcpResourcesTool, ReadMcpResourceTool, mcp__github__add_comment_to_pending_review, mcp__github__add_issue_comment, mcp__github__assign_copilot_to_issue, mcp__github__create_branch, mcp__github__create_or_update_file, mcp__github__create_pull_request, mcp__github__create_repository, mcp__github__delete_file, mcp__github__fork_repository, mcp__github__get_commit, mcp__github__get_file_contents, mcp__github__get_label, mcp__github__get_latest_release, mcp__github__get_me, mcp__github__get_release_by_tag, mcp__github__get_tag, mcp__github__get_team_members, mcp__github__get_teams, mcp__github__issue_read, mcp__github__issue_write, mcp__github__list_branches, mcp__github__list_commits, mcp__github__list_issue_types, mcp__github__list_issues, mcp__github__list_pull_requests, mcp__github__list_releases, mcp__github__list_tags, mcp__github__merge_pull_request, mcp__github__pull_request_read, mcp__github__pull_request_review_write, mcp__github__push_files, mcp__github__request_copilot_review, mcp__github__search_code, mcp__github__search_issues, mcp__github__search_pull_requests, mcp__github__search_repositories, mcp__github__search_users, mcp__github__sub_issue_write, mcp__github__update_pull_request, mcp__github__update_pull_request_branch, mcp__plugin_context-a8c_context-a8c__context-a8c-load-provider, mcp__plugin_context-a8c_context-a8c__context-a8c-execute-tool
model: sonnet
color: green
---

You are an expert code reviewer with deep knowledge of software engineering best practices, design patterns, and the Android ecosystem. You have particular expertise in Kotlin, Jetpack Compose, Hilt, Kotlin Coroutines, and Room, and you have strong knowledge of WordPress and WooCommerce APIs. Your role is to analyze code critically but constructively, helping developers improve their work.

## Scope

You review **recently written or modified code only**, not the entire codebase. Focus exclusively on the specific changes, diffs, or files that were just created or altered.

## Review Process

1. **Gather the Changes**: Start by running `git diff HEAD` (or `git diff` if changes are unstaged) to see what was modified. If specific files were mentioned, examine those files directly. If the diff is empty, check `git log --oneline -5` to find recent commits and diff against the appropriate base.

2. **Understand Context**: Before critiquing, understand *why* the changes were made. Read surrounding code to understand the intent and architecture. Check if the changes align with the project's MVVM architecture and module structure.

3. **Perform Multi-Dimensional Review**: Evaluate the changes across these dimensions:

### Correctness
- Logic errors, off-by-one errors, null safety issues
- Race conditions or concurrency problems (especially with coroutines)
- Edge cases not handled
- Resource leaks (unclosed streams, uncancelled coroutines, missing lifecycle cleanup)
- Incorrect API usage (Android SDK, WooCommerce/WordPress REST API)

### Architecture & Design
- MVVM pattern: ViewModels should extend `ScopedViewModel`, use `triggerEvent()` for one-shot events via `MultiLiveEvent`
- Proper dependency injection via Hilt (no service locators, proper scoping, modules in `di/` package)
- Separation of concerns (business logic not leaking into UI, data layer via FluxC)
- Fragments hosting Compose via `ComposeView` with `DisposeOnViewTreeLifecycleDestroyed`
- Navigation using XML nav graphs with NavController
- SOLID principles compliance

### Kotlin Best Practices
- Idiomatic Kotlin (scope functions, extension functions, sealed classes, data classes)
- Proper coroutine usage (appropriate dispatchers, structured concurrency, cancellation)
- Correct use of Flow vs StateFlow vs SharedFlow
- Proper nullability handling (avoid unnecessary `!!` operators)
- No wildcard imports

### Compose (when applicable)
- State hoisting: state up, events down — pass immutable values to composables
- `Modifier` must be the **first optional parameter**, named `modifier`, with default `Modifier`
- Do NOT acquire ViewModels inside composables; inject as parameter with default value
- Composable functions that emit content must return Unit and be named with PascalCase nouns
- Composable functions should not emit content at their top level — always wrap in a container (Column, Row, Box)
- Use `WooTheme` or `WooThemeWithBackground` (for proper light/dark support) as the root theme
- Always `remember` `mutableStateOf` / `derivedStateOf` inside composables
- Don't pass mutable types as parameters to composable functions
- Recomposition stability: avoid unnecessary recompositions, use proper `key` usage
- Use `UPPER_SNAKE_CASE` for constants and enums (not PascalCase)
- Set `contentDescription` for relevant icons and images; use `Modifier.semantics` for accessibility

### WooCommerce/WordPress API (when applicable)
- Correct endpoint usage and HTTP methods
- Proper authentication handling
- Pagination handling
- Error response handling and edge cases

### Security
- No hardcoded secrets, API keys, or credentials
- Proper input validation and sanitization
- Safe data handling (encryption where needed)
- No SQL injection or other injection vulnerabilities

### Testing
- Are new features covered by tests?
- Tests must extend `BaseUnitTest` (sets up `InstantTaskExecutorRule`, `CoroutineTestRule` with `UnconfinedTestDispatcher`)
- Use `testBlocking { }` helper for coroutine tests
- Test naming: backtick-wrapped BDD style — `` `given X, when Y, then Z` `` or `` `when X, then Y` ``
- Assertions: AssertJ (`assertThat(...).isEqualTo(...)`)
- Mocking: mockito-kotlin (`mock()`, `whenever()`, `verify()`)
- Compose tests: `ComposeTestRule` with finders/assertions/actions; use `waitUntil` not `Thread.sleep`

### Code Quality
- Max line length: 120 characters (exception: test names can be longer)
- Do NOT add comments to code (keep existing comments if already present)
- No FIXME in committed code — use TODO instead
- Readability and maintainability
- Dependencies managed via `gradle/libs.versions.toml`
- No unnecessary over-engineering or premature abstractions

## Output Format

Structure your review as follows:

### 📋 Review Summary
A 2-3 sentence overview of what was changed and the overall quality assessment.

### ✅ What Looks Good
List specific things done well. Be genuine — acknowledge good practices.

### 🔴 Critical Issues
Problems that could cause bugs, crashes, security vulnerabilities, or data loss. These MUST be fixed.
- For each issue: describe the problem, explain why it matters, and suggest a fix with code if possible.

### 🟡 Suggestions
Improvements that would enhance code quality, performance, or maintainability but aren't blocking.
- For each suggestion: describe the current approach, explain the improvement, and provide example code.

### 🟢 Nits
Minor style or preference items. Optional but would improve consistency.

### 📊 Overall Assessment
One of: **APPROVE** ✅ | **APPROVE WITH SUGGESTIONS** 🟡 | **REQUEST CHANGES** 🔴

## Guidelines

- Be specific: reference exact file names, line numbers, and code snippets
- Be constructive: always explain *why* something is an issue and suggest a concrete fix
- Be proportionate: don't block on style nits; reserve strong language for real problems
- Be respectful: critique the code, not the developer
- Prioritize: focus energy on critical issues over cosmetic ones
- Consider project context: respect the existing patterns and conventions in the codebase
- If you're unsure about a convention, check similar existing code in the project for precedent
- Do NOT suggest changes to code that wasn't modified in the current diff unless it's directly impacted by the changes
```

</details>

.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
```
hichamboushaba, line 2: Can we add `context-a8c` MCP too?

hichamboushaba (commented):
Great work @JorgeMucientes 👏 , I left some suggestions to enrich the discussion.

.agents/skills/pr/references/pr_template.md
ParaskP7: [SAME](https://github.com/woocommerce/woocommerce-android/pull/15382#discussion_r2826636836) (you could close mine)

.claude/settings.json
ParaskP7: True, and also, my 2-cents on that is that sometimes you might need those allowed, especially when you try something more automated when you want agents to be able to do such change themselves. But, having said that, maybe starting without and then check on the need as we go is wiser. 💭

ParaskP7:
Question (❓): About `Demo: AI Agent Emulator Verification with Screen Recording` and the fact it (hard) needs an emulator, it might be worth trying out with [Gradle Manage Devices](https://developer.android.com/studio/test/managed-devices) (GMD) and see if that will further help streamline this process for you. 🤔

FYI: The GMD promise this gives is that devs wouldn't even need to worry about creating/launching emulators, and if I remember correctly, they won't even need to have the emulator image downloaded. GMD will actually do it all of that, on the background; it will download the image (if not already available), create the emulator, launch the emulator, run the UI tests, stop the emulator, delete the emulator.

PS: For example, I am currently working on a skill and I wanted it to be able to do just that, but the project didn't have GMD setup, so what I told the skill was to configure GMD (if not configured already) and then smoke test launching the app, expand the below (just as an inpiration):

<details>
    <summary>Smoke Test via Gradle Managed Device</summary>

```md
## Step 6: Smoke Test via Gradle Managed Device (Local Only)

**This step uses Gradle Managed Devices exclusively. NEVER attempt to connect to,
check for, or use a physical device or standalone emulator (e.g., via `adb devices`).
All instrumented tests MUST run through the Gradle Managed Device infrastructure.**

**SKIP this step entirely when running as a background agent on a remote machine.**
Gradle Managed Devices require hardware virtualization and significant disk space
for system images, which remote/CI environments may not support. Proceed directly
to Step 7 in that case.

After unit tests pass, verify there are no runtime breakages from removed
`androidTestImplementation` or `implementation` dependencies by running a smoke
test on a Gradle Managed Device.

**Important:** The managed device configuration and smoke test class added in
this step are temporary scaffolding. Do NOT commit them. After the smoke test
passes or fails, revert all Step 6 changes to leave the working tree clean.

### 6a. Ensure a Gradle Managed Device is configured

Check the app module's `build.gradle(.kts)` for a `managedDevices` block inside
`android { testOptions { ... } }`. If none exists, add one:

    ```kotlin
    android {
        testOptions {
            managedDevices {
                localDevices {
                    create("pixel6Api34") {
                        device = "Pixel 6"
                        apiLevel = 34
                        systemImageSource = "google"
                    }
                }
            }
        }
    }
    ```

### 6b. Ensure a smoke test exists

Check for a simple androidTest class that launches the main activity and asserts
no crash. If none exists, create one in the app module's `androidTest` source set.
Adapt the activity class, test runner, DI setup (Hilt, Dagger, none), and assertions
to match the project's existing patterns — inspect existing androidTest classes for
conventions.

### 6c. Run the smoke test

    ```bash
    ./gradlew <deviceName>DebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=<package>.SmokeTest
    ```

Use the managed device name from Step 6a and the project's build variant.
```

.agents/skills/debug/SKILL.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Debug Build or Test Failures
+
+Diagnose and resolve build compilation errors or test failures.
+
+## For Build Failures
+
+1. **Run the build** to capture the full error output:
+   ```
+   ./gradlew assembleWasabiDebug 2>&1
```
malinajirka: ❓ Do we perhaps want to default to building just the Mobile app and not the watch app to speed up iterations?

.agents/skills/pr/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+description: Create a pull request following project conventions
+allowed-tools: Bash, Read, Grep, Glob
+user-invocable: true
+---
+
+# Create Pull Request
+
+Create a pull request following the project's PR conventions and template.
+
+## Steps
```
malinajirka, line 22: 💡 My skills also mentions:
```
## PR Title Format

- **Always prefix with `[WOOMOB-XYZ]`** where XYZ is the issue number
- Extract issue number from branch name (format: `issue/woomob-XYZ-...`)
- If issue number not available in branch name, ask the user for it

## PR Description Format

- **Always start with `Fixes WOOMOB-XYZ`** on the first line
```

Worth considering.

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,69 @@
+---
+name: review
+description: Review code changes against project architecture, style, and conventions
+allowed-tools: Bash, Read, Grep, Glob
+user-invocable: true
+---
+
+# Review Changes
+
+Review the current uncommitted or staged changes against the project's conventions and architecture rules.
```
malinajirka: ❓  Why just uncommitted? I usually run review before creating a PR not before each commit 🤔 .

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,69 @@
...
+- Compose screens live inside Fragments via `ComposeView` with `DisposeOnViewTreeLifecycleDestroyed`
+- ViewModels use `@HiltViewModel` + `@Inject constructor` with `SavedStateHandle` as the last parameter
+- Data flows through repositories — ViewModels never access Room or network directly
+- Navigation uses `NavController` via XML nav graphs
+
+### Kotlin Style
+- Max 120 character line length (test names excepted)
+- No wildcard imports
+- No `FIXME` (use `TODO`)
+- No new comments added to code
```
malinajirka: I've read a research that mentioned that being too strict leads to worse results - it's more efficient to provide context instead of strictly forcing the AI in some direction => e.g. here `no new comments added to code` can be interpreted that when it extracts a method it can't copy-paste the pre-existing comment => leading to AI deciding not to extract the method.

A phrasing leading to better results could be "Add comments only rarely when they provide a real value - e.g. explain business logic not clearly captured in the code." or something in that sense.

.agents/skills/test/SKILL.md
```diff
@@ -0,0 +1,49 @@
+---
```
malinajirka: I'd consider renaming it to `test-runner` for clarity.

.claude/settings.json
hichamboushaba: > my 2-cents on that is that sometimes you might need those allowed, especially when you try something more automated when you want agents to be able to do such change themselves. But, having said that, maybe starting without and then check on the need as we go is wiser.

Yes, I agree, but given that each of us has their own workflow, I think the best approach here is to start small, and then anyone can add what they need in their `~/.claude/settings.json` or `.claude/settings.local.json`

.claude/agents/test-writer.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Framework
+
+- JUnit 4 with mockito-kotlin for mocking and AssertJ for assertions
+- All tests MUST extend `BaseUnitTest` (from `libs/commons/src/testFixtures/`)
+- Use `testBlocking {}` for coroutine tests (wraps `runTest`)
+
+## Naming Convention
+
+Use backtick-wrapped BDD names:
```
malinajirka: I haven't tested this specific skill, but in my experience test skills with more examples perform significantly better. My skill is currently this:
```
---
name: kotlin-unit-tests
description: Writes clean, maintainable Kotlin unit tests. Use when writing unit tests, creating test files, adding tests, or modifying test code.
---

# Kotlin Unit Tests

Write clean, maintainable unit tests that are easy to read and modify.

## Core Principles

1. **Setup happy path defaults in `@Before`** - Each test should only override what it's testing
2. **One behavior per test** - Try to test one behavior per test to keep the test easy to read and maintain.
3. **Descriptive test names** - Test names should match "^(given .+, )?when .+, then .+$" regex.
4. **Consistency** - When adding or editing tests in existing test files, prefer consistency with the file over these rules.
5. **Don't mock data classes** - It's better to create a dummy data class than to mock it.

## Test Structure

```kotlin
@ExperimentalCoroutinesApi
class MyClassTest : BaseUnitTest() {

    private lateinit var sut: MyClass

    private val repository: MyRepository = mock()
    private val networkStatus: NetworkStatus = mock()
    private val tracker: AnalyticsTracker = mock()

    private val defaultId = 123L
    private val defaultModel = MyModel(id = defaultId, name = "Test")

    @Before
    fun setUp() = testBlocking {
        sut = MyClass(
            repository = repository,
            networkStatus = networkStatus,
            tracker = tracker,
            dispatchers = coroutinesTestRule.testDispatchers,
        )

        whenever(networkStatus.isConnected()).thenReturn(true)
        whenever(repository.fetchData(any())).thenReturn(Result.success(defaultModel))
    }

    @Test
    fun `given network not available, when loading data, then returns error`() = testBlocking {
        // GIVEN
        whenever(networkStatus.isConnected()).thenReturn(false)

        // WHEN
        val result = sut.loadData()

        //THEN
        assertThat(result.isFailure).isTrue()
    }

    @Test
    fun `given happy path, when loading data, then returns model`() = testBlocking {
        // WHEN
        val result = sut.loadData()

        // THEN
        assertThat(result.getOrNull()).isEqualTo(defaultModel)
    }
}
```

## Mock Configuration Patterns

### Pattern 1: Configure in @Before (preferred)
```kotlin
private val userRepository: UserRepository = mock()

@Before
fun setUp() {
    whenever(userRepository.getUser(any())).thenReturn(defaultUser)
}
```

### Pattern 2: Configure at declaration with `.also {}`
```kotlin
private val exception = mock<CustomException>().also {
    whenever(it.errorCode).thenReturn(ErrorCode.DEFAULT)
    whenever(it.message).thenReturn("Default error message")
}
```

### Pattern 3: Configure with `mock {}` block
```kotlin
private val selectedSite: SelectedSite = mock {
    on { get() } doReturn defaultSiteModel
}
```

## Test Naming

Test names should match "^(given .+, )?when .+, then .+$" regex.

**The "when" clause should describe the action (method call), not a condition.** Conditions go in "given".

### Bad: Condition in "when"
```kotlin
@Test
fun `when catalog generation completes, then returns success`()  // ❌ Wrong
```

### Good: Action in "when", condition in "given"
```kotlin
@Test
fun `given happy path, when syncCatalog, then returns success`()  // ✅ Correct
```

**Happy path tests should still include "given happy path"** for consistency, even when using default @Before setup:

```kotlin
@Test
fun `given happy path, when syncCatalog, then returns success`()  // ✅
```

## Comments in tests

Tests should be split into section with following comments `// GIVEN`, `// WHEN`, `// THEN`
sections (given is optional). Other comments should be added only when absolutely necessary, tests
should be self-documenting.

```kotlin
@Test
fun `given network fails, when loading data, then error state is emitted`() {
    // WHEN
    sut.loadData()

    // THEN
    assertThat(result.isFailure).isTrue()
}

@Test
fun `given network not available, when loading data, then error state is emitted`() {
    // GIVEN
    whenever(networkStatus.isConnected()).thenReturn(false)

    // WHEN
    sut.loadData()

    // THEN
    assertThat(result.isFailure).isTrue()
}
```


## Good vs Bad

### Good: Minimal, focused test
```kotlin
@Test
fun `given payment fails, when processing payment, then error is tracked`() = testBlocking {
    // GIVEN
    whenever(paymentProcessor.process(any())).thenReturn(Result.failure(Exception()))

    // WHEN
    sut.processPayment(order)

    // THEN
    verify(tracker).track(AnalyticsEvent.PAYMENT_FAILED)
}
```

### Bad: Repeating setup from @Before
```kotlin
@Test
fun `given payment fails, when processing payment, then error is tracked`() = testBlocking {
    // GIVEN
    whenever(networkStatus.isConnected()).thenReturn(true)
    whenever(userRepository.getUser()).thenReturn(mockUser)
    whenever(orderRepository.getOrder(any())).thenReturn(mockOrder)

    whenever(paymentProcessor.process(any())).thenReturn(Result.failure(Exception()))

    // WHEN
    sut.processPayment(order)

    // THEN
    verify(tracker).track(AnalyticsEvent.PAYMENT_FAILED)
}
```

## Sequential Mock Responses

When a mock needs to return different values on consecutive calls, use chained `thenReturn()` instead of `thenAnswer` with counters:

### Bad: Complex thenAnswer with counter
```kotlin
private suspend fun givenFailsThenSucceeds() {
    var callCount = 0
    whenever(repository.fetch()).thenAnswer {
        callCount++
        if (callCount == 1) Result.failure(Exception())
        else Result.success(data)
    }
}
```

### Good: Simple chained thenReturn
```kotlin
private suspend fun givenFailsThenSucceeds() {
    whenever(repository.fetch())
        .thenReturn(Result.failure(Exception()))
        .thenReturn(Result.success(data))
}
```

```

File: .claude/settings.json (hunk 2 above)
Comment (malinajirka) [reply .claude/settings.json]: I second this comment: `merge` `rebase` `push` (since it might decide to do force push) feel too risky.

Comment (ParaskP7) [reply .claude/settings.json]: > ...I think the best approach here is to start small...

I agree! 💯 👍

Comment (malinajirka) [reply .claude/settings.json]: I just noticed force push is explicitly disabled - TIL that it's possible to disable a specific flag while allowing the command with *. (Assuming it actually works)

File: .claude/agents/code-reviewer.md (hunk 5 above)
Comment (JorgeMucientes) [reply .claude/agents/code-reviewer.md]: I like it! Just curious, is there a reason you prefer `sonnet` model for the agent reviewer vs `opus`?

Comment (hichamboushaba) [reply .claude/agents/code-reviewer.md]: As I said previously in Slack, I was sharing this when I iterate more. For now, I'm using `sonnet` as I find it does a good job without taking a long time, as `opus` handles the planning and implementation, I think the review should be a less complex task.

But I think there is nothing wrong with using `opus` as you did for now, we can always iterate on this later.

File: .aiexclude (hunk 4 above)
Comment (JorgeMucientes) [reply .aiexclude:5]: Good question @ParaskP7 and I'm far from an expert but here are some reasons why some of those exclusions might make sense: 

>  I wonder if excluding these `build/` folders will do us any good

`build/ and */build/`. Why do you think would make agents less effective? The build/ folders contain generated bytecode, intermediate compilation artifacts, dex files, merged manifests, and other outputs that are massive (easily hundreds of MBs) and regenerated every build. Including them would bloat the agent's context with noise that has very low signal value. When an agent needs to analyze build errors, the relevant information comes from the Gradle console output imho. 

> PS.1: Same of ☝️ for the .gradle/ folder.

Anything valuable for an agent here? I feel this lands in a similar case as `build/` folder. 

> PS.2: I am also skeptical about the .buildkite/ and .github/ folders as well. I wouldn't exclude them, why would we, there isn't anything sensitive in there, is it? 🤔

`.github/` Yeap good catch here. This is definitely worth keeping imho, as it is not autogenerated content with each build. Moreover it has `PULL_REQUEST_TEMPLATE.md` file. As per `.buildkite/` it's proably worth keeping too. I don't think there's any sensitive data there and if any of the CI flows fails having that will probably help. 

> PS.3: Same of ☝️ for fastlane/.

Probably the same we can keep it for the same reasons as `.buildkite/` 👍🏼

Comment (ParaskP7) [reply .aiexclude:5]: Thanks for the reply @JorgeMucientes and I am with you on the whole `build/` and `.gradle/` folders, we might not need it all. And as you say, so we don't want to bloat the agent's `context`. But, then again, I am not sure on the agent's `context` side of things, if that would how it works (for me a mystery still), or if it just tries to load say load specific folder/file within `build/`, and only when and if it needs to... 🤔

---

So, I might then suggest:

1. Let's indeed ignore `.gradle/`.

2. But let's keep talking about the `build/` folder because (at minimum) I think these below will be useful to have:

`root` level `build/`, for at least `reports`:

image

`module level `*/build/`, for at least `outputs`, `reports` and `test-results`:

image

Wdyt? 🤔

PS: No matter what you'll decide to go with, we'll keep iterating on it anyway, so no bit deal for me, please go ahead with whatever you feel more comfortable atm. 🙏

File: .mcp.json (hunk 6 above)
Comment (JorgeMucientes) [reply .mcp.json:2]: Added! Just curious @hichamboushaba can you share a few examples were this MCP would be useful?

Comment (hichamboushaba) [reply .mcp.json:2]: A lot, actually. I use it more than any other MCP; it can fetch ticket details and documentation, and even access other repos on GitHub for implementation comparison.

File: .claude/agents/code-reviewer.md (hunk 5 above)
Comment (hichamboushaba) [reply .claude/agents/code-reviewer.md]: Something to add regarding `sonnet` here, the official code-review plugin from Claude [uses](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/code-review/commands/code-review.md) `sonnet` for the review part.

File: .mcp.json (hunk 6 above)
Comment (JorgeMucientes) [reply .mcp.json:2]: Hmm, yeah but you can do that without adding the MCP here @hichamboushaba.

**Comment by kidinov** [comment]
My quick take: we need to keep things balanced, as there are trade-offs between context consumed, other skills used (e.g. I use https://github.com/obra/superpowers) and the necessity to provide context manually. So taking this into consideration, I'd propose not to share in the repository those 3 skills:

 *  `debug`: It's mostly a generic table of "Unresolved reference → add import" and "NullPointerException → add mock stub". Any decent model already knows this. The project-specific parts (like the gradle
  commands) are already in AGENTS.md.
 *  `lint`: It's just `./gradlew detektAll` with or without `--auto-correct`, plus a list of common rules. That's already covered by AGENTS.md build commands + the detekt config file itself.
  * `test`: Similar story. It's basically "find the test file, run the right gradle command." The commands are in AGENTS.md, and the testing conventions are there too

File: .gitignore [hunk 13]
```diff
@@ -121,8 +121,8 @@ google-upload-credentials.json
 # Junie AI Assistant
 .junie/
 
-# Claude Code
+# AI Agents (shared config committed, personal config ignored)
 /CLAUDE.local.md
-.claude/
+.claude/settings.local.json
```
hichamboushaba: I would suggest the following (copied from the wpcom repo) as it's more generic and works for other files too:

```
# Agent config local files (Claude, Codex, etc.)
# Keep main config files under version control, ignore user-local overrides
**/CLAUDE.local.md
**/.claude/*.local.*
**/.codex/*.local.*
```

.claude/settings.json
JorgeMucientes: > I second this comment: merge rebase push (since it might decide to do force push) feel too risky.

Yeah those are already disabled (push is allowed but not force push as you well noticed) 👍🏼

hichamboushaba: > I second this comment: merge rebase push (since it might decide to do force push) feel too risky.

Even without push, `merge` and `rebase` might be destructive, if the branch wasn't pushed yet.

.agents/skills/review/SKILL.md
JorgeMucientes: Good catch! I think this part is confusing and a mistake that slipped through the cracks given right below in line 14, we are including all changes (uncommitted, staged, and "all commits in the branch") for review. 

I'll update this description section as it is misleading.

.aiexclude
JorgeMucientes: Done 👍🏼 [3d1a9f6](https://github.com/woocommerce/woocommerce-android/pull/15382/commits/3d1a9f6392b533f7e4ceef75525ab82967a4ef8c) . 
The `.aiexclude` now:
Excludes `build/ bulk` (intermediates, generated code, kotlin caches, etc.)
Keeps `build/reports/` at root level (detekt/lint reports)
Keeps `*/build/outputs/ (APKs)`, `*/build/reports/` (test/lint reports), and `*/build/test-results/` (JUnit XML results) at module level

JorgeMucientes:
Thank you all for your excellent input and suggestions @ParaskP7 @hichamboushaba @malinajirka and @kidinov 🙏🏼 

I've applied almost all of them and where I didn't I've replied with some extra reasoning to keep the discussion going. Please give it another go to ensure none of your feedback was missed (there where quite a few comments). 

As per this [suggestion](https://github.com/woocommerce/woocommerce-android/pull/15382#issuecomment-3926157579) @ParaskP7 . I think its a great suggestion as GMD could eliminate the "have an emulator running" prerequisite entirely. However, I think this is better addressed in a follow-up for a few reasons:

- Different problem space. This PR sets up agent configuration and conventions. Adding GMD changes `build.gradle` and introduces new infrastructure — that's a feature change with its own testing and team alignment needs. I'm happy to create a new issue and address that in a different task. 

- GMD and mobile-mcp solve different things imho. GMD automates emulator provisioning and runs instrumented tests. The verify-on-device skill uses mobile-mcp for interactive exploration (tapping, swiping, screenshots, recordings).

I also need to learn more about how GMD set up would look like. So for now I'm not applying anything related to that as part of this PR.

.aiexclude
ParaskP7: Awesome, thanks @JorgeMucientes ! 💯

ParaskP7:
> As per this https://github.com/woocommerce/woocommerce-android/pull/15382#issuecomment-3926157579 @ParaskP7 . I think its a great suggestion as GMD could eliminate the "have an emulator running" prerequisite entirely. However, I think this is better addressed in a follow-up for a few reasons:

For sure @JorgeMucientes , this was just me taking the opportunity because I saw a bit of emulator friction in there, definitely material for a follow-up and not at all needed for this exercise! 💯 (btw, thanks for considering this suggestion 🙇)

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+- Feature branches: `issue/ISSUEID-description` (e.g., `issue/1234-fix-order-list`)
+- Commit messages: < 100 characters, focus on what was done
+- MUST NOT include "Co-Authored-By" in commit messages
+- MUST NOT mention that code was generated by AI or any AI tool
+- PR template (`.github/PULL_REQUEST_TEMPLATE.md`): Description, Test Steps, Images/gif, RELEASE-NOTES.txt checkbox
+- Labels: at least one from Type / Status / Priority / Feature categories
+
+## Operational Rules
+
+- MUST NOT run static analysis (detekt, lint) or tests automatically during coding — only run when explicitly requested by the user
```
malinajirka: Are we sure about this? AutoCorrect can save a bunch of time and tokens - I'd argue we should run detekt before each build to ensure there are no formatting errors.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+- MUST confirm with the user before committing, pushing, or creating PRs — never assume ongoing permission from a previous approval
+- MUST NOT run `./gradlew` commands in the background unless explicitly told to
```
malinajirka, line 232: Doesn't this break auto-agent work? I'd expect the agent to buuld the app, verify tests, etc before it prompts me and tells me the job is done.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+- Commit messages: < 100 characters, focus on what was done
+- MUST NOT include "Co-Authored-By" in commit messages
+- MUST NOT mention that code was generated by AI or any AI tool
+- PR template (`.github/PULL_REQUEST_TEMPLATE.md`): Description, Test Steps, Images/gif, RELEASE-NOTES.txt checkbox
+- Labels: at least one from Type / Status / Priority / Feature categories
+
+## Operational Rules
+
+- MUST NOT run static analysis (detekt, lint) or tests automatically during coding — only run when explicitly requested by the user
+- MUST confirm with the user before committing, pushing, or creating PRs — never assume ongoing permission from a previous approval
```
malinajirka: I personally like to tell claude to create commits as it goes - it's easier to follow the train of thoughts, similarly to how we'd follow a train of thoughts of a regular developer. WDyt?

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
-- Compose tests: `ComposeTestRule` with finders/assertions/actions; use `waitUntil` not `Thread.sleep`
+- Mocking: mockito-kotlin (`mock()`, `whenever(...).thenReturn(...)`, `verify(...)`)
+- Compose UI tests: `ComposeTestRule` with finders/assertions/actions
+  - MUST use `waitUntil` for async — NEVER use `Thread.sleep`
+- What to test: ViewModels (state changes + events), Repositories (data flow), FluxC stores (action handling)
+- Test file locations mirror main source: `WooCommerce/src/test/`, `libs/<module>/src/test/`
+
+## Analytics
+
+- Add events to `AnalyticsEvent` enum in `WooCommerce/.../analytics/AnalyticsEvent.kt`
```
malinajirka: This can be confusing, POS has `WooPosAnalyticsEvent`. Might be worth mentioning here

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+- Track via injected `AnalyticsTrackerWrapper` — MUST NOT use the `AnalyticsTracker` singleton directly
```
malinajirka: POS has WooPosAnalyticsTracker, might be worth mentioning

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
-- Build debug: `./gradlew assembleWasabiDebug`
-- Install debug: `./gradlew installWasabiDebug`
-- Run detekt: `./gradlew detektAll --auto-correct`
+
+- Build debug APK: `./gradlew assembleWasabiDebug`
+- Install on device: `./gradlew installWasabiDebug`
+- Clean build: `./gradlew clean`
+- Run detekt: `./gradlew detektAll`
+- Run detekt with auto-correct: `./gradlew detektAll --auto-correct`
+- Android lint: `./gradlew lintVanillaRelease`
```
malinajirka: Do we really want to run lint on Vanilla? I'd prefer running everything on wasabi.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
 
 ## Build Commands
-- Build debug: `./gradlew assembleWasabiDebug`
-- Install debug: `./gradlew installWasabiDebug`
-- Run detekt: `./gradlew detektAll --auto-correct`
+
+- Build debug APK: `./gradlew assembleWasabiDebug`
+- Install on device: `./gradlew installWasabiDebug`
+- Clean build: `./gradlew clean`
+- Run detekt: `./gradlew detektAll`
```
malinajirka: I wouldn't mention detekt without auto-correct at all - agent should always run with autocorrect imo.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+.buildkite/                  CI pipeline configuration
+.agents/skills/              AI agent skills (shared across tools)
+```
 
 ## Build Commands
-- Build debug: `./gradlew assembleWasabiDebug`
-- Install debug: `./gradlew installWasabiDebug`
-- Run detekt: `./gradlew detektAll --auto-correct`
+
+- Build debug APK: `./gradlew assembleWasabiDebug`
```
malinajirka: We might want to explicitly mention :WooCommerce to ensure it builds the app not wear app.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+- Install on device: `./gradlew installWasabiDebug`
```
malinajirka: We might want to explicitly mention :WooCommerce to ensure it builds the app not wear app.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
-- **libs/cardreader** - Card reader functionality for payments
-- **libs/login** - Authentication and login flows
-- **libs/pos** - Point of sale functionality (POS)
-- **libs/commons** - Shared Compose components, utilities, and test fixtures
-- **libs/apifaker** - API mocking library for testing
-- **libs/detektrules** - Custom detekt rules for this project
+
+```
+┌─────────────────────────────────┐
+│  Fragment (hosts ComposeView)   │   UI layer — Jetpack Compose inside Fragments
```
malinajirka, line 60: We don't have fragments in POS - perhaps worth at least mentioning here so it's not confused.

AGENTS.md
```diff
@@ -1,71 +1,270 @@
...
+FluxC network DTOs → Room entities → domain models exposed to ViewModels.
+ViewModels never interact with Room or network directly — always through repositories.
+
+## Kotlin & Android Conventions
+
+- MUST use Kotlin for all new code
+- Max line length: 120 characters (exception: test names can be longer)
+- No wildcard imports
+- No `FIXME` — use `TODO` instead
+- MUST NOT add comments to new code (keep existing comments if present)
```
malinajirka: This will in my experience confuse the AI - e.g. if it want's to extract a method -> we should provide context on what is the goal, not restrict it like this.

.claude/agents/debugger.md
```diff
@@ -0,0 +1,47 @@
...
+description: Diagnoses build failures, test failures, and runtime issues
+model: sonnet
+tools: Read, Grep, Glob, Bash
+---
+
+You are a debugger for the WooCommerce Android project. Diagnose build and test failures systematically.
+
+## For Build Failures
+
+1. Run the build to capture the full error: `./gradlew assembleWasabiDebug 2>&1`
```
malinajirka: Similar to the below - I'd explicitly mention :WooCommerce so we avoid building wear.

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,119 @@
+---
```
malinajirka: The review plugin from anthropic is quite advanced - we might want to consider not adding a custom one at all - `code-review Plugin · claude-plugins-official · ✔ enabled`

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,69 @@
+---
```
malinajirka, line 1: ❓ Seems like we have `review` and `code-review` skills, what is the difference? Also as I mentioned below, I'd consider removing them both and using `code-review Plugin · claude-plugins-official · ✔ enabled`.

AGENTS.md
malinajirka: Actually, I think this doesn't belong to `agents.md`. It might be better to have `store-management-dev` skill and `pos-dev` skill, so we can easily differentiate between different architecture approaches and constraints.

AGENTS.md
JorgeMucientes: > AutoCorrect can save a bunch of time and tokens - I'd argue we should run detekt before each build to ensure there are no formatting errors.

What do you mean it can save a bunch of time and tokens? I'd say running detekt for every build does quite the oposite: 
- detektAll takes 30-60s. Running it before every build adds significant overhead when the agent may build multiple times during a task (fix → build → fix → build cycle). That's minutes of wasted time per iteration.
- Each detekt run produces output the agent has to process. Running it N times instead of once multiplies that cost for no benefit.

Additionally, I'd say the current rule ("run when explicitly requested") matches what developers do — you code first, then lint before the PR. Running detekt once at the end (before creating the PR) catches everything in one pass. I don't see much value in catching formatting errors mid-task (which btw are unlikely when AI codes) when auto-correct will fix them anyway at the end.

AGENTS.md
JorgeMucientes: Hmm, not sure about this either @malinajirka . This doesn't prevent incremental commits — it just requires the user to approve each one (you can of course choose to bypass this rule in each session). If we remove this guardrail, the agent might commit half-baked work, amend unrelated changes, or create a messy history.

AGENTS.md
JorgeMucientes: I don't have a strong opinion on this one, but this is the reasoning AI shared about this, and I think it makes sense, so imho we should keep this: 

> This rule doesn't prevent the agent from running gradle — it prevents running it in the background (i.e., with & or run_in_background). The agent can still run ./gradlew assembleWasabiDebug synchronously, wait for it to complete, read the output, and act on it.
The reason background gradle is problematic: the agent can't read the output reliably, may proceed before the build finishes, and can lose error output. Synchronous execution is better for agent workflows.
This rule does NOT break auto-agent work. The agent can still build → test → verify in sequence. It just does it synchronously, which is actually more reliable.

.claude/agents/code-reviewer.md
hichamboushaba: I didn't test the `code-review` plugin from Anthropic, but I think the advantage of a local defined code-review agent is that we can keep it focused on what matters most (Android, Kotlin, WordPress and WooCommerce APIs).
I can't say whether this will have a visible difference, maybe we can iterate on this as we go.
Funny point here: in Claude's [documentation](https://code.claude.com/docs/en/sub-agents), one of the examples they give for a subagent is `code-reviewer`.

malinajirka: The huge benefit of the plugin is that it spans multiple agents and is very thorough but still fast. Whenever I compared it to a local skill it was always magnitude better. With the good claude.md/agent.md I think it'll still perform better than a basic custom skill. Having said all that, we can try to iterate on our own skill and see - I can still continue using the plugin.

AGENTS.md
malinajirka: AI tends to notice formatting issues - then it tries to fix them and repeat (uses tokens) -> running detekt ensures there are 0 formatting issues with 0 tokens used.

AGENTS.md
malinajirka: I see - okay makes sense, thanks!

AGENTS.md
malinajirka: Isn't this exactly what a developer would do ? Commit as they go and then commit fixes, etc? Creating a history that one can follow to understand the follow up changes?

malinajirka: When you run 3 agents, approving each commit is super tedious - not mentioning cases when you run 10+ agents.

malinajirka: I'd personally remove this and if someone doesn't like it, they can create a custom rule/global claude.md.

AGENTS.md
malinajirka: Preventing agent from running tests also feels wrong - how can we create a long running loop if the agent can't verify its work? I'd suggest again leaving this for custom/local rules, not restricting it on a project level.

.claude/agents/code-reviewer.md
JorgeMucientes: Yeah I agree there's definitely some duplication there. But let's keep in mind the review skill lives in .agents/skills/ which is agent-agnostic — it works with Codex, Gemini, Copilot, or any AI tool that reads project files. The Anthropic code-review plugin only works in Claude Code.
So, I'd suggest keeping the `review.md` skill and remove the `code-reviewer.md` in favor of the plugin.

WDYT?

.agents/skills/review/SKILL.md
JorgeMucientes: Responded [here](https://github.com/woocommerce/woocommerce-android/pull/15382/changes#r2832986699)

.agents/skills/pr/SKILL.md
```diff
@@ -0,0 +1,61 @@
...
+user-invocable: true
+---
+
+# Create Pull Request
+
+Create a pull request following the project's PR conventions and template.
+
+## PR Title Format
+
+- **Always prefix with `[WOOMOB-XYZ]`** where XYZ is the issue number
```
irfano, line 14: I didn't realize this was our convention. Since we already include the Linear issue in the branch name and PR description, adding it to the title feels redundant. Also, the issue number is clickable in the description but not the title. Personally, I find project tags like `[CIAB: POS]` or `[Push Notifications]` much more useful.

.claude/agents/code-reviewer.md
malinajirka: Sounds good, thanks!

.agents/skills/pr/SKILL.md
malinajirka: I personally kind of like when I can see which issue is which from the PR overview page. Especially when I work on multiple PRs in one day and have multiple agent terminals open - I orient myself by the linear numbers.

JorgeMucientes: Hmm, I feel this falls more in the "personal preference" area. I think each one can add a local custom config here. 
I had added this line as per @irfano's feedback: 
`- Do NOT prefix with issue numbers — the issue is already linked in the branch name and PR description`

But maybe we can remove it and leave it up to the dev.

malinajirka: I think it'd be nice to be consistent and it's quite common practice to include issue number in the name -> there are even plugins for AS which make the issue number automatically clickable and navigate to Linear or any other tool. Is there any downside in having the issue number in the name?

irfano: > I think it'd be nice to be consistent

I agree.

> Is there any downside in having the issue number in the name?

The only drawback is that it crowds the title, leaving less room for the actual context.

a. Removing the Linear ID breaks your current workflow.
b. Keeping the ID doesn't really affect me or anyone else who doesn't use them.

Let's go with option **b**.

malinajirka: > crowds the title, leaving less room for the actual context.

Even very long titles are allowed on GitHub, e.g. :
Screenshot 2026-02-20 at 14 15 53

> b. Keeping the ID doesn't really affect me or anyone else who doesn't use them.
> Let's go with option b.

Thanks 🙇

irfano (approved):
Thanks for working this, Jorge! :shipit:

.claude/agents/code-reviewer.md
hichamboushaba: Just a remark regarding the `code-reviewer` agent vs the `code-review` plugin from Anthropic, I just tested the `code-review` one from Anthropic, and I think they serve different purposes, `code-review` can be invoked only against a PR, the agent `code-reviewer` can be used to review local changes (staged files) before committing them, which I find useful as it catches some issues earlier.

Just wanted to share this difference in case it's useful to someone of you.

malinajirka: The plugin can also be invoked against local changes - you just need to ask to do a local review.

hichamboushaba: Oh, I see, thanks! I'll test that. I just invoked it with `/code-review:code-review`; I didn't think to add any other input.

I'll try that and compare the results, but for my initial code review attempt, I wasn't impressed. The code-reviewer agent was more focused and returned correct actionable items, while the plugin returned one false positive only. But this was for just one attempt, I need to test more to confirm.

Also, one advantage of the local agent is that Claude is smart enough to invoke it automatically after the changes without waiting for my input, I think this can also be done using the plugin, but I think it has to be said explicitly.

malinajirka: > Also, one advantage of the local agent is that Claude is smart enough to invoke it automatically after the changes without waiting for my input, I think this can also be done using the plugin, but I think it has to be said explicitly.

This shouldn't be needed - it's a skill so AI can invoke it whenever it wants. If not, we can always encourage it by mentioning it in a skill file or agent.md.

> I'll try that and compare the results, but for my initial code review attempt, I wasn't impressed. The code-reviewer agent was more focused and returned correct actionable items, while the plugin returned one false positive only. But this was for just one attempt, I need to test more to confirm.

👍 I'll give the custom agent a chance - perhaps, it'll surprise me. I haven't used a custom solution like this in a few months and AI has evolved quite a bit, so I might be surprised.

hichamboushaba: > This shouldn't be needed - it's a skill so AI can invoke it whenever it wants. If not, we can always encourage it by mentioning it in a skill file or agent.md.

Yes, I'm aware, but In my very short experience with it, it didn't invoke it 😄, while with the agent, it did without me asking for it.

> 👍 I'll give the custom agent a chance - perhaps, it'll surprise me

Please let me know how it goes for you, as you used the plugin for longer than me 🙏

PR #15348, opened 2026-02-13
Implement "Connect to WordPress.com" setup steps UI

irfano:
Part of WOOMOB-1927

### Description

This PR adds the new Push Notifications connection steps screen and wires it from both My Store and Settings through a dedicated navigation graph. The new screen follows our Android-side visual style and reuses existing step indicator patterns. For now, tapping **Continue** on the Introduction screen navigates directly to this steps screen as a **temporary test flow**; the WordPress.com login step in between is intentionally skipped for now.  
Currently, the first step is set to `IN_PROGRESS` as a fake initial state, and step transitions will be updated once the real logic is added.

### Test Steps

1. Log in with a site with site credentials.
2. Open the flow from either My Store (card) or Settings (Enable Push Notifications).
3. Continue from the Introduction dialog.
4. Verify the temporary navigation lands on the steps screen.
5. Confirm that the first step appears as **In progress**, the other steps are initially **Not started**.

### Images/gif

|Current JP activation screen|New Woo PN screen|
|-|-|
|Screenshot_20260213_183332|Screenshot_20260213_200217|

|Dark Mode|All states|
|-|-|
|Screenshot_20260213_200320|Screenshot_20260213_200617|

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15385, opened 2026-02-18
Connect store to WordPress.com step

irfano:
### Description

This PR is a follow-up to #15348 and completes the **Connect to WordPress.com** step behavior in the new push notifications connection flow.

It wires the previously-collected login/connection context into the steps screen, executes the connection flow end-to-end, handles WebView fallback, and shows the correct success/error states.

#### Out of scope
- Plugin compatibility and later steps are not included in this PR (this PR focuses on Step 1: Connect to WordPress.com).
- If the store is already connected to WordPress.com, skipping the login flow earlier in the onboarding sequence will be added in a follow-up PR.

### Test Steps

1. Use a site that isn’t connected to WordPress.com.
2. Log in.
3. Enable the `WOO_PUSH_NOTIFICATIONS_SYSTEM` and `WOO_PUSH_NOTIFICATIONS_SYSTEM_M2` flags, then Restart.
4. Tap the "Never miss a new order card".
5. Complete the login flow.
6. Verify that the connection succeeded on the connection steps screen.

### Images/gif


hichamboushaba:
Hey @irfano, before reviewing this in full, I want to check some points first:

- I don't think we need to fetch Jetpack at this step, we need to do it in the intro step first as mentioned in the technical plan pe5sF9-4OM-p2 and discussed in this comment pe5sF9-4OM-p2#comment-5232. So I'd say remove that part from the ViewModel here, and assume that the site is not registered (use a hardcoded value of `JetpackConnectionStatus.AccountNotConnected`, we can then implement the fetching in the intro and pass the value.
- We don't need to think about WebView connection for this, I think we'll use only the native connection, as this is already supported by recent Woo versions, and we require updating the plugin anyway. So I think we can simplify the login quite a bit if we remove it, WDYT?

irfano:
Thanks for pointing them out. After our discussion (p1771499931208569-slack-C03L1NF1EA3), I made the necessary changes here. I didn’t include reordering the steps in this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/jetpack/main/JetpackActivationMainViewModel.kt
hichamboushaba, line 1: Maybe we can revert the changes of this file, and remove the `JetpackConnectionUrlResolver` since we don't need it now, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -20,19 +27,36 @@ import kotlinx.coroutines.flow.flowOf
...
+        initialValue = ConnectStoreStage.ConnectAccount,
+        key = KEY_CONNECT_STORE_STAGE
+    )
+
+    private val pendingConnectionStatus = savedStateHandle.getNullableStateFlow(
+        scope = viewModelScope,
+        initialValue = null,
+        clazz = JetpackConnectionStatus.AccountNotConnected::class.java,
+        key = KEY_PENDING_CONNECTION_STATUS
     )
```
hichamboushaba: I'm not sure why we need this? The Jetpack status will be passed from the previous screens, why do we need to keep it in the saved state again here?

We don't seem to be checking this value anywhere except when calling `jetpackActivationRepository.connectJetpackAccount`, so the value we'll get from the navigation will be enough.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -81,19 +107,85 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
...
+    }
+
+    private fun showConnectStoreStepComplete() {
+        pendingConnectionStatus.value = null
+        currentStep.update { current ->
+            current.copy(state = StepState.Success)
+        }
+    }
+
+    private fun showConnectStoreStepError(@StringRes messageRes: Int) {
```
hichamboushaba: This is not related to the connection, if we think these small utilify functions are useful, then maybe we can use a generic name, like `markStepAsFailed`, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -81,19 +107,85 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
...
+        return resolveConnectionErrorMessage(errorCode)
+    }
+
+    private fun resolveConnectionErrorMessage(errorCode: Int?): Int {
+        return if (errorCode == ERROR_CODE_FORBIDDEN) {
+            R.string.woo_push_notifications_connection_steps_error_connection_permission_message
+        } else {
+            R.string.woo_push_notifications_connection_steps_generic_error_message
+        }
+    }
```
hichamboushaba, line 179: IMO it would be more readable to combine these in the same function, both are called the same name, and one calls the other, and both are pretty small, WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -108,15 +200,17 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
 
     private fun getSiteAddress(): String {
         val site = selectedSite.get()
-        return StringUtils.getSiteDomainAndPath(site).ifBlank { site.name.orEmpty() }
+        return UrlUtils.removeScheme(site.url.orEmpty()).ifBlank {
+            StringUtils.getSiteDomainAndPath(site).ifBlank { site.name.orEmpty() }
+        }
```
hichamboushaba: The change doesn't make sense to me, `getSiteDomainAndPath` that we use now also depends on `site.url`, so if it's blank, both will be blank, and it already removes the scheme.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -81,19 +107,85 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
...
+            }
+        )
+    }
+
+    private suspend fun startConfirmConnectionStage() {
+        val siteUrl = requireNotNull(selectedSite.get().url?.takeIf { it.isNotBlank() }) { "Site URL missing" }
+
+        jetpackActivationRepository.fetchJetpackSite(siteUrl).fold(
+            onSuccess = {
+                showConnectStoreStepComplete()
```
hichamboushaba: We might want to call `advanceToNextStep` here to move to the next step.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModelTest.kt
hichamboushaba, line 1: The file structure was greatly modified (for example using `createViewModel` vs `setup`), and multiple unit tests were removed.

Can we double-check it and avoid unnecessary changes, please?

hichamboushaba (commented):
Nice work @irfano, I left some comments to address before approving this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/jetpack/main/JetpackActivationMainViewModel.kt
irfano: I forgot to remove this when I removed the WebView. I’ve fixed it in 9e91d1efb90d5a29382ef4a33a96740df4de575f.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: I overlooked this when updating the Jetpack status check behavior. 
`pendingConnectionStatus` made sense when we still had the `FetchStatus` step (it passed data to `ConnectAccount` and helped with restore).  
I removed it (a4fa1f25b93beac7a8b550d8707ec6b3dad89adc) and now `connectJetpackAccount` is called directly with `DEFAULT_CONNECTION_STATUS`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: Done: 2587f00a984b008cd581157b1522ec7a562a947e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: I agree! Done: b64cd846bae3666c152de05eb27ad29035b20803

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: That change was to fix tests that were broken due to the `StringUtils` dependency. However, the proper fix is to pass `StringUtils` via a constructor parameter. I’ve made the necessary change: 8f42505f6532e8b8e64834e6fdcbbb1d282bc000

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: 👍🏻  Added!

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModelTest.kt
irfano: I created this test class first, then rebased my PR onto your branch, which also included the same test class. It looks like I overwrote your changes during the rebase. I’ve updated the test class now.

hichamboushaba (approved):
Nice work @irfano, thanks for addressing my comments.

We need to simplify the logic further after p1771523624603299/1771522327.596419-slack-C03L1NF1EA3. We can handle it in a different PR.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2648,6 +2648,8 @@
     <string name="woo_push_notifications_connection_steps_in_progress">In progress</string>
     <string name="woo_push_notifications_connection_steps_not_started">Not started</string>
     <string name="woo_push_notifications_connection_steps_error">Error</string>
+    <string name="woo_push_notifications_connection_steps_error_connection_permission_message">You don’t have permission to connect this store to WordPress.com</string>
+    <string name="woo_push_notifications_connection_steps_generic_error_message">There was an error completing your request. Please try again or contact support if this error continues.</string>
```
hichamboushaba: I just noticed the duplication of this resource, we already have it below as `woo_push_notifications_connection_steps_generic_error`, I'll remove it.

PR #15387, opened 2026-02-19
[CIAB:POS] Fix bookings pagination and polish loading indicator

kidinov:
Closes: WOOMOB-2268

### Description

Fixes the bookings list pagination behavior where loading the next page caused a visual jump — the loading indicator disappeared before new items arrived, making the list scroll to the bottom with new items off-screen.

**Changes:**
- Fix the race condition between two state update paths during pagination. Now `observeBookings` flow handles clearing the loading state together with new items in a single state update, instead of `.onSuccess` clearing it prematurely
- Prevent showing the loading indicator when there are no more pages to load by exposing `hasMorePages` from `BookingListHandler`
- Redesign the pagination loading row to match the actual booking list item layout (time range, subtitle, badges) instead of the old order-style layout
- Move bookings above orders in the home screen menu

### Test Steps

1. Open POS and navigate to bookings (should be first in the menu now)
2. Scroll down the bookings list until a new page loads
3. Verify the loading indicator matches the booking item design (shimmer with time range, subtitle, and two badges)
4. Verify new items appear smoothly without the list jumping to the bottom
5. Scroll to the very end of the list — verify no stuck loading indicator when all pages are loaded

https://github.com/user-attachments/assets/f491384f-9feb-4ec1-9184-bd9750e676e1

- [ч] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
@@ -121,7 +121,8 @@ class WooPosBookingsViewModel @Inject constructor(
 
                 val currentContentState = _state.value as? WooPosBookingsState.Content
                 val paginationState = when (currentContentState?.paginationState) {
-                    WooPosPaginationState.Loading -> WooPosPaginationState.None
+                    WooPosPaginationState.Loading,
+                    WooPosPaginationState.Error -> WooPosPaginationState.None
                     else -> currentContentState?.paginationState ?: WooPosPaginationState.None
                 }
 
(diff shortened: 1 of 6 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -119,12 +119,19 @@ class WooPosBookingsViewModel @Inject constructor(
                     items.entries.find { it.key.id == id }?.value
                 }
 
+                val currentContentState = _state.value as? WooPosBookingsState.Content
+                val paginationState = when (currentContentState?.paginationState) {
+                    WooPosPaginationState.Loading -> WooPosPaginationState.None
```
malinajirka: 💡 Do we need to add `WooPosPaginationState.Error -> WooPosPaginationState.None` to clear pagination error?

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
kidinov: Good catch! Added in 792f35b — now both `Loading` and `Error` are reset to `None` when new data arrives from the flow.

PR #15388, opened 2026-02-19
[WOOMOB-2273] Refactor WooPos orders screen to separate single order and orders list views

malinajirka:
Fixes WOOMOB-2273

### Description
Refactored `WooPosOrdersScreen` composable to cleanly separate the single order mode from the orders list+details mode. Previously, the `isSingleOrderMode` boolean was threaded through `OrdersContent`, which conditionally hid/showed the list pane and adjusted weights. Now the screen branches at the top level into two distinct composables (`SingleOrderDetails` and `OrdersListWithDetails`), with shared order details logic extracted into a reusable `OrderDetailsPane`.

### Test Steps
1. Open POS
2. Open Order list
3. Smoke test it
4. Go back to POS
5. Open Bookings
6. Smoke test it

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
@@ -146,18 +146,24 @@ private fun WooPosOrdersScreen(
             .fillMaxSize()
     ) {
         when (state) {
-            is WooPosOrdersState.Content -> OrdersContent(
-                state = state,
-                isSingleOrderMode = isSingleOrderMode,
-                scrollToTopEvent = scrollToTopEvent,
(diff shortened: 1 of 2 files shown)


kidinov (approved):
LGTM. Not related to the PR, but maybe worth improving here - everything is jumping because the sizes are not correct for the loading of the title


https://github.com/user-attachments/assets/8bcc088c-42e0-47de-8393-3b7380ed8704

malinajirka:
Thanks!

> Not related to the PR, but maybe worth improving here - everything is jumping because the sizes are not correct for the loading of the title

I created an issue for this earlier: https://linear.app/a8c/issue/WOOMOB-2274/pos-booking-view-order-shimmer-loading-screen-doesnt-match-the-content. Let's fix it there - I don't think it's the size, but the fact that order list doesn't have toolbar so here when we added a toolbar -> all the items in the loading screen are pushed down unintentionally.

PR #15400, opened 2026-02-20
[Push Notifications] Check Jetpack connection status on Continue tap

hichamboushaba:
Closes WOOMOB-2270
Closes WOOMOB-2282
Closes WOOMOB-2265
Closes WOOMOB-1994

> [!NOTE]
> - Sorry for the size of the PR, the big part of the changes are Compose code and tests, I couldn't split it further. 
> - The text still mentions WordPress.com account, I didn't want to modify it in this PR, I'll update the wording in the PR that removes the login flow.
> - We don't have designs for the error screen, so I just used a generic error for now, we can update it if needed later.
> - We don't pass connection status to the steps screen yet, I'll handle that in a separate PR.

### Description

Check Jetpack connection status and WooCommerce version on the push notifications introduction screen to determine what to show the user.

**Fetching happens on screen open** , with a skeleton shimmer shown while loading. The `ViewState` is modeled as a sealed interface with five variants:

1. **Loading** → skeleton shimmer placeholder
2. **NotConnected** (site not registered / unknown) → intro screen with "Continue" button → WPCom login
3. **UpdateRequired** (connected, WC version incompatible) → "Update plugin" screen → connection steps
4. **GenericError** (connected + WC compatible, or fetch failure) → error screen with "Contact Support"
5. **ForbiddenError** (403) → error explaining the user needs to ask their store admin

A site is considered "connected" if it has site level connection with Jetpack, this matches: the Jetpack status is `AccountConnected` or `AccountNotConnected` with `JetpackSiteRegistrationStatus.REGISTERED`.

I also added a close button to the dialog.

#### Requirements
- Enable the PN M2 feature flag.
- Sign in using site credentials.

#### Jetpack not connected
1. Open the app
1. Tap on the Push Notification setup card in the dashboard.
1. Confirm the intro dialog shows the connection information.

#### Jetpack Connected and Woo outdated
1. Connect your site to Jetpack.
1. Make sure you have a Woo version less than `PUSH_NOTIFICATIONS_MIN_WC_VERSION` (set to 10.6.0 for now)
1. Tap on the Push Notification setup card in the dashboard.
1. Confirm the intro dialog shows the plugin update content.

#### Jetpack Connected and Woo up-to-date
1. Connect your site to Jetpack.
1. Modify `PUSH_NOTIFICATIONS_MIN_WC_VERSION` to be lower than your current plugin version.
1. Tap on the Push Notification setup card in the dashboard.
1. Confirm the intro dialog shows the error screen.

#### No permissions for checking Jetpack status
1. Sign in using a Shop Manager account.
1. Tap on the Push Notification setup card in the dashboard.
1. Confirm the intro dialog shows the forbidden error screen.

### Images/gif
Screenshot_20260220_144906
Screenshot_20260220_144827
Screenshot_20260220_144651
Screenshot_20260220_145013

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionDialog.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionDialog.kt
@@ -9,10 +9,12 @@ import androidx.fragment.app.viewModels
 import androidx.navigation.fragment.findNavController
 import com.woocommerce.android.R
 import com.woocommerce.android.extensions.navigateSafely
+import com.woocommerce.android.extensions.navigateToHelpScreen
 import com.woocommerce.android.ui.compose.composeView
 import com.woocommerce.android.ui.login.wpcom.WPComLoginMode
 import com.woocommerce.android.util.ChromeCustomTabUtils
 import com.woocommerce.android.viewmodel.MultiLiveEvent.Event.Exit
+import com.woocommerce.android.viewmodel.MultiLiveEvent.Event.NavigateToHelpScreen
 import dagger.hilt.android.AndroidEntryPoint
 import org.wordpress.android.util.DisplayUtils
 
@@ -34,11 +36,7 @@ class WooPushNotificationsIntroductionDialog : DialogFragment() {
         dialog?.window?.attributes?.windowAnimations = R.style.Woo_Animations_Dialog
 
         return composeView {
-            WooPushNotificationsIntroductionScreen(
-                onContinueClick = viewModel::onContinueClick,
-                onNotNowClick = viewModel::onNotNowClick,
-                onWhatIsWPComClick = viewModel::onWhatIsWPComClick
-            )
+            WooPushNotificationsIntroductionScreen(viewModel)
         }
     }
 
@@ -57,6 +55,15 @@ class WooPushNotificationsIntroductionDialog : DialogFragment() {
                             )
                     )
                 }
+
+                WooPushNotificationsIntroductionViewModel.NavigateToConnectionSteps -> {
+                    findNavController().navigateSafely(
+                        WooPushNotificationsIntroductionDialogDirections
+                            .actionWooPushNotificationsIntroductionDialogToWooPushNotificationsConnectionStepsFragment()
+                    )
+                }
+
+                is NavigateToHelpScreen -> navigateToHelpScreen(event.origin)
 
                 is WooPushNotificationsIntroductionViewModel.OpenUrlEvent -> {
                     ChromeCustomTabUtils.launchUrl(requireContext(), event.url)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
@@ -4,9 +4,6 @@ import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.asLiveData
 import com.woocommerce.android.AppUrls
 import com.woocommerce.android.extensions.isVersionAtLeast
-import com.woocommerce.android.model.JetpackConnectionStatus
-import com.woocommerce.android.model.JetpackSiteRegistrationStatus
-import com.woocommerce.android.model.JetpackStatus
 import com.woocommerce.android.support.help.HelpOrigin
 import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.jetpack.FetchJetpackStatus
@@ -27,16 +24,8 @@ class WooPushNotificationsIntroductionViewModel @Inject constructor(
     private val fetchActiveWCPluginVersion: FetchActiveWCPluginVersion,
     private val selectedSite: SelectedSite
 ) : ScopedViewModel(savedStateHandle) {
-
     companion object {
         const val PUSH_NOTIFICATIONS_MIN_WC_VERSION = "10.6.0" // TODO CHECK CORRECT VERSION LATER
-
-        private val JetpackStatus.isSiteConnected: Boolean
-            get() = when (jetpackConnectionStatus) {
-                is JetpackConnectionStatus.AccountConnected -> true
-                is JetpackConnectionStatus.AccountNotConnected ->
-                    jetpackConnectionStatus.siteRegistrationStatus == JetpackSiteRegistrationStatus.REGISTERED
-            }
     }
 
     private val _viewState = MutableStateFlow<ViewState>(ViewState.Loading)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
@@ -1,20 +1,97 @@
 package com.woocommerce.android.ui.pushnotifications.introduction
 
 import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.asLiveData
 import com.woocommerce.android.AppUrls
+import com.woocommerce.android.extensions.isVersionAtLeast
+import com.woocommerce.android.model.JetpackConnectionStatus
+import com.woocommerce.android.model.JetpackSiteRegistrationStatus
+import com.woocommerce.android.model.JetpackStatus
+import com.woocommerce.android.support.help.HelpOrigin
+import com.woocommerce.android.tools.SelectedSite
+import com.woocommerce.android.ui.jetpack.FetchJetpackStatus
+import com.woocommerce.android.ui.jetpack.FetchJetpackStatus.JetpackStatusFetchResponse
+import com.woocommerce.android.util.FetchActiveWCPluginVersion
 import com.woocommerce.android.viewmodel.MultiLiveEvent.Event
 import com.woocommerce.android.viewmodel.MultiLiveEvent.Event.Exit
 import com.woocommerce.android.viewmodel.ScopedViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
```diff
@@ -1,20 +1,91 @@
...
+            _viewState.value = ViewState.GenericError
+            return
+        }
+
+        if (wcVersion.isVersionAtLeast(PUSH_NOTIFICATIONS_MIN_WC_VERSION)) {
+            _viewState.value = ViewState.GenericError
+        } else {
+            _viewState.value = ViewState.UpdateRequired
+        }
+    }
```
hichamboushaba, line 81: This logic should be extracted somewhere else to be reusable for the setup screen too, we can handle this when working on plugin update task.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
```diff
@@ -1,20 +1,91 @@
...
+    private val fetchJetpackStatus: FetchJetpackStatus,
+    private val fetchActiveWCPluginVersion: FetchActiveWCPluginVersion,
+    private val selectedSite: SelectedSite
 ) : ScopedViewModel(savedStateHandle) {
+    companion object {
+        const val PUSH_NOTIFICATIONS_MIN_WC_VERSION = "10.6.0" // TODO CHECK CORRECT VERSION LATER
+    }
+
+    private val _viewState = MutableStateFlow<ViewState>(ViewState.Loading)
+    val viewState = _viewState.asLiveData()
```
irfano, line 32: We have a [rule](https://github.com/woocommerce/woocommerce-android/blob/trunk/AGENTS.md) (it doesn't exist in this branch yet):
> Don't use LiveData for new state in ViewModels — prefer StateFlow (existing LiveData is fine)

Would you like to expose the `StateFlow` like this:
```
val viewState: StateFlow<ViewState> = _viewState.asStateFlow()
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionDialog.kt
```diff
@@ -34,11 +36,7 @@ class WooPushNotificationsIntroductionDialog : DialogFragment() {
         dialog?.window?.attributes?.windowAnimations = R.style.Woo_Animations_Dialog
 
         return composeView {
-            WooPushNotificationsIntroductionScreen(
-                onContinueClick = viewModel::onContinueClick,
-                onNotNowClick = viewModel::onNotNowClick,
```
irfano, line 39: We have a rule in [AGENTS.md](https://github.com/woocommerce/woocommerce-android/blob/trunk/AGENTS.md):
`MUST NOT acquire ViewModels inside composables — inject as parameter with default value`

This PR changes the structure from passing separate parameters to passing the ViewModel. If you don’t have a reason for this change, I’d suggest reverting it here.

irfano (approved):
Nice work driving this change! 👏🏻
LGTM and it works as expected. I just added feedback based on our new rules in `AGENTS.md`. Please review them before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
hichamboushaba: I think the remark is about using LiveData as the backing state holder, not how it's exposed to the UI, the same file has this [line](https://github.com/woocommerce/woocommerce-android/blob/a3448bbeb73764927e4caa2c430228662c41d68a/AGENTS.md?plain=1#L95) 😄.

I have some remarks about exposing `LiveData` vs `StateFlow` to the UI layer, I recently discussed this topic with Adam [here](https://github.com/woocommerce/woocommerce-android/pull/14620#discussion_r2354609417).

Do you feel like we should move to expose StateFlow to the UI layer?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionDialog.kt
hichamboushaba: Unless I'm misunderstanding the rule, I don't think I'm breaking it. The ViewModel is still passed as a parameter; I believe the `not acquire` means just not calling the Composable `hiltViewModel` to create the ViewModel inside the Composable function. What I've done here is how we do it in all the other Composable screens in the app. Am I missing something? 🙂 

The `default` value doesn't apply when using Compose with Fragment IMO, it's just when using the Compose Navigation library.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
irfano: Okay, I guess I didn’t think about the rule in detail. The rule says we shouldn’t use LiveData for new state. Our state is already a `StateFlow`; we’re just exposing it as `LiveData` to Compose, which doesn’t violate the rule.

My comment was only because of the new rule in `AGENTS.md`, and it looks like we’re not breaking it, so it’s fine. 👍🏻

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionDialog.kt
irfano: You're completely right about the rule I quoted, I grabbed the wrong one! That said, our AGENTS.md does have a rule under [Common Pitfalls](https://github.com/woocommerce/woocommerce-android/blob/trunk/AGENTS.md#common-pitfalls) stating: `Don't pass ViewModel instances between composables — pass state and callbacks instead`. It also enforces `State hoisting: state up, events down (via lambdas)`.
Since the previous implementation was already passing events via lambdas, I'm curious if there's a specific reason for refactoring it to pass the ViewModel instead?

I'm aware we have many screens that use the same approach with `WooPushNotificationsIntroductionScreen(viewModel = viewModel)`, but I think we should try to follow the suggestion in `AGENTS.md` from now on. WDYT?"

irfano: If you think that passing the ViewModel into this top-level wrapper has advantages over extracting the state in the Fragment, maybe we should consider updating our guidelines in `AGENTS.md` to explicitly allow this pattern?

hichamboushaba: > I'm aware we have many screens that use the same approach with WooPushNotificationsIntroductionScreen(viewModel = viewModel), but I think we should try to follow the suggestion in AGENTS.md from now on. WDYT?

The point is that we has to convert from the `ViewModel` to the State somewhere (I mean calling `observeAsState`), what you suggest is doing it inside the Fragment's `onCreateView`, what we do in the codebase (I think everywhere) is to have the top-level Composable function of the screen accept the `ViewModel` itself, then it converts to the State (for example for our case [here](https://github.com/woocommerce/woocommerce-android/blob/cb34af4d170cd7b658adc6e0cf6349a55051830c/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionScreen.kt#L55-L62)), and calls another overloaded function that accepts the state and the callbacks. Personally, I prefer what we do, as it keeps the Compose logic out of the Fragment.
The approach of having two overloads (one taking the ViewModel and one taking the State and callbacks) is also used in the official Google samples ([for example](https://github.com/android/compose-samples/blob/baf2eb90c20b47b8e40b34104d3fab3e8ada58f9/Jetcaster/mobile/src/main/java/com/example/jetcaster/ui/player/PlayerScreen.kt#L119)).

> If you think that passing the ViewModel into this top-level wrapper has advantages over extracting the state in the Fragment, maybe we should consider updating our guidelines in AGENTS.md to explicitly allow this pattern?

We already have [documentation](https://github.com/woocommerce/woocommerce-android/blob/trunk/docs/compose.md) about Compose best practices, and I think it has examples and guidelines about both state hoisting and not acquiring the ViewModel, it has for example this part:

> State hoisting is basically moving all private state out of the @Composable functions to make composable functions stateless. Ideally, delegate data manipulation to the ViewModel or at least to the parent function that is calling the @Composable function.

So delegating to the `ViewModel` is explicitly listed as valid state hoisting.

The doc is already mentioned in the AGENTS.md, so if we feel we need to make this topic more clearer, I think we can do it in the doc itself. WDYT?

irfano: > what you suggest is doing it inside the Fragment's onCreateView, what we do in the codebase (I think everywhere) is to have the top-level Composable function

Agreed! We do this either in the Fragment or in a top-level composable wrapper. The top-level composable approach seems cleaner to me.

> The doc is already mentioned in the AGENTS.md, so if we feel we need to make this topic more clearer, I think we can do it in the doc itself. WDYT?

The rule that confuses me was `Don't pass ViewModel instances between composables`. But re-reading it, it says don't pass _between composables_, and here we're passing the ViewModel from a fragment to a composable, not between two composables. No need to update `AGENTS.md`. Thanks for the discussion @hichamboushaba, and sorry for the noise!

hichamboushaba: No worries @irfano, I enjoyed the discussion, and it's great to reflect on this points to clarify our common understanding.

PR #15401, opened 2026-02-20
[WOOMOB-2283] Hide pull-to-refresh indicator after booking state changes

malinajirka:
Fixes WOOMOB-2283

### Description
When a booking state changes (cancel, refund dismissed, note saved), the pull-to-refresh indicator was being shown unnecessarily. This happened because `onRefresh()` always sets `pullToRefreshState = Refreshing`, even for programmatic refreshes that don't need a visible indicator.

Added a `silent` flag to `onRefresh()` that skips setting the refreshing state. After cancel success, refund dialog dismissal, and booking note save, the data now refreshes silently without flashing the pull-to-refresh indicator. Also fixed cancel success not triggering a data refresh at all.

### Test Steps
1. Open POS > Bookings
2. Cancel a booking and confirm — verify the list updates without showing the pull-to-refresh indicator
3. Open a booking > Add a note > Save — verify the list refreshes without the indicator
4. Trigger a refund dialog and dismiss it — verify no indicator flash
5. Manually pull-to-refresh — verify the indicator still shows as expected

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
@@ -96,7 +96,7 @@ fun WooPosBookingsScreen(
 
     LaunchedEffect(cashPaymentResult.value) {
         if (cashPaymentResult.value) {
-            viewModel.onRefresh()
+            viewModel.onPullToRefresh()
             backStackEntry.savedStateHandle[BOOKING_CASH_PAYMENT_SUCCESS_KEY] = false
         }
     }
@@ -107,7 +107,7 @@ fun WooPosBookingsScreen(
 
     LaunchedEffect(cardPaymentResult.value) {
         if (cardPaymentResult.value) {
-            viewModel.onRefresh()
+            viewModel.onPullToRefresh()
             backStackEntry.savedStateHandle[BOOKING_CARD_PAYMENT_SUCCESS_KEY] = false
         }
     }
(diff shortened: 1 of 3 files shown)


kidinov (approved):
LGTM!

Small suggestion: instead of the silent boolean flag on onRefresh, we could split it into two methods — onPullToRefresh() for the UI-triggered pull-to-refresh (shows the indicator) and a private doRefresh() for silent background refreshes. This makes the intent clearer at each call site and avoids the boolean
  parameter:

```
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
index 8295f023d39..242348ff53c 100644
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
@@ -156,16 +156,18 @@ class WooPosBookingsViewModel @Inject constructor(
         )
     }
 
-    fun onRefresh(silent: Boolean = false) {
-        if (!silent) {
-            _state.value = when (val current = _state.value) {
-                is WooPosBookingsState.Content -> current.copy(
-                    pullToRefreshState = WooPosPullToRefreshState.Refreshing
-                )
-                else -> WooPosBookingsState.Loading
-            }
+    fun onPullToRefresh() {
+        _state.value = when (val current = _state.value) {
+            is WooPosBookingsState.Content -> current.copy(
+                pullToRefreshState = WooPosPullToRefreshState.Refreshing
+            )
+            else -> WooPosBookingsState.Loading
         }
 
+        doRefresh()
+    }
+
+    private fun doRefresh() {
         fetchJob?.cancel()
         loadMoreJob?.cancel()
         fetchResources()
@@ -260,7 +262,7 @@ class WooPosBookingsViewModel @Inject constructor(
         _state.value = currentState.copy(
             dialogState = WooPosBookingsState.Content.DialogState.Hidden
         )
-        onRefresh(silent = true)
+        doRefresh()
     }
 
     private fun handleCollectPayment() {
@@ -349,7 +351,7 @@ class WooPosBookingsViewModel @Inject constructor(
     }
 
     fun onBookingNoteSaved() {
-        onRefresh(silent = true)
+        doRefresh()
     }
 
     private fun handleBookingAction(action: WooPosBookingsState.BookingAction) {
@@ -437,7 +439,7 @@ class WooPosBookingsViewModel @Inject constructor(
                 )
             }
             if (result.isSuccess) {
-                onRefresh(silent = true)
+                doRefresh()
             }
         }
     }
```


Also what about payment flow?

PR #15402, opened 2026-02-20
[CIAB: POS] Hide collect payment button for refunded bookings

samiuelson:
### Description

WOOMOB-2285

When a booking is in "Cancelled" status with a "Refunded" payment status, the "Collect Payment" button was still being shown in the POS booking detail screen. This PR hides the collect payment button for refunded bookings, since collecting payment on an already-refunded booking doesn't make sense.

Adds an `isBookingRefunded` check alongside the existing `isBookingPaid` check so that `collectPaymentLabel` is set to `null` when the booking is both cancelled and refunded.

### Test Steps

1. Open a POS booking that is **Cancelled** with **Refunded** payment status
2. Verify the "Collect Payment" button is **not** shown
3. Open a POS booking that is **Cancelled** but **not refunded**
4. Verify the "Collect Payment" button **is** still shown
5. Open a **paid/completed** booking and verify collect payment is still hidden as before

#### Cancelled && Refunded
image

#### Cancelled 
image


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
```diff
@@ -143,13 +143,22 @@ class WooPosBookingViewStateMapper @Inject constructor(
         else -> false
     }
 
+    private fun isBookingRefunded(
```
kidinov: Small naming suggestion: `isBookingPaid` / `isBookingRefunded` sound like they only check the booking status, but they actually check a combination of booking status + payment status. This is a bit misleading.
                                                                                                                                                                                                                                                                                                                             
  Proposal:                                                                                                                                                                                                                                                                                                                  
```
  // Methods
  isBookingPaid(...)    → hasPaymentBeenCollected(...)
  isBookingRefunded(...) → hasPaymentBeenRefunded(...)

  // Local variables
  val isPaid = ...      → val isPaymentCollected = ...
  val isRefunded = ...  → val isPaymentRefunded = ...
```

kidinov: @samiuelson thinking further - 

**Suggestion:** instead of two boolean methods (`isBookingPaid` + `isBookingRefunded`), we could use a single method that returns an enum. The two booleans are really three mutually exclusive states: paid, refunded, or unpaid.                                                                                         
                                                     
  ```kotlin                                                                                                                                                                                                                                                                                                                  
  private enum class PaymentState { Paid, Refunded, Unpaid }            
                                                                                                                                                                                                                                                                                                                             
  private fun resolvePaymentState(                                                                                                                                                                                                                                                                                           
      bookingStatus: BookingEntity.Status,
      paymentStatus: PaymentStatus,
  ): PaymentState = when (bookingStatus) {
      BookingEntity.Status.Paid, BookingEntity.Status.Complete -> PaymentState.Paid
      BookingEntity.Status.Cancelled -> when (paymentStatus) {
          PaymentStatus.PAID -> PaymentState.Paid
          PaymentStatus.REFUNDED -> PaymentState.Refunded
          else -> PaymentState.Unpaid
      }
      else -> PaymentState.Unpaid
  }

  Then usage becomes:

  // actions list
  val paymentState = resolvePaymentState(booking.status, paymentStatus)
  if (paymentState == PaymentState.Paid) {
      add(WooPosBookingsState.BookingAction.IssueRefund(booking.orderId))
  }

  // payment section
  val paymentState = resolvePaymentState(booking.status, paymentStatus)
  paidWithLabel = if (paymentState == PaymentState.Paid) paymentInfo?.paymentMethodTitle else null,
  collectPaymentLabel = if (paymentState == PaymentState.Unpaid) totalAmount else null,
```

  This removes the two boolean methods, makes the states explicit, and avoids the !isPaid && !isRefunded check.

kidinov (approved):
LGTM - I left a code improvement suggestion

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
samiuelson: Good idea. Done in [3515c7f](https://github.com/woocommerce/woocommerce-android/pull/15402/commits/3515c7f1ee0eab16967f7e0eef4b9dbe9471374f)

PR #15403, opened 2026-02-20
[WOOMOB-2286] Change POS bookings sort to oldest-to-newest

malinajirka:
Fixes WOOMOB-2286

### Description
Changed POS bookings list sorting from newest-to-oldest to oldest-to-newest. The POS and main Bookings tab share the same Room `Bookings` table. When they used different sort orders, cached data from one screen would briefly display in the wrong order when opening the other, causing a visible "jump" as the fresh fetch replaced it. Switching POS to oldest-to-newest aligns with the main Bookings tab and matches user expectations (start of day → end of day).

### Test Steps
1. Open POS → Bookings — verify bookings are sorted from earliest to latest start time
2. Restart the app → open POS → Bookings — verify no visual content jump when fresh data loads
3. Compare the order with the main app Bookings tab — they should match


samiuelson (approved):
LGTM!

PR #15408, opened 2026-02-20
[Push Notifications] Remove WPCom login flow from push notifications setup

hichamboushaba:
Closes: WOOMOB-2287

> [!NOTE]
> I will handle code cleanup (removing the login changes) in a separate PR. 

## Description

Woo Push Notifications only requires a Jetpack **site connection** (site registration), not a WordPress.com **account** connection. The current flow unnecessarily routes users through a WPCom login before connecting the store, adding friction and complexity.

This PR removes the login flow and simplifies the connection step to only register the site.

### Changes

**Introduction screen:**
- Replaced `StartWPComLogin` event with `NavigateToConnectionSteps` — tapping "Continue" now goes directly to the connection steps, skipping WPCom login entirely
- Passes `isSiteConnectedToJetpack` to the connection steps so they can skip registration when the site is already connected

**Connection steps:**
- Replaced `connectJetpackAccount()` (site registration + account connection) with `registerSite()` (site registration only)
- Skips registration entirely when the site is already connected to Jetpack
- Removed two-stage `ConnectStoreStage` flow and `connectStoreStage` StateFlow

**Navigation:**
- Removed WPCom login include and action from the push notifications nav graph
- Removed `ShowPushNotificationsConnectionSteps` event and handlers from WPCom login fragments (2FA, magic link, password)
- Made `PushNotificationsSetup` login mode a no-op in `WPComLoginPostLoginViewModel`

**Repository:**
- Extracted `registerSite()` as a new public method in `JetpackActivationRepository`, reused by the existing `connectJetpackAccount()`

**Strings:**
- Removed "account" from WordPress.com references (the store still connects to WP.com, just not to a WPCom account)

#### Site not connected to Jetpack
1. Open a store that does not have Jetpack site connection → should see the introduction screen with "Continue" button
2. Tap "Continue" → should navigate directly to the connection steps screen (no WPCom login)
3. Connection steps should register the site

#### Site connected to Jetpack
1. Ensure your Woo version is lower than `PUSH_NOTIFICATIONS_MIN_WC_VERSION`
2. Connect your store to Jeptack.
5. Open push notifications setup intro
6. Tap on Continue → should skip registration step.

### Images/gif

https://github.com/user-attachments/assets/7a365859-4f77-4265-8468-60a99fa590cd


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
```diff
@@ -81,11 +81,7 @@ class WooPushNotificationsIntroductionViewModel @Inject constructor(
     }
 
     fun onContinueClick() {
-        if (_viewState.value is ViewState.UpdateRequired) {
-            triggerEvent(NavigateToConnectionSteps)
-        } else {
-            triggerEvent(StartWPComLogin)
-        }
+        triggerEvent(NavigateToConnectionSteps(isSiteConnectedToJetpack = _viewState.value !is ViewState.NotConnected))
```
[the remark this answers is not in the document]
hichamboushaba: `onContinueClick` can't be called from the other states, so I think this is fine.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -127,31 +116,23 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
...
     }
 
-    private suspend fun startConnectAccountStage() {
-        jetpackActivationRepository.connectJetpackAccount(
+    private suspend fun connectStoreToJetpack() {
+        if (navArgs.isSiteConnectedToJetpack) {
+            markCurrentStepAsCompleted()
+            advanceToNextStep()
+            return
+        }
```
hichamboushaba, line 124: For now, if the site is connected, we'll skip the task.

We can discuss with iOS team whether removing the task completely is better and update accordingly.

fastlane/resources/values/strings.xml
```diff
@@ -376,7 +376,7 @@
 
     <string name="my_store_widget_push_notifications_title">Never miss a new order</string>
     <string name="my_store_widget_push_notifications_menu_title">Enable push notifications</string>
-    <string name="my_store_widget_push_notifications_description">Connect your store to a WordPress.com account to get alerts for new orders and reviews.</string>
+    <string name="my_store_widget_push_notifications_description">Connect your store to WordPress.com to get alerts for new orders and reviews.</string>
```
irfano: ⚠️  We shouldn't update this `fastlane/.../strings.xml` in the PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
irfano: I think removing the task is a better approach. There’s no need to display an already completed step and make the screen more crowded.

irfano (approved):
LGTM! 👍🏻 
I voted to remove the WP.com connection stage if the site is already connected. If you agree, we can discuss this with the iOS team.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
hichamboushaba: Yes I agree, but let's first discuss with iOS team, and adjust in a separate PR when we agree, WDYT?

fastlane/resources/values/strings.xml
hichamboushaba: Good eye, thanks for catching this. I reverted it.

PR #15414, opened 2026-02-23
[WOOMOB-2291] Use store timezone across POS bookings

malinajirka:
### Description
Fixes WOOMOB-2291

POS bookings were using UTC for date/time operations — determining "today", formatting dates in the booking details, formatting time ranges, and computing API filter boundaries. This caused incorrect dates and times to be displayed for stores in non-UTC timezones.

This PR updates `WooPosBookingsViewModel`, `WooPosBookingViewStateMapper`, and `WooPosBookingTimeRangeFormatter` to use the store's timezone from `SelectedSite` for all date/time handling.

### Test Steps

1. In WP Admin → Settings → General, set the store timezone to **UTC-12** (Fiji) ([store-address]/wp-admin/edit.php?post_type=wc_booking&page=create_booking)
2. Create two bookings one that starts right after midnight and one that starts right before midnight.
3. Open POS → Bookings
4. Verify each booking appears on the correct date and with correct times (in Fiji's timezone)
5. Tap on view Order - verify the date and time on the order detail is correct.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -574,16 +579,17 @@ class WooPosBookingsViewModel @Inject constructor(
...
+            formattedDate = formatted,
             selectedDateMillis = millis,
         )
     }
 
     private fun dateRangeForDate(date: LocalDate): BookingsFilterOption.DateRange {
-        val start = date.atTime(LocalTime.MIDNIGHT).atOffset(ZoneOffset.UTC).toInstant()
-        val end = date.atTime(LocalTime.MAX).atOffset(ZoneOffset.UTC).toInstant()
+        val start = date.atStartOfDay(storeZoneId).toInstant()
+        val end = date.atTime(LocalTime.MAX).atZone(storeZoneId).toInstant()
```
kidinov, line 595: Do we need to do this here - maybe pass `date` to `format` and there already make start and end. Then we can get rid of `dateRangeForDate` method

kidinov (approved):
LGTM! Left one minor suggestion!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
malinajirka: I tried that but it feels a bit off since we'd be passing the info in view state. I think ideally, we'd move this out of the VM to some data utility class that consistently ensures timezone are handled correctly across the POS. But that's out of scope here.

PR #15415, opened 2026-02-23
[WOOMOB-2296] Fix bookings showing no bookings screen before loading

kidinov:
### Description
Fixes WOOMOB-2296

When opening POS bookings, a "no bookings" empty screen flashed briefly before bookings loaded. This happened because `fetchBookings()` onSuccess handler set the state to `Empty` before Room had a chance to deliver the fetched data. PR #15390 (date filtering) surfaced this race condition since `cleanAndUpsertBookings` in Room now triggers a re-emission that arrives after the `Empty` state was already set.

**The fix:** When `fetchBookings()` succeeds and no bookings are available yet, set `Content(NothingFound)` with the date selector instead of the full-screen `Empty` state. This way the UI shows the date picker and a contextual "no bookings for this date" message instead of a generic empty screen.

**Dead code cleanup:** The `Empty` state became unreachable after this fix. With date filtering always active, the bookings screen always has a selected date and should show the date selector - even when there are no results. The full-screen `Empty` state (which had no date selector) no longer makes sense, so it was removed along with its composable, preview, test, and string resources.

**Other improvements:**
- Renamed `Content.Items.Searching` to `Content.Items.Loading` since there is no search in POS bookings - this state represents loading items when switching dates
- `Loading` state now shows the date selector so users always see which date is selected

### Test Steps
1. Open POS bookings on a site with bookings for today
2. Verify bookings load without flashing a "no bookings" screen
3. Switch to a date with no bookings - verify "no bookings for this date" message shows with date selector
4. Switch back to a date with bookings - verify they load correctly
5. Pull to refresh - verify it works

### Images/gif

https://github.com/user-attachments/assets/82be6308-35b7-41d5-b0cd-ed8e5aecf3d6

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListHandler.kt
@@ -60,7 +60,7 @@ class BookingListHandler @Inject constructor(
         searchQuery: String? = null,
         filters: BookingFilters = BookingFilters.EMPTY,
         sortBy: BookingListSortOption
-    ): Result<Unit> = mutex.withLock {
+    ): Result<Int> = mutex.withLock {
         // Reset pagination attributes
         page.value = 1
         _canLoadMore.set(true)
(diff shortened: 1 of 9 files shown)


kidinov:
@malinajirka there is a date picker and some dead code in `WooPosBookingsDateSelector` - is this WIP?

malinajirka (approved):

I kind of hoped this was solved on the `BookingListHandler` layer similar to `WooPosCouponsListViewStateManager` but since it's not, I'd consider returning the number of updated items from `loadBookings` => show empty state only when it's 0 (for any other case, don't do anything in onSuccess). Your proposed solution minimizes the issue, but the empty state still sometime flashes. Wdyt? 

> @malinajirka there is a date picker and some dead code in WooPosBookingsDateSelector - is this WIP?

I believe this PR has the latest changes and I haven't planned any other must-does. We plan to adjust designs of the calendar view and perhaps display it docked, but nothing else than that.

I don't see any dead code though 🤔 - just one warning that feels like a false positive about `showDatePicker = false`. I think we don't need to set the showDatePicker = false twice, is that what you meant?

kidinov:
> I kind of hoped this was solved on the BookingListHandler layer similar to WooPosCouponsListViewStateManager but since it's not, I'd consider returning the number of updated items from loadBookings => show empty state only when it's 0 (for any other case, don't do anything in onSuccess). Your proposed solution minimizes the issue, but the empty state still sometime flashes. Wdyt?

Good catch - I was testing it on the site without todays bookings therefore didn't notice the blinking. Implmeneted here f4029700fb59cea37e3f3215dad5ee938cb60ed3

> I don't see any dead code though 🤔 - just one warning that feels like a false positive about showDatePicker = false. I think we don't need to set the showDatePicker = false twice, is that what you meant?

It looks like that I had some half baked version from the trunk here, and there was code to show the date picker, but it was not showing. I merged the latest trunk, and it looks all good there

PR #15417, opened 2026-02-23
[Push Notifications] Handle plugin update in push notifications setup

hichamboushaba:
Closes WOOMOB-1996

### Description
- Extract `CheckWooPluginPushNotificationsSupport` use case to check WC plugin version compatibility for push notifications
- When the intro screen detects an outdated plugin (`UpdateRequired`), navigate to the connection steps screen with `shouldAutoOpenUpdatePlugin=true` which auto-opens the plugin update web view
- When the connection steps screen detects an outdated plugin during normal flow, show a version-specific error message with an "Update plugin" button and a secondary "Try again" button
- After returning from the plugin update web view, the plugin compatibility step is automatically retried

This flow follows what was discussed here p1771936947313419-slack-C03L1NF1EA3

#### Requirements:
- Make sure the PN M2 feature flag is enabled.
- Modify the value of `PUSH_NOTIFICATIONS_MIN_WC_VERSION` to 10.5.2 (the current version of Woo)

#### Site disconnected
1. Install [WC 10.5.0](https://downloads.wordpress.org/plugin/woocommerce.10.5.0.zip) manually
2. Make sure the site is not connected to Jetpack.
3. Open the app and start PN setup.
4. Confirm the setup screen shows the Update Required error screen (WlcBAWKwDmaQFvRTx2vxTk-fi-179_6267)
5. Tap on Update plugin.
6. Update the plugin, then tap on Done (or close the WebView)
7. Confirm that the app confirms the compatibility, and continues to next steps.

#### Site already connected
1. Install [WC 10.5.0](https://downloads.wordpress.org/plugin/woocommerce.10.5.0.zip) manually
2. Make sure the site is connected to Jetpack.
3. Open the app and start PN setup.
4. Confirm the WebView is auto-opened for the plugin update.
5. Update the plugin, then tap on Done (or close the WebView)
6. Confirm that the app confirms the compatibility, and continues to next steps.

### Images/gif
https://github.com/user-attachments/assets/b6c9be6c-100b-4225-90f8-f7a52b44f665


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/283">24.3</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/connection/WooPushNotificationsConnectionStepsViewModel.kt
```diff
@@ -202,6 +246,7 @@ class WooPushNotificationsConnectionStepsViewModel @Inject constructor(
     ) {
         val isDone = steps.all { it.state == StepState.Success }
         val isError = steps.any { it.state is StepState.Error }
+        val isPluginUpdateRequired = steps.first { it.type == StepType.CheckPluginCompatibility }.state is StepState.Error
```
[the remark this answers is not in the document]
hichamboushaba: Good point, this was addressed, but in a way that doesn't rely on String comparison.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/common/webview/AuthenticatedWebViewScreen.kt
```diff
@@ -43,10 +43,18 @@ private fun AuthenticatedWebViewScreen(
...
-                )
+                ),
+                actions = {
+                    if (viewState.displayMode == DisplayMode.TASK) {
+                        WCTextButton(
+                            onClick = onClose,
+                            text = stringResource(id = R.string.done)
+                        )
+                    }
+                }
```
hichamboushaba, line 57: I updated the WebView screen to show a `Done` button when using the new display mode `TASK` that I added. I think it's clearer than the `X` button we use for navigation, especially since we need the user to know they need to tap this when finishing.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
```diff
@@ -68,20 +65,34 @@ class WooPushNotificationsIntroductionViewModel @Inject constructor(
...
-            _viewState.value = ViewState.GenericError
-            return
-        }
-
-        if (wcVersion.isVersionAtLeast(PUSH_NOTIFICATIONS_MIN_WC_VERSION)) {
-            _viewState.value = ViewState.GenericError
-        } else {
-            _viewState.value = ViewState.UpdateRequired
+        when (checkWCPluginSupport()) {
+            Compatible -> _viewState.value = ViewState.GenericError
```
JorgeMucientes, line 91: I wonder if displaying the generic error screen when the plugin is compatible but token registration fails:

https://github.com/user-attachments/assets/719327b2-d684-4097-a33b-104e6f1a2230

Is better than displaying the screen where 2 steps (plugin compatibility and WP.com connection) are success and the token registration step is the one failing. That seems more informative and could be more helpful in case the user needs to reach out to support and share a screenshot. WDYT?

JorgeMucientes (approved):
Great work @hichamboushaba code looks good and everything works as expected. Just left a question to discuss, but definitely out of scope for this PR. :shipit:

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/pushnotifications/CheckWooPluginPushNotificationsSupportTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test
+    fun `given newer version, when invoked, then returns Compatible`() = testBlocking {
+        whenever(fetchActiveWCPluginVersion()).thenReturn("11.0.0")
+
+        val result = sut()
+
+        assertThat(result).isEqualTo(Result.Compatible)
+    }
```
JorgeMucientes, line 34: Oh one thing I missed commenting. Isn't this test redundant to the above? Given the logic from `CheckWooPluginPushNotificationsSupport` they are pretty much testing the same thing twice

hichamboushaba: I don't know, I'd say it's not redundant, one is testing `==` and one is testing `>`, the test shouldn't care about the internal implementation of the system that we are testing, just the outcome.
Internally we are testing `isVersionAtLeast`, if someone changes its logic and broke the `==` part, but the `>` was still working, the two tests is the only way to catch this.

WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/pushnotifications/introduction/WooPushNotificationsIntroductionViewModel.kt
hichamboushaba: Good point, I think this is worth a discussion in Slack with the iOS team to discuss the pros and cons. Personally I was planning to start a discussion that suggests trying to register a token in this situation before showing the error, what you are suggesting is good too, we just need to decide what to show in the screen, because currently it shows "Update plugin".

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/pushnotifications/CheckWooPluginPushNotificationsSupportTest.kt
JorgeMucientes: Fair enough, that makes sense. Let's keep it 👍🏼

PR #15423, opened 2026-02-24
[WOOMOB-2291] Fix double-offset in POS bookings timezone handling

malinajirka:
### Description
Fixes WOOMOB-2291
Fixes WOOMOB-2305

**Why another PR after #15414?** PR #15414 switched all date/time operations from UTC to the store timezone. However, the WooCommerce Bookings API returns Unix timestamps where the value represents **site-local time encoded as UTC** (fake UTC). For example, an 11:00 AM Vilnius booking (UTC+2) has a timestamp that decodes to `11:00 UTC` — not `09:00 UTC` as true UTC would be.

This means #15414 introduced a **double-offset problem**: the formatters applied the store offset on top of timestamps that were already in local time, causing displayed times to be shifted by the store's GMT offset (e.g., an 11:00 AM booking showing as 1:00 PM in UTC+2).

**Why did manual testing of #15414 look correct?** The bookings used for testing were created via the date picker for Booking Products on the web, which has a bug - it uses some seamingly random timezone when storing the dates. Date picker for Booking **Services** stores the dates as fake UTC, which is what exposed the double-offset issue.

**The fix:** Since the API timestamps *are* local time (just labeled as UTC), we interpret them as UTC everywhere — for display formatting and for API request boundaries. The one exception is determining "today" (`selectedDate`), which correctly uses the store timezone to know what wall-clock day it is at the store.

**Concretely this PR:**
- Reverts `dateRangeForDate` in `WooPosBookingsViewModel` to use `ZoneOffset.UTC` for API filter boundaries
- Removes `SelectedSite` dependency from `WooPosBookingTimeRangeFormatter`, `WooPosBookingViewStateMapper`, and `WooPosDateFormatter` — all now use UTC
- Renames `formatShortDateInStoreTimeZone` → `formatShortDate` (since it's no longer store-timezone-aware)

### Test Steps

1. In WP Admin → Settings → General, set the store timezone to a non-UTC zone (e.g., **UTC-8** or **Los Angeles**)
2. Create a booking for tomorrow at 6:00 AM  **Make sure to use Booking Service not Booking Product**
3. Open POS → Bookings
4. Verify the booking appears on the correct date with the correct time (6:00 AM, not shifted by 8 hours)
5. Navigate to a different day and back — verify the date selector and booking list are correct
6. Tap a booking → verify the detail view shows the correct date and time
7. Tap "View Order" → verify the booking info line in order details shows the correct date/time
8. Create a booking for 6pm - double check it's still shown on the correct date in the POS


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingTimeRangeFormatter.kt
```diff
@@ -4,21 +4,14 @@ import android.icu.text.DateIntervalFormat
...
 import java.time.Instant
-import java.time.ZoneId
 import javax.inject.Inject
 
-class WooPosBookingTimeRangeFormatter @Inject constructor(
-    private val selectedSite: SelectedSite,
-) {
-    private val storeZoneId: ZoneId by lazy { selectedSite.get().clock.zone }
-
+class WooPosBookingTimeRangeFormatter @Inject constructor() {
```
kidinov, line 10: Maybe to add a comment that explains why UTC is used? And link this PR as it has a nice explanation as well?

kidinov (approved):
LGTM! 

A bit scary that when the backend is fixed, then we will show wrong dates in the app until everyone gets the update, but I guess there is no way around this in this situation

malinajirka:
Thanks for the review @kidinov 

> A bit scary that when the backend is fixed, then we will show wrong dates in the app until everyone gets the update, but I guess there is no way around this in this situation

I don't think they'll update the behavior - from what I understood this is the desired behavior they implemented on purpose.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15426, opened 2026-02-24
[WOOMOB-2306] Add WordPress.com TOS to push notifications intro screen

hichamboushaba:
### Description
Fixes WOOMOB-2306

Adds the WordPress.com Terms of Service footer to the push notifications introduction screen when the site is not connected. Uses the existing reusable `WPComConsent` component with `forJetpackSetup = false`.

### Test Steps
1. Trigger the push notifications introduction screen (when site is not connected)
2. Verify the WordPress.com TOS footer appears below the \"Not now\" button
3. Verify the TOS footer does NOT appear when the update is required (UpdateRequired state)
4. Tap the \"Terms of Service\" link and verify it opens the WordPress.com TOS in a browser

(Alternatively just check the Compose preview)

### Images/gif
Screenshot_20260224_180354


irfano (approved):
LGTM! 👍🏻

PR #15430, opened 2026-02-25
[WOOMOB-2317] Clear push notification registration state after WooCommerce plugin downgrade

hichamboushaba:
### Description
Fixes WOOMOB-2317

When the WooCommerce plugin is downgraded to a version below 10.6.0 after the app has already registered for self-driven push notifications, the app now treats the site as "not registered". This is achieved by folding a cached plugin compatibility check (using the local DB, no network call) into both `isWooPushTokenRegisteredForSite` and `observeWooPushTokenRegisteredForSite` in `PushNotificationRepository`, so that dashboard and settings entry points reappear correctly after a downgrade.

`CheckWooPluginPushNotificationsSupport` was also moved from `ui/pushnotifications` to `notifications/push` for better layering (it has no UI concerns), and extended with a `forceRefresh` parameter (`true` = network fetch, `false` = cached DB read).

### Test Steps
1. Make sure the Push Notifications feature flag is enabled.
2. Install the app on a site running WooCommerce ≥ 10.6.0 (from this post p91TBi-dyW-p2)
3. Make sure the site is connected to Jetpack.
4. Open the app; the push notifications registration should succeed, and entry points should be hidden.
5. Downgrade the WooCommerce plugin to a version below 10.6.0 on the site.
6. Reopen the app (site switch or app restart triggers plugin cache refresh via `SiteObserver`).
7. Verify the push notification entry points (dashboard card / settings row) reappear, indicating the site is treated as "not registered".

**Note:** the dashboard card might not appear on first launch after the downgrade, because the DB is not observer, we get a single status when the screen is opened, where there might be a race condition that the fetch is not done yet. I believe this is an accepted limitation, the refactoring to observe the DB is not justified, please let me know what you think.


WooCommerce/src/test/kotlin/com/woocommerce/android/notifications/push/CheckWooPluginPushNotificationsSupportTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Test
+    fun `given forceRefresh=false, when invoked, then does not call fetchActiveWCPluginVersion`() = testBlocking {
+        whenever(getWooCorePluginCachedVersion()).thenReturn("10.6.0")
+
+        sut(forceRefresh = false)
+
+        verify(fetchActiveWCPluginVersion, never()).invoke()
+    }
```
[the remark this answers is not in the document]
hichamboushaba: I don't think that's the role of `verifyBlocking`, it's just a way to verify a suspendable call in non-suspendable tests, given our test is already suspendable, then `verify` is fine.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -143,14 +144,22 @@ class PushNotificationRepository @Inject constructor(
...
     }
 
     fun observeWooPushTokenRegisteredForSite(siteId: Long): Flow<Boolean> {
         val tokenKey = getPushTokenKeyForSite(siteId)
-        return pushNotificationsDataStore.data.map { preferences -> preferences[tokenKey] != null }
+        return pushNotificationsDataStore.data.map { preferences ->
+            val isTokenStored = preferences[tokenKey] != null
+            val supportResult = checkWooPluginPushNotificationsSupport(forceRefresh = false)
+            // Treat errors as "compatible" to avoid hiding entry points during temporary failures
+            val isPluginCompatible = when (supportResult) {
```
irfano, line 163: np: a note regarding coding style, `val isPluginCompatible = supportResult !is CheckWooPluginPushNotificationsSupport.Result.UpdateRequired` could be simpler here. Using `when` is safer for future changes, but I don't think the risk is high in this case, and a single line is cleaner than 5 lines. 
_Feel free to ignore._

irfano (approved):
LGTM! 👍🏻  Thanks for addressing this!

> Note: the dashboard card might not appear on first launch after the downgrade, because the DB is not observer, we get a single status when the screen is opened, where there might be a race condition that the fetch is not done yet. I believe this is an accepted limitation, the refactoring to observe the DB is not justified, please let me know what you think.

I noticed it too, it didn’t appear on my first run. I think we shouldn’t address this in this PR. It’s an edge case, and we can revisit it later if it becomes necessary.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
hichamboushaba: It was like that before, and Copilot suggested the `when` syntax, I found it more readable and I applied it, so we are back to a full circle here 😄.

I'll leave it as it's now, I think this is a minor point.

PR #15434, opened 2026-02-25
[WOOMOB-2297] Fix POS DataStore not cleared on logout and self-hosted site key collision

malinajirka:
Fixes WOOMOB-2297

### Description
Do not merge label - **let's merge into 24.3** which will go through a regular beta testing cycle.

The POS DataStore (sync timestamps, preferences) was not cleared during logout. When a user logged out and back in to the same site, the stale sync timestamps made POS skip the catalog sync — even though the product database was empty for the new session — resulting in an empty product list until a manual full sync was triggered.

Additionally, sync timestamp DataStore keys used `site.siteId` which is `0` for self-hosted sites, causing all self-hosted sites to share the same key prefix.

**Fix 1 — Clear POS DataStore on logout:**
- Added `WOO_POS` qualifier to the POS DataStore (following the existing `DataStoreQualifier` pattern)
- Injected the qualified DataStore into `AccountRepository` and clear it during `cleanup()`

**Fix 2 — Use `remoteId()` for DataStore keys:**
- Changed `site.siteId` → `site.remoteId().value` in `WooPosSyncTimestampRepository.buildSiteSpecificKey()`, which correctly returns `siteId` for WP.com/Jetpack sites and `selfHostedSiteId` for self-hosted sites.

### To test
1. Log in to a site with POS enabled and open POS — products load after initial sync
2. Log out
3. Log back in to the same site and open POS
4. **Before fix:** POS shows empty product list (sync is skipped because stale timestamp says "recently synced")
5. **After fix:** POS triggers a fresh full sync and products appear


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/datastore/WooPosSyncTimestampRepository.kt
```diff
@@ -104,7 +106,7 @@ class WooPosSyncTimestampRepository @Inject constructor(
     private fun buildSiteSpecificKey(key: String): Preferences.Key<String>? {
         val site = selectedSite.getOrNull()
         return if (site != null) {
-            stringPreferencesKey("${site.siteId}-$key")
+            stringPreferencesKey("${site.remoteId().value}-$key")
```
kidinov, line 109: Shell we do the same in `WooPosPreferencesRepository` then?

malinajirka: We use `localsiteid` there, so it's not necessary.

kidinov (approved):
LGTM!

PR #15435, opened 2026-02-25
[WOOMOB-2221] Fetch and display booking location from product data in POS

kidinov:
### Description
Fixes WOOMOB-2221

The POS booking details screen has a location row that was always empty because the booking API doesn't include location data. Location is a product-level field (`booking_location`) exposed by the WooCommerce Bookings plugin on the product REST endpoint.

This PR adds a lightweight fetch for the product's `booking_location` field (using `_fields` to minimize the response) and wires it through to the POS booking details. Locations are cached by productId and fetched in parallel for all visible bookings.

### Test Steps
1. Set up a site with WooCommerce Bookings plugin
2. Create a bookable product with a location set (e.g. "238 Willow Creek Drive")
3. Create a booking for that product
4. Open WooPos and navigate to the Bookings tab
5. Select the booking to see its details
6. Verify the location row shows the correct address
7. Verify a booking for a product without a location set does not show the location row

### Images/gif
image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
@@ -180,6 +180,18 @@ class BookingsRepository @Inject constructor(
         }
     }.await()
 
+    suspend fun fetchProductBookingLocation(productId: Long): Result<String?> {
+        val result = bookingsStore.fetchProductBookingLocation(
+            site = selectedSite.get(),
+            productId = productId
+        )
+        return if (result.isError) {
+            Result.failure(WooException(result.error))
+        } else {
+            Result.success(result.model)
+        }
+    }
+
     suspend fun markAsPaid(
         bookingId: Long,
     ): Result<Unit> = appCoroutineScope.async {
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

kidinov:
@malinajirka @samiuelson do you think this is worth merging before the cut off?

malinajirka:
I'm fine merging it (after I run manual tests). The only thing I'm not sure about is whether it's worth it from the POV that this will be provided by the booking API in the future version - so we'll have to revert and re-implement. But I guess that's fine, wdyt?

kidinov:
>  I'm fine merging it (after I run manual tests). The only thing I'm not sure about is whether it's worth it from the POV that this will be provided by the booking API in the future version - so we'll have to revert and re-implement. But I guess that's fine, wdyt?

It feels that we will reimplement/remove a lot when the v4 API is ready, including this. My main concern is that we add one more network call which slows the implementation down a bit

samiuelson (approved):
LGTM!

PR #15439, opened 2026-02-26
[WOOMOB-2277] Prevent crash when re-adding a hidden site

irfano:
### Description
Fixes WOOMOB-2277

When a user enters a site address that already exists in the account but is currently hidden in Site Picker, the app can crash during the post-add auto-login flow.

In practice, the user usually does not realize that store is hidden and simply enters the same address again. This branch updates that behavior: if the re-entered site already exists but is hidden, it is restored to visible before continuing, instead of ending in a crash path.

### Test Steps
1. Please validate the reproduction first on `trunk`.
2. On `trunk`, sign in with an account that has at least two Woo sites.
3. Open Site Picker, tap **Edit**, hide one Woo site, and save.
4. In Site Picker, tap **Add a store**.
5. Enter the same site address as the hidden site and complete the flow.
6. Confirm `trunk` can hit the crash scenario (`NoSuchElementException` in Site Picker auto-login path).
7. Switch to `issue/WOOMOB-2277-fix-site-switching-crash` and repeat the same flow.
8. Confirm there is no crash and the previously hidden site is visible again in Site Picker.

### Images/gif
**Crash record**

[Screen_recording_20260226_160340.webm](https://github.com/user-attachments/assets/1c76f49c-00a3-4ff1-8b90-c574f8779cf0)


hichamboushaba (approved):
Great catch, thanks for fixing it @irfano, works as expected.

hichamboushaba:
Before merging this, do you want maybe to add a release note about it?

PR #15432, opened 2026-02-25
Derive booking payment status from order data

JorgeMucientes:
## Description

Closes: WOOMOB-2315

Equivalent PR on iOS: https://github.com/woocommerce/woocommerce-ios/pull/16725

The bookings plugin conflates booking status with payment status. When an order is marked as "completed", the booking status changes to "complete" — but the payment status shown in the app was derived from booking status, which incorrectly displayed "unpaid" for completed bookings.

This PR derives payment status from the **order data** instead of the booking status field. This matches how CIAB desktop admin and POS already work. However, POS has a slight difference, it doesn't handle the `partial refund` label so in step number 4 of testing steps, when refunding the order partially, both of the appointments will show as fully refunded <-- @kidinov is that intentional?

Desktop experience:

Screenshot 2026-02-26 at 07 23 24

Screenshot 2026-02-26 at 07 26 29

booking_list_all_tab_payment_status

### Changes

1. **New shared `PaymentStatusResolver`** in the bookings feature package — replicates the logic POS already uses (checks `refundTotal`, `datePaid`, `orderStatus` from `OrderEntity`)
2. **New shared `PaymentStatus` enum** (`PAID`, `UNPAID`, `FAILED`, `REFUNDED`, `PARTIALLY_REFUNDED`) in the bookings package
3. **POS now uses a typealias** to the shared `PaymentStatus` — `WooPosPaymentStatusResolver` delegates to the shared resolver, zero behavior change
4. **Removed `BookingStatus` sealed interface** — replaced with `PaymentStatus` enum + `isCancelled: Boolean` in `BookingSummaryModel`
5. **`BookingMapper`** now receives `PaymentStatus` as a parameter (resolved by ViewModels)
6. **`BookingDetailsViewModel`** and **`BookingListViewModel`** call `PaymentStatusResolver.resolve(orderId)` to compute status
7. **`BookingPaymentSection`** "Mark as Paid" button visibility now driven by `PaymentStatus.UNPAID`
8. **New status tags**: Refunded, Partially Refunded, Failed

No extra network calls needed — `BookingsStore.fetchBookings()` already fetches linked orders.

## Test Steps

1. Create a booking in WooCommerce with an unpaid order → verify "Unpaid" tag in booking list and detail
2. Mark the order as paid → verify "Paid" tag appears
3. Process a full refund → verify "Refunded" tag
[...]

PR #15371, opened 2026-02-17
[CIAB:POS] Implement booking list items as per design

kidinov:
### Description

Redesign booking list items to match the Figma spec:

- Replace booking status badges with payment status badges (Paid, Unpaid, Failed, Refunded, Partially Refunded) computed from order data
- Add separate Cancelled booking badge shown alongside payment status
- Map Unknown attendance state to Unattended
- Fall back to email when customer name is missing in the subtitle
- Allow subtitle to wrap to 2 lines

Payment status is resolved from `OrderEntity` fields (`refundTotal`, `datePaid`, `status`) via a new `WooPosPaymentStatusResolver`, avoiding a Room migration.

The badge logic is maybe no final therefore I create an issue to come back to it if needed

WOOMOB-2233

### Test Steps

1. Open Woo POS bookings screen
2. Verify payment status badges show correctly (Paid = gray, Unpaid/Failed = red)
3. Verify cancelled bookings show a blue "Cancelled" badge next to attendance and payment badges
4. Verify bookings without customer name show email in the subtitle
5. Verify long subtitles wrap to 2 lines instead of truncating on 1

image

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15440, opened 2026-02-26
[WOOMOB-2331] Fix payment status resolver using line item total instead of order total

kidinov:
### Description
Fixes WOOMOB-2331

`WooPosPaymentStatusResolver` was receiving the booking's line item total as `orderTotal`, but comparing it against the order-level `refundTotal`. For multi-booking orders with partial refunds, this incorrectly showed all bookings as fully refunded. Now the resolver reads `order.total` from `OrderEntity` directly.

### Test Steps
1. Add 2 appointments to the cart in the storefront (creates 2 bookings in the same order)
2. Complete the order—both bookings show as Paid in POS
3. Refund one of the bookings (partial refund at order level)
4. Verify both bookings show as **Partially Refunded** (not Refunded)

### Images/gif
image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/PaymentStatusResolver.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/PaymentStatusResolver.kt
@@ -9,13 +9,11 @@ class PaymentStatusResolver @Inject constructor(
     private val orderStore: WCOrderStore,
     private val selectedSite: SelectedSite,
 ) {
-    suspend fun resolve(orderId: Long, orderTotal: BigDecimal? = null): PaymentStatus {
+    suspend fun resolve(orderId: Long): PaymentStatus {
         val order = orderStore.getOrderByIdAndSite(orderId, selectedSite.get())
             ?: return PaymentStatus.UNPAID
 
-        val total = orderTotal
-            ?: order.total.toBigDecimalOrNull()
-            ?: BigDecimal.ZERO
+        val total = order.total.toBigDecimalOrNull() ?: BigDecimal.ZERO
         return computeStatus(order.refundTotal.abs(), total, order.datePaid, order.status)
     }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingViewStateMapper.kt
@@ -37,7 +37,6 @@ class WooPosBookingViewStateMapper @Inject constructor(
         val timeRange = formatTimeRange(booking)
         val paymentStatus = paymentStatusResolver.resolve(
             orderId = booking.orderId,
-            orderTotal = booking.order.paymentInfo?.total,
         )
 
         return WooPosBookingsState.BookingItemViewState(
@@ -72,7 +71,6 @@ class WooPosBookingViewStateMapper @Inject constructor(
 
         val paymentStatus = paymentStatusResolver.resolve(
             orderId = booking.orderId,
-            orderTotal = booking.order.paymentInfo?.total,
         )
 
         return WooPosBookingsState.BookingDetailsViewState(
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosPaymentStatusResolver.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosPaymentStatusResolver.kt
@@ -1,13 +1,12 @@
 package com.woocommerce.android.ui.woopos.bookings
 
 import com.woocommerce.android.ui.bookings.PaymentStatusResolver
-import java.math.BigDecimal
 import javax.inject.Inject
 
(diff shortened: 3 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka:
@kidinov the code looks good to me, however, I'm wondering about the behavior. Was this discussed somewhere? I'd expect a different behavior:
E.g. Order with Booking 1 and Booking 2 -> I refund Booking 1 -> Order is marked as partially refunded, Booking 1 as refunded, and booking 2 as paid.

Wdyt?

kidinov:
@malinajirka  👋

That comes from here: https://github.com/woocommerce/woocommerce-android/pull/15432

And yes, I think you are right. The problem is that to do this, we'd need to load all refunds for the order via `WCRefundStore`, match each `WCRefundItem.itemId` to 
  the booking's `orderItemId`, sum the per-item refund amounts, and compare against the line item total. It's doable but pretty hacky since we'd be reconstructing payment status from scattered data that wasn't
  designed for this.                                                                                                                                                                                               
                                                                                                                                                                                                                   
  The proper solution is the v4 bookings API, which already includes `payment_status` per booking (see [this comment](https://github.com/woocommerce/woocommerce-android/pull/15371#discussion_r2821286406) where Hicham shared the ciab-admin PR that adds it). Once we migrate to v4, we just read the field and don't need any of this local logic.

  For now I'd suggest we keep the current order-level approach as a "good enough" approximation. Multi-booking orders with partial refunds are an edge case (POS typically creates one booking per order), and
spending time on a local workaround that v4 will replace doesn't seem worth it. Wdyt?

JorgeMucientes:
Hey @kidinov given the changes added in [PR](https://github.com/woocommerce/woocommerce-android/pull/15432), I'd suggest basing the changes you applied in this PR to branch `issue/WOOMOB-2315-payment-status-booking-screens`. The reason is that I've refactored [PaymentStatusResolver](https://github.com/woocommerce/woocommerce-android/pull/15432/changes#diff-267e41d3ff0e8d3af8f1e92e80d87e88728565fef54c00aa15b9badad7b994cbR8) so that it's used both in POS and Booking list in store management. I kept the POS logic as it was, simply moved the file so we could reuse that. 
Wdyt?

malinajirka:
The API doesn't handle this either. I refunded the first booking (12-12:15am) but it cancelled the second booking  (1-1:15am). Moreover, I can see `collect payment` on the cancelled/refunded booking now -> when I tap on collect payment, it allowed me to charge for the whole order again (each booking is $11, there are 2 bookings total, but I ended up with net payment $33 (4x11 booking - 1x11 refunded booking. => 33).

Screenshot 2026-02-26 at 11 01 01


I think we need to somehow disable orders with more than one booking in the app - otherwise, these flows will be broken in this version of the app forever. Wdyt?

malinajirka:
Perhaps, we could at least hide `collect payment` on partially refunded order/bookings => that would solve at least on part of the issue.

kidinov:
@JorgeMucientes 

> Hey @kidinov given the changes added in https://github.com/woocommerce/woocommerce-android/pull/15432, I'd suggest basing the changes you applied in this PR to branch issue/WOOMOB-2315-payment-status-booking-screens. The reason is that I've refactored [PaymentStatusResolver](https://github.com/woocommerce/woocommerce-android/pull/15432/changes#diff-267e41d3ff0e8d3af8f1e92e80d87e88728565fef54c00aa15b9badad7b994cbR8) so that it's used both in POS and Booking list in store management. I kept the POS logic as it was, simply moved the file so we could reuse that.
Wdyt?

Good point. Pointing on your branch now. Removed `orderTotal` as param from `PaymentStatusResolver` as well

> I think we need to somehow disable orders with more than one booking in the app - otherwise, these flows will be broken in this version of the app forever. Wdyt?

Disabled orders is also sort of a broken flow too, and this also will be forever in this version as well. I think we should limit our sets of workarounds to the API bugs. Let's see what Katie says, but I think the right call for now will be to not support creating multiple bookings in 1 order on the customer facing frontend

The collect button for the canceled booking is expected - 0fzXPGPRTPyjU3ExPLW6To-fi-799_11433#1629382638. I am not sure about partially refunded status though, but considering the above, I would not do anything just yet

malinajirka:
> Disabled orders is also sort of a broken flow too, and this also will be forever in this version as well.

It is not the best UX but it is not a broken flow -> e.g. merchant can't overcharge the customer, whereas in the current state of the app they can.

hichamboushaba:
> The proper solution is the v4 bookings API, which already includes payment_status per booking (see https://github.com/woocommerce/woocommerce-android/pull/15371#discussion_r2821286406 where Hicham shared the ciab-admin PR that adds it). Once we migrate to v4, we just read the field and don't need any of this local logic.

Just adding here one point, the API behaves the same, so migrating to it won't change the behavior, CIAB admin shows the same thing as the apps with a similar scenario:

Screenshot 2026-02-26 at 15 25 58

I think this could be by design: the payment is related to the order, not the booking, so when we say "partially refunded", we mean the order. The merchant can differentiate the refunded booking from the others by checking the canceled tags.
I am unsure if this solution is clear from a UX perspective, but I wanted to share my understanding of the situation.

PR #15446, opened 2026-02-26
[WOOMOB-2341] Use 1-second inclusive bounds for bookings date range

irfano:
### Description
Fixes WOOMOB-2341
This updates bookings date-range query normalization from +/-1 minute to +/-1 second when building request parameters.
The matches the duration with iOS.

### Test Steps
1. Log in to a CIAB site.
2. Note a time from a booking, (for example at`07:09` on a given date, `2026-02-22 07:09:00`).
3. Open Bookings (or POS Bookings) and set the date filter start to `07:09`.
4. Verify the `07:09` booking is visible in the list.
5. Change start to `07:10` and verify the same booking is no longer included.

### Images/gif
N/A

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -10,7 +10,6 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
 import org.wordpress.android.fluxc.utils.extensions.filterNotNull
 import org.wordpress.android.fluxc.utils.extensions.putIfNotNull
-import java.time.ZoneOffset
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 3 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
```diff
@@ -162,13 +162,13 @@ class BookingsRestClient @Inject constructor(
         if (paymentStatus != null) TODO()
         if (customer != null) set("customer", customer.customerId.toString())
         if (dateRange != BookingsFilterOption.DateRange.DEFAULT) {
```
malinajirka, line 163: Have you considered moving this out of the Rest Client? It feels like a logic specific to a single screen/use-case.

For example, in the POS we request bookings for a specific date -> our request is `after=00:00` `before=23.59.59` => if we add a second, we end up with `before=00:00:01 => which would include booking from a different date. 

If you believe it belongs here, we can pass an extra parameter for POS to skip it. Let me know, thanks!

malinajirka: Actually, thinking about it, the range in POS also isn't optimal -> we should likely do `before 00:00 of the next day` and `after 23:59:999 of the previous day` - but still feels like it should be fixed in the VM layer where we create the range.

irfano: > Have you considered moving this out of the Rest Client?

I was already working on it, and sorry for requesting review too early. 😅 
I moved the adjustment out of `BookingsRestClient` into `BookingsStore`, so the REST client now only serializes raw filter values.

image

The +/-1s adjustment is still needed as a workaround for API semantics: we currently have `after`/`before`, but not an inclusive `on` filter that web currently has.
So we’re essentially tweaking the query parameters to support the `on` option as well. I’d avoid doing this in the VM, because these adjusted values are being used only for endpoint calls. If we do it in the VM, we risk leaking manipulated dates into UI state or cached filters. Keeping it close to the API call is safer.

Does this approach look reasonable to you?

irfano: Here is the Slack discussion (p1763116036412419-slack-C09FHQNQERG) where we decided on the logic.

malinajirka: 👍 Cool, thanks!

malinajirka (approved):
Works as expected, and I can confirm it fixes even the behavior in the POS, thank you!

PR #15451, opened 2026-02-27
[WOOMOB-2339] Hide attendance badge on cancelled bookings and refactor AttendanceSection

samiuelson:
### Description
Fixes WOOMOB-2339

Cancelled bookings were incorrectly showing an attendance status badge. This PR hides the attendance badge when a booking is cancelled.

### Test Steps
1. Open POS bookings screen
2. Find a cancelled booking — verify no attendance badge (Attended/Unattended) is shown in the list item or details
3. Find a non-cancelled booking — verify the attendance badge still appears
4. Toggle attendance on a non-cancelled booking — verify it updates optimistically and persists after reload
5. Toggle attendance, then simulate a network failure — verify the selection reverts

### Images/gif
https://github.com/user-attachments/assets/7402a798-7b58-426e-84bf-7b3e7b00e31a


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsState.kt
```diff
@@ -102,7 +106,7 @@ sealed class WooPosBookingsState {
         val isSelected: Boolean,
         val paymentStatus: PaymentStatus,
         val isCancelled: Boolean,
-        val attendanceBadge: AttendanceState = AttendanceState.UNATTENDED,
+        val attendanceBadge: AttendanceState? = null,
```
kidinov, line 109: I understand that it doesn't make sense to change attendance after a booking was canceled, but I’m not sure why we would need to do special work and hide attendance status if it's canceled.

Let’s say a person came to the business, they were marked as "Attended" and then something happened and it had to be canceled. Why hide this fact that the person showed up? I think in some cases it may be useful for a merchant to know that. What?

kidinov: Only reason I see we could hide it is that in the list it’s not  spaced enough, but IMO we should still show it in the title of the booking details

We do show it on the web

Image

kidinov (approved):
LGTM! I left one suggestion.

Also, do you want to merge it as hotfix? If you do then we should target release branch but do not release it yet as it will be another branch made of the release one later.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsState.kt
samiuelson: Fair point. I think it would be good to have attendence "badge" view somewhere in booking details in case of canceled booking. As for the attendence toggle, I think it's desired to hide it which is consistent with ciab admin, where it's impossible to edit attendence if the booking has been canceled.

PR #15452, opened 2026-02-27
[WOOMOB-2347] Optimize booking location loading

kidinov:
### Description
Fixes WOOMOB-2347

Do not merge because probably we will need to rebase that to the hotfix branch

Optimizes booking location loading and removes the loading flash when switching dates.

**Location caching:** Previously, the location cache was cleared on every date change, forcing a re-fetch even for already-seen products. Since locations are per-product (not per-date), the cache now persists across date changes. Bookings are also shown immediately while missing locations are fetched in the background.

**No loading flash on date change:** Previously, switching dates would briefly show a loading state because `Items.Loading` was set before the Room flow re-queried. Now the current bookings stay visible until the new ones arrive from the database flow. The "nothing found" state is also suppressed while a fetch is still in progress.

@malinajirka wdyt about the second one? It feels it's better to it like that when we switch between dates with bookings, but when there is a date without booking the it feels a bit off. Not sure

### Test Steps
1. Open POS Bookings screen
2. Navigate between dates with bookings that share the same products
3. Verify bookings appear immediately without a loading flash
4. Verify locations show up correctly (either instantly from cache or shortly after)
5. Navigate to a date with no bookings — verify "nothing found" appears after fetch completes, not before

### Images/gif

https://github.com/user-attachments/assets/62ada4de-0a16-49a0-89a5-20076149fd56

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
@@ -161,9 +161,10 @@ class WooPosBookingsViewModel @Inject constructor(
                     return@collectLatest
                 }
 
-                val isFetchingContent = current is WooPosBookingsState.Content &&
-                    (current.items is WooPosBookingsState.Content.Items.Loading || fetchJob?.isActive == true)
-                if (bookings.isEmpty() && isFetchingContent) {
+                if (bookings.isEmpty() && current is WooPosBookingsState.Content && fetchJob?.isActive == true) {
+                    _state.value = current.copy(
+                        items = WooPosBookingsState.Content.Items.Loading,
+                    )
                     return@collectLatest
                 }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
@@ -105,8 +105,7 @@ class WooPosBookingsViewModel @Inject constructor(
                             message = error.message ?: "Failed to load bookings"
                         )
                     }
-                    current is WooPosBookingsState.Content &&
-                        current.items is WooPosBookingsState.Content.Items.Loading -> {
+                    current is WooPosBookingsState.Content -> {
                         _state.value = current.copy(
                             items = WooPosBookingsState.Content.Items.Error(
                                 title = resourceProvider.getString(
@@ -126,8 +125,7 @@ class WooPosBookingsViewModel @Inject constructor(
                     current is WooPosBookingsState.Loading -> {
                         _state.value = buildNothingFoundState()
                     }
-                    current is WooPosBookingsState.Content &&
-                        current.items is WooPosBookingsState.Content.Items.Loading -> {
+                    current is WooPosBookingsState.Content -> {
                         _state.value = current.copy(
                             items = WooPosBookingsState.Content.Items.NothingFound(
                                 title = resourceProvider.getString(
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/281">24.2</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
```diff
@@ -227,10 +224,44 @@ class WooPosBookingsViewModel @Inject constructor(
...
+                }
+                if (hasMissingLocations) {
+                    fetchBookingLocations(bookings)
+                    rebuildStateWithLocations(bookings, resourcesMap)
+                }
             }
         }
     }
 
+    private suspend fun rebuildStateWithLocations(
```
malinajirka: 🔍 I'd consider moving this method lower in the class so the class reads more from top to bottom.

malinajirka (commented):
Thanks!

> No loading flash on date change: Previously, switching dates would briefly show a loading state because Items.Loading was set before the Room flow re-queried. Now the current bookings stay visible until the new ones arrive from the database flow. The "nothing found" state is also suppressed while a fetch is still in progress.

I'm personally not a fan of this approach - I like to see something is happening as a user. I'd personally prefer 200ms artificial loading to avoid flashes, rather then not having a progress indicator at all.

malinajirka (approved):

malinajirka:
I also realized one more issue - if we don't display any loading, what happens if there are identical bookings on two days -> we don't display booking date on the row items, so the user won't even know if the data has been replaced already or not. This could be happening quite often, considering many services are full or have peak hours and allow booking only specific time slots.

kidinov:
@malinajirka 

Good point about the identical bookings case; I didn't think about that.
                                                                                                                                                                                                                   
The actual root cause of the delay is flatMapLatest in BookingListHandler. When filters change, it cancels the old Room Flow and re-subscribes to a new one. Even though data is already in the DB, Room's flow subscription is async (query runs on Room's executor thread), so there's a gap where no data is available. That gap is what was causing the flash.

I fixed it differently now:
  - Added a one-shot DB query that runs before subscribing to the Room flow. So inside flatMapLatest, we first emit the cached snapshot, then subscribe to the flow for ongoing updates.
  - In the ViewModel, if the snapshot comes back empty (data not cached yet), we show Items. Loading properly instead of keeping old bookings.

  So now:
  - Data cached → shows instantly, no loading, no flash
  - Data not cached → loading indicator until the fetch populates the DB

That's all ugly code-wise but it appears that the only way to solve it is if we go with observing the database

5047e348f2046cc22d671f0caca34b91333ee5c1

4327e8921ffbfc35f9044fb260858c95e27decc7

wdyt?



https://github.com/user-attachments/assets/c4a9bb26-960b-47f4-b211-41b951fe1086

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
kidinov: Moved it in da3b2bcd8c1 - rebuildStateWithLocations is now right after fetchBookingLocations.

malinajirka:
> Even though data is already in the DB, Room's flow subscription is async (query runs on Room's executor thread), so there's a gap where no data is available. That gap is what was causing the flash.

Hmm, it's quite weird, especially since on Coupons I don't see any flash and we observe the database there as well.

> That's all ugly code-wise but it appears that the only way to solve it is if we go with observing the database

TBH I still prefer the loading state more as a user - it's clear that something has happened after I tapped on the button. Depends on how fast the flash is on your side, on my side it's roughly 100ms - so it is not super super fast which would look broken but this feels just like the app was doing something (loading data from db) and now presents the result. 🤔 

Either case, I'm fine merging this if you believe it improves the UX, but I would suggest avoiding such change as a hotfix. This feels like a risky change to how, when, and what we load to the UI. Wdyt?

PR #15453, opened 2026-02-27
[WOOMOB-2342] Cancel previous attendance update request on rapid taps

samiuelson:
### Description
Fixes WOOMOB-2342

Tracks the attendance update coroutine job and cancels any in-flight request before launching a new one. This prevents duplicate concurrent API calls when the user taps the attendance toggle rapidly.

### Test Steps
1. Open a booking with an editable attendance status (Confirmed/Paid)
2. Tap / switch the attendance toggle rapidly several times
3. Verify no glitches or delays are visible

### Images/gif
N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
```diff
@@ -135,18 +135,18 @@ class BookingsRepository @Inject constructor(
     suspend fun updateAttendanceStatus(
         bookingId: Long,
         attendanceStatus: BookingEntity.AttendanceStatus,
-    ): Result<Unit> = appCoroutineScope.async {
```
samiuelson, line 150: 💡 appCoroutineScope ignores parent coroutine scope from VM.

malinajirka: I'm guessing here, but I think the app scope might have been intentional here - we still want to change the attendance if the user taps and leaves the screen, wdyt? I'm fine removing it though, since it's much easier to reason about and if the request was sent - it doesn't matter that the app doesn't receive the result, technically the operation will still complete.

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
samiuelson: I used app-scoped coroutine in the other place this function is called, for consistency: [d5630f8](https://github.com/woocommerce/woocommerce-android/pull/15453/commits/d5630f8510b79441cbc9971663c11ec166ba00f7)

PR #15454, opened 2026-02-27
[WOOMOB-2350] Show order totals on POS card payment screen

malinajirka:
Fixes WOOMOB-2350

Do not merge label - this will likely be merged into a hotfix branch when it is available.

### Description

- Adds an order summary section (subtotal, discount, taxes, total) to the POS card payment screen. The summary is shown in all `Collecting` states (preparing, ready for payment, reader disconnected) and is loaded from Room using existing infrastructure. 

- Changes the booking detail "Collect Payment · $X" button to just "Collect payment" in a hotfix-compatible way (no string changes).

- Temporarily hides Payment section on Booking detail to avoid confusing UX until we discuss a proper UX. 

### Test Steps
1. Open POS Bookings → select an unpaid booking (Notice Payment section is not visible)
2. Tap "Collect Payment" (Noticed amount is not visible on the button)
3. Verify the order summary rows (Subtotal, Taxes, Total) appear below the status icon/text
5. Disconnect the card reader → verify the order summary still shows on the disconnected state
6. Verify the order summary does NOT appear on payment-in-progress or payment-failed states

### Images/gif

https://github.com/user-attachments/assets/328d46ac-a9c6-4bc5-89dd-077adb1f4304


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/281">24.2</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
```diff
@@ -435,16 +463,90 @@ private fun CardPaymentCenteredLayout(
         ) {
             content()
         }
+        summary()
```
kidinov: Shall it be centered between the content and the buttons?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
```diff
@@ -421,6 +448,7 @@ private fun CardPaymentFailed(
 private fun CardPaymentCenteredLayout(
     modifier: Modifier = Modifier,
     content: @Composable () -> Unit,
+    summary: @Composable ColumnScope.() -> Unit = {},
```
kidinov: What's the point of having default value here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/details/WooPosBookingDetails.kt
```diff
@@ -57,6 +57,8 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosSpa
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTheme
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
 
+private const val SHOW_PAYMENT_SECTION = false
```
kidinov, line 60: In my opinion, for such things, we should use a version control system instead of relying on dead code with a flag

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
malinajirka: I've created an issue for this - https://linear.app/a8c/issue/WOOMOB-2381/pos-booking-payment-screen-totals-section-should-be-vertically. It requires a bit more changes to the layout so I didn't want to include it as a hotfix.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
malinajirka: Updated in 9941084a687ca2f7923cccab7370563b77613b20

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -93,6 +106,24 @@ class WooPosCardPaymentViewModel @Inject constructor(
         }
     }
 
+    private fun loadOrderTotals() {
+        viewModelScope.launch {
+            val orderEntity = orderStore.getOrderByIdAndSite(orderId, selectedSite.get())
```
kidinov: CC's comment -
```
ViewModel directly accesses WCOrderStore, OrderMapper, and SelectedSite, bypassing the repository layer (AGENTS.md says "ViewModels never interact with Room or network directly — always through repositories")                                                                                                        
                                                                                                                                                                                                                                                                                                                             
  loadOrderTotals() calls orderStore.getOrderByIdAndSite() directly from the ViewModel without withContext(Dispatchers.IO), meaning the Room DB query runs on the main thread. Every other equivalent call in the woopos module (WooPosTotalsRepository, WooPosCashPaymentRepository, WooPosEmailReceiptRepository) wraps
  this in withContext(Dispatchers.IO) inside a repository. The fix is to route through an existing repository (e.g. WooPosTotalsRepository.getOrderById) or create one.
```

Probably make sense?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/details/WooPosBookingDetails.kt
malinajirka: I would generally agree, but since we'll re-enable it as soon as we merge to trunk, it feels like an overkill. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -57,6 +66,9 @@ class WooPosCardPaymentViewModel @Inject constructor(
     private val _navigationEvent = MutableSharedFlow<WooPosNavigationEvent>()
     val navigationEvent: SharedFlow<WooPosNavigationEvent> = _navigationEvent.asSharedFlow()
 
+    private val _orderTotals = MutableStateFlow<WooPosOrderTotalsViewState?>(null)
```
kidinov: Consider merging orderTotals into WooPosCardPaymentState instead of keeping it as a separate flow. Two independent flows can get out of sync (e.g. totals load slower than the state transitions to Collecting, causing the summary to pop in). Also, orderTotals being nullable adds unnecessary optionality the order  
  is already in Room at this point, so it should always be available.
                                                                                                                                                                                                                                                                                                                            

```
  sealed class WooPosCardPaymentState {                                                                                                                                                                                                                                                                                      
      data object Initiating : WooPosCardPaymentState()                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                             
      sealed class Collecting : WooPosCardPaymentState() {                                                                                                                                                                                                                                                                   
          abstract val orderTotals: WooPosOrderTotalsViewState

          data class Preparing(
              val title: String,
              val subtitle: String,
              override val orderTotals: WooPosOrderTotalsViewState,
          ) : Collecting()

          data class ReadyForPayment(
              val title: String,
              val subtitle: String,
              override val orderTotals: WooPosOrderTotalsViewState,
          ) : Collecting()

          data class ReaderDisconnected(
              val title: String,
              val subtitle: String,
              val actionButtonLabel: String,
              override val orderTotals: WooPosOrderTotalsViewState,
          ) : Collecting()
      }

      data class PaymentInProgress(
          val title: String,
          val subtitle: String,
      ) : WooPosCardPaymentState()

      data class PaymentFailed(
          val title: String,
          val subtitle: String,
          val actionButtonLabel: String? = null,
          val isDismissButtonVisible: Boolean,
      ) : WooPosCardPaymentState()
  }
```


Also, I'd keep `WooPosOrderTotalsViewState` in this file - seems unnecessarily extracted to a new one being part of the same screen state

kidinov (approved):
LGTM! I left a couple of suggestions - please take a look

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
malinajirka: Good call - fixed in 37de395fd46660ff5df2dba95bed58f7daf84ca2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/details/WooPosBookingDetails.kt
kidinov: Basically it's `git revert hash` VS adding a new code here and then removing it. I'd say git revert is easier but no big deal obviosly

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
malinajirka: > Also, orderTotals being nullable adds unnecessary optionality the order
is already in Room at this point, so it should always be available.

Hmm, this made me realize that there is no guarantee the order is in database 😞 . 

I like the idea of merging the states though. For some reason I thought that we can't merge them since we need to show the totals when reader is not connected - but that's exactly what `ReaderDisconnectedState` is for, so merging it is a good idea!

I'll think about this tomorrow - we fetch the order in Preparing state, but only when reader is connected.

malinajirka: I've updated this, could you please review the new commits @kidinov

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -57,14 +61,44 @@ class WooPosCardPaymentViewModel @Inject constructor(
...
     init {
-        observeCardReaderStatus()
+        viewModelScope.launch {
+            if (loadOrderTotals()) {
+                observeCardReaderStatus()
+            }
+        }
+    }
+
+    private suspend fun loadOrderTotals(): Boolean {
```
kidinov: np: that a bit of a code smell that it returns Boolean - It's using a return value for flow control, essentially saying "did it work?" so the caller knows whether to continue. This mixes two concerns: loading data and controlling init flow. 

Maybe return `WooPosOrderTotalsViewState?` instead of `Boolean`. The caller decides what to do with the result. This removes the side effect of setting orderTotals inside the function and makes the data flow explicit:

```
  init {
      viewModelScope.launch {
          val totals = loadOrderTotals()
          if (totals != null) {
              orderTotals = totals
              observeCardReaderStatus()
          }
      }
  }

  private suspend fun loadOrderTotals(): WooPosOrderTotalsViewState? {
      val order = cardPaymentRepository.fetchOrGetOrder(orderId) ?: run {
          _state.value = WooPosCardPaymentState.PaymentFailed(...)
          return null
      }
      return WooPosOrderTotalsViewState(
          subtotal = priceFormat(order.productsTotal),
          ...
      )
  }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
```diff
@@ -451,11 +465,91 @@ private fun CardPaymentCenteredLayout(
             ) {
                 content()
             }
+            summary()
```
kidinov: On some screens (this one, by the way, is quite large), we have the text that doesn't fit


Image

malinajirka: Tbh, I'm not sure there is a solution other than making it scrollable which sucks... There simply might not be enough space. I believe the size of the display doesn't matter here, what matters is the ratio.

kidinov:
```
@Composable
private fun CardPaymentCenteredLayout(
    onBackClicked: (() -> Unit)? = null,
    content: @Composable () -> Unit,
    summary: @Composable ColumnScope.() -> Unit,
    buttons: @Composable ColumnScope.() -> Unit,
) {
    Column(
        modifier = modifier.fillMaxSize(),
        horizontalAlignment = Alignment.CenterHorizontally,
    ) {
        if (onBackClicked != null) {
            WooPosToolbar(
                titleText = stringResource(R.string.woopos_card_payment_toolbar_title),
                onBackClicked = onBackClicked,
            )
        } else {
            Spacer(modifier = Modifier.height(56.dp))
        }
        Column(
            modifier = Modifier
                .fillMaxSize()
                .padding(
                    top = WooPosSpacing.XLarge.value,
                    bottom = WooPosSpacing.XXXLarge.value,
                ),
            horizontalAlignment = Alignment.CenterHorizontally,
        ) {
            content()
            Spacer(modifier = Modifier.weight(1f))
            summary()
            Spacer(modifier = Modifier.weight(1f))
            buttons()
        }
    }
}
```

Maybe something simple like that would largely work for both issues (non centered summary as well).


Image

kidinov: But yeah, content has to be scrollable if there is not enough vertical height

malinajirka: The example you shared will somewhat work - but looks off when the `summary` is not visible -> e.g. payment failed screen.

kidinov: Yeah, for this case we'd need to make summary nullable and when it's null then to adjust the layout depends on that

malinajirka: okay, I think I got it - let me quickly test.

malinajirka: @kidinov Tbh I couldn't come up with a solution without downsides. I ended up putting the content into a scrollview to ensure it is usable on all devices with all font sizes.

https://github.com/user-attachments/assets/eab4d636-ae2c-4d65-94f6-7ef7cbd6d337


https://github.com/user-attachments/assets/cab8a042-ddb8-46ab-b29a-5962a896d003

I also played with keeping the buttons always visible on a sheet, and let only the content + buttons scroll -> similarly to what we do with Collect Payment button on the Booking detail. That worked pretty well too, but I'm not sure it's any better since then we can't evenly distribute the content/totals/button when there is enough space for all.

Wdyt @kidinov ?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
malinajirka: Updated in  77448d4b0d1c52915347e52c937f83c847ac5b13

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentScreen.kt
kidinov: [Refactor_card_payment_layout_to_use_spread_constraints_and_improve_scrolling_behavior.patch](https://github.com/user-attachments/files/25737822/Refactor_card_payment_layout_to_use_spread_constraints_and_improve_scrolling_behavior.patch)


What do you think about this approach? It's scrollable content when there is no space available, and spreads when there is enough. It's more complex than what I expected, though. 🥲

malinajirka: I like the idea, but it still cuts-off the text in some cases.

Screenshot 2026-03-04 at 12 48 26

I also tried implementing a smart image resize - resize the image between 120-240dp based on available space and make the content scrollable if it still doesn't fit with 120dp sized image. However, it requires an extra recomposition to calculate the sizes and is very complex, so imo not worth it -> considering, we'll likely replace this screen with the regular POS checkout flow in the upcoming months.

So if you agree, I'll go with the solution that is currently pushed in this branch, wdyt @kidinov ?

kidinov: > I like the idea, but it still cuts-off the text in some cases.

Yes, it does, but that scrollable is therefore reachable. We have a trade-off between cutting the buttons, which are actionable, and have to be clicked almost always, and cutting the text, which in most of the cases is not needed, as the merchant would just know what to do as they would learn it from the first go

Having said that, with our current minimum screen size requirement, the cash button seems to be always at least partially visible even in the current implementation, so I think it's not a big deal

> So if you agree, I'll go with the solution that is currently pushed in this branch, wdyt @kidinov ?

Sure, lets go ahead

malinajirka: > Yes, it does, but that scrollable is therefore reachable. 

I see what you mean, in that case, I'd personally prefer the approach where buttons are on a sheet and are always visible -> with totals and card reader section being scrollable. Same as Collect Payment button behaves on the Booking detail. While I see a point that Totals is more important then Reader info, it feels weird when it's cut off like this in the middle of the screen.

We'd need to combine your approach with what [I tried above](https://github.com/woocommerce/woocommerce-android/pull/15454#discussion_r2882345843) and I believe it might work. However, considering `with our current minimum screen size requirement, the cash button seems to be always at least partially visible` I'm inclined towards merging this as is.

PR #15460, opened 2026-03-02
[WOOMOB-2344] Fetch bookings data only when local cache is empty

hichamboushaba:
### Description
Fixes WOOMOB-2344

Previously, `ObserveBookingsVisibility` always fetched bookable products and bookings from the network on every subscription start (via `onStart`). This caused an overhead of two additional requests on each app launch.

This PR changes the logic to make the fetch only when the cache is empty, the downside of this is that in case all bookable products or bookings get deleted (a rare case IMO), then we'll keep displaying the bookings tab until the user opens products or bookings screen. I think the advantages outweigh this downside, WDYT?

### Test Steps
1. Start with a fresh install.
2. Open the app on a CIAB site with bookable products
3. Check the logs and confirm `BookingsStore: fetchBookings` is logged.
4. Confirm the bookings tab appears after the fetch.
6. Close and re-open the app.
7. Confirm `BookingsStore: fetchBookings` is not logged.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsVisibilityTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsVisibilityTest.kt
@@ -17,6 +17,7 @@ import org.mockito.kotlin.any
 import org.mockito.kotlin.anyOrNull
 import org.mockito.kotlin.eq
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.wordpress.android.fluxc.model.SiteModel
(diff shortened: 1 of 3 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsVisibilityTest.kt
irfano: This test name refers to `onStart`, but we no longer use it in `ObserveBookingsVisibility`.

irfano (approved):
LGTM! It's a clean fix.

> I think the advantages outweigh this downside, WDYT?

Agreed!

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/bookings/tab/ObserveBookingsVisibilityTest.kt
hichamboushaba: Good catch, test removed.

PR #15462, opened 2026-03-02
[WOOMOB-2379] Hide Connect Store step when site is already connected

hichamboushaba:
Fixes WOOMOB-2379

### Description

When a site is already connected to Jetpack, the push notification setup flow now hides the "Connect Store" step entirely instead of showing it as auto-completed. The screen also displays context-appropriate title and subtitle text — "Set up push notifications" instead of "Connect to WordPress.com".

### Test Steps
1. Log into a site that is already connected to Jetpack (e.g., a Jurassic Ninja site)
2. Enable Push Notifications feature flags
3. Tap the push notification banner on the My Store screen
4. On the introduction dialog, tap "Update plugin"
5. Verify the connection steps screen shows:
   - Title: "Set up push notifications"
   - Subtitle: "Please wait while we set up push notifications for {site}."
   - Only 2 steps: "Check plugin compatibility" and "Enable push notifications" (no "Connect Store" step)

### Images/gif
Screenshot_20260302_134116


irfano (approved):
LGTM! 👍🏻

PR #15465, opened 2026-03-03
[WOOMOB-2347] Optimize booking location loading

kidinov:
Continuation of #15452 (closed because `release/24.2` branch was deleted), now targeting `release/24.2.1`.

### Description
Fixes WOOMOB-2347

Optimizes booking location loading and removes the loading flash when switching dates.

**Location caching:** Previously, the location cache was cleared on every date change, forcing a re-fetch even for already-seen products. Since locations are per-product (not per-date), the cache now persists across date changes. Bookings are also shown immediately while missing locations are fetched in the background.

**No loading flash on date change:** Previously, switching dates would briefly show a loading state because `Items.Loading` was set before the Room flow re-queried. Now the current bookings stay visible until the new ones arrive from the database flow. The "nothing found" state is also suppressed while a fetch is still in progress.

### Test Steps
1. Open POS Bookings screen
2. Navigate between dates with bookings that share the same products
3. Verify bookings appear immediately without a loading flash
4. Verify locations show up correctly (either instantly from cache or shortly after)
5. Navigate to a date with no bookings — verify "nothing found" appears after fetch completes, not before

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingsRepository.kt
@@ -61,6 +61,18 @@ class BookingsRepository @Inject constructor(
             order = order
         )
 
+    suspend fun getBookingsList(
+        limit: Int? = null,
+        filters: BookingFilters? = null,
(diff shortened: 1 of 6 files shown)


malinajirka (approved):
LGTM, thanks!

PR #15471, opened 2026-03-03
[CIAB] Add analytics tracking to booking screens

hichamboushaba:
### Description
Fixes WOOMOB-1999

Adds analytics tracking to the non-POS booking screens (list, detail, and filter) to align with the CIAB Mobile Tracking Plan. This includes renaming existing main-tab booking events from past tense to present tense (`_selected` → `_select`) and adding 15 new analytics events covering user interactions across the booking list, booking detail, and booking filter screens.

### Test Steps
1. Open the Bookings tab → verify `main_tab_bookings_select` is tracked
2. Tap the Bookings tab again while already on it → verify `main_tab_bookings_reselect` is tracked
3. Verify `booking_list_view` fires automatically with `selected_tab`, `is_default_tab`, `is_list_empty`, `is_filtered`
4. Switch between Today/Upcoming/All tabs → verify `booking_list_tab_select` with `selected_tab` property
5. Tap a booking → verify `booking_list_booking_tap` with `is_search_active`, `is_filtering_active`, `selected_tab`
6. Tap Sort → verify `booking_list_sort_by_tap` is tracked
7. Select a sort option → verify `booking_list_sort_by_option_tap` with `sort_option` property
8. Tap Filters → verify `booking_list_filters_tap` is tracked
9. Apply filters → verify `booking_list_apply_filters` with `selected_filters` property
10. Tap search → verify `booking_list_search_tap` is tracked
11. On booking detail, tap Add Note → verify `booking_detail_add_note_tap`
12. On booking detail, tap View Order → verify `booking_detail_view_linked_order_tap`
13. On booking detail, toggle attendance → verify `booking_detail_attendance_status_update` with `booking_status`
14. On booking detail, cancel booking → verify `booking_detail_cancel_booking`
15. Force a network error while fetching bookings → verify `booking_list_failed_to_fetch_bookings` with `error_code`, `error_description`
16. Force a network error while toggling attendance or cancelling → verify `booking_list_failed_to_update_booking_details` with `action`, `error_code`, `error_description`

### Images/gif
N/A — analytics-only changes, no UI modifications.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>BookingAnalyticsHelper</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/283">24.3</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -229,7 +228,20 @@ class BookingListViewModel @Inject constructor(
             searchQuery = fetchParams.searchQuery,
             filters = fetchParams.prepareFilters(),
             sortBy = fetchParams.sortOption
-        ).onFailure {
+        ).onSuccess {
+            trackBookingListView(fetchParams)
```
hichamboushaba: There is a pending question about when to track this event pe5sF9-4Mi-p2#comment-5277

hichamboushaba: I updated the logic in 20814f7aa23f7cea5865582930b44adf3da7b8fa to follow the approach suggested in the P2, but this is not confirmed 100%, so it may still be updated.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -295,6 +327,23 @@ class BookingDetailsViewModel @Inject constructor(
...
+        is WooException -> error.type.name
+        else -> javaClass.simpleName
+    }
+
     data class NavigateToOrder(val orderId: Long) : MultiLiveEvent.Event()
     data class NavigateToBookingNote(val bookingId: Long) : MultiLiveEvent.Event()
+
+    companion object {
+        private const val KEY_BOOKING_STATUS = "booking_status"
+        private const val KEY_ACTION = "action"
```
AdamGrzybkowski: It's okay for now, but if we build some quick actions to update bookings from the list, we will need a better way to structure the analytics-related code rather than having those private values.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
```diff
@@ -295,6 +327,23 @@ class BookingDetailsViewModel @Inject constructor(
...
+        BookingAttendanceStatus.Attended -> "attended"
+        BookingAttendanceStatus.Unattended -> "unattended"
+    }
+
+    private fun Throwable.errorCode(): String? = (this as? WooException)?.error?.apiErrorCode
+
+    private fun Throwable.errorType(): String? = when (this) {
+        is WooException -> error.type.name
+        else -> javaClass.simpleName
+    }
```
AdamGrzybkowski: 💡  I think it would be nice to reuse those in `BookingListViewModel`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -304,6 +340,38 @@ class BookingListViewModel @Inject constructor(
...
+    }
+
+    fun trackBookingListView() = launch {
+        val state = _state.first()
+
+        analyticsTrackerWrapper.track(
+            AnalyticsEvent.BOOKING_LIST_VIEW,
+            mapOf(
+                KEY_SELECTED_TAB to state.tabState.selectedTab.toAnalyticsValue(),
+                KEY_IS_DEFAULT_TAB to (!didUserSwitchTab).toString(),
```
AdamGrzybkowski: ⚠️ The default tab is `Today`, right? That means if the user opens the booking list, goes to `All` and then back to `Today`, we would send an event with `is_default_tab=false` which wouldn't be true. 

```
2026-03-05 11:20:04.790 18222-18389 WooCommerce-UTILS       com.woocommerce.android.dev          I  🔵 Tracked: woocommerceandroid_booking_list_view, Properties: {"selected_tab":"today","is_default_tab":"false","is_list_empty":"false",...}
```

Or am I understanding this param wrong?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
```diff
@@ -304,6 +340,38 @@ class BookingListViewModel @Inject constructor(
...
+                KEY_IS_LIST_EMPTY to state.contentState.bookings.isEmpty().toString(),
+                KEY_IS_FILTERED to (state.controlsState.enabledFiltersCount > 0).toString()
```
AdamGrzybkowski: ⚠️ This property has the wrong value after applying filters. It's likely invoked before we get the new value in `enabledFiltersCount`.

Steps: 
1. Open filters
2. Apply some and tap Show Bookings
3. Verify the event `woocommerceandroid_booking_list_view` has `"is_filtered":"false"`

AdamGrzybkowski (changes requested):
Mostly works but I think we have two issues (or at least one!) that need to be fixed before merging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
hichamboushaba: The property role is to say whether the user is still on the default tab or did change it, even if they return to the default one, they already did change it, there was a discussion about this here:
pe5sF9-4Mi-p2#comment-5275

(I don't have a strong opinion of the behavior, I just applied the plan that was already in place)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
hichamboushaba: Yes, I don't like it either. We define all keys in `AnalyticsTracker` and it doesn't scale well, I can move the keys there, but I think we are just making the situation worse by doing that.

I tried a different approach in the commit 52a7124, it's a bit similar to how iOS app does tracking, but I didn't go as far as defining extensions for all events. Let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
hichamboushaba: The commit 52a7124 that I added for the other issue handles this too.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/details/BookingDetailsViewModel.kt
AdamGrzybkowski: Looks good 👍 Thanks

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
AdamGrzybkowski: So I got it wrong, thanks for the explanation.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
hichamboushaba: Great catch @AdamGrzybkowski, the commit c3f5ed8 should fix this.

hichamboushaba:
Thanks @AdamGrzybkowski for the great review, I tried addressing all comments, ready for another round.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
AdamGrzybkowski: Interesting, this indeed fixed it in one scenario when tapping `Show bookings` directly from the specific filter page. But when I first go back to the main filter page, and only after that tap `Show bookings` the property is not correct. I guess it's best to show on the recording: 


https://github.com/user-attachments/assets/942b04ad-0797-4711-a19a-bad611603cf4

hichamboushaba: Interesting, what effect does this could have? the filter details screen is inside a NavHost in the filter list screen, so normally we are calling the same callback 🤔

On my side I can't reproduce it:

https://github.com/user-attachments/assets/befa7d60-5167-4444-bbd7-e3487976db6d

Can you test the following patch and see if it makes any difference:

```patch
IDEA additional info:
<+>UTF-8
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt	(revision 2620a781ecb36010812d39dcf293a7c4677c9012)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt	(date 1772713972607)
@@ -346,6 +346,7 @@
 
     fun trackBookingListView() = launch {
         val state = _state.first()
+        val enabledFiltersCount = bookingFilterRepository.bookingFiltersFlow.first().enabledFiltersCount
 
         analyticsTrackerWrapper.track(
             AnalyticsEvent.BOOKING_LIST_VIEW,
@@ -353,7 +354,7 @@
                 BookingAnalyticsHelper.KEY_SELECTED_TAB to state.tabState.selectedTab.toAnalyticsValue(),
                 BookingAnalyticsHelper.KEY_IS_DEFAULT_TAB to (!didUserSwitchTab).toString(),
                 BookingAnalyticsHelper.KEY_IS_LIST_EMPTY to state.contentState.bookings.isEmpty().toString(),
-                BookingAnalyticsHelper.KEY_IS_FILTERED to (state.controlsState.enabledFiltersCount > 0).toString()
+                BookingAnalyticsHelper.KEY_IS_FILTERED to (enabledFiltersCount > 0).toString()
             )
         )
     }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingAnalyticsHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+    companion object {
+        const val KEY_BOOKING_STATUS = "booking_status"
+        const val KEY_ACTION = "action"
+        const val KEY_SELECTED_TAB = "selected_tab"
+        const val KEY_IS_SEARCH_ACTIVE = "is_search_active"
+        const val KEY_IS_FILTERING_ACTIVE = "is_filtering_active"
+        const val KEY_SORT_OPTION = "sort_option"
+        const val KEY_IS_DEFAULT_TAB = "is_default_tab"
+        const val KEY_IS_LIST_EMPTY = "is_list_empty"
+        const val KEY_IS_FILTERED = "is_filtered"
```
AdamGrzybkowski, line 43: Nitpick, but I noticed we use two different keys for the same thing. This and `KEY_IS_FILTERING_ACTIVE`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
AdamGrzybkowski: Ok so it's not about the flow directly but about the time it takes. If it takes long enough (5 sec) you should be able to reproduce. This is likely related to the default timeout in the `asLiveData()` extension. After that timeout, we will only get the cached/replayed value in onStart. 

The above fix solves the problem, as it's not using cached values.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingAnalyticsHelper.kt
hichamboushaba: What do you mean? the other property is for a different event. Do you mean we should use the same wording for the property? If yes, then we need to discuss with the team, as iOS already implemented tracking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
hichamboushaba: Oh, I see, great find, thanks again.
This means the logic is not working well even for the other properties, the last commit 83deb50 should fix this now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/BookingAnalyticsHelper.kt
AdamGrzybkowski: Yes, the wording was confusing, given it's the same value for a different key. I'm aware it's already used so just mentioning it.

AdamGrzybkowski (approved):
All good now!

PR #15472, opened 2026-03-03
Remove new strings.xml entries from hotfix branch

kidinov:
### Description
Removes two new string resources (`woopos_totals_toolbar_title` and `woopos_card_payment_toolbar_title`) that were added in the hotfix branch. Since **strings.xml cannot be changed in a hotfix**, the code now reuses the existing `R.string.payment` which has the same "Payment" value. `woopos_totals_toolbar_title` was unused entirely.

### Test Steps
1. Open POS and go to a card payment flow
2. Verify the toolbar title still shows "Payment"

### Images/gif
N/A


malinajirka (approved):
TY

PR #15473, opened 2026-03-03
[WOOMOB-2348] Show error toast on attendance status update failure

kidinov:
### Description
Fixes WOOMOB-2348

When the attendance status optimistic update fails, the UI was silently rolling back without telling the user. This adds an error toast on failure so the user knows the update didn't go through.

Also, about cancelling the in-flight request when the user toggles again - this doesn't work well here because `updateAttendanceStatus` in the repository runs the actual network call in `appCoroutineScope`. Cancelling the viewmodel's job only cancels the await(), but the request keeps running and writes the stale value to
  DB. Then `observeBookings` picks it up and overwrites the optimistic UI state. Fixing this properly would need an override mechanism in the observer, which adds a lot of complexity, which may not be worth it as no common use case.
 
@malinajirka if you have any ideas on how to improve that, let me know

### Test Steps
1. Open WooPos bookings screen
2. Select a booking with editable attendance status
3. Toggle the attendance switch while in airplane mode
4. Verify that the toggle rolls back and an error toast is shown

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsViewModel.kt
@@ -468,6 +468,7 @@ class WooPosBookingsViewModel @Inject constructor(
                     selectedDetails = reverted,
                     items = updateItemsWithDetails(rollbackState.items, reverted),
                 )
+                _toastEvent.emit(resourceProvider.getString(R.string.booking_attendance_status_error))
             }
         }
     }


malinajirka (approved):
LGTM, thanks!

> Also, about cancelling the in-flight request when the user toggles again - this doesn't work well here because updateAttendanceStatus in the repository runs the actual network call in appCoroutineScope. Cancelling the viewmodel's job only cancels the await(), but the request keeps running and writes the stale value to
DB. Then observeBookings picks it up and overwrites the optimistic UI state. Fixing this properly would need an override mechanism in the observer, which adds a lot of complexity, which may not be worth it as no common use case.

I see two options:
1. Replace the app scope with local scope. That's what Sam originally suggested and I mentioned/discouraged him with argument that it might have been intentional to use app scope to ensure the request completes even when user leaves the screen. However, considering we don't write the result anywhere - as soon as the app sends the request, the server processes it -> so the fact the job survives on Android doesn't impact the result what so ever.
2. Return Deferred from the updateAttendance and then cancel it directly.

kidinov:
@malinajirka thanks for the proposal!

Option 1 would help cancel in-flight requests, but it doesn't fully solve the problem. If a previous request already completed and wrote to the DB before the user toggles again, observeBookings still picks up that stale value and overwrites the UI. So we'd still need an override mechanism in the observer on top of switching to local scope.

Also, all mutations in BookingsRepository (updateNote, cancelBooking, markAsPaid) use the same appCoroutineScope.async { }.await() pattern. Changing just `updateAttendanceStatus` would be inconsistent, and changing all of them is a bigger scope

Option 2 has the same observer problem, plus it leaks concurrency details (Deferred) into the repository API.

I think for now the toast is good enough since rapid toggling is not a common scenario. If it becomes a real issue, we can revisit it with a proper observer override mechanism

malinajirka:
I see, thanks for elaborating on it. I agree, fixing this properly would require a deeper change.

PR #15477, opened 2026-03-04
[WOOMOB-1820] Fix pull-to-refresh not working on empty booking list

AdamGrzybkowski:
### Description
Fixes WOOMOB-1820

The empty state view in the booking list was not wrapped in a pull-to-refresh container, so users couldn't trigger a refresh when there were no bookings. This wraps the `EmptyView` in `WCPullToRefreshBox` to enable pull-to-refresh on the empty state, consistent with the non-empty list behaviour.

### Test Steps
1. Open the app and navigate to the Bookings list.
2. Ensure there are no bookings (or use a store/account with no bookings).
3. Pull down on the empty state screen.
4. Verify that the pull-to-refresh indicator appears and a refresh is triggered.

### Images/gif
N/A


hichamboushaba (approved):
Looks good, thanks @AdamGrzybkowski

PR #15483, opened 2026-03-04
[WOOMOB-2394] Booking details location

AdamGrzybkowski:
Closes: WOOMOB-2394

### Description

The booking details screen was showing a hardcoded location string. This PR replaces it with the actual location fetched from the Product API’s booking_location field, and caches it locally in Room for subsequent visits.

**What changed:**

1. Replaced the hardcoded location in BookingMapper with a dynamic BookingLocationStatus parameter
2. Added a skeleton loading state for the location field (matching the existing staff member pattern)
3. `BookingsStore.fetchProductBookingLocation()` now accepts an optional bookingId — when provided, it checks the DB cache first and persists the fetched location automatically. The ViewModel gets updates reactively via Room observation
4. When location is unavailable, a dash (-) is displayed

**Location field in BookingEntity**

This PR adds a location column to BookingEntity in Room. The Bookings REST API doesn’t currently include location, so this field is populated from a separate Product API call and cached locally.

If the backend adds location to the booking response in the future, this field is already in place — we’d just need to map it in BookingDtoMapper.toEntity() instead of relying on the Product endpoint.

The updated fetchProductBookingLocation(productId, bookingId) signature is backward-compatible — bookingId defaults to null, so existing callers like `WooPosBookingsViewModel` (which uses its own in-memory cache keyed by productId) are unaffected. POS could potentially adopt the DB caching by passing bookingId in the future if needed.

Open to feedback on this approach — happy to explore alternatives if there’s a better way to handle the caching layer.

### Test Steps
1. Open a booking that has a product with a booking_location set
6. Observe the skeleton loading state briefly, then the location appears
7. Close and reopen the same booking — location should appear immediately (cached)
8. Open a booking whose product has no booking_location — should show -

### Images/gif

https://github.com/user-attachments/assets/0b2b69ab-9d7f-44d1-ac61-174a6e7b7d84

- [x]  I have considered if this change warrants release notes and have added them to RELEASE-NOTES.txt if necessary. Use the “[Internal]” label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
@@ -20,6 +20,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import com.woocommerce.android.R
+import com.woocommerce.android.extensions.orNullIfEmpty
 import com.woocommerce.android.ui.bookings.details.AttendanceUpdateStatus
 import com.woocommerce.android.ui.bookings.details.CancelStatus
 import com.woocommerce.android.ui.compose.animations.SkeletonView
@@ -73,9 +74,29 @@ fun BookingAppointmentDetails(
                 }
             }
             AppointmentDetailsRow(
-                label = R.string.booking_appointment_label_location,
-                value = model.location
-            )
+                label = R.string.booking_appointment_label_location
+            ) {
+                when (model.location) {
+                    is BookingLocationStatus.Loaded, is BookingLocationStatus.Unavailable -> {
+                        Text(
+                            text = (model.location as? BookingLocationStatus.Loaded)?.location?.orNullIfEmpty() ?: "-",
+                            style = MaterialTheme.typography.bodyMedium,
+                            color = MaterialTheme.colorScheme.onSurfaceVariant,
+                            maxLines = 1,
+                            overflow = TextOverflow.Ellipsis
+                        )
+                    }
+
+                    BookingLocationStatus.Loading -> {
+                        SkeletonView(
+                            width = 80.dp,
+                            height = with(LocalDensity.current) {
+                                MaterialTheme.typography.bodyMedium.fontSize.toDp()
+                            },
+                        )
+                    }
+                }
+            }
             AppointmentDetailsRow(
                 label = R.string.booking_appointment_label_duration,
                 value = model.duration,
@@ -166,7 +187,7 @@ data class BookingAppointmentDetailsModel(
     val date: String,
     val time: String,
     val staff: BookingStaffMemberStatus?,
-    val location: String,
+    val location: BookingLocationStatus,
     val duration: String,
     val cancelButtonVisible: Boolean,
     val cancelStatus: CancelStatus,
@@ -179,6 +200,12 @@ data class BookingAppointmentDetailsModel(
     val attendanceButtonVisible: Boolean = isAttendanceStatusEditable
     val attendanceButtonEnabled: Boolean = attendanceUpdateStatus != AttendanceUpdateStatus.InProgress
     val attendanceInProgressShown: Boolean = attendanceUpdateStatus == AttendanceUpdateStatus.InProgress
+}
+
+sealed interface BookingLocationStatus {
+    data object Loading : BookingLocationStatus
+    data class Loaded(val location: String) : BookingLocationStatus
+    data object Unavailable : BookingLocationStatus
 }
 
 sealed interface BookingStaffMemberStatus {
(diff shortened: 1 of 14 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingDtoMapper.kt
```diff
@@ -28,7 +28,8 @@ internal class BookingDtoMapper @Inject constructor(
 
     suspend fun BookingDto.toEntity(
         localSiteId: LocalId,
-        orderEntity: OrderEntity?
+        orderEntity: OrderEntity?,
+        existingLocation: String? = null,
```
AdamGrzybkowski, line 32: This is used to keep the previously stored location while refreshing the booking.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/283">24.3</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -194,18 +196,33 @@ class BookingsStore @Inject internal constructor(
...
             val response = bookingsRestClient.fetchProductBookingLocation(site, productId)
             when {
                 response.isError -> WooResult(response.error)
-                response.result != null -> WooResult(response.result.bookingLocation)
+                response.result != null -> {
+                    val location = response.result.bookingLocation
+                    if (bookingId != null) {
+                        bookingsDao.updateLocation(site.localId(), bookingId, location)
+                    }
+                    WooResult(location)
```
AdamGrzybkowski: Updating the location here means we get it for free in the VM, as we are already colleting all the changes from the DB. VM only needs to trigger the fetch, and that's all.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
```diff
@@ -194,13 +196,26 @@ class BookingsStore @Inject internal constructor(
...
+            }
             val response = bookingsRestClient.fetchProductBookingLocation(site, productId)
             when {
                 response.isError -> WooResult(response.error)
-                response.result != null -> WooResult(response.result.bookingLocation)
+                response.result != null -> {
+                    val location = response.result.bookingLocation?.takeIf { it.isNotBlank() }
+                    if (bookingId != null) {
+                        bookingsDao.updateLocation(site.localId(), bookingId, location)
+                    }
```
AdamGrzybkowski, line 221: A blank location will be treated as `null` to trigger the fetch next time. WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAppointmentDetails.kt
```diff
@@ -73,9 +73,29 @@ fun BookingAppointmentDetails(
...
+                            maxLines = 1,
+                            overflow = TextOverflow.Ellipsis
+                        )
+                    }
+
+                    BookingLocationStatus.Loading -> {
+                        SkeletonView(
+                            width = 80.dp,
+                            height = with(LocalDensity.current) {
+                                MaterialTheme.typography.bodyMedium.fontSize.toDp()
```
irfano, line 93: TIL! That's a smart way to size skeleton views.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
irfano, line 56: `BookingDto.toEntity()` is called from `fetchBookings()` without `existingLocation`, while other booking write paths do preserve it. So list sync can overwrite a previously cached `location` with `null`, which matches the attached video where Booking Details location falls back to `"-"` after refresh/reopen. Can we pass through the existing DB `location` in this call path as well?

https://github.com/user-attachments/assets/8df20092-921b-4e0f-93cd-ec4411e343ed

irfano (approved):
Thanks for the clear PR description. I like the overall caching approach and the backward-compatible API update.

I’m requesting changes for one issue already noted in the inline comment. Other than that, everything looks good from my side.

irfano (changes requested):
I accidentally approved this in my previous comment. I'm requesting changes because the issue in the video is a blocker.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
AdamGrzybkowski: Good catch! Should be fixed now

irfano (approved):

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsStore.kt
irfano: Nice! It's fixed now.

PR #15485, opened 2026-03-05
Omit template_id for WC >= 10.7.0 in POS receipt emails

staskus:
## Description

WC 10.7.0+ supports auto-selecting the best email template when `template_id` is omitted from `POST /orders/{id}/actions/send_email` ([woocommerce/woocommerce#63556](https://github.com/woocommerce/woocommerce/pull/63556)). This fixes receipt emails for bookings and refunded orders where the `customer_pos_completed_order` template is not available.

Mirrors the iOS implementation: [woocommerce-ios#16781](https://github.com/woocommerce/woocommerce-ios/pull/16781).

The change threads an optional `templateId` parameter through `OrderRestClient`, `WCOrderStore`, and `WooPosEmailReceiptRepository`. The repository now checks the WC plugin version and decides:

- **WC < 10.0**: legacy path - update order email, then call `send_order_details`
- **WC 10.0 - 10.6.x**: call `send_email` with `template_id: "customer_pos_completed_order"`
- **WC >= 10.7.0**: call `send_email` without `template_id` (server auto-selects)

## Test Steps

1. Connect to a store running WC >= 10.7.0
2. Complete a POS order and send a receipt email
3. Verify the email is received successfully
4. For bookings: complete a booking order via POS and send receipt - should now work

Automated tests cover all version ranges.

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -951,7 +951,7 @@ class OrderRestClient @Inject constructor(
         orderId: Long,
         email: String,
         forceEmailUpdate: Boolean,
-        templateId: String? = null
+        templateId: String?
     ): WooPayload<Unit> {
         val response = wooNetwork.executePostGsonRequest(
             site = site,
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -950,17 +950,20 @@ class OrderRestClient @Inject constructor(
         site: SiteModel,
         orderId: Long,
         email: String,
-        forceEmailUpdate: Boolean
+        forceEmailUpdate: Boolean,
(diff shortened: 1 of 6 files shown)


staskus:
I appreciate if someone could review this. It requires the newest WooCommerce on a trunk to test.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -950,17 +950,20 @@ class OrderRestClient @Inject constructor(
         site: SiteModel,
         orderId: Long,
         email: String,
-        forceEmailUpdate: Boolean
+        forceEmailUpdate: Boolean,
+        templateId: String? = null
```
kidinov: Right now, the default silently means "let the server auto-select", which is a meaningful behavioural choice that should be (maybe) explicit at the call site. If someone adds a new caller in the future and forgets to pass template Id, they'd get auto-selection without realizing it, which may or may not be what they want. 

So I'd consider removing default values in both rest and store levels. Wdyt?

kidinov (approved):
Looks and tests good! Thank you! I left one minor comment, please take a look

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
staskus: That's a fair point. I'll make passing templateId explicit.

PR #15489, opened 2026-03-05
[WOOMOB-2397][POS Refunds M2] Part 2: Refunded items UI 

samiuelson:
## Description

Adds the UI layer for displaying refunded line items in POS order details. This is the second of two PRs for WOOMOB-2397, building on #15488.

**Changes:**
- Add `ProductsSection` composable that handles `LineItemsState` (loading shimmer / loaded items)
- Add `RefundedProductsSection` composable for displaying refunded items
- Add `ProductsShimmer` composable for loading state placeholder
- Extract `title` parameter in `OrdersProducts` for reuse between products and refunded sections
- Update previews with refunded items sample data
- Add "Refunded Products" string resource

## Testing instructions

1. Open POS and complete an order
2. Issue a partial refund for one item
3. Open order details
4. Verify "Products" section shows non-refunded items
5. Verify "Refunded Products" section appears with refunded items
6. Verify shimmer loading state appears briefly while refund data loads

## Images/Gif
image

https://github.com/user-attachments/assets/eafc4355-508c-43aa-858d-ae7aaea06c41

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
@@ -91,12 +91,9 @@ fun WooPosOrderDetails(
 
         Spacer(Modifier.height(WooPosSpacing.Large.value))
 
-        val lineItemsState = details.lineItems
-        if (lineItemsState is WooPosOrdersState.OrderDetailsViewState.Computed.Details.LineItemsState.Loaded &&
-            lineItemsState.items.isNotEmpty()
-        ) {
-            OrdersProducts(lineItems = lineItemsState.items)
-        }
+        ProductsSection(lineItems = details.lineItems)
+
+        RefundedProductsSection(refundedLineItems = details.refundedLineItems)
 
         Spacer(Modifier.height(WooPosSpacing.Medium.value))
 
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
```diff
@@ -166,11 +163,56 @@ private fun OrdersHeader(details: WooPosOrdersState.OrderDetailsViewState.Comput
...
+}
+
+@Composable
+private fun RefundedProductsSection(
+    refundedLineItems: WooPosOrdersState.OrderDetailsViewState.Computed.Details.LineItemsState
+) {
+    when (refundedLineItems) {
+        is WooPosOrdersState.OrderDetailsViewState.Computed.Details.LineItemsState.Loading -> {
+            Spacer(Modifier.height(WooPosSpacing.Medium.value))
+            ProductsShimmer(
```
malinajirka, line 191: 💡 Do we know at this point that the order has refunded items? If not, I'd consider hiding the shimmer effect for refunded section, since most orders won't (hopefully) have any refunded items. Those with refunded items will likely have all items refunded (one shown section only). This should reduce jumping UI, wdyt?

malinajirka (approved):
Thanks @samiuelson !

The changes look good and work as expected. The only thing which feels a bit off to me as a user - not technically related to this PR - is that the product row is without taxes:

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
samiuelson: Good idea. I made order refund state check more granular to skip shimmer for Products section in case the order has been fully refunded: [87c10d5](https://github.com/woocommerce/woocommerce-android/pull/15489/commits/87c10d5cefbce86ae892aca9bed2b770fc344eaf). This reduces UI "jumping".

samiuelson:
Thanks for review, @malinajirka!

> The changes look good and work as expected. The only thing which feels a bit off to me as a user - not technically related to this PR - is that the product row is without taxes

I'm not sure if it's a good idea, because this would make the display logic inconsistent across Products and Refunded Products sections. In Products totals are displayed without taxes as well.

In the next milestone M3, we are going to show more details of refund:
image
I think this will solve the issue. Let me know what you think!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
malinajirka: I haven't tested this after the latest commit, but when the order doesn't have any refunded items, the loading screen shows both Products and Refunded Products sections with shimmer effect, doesn't it? I'm just assuming this since the `Loaded` state skips the section completely, but the `Loading` state doesn't seem to.  Wdyt?

malinajirka:
>> The changes look good and work as expected. The only thing which feels a bit off to me as a user - not technically related to this PR - is that the product row is without taxes

> I'm not sure if it's a good idea, because this would make the display logic inconsistent across Products and Refunded Products sections. In Products totals are displayed without taxes as well.

I wasn't suggesting we should show tax for refunded items but not show it for product items. I just pointed out that this part feels weird to me as a user. 

When I have an order with 5 products and no refunds, it's relatively natural, that product rows are without taxes and all taxes are sumed up below. However, when I have an order with 1 product and 2 refunded products, it feels a bit weird. I'm not sure what the solution is though - especially, since the backend likely doesn't tell us per-item-tax.

Screenshot_1773072434

P.S. I noticed that the Apple ring on this order has negative sign -> we might want to consider using `abs` here as well.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
samiuelson: > I haven't tested this after the latest commit, but when the order doesn't have any refunded items, the loading screen shows both Products and Refunded Products sections with shimmer effect, doesn't it?

This is not the case. In case the order doesn't have any refunded items the shimmer effect is not displayed at all, and LineItemsState.Loaded state is emitted immediately. The logic behind it is in `mapOrderDetailsWithoutActions()`.

malinajirka: Ahh cool, thanks!

samiuelson:
> When I have an order with 5 products and no refunds, it's relatively natural, that product rows are without taxes and all taxes are sumed up below. However, when I have an order with 1 product and 2 refunded products, it feels a bit weird. I'm not sure what the solution is though - especially, since the backend likely doesn't tell us per-item-tax.

Got it. To me this feels good enough, assuming we'll have the total taxes shown in the list/details of historical refunds in the Totals section. Though, I'll ask for a designer's feedback during call for testing, likely tomorrow.

> P.S. I noticed that the Apple ring on this order has negative sign -> we might want to consider using `abs` here as well.

This has been fixed within 5f8c4f9f3357c18d6f9c80f9020343f8ba4b7ccb.

PR #15492, opened 2026-03-05
[WOOMOB-2323] Match CIAB booking badge styles

irfano:
### Description
Fixes WOOMOB-2323
Updates CIAB bookings badges to match the current design styling in the bookings UI. This doesn't affect POS badges, I'll update those once we approve the styling in this PR.

### Changes

558974309-04823cbe-ab02-4e64-86e5-624a1ea586e8

### Test Steps
1. Open a CIAB site with bookings
2. Navigate to the bookings list.

### Images/gif
|Before|After|Web|
|-|-|-|
|before|after|image|

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusTag.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusTag.kt
@@ -92,7 +92,11 @@ fun BookingAttendanceStatus.backgroundColor(): Color = if (isOutlined()) {
 }
 
 @Composable
-fun BookingAttendanceStatus.textColor(): Color = colorResource(R.color.tagView_text)
+fun BookingAttendanceStatus.textColor(): Color = if (isOutlined()) {
+    colorResource(R.color.color_on_surface_high)
+} else {
+    colorResource(R.color.tagView_text)
+}
 
 @Composable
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusTag.kt
```diff
@@ -83,10 +85,27 @@ fun BookingAttendanceStatus?.text(): String {
...
+}
 
 @Composable
-fun BookingAttendanceStatus.textColor(): Color = colorResource(R.color.tagView_text)
+fun BookingAttendanceStatus.border(): BorderStroke? = if (isOutlined()) {
+    BorderStroke(1.dp, colorResource(R.color.tag_border_booking_outlined))
+} else {
+    null
+}
+
```
AdamGrzybkowski: Those could be private. Are we planning to reuse them?

AdamGrzybkowski (approved):
LGTM 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/compose/BookingAttendanceStatusTag.kt
irfano: Good point. No reuse planned for now, so I made those helpers private in e35851f230f.

PR #15493, opened 2026-03-06
Fix analytics inconsistencies between Android app and wp-admin dashboard

JorgeMucientes:
Fixes: WOOMOB-2441

## Description

Fixes analytics data inconsistencies between the Android app's **My Store** dashboard and the **wp-admin Analytics** dashboard. Two root causes were identified and fixed:

### 1. Hardcoded `date_type=date_created` override

The Android app was sending `date_type=date_created` in every call to `/wc-analytics/reports/revenue/stats`, overriding the store's configured `woocommerce_date_type` setting. wp-admin does **not** send this parameter, letting the server use the store's setting (default: `date_paid`).

**Fix:** Remove the `dateType` parameter and `date_type` from the API request entirely, matching wp-admin behavior.

### 2. Wrong revenue metric displayed

The My Store dashboard was showing `total_sales` as the "Revenue" headline, while wp-admin shows `net_revenue` ("Net sales"). These are different metrics:
- `total_sales` = gross_sales − refunds − coupons **+ taxes + shipping** ($122.56)
- `net_revenue` = gross_sales − refunds − coupons **− taxes − shipping** ($125.00)

**Fix:** Display `netRevenue` instead of `totalSales` in `DashboardStatsViewModel` and `DashboardStatsView`. Also updated the label displayed in the dashboard

### Files changed

| File | Change |
|------|--------|
| `OrderStatsRestClient.kt` | Removed `dateType` param and `date_type` from API request |
| `WCStatsStoreTest.kt` | Updated mock matchers for removed parameter |
| `DashboardStatsViewModel.kt` | Map `netRevenue` instead of `totalSales` |
| `DashboardStatsView.kt` | Render `netRevenue` instead of `totalSales` |

### Related

- Original commit introducing `date_type=date_created`: [`c2c97f4`](https://github.com/woocommerce/woocommerce-android/commit/c2c97f4712167cffc053a0c0835f25720f1819b0)
- Server-side `date_type` handling: [Orders/Stats/DataStore.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/API/Reports/Orders/Stats/DataStore.php)

## Testing instructions

1. Open the app and navigate to **My Store** dashboard
2. Select a custom date range with known order data
3. Compare the **Revenue** and **Orders** numbers with the same date range in **wp-admin → Analytics → Revenue**
4. Verify that the app now shows `net_revenue` (matching wp-admin's "Net sales") instead of `total_sales`
5. Verify that no `date_type` parameter is sent in the API request (check network logs)
6. Verify top performers card matches wp-admin -> Analytics -> Top performers 

https://github.com/user-attachments/assets/b5ec5e19-0158-4913-a713-f26ae55731e0

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsView.kt
```diff
@@ -545,9 +545,9 @@ class DashboardStatsView @JvmOverloads constructor(
     private fun updateChartView() {
         val wasEmpty = binding.chart.lineData?.let { it.dataSetCount == 0 } ?: true
 
-        val grossRevenue = revenueStatsModel?.totalSales ?: 0.0
+        val netRevenue = revenueStatsModel?.netRevenue ?: 0.0
         val revenue = currencyFormatter.getFormattedAmountZeroRounded(
-            grossRevenue,
+            netRevenue,
             revenueStatsModel?.currencyCode.orEmpty()
```
AdamGrzybkowski, line 551: ⚠️ Should we also apply this change to the Wear App? I see that [WearableConnectionRepository](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/wear/WearableConnectionRepository.kt) is using `totalSales` as well.

AdamGrzybkowski (commented):
Looks good, but left one comment to make sure all supported devices are aligned.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsView.kt
JorgeMucientes: Good catch @AdamGrzybkowski I've addressed the missing changes in the last 3 commits. Although the wear app is as deprecated barely used product i think is good we keep consistency.

AdamGrzybkowski (approved):
Looks good! One test in the wear module needs some updates to pass

PR #15494, opened 2026-03-06
[WOOMOB-2442] Fix incremental sync missing product updates for sites with negative UTC offsets

malinajirka:
### Description
Fixes WOOMOB-2442

During incremental sync, the POS sends a `modified_after` parameter to filter products changed since the last sync. The stored timestamp comes from the server's response header and is always in **UTC**. However, the WooCommerce REST API compares `modified_after` against `date_modified`, which is stored in the **site's local timezone**. For sites with negative UTC offsets, the UTC value is always ahead of the local `date_modified`, causing recently modified products to be excluded from the sync response.

This fix converts the UTC timestamp to site-local time (using `SiteModel.timezone`) at the REST client level before sending the `modified_after` parameter, so all callers (products, variations, catalog size check) automatically get the correction.

### Test Steps
1. Configure a WooCommerce site with a negative UTC offset (e.g., UTC-6, America/Chicago)
2. Open the POS and wait for the initial full sync to complete
3. On the site's admin, modify a product (e.g., change the price)
4. Return to the POS and trigger an incremental sync (pull to refresh or wait for periodic sync)
5. Verify the modified product is updated in the POS catalog
6. Repeat with a site in a positive UTC offset (e.g., UTC+5:30) and a UTC+0 site to confirm no regressions

### Images/gif
N/A

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
@@ -10,6 +10,10 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.CoreProductStatus
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.ProductApiResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
+import java.time.LocalDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+import java.util.Locale
 import javax.inject.Inject
 
 class WooPosProductRestClient @Inject constructor(
@@ -24,6 +28,11 @@ class WooPosProductRestClient @Inject constructor(
         private const val VARIATIONS_FIELDS = "id,parent_id,description,sku,global_unique_id,status,price," +
             "regular_price,sale_price,date_modified,stock_quantity,stock_status,manage_stock," +
             "backordered,attributes,image,downloadable,name,type"
+
+        private const val SECONDS_PER_HOUR = 3600
+
+        private val API_DATE_FORMATTER = DateTimeFormatter
+            .ofPattern("yyyy-MM-dd'T'HH:mm:ss", Locale.US)
     }
 
     suspend fun fetchProducts(
@@ -38,7 +47,7 @@ class WooPosProductRestClient @Inject constructor(
         val params = buildBaseParams(
             pageSize = pageSize,
             page = page,
-            modifiedAfter = modifiedAfter,
+            modifiedAfter = modifiedAfter?.let { adjustUtcToSiteLocalTime(it, site.timezone) },
             fields = PRODUCT_FIELDS,
             includeStatus = includeStatus,
             posProductsOnly = posProductsOnly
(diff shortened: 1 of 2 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
```diff
@@ -151,4 +160,22 @@ class WooPosProductRestClient @Inject constructor(
...
+     * headers), we must convert to site-local time before sending. Without this, sites with
+     * negative UTC offsets will miss recently modified products because the UTC value is always
+     * ahead of their local `date_modified`.
+     */
+    internal fun adjustUtcToSiteLocalTime(utcDateString: String, siteGmtOffset: String?): String {
+        val offsetHours = siteGmtOffset?.toDoubleOrNull() ?: return utcDateString
+        if (offsetHours == 0.0) return utcDateString
+
+        val offsetSeconds = (offsetHours * SECONDS_PER_HOUR).toInt()
+        val zoneOffset = ZoneOffset.ofTotalSeconds(offsetSeconds)
```
malinajirka: 📓 We need to convert to seconds and offset seconds, since `ofHours` doesn't support fractional offsets like 5.5 (UTC+5:30, India).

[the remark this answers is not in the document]
malinajirka: Technically correct for arbitrary doubles, but in practice it doesn't matter here. WordPress timezone offsets are always multiples of 0.25 hours (-12, -5.5, 5.75, etc.). Values like 0.25, 0.5, 0.75 are exactly representable in IEEE 754 double, so 0.25 * 3600 = 900.0, 0.5 * 3600 = 1800.0, etc. — no precision loss, truncation and rounding produce  identical results.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
```diff
@@ -151,4 +160,22 @@ class WooPosProductRestClient @Inject constructor(
...
+        val utcInstant = LocalDateTime.parse(utcDateString, API_DATE_FORMATTER)
+            .toInstant(ZoneOffset.UTC)
+        return API_DATE_FORMATTER.withZone(zoneOffset).format(utcInstant)
```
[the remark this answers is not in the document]
malinajirka: Fixed

samiuelson (approved):
LGTM

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
```diff
@@ -24,6 +30,11 @@ class WooPosProductRestClient @Inject constructor(
         private const val VARIATIONS_FIELDS = "id,parent_id,description,sku,global_unique_id,status,price," +
             "regular_price,sale_price,date_modified,stock_quantity,stock_status,manage_stock," +
             "backordered,attributes,image,downloadable,name,type"
+
+        private const val SECONDS_PER_HOUR = 3600
```
samiuelson: We could be using `1.hours.inWholeSeconds` from std lib instead. It's a matter of taste though.

samiuelson (approved):

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/pos/WooPosProductRestClient.kt
malinajirka: Updated in c3c316659cafe82202a340eebbce1905b57ff550

PR #15498, opened 2026-03-06
[WOOMOB-2413] Hide push notifications benefit behind flag

hichamboushaba:
Fixes WOOMOB-2413

### Description
Hides the Push Notifications benefit from the Jetpack installation dialog when the WOO_PUSH_NOTIFICATIONS_SYSTEM_M2 feature flag is enabled. This ensures users do not see the benefit if the new push notification system is active.

> [!NOTE]
> I didn't add a unit test for the new behavior because it'll require creating a new use case wrapping the WOO_PUSH_NOTIFICATIONS_SYSTEM_M2 feature flag logic, and given there is already some pending work to make feature flags testable without the use cases (#15299), I decided it's not worth to cause more divergence.

### Test Steps
1. Sign in using a Jetpack CP site or using Site credentials.
2. Open the app and navigate to Settings > Developer Options.
3. Enable the WOO_PUSH_NOTIFICATIONS_SYSTEM_M2 flag.
4. Go to Jetpack installation dialog (from Settings > Install Jetpack).
5. Verify that the Push Notifications benefit is hidden.
6. Disable the flag and verify the benefit reappears.

### Images/gif
| Before | After |
| --- | --- |
| jetpack-benefits-after-settings | jetpack-benefits-flag-enabled |

- [x] I have considered if this change warrants release notes and have added them to RELEASE-NOTES.txt if necessary. Use the [Internal] label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jetpack/benefits/JetpackBenefitsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jetpack/benefits/JetpackBenefitsScreen.kt
@@ -83,15 +83,17 @@ fun JetpackBenefitsScreen(
             )
             Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.major_200)))
 
-            // Push notifications
-            BenefitEntry(
-                icon = R.drawable.ic_alarm_bell_ring,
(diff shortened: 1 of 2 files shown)


AdamGrzybkowski (approved):
Works as described :shipit:

PR #15499, opened 2026-03-09
[WOOMOB-2421] Fix product detail webview URL path

hichamboushaba:
Fixes WOOMOB-2421

### Description
Fix the product detail webview URL by adding the missing `admin.php` segment and using the correct `/woocommerce/services/edit/` path instead of `/woocommerce/products/edit/`.

### Test Steps
1. Use a CIAB site.
2. Open a bookable product in the app
3. Verify the webview loads correctly with the proper admin URL
4. Confirm the product edit page renders as expected

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/webview/ProductDetailWebViewViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/webview/ProductDetailWebViewViewModel.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.ui.products.details.webview
 
+import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.liveData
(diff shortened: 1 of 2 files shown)


irfano:
@hichamboushaba , I've already opened an issue for this (#15491: I previously added `services` and now I've also added `admin.php` to the link).  I also renamed the `ProductDetailWebView` → `BookableServiceDetailWebView`. If that renaming makes sense to you, let's proceed with my PR.

hichamboushaba:
@irfano thanks for the ping, sorry I didn't see your PR.

Regarding the rename, I don't have a strong opinion, but just explaining my rationale for using the generic names in the first place, I wanted to have this as reusable and easy to extend, so if we want to use the same pattern for another product type, we can do it quickly by just updating the logic in two areas: in `openInWebViewIfNeeded` and in the `ProductDetailWebViewModel` to construct the correct URL.

As I said, it's not a strong opinion, so feel free to close this PR if my point doesn't make sense to you.

AdamGrzybkowski:
FYI. I believe that the WOOMOB-2421 is about supporting the `bookable` type together with the `bookable-service` [here](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailViewModel.kt#L435). From the walkthrough, you could see that the `bookable` type would be opened in the app's native view, but it doesn't support it fully. 

I'm not fully sure how to create such product though 🤔

irfano:
@hichamboushaba , I closed my PR after your comment. Let's proceed with this one, and don't forget to target the release branch.

AdamGrzybkowski (approved):
:shipit:

PR #15504, opened 2026-03-10
[WOOMOB-2449] Fix attendance status filter to be single-select

AdamGrzybkowski:
### Description
Fixes WOOMOB-2449

The attendance status filter in the bookings list allowed multi-select, which caused a 400 API error when multiple statuses were sent (`attendance_status is not of type string`). This PR makes the filter single-select so only one attendance status (or "Any") can be active at a time.

Changes:
- Changed `BookingAttendanceStatusFilterUiState` to hold a single `AttendanceStatus?` instead of a `Set`
- Simplified the ViewModel selection logic to replace (not toggle) the selected status
- Updated `BookingFilterRepository` to read only the first stored value from DataStore (backwards compatible with previously stored multi-value data)
- Simplified API serialization in `BookingsRestClient` to send a plain string

Note: I've decided to keep `AttendanceStatuses` since it's heavily used and in case we need to add multiselect again, it should be significanlty easier. 

### Test Steps
1. Open the Bookings list
2. Tap the filter icon and navigate to the Attendance Status filter
3. Select "Attended" — verify the checkmark appears next to "Attended" only
4. Select "Unattended" — verify "Attended" is deselected and "Unattended" is now selected
5. Tap "Unattended" again — verify it stays selected (no toggle-off)
6. Select "Any" — verify the filter is cleared
7. Tap "Show Bookings" — verify no API error occurs

### Images/gif

https://github.com/user-attachments/assets/500c5144-4c8e-4cee-bf89-3f53463c4b25

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -149,7 +149,7 @@ class BookingsRestClient @Inject constructor(
             set("product", serviceEvents.values.joinToString(",") { it.productId.toString() })
         }
         if (attendanceStatuses != BookingsFilterOption.AttendanceStatuses.DEFAULT) {
-            attendanceStatuses.values.firstOrNull()?.let {
+            attendanceStatuses.values.singleOrNull()?.let {
                 set("attendance_status", it.key)
             }
         }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
@@ -149,9 +149,9 @@ class BookingsRestClient @Inject constructor(
             set("product", serviceEvents.values.joinToString(",") { it.productId.toString() })
         }
         if (attendanceStatuses != BookingsFilterOption.AttendanceStatuses.DEFAULT) {
-            set(
-                "attendance_status",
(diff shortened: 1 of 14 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
```diff
@@ -149,9 +149,9 @@ class BookingsRestClient @Inject constructor(
             set("product", serviceEvents.values.joinToString(",") { it.productId.toString() })
         }
         if (attendanceStatuses != BookingsFilterOption.AttendanceStatuses.DEFAULT) {
-            set(
-                "attendance_status",
-                attendanceStatuses.values.joinToString(FILTER_QUERY_PARAMETER_SEPERATOR) { it.key })
+            attendanceStatuses.values.singleOrNull()?.let {
```
hichamboushaba: > Note: I've decided to keep AttendanceStatuses since it's heavily used and in case we need to add multiselect again, it should be significanlty easier.

I believe it's unlikely we will have more attendance statuses than the two we currently have, as they are logically the only valid options. From a YAGNI perspective, I would also suggest changing `attendanceStatuses` to a single value in the networking layer to make the code easier to understand.

If, in the unlikely event, we do add more attendance statuses and the API supports multi-select options, we can reintroduce the logic. I think that would be a simple change.

hichamboushaba (approved):
Nice work @AdamGrzybkowski, I'm approving, I left a single comment, but it's not a blocker, you can ignore it if you don't agree.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/bookings/BookingsRestClient.kt
AdamGrzybkowski: Yeah, I was 50/50 on that. Done in https://github.com/woocommerce/woocommerce-android/pull/15504/changes/e7b0d80d42eae5a338c65584b2b07e5eb43983c3

hichamboushaba (approved):
Great work.

PR #15479, opened 2026-03-04
[WOOMOB-2395] Enable POS coupons for CIAB sites

malinajirka:
### Description
Fixes WOOMOB-2395

Enables coupons on POS for CIAB sites regardless of site settings. CIAB sites short-circuit to enabled without making the remote settings request. Non-CIAB sites continue to respect the site settings as before.

### Test Steps
1. Open POS on a CIAB site
2. Verify coupons tab is available and functional

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15513, opened 2026-03-11
[WOOMOB-2456] Revert force-enabled coupons on CIAB sites

malinajirka:
### Description
Fixes WOOMOB-2456

Two changes to `CachedCouponEnabledChecker`:

1. **Remove CIAB coupon override**: The API for checking coupon availability has been fixed for CIAB sites. Previously (#15479), we short-circuited CIAB sites to always return `true` for coupons, bypassing the remote settings check. Now that the API works correctly, all sites (including CIAB) rely on the API result again.

2. **Fix cache invalidation on site switch**: The cached coupon-enabled value was not invalidated when the selected site changed, meaning a user switching sites could potentially see stale coupon availability from the previous site (given the same instance of the cache was used). The cache now tracks the site ID and refetches when the site changes.

### Test Steps
1. Open POS on a CIAB site with coupons enabled → verify coupons are available
2. Open POS on a site with coupons disabled → verify coupons are not available
3. Switch between sites → verify coupon availability reflects the current site's settings

### Images/gif
N/A


kidinov (approved):
LGTM!

PR #15514, opened 2026-03-11
WOOMOB-2467: Fix payment flow not navigating back from Bookings

AdamGrzybkowski:
### Description

Fixes a bug where completing the Collect Payment flow (via Cash) when navigating from Bookings → View Order → Order Details would leave the user stuck on the payment method selection screen instead of navigating back.

**Root cause:** The `NavigateBackToOrderList` handler used a nav action that pops to the `orders` (Order List) destination. When the user entered Order Details from Bookings, the Order List is not in the navigation backstack, so the pop silently failed.

**Fix:** Use `popBackStack` directly instead of the nav action, with a fallback to `orderDetailFragment` when the `orders` destination is not in the backstack.

### Test Steps

1. Open WooPOS Bookings or navigate to a Booking with an unpaid order
2. Tap "View Order" to open Order Details
3. Tap "Collect Payment"
4. Complete the flow via Cash option
5. **Expected:** Payment flow closes and Order Details is shown
6. **Verify the normal flow still works:** From the Orders list, open an unpaid order → Collect Payment → Cash → should return to the Order List

### Images/gif
N/A

- [ ] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
@@ -230,8 +230,11 @@ class SelectPaymentMethodFragment : BaseFragment(R.layout.fragment_select_paymen
                         // We should pop the back stack to show the [OrderDetailsFragment].
                         findNavController().popBackStack()
                     } else {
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15515, opened 2026-03-11
[WOOMOB-2467] Fix payment flow not navigating back from Bookings

AdamGrzybkowski:
### Description
Fixes WOOMOB-2467

When completing the Collect Payment flow (via Cash) after navigating from Bookings → View Order → Order Details, the user gets stuck on the payment method selection screen instead of navigating back.

**Root cause:** The `NavigateBackToOrderList` handler used a nav action with `popUpTo="@+id/orders"`. When the user entered Order Details from Bookings, the Order List destination is not in the navigation backstack, so the pop silently failed.

**Fix:** Replace the nav action with `popBackStack(R.id.orders)`, falling back to `popBackStack(R.id.orderDetailFragment)` when the orders destination is not found. Also removed the now-unused nav action from `nav_graph_payment_flow.xml`.

### Test Steps

**From Booking Details (phone):**
1. On a phone, open WooPOS Bookings and select an unpaid booking
2. Tap "View Order" to open Order Details
3. Tap "Collect Payment" → complete via Cash
4. Verify: payment flow closes and Order Details is shown

**From Booking Details (tablet):**
5. On a tablet, repeat steps 1-3
6. Verify: payment flow closes and Order Details is shown

**From Order Details (phone):**
7. On a phone, open the Orders list and select an unpaid order
8. Tap "Collect Payment" → complete via Cash
9. Verify: payment flow closes and the Order List is shown

**From Order Details (tablet):**
10. On a tablet, repeat steps 7-8
11. Verify: payment flow closes and Order Details is shown in the detail pane

### Images/gif

https://github.com/user-attachments/assets/8e5c1331-3c58-4a28-b663-6347f0f23a8b

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
@@ -230,8 +230,11 @@ class SelectPaymentMethodFragment : BaseFragment(R.layout.fragment_select_paymen
                         // We should pop the back stack to show the [OrderDetailsFragment].
                         findNavController().popBackStack()
                     } else {
-                        SelectPaymentMethodFragmentDirections.actionSelectPaymentMethodFragmentToOrderList().run {
-                            findNavController().navigateSafely(this)
+                        val popped = findNavController().popBackStack(R.id.orders, false)
+                        if (!popped) {
+                            // orders destination is not in the backstack (e.g. when navigating
+                            // from Bookings), fall back to popping to OrderDetailFragment
+                            findNavController().popBackStack(R.id.orderDetailFragment, false)
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
```diff
@@ -230,8 +230,11 @@ class SelectPaymentMethodFragment : BaseFragment(R.layout.fragment_select_paymen
...
                         findNavController().popBackStack()
                     } else {
-                        SelectPaymentMethodFragmentDirections.actionSelectPaymentMethodFragmentToOrderList().run {
-                            findNavController().navigateSafely(this)
+                        val popped = findNavController().popBackStack(R.id.orders, false)
+                        if (!popped) {
+                            // orders destination is not in the backstack (e.g. when navigating
+                            // from Bookings), fall back to popping to OrderDetailFragment
+                            findNavController().popBackStack(R.id.orderDetailFragment, false)
                         }
```
hichamboushaba: Question, if we want to keep the same behavior as before, wouldn't something like the following suggestion work? To make sure we always pop to the parent of order details fragment?


```suggestion
                        findNavController().popBackStack(R.id.orderDetailFragment, true)
}
```

hichamboushaba (approved):
Nice catch @AdamGrzybkowski, I left a question, but it's not a blocker.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodFragment.kt
AdamGrzybkowski: I like it! Applied in https://github.com/woocommerce/woocommerce-android/pull/15515/commits/7196a8b165d8b2b5cb655050b67ab3829c1d472b

hichamboushaba (approved):

PR #15516, opened 2026-03-11
[WOOMOB-2466] Fix issues found during FTS call for testing

kidinov:
### Description
Fixes WOOMOB-2466

Fixes 4 issues found during the pdfdoF-8Nh-p2 for POS FTS search on Android:

1. **Crash: duplicate LazyColumn keys** — search results could contain items with the same numeric ID across different types (e.g. variable product and variation). Fixed by using composite keys with type prefix (`simple_123`, `variation_search_456`) @staskus could you please validate this one? I could not reproduce this as I probably don't have overlaps in the ids in my stores

2. **Search results not filtered by product_type and status** — FTS index included grouped and draft products. Fixed by filtering at index build time, matching iOS approach. Uses `WooPosProductsTypesFilterConfig` as single source of truth.

3. **Recent searches not saved for variable products** — when tapping a variable product in search, the toolbar state switched to VariationList before the search could be saved. Fixed by saving the search at tap time.

4. **Search UX improvements** — skip full-screen loading state for local FTS search (results are fast enough). Skip debounce for FTS. Enable item animations in search results list so items animate in/out as results change. @malinajirka wdyt about that?

### Test Steps
1. Search for a term that returns many results and scroll down — no crash
2. Search for a grouped or draft product — it should not appear in results
3. Search for a product, tap a variable product, go back — the search term should appear in recent searches
4. Type in search — results should appear immediately without loading flicker (local catalog mode)
5. Verify remote search still shows loading state when local cache is empty

### Images/gif

https://github.com/user-attachments/assets/00f8b53d-191c-4aba-9772-91f5aea664d8

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt
@@ -4,14 +4,6 @@ sealed class WooPosItemSelectionViewState(
     open val id: Long,
     open val name: String
 ) {
-    val uniqueKey: String
-        get() = when (this) {
-            is Product.Simple -> "simple_$id"
-            is Product.Variable -> "variable_$id"
-            is Product.Variation -> "variation_$id"
-            is Product.VariationSearchResult -> "variation_search_$id"
-            is Coupon -> "coupon_$id"
-        }
     sealed class Product(
         override val id: Long,
         override val name: String,
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt
@@ -4,6 +4,14 @@ sealed class WooPosItemSelectionViewState(
     open val id: Long,
     open val name: String
 ) {
+    val uniqueKey: String
+        get() = when (this) {
+            is Product.Simple -> "simple_$id"
+            is Product.Variable -> "variable_$id"
+            is Product.Variation -> "variation_$id"
+            is Product.VariationSearchResult -> "variation_search_$id"
+            is Coupon -> "coupon_$id"
+        }
     sealed class Product(
         override val id: Long,
         override val name: String,
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
@@ -5,7 +5,6 @@ import com.google.gson.JsonSyntaxException
 import com.google.gson.reflect.TypeToken
 import com.woocommerce.android.ui.woopos.common.data.WooPosProductsTypesFilterConfig
 import com.woocommerce.android.ui.woopos.common.util.WooPosLogWrapper
-import org.wordpress.android.fluxc.store.WCProductStore.ProductFilterOption
 import com.woocommerce.android.util.WooLog
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.SiteModel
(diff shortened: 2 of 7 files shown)


staskus:
> @staskus could you please validate this one? I could not reproduce this as I probably don't have overlaps in the ids in my stores

@kidinov still crashes.

Claude code fixes it this way and then it doesn't:

image

kidinov:
@staskus Thank you!


@malinajirka there is a theory regarding how in general we can get duplicates:
```
  The flow works like this:                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                           
  1. PaginatedSyncHelper.fetchAllPages() accumulates items from all API pages into a flat list. The WooCommerce REST API can return the same product/variation on consecutive page boundaries (this is a known WordPress REST API pagination behavior when items share the same sort key).
  2. When these go to the products/variations Room tables, upsertProducts / upsertVariations handles duplicates via ON CONFLICT REPLACE (primary key).
  3. But the same list goes to buildFtsEntities / buildFtsEntitiesForIncremental, which just does:
  val productFtsEntities = eligibleProducts.map { it.toFtsEntity(siteIdString) }
  3. No dedup. And ftsDao.insertAll() is a plain @Insert — FTS4 virtual tables in SQLite don't support UNIQUE constraints or ON CONFLICT strategies. Every INSERT creates a new row.
  4. When searchProductsFts runs, the FTS query returns both duplicate rows. Both map to the same entity from the products/variations map, producing two identical items in the result list with the same id and same type.
  5. In the UI, both items generate the same uniqueKey (e.g., variation_search_5503) → crash.
 ```

Does it make any sense? Also, as soon as we are going to use the file, this should be gone? Additionally, I am even wondering if all this is caused by moving to another way of syncing 🤔 

In any case, it feels like it's the right call to do deduplication before inserting into the FTS table - d8d7294e6d64f739e723457ab41648f1c5faf9d0

@staskus I am curious if you won't take the latest changes and will go to the settings and force the full sync. Will it still be crashing? 

After that, could you please verify that it's fixed in the latest commit after the full sync as well?

**Comment by staskus** [comment]
@kidinov, you're right, even without the last commit after full resync, there's no more crash.

File: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt [hunk 1]
```diff
@@ -4,6 +4,14 @@ sealed class WooPosItemSelectionViewState(
     open val id: Long,
     open val name: String
 ) {
+    val uniqueKey: String
```
malinajirka: 📓 I believe this shouldn't be necessary - all use post_id under the hood, so they are unique across the site. Having said that, I guess there is no harm.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
```diff
@@ -228,6 +245,17 @@ class WooPosLocalCatalogSyncWithFts @Inject constructor(
         }
     }
 
+    private val allowedStatus = filterConfig.filters[ProductFilterOption.STATUS]
```
malinajirka, line 248: 🔍 Would be nice if we could extract the logic and re-use it here as well as WooPosSearchByIdentifier. Having said that, I guess we are reusing the filterConfig, so that's hopefully enough.

malinajirka (approved):
Thanks @kidinov, the changes look good to me and the app seems to behave as expected.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemSelectionViewState.kt
kidinov: You're right, IDs are post_ids so they're unique site-wide. Removed it in d4dafe45f13.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncWithFts.kt
kidinov: Yeah, the models are different (Room entities vs domain models) and the checks have subtle differences, so extracting it would add more abstraction than it saves. Sharing the filterConfig should be enough for now.

PR #15519, opened 2026-03-11
[WOOMOB-2472] Apply booking filters on Today and Upcoming tabs

AdamGrzybkowski:
### Description
Fixes WOOMOB-2472

User-selected filters (team members, customer, attendance status, service/events, etc.) were only applied on the **All** tab. The **Today** and **Upcoming** tabs discarded all user-selected filters because `prepareFilters()` was constructing a new `BookingFilters` object instead of copying the existing user filters.

- Changed `BookingFilters(` to `filters.copy(` for Today/Upcoming tabs so user-selected filters are preserved
- Tab-specific overrides (`dateRange` and `excludedBookingStatuses`) are still applied as before

### Test Steps
1. Open the Bookings screen on the **Today** tab
2. Tap the filter button and select a filter (e.g., a specific team member or customer)
3. Verify the booking list is filtered correctly
4. Switch to the **Upcoming** tab
5. Verify the filters are still applied (same filtered results, not all bookings)
6. Switch to the **All** tab and verify filters still work as before
7. Clear filters and verify all tabs show unfiltered results

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/bookings/list/BookingListViewModel.kt
@@ -338,7 +338,7 @@ class BookingListViewModel @Inject constructor(
 
     private fun FetchParams.prepareFilters(): BookingFilters = when (selectedTab) {
         BookingListTab.Today,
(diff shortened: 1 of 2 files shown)


hichamboushaba (approved):
Great catch, it seems this was a regression when we updated logic to exclude canceled bookings from the list. Good work adding unit tests.

PR #15522, opened 2026-03-12
[WOOMOB-2480] Fix symbol-only FTS search showing error instead of empty results

kidinov:
### Description
Fixes WOOMOB-2480

When searching for symbol-only queries (like `-`, `!`, `@`, `#`, `$`) in POS, the UNICODE61 tokenizer strips all non-letter/non-digit characters, producing a blank formatted query. Previously this returned `Result.failure`, which propagated as "Unable to load products" error. Now it returns an empty success so the UI correctly shows "No search results".

### Test Steps
1. Open POS and go to the product search
2. Search for `-` or `!` or `@#$`
3. Verify it shows "No search results" instead of "Unable to load products"
4. Search for `t-shirt` — verify it still finds products as before

### Images/gif

https://github.com/user-attachments/assets/efc2eb4c-29be-4332-b0d2-0de3e1e014b2


malinajirka (approved):
LGTM, thanks!

PR #15525, opened 2026-03-12
[Woo POS][Refunds] Fix refund details dialog not opening when switching between orders

samiuelson:
## Description
Fixes refund details dialog not opening when switching between orders in POS.

When switching orders (A → B → A), `startSideLoadActionsIfNeeded` returned early for Order A on re-selection because its actions were already `Loaded`. This skipped `sideLoadActions` which is responsible for populating `cachedRefundData`. The cache was left pointing to Order B, so the refund dialog's guard (`cached.orderId != selectedDetails.id`) silently blocked it from opening.

Added `ensureRefundDataCached()` that runs when the early return is taken. It checks if the cache already matches the order — if so, it's a no-op. If not, it fetches the order and its refunds from local cache (`forceRefresh = false`) and rebuilds `cachedRefundData`.

### Code cleanup
Renamed functions in `WooPosOrdersViewModel` to better reflect their actual behavior:
- `sideLoadActions` → `sideLoadActionsAndRefundData` (also fetches refunds and caches refund data)
- `sideLoadActionsJob` → `sideLoadJob` (shorter, matches broader scope)
- `startSideLoadActionsIfNeeded` → `sideLoadOrderExtras` (always loads bookings + conditionally loads actions)
- `getOrderForSideLoading` → `resolveOrder` (generic order resolver, not specific to side-loading)
- `isAlreadySelectedWithActions` → `isOrderFullyLoaded` (describes intent, not mechanism)
- `updateStoredRefundDataForOrder` → `cacheRefundDataFromLazyState` (explicit about data source and action)

Extracted `cacheRefundData` helper to centralize the 5 direct `CachedRefundData` writes.

## Test Steps
1. Go to POS > Orders
2. Select an order with refunds, tap a refund row → dialog opens ✓
3. Select a different order with refunds, tap a refund row → dialog opens ✓
4. Switch back to the first order, tap a refund row → dialog opens ✓ (was broken before)


kidinov (commented):
@samiuelson there is something wrong with the order selection. Notice that it selects the first order after the refund flow is canceled

https://github.com/user-attachments/assets/681e9adc-4e2e-440c-9031-668914fb08ed

samiuelson:
> @samiuelson there is something wrong with the order selection. Notice that it selects the first order after the refund flow is canceled
> 

I wasn't able to effectively reproduce this, but looking into the code, I can see that order refresh logic was lacking cancelation when selecting another order. Fixed in [a45ae4e](https://github.com/woocommerce/woocommerce-android/pull/15525/commits/a45ae4e8002a631b327481bdee36461e790fc0e8)

kidinov (approved):
LGTM!

PR #15526, opened 2026-03-12
[Woo POS][Refunds] Fix refund time showing 12:00 AM instead of actual time

samiuelson:
WOOMOB-2488

### Description

In POS > Orders > Order details, refund rows in the totals section always showed "12:00 AM" instead of the actual refund time.

**Root cause:** `RefundMapper.fromFormattedDate()` used a date-only format (`yyyy-MM-dd`) to parse the API's ISO 8601 datetime string (`2024-01-15T14:30:00`), discarding the time portion and defaulting to midnight.

**Fix:** Try parsing with `yyyy-MM-dd'T'HH:mm:ss` first, falling back to the date-only format for backward compatibility.

### Test Steps

1. Open POS and create an order
2. Complete the order, then issue a refund
3. Navigate to Orders > tap the order > scroll to the totals section
4. Verify the refund row shows the correct time (not "12:00 AM")

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
@@ -13,6 +13,7 @@ import java.util.Locale
 import javax.inject.Inject
 
 const val DATE_FORMAT_DAY = "yyyy-MM-dd"
+private const val DATE_FORMAT_DATE_TIME = "yyyy-MM-dd'T'HH:mm:ss"
 
 class RefundMapper @Inject constructor(private val gson: Gson) {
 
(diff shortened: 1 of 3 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
```diff
@@ -68,7 +69,12 @@ class RefundMapper @Inject constructor(private val gson: Gson) {
         if (date.isEmpty()) {
             return null
         }
-        val dateFormat = SimpleDateFormat(DATE_FORMAT_DAY, Locale.ROOT)
-        return dateFormat.parse(date)
+        val dateTimeFormat = SimpleDateFormat(DATE_FORMAT_DATE_TIME, Locale.ROOT)
```
malinajirka: I think we might want to consider using DateTimeFormatter - it's what we use in other parts of the POS codebase + it gives us some benefits. On this note, I'd consider using `WooPosDateFormatter` here, so we have all the date parsing in one place.

P.S. If I recall correctly, it's a best practice to cache (and synchronize) SimpleDateFormatter, since it's creation isn't cheap. DateTimeFormatter doesn't need to be manually synchronized, but I think it might still be a good practice to cache it. But this part feels out of scope for this PR.

malinajirka (approved):
Thanks, I've left a non-blocking comment.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
samiuelson: Replaced SimpleDateFormat with DateTimeFormatter. I added shared formatter instances to DateUtils in fluxc-plugin so they can be reused across the module. Couldn't use WooPosDateFormatter since it lives in the app module and RefundMapper is in fluxc-plugin. [ce3ed18](https://github.com/woocommerce/woocommerce-android/pull/15526/commits/ce3ed1838dabe45e7b26146d9571f4c03f22cb6e)

PR #15528, opened 2026-03-12
Fix $0.00 refunds not tracked in POS UI

samiuelson:
Closes WOOMOB-2490

### Description

After refunding a $0.00 item through the POS refund flow, the refund was created successfully in wp-admin but not reflected in the POS UI. Previously refunded $0.00 items still appeared in the refundable items list.

**Root cause:** `WooPosRetrieveOrderRefunds` short-circuited with an empty list when `order.refundTotal == 0`. Since $0.00 refunds don't change the order's `refundTotal`, this optimization incorrectly skipped fetching refunds for orders that only have $0.00 refunds.

**Fix:** Remove the `refundTotal == 0` early return so refunds are always fetched. When `forceRefresh = false`, the local cache is checked first (fast), and an API call only happens if the cache is empty — so the performance impact is minimal.

### Test Steps

1. In POS, create an order containing a $0.00 item
2. Complete the order, then refund the $0.00 item
3. Verify the refund appears in the POS UI and the item no longer appears in the refundable items list

### Images/gif
image

image


kidinov (approved):
LGTM!

PR #15530, opened 2026-03-12
Use visible Woo sites for notification filtering

irfano:
### Description
Fixes WOOMOB-2113
This updates `NotificationMessageHandler` to use the Woo site visibility source when filtering self-driven notifications for hidden stores.
It avoids relying on `SiteStore.visibleSites`, which is based on deprecated WP.com visibility data, and adds coverage for visible and hidden site flows.

### Test Steps
Testing isn’t currently possible because the Woo push server isn’t active yet.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
@@ -93,13 +93,12 @@ class NotificationMessageHandler @Inject constructor(
         val pushUserId = messageData[PUSH_ARG_USER]
 
         // We need to filter out duplicate notifications from the WPCOM system
-        if (isRegisteredForWooPush) {
+        if (!isWooSiteVisible(notification.remoteSiteId)) {
+            wooLog.w(NOTIFICATIONS, "Skipping notification, site ${notification.remoteSiteId} is not visible")
+            return
+        } else if (isRegisteredForWooPush) {
             if (notification.remoteNoteId > 0L) {
                 wooLog.d(NOTIFICATIONS, "Skipping WPCOM notification, already registered with Woo Core")
-                return
-            }
-            if (!isWooSiteVisible(notification.remoteSiteId)) {
-                wooLog.w(NOTIFICATIONS, "Skipping notification, site ${notification.remoteSiteId} is not visible")
                 return
             }
         } else if (!accountStore.hasAccessToken()) {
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -98,7 +98,7 @@ class NotificationMessageHandler @Inject constructor(
                 wooLog.d(NOTIFICATIONS, "Skipping WPCOM notification, already registered with Woo Core")
                 return
             }
-            if (siteStore.visibleSites.none { it.siteId == notification.remoteSiteId }) {
+            if (!isWooSiteVisible(notification.remoteSiteId)) {
```
hichamboushaba: @irfano should we move this check out of the `if (isRegisteredForWooPush)` check? I mean if a user hides a site on the site picker, they most likely are not interested on its notifications too, WDYT?

hichamboushaba: OK, so answering myself here, we do disable WPCom notifications for sites when the user hides them, so technically moving this out of the `if` check is not needed.
However, I still believe that extracting it from the `if` check would be a beneficial change, as it would make the code easier to follow, in my opinion.

hichamboushaba (approved):
Thanks @irfano, I left a comment above, but it's not a blocker.

hichamboushaba:
@itsmeichigo I think iOS will need a similar logic, I saw you [added](https://github.com/woocommerce/woocommerce-ios/pull/16606) a service extension to block WPCom notifications if the site is registered with Woo PN, so maybe you can reuse it for this too to block PNs for hidden sites.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
irfano: Yes, this is not strictly necessary because WPCom notifications are already disabled. Still, there’s no harm in applying this change to all stores. There may be a delay when disabling WPCom notifications takes effect, and this check could help during that window.
I made the change in 713048699bb179f3e8426ea8162c9c7cae11dd1c.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -93,15 +93,14 @@ class NotificationMessageHandler @Inject constructor(
         val pushUserId = messageData[PUSH_ARG_USER]
 
         // We need to filter out duplicate notifications from the WPCOM system
-        if (isRegisteredForWooPush) {
+        if (!isWooSiteVisible(notification.remoteSiteId)) {
+            wooLog.w(NOTIFICATIONS, "Skipping notification, site ${notification.remoteSiteId} is not visible")
+            return
+        } else if (isRegisteredForWooPush) {
```
hichamboushaba: np, given we early return, I think the `else` here (and also below) is not needed, it just adds indentation, WDYT?

irfano: Agreed! Done: 26947b597b0d6e2319f13c3875cc9fbdc570ed66

PR #15536, opened 2026-03-13
[WellSQL Migration] Migrate `AccountModel` to Room

ParaskP7:
Closes: [AINFRA-538](https://linear.app/a8c/issue/AINFRA-538)

<sup>Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com></sup>

---

### Description

This PR migrates the `AccountModel` table from [WellSQL](https://github.com/wordpress-mobile/wellsql) to [Room](https://developer.android.com/training/data-storage/room) database as part of the ongoing database migration effort.

**Test Phase:**
- Replace manual Room database builders with test rules (a01fd15c5c4)
- Move account tests and site test utils to fluxc module (39f5eb7e424)
- Use AssertJ instead of deprecated JUnit assert in account test (4380435c02f)

**Cleanup Phase:**
- Remove unused `getAllAccounts()` from `AccountSqlUtils` (620ed92bfa1)

**Refactor Phase:**
- Extract account persistence into `AccountStorePersistence.kt` (20c71dcacd4)
    - Wraps all `AccountSqlUtils` calls behind a Kotlin class
    - `AccountStore.java` delegates persistence to the new class
    - Uses `runBlocking` bridge for suspend DAO calls from Java EventBus handlers (same pattern as `ThemeStore.kt`)

**Migration Phase:**
- Migrate `AccountModel` table from WellSQL to Room (30117aa7375)
    - Introduce `AccountEntity` for Room
    - Introduce `AccountDao` with comprehensive tests
    - Introduce `AccountMapper` for domain ↔ entity conversion with tests
    - Register entity and DAO in `WPAndroidDatabase` (v36 → v37)
    - Update `AccountStorePersistence` to use DAO directly
    - Delete `AccountSqlUtils`
    - Delete `WellSqlTestUtils.java` (broken after annotation removal, all callers redundant)
- Drop `AccountModel` table from `wp-fluxc` WellSQL database (cc4cd6e4b9a)

**Bug Fix:**
- Inject `AccountStorePersistence` into `SiteSqlUtils` for account setup in tests (5d42165bd77)
    - Root cause: `SiteSqlUtils` and `SiteStoreUnitTest` were using `AccountSqlUtils` directly, which was deleted in the migration. Updated to use `AccountStorePersistence` + Room DAO instead.

**Additional Cleanup:**
- Remove leftover WellSQL `@Column` annotations from `WCProductShippingClassModel` (28a3493768d)
    - Already a Room `@Entity` since `fbaadf36dcb`, annotations were stale leftovers
- Remove leftover WellSQL setup from leaderboards tests (5b4e47a2f6d)

**Self-Healing Account Data (No Row Migration Needed):**

All WellSQL tables migrated so far have been "stateless" in the sense that their rows are fetched from the server on demand and can be safely discarded — the app simply re-fetches them. `AccountModel` is different: it holds the signed-in user's identity (display name, avatar, email, privacy settings) and is loaded once at login time, making it similar to the upcoming `SiteModel` migration. This raised the question of whether existing rows need to be migrated between databases to avoid data loss on upgrade.

After investigation and testing, the answer is no — the account data is self-healing:
- The OAuth access token is stored in `SharedPreferences` (independent from the database), so removing the WellSQL table does not cause logout
- On every app start, `AccountStore.onAccountChanged()` triggers a `/me` API call that repopulates the Room table with fresh account data from the server
- For offline upgrades, the app degrades gracefully: the avatar may be empty and some account-derived fields blank, but the selected site (stored in `SiteModel`, separate table) still displays correctly. Everything self-heals on the next network connection.

---

#### Database Migration Verification
1. Build and install this update
2. Open Android Studio → App Inspection → Database Inspector
3. Verify `AccountModel` table NO LONGER exists in `wp-fluxc` database
4. Verify `AccountEntity` table EXISTS in `wp-android-database`
5. Compare all 21 column values match the expected account data

#### Profile & Identity Display
1. Navigate to **More Menu** (hamburger) → Verify avatar loads and display name is correct
2. Navigate to **Settings** → Verify display name and username shown correctly
3. Open **Site Picker** → Verify user info card shows name and avatar

#### Account Settings & Privacy
1. Navigate to **Settings → Privacy** → Toggle analytics on/off → Restart app → Verify toggle state persists

#### Login & Logout
1. Sign out → Verify returns to login screen and account data is cleared
2. Sign back in → Verify fresh account data loads correctly (avatar, name, email, settings)

#### Offline Upgrade Path
1. Enable airplane mode
2. Force-stop and reopen the app
3. Verify the app loads without crash — site name still shows, avatar may be empty
4. Disable airplane mode → Verify account data self-heals (avatar loads, name appears)

---

### Images/gif

`BEFORE`

AccountModel-BEFORE

`AFTER`

AccountModel-AFTER

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/dao/AccountDao.kt
```diff
@@ -0,0 +1,50 @@
...
+    suspend fun getDefaultAccount(): AccountEntity? =
+        getAccountById(DEFAULT_ACCOUNT_LOCAL_ID)
+
+    @Query(
+        """
+        SELECT * FROM AccountEntity
+        WHERE id = :id
+        """
+    )
+    protected abstract suspend fun getAccountById(id: Int): AccountEntity?
```
hichamboushaba: Minor point here, is there any added value for the protected function `getAccountById`? If not, then we can use something like the following to use a single function, same for the other functions, WDYT?

```suggestion
    @Query(
        """
        SELECT * FROM AccountEntity
        WHERE id = $DEFAULT_ACCOUNT_LOCAL_ID
        """
    )
    abstract suspend fun getDefaultAccount(): AccountEntity?
```

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/AccountStorePersistence.kt
```diff
@@ -0,0 +1,30 @@
+package org.wordpress.android.fluxc.persistence
+
+import kotlinx.coroutines.runBlocking
+import org.wordpress.android.fluxc.model.AccountModel
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class AccountStorePersistence @Inject constructor(
+    private val database: WPAndroidDatabase,
```
hichamboushaba, line 10: np, just one point here, if we want to stay consistent with how we access other Daos, maybe we can add the `accountDao` to the `DatabaseModule` and inject it directly? I'm suggesting this just for consistency.

hichamboushaba (approved):
Nice work @ParaskP7, the changes work well.
I left two minor comments, but they are optional, feel free to ignore them.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/AccountStorePersistence.kt
ParaskP7: This is a great pointer @hichamboushaba and thanks for bringing this forward! 💯

The TL;DR is that this was a calculated decision — `AccountDao` is declared as `internal` (same as `ThemeDao`, `ListDao`, `NotificationDao`, `WhatsNewDao`), and the `accountDao()` accessor on `WPAndroidDatabase` is also `internal`. Injecting `WPAndroidDatabase` instead of `AccountDao` directly allows `AccountStorePersistence` to remain a `public` class that can be constructed from tests in other modules (`fluxc-tests`, `fluxc-plugin`) without exposing the DAO outside the `fluxc` module boundary.

---

Now, If we injected `AccountDao` directly via `DatabaseModule`, the production DI would work (Hilt generates code within the same module), but test classes in `fluxc-tests` and `fluxc-plugin` manually construct `AccountStorePersistence` and wouldn't be able to pass an `AccountDao` — Kotlin blocks `internal` access across modules, and from Java the only way is through the mangled name (`accountDao$fluxc_debug()`).

FYI: `ThemeStore` uses the same pattern for the same reason — it injects `WPAndroidDatabase` and accesses `database.themeDao()` internally.

Wdyt? 💭

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/dao/AccountDao.kt
ParaskP7: Oh, this is a great suggestion @hichamboushaba , I like and will do, thanks so much! 🥇

ParaskP7:
Thanks for reviewing and testing this @hichamboushaba , much appreciated! 🙇 ❤️ 🚀

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/dao/AccountDao.kt
ParaskP7: FYI: This is now done: f514305cafaf8bf5392965320e48e79bc60e32df Cc @hichamboushaba

ParaskP7:
FYI: I just changed the milestone from `24.4` (this Friday, `20/03`)  to `24.5` (next sprint, `03/04`) and thus, my plan is to merge this on Monday (`23/03`), just to avoid merging this so close to this eminent code freeze/release, and give us enough time to dev-battle-test it before the next code freeze/release. 😊

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/AccountStorePersistence.kt
hichamboushaba: I see, sounds good @ParaskP7, thanks for the explanation.

ParaskP7: Awesome, thank YOU again @hichamboushaba ! 🙇 ❤️ 🚀

PR #15537, opened 2026-03-13
Make trunk prototype builds on-demand and add nightly build pipeline

AliSoftware:
Closes [AINFRA-2170](https://linear.app/a8c/issue/AINFRA-2170/set-up-nightly-trunk-prototype-builds-for-woo-mobile)
iOS counterpart: https://github.com/woocommerce/woocommerce-ios/pull/16818

## Description

 - Mobile Prototype Builds are now on-demand on `trunk` (gated behind an input step) but still built automatically on PRs.
 - But we added a scheduled pipeline for nightly prototype builds, triggered on `trunk` by a Buildkite schedule to run every day at midnight (see https://github.com/Automattic/buildkite-ci/pull/827)
 - Wear OS Prototype Builds remain on-demand on both trunk and PRs. Nightly builds don't build Wear OS at this point since they are tested more rarely, and if one needs one they can still generate it on demand from `trunk`

This was discussed in p1773397624341559/1773394983.779039-slack-CC7L49W13 to align with WCiOS and provide a balance between having builds from `trunk` all while not overwhelming CI usage too much for building unnecessary.

### Before

| Prototype Builds | Mobile app | Wear app |
|---|---|---|
| On PRs | ✅ Always | 🙋 On Demand |
| On `trunk` commits | ✅ Always | 🙋 On Demand |
| Nightly Builds | ❌ N/A | ❌ N/A |

### After

| Prototype Builds | Mobile app | Wear app |
|---|---|---|
| On PRs | ✅ Always | 🙋 On Demand |
| On `trunk` commits | 🙋 On Demand | 🙋 On Demand |
| Nightly Builds | 🕐 [Every day at midnight](https://github.com/Automattic/buildkite-ci/pull/827/changes#diff-3639440759f4caf7238c60efa31dd20ce64b6c08a2c2f25ff01d943e84f5d3cc) | ❌ N/A |

## Testing

 - [x] Verify a Mobile Prototype Build is created on this PR and you get the PR comment for it
 - [x] Verify [the CI build for this PR shows](https://buildkite.com/automattic/woocommerce-android/builds/latest?branch=ainfra-2170%2Fnightly-prototype-builds) a "Create Wear Prototype Build?" input step
    image
 - [x] Click on the "Create Wear Prototype Build?" step, then on the "Continue" button to confirm
    image
 - [ ] Validate the Wear Prototype build is generated for Wear and you get the PR comment for it too.

> [!NOTE]
> Unfortunately, it's not going to be possible to test the cases of `trunk` builds nor scheduled nightly builds until after this PR and https://github.com/Automattic/buildkite-ci/pull/827 land.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -83,7 +83,7 @@ steps:
           - "WooCommerce/build/outputs/apk/**/*"
 
       - input: Create a Wear Prototype Build?
-        prompt: Share a Wear Prototype Build via Firebase App Distribution?
+        prompt: Share a Wear Prototype Build? (it will be uploaded to AWS S3)
         key: wear_prototype_build_triggered
       - label: ":amazon-s3: Wear Prototype Build"
         depends_on: wear_prototype_build_triggered
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -66,20 +66,27 @@ steps:
 
   ########################################
   - group: "🚀 Prototype Build"
+    # Mobile Prototype Builds are on-demand on `trunk`, but always built on PRs
+    # WearOS Prototype Builds are on-demand both on `trunk` and PRs
     steps:
+      - input: Create a Mobile Prototype Build?
+        prompt: Share a Mobile Prototype Build via Firebase App Distribution?
+        key: mobile_prototype_build_triggered
+        if: build.branch == "trunk"
       - label: ":firebase: Mobile Prototype Build"
+        # Buildkite considers this implicitly satisfied if `depends_on` step is skipped due to `if:` condition, so this will run unconditionally on PRs
+        # https://buildkite.com/docs/pipelines/configure/dependencies#how-skipped-steps-affect-dependencies
+        depends_on: mobile_prototype_build_triggered
         command: .buildkite/commands/prototype-build.sh WooCommerce
         plugins: [$CI_TOOLKIT]
         artifact_paths:
           - "WooCommerce/build/outputs/apk/**/*"
 
-      - block: Build Watch Prototype?
-        prompt: Trigger an on-demand Wear OS prototype build
-        key: watch_prototype_build_triggered
-        depends_on: ~
-
-      - label: ":amazon-s3: Watch Prototype Build"
-        depends_on: watch_prototype_build_triggered
+      - input: Create a Wear Prototype Build?
+        prompt: Share a Wear Prototype Build via Firebase App Distribution?
+        key: wear_prototype_build_triggered
+      - label: ":amazon-s3: Wear Prototype Build"
+        depends_on: wear_prototype_build_triggered
         command: .buildkite/commands/prototype-build.sh WooCommerce-Wear
         plugins: [$CI_TOOLKIT]
         artifact_paths:
--- a/.buildkite/schedules/nightly-prototype-build.yml
+++ b/.buildkite/schedules/nightly-prototype-build.yml
@@ -10,3 +10,9 @@ steps:
     plugins: [$CI_TOOLKIT]
     artifact_paths:
       - "WooCommerce/build/outputs/apk/**/*"
+
+notify:
+  - slack:
+      channels:
+        - "#android-core-notifs"
+      if: "build.state == 'failed'"
--- a/.buildkite/schedules/nightly-prototype-build.yml
+++ b/.buildkite/schedules/nightly-prototype-build.yml
@@ -1,0 +1,12 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+agents:
+  queue: "android"
+
+steps:
+  - label: ":firebase: Nightly Mobile Prototype Build"
+    command: .buildkite/commands/prototype-build.sh WooCommerce
+    plugins: [$CI_TOOLKIT]
+    artifact_paths:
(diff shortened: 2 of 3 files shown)


.buildkite/schedules/nightly-prototype-build.yml
```diff
@@ -0,0 +1,12 @@
...
+
+agents:
+  queue: "android"
+
+steps:
+  - label: ":firebase: Nightly Mobile Prototype Build"
+    command: .buildkite/commands/prototype-build.sh WooCommerce
+    plugins: [$CI_TOOLKIT]
+    artifact_paths:
+      - "WooCommerce/build/outputs/apk/**/*"
```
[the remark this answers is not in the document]
AliSoftware: This seems like a good suggestion, @malinajirka wdyt, would you want notifs in that Slack channel for nightly builds? Only if those nightly builds fail, or both if they fail or succeed?

.buildkite/pipeline.yml
```diff
@@ -66,20 +66,27 @@ steps:
...
-        prompt: Trigger an on-demand Wear OS prototype build
-        key: watch_prototype_build_triggered
-        depends_on: ~
-
-      - label: ":amazon-s3: Watch Prototype Build"
-        depends_on: watch_prototype_build_triggered
+      - input: Create a Wear Prototype Build?
+        prompt: Share a Wear Prototype Build via Firebase App Distribution?
+        key: wear_prototype_build_triggered
+      - label: ":amazon-s3: Wear Prototype Build"
```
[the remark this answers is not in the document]
AliSoftware: Nice catch on the discrepancy. But actually the emoji is right — Wear Prototype Builds are uploaded to S3 not to Firebase, because Firebase doesn't support Wear apps — so the `prompt` of the `input` step is the one being misleading, not the emoji.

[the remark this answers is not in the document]
AliSoftware: Fixed in https://github.com/woocommerce/woocommerce-android/pull/15537/commits/286d3f4be5fc6f61d7d82cc5443902d6cf10fcb0

.buildkite/pipeline.yml
```diff
@@ -66,20 +66,27 @@ steps:
...
           - "WooCommerce/build/outputs/apk/**/*"
 
-      - block: Build Watch Prototype?
-        prompt: Trigger an on-demand Wear OS prototype build
-        key: watch_prototype_build_triggered
-        depends_on: ~
-
-      - label: ":amazon-s3: Watch Prototype Build"
-        depends_on: watch_prototype_build_triggered
+      - input: Create a Wear Prototype Build?
```
iangmaia, line 85: Maybe I'm mixing things up, but do we want this job to depend  on the app build? Otherwise, perhaps we should add back the `depends_on: ~`? 🤔

iangmaia: Ah nvm, I think it's not really necessary.

AliSoftware: The previous `depends_on: ~` meant "depend on nothing" (`~` is [the marker for `null` in YAML](https://yaml.org/type/null.html)). It was necessary before because the previous YAML used `block:` step instead of `input`, and those 2 step types are very similar with the main difference being that `block` steps add implicit dependencies to the step before and after it, while `input` doesn't. So having a `block` (with implicit dependencies) then use `depends_on: ~` (to remove those implicit dependencies) was a bit of a convoluted way to do what `input` steps already do.

.buildkite/schedules/nightly-prototype-build.yml
```diff
@@ -0,0 +1,18 @@
...
+
+notify:
+  - slack:
+      channels:
+        - "#android-core-notifs"
+      if: "build.state == 'failed'"
```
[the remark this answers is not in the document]
AliSoftware: Personally I think this would be overkill, but up to the Woo team to decide and tell us what they prefer 🙂

iangmaia (approved):
Confirmed I was able to run the Wear app on demand in Buildkite and we got a [PR comment](https://github.com/woocommerce/woocommerce-android/pull/15537#issuecomment-4057169629) for the prototype. I haven't tested it myself but it looks correct.

👍

.buildkite/schedules/nightly-prototype-build.yml
```diff
@@ -0,0 +1,18 @@
...
+  - label: ":firebase: Nightly Mobile Prototype Build"
+    command: .buildkite/commands/prototype-build.sh WooCommerce
+    plugins: [$CI_TOOLKIT]
+    artifact_paths:
+      - "WooCommerce/build/outputs/apk/**/*"
+
+notify:
+  - slack:
+      channels:
+        - "#android-core-notifs"
```
ParaskP7, line 17: Btw @AliSoftware and just a thought on this, just because I am not sure if this `#android-core-notifs` is the right channel. Why? My thinking below: 🤔

1. `#android-core-notifs` is used for Android Core related topics in general and this feels more like a Woo Android topic. My worry below: 💭
    - It might clatter the channel.
    - We (Android Core) shouldn't be responsible to fix such build errors, if and when they occur.
    - Product engineers might fail to see such build failures in time, us needing to ping them, just to bring them to this channel and raise awareness of it.
2. `woo-android-notifs` is used for Woo Android topics and might be a better fit for a quick resolve.

Cc @JorgeMucientes here as I would like to pick your brain on it too. 🙏

ParaskP7: FYI: #16144 _(long overdue)_

PR #15545, opened 2026-03-16
[WOOMOB-2506] Disable order status editing for CIAB sites

hichamboushaba:
### Description
Fixes WOOMOB-2506

Prevents CIAB site users from changing order status in both the Order Detail Viewer and Order Detail Editor. A new `ReadOnly` mode is added to `OrderDetailOrderStatusView` which hides the edit pencil icon and disables click listeners on the status container. The mode is selected by checking `CIABSiteGateKeeper.isFeatureSupported(OrderStatusEditing)` — CIAB sites get `ReadOnly`, non-CIAB sites retain the existing `OrderEdit` behavior.

Also removes the unused `OrderCreation` mode from `OrderDetailOrderStatusView` — it was never passed to `initView()` since order creation hides the entire status view.

### Test Steps
1. Switch to a **CIAB site**
2. Open an existing order → verify the status tag is visible but there is **no edit pencil icon** and tapping the status area does nothing
3. Tap **Edit** on the order → verify the status section in the editor is also read-only (no pencil icon)
4. Switch to a **non-CIAB site**
5. Open an existing order → verify the **edit pencil icon is visible** and tapping it opens the "Change order status" dialog
6. Tap Edit on the order → verify status editing works as before

### Images/gif
| CIAB | Non-CIAB |
| --- | --- |
| Screenshot_20260316_171933 | Screenshot_20260316_165537 |
| ciab_order_detail_padding_fix | non_ciab_order_detail_edit_visible |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailOrderStatusView.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailOrderStatusView.kt
@@ -25,8 +25,6 @@ class OrderDetailOrderStatusView @JvmOverloads constructor(
 ) : MaterialCardView(ctx, attrs, defStyleAttr) {
     private val binding = OrderDetailOrderStatusBinding.inflate(LayoutInflater.from(ctx), this)
 
-    private var mode: Mode = Mode.OrderEdit
-
     fun updateStatus(orderStatus: OrderStatus) {
         binding.orderStatusOrderTags.contentDescription =
             context.getString(R.string.orderstatus_contentDesc_withStatus, orderStatus.label)
@@ -35,16 +33,8 @@ class OrderDetailOrderStatusView @JvmOverloads constructor(
 
     fun updateOrder(order: Order) {
         binding.orderStatusSubtitle.text = getFormattedDate(order.dateCreated)
-
-        when (mode) {
-            Mode.OrderEdit, Mode.ReadOnly -> {
-                binding.orderStatusHeader.text =
-                    order.getBillingName(context.getString(R.string.orderdetail_customer_name_default))
-            }
-            Mode.OrderCreation -> {
-                binding.orderStatusHeader.isVisible = false
-            }
-        }
+        binding.orderStatusHeader.text =
+            order.getBillingName(context.getString(R.string.orderdetail_customer_name_default))
 
         updatePosTag(order)
     }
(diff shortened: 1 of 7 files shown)


hichamboushaba:
\ciab_order_detail_readonly

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailOrderStatusView.kt
```diff
@@ -80,16 +62,15 @@ class OrderDetailOrderStatusView @JvmOverloads constructor(
...
             }
-            Mode.OrderCreation -> {
-                with(binding.orderStatusEditImage) {
-                    isVisible = true
-                    setOnClickListener(editOrderStatusClickListener)
-                }
+            Mode.ReadOnly -> {
+                binding.orderStatusEditImage.isVisible = false
+                binding.orderStatusContainer.isClickable = false
+                binding.orderStatusContainer.isFocusable = false
```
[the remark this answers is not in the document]
hichamboushaba: Addressed

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/views/OrderDetailOrderStatusView.kt
```diff
@@ -59,18 +49,10 @@ class OrderDetailOrderStatusView @JvmOverloads constructor(
...
-            Mode.OrderEdit -> {
-                "${date.getMediumDate(context)}, ${date.getTimeString(context)}"
-            }
-        }
+        return "${date.getMediumDate(context)}, ${date.getTimeString(context)}"
     }
 
-    fun initView(mode: Mode, editOrderStatusClickListener: EditStatusClickListener) {
-        this.mode = mode
+    fun initView(mode: Mode, editOrderStatusClickListener: EditStatusClickListener = {}) {
```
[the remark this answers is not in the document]
hichamboushaba: Addressed

irfano (changes requested):
Requesting changes for two issues:

- Bulk status update in the order list is still available. CIAB users can still change orders from there.
Image

- There is also a UI regression in the status section. Please check my before/after screenshots below. Paddings were changed unintentionally.

**Before | After**
Image
Image

Please address these before merging.

hichamboushaba:
Thanks @irfano for the review, good catches, both issues should be fixed:

| | |
| - | - |
| Screenshot_20260317_144717 | Screenshot_20260317_144643 |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListFragment.kt
```diff
@@ -120,11 +122,22 @@ class OrderListFragment :
...
-        maxSelectionCount = OrderListViewModel.BULK_UPDATE_COUNT_LIMIT
-    )
+    private val selectionPredicate by lazy {
+        MutableMultipleSelectionPredicate<Long>(
+            // The only option we support with multi-select is order status editing, so disable multi-select
+            // if the feature is not supported
+            maxSelectionCount = if (ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.OrderStatusEditing)) {
+                OrderListViewModel.BULK_UPDATE_COUNT_LIMIT
+            } else {
+                0
```
irfano: Setting `maxSelectionCount` to `0` feels a bit hacky to me. What do you think about gating `addSelectionTracker()`, which we call in `onViewCreated()`, instead?

hichamboushaba: Hmm, TBH I don't see a big of difference, the selection predicate role is to control whether selection is enabled or not, so using it doesn't seem hacky to me 😅 .
I applied your suggestion in ff95fbb5a2e937fe0505eaa13e91e84568ce9c25, it's a good approach too.

irfano: Yes, the behavior is the same. But it seems clearer to me not to call `addSelectionTracker()` or initialize a `SelectionTracker` if we won't use it..
Btw, this broke order detail navigation because `OrderDetailAdapter` doesn't navigate when `tracker` is `null`. Check this patch for the fix.

[Handle_null_selection_tracker_in_order_list_tap_guard.patch](https://github.com/user-attachments/files/26062323/Handle_null_selection_tracker_in_order_list_tap_guard.patch)

hichamboushaba: Patch applied.

irfano (approved):
LGTM! 👍🏻

PR #15551, opened 2026-03-18
Document cross-tool local agent notes

kidinov:
### Description
Document that `AGENTS.md` is the shared agent instruction file for Codex and other agentic AI tools, and that `LOCAL_AGENT_NOTES.md` can be used for repository-local preferences when present.

### Test Steps
1. Open `AGENTS.md`.
2. Confirm the new Cross-Tool Instructions section explains the shared role of `AGENTS.md` and `LOCAL_AGENT_NOTES.md`.
3. Confirm no code or runtime behavior changed.

### Images/gif
N/A

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,8 +4,8 @@ This file provides guidance to AI agents when working with code in this reposito
 
 ## Cross-Tool Instructions
 
-- Use `LOCAL_AGENT_NOTES.md` for repository-local AI preferences when present
-- `CLAUDE.local.md` may point to `LOCAL_AGENT_NOTES.md` for backward compatibility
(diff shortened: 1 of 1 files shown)


malinajirka (approved):
I'm slightly worried this might backfire => push us towards having custom setups. I'm thinking ideally, we'd all improve the shared setup together so everyone can benefit from the best possible setup. Wdyt?

kidinov:
> I'm slightly worried this might backfire => push us towards having custom setups. I'm thinking ideally, we'd all improve the shared setup together so everyone can benefit from the best possible setup. Wdyt?

That’s certainly true, but it doesn’t imply that we can’t have our local “settings.” In fact, they’re valuable for experimentation and things that heavily depend on individual preferences. While I agree that we should largely share good practices, it doesn’t mean we should prohibit local settings

PR #15557, opened 2026-03-19
[WOOMOB-2236] Parse CIAB fulfillment status from order meta

irfano:
### Description
Fixes WOOMOB-2236
Parse CIAB fulfillment status from order metadata and expose it on the order model.
This preserves `_fulfillment_status` in cached orders and maps the known Woo values into a dedicated fulfillment status enum with a safe fallback for unknown values.

### Test Steps
1. Fresh install the app.
2. Sign in to a CIAB site.
3. Open the order list.
4. If you are using the optional debug patch below, apply the patch and verify Logcat contains lines like `OrderListItemDataSource > orderId=190 raw_fulfillment_status=unfulfilled mapped_fulfillment_status=UNFULFILLED`.
5. Check the same orders in Woo web and confirm the fulfillment values match what the app reports.

### Images/gif
N/A

### Optional Debug Patch
```diff
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListItemDataSource.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListItemDataSource.kt
index e147e19c019..b0593a21fa6 100644
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListItemDataSource.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListItemDataSource.kt
@@ -2,6 +2,7 @@ package com.woocommerce.android.ui.orders.list
 
 import com.woocommerce.android.R
 import com.woocommerce.android.extensions.getBillingName
+import com.woocommerce.android.model.Order
 import com.woocommerce.android.model.TimeGroup
 import com.woocommerce.android.model.TimeGroup.GROUP_FUTURE
 import com.woocommerce.android.model.TimeGroup.GROUP_OLDER_MONTH
@@ -16,6 +17,7 @@ import com.woocommerce.android.ui.orders.list.OrderListItemUIType.LoadingItem
 import com.woocommerce.android.ui.orders.list.OrderListItemUIType.OrderListItemUI
 import com.woocommerce.android.ui.orders.list.OrderListItemUIType.SectionHeader
 import com.woocommerce.android.util.DateUtils
+import com.woocommerce.android.util.WooLog
 import com.woocommerce.android.viewmodel.ResourceProvider
 import kotlinx.coroutines.runBlocking
 import org.wordpress.android.fluxc.Dispatcher
@@ -23,6 +25,8 @@ import org.wordpress.android.fluxc.generated.WCOrderActionBuilder
 import org.wordpress.android.fluxc.model.LocalOrRemoteId
 import org.wordpress.android.fluxc.model.WCOrderListDescriptor
 import org.wordpress.android.fluxc.model.WCOrderSummaryModel
+import org.wordpress.android.fluxc.model.metadata.WCMetaData.OrderFulfillmentMetadataKeys.FULFILLMENT_STATUS
+import org.wordpress.android.fluxc.model.metadata.get
 import org.wordpress.android.fluxc.model.list.datasource.ListItemDataSourceInterface
 import org.wordpress.android.fluxc.store.WCOrderStore
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrderListPayload
@@ -67,6 +71,13 @@ class OrderListItemDataSource @Inject constructor(
                 if (order == null) {
                     LoadingItem(orderId)
                 } else {
+                    val rawFulfillmentStatus = order.metaData[FULFILLMENT_STATUS]?.valueAsString
+                    val fulfillmentStatus = Order.FulfillmentStatus.fromApiValue(rawFulfillmentStatus)
+                    WooLog.d(
+                        WooLog.T.ORDERS,
+                        "OrderListItemDataSource > orderId=${order.orderId} raw_fulfillment_status=$rawFulfillmentStatus " +
+                            "mapped_fulfillment_status=$fulfillmentStatus"
+                    )
                     @Suppress("DEPRECATION_ERROR")
                     OrderListItemUI(
                         orderId = order.orderId,
```

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
@@ -71,7 +71,7 @@ data class Order(
                 "fulfilled" -> FULFILLED
                 "partially_fulfilled" -> PARTIALLY_FULFILLED
                 "unfulfilled" -> UNFULFILLED
-                null, "", "no_fulfillments" -> NO_FULFILLMENTS
+                null -> NO_FULFILLMENTS
                 else -> UNKNOWN
             }
         }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
@@ -52,10 +52,29 @@ data class Order(
     val giftCardDiscountedAmount: BigDecimal?,
     val shippingTax: BigDecimal,
     val salesChannel: SalesChannel,
+    val fulfillmentStatus: FulfillmentStatus = FulfillmentStatus.NO_FULFILLMENTS,
 ) : Parcelable {
     enum class SalesChannel {
         POS,
         NON_POS
+    }
+
+    enum class FulfillmentStatus {
+        FULFILLED,
+        PARTIALLY_FULFILLED,
+        UNFULFILLED,
+        NO_FULFILLMENTS,
+        UNKNOWN;
+
+        companion object {
+            fun fromApiValue(value: String?): FulfillmentStatus = when (value) {
+                "fulfilled" -> FULFILLED
+                "partially_fulfilled" -> PARTIALLY_FULFILLED
+                "unfulfilled" -> UNFULFILLED
+                null, "", "no_fulfillments" -> NO_FULFILLMENTS
+                else -> UNKNOWN
+            }
+        }
     }
 
     @IgnoredOnParcel
@@ -393,7 +412,8 @@ data class Order(
                 selectedGiftCard = "",
                 giftCardDiscountedAmount = null,
                 shippingTax = BigDecimal(0),
-                salesChannel = SalesChannel.NON_POS
+                salesChannel = SalesChannel.NON_POS,
+                fulfillmentStatus = FulfillmentStatus.NO_FULFILLMENTS
             )
         }
 
--- a/libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderTest.kt
+++ b/libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderTest.kt
@@ -240,6 +240,24 @@ internal class StripOrderTest {
             .doesNotContain(redundantMemberKey)
     }
 
+    @Test
+    fun `when stripping order, then _fulfillment_status meta data is preserved`() = runTest {
+        // given
+        val metaDataFromRemote = listOf(
+            WCMetaData(1, redundantMemberKey, "sample value"),
+            WCMetaData(2, WCMetaData.OrderFulfillmentMetadataKeys.FULFILLMENT_STATUS, "fulfilled")
+        )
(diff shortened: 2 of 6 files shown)


libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderTest.kt
```diff
@@ -240,6 +240,24 @@ internal class StripOrderTest {
             .doesNotContain(redundantMemberKey)
     }
 
+    @Test
+    fun `when stripping order, then _fulfillment_status meta data is preserved`() = runTest {
+        // given
```
JorgeMucientes, line 245: Minor np, I think the given when then commments in tests are probably not needed

WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
```diff
@@ -52,12 +52,31 @@ data class Order(
     val giftCardDiscountedAmount: BigDecimal?,
     val shippingTax: BigDecimal,
     val salesChannel: SalesChannel,
+    val fulfillmentStatus: FulfillmentStatus = FulfillmentStatus.NO_FULFILLMENTS,
```
JorgeMucientes, line 55: Just double checking. When we don't have a fullfillment status do we want to default to `NO_FULFILLMENTS` or to `UNKNOWN` instead?

JorgeMucientes (approved):
Nicely done! Parsing works as expected :shipit: 
I left one question just to make sure the parsing to `NO_FULFILLMENTS` is intentional. From the web desktop experience what I see is they display a big dash "-" when `fulfillment_status` is null.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderTest.kt
irfano: I think they can be useful sometimes, since these comments help separate parts of the test case when reading the code. Plus, we use these comments in many other places as well, including this class. So, I’d like to keep them to maintain consistency.

WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt
irfano: Good questions, @JorgeMucientes! I had mapped `"no_fulfillments"` and `""` to `FulfillmentStatus.NO_FULFILLMENTS`. However, the endpoint actually returns neither an empty string (`""`) nor `"no_fulfillments"`. I initially thought it could return  `"no_fulfillments"` based on [this code](https://github.com/woocommerce/woocommerce/blob/d26420ad4c6ff39ec7cca7853954e1fa96ab2cc5/plugins/woocommerce/src/Admin/Features/Fulfillments/FulfillmentUtils.php#L498C5-L498C20)), but it turns out it doesn't return any status when there are no fulfillments. So I updated the code and removed mapping of `""` and `"no_fulfillments"` to `FulfillmentStatus.NO_FULFILLMENTS`: 2d999e18fbc47d45f9e50bdd1de7c830ff050ae3

PR #15558, opened 2026-03-20
Add openssl gem to fix translations download job

AdamGrzybkowski:
### Description
CI fails to download translations. This PR adds openssl gem to fix that. 

### Images/gif

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,3 +22,10 @@ gem 'fastlane-plugin-wpmreleasetoolkit', '~> 13.8'
 group :screenshots, optional: true do
   gem 'rmagick', '~> 4.1'
 end
+
+# To avoid errors like:
+#
+# SSL_connect returned=1 errno=0 peeraddr=3.5.132.155:443 state=error: certificate verify failed (unable to get certificate CRL)
+#
+# See https://github.com/ruby/openssl/issues/949
+gem 'openssl', '~> 4.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -271,6 +271,7 @@ GEM
       faraday (>= 1, < 3)
       sawyer (~> 0.9)
     open4 (1.3.4)
+    openssl (4.0.1)
     options (2.3.2)
     optparse (0.8.1)
     os (1.1.4)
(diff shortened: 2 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/283">24.3 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Gemfile
```diff
@@ -22,3 +22,10 @@ gem 'fastlane-plugin-wpmreleasetoolkit', '~> 13.8'
 group :screenshots, optional: true do
   gem 'rmagick', '~> 4.1'
 end
+
+# To avoid errors like:
+#
+# SSL_connect returned=1 errno=0 peeraddr=3.5.132.155:443 state=error: certificate verify failed (unable to get certificate CRL)
+#
+# See https://github.com/ruby/openssl/issues/949
+gem 'openssl', '~> 4.0'
```
AliSoftware, line 31: I'm not entirely sure we need to pin openssl to 4.x vs let it use latest 3.x (to be honest I'm not sure if one needs the major version of OpenSSL gem to be in line with the major version of Ruby — and since we're using Ruby 3.2.2 expect to use OpenSSL gem 3.x too — or if there's no such constraint.

Did you actually run `bundle install` locally to have the `Gemfile.lock` be updated (which would prove that pinning this version still worked without issue)? Or were the changes in the `Gemfile.lock` made by an AI agent editing the file directly (which means it didn't prove that version would be compatible or not)?

I guess we can also just merge and see if gems will still be happy to build next time we have a job that actually needs to run `bundle install` on CI… (as in this tooling PR most of the CI jobs are skipped due to no change in codebase)

AliSoftware (approved):
Approving to unblock, with a question

Gemfile
AdamGrzybkowski: >Did you actually run bundle install locally to have the Gemfile.lock be updated (which would prove that pinning this version still worked without issue)?

I did run `bundle install`.

AliSoftware: Cool, then TIL that openssl 4 is not just for Ruby 4 but works with Ruby 3 too 👍

:shipit:

PR #15565, opened 2026-03-23
[WOOMOB-2528] POS CIAB feature gating and IPP hiding

kidinov:
### Description
Fixes WOOMOB-2528
Fixes WOOMOB-2530
Fixes WOOMOB-2538

Implements POS and IPP feature gating for CIAB sites per the pdfdoF-8QX-p2.

@samiuelson @malinajirka sorry the PR initially was small but due to the changing requirements and proposed improvements got much bigger 🥲

#### POS Gating
CIAB sites without a Pro plan see an ineligibility screen ("Pro plan required") with "Learn More" (opens pricing page) and "Exit POS" buttons. Pro plan CIAB users and non-CIAB users are not affected.

Eligible plan slugs: `woo_hosted_pro_plan_monthly`, `woo_hosted_pro_plan_yearly`.

#### IPP Hiding
IPP is disabled **only for non-Pro CIAB users**. Pro plan CIAB users can use IPP normally. Non-CIAB sites are not affected.

What is hidden for non-Pro CIAB users:
- Card reader and Tap to Pay rows in the payment methods screen
- Payments settings screen from the main menu

What is shown instead:
- A footer in the payment methods screen with the Pro upgrade copy and a "Learn more" link (same URL as POS ineligibility screen)
- The Collect Payment button in order details **stays visible** (tapping it shows cash + upgrade footer)

#### Refresh After Upgrade
When the user taps "Learn more", goes to the browser, upgrades, and comes back — the app refetches the site from the backend and re-checks eligibility. This works on both the POS eligibility screen and the payment methods screen.

#### Authenticated WebView
Both "Learn More" links (POS eligibility screen and payment methods screen) now open in an authenticated WebView instead of a Chrome Custom Tab. This ensures the user is already logged in on the WordPress.com upgrade page.

- **Payment selection screen**: Uses the existing `AuthenticatedWebViewFragment` via `LaunchUrlInAuthenticatedWebView` event (only for the CIAB upgrade link; the regular IPP learn more link still uses Chrome Custom Tab since it goes to public woocommerce.com docs)
- **POS eligibility screen**: Uses a new reusable `WooPosAuthenticatedWebView` composable with its own Compose navigation route (`OpenWebView` nav event), backed by `WCWebView` + `WebViewAuthenticator`

#### Analytics
- CIAB upgrade "Learn more" tap on the payment selection screen fires `IN_PERSON_PAYMENTS_LEARN_MORE_TAPPED` with source `ciab_upgrade` (reuses existing event, distinct from the regular `payment_methods` source)

#### Changes
- Add `CiabPlanUpgradeRequired` reason to `NonLaunchabilityReason`
- Add CIAB plan check in `WooPosCanBeLaunchedInTab` (before version/currency checks)
- Make `CIABSiteGateKeeper` plan-aware: `InPersonPayments` checks plan slug, not just CIAB status
- Extract shared `CIAB_PRO_PLAN_SLUGS` constant used by both gatekeeper and POS launchability
- Update eligibility screen with sealed state hierarchy (`RetryableIneligible` / `CiabPlanUpgradeRequired`)
- Replace `LearnMoreIpp` data class with sealed interface (`Standard` / `CiabUpgrade` / `Hidden`)
- Add refresh-on-resume with site refetch for both POS eligibility and payment methods screens
- Separate `recheckEligibility()` from `retryEligibilityCheckTapped()` to avoid false retry analytics on resume
- Rename `CIABAffectedFeature.WooPayments` to `InPersonPayments`
- Fix "point of sale" casing to lowercase
- Add `pos_ineligible_ui_learn_more_tapped` analytics event
- Use authenticated WebView for CIAB upgrade links (payment selection + POS eligibility)
- Add reusable `WooPosAuthenticatedWebView` component with navigation support
- Track CIAB upgrade learn more tap with `ciab_upgrade` source

### Test Steps

**POS gating (non-Pro CIAB site):**
1. Use a CIAB site with a free plan (slug `woo_hosted_free_plan`)
2. Open POS tab — ineligibility screen shows "Pro plan required"
3. Tap "Learn More" — opens pricing page in an **authenticated WebView** (user should be logged in)
4. Tap "Exit POS" — exits POS
5. There should be no "Retry" button on this screen

**POS gating (Pro CIAB site):**
6. Use a CIAB site with Pro plan (`woo_hosted_pro_plan_monthly` or `_yearly`)
7. POS should launch normally

**IPP hiding (non-Pro CIAB site):**
8. Open an unpaid order on a non-Pro CIAB site
9. Tap "Collect Payment" — the button should be visible
10. On the payment methods screen: card reader and Tap to Pay rows should be hidden, cash should be visible
11. A footer with the Pro upgrade copy and "Learn more" link should appear below the payment methods
12. Tap "Learn more" — opens pricing page in an **authenticated WebView**
13. The Payments item in More Menu should be hidden

**IPP (Pro CIAB site):**
14. On a Pro CIAB site, card reader, Tap to Pay, and Payments menu should all be visible

**Non-CIAB site:**
15. On a non-CIAB site, POS and IPP should work as before, no changes

**Refresh after upgrade:**
16. On POS ineligibility screen, tap "Learn More", come back — screen should re-check eligibility
17. On payment methods screen (non-Pro CIAB), tap "Learn more" in the footer, come back — screen should refresh

### Images/gif

https://github.com/user-attachments/assets/973b3566-66d7-48e8-9387-9f1c3c737efb

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
@@ -2,7 +2,7 @@ package com.woocommerce.android.ciab
 
 enum class CIABAffectedFeature {
     Blaze,
-    WooPayments,
+    InPersonPayments,
     WooShippingSplitShipments,
     GroupedProducts,
     VariableProducts,
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
@@ -27,10 +27,24 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
         return !CIAB_PRO_PLAN_SLUGS.contains(site.planProductSlug)
     }
 
+    fun buildPlanUpgradeUrl(): String {
+        val siteUrl = selectedSite.getOrNull()?.url
+        val siteSlug = siteUrl
+            ?.removePrefix("https://")
+            ?.removePrefix("http://")
+            ?.trimEnd('/')
+        return if (siteSlug != null) {
+            "$LEARN_MORE_BASE_URL?siteSlug=$siteSlug"
+        } else {
+            LEARN_MORE_BASE_URL
+        }
+    }
+
     companion object {
         val CIAB_PRO_PLAN_SLUGS = setOf(
             "woo_hosted_pro_plan_monthly",
             "woo_hosted_pro_plan_yearly",
         )
+        private const val LEARN_MORE_BASE_URL = "https://wordpress.com/setup/woo-hosted-plans/"
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
@@ -9,7 +9,7 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
     ): Boolean {
         return when (feature) {
             CIABAffectedFeature.POS -> true
-            CIABAffectedFeature.InPersonPayments -> !isCurrentSiteNonProCIAB()
+            CIABAffectedFeature.InPersonPayments -> !isCurrentSiteCIAB() || isCurrentSiteProCIAB()
             else -> !isCurrentSiteCIAB()
         }
     }
@@ -21,10 +21,9 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
     fun isCurrentSiteCIAB(): Boolean =
         selectedSite.getOrNull()?.isCIABSite() ?: false
 
-    private fun isCurrentSiteNonProCIAB(): Boolean {
+    private fun isCurrentSiteProCIAB(): Boolean {
         val site = selectedSite.getOrNull() ?: return false
-        if (!site.isCIABSite()) return false
-        return !CIAB_PRO_PLAN_SLUGS.contains(site.planProductSlug)
+        return site.isCIABSite() && CIAB_PRO_PLAN_SLUGS.contains(site.planProductSlug)
     }
 
     fun buildPlanUpgradeUrl(): String {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
@@ -8,8 +8,7 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
         feature: CIABAffectedFeature
     ): Boolean {
         return when (feature) {
-            CIABAffectedFeature.POS,
-            CIABAffectedFeature.WooPayments -> true
+            CIABAffectedFeature.POS -> true
 
             else -> !isCurrentSiteCIAB()
         }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
@@ -9,7 +9,7 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
     ): Boolean {
         return when (feature) {
             CIABAffectedFeature.POS -> true
-
+            CIABAffectedFeature.InPersonPayments -> !isCurrentSiteNonProCIAB()
             else -> !isCurrentSiteCIAB()
         }
     }
@@ -20,4 +20,17 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
 
     fun isCurrentSiteCIAB(): Boolean =
         selectedSite.getOrNull()?.isCIABSite() ?: false
+
+    private fun isCurrentSiteNonProCIAB(): Boolean {
+        val site = selectedSite.getOrNull() ?: return false
+        if (!site.isCIABSite()) return false
+        return !CIAB_PRO_PLAN_SLUGS.contains(site.planProductSlug)
+    }
+
+    companion object {
+        val CIAB_PRO_PLAN_SLUGS = setOf(
+            "woo_hosted_pro_plan_monthly",
+            "woo_hosted_pro_plan_yearly",
+        )
+    }
 }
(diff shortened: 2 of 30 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABAffectedFeature.kt
```diff
@@ -2,7 +2,7 @@ package com.woocommerce.android.ciab
 
 enum class CIABAffectedFeature {
```
staskus, line 3: It's something iOS devs could take as well. 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
```diff
@@ -21,4 +20,31 @@ class CIABSiteGateKeeper @Inject constructor(private val selectedSite: SelectedS
 
     fun isCurrentSiteCIAB(): Boolean =
         selectedSite.getOrNull()?.isCIABSite() ?: false
+
+    private fun isCurrentSiteNonProCIAB(): Boolean {
+        val site = selectedSite.getOrNull() ?: return false
+        if (!site.isCIABSite()) return false
```
staskus: It's a bit hard to read those double-negations, but I understand that we want to enable IPP for:
- Non-CIAB sites
- Pro CIAB sites

so I guess `CIABAffectedFeature.InPersonPayments -> !isCurrentSiteNonProCIAB()` is fine

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodViewModel.kt
```diff
@@ -499,11 +529,23 @@ class SelectPaymentMethodViewModel @Inject constructor(
             wooCommerceStore.getSiteSettings(selectedSite.get())?.currencyCode ?: ""
         }
 
+    fun onResumed() {
+        if (!hasOpenedCiabLearnMore) return
```
staskus, line 533: Not sure if there's a pattern for this on Android:

The hasOpenedLearnMore boolean flag + onResumed() check is used for both POS and IPP. Would it be simpler to pass an onReturnFromLink callback through the state instead? That way, the "refresh on return" behavior is defined where the link is built, rather than scattered across a flag, a lifecycle observer, and an onResumed method?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
```diff
@@ -81,47 +112,73 @@ fun WooPosEligibilityScreen(
...
+                    state = WooPosButtonState.ENABLED,
+                    modifier = buttonModifier,
+                )
+            }
+
+            is WooPosEligibilityRetryState.Loading -> {
+                WooPosButton(
+                    text = stringResource(id = R.string.woopos_eligibility_retry_check_label),
+                    onClick = onRetry,
+                    state = WooPosButtonState.LOADING,
```
staskus, line 156: When loading, the title disappears:

Image

After loading finishes, everything disappears for half a second before transitioning to POS:

Image

We _could_ show the same full-screen loading state for the whole view, not just for the button. We do that on iOS. Maybe that would make it a bit easier?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -4032,6 +4032,10 @@
     <string name="woopos_eligibility_screen_unable_to_load">Unable to load</string>
     <string name="woopos_eligibility_screen_US_country_name">the United States</string>
     <string name="woopos_eligibility_screen_UK_country_name">the United Kingdom</string>
+    <string name="woopos_eligibility_ciab_title">Pro plan required</string>
+    <string name="woopos_eligibility_reason_ciab_plan_upgrade">Accept payments in person for just 2.70% + $0.10 per transaction. Upgrade to Pro to access tap-to-pay on your phone and our full point of sale system with real-time inventory and order syncing.</string>
+    <string name="woopos_eligibility_reason_ciab_plan_upgrade_html">Accept payments in person for just 2.70% + $0.10 per transaction. &lt;b&gt;Upgrade to Pro&lt;/b&gt; to access tap-to-pay on your phone and our full point of sale system with real-time inventory and order syncing.</string>
```
staskus, line 4037: Nice touch on bolding it 👍

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityViewModelTest.kt
```diff
@@ -1,5 +1,7 @@
 package com.woocommerce.android.ui.woopos.eligibility
 
+import com.woocommerce.android.ciab.CIABSiteGateKeeper
```
staskus, line 4: Could consider adding some tests for all those cases 👍

staskus (approved):
✅ The changes are good and clear. Works as expected. The business logic matches the iOS. 

I like how Android deals with CIAB-affected features a bit cleaner than iOS. Something to take away. 

Left a couple smaller comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ciab/CIABSiteGateKeeper.kt
kidinov: Good point! Simplified by flipping to a positive helper `isCurrentSiteProCIAB()`, so it now reads: `!isCurrentSiteCIAB() || isCurrentSiteProCIAB()` — "IPP is supported if the site is not CIAB, or if it's CIAB Pro."

fa86fd85ca1

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodViewModel.kt
kidinov: Yeah, it's not the prettiest pattern. Unfortunately on Android there's no clean way around it — we need to react to the lifecycle resume after returning from the webview, and a flag + onResumed() is the standard way to do that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
kidinov: Good catch! Fixed by:
- Keeping title and suggestion text visible during loading (preserved from the previous `Ineligible` state)
- Showing a loading button during `Eligible` transition instead of empty content
- Starting with `null` state (nothing rendered until initialized) so `Loading` fields can be non-nullable

Full-screen loading for the `Eligible` case felt like overkill since most users won't actually upgrade — they'll just return to the same screen.

a8890916806

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityViewModelTest.kt
kidinov: Added a couple of tests for the CIAB-specific flows: plan upgrade state initialization and the learn more → onResumed recheck path.

c53178dcc3b

PR #15572, opened 2026-03-26
[WOOMOB-2445] Add edit button for POS receipt information

kidinov:
### Description
Fixes WOOMOB-2445

Add an edit pencil icon button to the POS receipt information card in settings. Tapping it opens the POS settings page in an authenticated WebView where the user can edit receipt details (store name, address, phone, email, refund policy). Receipt data is re-fetched when returning from the webview. Uses the correct URL based on whether the site is CIAB or not.

Based on #15565 for the `WooPosAuthenticatedWebView` component and `OpenWebView` navigation event.

### Test Steps
1. Open POS > Settings > Store
2. In the "Receipt information" card, tap the pencil icon
3. Verify it opens the POS settings page in an authenticated WebView
4. Edit some receipt info and save
5. Go back — verify the receipt info refreshes with the updated values
6. (CIAB site) Verify the URL uses `next-admin` path
7. (Non-CIAB site) Verify the URL uses `wc-settings&tab=point-of-sale` path

### Images/gif

https://github.com/user-attachments/assets/06c8e697-c9e7-4c10-9655-b2ece65c77ab


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
```diff
@@ -51,4 +82,12 @@ class WooPosSettingsStoreViewModel @Inject constructor(
             _state.value = _state.value.copy(receiptState = newReceiptState)
         }
     }
+
+    data class EditReceiptTarget(val url: String)
+
+    private companion object {
+        const val POS_SETTINGS_PATH = "/wp-admin/admin.php?page=wc-settings&tab=point-of-sale"
```
staskus: I'll be doing some auto-redirection on WOOMOB-2482 from wp-admin to ciab to happen automatically. I could include the POS settings page as well.

kidinov: That would be great!

Let's see if this direction in general finds any support

p1774519029850439/1772797585.357519-slack-C070SJRA8DP

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
```diff
@@ -2,26 +2,56 @@ package com.woocommerce.android.ui.woopos.settings.details.store
...
 
+    private val _openEditReceiptEvent = MutableSharedFlow<EditReceiptTarget>()
+    val openEditReceiptEvent: SharedFlow<EditReceiptTarget> = _openEditReceiptEvent.asSharedFlow()
+
     init {
         loadStoreData()
         loadReceiptData()
     }
 
+    fun onEditReceiptClicked() {
```
malinajirka, line 39: 💡 We can consider adding an analytics tracking here. All comparable user-initiated actions in POS settings are tracked (`StoreDetailsTapped`, `HardwareTapped`, `GetSupportTapped`, etc.). Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
```diff
@@ -2,26 +2,56 @@ package com.woocommerce.android.ui.woopos.settings.details.store
...
+        viewModelScope.launch {
+            val url = buildReceiptSettingsUrl()
+            _openEditReceiptEvent.emit(EditReceiptTarget(url))
+        }
+    }
+
+    fun onScreenResumed() {
```
malinajirka, line 47: 🔍 np: Since the `init` block also calls `loadReceiptData()`, and `ON_RESUME` fires on initial display too, receipt data will be fetched twice on first screen appearance. This is harmless (the second call just overwrites the first result) and the ON_RESUME is primarily for refreshing after returning from the WebView — just flagging for awareness.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
```diff
@@ -51,4 +82,12 @@ class WooPosSettingsStoreViewModel @Inject constructor(
             _state.value = _state.value.copy(receiptState = newReceiptState)
         }
     }
+
+    data class EditReceiptTarget(val url: String)
+
+    private companion object {
```
malinajirka: 💡 We might want to consider moving this to WooPosAppUrls, wdyt?

malinajirka (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
kidinov: Good catch, added `EditReceiptTapped` tracking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
kidinov: Good point. Removed `loadReceiptData()` from `init` — `onScreenResumed()` already covers initial display, so no need to duplicate it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
kidinov: @iamgabrielma probably the same should be added on iOS then

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/store/WooPosSettingsStoreViewModel.kt
kidinov: Moved the constants there. Thanks!

PR #15574, opened 2026-03-26
[WOOMOB-2544] Run incremental sync after manual full sync from Settings

malinajirka:
### Description
Fixes WOOMOB-2544

When a user triggers a manual full catalog sync from POS Settings > Local Catalog, newly added products may not appear because:
1. The incremental sync was not run after the full sync
2. The product list was not refreshed after syncing

The background `WooPosLocalCatalogSyncWorker` correctly runs incremental sync after full sync, but the Settings ViewModel did not. This adds the missing incremental sync call after a successful full sync and sends a `RefreshProductList` event so the product list picks up the newly synced products.

**Note:** Long term, the product list should observe the local database directly so manual refresh events won't be necessary. This is currently blocked by the need to support the original paginated API approach, whose architecture isn't compatible with reactive database observation.

### Test Steps
1. Open POS Settings > Local Catalog
2. Add a new product to the store (e.g. via web admin)
3. Trigger a manual full catalog sync from the Settings screen
4. Verify the newly added product appears in the catalog

### Images/gif
N/A


staskus (approved):
👍 Works as expected now, thanks for the fix.

PR #15575, opened 2026-03-27
[WOOMOB-359] Fix customer mapper copying billing address into shipping fields

kidinov:
### Description
Partially addresses WOOMOB-359

`WCCustomerMapper.mapToModel()` was mapping `dto.billing` fields into all shipping address fields instead of `dto.shipping`. The reverse mapping in `mapToDTO()` had the same bug — it wrote `model.billing*` fields into the `Shipping` DTO.

This meant that when a **registered customer** was selected during order creation, the app fetched their full data from `/wc/v3/customers/{id}` (which returns separate billing and shipping objects), but then threw away the shipping data and duplicated billing into both addresses.

**What this fixes:** Registered customers (those with a WordPress user account, `user_id > 0`) now correctly get their shipping address populated from the API response.

**What this does NOT fix:** Guest customers found via the analytics API (`wc-analytics/reports/customers`) still show billing data in both address fields. This is because the analytics API returns only a single flat set of address fields — there is no shipping address in the response. Fixing that requires either backend changes or switching to a different API endpoint (see the peaMlT-16R-p2 for details).

The existing tests used identical values for billing and shipping, so they never caught this bug. Updated them to use distinct values.

### Test Steps
1. On a WooCommerce site, ensure you have a **registered customer** with **different** billing and shipping addresses
2. Start a new order creation
3. Tap "Add customer details"
4. Select the registered customer (the one with a username shown)
5. Verify that **Shipping Address** shows the actual shipping data
6. Verify that **Billing Address** shows the actual billing data
7. Confirm the two addresses are different

### Images/gif
Screenshot 2026-03-27 at 15 13 17
image


malinajirka (approved):
LGTM, thanks!

PR #15582, opened 2026-03-30
[Hack Week] Add Settings entry point for Troubleshoot Connection

hichamboushaba:
### Description
Hack Week project to improve the Troubleshooting (Connectivity Tool) screen, bringing Android closer to iOS feature parity.

**Changes in this PR:**
- Move the connectivity tool package out of  to  since it's no longer orders-specific
- Add a "Troubleshoot Connection" entry point in Settings, visible for non-WPCom sites (matching iOS placement below Help & support)
- Skip the WordPress.com servers check for app-password sites, since they don't authenticate via WPCom
- Add  (now reflected by the /WPCom naming) to the Compose screen to conditionally hide the WPCom check card
- Rename  references to  in the connectivity tool code for clarity (e.g.,  → ,  → )

> [!NOTE]
> The JsonSyntaxException / parsing behavior was tightened in : code and KDoc were aligned so parsing failures are treated as  and JSON null handling was adjusted so callers receive  when the JSON is null. This fixes a classification issue where parse errors were being reported as generic errors in some flows (which also affected Sentry categorization).

> [!NOTE]
> One temporary behavior change:  responses are mapped to  on some sites for parity with the previous behaviour, but when running on app-password-only sites we map it to  (so the Troubleshoot screen is not shown for WP.com-hosted sites using app-passwords). This is intended as an interim change while we align Android and iOS behavior more closely.

### Test Steps
1. Log in with a **Jetpack-connected** site
2. Go to Menu → Settings
3. Verify "Troubleshoot Connection" appears below "Help & support"
4. Tap it → verify 4 checks run (Internet, WPCom Servers, Site, Orders)
5. Switch to an **app-password** site
6. Go to Menu → Settings → Troubleshoot Connection
7. Verify only 3 checks run (WPCom Servers check is skipped)
8. Trigger a timeout from Orders on a WP.com-hosted site and confirm the Orders view shows the generic error snackbar (no Troubleshoot screen opened)
9. Verify "No connection issues" summary appears when all checks pass
10. Verify the option is **not** visible for WPCom-hosted sites

### Images/gif

| Settings | Jetpack site (4 checks) | App-password site (3 checks) |
|----------|------------------------|------------------------------|
| settings_troubleshoot_connection | connectivity_tool_screen | connectivity_tool_app_password |

- [ ] I have considered if this change warrants release notes and have added them to  if necessary. Use the "[Internal]" label for non-user-facing changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ArrayOrObjectDeserializer.kt
```diff
@@ -20,17 +20,17 @@ import java.lang.reflect.Array as JArray
...
 class ArrayOrObjectDeserializer : JsonDeserializer<Array<*>> {
     override fun deserialize(
         json: JsonElement,
         typeOfT: Type,
         context: JsonDeserializationContext
-    ): Array<*>? {
+    ): Array<*> {
         val componentType: Type = (typeOfT as? GenericArrayType)?.genericComponentType
-            ?: throw JsonParseException("Expected a GenericArrayType, but got: $typeOfT")
+            ?: throw JsonSyntaxException("Expected a GenericArrayType, but got: $typeOfT")
```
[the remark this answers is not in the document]
hichamboushaba: Good point, I updated the commit to keep the `null` part, I fixed the logic to make sure we return `null` when json is null, before we were returning `emptyList()`.

Side note: this has no effect, just clearer. Gson [implementation](https://github.com/google/gson/blob/1fa9b7a0a994b006b3be00e2df9de778e71e6807/gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java#L92-L94) already returns `null` when json is null without checking the type adapter, so our logic is never used, and this can be confirmed by the test `when json is null, then returns null array` that was green even when we returned `emptyList()`.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ArrayOrObjectDeserializer.kt
```diff
@@ -30,11 +30,11 @@ class ArrayOrObjectDeserializer : JsonDeserializer<Array<*>> {
         context: JsonDeserializationContext
     ): Array<*>? {
         val componentType: Type = (typeOfT as? GenericArrayType)?.genericComponentType
-            ?: throw JsonParseException("Expected a GenericArrayType, but got: $typeOfT")
+            ?: throw JsonSyntaxException("Expected a GenericArrayType, but got: $typeOfT")
```
hichamboushaba, line 33: This is the point I mentioned in the PR description, it's unrelated to the changes, but while I was testing some flows, I noticed that when faked some responses to trigger parsing errors, they were treated as generic errors, and the cause was this.
This also means that our Sentry graph is missing some issues that goes through this deserializer.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ArrayOrObjectDeserializer.kt
```diff
@@ -73,7 +73,7 @@ private fun jsonToList(
     json: JsonElement,
     itemType: Type,
     context: JsonDeserializationContext
-): List<Any?> {
+): List<Any?>? {
```
hichamboushaba, line 76: More on this change in this [thread](https://github.com/woocommerce/woocommerce-android/pull/15582/changes#r3010506416)

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
```diff
@@ -545,6 +545,7 @@ enum class AnalyticsEvent(override val siteless: Boolean = false) : IAnalyticsEv
     PRIVACY_SETTINGS_SHARE_INFO_LINK_TAPPED,
     PRIVACY_SETTINGS_THIRD_PARTY_TRACKING_INFO_LINK_TAPPED,
     SETTINGS_DOMAINS_TAPPED,
+    SETTINGS_TROUBLESHOOT_CONNECTION_TAPPED,
```
hichamboushaba, line 548: This event matches the one used for iOS

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/connectivitytool/useCases/StoreConnectionCheckUseCase.kt
```diff
@@ -30,7 +34,8 @@ class StoreConnectionCheckUseCase @Inject constructor(
     private fun WooResult<WCSSRModel>.parseError() =
         when (error.type) {
             WooErrorType.TIMEOUT -> Failure(FailureType.TIMEOUT)
-            WooErrorType.API_NOT_FOUND -> Failure(FailureType.JETPACK)
+            WooErrorType.API_NOT_FOUND ->
+                if (isAppPasswordSite) Failure(FailureType.GENERIC) else Failure(FailureType.JETPACK)
```
hichamboushaba, line 38: This is a temporary change, using `JETPACK` for `API_NOT_FOUND` doesn't match what iOS [does](https://github.com/woocommerce/woocommerce-ios/blob/c057b2272c6e8150ea901d77636bc033d5aeba6c/WooCommerce/Classes/ViewRelated/Connectivity%20Tool/ConnectivityToolViewModel.swift#L272-L275), but given it was like this I decided to now modify it for now, I just handled app passwords given it's the new thing.

I'll try to revisit this when I'm done with the other changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/connectivitytool/ConnectivityToolScreen.kt
```diff
@@ -39,40 +39,42 @@ import com.woocommerce.android.R
...
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckCardData.StoreOrdersConnectivityCheckData
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckCardData.WPComConnectivityCheckData
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckStatus.Failure
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckStatus.InProgress
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckStatus.NotStarted
+import com.woocommerce.android.ui.connectivitytool.ConnectivityCheckStatus.Success
 
 @Composable
-fun OrderConnectivityToolScreen(viewModel: OrderConnectivityToolViewModel) {
+fun ConnectivityToolScreen(viewModel: ConnectivityToolViewModel) {
```
irfano, line 52: Connectivity tool sounds like something that helps establish a connection. What about `TroubleshootConnectionScreen`?

irfano (approved):
LGTM! 👍🏻
I left some minor comments. Two more notes:
1. You disabled the Troubleshoot screen for WP.com-hosted sites, but it’s still accessible from the order list when a timeout occurs (see OrderList, line 461). Do you plan to apply the same logic there as well?
2. Let's make the icons white in dark mode.

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/connectivitytool/ConnectivityToolScreen.kt
hichamboushaba: I agree, but if you don't mind, I'll try to handle renaming in a separate PR, I have 4 PRs that depend on this branch, and renaming the screen would mean ton of merge conflicts to handle.

hichamboushaba:

> You disabled the Troubleshoot screen for WP.com-hosted sites, but it’s still accessible from the order list when a timeout occurs (see OrderList, line 461). Do you plan to apply the same logic there as well?

I did this just to follow iOS, I intend to discuss this with the team later, I think the tool could be useful for WordPress.com sites too.

> Let's make the icons white in dark mode.

I'll try to address this in the next PRs.

PR #15533, opened 2026-03-13
[WOOMOB-2486] Add FedEx support to shipping label flow

irfano:
### Description
Fixes WOOMOB-2486

Adds the remaining hardcoded FedEx support needed in the Android shipping label flow. This wires FedEx predefined carrier packages into package selection, adds the FedEx pickup URL for purchased labels, and fixes the carrier tab casing to display `FedEx`.

### Test Steps
1. Login to https://mc.a8c.com/secret-store/?secret_id=14089 with store credentials.
2. Navigate to Orders and open order `#75 Guest`.
3. Tap `Create Shipping Label`.
4. Tap `Select a Package`, open the `Carrier` tab, and verify `FedEx` appears with carrier packages.
5. Return to the rates step and verify the FedEx carrier tab label is shown as `FedEx`.
6. Select a FedEx rate and purchase the label.
7. In the purchased label state, tap `Schedule pickup`.
8. Verify the FedEx pickup scheduling page opens.

### Images/gif
Package selection with FedEx predefined packages Shipping rates with FedEx carrier tab casing

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15587, opened 2026-03-30
[WOOMOB-2574] Disable FedEx support locally

irfano:
### Description
Fixes WOOMOB-2574

Disables FedEx in the Android shipping labels flow behind a temporary feature flag.
This hides FedEx in both the package carrier tabs and the shipping rates flow while the mobile support remains turned off.

Slack discussion:

### Test Steps
1. Open an order that can be used to create a shipping label (The site should be connected to staging server).
2. Start the shipping label flow.
3. Go to the package selection screen.
4. Open the `Carrier` tab.
5. Confirm FedEx is not shown in the carrier list.
6. Continue to the shipping rates step.
7. Confirm FedEx rates are not shown.

### Images/gif
|Before|After|
|-|-|
|Screenshot_20260330_232148|Screenshot_20260330_232009|

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
@@ -34,4 +34,5 @@ enum class FeatureFlag(
     AGE_ELIGIBILITY_CHECKS("age_eligibility_checks", localValue = PackageUtils.isDebugBuild()),
     WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1("woo_self_driven_push_notifications_m1", localValue = false),
     WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M2("woo_self_driven_push_notifications_m2", localValue = false),
+    WOO_SHIPPING_FEDEX("woo_shipping_fedex", localValue = false),
 }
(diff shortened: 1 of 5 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
```diff
@@ -34,4 +34,5 @@ enum class FeatureFlag(
     AGE_ELIGIBILITY_CHECKS("age_eligibility_checks", localValue = PackageUtils.isDebugBuild()),
     WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1("woo_self_driven_push_notifications_m1", localValue = false),
     WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M2("woo_self_driven_push_notifications_m2", localValue = false),
+    WOO_SHIPPING_FEDEX("woo_shipping_fedex", localValue = false),
```
AdamGrzybkowski, line 37: ❓ Any reasons for not taking advantage of the new unified feature flag handling and setting the local value to `true` while disabling the feature remotely? Is there any benefit of having it done in a way that requires an extra app update?

AdamGrzybkowski (approved):
Looks good! One question though - couldn't the backend simply filter out FedEx until its ready? 🤔

AdamGrzybkowski:
@irfano Should this PR target 24.4?

irfano:
> @irfano Should this PR target 24.4?

No, because this feature was added in #15533.

WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
irfano: We set `localValue` to `false` because the FedEx implementation is still missing on our side. Setting `localValue` to false means this version does not support FedEx. 
We will set it to `true` after completing WOOMOB-2572.

irfano:
> One question though - couldn't the backend simply filter out FedEx until its ready? 🤔

I’m not sure, actually. If we add this logic to the backend, it would always need to keep that logic to avoid sending FedEx to Android 24.5 and iOS 24.4. Also, I’m not sure whether the backend can filter based on app version. Disabling it on the client side seems like the better option to me.

PR #15592, opened 2026-03-31
[WOOMOB-2590] Open WC Admin in Chrome Custom Tab for CIAB sites

hichamboushaba:
### Description
Fixes WOOMOB-2590

CIAB hides the navigation sidebar when detecting a mobile user agent, which breaks the WC Admin experience where navigation to other screens is expected. For CIAB sites, WC Admin now opens in a Chrome Custom Tab instead of the in-app authenticated WebView. Non-CIAB sites retain the existing behavior.

### Test Steps
1. Log in to a **CIAB site**
2. Go to More menu → tap **WC Admin**
3. Verify it opens in a **Chrome Custom Tab** (browser overlay with toolbar showing the URL)
4. Verify the WC Admin sidebar navigation is visible and functional
5. Log in to a **non-CIAB site** (self-hosted + Jetpack)
6. Go to More menu → tap **WC Admin**
7. Verify it opens in the **in-app authenticated WebView** (existing behavior)

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
@@ -421,12 +421,20 @@ class MoreMenuViewModel @Inject constructor(
 
     private fun onViewAdminButtonClick() {
         trackMoreMenuOptionSelected(VALUE_MORE_MENU_ADMIN_MENU)
-        triggerEvent(
-            MultiLiveEvent.Event.LaunchUrlInAuthenticatedWebView(
-                url = selectedSite.get().adminUrlOrDefault,
-                screenTitle = UiString.UiStringRes(R.string.more_menu_button_wс_admin)
+        val site = selectedSite.get()
+        val adminUrl = site.adminUrlOrDefault
+        if (site.isCIABSite) {
+            // CIAB hides the navigation sidebar when detecting a mobile user agent,
+            // which breaks the WC Admin experience. Open in Chrome Custom Tab instead.
+            triggerEvent(MultiLiveEvent.Event.LaunchUrlInChromeTab(url = adminUrl))
+        } else {
+            triggerEvent(
+                MultiLiveEvent.Event.LaunchUrlInAuthenticatedWebView(
+                    url = adminUrl,
+                    screenTitle = UiString.UiStringRes(R.string.more_menu_button_wс_admin)
+                )
             )
-        )
+        }
     }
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/285">24.5</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/moremenu/MoreMenuViewModel.kt
```diff
@@ -421,12 +421,20 @@ class MoreMenuViewModel @Inject constructor(
...
-            MultiLiveEvent.Event.LaunchUrlInAuthenticatedWebView(
-                url = selectedSite.get().adminUrlOrDefault,
-                screenTitle = UiString.UiStringRes(R.string.more_menu_button_wс_admin)
+        val site = selectedSite.get()
+        val adminUrl = site.adminUrlOrDefault
+        if (site.isCIABSite) {
+            // CIAB hides the navigation sidebar when detecting a mobile user agent,
+            // which breaks the WC Admin experience. Open in Chrome Custom Tab instead.
+            triggerEvent(MultiLiveEvent.Event.LaunchUrlInChromeTab(url = adminUrl))
+        } else {
```
hichamboushaba, line 430: I decided to apply this only for CIAB, because the internal WebView offers a better UX with the auto-authentication that we have.

AdamGrzybkowski (approved):
Changes look good to me 👍 

One thing is that we lose the auto authentication now. I know we discussed this briefly with our team, but shouldn't we discuss this trade-off with the rest of the mobile team?

hichamboushaba:
> One thing is that we lose the auto authentication now. I know we discussed this briefly with our team, but shouldn't we discuss this trade-off with the rest of the mobile team?

I don't know, I think this was clear as a tradeoff, the auto authentication was a new addition to this flow just recently (in this [PR](https://github.com/woocommerce/woocommerce-android/pull/14826)), before we were using Chrome Custom tabs for all users, so IMO this is not a critical change that requires more discussions, if the user was already logged in on their browser, they'll be auto authenticated anyway, WDYT?

AdamGrzybkowski:
Makes sense! :shipit:

PR #15600, opened 2026-04-02
Fix media picker toolbar using purple instead of app colors

kidinov:
Closes: WOOMOB-2596

## Description

The WordPress Media Library picker toolbar shows a **purple background** (the app's `color_primary`) instead of the standard toolbar color used everywhere else in the app.

The mediapicker library defines `color_media_picker_toolbar_background` as `@color/color_primary`. Since Woo's `color_primary` is purple, the toolbar ends up purple. The fix overrides this color resource to use `color_toolbar` instead (white in light mode, dark in dark mode).

## Screenshot (before)

image

## Test Steps

1. Go to **Products → tap a product → Images → Add image → Choose from device**
2. Verify the Media Library toolbar is white (light mode) or dark (dark mode), not purple

## RELEASE-NOTES.txt

- [x] I have considered whether a release note is required — this is a minor visual fix, no release note needed

### DIFF
--- a/WooCommerce/src/main/AndroidManifest.xml
+++ b/WooCommerce/src/main/AndroidManifest.xml
@@ -150,7 +150,7 @@
             android:windowSoftInputMode="adjustResize" />
         <activity
             android:name="org.wordpress.android.mediapicker.ui.MediaPickerActivity"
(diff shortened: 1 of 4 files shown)


malinajirka (approved):
LGTM

PR #15603, opened 2026-04-02
[HACK WEEK] Migrate ReviewListFragment to Jetpack Compose

AdamGrzybkowski:
### Description

> [!NOTE]
> This work was initially started during our meet-up for fun, but the initial results were good enough to polish it a bit and open a PR. It won't hurt to delete some XML. 

Migrate the ReviewListFragment from XML/RecyclerView to Jetpack Compose, along with extracting and improving the shared ReviewListItem composable.

**Note:** The `MaterialContainerTransform` shared element transition from the review list to the detail screen is no longer available. This transition relied on passing a `View` reference via `FragmentNavigatorExtras`, which is not possible from a Compose UI. The detail screen (`ReviewDetailFragment`) still uses XML, so Compose's `SharedTransitionLayout` cannot be used either. There's not much we can do at the moment.

#### Commits

> [!IMPORTANT]
> The first commit changes the UI a bit by matching the item UI in the dashboard with the Review list screen. Unless the UI was different on purpose, that should be a free bonus from reusing the same component. 

1. **Update ReviewListItem style to match XML version** — Fix icon tint to use pending status (purple for HOLD), use `FontWeight.Medium` for title, use color resources for title/description colors to match the XML adapter, add `maxLines = 2`, surface background, and `showDivider`/color parameters for customizability.

2. **Extract ReviewListItem to shared composable** — Move `ReviewListItem` from `DashboardReviewsCard`'s private composable into its own file using Material 3 APIs, so it can be reused by both the dashboard card and the review list screen. Includes `@Preview` composables for all visual states.

3. **Fix initial load not showing reviews when cache is empty** — Move `fetchReviewList` call inside the cache-check coroutine to avoid a race where the fetch completes and sets `isSkeletonShown = false` before the cache check sets it to `true`. This bug existed in the XML version as well.

4. **Migrate ReviewListFragment to Jetpack Compose** — Replace the XML RecyclerView implementation with a Compose `ReviewListScreen`. The fragment is now a thin Compose host. Uses Material 3 components, `WCPullToRefreshBox`, `InfiniteListHandler`, `WCSwitch`, sticky time-group headers, and proper empty/skeleton/loading states.

### Test Steps

1. Open the app and enable `recent reviews` dashboard card
2. Verify the UI
3. Go to Menu -> Reviews
4. Play with the UI - refresh/toggle unread/tap item/scroll etc.
5. Verify all works as previously

### Images/gif

| | Before | After |
|---|---|---|
| Dashboard Card |  Screenshot_20260402_125444 | Screenshot_20260403_152657 |
| Review List | Screenshot_20260402_125454 | Screenshot_20260403_152703 |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/reviews/ReviewListItem.kt
```diff
@@ -0,0 +1,259 @@
...
+
+                if (review.rating > 0) {
+                    AndroidView(
+                        factory = { context ->
+                            RatingBar(context, null, androidx.appcompat.R.attr.ratingBarStyleSmall)
+                        },
+                        update = { ratingBar ->
+                            ratingBar.rating = 100F
+                            ratingBar.numStars = review.rating
+                        }
```
hichamboushaba, line 118: I know this is not new here, but maybe we can just add a comment explaining why we set `rating` to `100F`, at first glance it wasn't clear to me that it's to hide the unfilled stars

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/reviews/ReviewListFragment.kt
```diff
@@ -1,123 +1,87 @@
...
-                    super.onScrollStateChanged(recyclerView, newState)
-
-                    if (!recyclerView.canScrollVertically(1)) {
-                        viewModel.loadMoreReviews()
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        return ComposeView(requireContext()).apply {
```
hichamboushaba: np, we have a helper function `composeView` that you can use to reduce the boilerplate.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/reviews/ReviewListFragment.kt
```diff
@@ -1,123 +1,87 @@
...
+                            onRefresh = {
+                                AnalyticsTracker.track(AnalyticsEvent.REVIEWS_LIST_PULLED_TO_REFRESH)
+                                viewModel.forceRefreshReviews()
+                            },
+                            onLoadMore = { viewModel.loadMoreReviews() },
+                            onUnreadFilterChanged = { viewModel.onUnreadReviewsFilterChanged(it) },
+                            onLearnMoreClick = {
+                                ChromeCustomTabUtils.launchUrl(requireActivity(), AppUrls.URL_LEARN_MORE_REVIEWS)
+                            }
+                        )
```
hichamboushaba: To stay consistent with the codebase conventions, WDYT about having a toplevel `ReviewListScreen(viewModel)` that handles this logic?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/reviews/ReviewListFragment.kt
```diff
@@ -1,123 +1,87 @@
...
+                    val viewState by viewModel.viewStateData.liveData.observeAsState()
+                    val reviews by viewModel.reviewList.observeAsState()
+
+                    viewState?.let { state ->
+                        ReviewListScreen(
+                            reviews = reviews.orEmpty(),
+                            viewState = state,
+                            onReviewClick = { review -> onReviewClick(review) },
+                            onRefresh = {
+                                AnalyticsTracker.track(AnalyticsEvent.REVIEWS_LIST_PULLED_TO_REFRESH)
```
hichamboushaba: np, I know this was like this before, but we can probably move the tracking to the ViewModel.

hichamboushaba (approved):
Nice work @AdamGrzybkowski, it's good to see old parts of the app embracing Compose, and it would be great to build on top of this to migrate `ProductReviewsFragment` too in the future.

I left few comments, but none is blocking, so I'm pre-approving.

PR #15607, opened 2026-04-02
Add UI review skill and screenshot support

irfano:
### Description
This PR adds an internal `ui-review` skill.

This skill takes screenshots of UI changes under different conditions, lists them in a single report, and flags the problematic ones.

The reviewed conditions are:
- Light
- Dark
- Large Font
- Landscape
- RTL
- Data variation 1 - Long strings
- Data variation 2 - Null data
- Empty list and other screen-specific data variations when they are meaningful

The skill uses Android's screenshot testing tool and the screenshots generated by that tool.
During a run, it creates temporary previews in a temp file, runs screenshot generation for those previews, and then builds a report from a fixed template using the generated screenshots.

### Test Steps
1. Make a small visual change in a Compose screen and run the `ui-review` skill in diff mode.
2. Confirm that it creates a report at `WooCommerce/build/reports/ui-review/index.html`.
3. Confirm that the report lists the generated screenshots and flags visible issues when there is a visual problem.
4. Without relying on a diff, run the `ui-review` skill again by giving it a specific Compose target.
5. Confirm that it reviews the provided target and generates the report for that target.

### Images/gif
Screenshot 2026-04-02 at 22 03 27


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/build.gradle
```diff
@@ -77,6 +78,7 @@ def versionProperties = loadPropertiesFromFile(file("${rootDir}/version.properti
 
 android {
     namespace = "com.woocommerce.android"
+    experimentalProperties["android.experimental.enableScreenshotTest"] = true
```
malinajirka: This is also set in `gradle.properties` (line added by this PR). The existing pattern in this project is to set experimental properties only in `gradle.properties` (see `android.experimental.enableTestFixturesKotlinSupport`). Consider removing this line to keep a single source of truth.

WooCommerce/build.gradle
```diff
@@ -215,6 +217,9 @@ android {
 }
 
 dependencies {
+    screenshotTestImplementation(libs.androidx.compose.ui.tooling.main)
```
malinajirka: This may be redundant — `implementation(libs.androidx.compose.ui.tooling.main)` already exists in the dependencies block (around line 427), which makes `ui-tooling` available to all source sets including `screenshotTest`. Can you verify if this is actually needed?

gradle/libs.versions.toml
```diff
@@ -94,6 +94,7 @@ mpandroidchart = 'v3.1.0'
 photoview = '2.3.0'
 robolectric = "4.16.1"
 sentry = '6.1.0'
+screenshot = '0.0.1-alpha13'
```
malinajirka: This is a very early alpha (`0.0.1-alpha13`). Worth verifying it's compatible with the project's AGP version (`8.13.2`). Consider adding a comment noting this is experimental and used only for the internal `ui-review` AI skill.

WooCommerce/build.gradle
malinajirka: This duplicates `android.experimental.enableScreenshotTest=true` already set in `gradle.properties` (line 13). Having both is confusing — if someone sets one to `false` to disable screenshot testing, the other silently overrides it.

Since `gradle.properties` is the standard place for global Gradle/AGP properties and matches how other experimental flags are set in this project, I'd remove this line and keep only the `gradle.properties` entry.

```suggestion
```

gradle/libs.versions.toml
JorgeMucientes: Minor but this is now in [alpha14 version](https://developer.android.com/studio/preview/compose-screenshot-testing-release-notes#alpha14), perhaps we can update it before merging.

WooCommerce/build.gradle
```diff
@@ -2,6 +2,7 @@ import io.sentry.android.gradle.extensions.InstrumentationFeature
 
 plugins {
     alias(libs.plugins.android.application)
+    alias(libs.plugins.compose.screenshot)
```
JorgeMucientes: This would apply the `compose-screenshot` plugin unconditionally, so the screenshotTest* source set and screenshot tasks are wired into every build/sync — not just when the UI review skill runs. The skill only needs an on-demand updateVanillaDebugScreenshotTest (see .agents/skills/ui-review/SKILL.md:184-190), so this is broader than necessary. Combined with android.experimental.enableScreenshotTest=true in gradle.properties:13 and the experimental flag at line 81, the alpha screenshot infra is on globally. Worth gating the plugin (and the deps at lines 220-221) behind a property so normal local/CI builds don't carry it when the skill isn't in use.

.agents/skills/ui-review/SKILL.md
```diff
@@ -0,0 +1,303 @@
...
+## Cleanup
+
+Before finishing:
+
+1. Delete temporary preview source files
+2. Copy the screenshots needed by the report into `WooCommerce/build/reports/ui-review/assets`
+3. Delete `WooCommerce/src/screenshotTestVanillaDebug`
+4. Delete `WooCommerce/build/outputs/screenshotTest-results`
+5. Delete `WooCommerce/build/reports/screenshotTest`
+6. Delete tool-generated screenshot test reports under `WooCommerce/build/reports/tests` for this run
```
JorgeMucientes: Could potential issues to review here. 

Step 3 deletes the entire WooCommerce/src/screenshotTestVanillaDebug directory, not just the temporary preview file(s) the skill created in step 1. Today nothing is committed there, but as soon as anyone adds real screenshot tests to that source set, this will wipe them on every skill run. I'd either scope step 3 to the specific temp file(s), or drop it since step 1 already covers them.

Same caution for step 6 (line 303): WooCommerce/build/reports/tests is the shared Gradle test reports directory, so "delete tool-generated screenshot test reports for this run" leaves it to the agent to figure out which subset is safe to remove — easy to over-delete unrelated test reports. Worth being explicit about the exact path(s) to delete.

JorgeMucientes (changes requested):

gradle/libs.versions.toml
irfano: Fixed in a3d9300f615.

WooCommerce/build.gradle
irfano: Fixed: 4a2c9bb44890d205add0475cf236fa065ddcce7a

.agents/skills/ui-review/SKILL.md
irfano: Agreed. I’ll scope cleanup to only the UiReviewTemp-generated files and the specific screenshot task report path so we don’t risk deleting unrelated screenshot tests or Gradle reports.

irfano:
Thanks for the valuable feedback. I addressed all of it, and it’s ready for another round of review.

JorgeMucientes (approved):
Looks great @irfano thanks for working n this skill and addressing the changes 👍🏼

PR #15610, opened 2026-04-03
Migrate booking analytics to EventHorizon codegen

adborbas:
### Description

Replaces 14 hand-written `AnalyticsEvent` enum entries with generated EventHorizon event types from the `eventhorizonschemas` YAML schema. A new `track(Trackable)` method on `AnalyticsTrackerWrapper` routes codegen events through the same Tracks pipeline as existing events, so ViewModels use a single tracking entry point for both old and new event types. Also adds tracking for `BookingDetailRefundTapEvent` in `issueRefund()`, which was defined in the schema but not previously tracked on Android. Two error events (`BOOKING_LIST_FAILED_TO_FETCH_BOOKINGS`, `BOOKING_LIST_FAILED_TO_UPDATE_BOOKING_DETAILS`) remain on the old pattern since they pass `Throwable` context.

Mirrors [woocommerce-ios#16883](https://github.com/woocommerce/woocommerce-ios/pull/16883).

### Test Steps

1. Open Bookings tab — verify `booking_list_view` fires with correct `selected_tab`, `is_default_tab`, `is_list_empty`, `is_filtered` properties
2. Switch tabs, tap a booking, sort, filter, search — verify corresponding events fire
3. Open booking details — update attendance, cancel booking, add note, view linked order, issue refund — verify detail events fire
4. Trigger error scenarios (e.g. offline) — verify error events still fire with `error_code`/`error_type`

### Images/gif

N/A — analytics-only change, no UI modifications.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/EventHorizonAnalyticsEvent.kt
```diff
@@ -0,0 +1,13 @@
+package com.woocommerce.android.analytics
+
+import com.automattic.eventhorizon.Trackable
+
+/**
+ * Adapts EventHorizon-generated [Trackable] events to the app's [IAnalyticsEvent] interface,
+ * so they can flow through [AnalyticsTracker] and the existing Tracks pipeline.
+ */
+class EventHorizonAnalyticsEvent(private val trackable: Trackable) : IAnalyticsEvent {
```
irfano: `val` can be removed from here since it's never used as a property.

```suggestion
class EventHorizonAnalyticsEvent(trackable: Trackable) : IAnalyticsEvent {
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -760,10 +766,13 @@ class MainActivity :
             ORDERS -> AnalyticsEvent.MAIN_TAB_ORDERS_SELECTED
             PRODUCTS -> AnalyticsEvent.MAIN_TAB_PRODUCTS_SELECTED
             POS -> AnalyticsEvent.MAIN_TAB_POS_SELECTED
-            BOOKINGS -> AnalyticsEvent.MAIN_TAB_BOOKINGS_SELECT
             MORE -> AnalyticsEvent.MAIN_TAB_HUB_MENU_SELECTED
+            BOOKINGS -> {
+                analyticsTrackerWrapper.track(MainTabBookingsSelectEvent)
```
irfano: There’s a regression here. We were sending `KEY_HORIZONTAL_SIZE_CLASS` with the `MAIN_TAB_BOOKINGS_SELECT` and `MAIN_TAB_BOOKINGS_RESELECT` events.

It looks like this parameter isn’t used on iOS. I think we should update EventHorizon and add this parameter there for Android.

irfano (changes requested):
Thank you for opening the first PR to use EventHorizon, @adborbas! I left some comments. If you haven’t already confirmed that `KEY_HORIZONTAL_SIZE_CLASS` is unnecessary and can be removed, we should update the EventHorizon parameters.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
adborbas: Once https://github.com/Automattic/EventHorizonSchemas/pull/53 is merged I'll update this PR.

irfano (approved):
This is a valuable addition to the repo. Thanks for handling this, @adborbas!

I confirmed that `horizontal_size_class` has now been added. LGTM! 👍🏻

PR #15611, opened 2026-04-03
Add feature flag override panel to Help screen for logged-out state

RafaelKayumov:
## Description

Adds a "Developer" section to the Help & Support screen (accessible during the login flow) that lets developers override feature flags **before logging in**. This avoids the painful restart cycle for flags that are read once at app launch.

Android counterpart of woocommerce/woocommerce-ios#16882.

**What changed:**
- New `LOGGED_OUT_FF_PANEL` local feature flag (enabled for debug builds)
- `DevFeatureFlagsViewModel` gains a `skipRemoteLoad` parameter via `SavedStateHandle` — when `true`, it skips `awaitRemoteFlagsLoaded()` which would hang indefinitely without authentication
- New `DevFeatureFlagsActivity` as a standalone entry point for the feature flags screen, launchable from outside the main nav graph
- `HelpActivity` gets a conditional "DEVELOPER" section with an "Override Feature Flags" row

**Behavior details:**
- The developer section **only appears in logged-out state** — when logged in, the FF panel is already accessible via Settings → Developer Options
- All feature flags are shown, but **remote value loading is skipped** in logged-out state (no authenticated network fetch) — rows show "Local: Enabled/Disabled" instead of "Remote: ..."
- Overrides are stored in SharedPreferences and checked before any remote fetch, so they work without authentication context

## Demo

https://github.com/user-attachments/assets/b6f60810-4202-4520-9ff5-ad5e6d43616a

## Test Steps

1. Run a debug build
2. On the login prologue screen, tap "Help"
3. Verify a "DEVELOPER" section appears with "Override Feature Flags" row
4. Tap it → verify all feature flags are listed with their local values
5. Toggle a flag, go back, proceed with login → verify the override persists
6. Open Help from a logged-in state → verify the "DEVELOPER" section is **not** shown
7. Verify the developer section does **not** appear in release builds

---

### DIFF
--- a/WooCommerce/src/main/res/values/strings.xml
+++ b/WooCommerce/src/main/res/values/strings.xml
@@ -2383,9 +2383,9 @@
     <string name="support_system_status_report_copied_to_clipboard">System status report copied to clipboard</string>
     <string name="support_system_status_report_error_copy_to_clipboard">Error copying SSR to clipboard</string>
 
-    <string name="support_developer_section_title">DEVELOPER</string>
-    <string name="support_override_feature_flags">Override Feature Flags</string>
-    <string name="support_override_feature_flags_detail">Toggle feature flags before logging in</string>
+    <string name="support_developer_section_title" translatable="false">DEVELOPER</string>
+    <string name="support_override_feature_flags" translatable="false">Override Feature Flags</string>
+    <string name="support_override_feature_flags_detail" translatable="false">Toggle feature flags before logging in</string>
     <string name="support_system_status_report_share_error">Unable to share System Status Report</string>
     <string name="support_system_status_report_fetch_error">Error fetching SSR. Please check WooCommerce -&gt; Status in wp-admin.</string>
     <string name="help">Help</string>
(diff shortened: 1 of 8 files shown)


WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2382,6 +2382,10 @@
     <string name="support_system_status_report_share_label">Share system status report</string>
     <string name="support_system_status_report_copied_to_clipboard">System status report copied to clipboard</string>
     <string name="support_system_status_report_error_copy_to_clipboard">Error copying SSR to clipboard</string>
+
+    <string name="support_developer_section_title">DEVELOPER</string>
+    <string name="support_override_feature_flags">Override Feature Flags</string>
+    <string name="support_override_feature_flags_detail">Toggle feature flags before logging in</string>
```
hichamboushaba: np, I would mark these with `translatable="false"`, I think translating them is not needed.

hichamboushaba (approved):
Nice work @RafaelKayumov 💯, welcome to the Android world!

I left one minor comment, but it's not a blocker.

WooCommerce/src/main/res/values/strings.xml
RafaelKayumov: Good point. I have same strings non-localizable in iOS client as well. Done in [6f7f02a](https://github.com/woocommerce/woocommerce-android/pull/15611/commits/6f7f02ab7c33eb085e2be22394708305dc163d42)

PR #15620, opened 2026-04-07
[WOOMOB-2543] Fix POS refresh catalog banner persisting after manual sync

kidinov:
### Description
Fixes WOOMOB-2543

The "Refresh catalog" banner in POS persists after performing a manual sync from settings or pull-to-refresh. The banner state was only evaluated once during ViewModel initialization and never re-checked when sync completed.

The fix reacts to `RefreshProductList` and `SettingsEvent` in `WooPosItemsViewModel` by calling `refreshBannerState()`, so the banner hides once the sync timestamp is updated.

### Test Steps
1. Open POS on a device/emulator - this triggers an initial sync
2. **Force stop** the app (Settings > Apps > WooCommerce > Force Stop)
3. **Turn off network** (airplane mode) to prevent the background sync worker from running
4. Change the device date forward by **8+ days** (Settings > System > Date & time > disable auto date)
5. Open the app and navigate to POS - the "Refresh catalog" banner should appear
6. **Turn network back on**
7. Go to POS **Settings > Product catalog** and tap **"Update catalog"**
8. Navigate back to the POS home screen
9. **Verify** the banner is now **gone**
10. Repeat steps 2-5 to show the banner again, then **pull-to-refresh** on the product list
11. **Verify** the banner disappears after the refresh completes

### Images/gif
N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsViewModel.kt
```diff
@@ -162,8 +161,10 @@ class WooPosItemsViewModel @Inject constructor(
                     is ParentToChildrenEvent.BarcodeEvent,
                     is ParentToChildrenEvent.MissingVariationEvent,
                     is ParentToChildrenEvent.RemoveProductsClicked,
-                    is ParentToChildrenEvent.ProductsRemoved,
-                    is ParentToChildrenEvent.SettingsEvent -> Unit
+                    is ParentToChildrenEvent.ProductsRemoved -> Unit
+
+                    ParentToChildrenEvent.RefreshProductList,
+                    is ParentToChildrenEvent.SettingsEvent -> refreshBannerState()
```
malinajirka: Consider matching the specific subtype `is ParentToChildrenEvent.SettingsEvent.RetrySyncRequested` instead of the parent sealed class. Today `SettingsEvent` only has `RetrySyncRequested`, but when new subtypes are added (e.g., `ThemeChanged`, `CurrencyChanged`), they would all unnecessarily trigger `refreshBannerState()`.

```suggestion
                    ParentToChildrenEvent.RefreshProductList,
                    is ParentToChildrenEvent.SettingsEvent.RetrySyncRequested -> refreshBannerState()
```

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsViewModel.kt
kidinov: Done in 1b915e585df

PR #15621, opened 2026-04-07
[WOOMOB-2566] Display sync progress indicator during initial POS sync

kidinov:
### Description
Fixes WOOMOB-2566

The initial POS catalog sync can take several minutes on large catalogs but showed no progress to the user. This PR surfaces the backend's generation progress on the splash screen.

The sync goes through three visual phases:
1. **Syncing** - spinner with "Syncing catalog" (before first poll response)
2. **Preparing** - "Preparing catalog..." (backend is queuing/scheduling generation)
3. **Progress** - "131 of 4512 items" (backend is actively generating the catalog)

Progress is piped from `WooPosFileBasedSyncAction` (where polling happens) through `WooPosLocalCatalogSyncRepository` and `WooPosProductsDataSource.prepopulateCache()` (via `channelFlow`) to the splash screen.

### Test Steps
1. Set up a store with a large product catalog (e.g. ~4500 items)
2. Clear app data or use a fresh install to trigger the initial sync
3. Launch POS
4. Observe the splash screen shows "Preparing catalog..." initially
5. After a moment, observe it transitions to "131 of 4512 items" (numbers update as sync progresses)
6. Once sync completes, POS home screen loads normally

### Images/gif

https://github.com/user-attachments/assets/67d1a8e5-1e4c-45e5-a59b-0cbffedb47c8


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFileBasedSyncAction.kt
```diff
@@ -44,6 +54,7 @@ class WooPosFileBasedSyncAction @Inject constructor(
     }
 
     suspend fun syncCatalog(site: SiteModel): WooPosFileBasedSyncResult {
+        _syncState.value = null
```
malinajirka: **Stale state on retry:** `_syncState` is reset here at the start of `syncCatalog()`, but never after completion or failure. Since `WooPosFileBasedSyncAction` lives in a `@Singleton` repository, the last progress value (e.g., `Progress(4512, 4512)`) persists in the `StateFlow` after sync finishes.

On retry, the `progressJob` in `WooPosProductsDataSource` starts collecting `syncState` _before_ `syncCatalog()` is re-invoked. It immediately receives the stale value, causing a brief flash of old progress data.

Consider resetting in a `finally` block:

```suggestion
        _syncState.value = null
        try {
```

(and add a matching `finally { _syncState.value = null }` after the existing body)

malinajirka, line 57: ~🔍  `_syncState` is reset to `null` here at the start, but it's never reset after sync completes or fails. So it retains the last value (e.g. `Progress(4500, 4512)`). On a retry, the new collector in `prepopulateCache()` would immediately receive that stale value from the `StateFlow` and briefly flash old progress numbers before this reset kicks in.~

~Probably not noticeable in most cases, but a `finally { _syncState.value = null }` wrapping the body of this function would make it clean. Wdyt?~

malinajirka (approved):
LGTM, left one nitpick.


P.S. I forgot to hit submit on this review, sorry...

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosFileBasedSyncAction.kt
malinajirka: This is actually not visible on UI even when reproduced, so not worth updating.

PR #15622, opened 2026-04-07
[WOOMOB-2588] Add bookable service product type to picker for CIAB sites

hichamboushaba:
### Description
Fixes WOOMOB-2588

CIAB sites cannot currently create a bookable service product type from the mobile app. This adds a "Service" option to the product type picker (visible only on CIAB sites) that creates an auto-draft product via REST API and opens an authenticated webview for service creation.

**How it works:**
1. User taps "Service" in the product type picker
2. A loading indicator is shown while an auto-draft bookable-service product is created via the WooCommerce REST API
3. The webview opens to the CIAB admin service editor (`/woocommerce/services/edit/{productId}`)
4. When the user dismisses the webview, the product list refreshes to show the new service

**Key implementation details:**
- `BookableServiceCreation` added to `CIABAffectedFeature` with CIAB-only gating
- `ProductDetailRepository.createAutoDraftProduct()` uses `dispatchAndAwait` to create the product via FluxC
- Auto-draft products are excluded from local caching (`WCProductStore`) to prevent them from appearing in the product list before the API refresh
- Product list refreshes on `WEBVIEW_DISMISSED` notice from `AuthenticatedWebViewFragment`

### Test Steps
1. Log in to a CIAB site
2. Navigate to Products → tap the "+" FAB
3. Verify "Service" appears in the product type picker with a gray calendar icon
4. Tap "Service" — verify a loading spinner appears
5. Verify an authenticated webview opens showing the CIAB admin service editor
6. Enter a service name (e.g., "Acme Haircut Service") and tap "Save"
7. Navigate back — verify the product list refreshes and shows the new service
8. Log in to a non-CIAB site
9. Navigate to Products → tap the "+" FAB
10. Verify "Service" does NOT appear in the product type picker

### Images/gif

https://github.com/user-attachments/assets/92b04a16-1882-40d8-88dc-d61bf3af535a


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListFragment.kt
```diff
@@ -529,6 +531,12 @@ class ProductListFragment :
     }
 
     private fun setupResultHandlers() {
+        // Refresh the product list when returning from a webview-based product creation flow
+        // (e.g., bookable service creation on CIAB sites)
+        handleNotice(AuthenticatedWebViewFragment.WEBVIEW_DISMISSED) {
```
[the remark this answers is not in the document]
hichamboushaba: Actually the documentation is outdated here, this was required when few years ago we had `TopLevelFragments` not part of the navigation graph, now, they are restaured correctly after configuration changes.
I updated the functions documentation in [ecdd5ba](https://github.com/woocommerce/woocommerce-android/pull/15622/commits/ecdd5baaf2d3c7de6bcf28ffb47c706c6f037653)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/typesbottomsheet/ProductTypesBottomSheetViewModel.kt
```diff
@@ -91,4 +128,9 @@ class ProductTypesBottomSheetViewModel @Inject constructor(
         val isVirtual: Boolean = false,
         val isVisible: Boolean = true
     ) : Parcelable
+
+    companion object {
+        @VisibleForTesting
+        const val BOOKABLE_SERVICE_EDIT_PATH = "admin.php?page=next-admin&p=/woocommerce/services/edit"
```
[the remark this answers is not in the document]
hichamboushaba: Addressed in [fe9cb37](https://github.com/woocommerce/woocommerce-android/pull/15622/commits/fe9cb37bd1502f9c87760a9987fdb8e72a4ca342)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRepository.kt
```diff
@@ -189,6 +193,27 @@ class ProductDetailRepository @Inject constructor(
...
+        val model = WCProductModel().copy(
+            type = type.value,
+            status = "auto-draft"
+        )
+        val payload = WCProductStore.AddProductPayload(
+            site = selectedSite.get(),
+            product = model
+        )
+
+        val result: OnProductCreated = dispatcher.dispatchAndAwait(
```
irfano, line 206: `dispatchAndAwait()` has no timeout here. If the create request never returns, this coroutine never resumes and the loading spinner can stay visible indefinitely. It would be safer to handle a timeout in this flow as well.

irfano (changes requested):
- I think we should also update the flow for editing a product’s type. When changing a simple product’s type to service, we should close the native product detail screen and open the web view. We can handle this in a separate PR.

https://github.com/user-attachments/assets/83564916-4ad3-4f0c-b4fb-64e35ea9ee46

- We should also update the native list after a service is trashed in the web view.

https://github.com/user-attachments/assets/8e463b50-30e1-4452-aa44-8b9acdadbbf7

hichamboushaba:

> I think we should also update the flow for editing a product’s type. When changing a simple product’s type to service, we should close the native product detail screen and open the web view. We can handle this in a separate PR.

Nice catch, I didn't think about this. Opening the WebView won't be straightforward, because we don't save the product after changing the type, we save it only when the user explicitly taps on Save. So instead of opening WebView, I decided to hide the type when the bottom sheet is opened during the edit flow (2d997f6), WDYT?

> We should also update the native list after a service is trashed in the web view.

Good catch, but I don't think this is related to the changes of this PR, let's log an issue to address it separately.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRepository.kt
hichamboushaba: The network API call has a timeout, so we'll get a response no matter what 🤔, can you explain why you think we might need a timeout here? I don't think we've never needed specific timeouts with `dispatchAndAwait`.

irfano: You're right, I was misled by the fact that this class already uses `callAndWaitUntilTimeout` and `suspendCoroutineWithTimeout` in other dispatcher-based flows, so I assumed the same concern applied here. After tracing this one end to end, I agree it does not need an extra timeout. I'll resolve this.

irfano (approved):
> I decided to hide the type when the bottom sheet is opened during the edit flow (https://github.com/woocommerce/woocommerce-android/commit/2d997f6986758c196e4413ad92ce3f5779ce7ff2), WDYT?

👍🏻 Hiding the type option is fine!

> Good catch, but I don't think this is related to the changes of this PR, let's log an issue to address it separately.

👍🏻 Opened WOOMOB-2640

Everything looks good! 🚢

PR #15623, opened 2026-04-07
[WOOMOB-2583] Hide Inbox for CIAB sites

AdamGrzybkowski:
### Description
Fixes WOOMOB-2583

Hides the Inbox feature for CIAB sites in two places:
1. **More Menu** — `MoreMenuRepository.isInboxEnabled()` now returns `false` for CIAB sites via `CIABSiteGateKeeper`
2. **Dashboard** — New `ObserveInboxWidgetStatus` returns `Hidden` for CIAB sites, preventing the Inbox card from appearing or being added

### Test Steps
1. Log in with a CIAB site
2. Open the More Menu → verify Inbox button is not shown
3. Open the Dashboard → verify the Inbox card is not visible
4. Open the Dashboard widget editor → verify the Inbox widget shows as unavailable/hidden
5. Log in with a non-CIAB site → verify Inbox is visible in both the More Menu and Dashboard as before

### Images/gif

| non-CIAB | CIAB |
|---|---|
| Screenshot_20260407_160746 | Screenshot_20260407_160715 |
| Screenshot_20260407_160740 | Screenshot_20260407_160707 |

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/moremenu/domain/MoreMenuRepositoryTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/moremenu/domain/MoreMenuRepositoryTest.kt
@@ -79,7 +79,7 @@ class MoreMenuRepositoryTest : BaseUnitTest() {
     }
 
     @Test
-    fun `given CIAB site, when checking inbox, then inbox is disabled`() = testBlocking {
+    fun `given Inbox feature is unsupported, when checking inbox, then inbox is disabled`() = testBlocking {
         // GIVEN
         whenever(selectedSite.exists()).thenReturn(true)
         whenever(ciabSiteGateKeeper.isFeatureUnsupported(CIABAffectedFeature.Inbox)).thenReturn(true)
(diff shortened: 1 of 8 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/moremenu/domain/MoreMenuRepositoryTest.kt
```diff
@@ -64,4 +77,17 @@ class MoreMenuRepositoryTest {
         // then
         assertThat(isUpgradesEnabled).isFalse
     }
+
+    @Test
+    fun `given CIAB site, when checking inbox, then inbox is disabled`() = testBlocking {
```
irfano: For this test case, the external condition is that CIABAffectedFeature.Inbox is unsupported, so I think this name would be better:

```suggestion
    fun `given Inbox feature is unsupported, when checking inbox, then inbox is disabled`() = testBlocking {
```

irfano (approved):
I just left a minor comment, but everything looks good. 👍🏻

PR #15624, opened 2026-04-07
[Migrate] Move `SiteModel` from WellSQL to Room & Remove `WellSQL`

ParaskP7:
Closes: [AINFRA-546](https://linear.app/a8c/issue/AINFRA-546)

<sup>Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com></sup>

---

FYI: The `unit-test-exemption` label was added purely because of [this](https://github.com/woocommerce/woocommerce-android/pull/15624#issuecomment-4199838543) false-positive Dangermattic ⚠️ warning: `Class OnSiteRemoved is missing tests, but unit-tests-exemption label was set to ignore this.`

---

### Open Questions for Team Review

This PR has three known limitations, ordered by impact. Feedback is welcome — particularly on the first.

> [!WARNING]
> **1. Application password users will be logged out on upgrade**
>
> Users who logged in via application passwords will be fully logged out after this upgrade. WP.com users are unaffected — they stay logged in and their sites self-heal via `FETCH_SITES`.
>
> **Tracks data (last 12 months):** Successful logins by flow show that **~99.99% of logins are WP.com** (`epilogue` flow: ~200-250 uniques/day). Non-WP.com flows registered only 4 total occurrences in the entire year: `login_password` (1), `login_store_creds` (1), `prologue` (2). A row migration would fix this but adds complexity for an effectively non-existent user segment.
>
> <details>
> <summary>Tracks Trends query used</summary>
>
> - Event: `woocommerceandroid_unified_login_step`
> - Filter by property: `step` = `success`
> - Segment data by: Property → `flow`
> - Time range: Last 12 Months
> - Stat: Uniques
> </details>
>
> **Recommendation:** Accept the logout for app-password users. The impact is negligible based on actual usage data. These users would simply re-enter their store address and credentials once after the upgrade.

> [!WARNING]
> **2. Offline upgrade shows an empty site picker**
>
> If a WP.com user upgrades while offline, they see an empty site picker with no sites. The app does not crash, but it also does not self-heal when connectivity is restored — the user must tap "Try another account" (which logs them out) to recover. Same root cause: `SiteEntity` table starts empty and needs a network fetch.
>
> **Recommendation:** Accept this behavior. Very few users will upgrade while offline and open the app before connectivity returns. Row migration would also fix this, but the edge case doesn't justify the added complexity.

> [!WARNING]
> **3. Cross-database foreign key constraints are permanently lost**
>
> Five WC entities previously had `FOREIGN KEY ... ON DELETE CASCADE` constraints to `SiteModel` in WellSQL. These are permanently lost because `SiteEntity` lives in `wp-android-database` while WC entities live in `wc-android-database`. See [`docs/site-entity-foreign-keys.md`](https://github.com/woocommerce/woocommerce-android/blob/migrate/site-model-to-room/docs/site-entity-foreign-keys.md) for the full analysis.
>
> **Recommendation:** Accept orphaned data. WC data for deleted sites is harmless (never queried without a valid site ID) and is overwritten if the site is re-added. Those 5 WC entities have already been operating without FK constraints since their respective Room migrations, and 27 other WC entities never had them either. Merging databases or moving `SiteEntity` is not worth the architectural cost.

---

### Description

This PR migrates `SiteModel` — the **last remaining WellSQL table** — to Room, then completely removes WellSQL from the project. With `SiteModel` being the most complex model (99 fields, 6-strategy insert/update matching logic, central to virtually every feature), and 126 files changed across 20 commits, this is a large PR. However, **it is not complicated if reviewed commit-by-commit**. Most of the diff is legacy cleanup (~3,259 lines deleted) and dependency removal (~2,575 lines from `WellSqlConfig.kt` alone). The key commits that need careful review are called out below.

**How to review:** Go commit-by-commit. Skim `Cleanup:`, `Test:`, and most `Refactor:` commits (they are straightforward). Focus review effort on:
- 07afccd55a1 — **The actual migration** (most important commit, needs full focus)
- 0edf122413c — Extraction of `insertOrUpdateSite` into `SiteStorePersistence`
- cbb613dde63 — Routing all site persistence through `SiteStore`
- 6a53eb4e43e — The `withTransaction` race condition fix
- 4b9e4448ae1 — WellSQL removal (large diff but straightforward deletions, also see [project dependencies changes](https://github.com/woocommerce/woocommerce-android/pull/15624#issuecomment-4199877015))
- e1852c3a762 — FK limitation documentation

---

**Cleanup Phase:**
- Remove unused site related action/store/rest/sql code (3d66dcc1544)
- Remove unused site store/sql/dao code (3fca69ade29)

**Test Phase:**
- Move site tests and test utils to fluxc module (ba04ae4cc79)
- Consolidate site test utils into TestSiteSqlUtils (21a75f595a9)
- Replace deprecated RuntimeEnvironment with ApplicationProvider (6bd8f040e20)
- Remove OrderTestUtils and inline helpers at call sites (e3c9be6fc87)
- Convert SiteStoreUnitTest from java to kotlin (a57c53bd2cf, 34ca34493bf, de0f5ff84b4)
- Remove TestSiteSqlUtils and inline site creation in tests (25a93b80da6)

**Refactor Phase:**
- Extract site persistence into SiteStorePersistence (0edf122413c)
- Route all site persistence through SiteStore (cbb613dde63)
- Convert DuplicateSiteException from object to class (5ac0afb5416)
- Make SiteStorePersistence.insertOrUpdateSite non-null (4a3c573f1f1)
- Inline AppLog.T.DB usage and suppress detekt (162c60ee02b)

**Docs Phase:**
- Add KDoc on SiteStorePersistence.insertOrUpdateSite (e838f154adc)
- Document cross-database foreign key limitation (e1852c3a762)

**Migration Phase:**
- Migrate `SiteModel` from WellSQL to Room (07afccd55a1)
    - Introduce `SiteEntity` (99 fields, auto-generated PK) with KDoc explaining why `id` instead of `siteId`
    - Introduce `SiteDao` with read/write/compound operations
    - Introduce `SiteMapper` for domain ↔ entity conversion
    - Add `SiteMapperTest`, `SiteDaoTest`, `SiteStorePersistenceTest`
    - Register entity and DAO in `WPAndroidDatabase` (v37 → v38)
    - Update `SiteStore` to use Room DAO with `runBlocking` for synchronous API and `coroutineEngine.launch()` for EventBus handlers
    - Delete `SiteSqlUtils` and `UpdateAllExceptId`
    - Remove `@Table`/`@Column`/`@PrimaryKey` WellSQL annotations from `SiteModel.java`
    - Update all affected tests across fluxc, fluxc-plugin, and WooCommerce modules

**Deps Phase:**
- Remove WellSQL from the project (4b9e4448ae1)
    - Delete `WellSqlConfig.kt` (~2,400 lines of migration history across 244 database versions), `WooWellSqlConfig.kt`, and `SingleStoreWellSqlConfigForTests.java`
    - Remove `WellSql.init()` from `WooCommerce`, `WooCommerceWear`, and `WooCommerceTest`
    - Remove wellsql deps from `build.gradle`, `libs.versions.toml`, `settings.gradle`, `dependabot.yml`, and proguard rules

**Fix Phase:**
- Wrap site insert-or-update in Room transaction (6a53eb4e43e)
    - Prevents a TOCTOU race condition when two concurrent `FETCH_SITES` responses interleave their DAO calls, causing a `UNIQUE(siteId, url)` constraint crash

---

### Self-Healing Site Data

Unlike previous WellSQL migrations where rows were "stateless" (fetched on demand, safely discarded), `SiteModel` holds the user's site list which determines login state and site selection. This raised the question of whether existing rows need to be migrated between databases.

After investigation and testing across three account types:

**WP.com account with single WOO site** — Fully self-healing. The OAuth access token is stored in `SharedPreferences` (independent from the database), so `hasAccessToken()` returns `true` and the user stays logged in. On app start, `FETCH_SITES` repopulates the Room table with fresh site data from the server. Since there's only one WOO site, auto-login selects it automatically — the user lands on the dashboard without any interaction.

**WP.com account with multiple WOO sites** — Self-healing with minor UX regression. Same mechanism as above, but since there are multiple WOO sites, the app shows the site picker for the user to re-select their site. After selection, everything works normally. Tested with 207 sites (A8C account) — all sites migrated correctly with identical data.

**Application password accounts** — NOT self-healing.

> [!WARNING]
> Users who logged in via application passwords will be **fully logged out** on upgrade. This happens because `hasAccessToken()` returns `false` (no WP.com OAuth token) and the login state depends on the selected site existing in the database. Since the `SiteEntity` table starts empty after migration, `selectedSite.connectionType` returns `null`, and `isUserLoggedIn()` returns `false`.
>
> A **row migration** (copying site data from the old WellSQL database to Room during upgrade) would fix this. This is pending team discussion — feedback welcome on whether the UX impact is acceptable or if row migration should be added before merge.

---

### Foreign Key Limitation

> [!WARNING]
> Five WC entities previously had `FOREIGN KEY(LOCAL_SITE_ID) REFERENCES SiteModel(_id) ON DELETE CASCADE` constraints in WellSQL. These constraints are **permanently lost** because `SiteEntity` lives in `wp-android-database` while all WC entities live in `wc-android-database` — SQLite FK constraints only work within a single database file.
>
> The stale FK TODO comments and `@Ignore` cascade delete tests have been removed. See [`docs/site-entity-foreign-keys.md`](docs/site-entity-foreign-keys.md) for the full analysis, including all 32 WC entities and 4 WP entities that reference site IDs, and possible solutions (programmatic cascade delete, database merge, or accepting orphaned data). Feedback welcome on the preferred approach.

---

#### Database Migration Verification
1. Build and install this update (upgrade over `trunk`)
2. Open Android Studio → App Inspection → Database Inspector
3. Verify `SiteModel` table **NO LONGER** exists in `wp-fluxc` database
4. Verify `SiteEntity` table **EXISTS** in `wp-android-database`
5. Compare column values match expected site data (siteId, url, name, hasWooCommerce, etc.)

#### WP.com Account with Single WOO Site
1. Install `trunk`, sign in with a WP.com account that has exactly one WooCommerce site
2. Build and install this branch (upgrade)
3. Verify: no crash, no site picker, dashboard loads directly with correct site
4. Navigate to Orders, Products, Settings — verify everything works

#### WP.com Account with Multiple WOO Sites
1. Install `trunk`, sign in with a WP.com account that has multiple WooCommerce sites
2. Build and install this branch (upgrade)
3. Verify: no crash, site picker shown with all WOO sites listed
4. Select a site → verify dashboard loads with correct site data
5. Navigate to Orders, Products, Settings — verify everything works

#### Application Password Account
1. Install `trunk`, sign in via "Enter your store address" with a self-hosted site
2. Build and install this branch (upgrade)
3. Verify: user is redirected to login screen (expected — not self-healing)
4. Sign in again → verify everything works normally after re-login

#### Offline Upgrade Path

> [!WARNING]
> Upgrading while offline shows an empty site picker with no sites to choose from. The app does not crash, but it also does **not self-heal when connectivity is restored** — the user is stuck on the empty picker and must tap "Try another account" (which logs them out) to recover. This is another argument in favor of row migration, which would preserve site data locally regardless of network state.

1. Install `trunk`, sign in with a WP.com account
2. Enable airplane mode
3. Build and install this branch (upgrade)
4. Verify: no crash, but site picker shows 0 sites
5. Disable airplane mode → verify sites do **not** auto-refresh (user must tap "Try another account" to recover)

#### Fresh Install
1. Uninstall the app completely
2. Build and install this branch
3. Sign in with a WP.com account
4. Verify dashboard loads and all features work normally

---

### Images/gif

image

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>OnSiteRemoved</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainPresenter.kt
```diff
@@ -75,7 +75,7 @@ class MainPresenter @Inject constructor(
         dispatcher.dispatch(AccountActionBuilder.newUpdateAccessTokenAction(UpdateTokenPayload(token)))
     }
 
-    override fun hasMultipleStores() = wooCommerceStore.getWooCommerceSites().size > 0
+    override suspend fun hasMultipleStores() = wooCommerceStore.getWooCommerceSites().size > 0
```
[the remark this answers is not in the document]
ParaskP7: Good catch on the inconsistency, but this is pre-existing — our change only added `suspend` to the signature. The `> 0` vs `> 1` discrepancy exists on `trunk` today.

That said, `hasMultipleStores()` is actually dead code — it has no callers anywhere in the codebase. Both the method and the inconsistency could be removed entirely, but that's a separate cleanup outside the scope of this migration PR.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteStorePersistence.kt
```diff
@@ -0,0 +1,147 @@
...
+                    }
+                }
+            }
+
+            // If the site is a self hosted, maybe it's already in the DB as a Jetpack site, and we don't want to create
+            // a duplicate.
+            if (existingSite == null) {
+                val forcedHttpXmlRpcUrl = "http://" + UrlUtils.removeScheme(site.xmlRpcUrl)
+                val forcedHttpsXmlRpcUrl = "https://" + UrlUtils.removeScheme(site.xmlRpcUrl)
+                val siteResult = siteDao.getByXmlRpcUrl(forcedHttpXmlRpcUrl, forcedHttpsXmlRpcUrl).firstOrNull()
```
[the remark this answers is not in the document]
ParaskP7: This logic is a direct port from the original `SiteSqlUtils.insertOrUpdateSite()` which had the same null handling characteristics. The WellSQL version used these values identically without null guards, and it never caused issues in production.

Adding defensive null normalization would be a behavior change beyond the scope of this migration, which aims to preserve the existing logic as-is. If we want to harden this, it should be a separate follow-up.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/dao/SiteDao.kt
```diff
@@ -0,0 +1,133 @@
...
+     * @param origin the origin value to filter by (e.g., [SiteModel.ORIGIN_WPCOM_REST])
+     * @param siteIds the remote site IDs to keep
+     */
+    @Transaction
+    open suspend fun deleteByOriginNotInList(origin: Int, siteIds: List<Long>) {
+        val localSites = getByOrigin(origin)
+        val toDelete = localSites.filter { it.siteId !in siteIds }
+        for (site in toDelete) {
+            deleteByLocalId(site.id)
+        }
```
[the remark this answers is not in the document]
ParaskP7: Fair point about `List` vs `Set` for the `!in` check. However, the site count is small in practice — even the largest account we tested had 207 sites. The O(n*m) cost is negligible at this scale.

The previous WellSQL implementation (`removeWPComRestSitesAbsentFromList`) used the same pattern — iterate local sites, filter against the server list, delete individually. This migration preserves that behavior.

A single SQL `DELETE ... WHERE siteId NOT IN (...)` would be cleaner, but SQLite has a variable limit (`SQLITE_MAX_VARIABLE_NUMBER = 999`) that would require chunking for large lists. Converting to a `Set` is a reasonable quick win though — happy to add that if the reviewer feels strongly about it.

libs/fluxc/src/test/java/org/wordpress/android/fluxc/persistence/dao/SiteDaoTest.kt
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    @Test
+    fun `given site inserted, when get by local id, then returns site`() = runTest {
+        val sideId = 100L
+        val url = "https://example.com"
+        val entity = createSiteEntity(siteId = sideId, url = url)
+        dao.insert(entity)
+        val inserted = dao.getByRemoteId(sideId).first()
+
```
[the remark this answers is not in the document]
ParaskP7: Good catch — fixed the `sideId` → `siteId` typo (appeared in two tests). Thanks! See 11aa9e72194def9828dff4d19182659a12a3dfb6.

ParaskP7:
FYI: In addition to @samiuelson which was auto-added to this requested review via `@woocommerce/android-developers` as a code owner, I am manually adding a second reviewer on this, @malinajirka , just because of this PR's (possible) impact.

PS: Maybe it is even worth adding more reviewers, or at least mention `@woocommerce/android-developers` as a whole, just so that everyone can at least skim this PR and contribute on the 3 x ⚠️ warnings discussion mentioned on-top. But, I'll leave that last part for @samiuelson and @malinajirka to decide. 🙏

THANK YOU ALL

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/CookieNonceAuthenticator.kt
```diff
@@ -55,7 +55,7 @@ class CookieNonceAuthenticator @Inject constructor(
         val usingSavedRestUrl = site.wpApiRestUrl != null
         if (!usingSavedRestUrl) {
             site.wpApiRestUrl = discoverApiEndpoint(site.url)
-            (siteSqlUtils::insertOrUpdateSite)(site)
+            siteStore.insertOrUpdateSite(site)
```
malinajirka: 💡 `siteStore.insertOrUpdateSite(site)` is declared `@Throws(DuplicateSiteException::class)` - should we catch it here to ensure the request isn't aborted?

Same applies to the second call on line 74.

```suggestion
            try {
                siteStore.insertOrUpdateSite(site)
            } catch (e: SiteStorePersistence.DuplicateSiteException) {
                AppLog.w(AppLog.T.API, "Duplicate site detected while saving wpApiRestUrl", e)
            }
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -284,7 +280,7 @@ open class WooCommerceStore @Inject internal constructor(
                     // Persist the Application Passwords auhtorization URL
                     site.applicationPasswordsAuthorizeUrl = response.result.authentication
                         ?.applicationPasswords?.endpoints?.authorization
-                    siteSqlUtils.insertOrUpdateSite(site)
+                    siteStore.insertOrUpdateSite(site)
```
malinajirka: 💡 Same issue as `CookieNonceAuthenticator` — `insertOrUpdateSite` can throw `DuplicateSiteException` and this call has no try-catch. This is a side-effect persistence call (saving the application passwords authorization URL), so a duplicate site shouldn't crash the API version check. Wrapping in try-catch with a log would suffice. Wdyt?

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/SiteStoreUnitTest.kt
```diff
@@ -0,0 +1,664 @@
...
+    @Throws(DuplicateSiteException::class)
+    fun testSearchSitesByNameMatching() = test {
+        val wpComSite1 = SiteUtils.generateWPComSite()
+        wpComSite1.name = "Doctor Emmet Brown Homepage"
+        val wpComSite2 = SiteUtils.generateWPComSite()
+        wpComSite2.name = "Shield Eyes from light"
+        wpComSite2.siteId = 557
+        val wpComSite3 = SiteUtils.generateWPComSite()
+        wpComSite3.name = "I remember when this was all farmland as far as the eye could see"
+        wpComSite2.siteId = 558
```
malinajirka: 🔍 np: This looks like a pre-existing bug carried from the Java version — line 328 sets `wpComSite2.siteId = 558` but should be `wpComSite3.siteId = 558`. As written, `wpComSite3` keeps `siteId = 556` (same as `wpComSite1`) and shares the same unique key `(556, "")`, so inserting site3 overwrites site1 rather than creating a 3rd row. The test still passes because the overwritten name also contains "eye", but it's testing 2 sites instead of the intended 3.

```suggestion
        wpComSite3.siteId = 558
```

malinajirka:
Thanks so much for working on this @ParaskP7 !

> 1. Application password users will be logged out on upgrade
Users who logged in via "Enter your store address" (application passwords) will be fully logged out after this upgrade. WP.com users are unaffected — they stay logged in and their sites self-heal via FETCH_SITES.
Tracks data (last 12 months): Successful logins by flow show that ~99.99% of logins are WP.com (epilogue flow: ~200-250 uniques/day). Non-WP.com flows registered only 4 total occurrences in the entire year: login_password (1), login_store_creds (1), prologue (2). A row migration would fix this but adds complexity for an effectively non-existent user segment.
Tracks Trends query used
Recommendation: Accept the logout for app-password users. The impact is negligible based on actual usage data. These users would simply re-enter their store address and credentials once after the upgrade.

I'm having hard time believing this - we encounter app-pwd passwords users in support relatively often. The very first sentences is off => `Enter your store address` is the only login flow we have, that's not where we differentiate between app pwd vs .com login.

This could be a blocker for the migration as is tbh, I believe we can't afford logging out all app pwd users -> the support queue would likely be crazy and we'd likely lose quite a few users. @JorgeMucientes what is the best way of checking which login users use? Do you think logging out all app pwd users is something worth considering or a complete no-go?

ParaskP7:
👋 @malinajirka and thank you a lot for taking a first look at that! 🙇 

Before answering your concerns let me preface this by saying that:

1. I wanted to raise a PR with as little overload (row migration) as possible, just to get us start with this discussion, and, so you know what to expect of this last migration (the overall and minimum of code changes).
2. Everything is doable, I can think of multiple ways to take this forward, this is just the base, and if need be, we can take this base and create another PR, redoing things (to incorporate row migration) and close this one. We could also continue working on this PR to (somehow force) self-heal on application passwords users (needs investigation).
3. I've only identified 2 main flows so far, with/without application passwords, plus 2 extra flows for WP.com accounts, with single/multiple Woo sites. I am sure there are more, let's discuss those extra flows I missed so we could cover/test as many cases as possible before committing to a solution.

> I'm having hard time believing this - we encounter app-pwd passwords users in support relatively often.

This `Tracks Trends` query I used was the best I could think of, and probably, from what you say, it is the wrong query, or at leat a misleading one. Do you have any other query in mind? 🙏 (I see you asked @JorgeMucientes thanks)

> The very first sentences is off => Enter your store address is the only login flow we have, that's not where we differentiate between app pwd vs .com login.

True, you're right and I've removed this, moving `application passwords` out of parenthesis, making this sentence read as: `Users who logged in via application passwords will be fully logged out after this upgrade.` 🙏

> This could be a blocker for the migration as is tbh...

No worries of it being a blocker, we can adapt as need be, redo it all, this was just the quickest way to migrate this table from WellSQL into Room, just like I did for the `Account` table (#15536) for which I also thought it might need a row migration (but in the end it didn't).

> ...I believe we can't afford logging out all app pwd users -> the support queue would likely be crazy and we'd likely lose quite a few users.

This is true, especially if we have hundreds/thousands of those... 💯 (not sure of the numbers, plus, the ration to WP.com users)

> @JorgeMucientes what is the best way of checking which login users use? Do you think logging out all app pwd users is something worth considering or a complete no-go?

🙏 x 🙏 ^ 🙏 (great questions)

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteMapper.kt
```diff
@@ -0,0 +1,213 @@
+package org.wordpress.android.fluxc.persistence
+
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.persistence.entity.SiteEntity
+import javax.inject.Inject
+
+class SiteMapper @Inject constructor() {
+    @Suppress("LongMethod")
+    fun toEntity(model: SiteModel) = SiteEntity(
```
malinajirka, line 9: ❓ model.toEntity().toModel() returns a different  model, since null fields are changed to empty strings. I don't see any issue with it, just wanted to raise it since it feels like a potential code-smell. Can you foresee any issues?

malinajirka (commented):
Thanks so much for working on this @ParaskP7! Overall, the migration looks solid, however, there are a few points I believe are worth discussing.

1. Shared Preferences - We use local site id in many different preferences as part of the key in both store management and POS. We even store `AppPrefs.SELECTED_SITE_LOCAL_ID` directly. Unless I'm missing something, we might need to handle migration of these manually. Unless the plan is to intentionally ignore them and let the app handle everything as if it user just logged in (or we could purge them) - if that's the case though, we need to ensure the same site ID won't be generated in room => as we might end up assigning values to different sites which would likely have negative consequences.

2. When it comes to the Foreign Key => I'm wondering if it still makes sense to keep two separate databases. The original reasoning (AFAICT) was that WP app didn't need Woo tables, but now this is gone, since each app uses their own "copy" of FluxC. I understand it might be above scope of this PR, but I believe it's worth considering. Wdyt?

3. Data Migration - last but not least, as I mentioned above, I believe we need to double-check the app password usage. I created a funnel and it seems roughly 40% of logins in the last 30 days are Application password logins. It's also worth mentioning that AFAIU we use Application Password even for Jetpack users now, so we can communicate directly with the site without going through Jetpack Proxy - having said that, it's quite likely the recovery will work fine for this particular case, but worth double checking.

malinajirka:
> Recommendation: Accept orphaned data. WC data for deleted sites is harmless (never queried without a valid site ID) and is overwritten if the site is re-added.

Correct me if I'm wrong, but I believe the same issue I mentioned with the Shared Preferences key might happen here 🤔 => Room might generate an ID, let's say 10, for site A, but in WellSQL 10 was used for site B => we might end up associating data of site B to site A. Wdyt?

hichamboushaba:
> 1. Application password users will be logged out on upgrade
> Users who logged in via application passwords will be fully logged out after this upgrade. WP.com users are unaffected — they stay logged in and their sites self-heal via FETCH_SITES.
> Tracks data (last 12 months): Successful logins by flow show that ~99.99% of logins are WP.com (epilogue flow: ~200-250 uniques/day). Non-WP.com flows registered only 4 total occurrences in the entire year: login_password (1), login_store_creds (1), prologue (2). A row migration would fix this but adds complexity for an effectively non-existent user segment.
> Tracks Trends query used
Recommendation: Accept the logout for app-password users. The impact is negligible based on actual usage data. These users would simply re-enter their store address and credentials once after the upgrade.

> Shared Preferences - We use local site id in many different preferences as part of the key in both store management and POS. We even store AppPrefs.SELECTED_SITE_LOCAL_ID directly. Unless I'm missing something, we might need to handle migration of these manually. Unless the plan is to intentionally ignore them and let the app handle everything as if it user just logged in (or we could purge them) - if that's the case though, we need to ensure the same site ID won't be generated in room => as we might end up assigning values to different sites which would likely have negative consequences.

Thanks @ParaskP7 and @malinajirka for the ping, I didn't review the PR yet, but just looking at these two sentences above, and this matches the issue that I mentioned before (p1747219327074329/1747214664.229459-slack-C6H8C3G23).
I think this issue is a blocker, as Jirka mentioned, around 40% of users use app passwords for login, but even for other users, losing the selected site or ending up on a different site is a big impact too.

What's blocking from working on a real migration that preserves the IDs? I think there are few ways to handle this:
- Use `sql` directly to update the Room DB (I mean inserting the sites) when handling the `WellSql` migration.
- Defer the WellSql delete to a later version, and for now just read the data from WellSql when handling the Room migration.

WDYT?

ParaskP7:
Thanks @malinajirka and @hichamboushaba for the thorough feedback, much appreciated — you're both right, and this has shifted my thinking significantly. ❤️ 🙇 🙏 Let me respond to each point:

## 1. App password usage is much higher than I thought

@malinajirka, your funnel shows ~40% of logins are app passwords — that's a massive discrepancy from the ~0.01% I measured. The gap is likely because (as you noted) we may now use Application Passwords for Jetpack users too, not just self-hosted — so my query on `login_password` / `login_store_creds` steps would have undercounted the true app-password login volume. That changes the calculus entirely.

Combined with the fact about the **shared preferences + local site ID coupling** (`SELECTED_SITE_LOCAL_ID` and other prefs keyed by local site ID), accepting data loss is no longer acceptable — even for WP.com users, a changed local ID would cause preferences to attach to the wrong site.

**Conclusion: row migration is a must.** Accepting this as a blocker. ❌

Request (🙏): Could you (or @hichamboushaba) share the funnel link/config if possible? Purely for my own education — I'd like to see how the measurement was set up and replicate it on my end.      

## 2. Foreign keys and two databases

@malinajirka, totally agree the original "WP app doesn't need Woo tables" reasoning no longer applies — each app has its own FluxC copy now. Rather than doing a full merge of both databases at once, I'd propose a more incremental approach:

1. **In this PR**: migrate `SiteModel` into `wc-android-database` (instead of `wp-android-database`). This makes it possible to restore the FK constraints for the 5 WC entities that previously had them (`WCOrderSummaryModel`, `WCProductTagModel`, `WCProductShippingClassModel`, `WCProductCategoryModel`, `WCProductReviewModel`).
2. **Follow-up PRs**: migrate the remaining 15 `wp-android-database` tables into `wc-android-database` over time, one by one, until we're down to a single database.

This avoids a risky one-shot merge while still moving us toward the "single database" end state. WDYT? 🤔

## 3. Row migration approach

@hichamboushaba, thanks for the concrete suggestions. Between the two:

- **Option A (inline SQL during Room migration):** Read from WellSQL → insert into Room in the same migration step. Pros: single-step, clean cutover. Cons: requires accessing the WellSQL DB file directly from Room's migration callback, which is awkward.
- **Option B (defer WellSql delete):** Keep WellSQL readable for one version, do the copy at app start, drop in a follow-up. Pros: safer, reversible. Cons: two-step release, we keep WellSQL around longer.

**My plan:** go with **Option B** — it's lower-risk and lets us validate the row migration in production before removing WellSQL. Before closing this PR, I'll first address all the remaining code-related comments from @malinajirka's review so the follow-up PR starts from a cleaner base without duplicate noise (see ee179bb65bc96ed60ac92921a23b3b272fb6a25d + 9bf6bf8b2f0c611ca70d76874763c4ac3c861f8b, plus 3f4fa636303609145176d57ad0fd560cef39f517). Then:

1. Close this PR and open a new one, keeping all commits **except** 4b9e4448ae109e3946c35d2d0e28754c787a3f43 (Remove WellSQL) and e1852c3a7622d20fb5d98d82e30a20a8a6dd64b7 (Document FK).
2. Keep the `SiteModel` WellSQL annotations as part of 07afccd55a1ee042e2d8e80839594ded7aa62acf (Migration), so we can still read from `wp-fluxc`. Also update that commit to target `wc-android-database` instead of `wp-android-database`, per point 2 above.
3. Re-add the ForeignKey constraints on the 5 WC entities (`WCOrderSummaryModel`, `WCProductTagModel`, `WCProductShippingClassModel`, `WCProductCategoryModel`, `WCProductReviewModel`) now that `SiteEntity` lives in the same database, and re-enable the 4 cascade delete tests that were removed. 
4. Add a one-time row migration step that copies all `SiteModel` rows from WellSQL into `SiteEntity`, preserving the local `id` so `SharedPreferences` keys remain valid.
5. Ship the new PR with Room as the source of truth, WellSQL still present but otherwise unused.
6. Follow-up PR (next version) to remove WellSQL entirely.

How does that all sound to you both? Would appreciate your feedback on the overall approach before I kick off the follow-up work. 🙏

---

FYI: One heads-up on timing: I'm starting **parental leave on Monday**, so I won't personally have time to fully work on this new PR. Most likely @wzieba will pick this up on his return from sabbatical (`May 4th`). Otherwise, I'll take it back when I return and finish the work I started (`May 20th`). ⏳

libs/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteMapper.kt
ParaskP7: Good observation @malinajirka — the round-trip `model.toEntity().toModel()` does lose the null/empty distinction because `SiteEntity` uses non-null `String` fields (following the `AccountEntity` pattern) and the mapper uses `.orEmpty()` to bridge the gap.

I don't foresee functional issues:

1. The Java `SiteModel` getters are not `@Nullable`-annotated (e.g. `public String getUrl()`, `public String getName()`, `public String getXmlRpcUrl()`), and setters are `@NonNull`-annotated (e.g. `public void setUrl(@NonNull String url)`). From Kotlin's perspective these are platform types (`String!`), which callers almost universally treat as non-null. A quick grep across the codebase shows `site.url`, `site.name`, `site.iconUrl`, etc. used without null checks in most call sites.
2. In practice, after a successful `FETCH_SITES` most fields come back populated or treated as empty, so the null→empty conversion only rarely affects real data. That said, I haven't exhaustively verified every field against the server's JSON schema, so there may be edge cases where a genuine `null` from the server becomes an empty string after round-trip. If that matters for a specific field, happy to revisit.
3. WellSQL behaved similarly in practice: `@Column` fields default to `null` for strings but the app code rarely relied on that distinction.

If there's a specific scenario where null-vs-empty matters that I'm missing, happy to revisit — otherwise I'd treat this as an acceptable normalization. Wdyt? 🤔

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/CookieNonceAuthenticator.kt
ParaskP7: Nice catch @malinajirka , thanks and fixed: ee179bb65bc96ed60ac92921a23b3b272fb6a25d

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
ParaskP7: Nice catch @malinajirka , thanks and fixed: ee179bb65bc96ed60ac92921a23b3b272fb6a25d

libs/fluxc/src/test/java/org/wordpress/android/fluxc/store/SiteStoreUnitTest.kt
ParaskP7: What an eye @malinajirka , test typo fixed: 9bf6bf8b2f0c611ca70d76874763c4ac3c861f8b

malinajirka:
Thanks @ParaskP7 ! I haven't gone through all your comments and it's too late to think everything through. However, before you start on options B, I wanted to share my train of thoughts:

> Option B (defer WellSql delete): Keep WellSQL readable for one version, do the copy at app start, drop in a follow-up. Pros: safer, reversible. Cons: two-step release, we keep WellSQL around longer.

If we merge into lets say 24.7 and we remove WellSQL in 24.9 all users who are on 24.6- and update directly to 24.9 would not migrate properly, right? Or am I missing something? 

I'm thinking that we can remove WellSQL right away, but keep the database files untouched => then we can migrate to room using SQLite direct access instead of going through WellSQL => that should allow us to migrate no matter how many versions users skip and allows us to remove WellSql right away. Wdyt?

ParaskP7:
Great point @malinajirka — and I realize now this is what @hichamboushaba was actually suggesting with his "use `sql` directly" option. I misread it earlier as a Room-migration-callback-while-WellSQL-is-still-loaded pattern, but he clearly meant raw SQLite access to the `wp-fluxc` file, which is exactly what you're describing.

You're absolutely right about the skip-version gap: if we migrate in 24.7 and remove WellSQL in 24.9, any user going from 24.6 → 24.9 directly (e.g. via Play Store bulk update or coming back from an uninstall with cached install data) would skip the WellSQL read step entirely and land with an empty `SiteEntity` table. That defeats the whole purpose of doing the row migration in the first place.

So the approach is:

- Keep the WellSQL **database file** (`wp-fluxc`) on disk untouched
- Remove the WellSQL **library dependency** immediately
- In the Room migration, open `wp-fluxc` directly via `SupportSQLiteDatabase` / `SQLiteDatabase.openDatabase(...)`, read the `SiteModel` rows with raw SQL, and insert them into `SiteEntity`
- Leave the `wp-fluxc` file on disk — cleanup can happen in a future follow-up release

This is version-agnostic — as long as the old `wp-fluxc` file exists on disk, the migration runs correctly regardless of which version the user is coming from. And we get to drop the WellSQL dependency in the same PR, no two-step release needed.

Updated plan:

1. Close this PR and open a new one, keeping all commits **except** 4b9e4448ae109e3946c35d2d0e28754c787a3f43 (Remove WellSQL) and e1852c3a7622d20fb5d98d82e30a20a8a6dd64b7 (Document FK)
2. Update the migration commit (07afccd55a1ee042e2d8e80839594ded7aa62acf) to target `wc-android-database` instead of `wp-android-database`, and remove the `SiteModel` WellSQL annotations (no longer needed since we won't read via WellSQL)
3. Re-add the `ForeignKey` constraints on the 5 WC entities and re-enable the 4 cascade delete tests
4. Add a one-time row migration step that opens `wp-fluxc` directly via raw SQLite, reads all `SiteModel` rows, and inserts them into `SiteEntity` preserving the local `id`. Leave the `wp-fluxc` file on disk — cleanup of the now-unused database file can happen in a future follow-up release once we've verified the migration runs cleanly in production.
5. Re-add the WellSQL library removal commit (now safe since we no longer depend on WellSQL runtime, only on the raw `wp-fluxc` file during migration)

WDYT? If this matches what you had in mind, I'll leave the instructions here for @wzieba / future me to pick up.

malinajirka:
Thanks for the clear summary! 

> So the approach is:
Keep the WellSQL database file (wp-fluxc) on disk untouched
Remove the WellSQL library dependency immediately
In the Room migration, open wp-fluxc directly via SupportSQLiteDatabase / SQLiteDatabase.openDatabase(...), read the SiteModel rows with raw SQL, and insert them into SiteEntity
Leave the wp-fluxc file on disk — cleanup can happen in a future follow-up release

Yeah, that exactly matches what I was thinking.

> Close this PR and open a new one

Fine with me, or we can revert the commits.

> Add a one-time row migration step that opens wp-fluxc directly via raw SQLite, reads all SiteModel rows, and inserts them into SiteEntity preserving the local id. Leave the wp-fluxc file on disk — cleanup of the now-unused database file can happen in a future follow-up release once we've verified the migration runs cleanly in production.

Yeah, one thing we'll need to keep in mind is that if we change SiteModel table in Room  in the future(which is quite rare), we might need to update the Sql2Room migration script as well (depends on the type of change). I'm not sure how much we want to worry about this, but I thought it's better to bring it up for completeness. E.g. if we rename `id` to `super_id`, the migration script needs to be updated as well, otherwise again users on 24.6- who directly go to x.y version wouldn't be able to migrate.

ParaskP7:
Thanks @malinajirka and it's a deal then! 🤝

> Yeah, that exactly matches what I was thinking.

👍

> Fine with me, or we can revert the commits.

👍 

> Yeah, one thing we'll need to keep in mind is that if we change SiteModel table in Room in the future(which is quite rare), we might need to update the Sql2Room migration script as well (depends on the type of change). I'm not sure how much we want to worry about this, but I thought it's better to bring it up for completeness. E.g. if we rename id to super_id, the migration script needs to be updated as well, otherwise again users on 24.6- who directly go to x.y version wouldn't be able to migrate.

Good edge case to think about, yes, also, I agree, I don't think we should worry too much on it, I for sure don't expect that to happen. Having said that, you know best and I see this "table" being touched a couple of times each year for (at least) the past 2 years (mainly new fields/columns being added), so can't be too sure. We could add a BIG warning comment on top of this `SiteModel` class, that could help the author (and through him AI), to be cautious about this. We could also add DOCS on it, within `AGENTS.md`, to help AI bubble that up while new such code is being written, or reviewed later. I am sure there can be other ideas as well... 💭

No matter, this is a delicate migration, and we can't keep this (now) single table `wp-fluxc` database file forever. At some point we'll have to get rid of this "temporal" file and live with the fact that a few "rusty" users will end-up being logged-out on an update, say X versions later, only because of our need to now update `SiteModel` and its `SiteEntity`. I am not sure we can avoid that indefinitely, can we? 🤔

hichamboushaba:
Thank you both, nice discussion.

> Request (🙏): Could you (or @hichamboushaba) share the funnel link/config if possible? Purely for my own education — I'd like to see how the measurement was set up and replicate it on my end.

We have two ways here:
- Either using Tracks trends, check for the event `woocommerceandroid_signed_in` and segment by `dotcom_user`, this will even show you that app passwords are now used more than WordPress.com (this was surprising to me).
- To have a more accurate result, you need to fetch the actual usage of the app users, not the login usage, you can do so by checking Superset directly using a query like this 3af7d-pb/

-----

**Note:** Regarding this:
> Non-WP.com flows registered only 4 total occurrences in the entire year: login_password (1), login_store_creds (1), prologue (2)

For me, even when using `woocommerceandroid_unified_login_step`, I see high usage of `login_store_creds`, I'm not sure what caused it to show you only 4 occurences.

ParaskP7:
Thanks for sharing all that @hichamboushaba and yea, it seems evident that I am a bit rusty on the `Tracks Trends` and `Tracks Funnel` side of things. Yours and @malinajirka's help was really appreciated so that we can drive this into the right conclusion, which is now agreed 🤝 that we need to go through the data migration flow for `SiteModel`. 💯

FYI: As per all the discussions that happen in this PR, I'll now proceed to close this. @wzieba and/or me will open another such PR later, with data migration this time. We'll ping you both on it again when the time comes. Thank YOU again, BOTH! 🙇

PR #15625, opened 2026-04-07
Split and improve AI agent skills for POS and Store

malinajirka:
### Description
Restructures the AI agent skills to properly separate POS and Store Management patterns, and adds a high-level architecture overview to `AGENTS.md`.

Previously, some skills (like `analytics`) mixed POS and Store patterns, leading to incorrect guidance. The POS architecture was also not documented at a high level — only individual patterns were covered.

**New skills:**
- `pos` — POS architecture, ViewModel, event bus, Compose UI, design system
- `pos-analytics` — `WooPosAnalyticsEvent` sealed class, `WooPosAnalyticsTracker`, `WooPosAnalyticsEventConstant`
- `pos-tests` — `WooPosCoroutineTestRule`, `runTest`, event bus mocking, analytics verification
- `store-analytics` — `AnalyticsEvent` enum, `AnalyticsTrackerWrapper`, `KEY_`/`VALUE_` constants
- `store-compose` — Fragment hosting, `composeView` helper, WooTheme, preview annotations
- `store-viewmodel` — `ScopedViewModel`, Hilt, StateFlow, `triggerEvent`, navArgs
- `store-tests` — `BaseUnitTest`, `testBlocking`, `captureValues`, AssertJ

**Updated `AGENTS.md`:**
- Added architecture overview explaining the two-app structure (Store Management vs POS)
- Each section describes the stack, code location, and which skills to use
- Removed implementation details that are now covered by dedicated skills (ScopedViewModel pattern, Compose UI patterns, testing patterns, analytics, `@docs/` references)
- Common Pitfalls trimmed to universal rules; app-specific pitfalls moved to skills

**Other updates:**
- `.claude/settings.json` — added new skills to allowed tools, removed stale entries
- `pr/SKILL.md` — minor updates

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          5 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosMyFeatureViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooPosMyFeatureViewModelTest</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>MyFeatureViewModelTest</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>ExampleViewModelTest</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

AGENTS.md
```diff
@@ -253,20 +167,6 @@ fun `given user is logged in, when refresh is pulled, then data reloads`() = tes
...
-
-| Skill | Triggers |
-|-------|----------|
-| `review` | "review changes", "check my code", "quality check" |
-| `pr` | "create PR", "open PR", "pull request" |
-| `verify-on-device` | "verify on emulator", "test on device", "take screenshot" |
-
 ## Further Reading
 
-@docs/compose.md
```
kidinov, line 269: What I liked with the linking of the docs is that we have 1 source of truth for humans and for the AIs on codings styles, approaches and so on

.claude/settings.json
```diff
@@ -21,7 +21,19 @@
       "Bash(gh issue *)",
       "Bash(gh api *)",
       "Bash(adb *)",
-      "Bash(emulator *)"
+      "Bash(emulator *)",
+      "Skill(write-tests)",
```
kidinov: Skill(write-tests), Skill(write-pos), Skill(write-compose), Skill(write-viewmodel) - but the actual skill names are store-tests, pos, store-compose, store-viewmodel. These allowlist entries don't match.

AGENTS.md
malinajirka: The main problem with linking is that it always adds all the docs to the context -> I don't mind having links per se, but we shouldn't link all docs from agents.md which is load in each instance.

It's also worth noting is that Claude needs in the context only things which are not common knowledge/best practice => e.g. compose best practices recommended by google don't need to be in the context, since the model was trained on them. Adding such things pollutes the context and reduces quality of the output.

kidinov: That's correct but all the linked docs are our project-specific docs, not google specific right? And largely, they duplicate of what has been added in this PR

.agents/skills/pos/SKILL.md
```diff
@@ -0,0 +1,235 @@
+---
```
kidinov, line 1: I think skills like pos, store-compose, and store-viewmodel don't fit the skill model well. Skills are meant to be task-oriented — step-by-step guides for a specific workflow, like "add an analytics event" or "write a test." These three are more like architecture reference docs that describe how things work.      
                                                                                                                                                                                                                                                                                                                             
The problem is practical: an agent asked to "fix a bug in WooPosItemsViewModel" has no reason to load the pos skill before it starts working. But without it, AGENTS.md no longer has enough detail to prevent the agent from using store patterns like ScopedViewModel or triggerEvent().                                 
                                                                                                                                                                                                                                                                                                                             
Reference/architecture content should stay in AGENTS.md (or linked docs) since it's always loaded. Skills should be reserved for things like pos-analytics, pos-tests, store-analytics, store-tests, pr — actual workflows with concrete steps.

wdyt?

AGENTS.md
malinajirka: We could link the docs from each of the relevant skills. Before we invest time into it, is it truly worth it? Looking at the existing @docs, they seem quite outdated and my personal bet is not many people read them. Perhaps, we could consider simply removing them and having the skills as the source of truth? Alternatively, we could consider asking claude to extract the knowledge into the @docs files -> and reference them from the skills. That way, it'd be up-to-date. My main point is that linking the existing docs feels a bit worthless since they don't even match reality/code anymore. Wdyt?

.agents/skills/pos/SKILL.md
malinajirka: I believe what you are describing are commands, at least in my understanding. I mean, skills kind of replaced them, but I don't think it was 1:1 one replacement => skills are imo equivalent of human skill => it can be a skill to prepare a coffee (step2step instructions) as well as a skill to navigate in POS codebase (knowledge/experience skill).

> The problem is practical: an agent asked to "fix a bug in WooPosItemsViewModel" has no reason to load the pos skill before it starts working. 

This is a flaw in the description of the skill, not in the concept IMO. We should mention in the description that the skill should be used not just when "Use when writing, editing, exploring, debugging, or reviewing WooPos-prefixed classes or files under ui/woopos/." but also when fixing bugs, implementing features, etc. . 

Wdyt?

kidinov: That's a fair point, and I'm not 100% sure about this either - just thinking out loud.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                             
I agree skills don't have to be strictly step-by-step commands. The coffee analogy makes sense. And yeah, improving the description would help with triggering.                                                                                                                                                            
                                              
My worry is more about the failure mode. If a workflow skill like pr doesn't load, the agent just creates a PR - easy to fix. But if the pos skill doesn't load and the agent starts writing code with ScopedViewModel or triggerEvent() in POS, that's wrong code generated and more likely that will have bigger negative impact. Also harder to notice
                                                                                              
Even with a perfect description, there are cases where the agent might not think to load it - like a simple rename or a one-line fix in a WooPos file. It just won't feel "POS-related enough" to trigger.                                                                                                                 
                                                                                                                                                                                                                                                                                                                             
So I was thinking maybe a middle ground - keep the key architectural constraints (what NOT to use, what to use instead) in AGENTS.md so they're always loaded as guard rails, and then have task-oriented POS skills (pos-analytics, pos-tests, etc.) for specific workflows. That way the basics are always enforced and the detailed guidance loads when needed.                                                                                                                                                                                                                                                                                   
                                                                                              
But again, not fully sure - maybe better descriptions would be enough in practice. Wdyt?

.claude/settings.json
malinajirka: Fixed in 3d7f0f1

.agents/skills/pos/SKILL.md
malinajirka: Good point about the failure mode. Added a POS vs Store quick-reference table and an explicit guard rail ("if working in POS code: do NOT use ScopedViewModel, triggerEvent, etc.") directly in the "How to tell which one you're in" section of AGENTS.md. This way the critical constraints are always loaded without needing the skill, while the detailed patterns stay in the skills. Fixed in 9c3c0b9.

AGENTS.md
malinajirka: @JorgeMucientes wdyt about this?

kidinov: > Perhaps, we could consider simply removing them and having the skills as the source of truth?

Docs and skill IMO are quite 2 different concepts, and I still think open-source project should have code conventions, "how to start" and so on. Also, those skills are something hot and used today, but tomorrow all that can be completely outdated or format changes to something entirely new, and they will get unused.

> Alternatively, we could consider asking claude to extract the knowledge into the https://github.com/docs files -> and reference them from the skills. That way, it'd be up-to-date.

Yeah, I'd rather do that

malinajirka: 👍 Cool, sounds good - I'll update it to update docs and reference them. I agree with your train of thoughts, good point that when skills get deprecated, we'd need to move it somehow somewhere again.

.claude/settings.json
```diff
@@ -22,6 +22,18 @@
       "Bash(gh api *)",
       "Bash(adb *)",
       "Bash(emulator *)",
+      "Skill(store-tests)",
```
kidinov, line 25: Skill(pos) appears twice, Skill(pr) appears twice, and Skill(woo-android-code-review) and (simplify?) references a skill that doesn't exist

docs/tracking-events.md
```diff
@@ -1,55 +0,0 @@
-# Tracking Events
```
kidinov, line 1: README.md:76 still links to this file so probably should be updated

docs/pos-architecture.md
```diff
@@ -0,0 +1,251 @@
...
+```kotlin
+sealed class WooPosMyFeatureUIEvent {
+    data class ItemClicked(val id: Long) : WooPosMyFeatureUIEvent()
+    data object BackClicked : WooPosMyFeatureUIEvent()
+}
+```
+
+## State Classes
+
+All POS state classes MUST be `@Parcelize` + `: Parcelable` for process-death recovery:
```
kidinov: np: this maybe not correct - WooPosActivity.onCreate passes null to super.onCreate() to force a fresh start, and several real POS state classes (WooPosSplashState, WooPosOrdersState, WooPosBookingsState) are not Parcelable. Maybe reword as "use Parcelable when state goes through saved state / nav args"?

docs/store-viewmodel-patterns.md
```diff
@@ -0,0 +1,200 @@
...
+        const val RESULT_KEY = "my_feature_result"
+    }
+}
+```
+
+### Key Rules
+
+1. Extend `ScopedViewModel(savedStateHandle)`
+2. Annotate with `@HiltViewModel` + `@Inject constructor`
+3. `SavedStateHandle` MUST be the first constructor parameter
```
kidinov, line 59: I think this is correct now, but just validating - the old AGENTS.md said "MUST be the last parameter" while here "MUST be the first". Codebase actually has both (~72 files first, ~13 last)

docs/store-compose.md
```diff
@@ -0,0 +1,165 @@
+# Store App — Jetpack Compose Guidelines
+
+> POS has its own Compose patterns — see [POS Architecture](pos-architecture.md).
+
+We follow the official [Compose API guidelines for App development](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#api-guidelines-for-jetpack-compose) and Google's Compose best practices. This doc covers only **project-specific** conventions and patterns.
+
+## Content
```
kidinov, line 7: The accessibility section (semantics, content descriptions, headings, state descriptions) and "always provide Modifier" rationale are gone. Probably worth keeping accessibility somewhere - neither store nor POS doc covers it now

docs/pull-request-guidelines.md
```diff
@@ -2,43 +2,36 @@
 
 ## Branch Naming
 
-Feature or fix branch names should use the pattern `issue/ISSUEID-description` where the `ISSUEID` is the Github issue number. For example, if the issue number is 1000 and the issue is related to an error with the order list, an appropriate branch name would be `issue/1000-order-list-error`.
+Feature or fix branch names should use the pattern `issue/ISSUEID-description` where the `ISSUEID` is the Linear issue identifier (e.g., `WOOMOB-1234`). For example, `issue/woomob-1234-order-list-error`.
 
-If there is no Github issue, you can use prefixes like `feature/` or `fix/`.
-
-> NOTE: Features are typically large PRs (pull requests) that contain smaller PRs that build a feature and are labeled as an `epic`.
+If there is no Linear issue, you can use prefixes like `feature/` or `fix/`.
```
kidinov, line 7: np: Small missmatch with pr/SKILL.md: here it allows feature//fix/ branches without a Linear issue, but the skill hardcodes [WOOMOB-XYZ] and Fixes WOOMOB-XY with no fallback. Skill needs a no-issue path as well

.agents/skills/pr/SKILL.md
kidinov: It adds ### Description / ### Test Steps / ### Images/gif etc here. If `.github/PULL_REQUEST_TEMPLATE.md` ever changes, the skill will silently drift. Probably better to instruct the skill to read the template at runtime.

kidinov (approved):
LGTM! I left a few suggestions, please take a look

.claude/settings.json
malinajirka: Simplify is one of the basic skills provided by Anthropic. The rest fixed in c69696567711f40adace98376f69a46e287dc85b. Good catch!

docs/tracking-events.md
malinajirka: Fixed in a8697d9455e8a48dca28ff634f371d3a54ef191e

docs/pos-architecture.md
malinajirka: Updated in f1c6d44260e207b917bc47913bd22874e72ee77e

docs/store-viewmodel-patterns.md
malinajirka: Yeah, this is intentional change - it should go first.

docs/store-compose.md
malinajirka: These are general non-project-specific guidelines, so based on recommendation, we shouldn't include them unless there is a specific reason for it. I'd personally keep it lean for now and we can add it if we notice Claude is not adding something it should. Let me know if you'd prefer to add it anyway.

docs/pull-request-guidelines.md
malinajirka: I have been using the skill without any issues so far, I think we can keep it as is. I kind of like that we push claude a bit towards trying to create the PR with woomob - it'll ask the user for it if it doesn't find it => user can always reply to create a PR without it.

.agents/skills/pr/SKILL.md
malinajirka: Good point, fixed in d3f77eef26ae66d3f7b2b0d7d4b00c51a03097bb

PR #15626, opened 2026-04-07
[WOOMOB-2514] Fix startup crash after selected site is invalidated in background

irfano:
### Description
Fixes WOOMOB-2514

When a self-hosted site's application password is revoked while the app is backgrounded, the logout flow clears the selected site. If process death happens before the user returns, Android restores stale `MainActivity` saved state on the next launch, but there's no selected site anymore, so the app crashes before the normal recovery path can take over.

**Fix:** Skip restoring `savedInstanceState` in `MainActivity.onCreate()` when no selected site exists. This lets the app start fresh and fall through to the site picker as expected. A static `SelectedSite.hasSelectedSiteId(context)` helper was added so this check can run before Hilt injection.

### Test Steps
1. Build `trunk` first and confirm the crash is reproducible with the steps below.
2. Apply the local repro patch below and build the app: [Schedule_a_one-time_background_sync_when_the_app_goes_to_background.patch](https://github.com/user-attachments/files/26542882/Schedule_a_one-time_background_sync_when_the_app_goes_to_background.patch)
3. Log in with a self-hosted site using the web authorization flow (enter incorrect credentials first to trigger app-password setup).
4. In wp-admin, open Users then Profile, keep the Application Passwords section ready.
5. Background the app, then revoke the Woo app password from wp-admin within 10 seconds.
6. Verify in logs: the regular sync fails, the app password is deleted, and logout starts.
7. Run `adb shell am kill com.woocommerce.android.dev` to simulate process death.
8. Reopen the app.
9. **trunk:** app crashes. **This branch:** app opens to the site picker without crashing.

### Images/gif
N/A


libs/commons/src/main/java/com/woocommerce/android/tools/SelectedSite.kt
```diff
@@ -33,6 +33,16 @@ class SelectedSite @Inject constructor(
         const val SELECTED_SITE_LOCAL_ID = "SELECTED_SITE_LOCAL_ID"
 
         fun getEventBus(): EventBus = EventBus.getDefault()
+
+        fun getSelectedSiteId(context: Context): Int {
```
hichamboushaba: np, I believe this can be private, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -320,7 +320,8 @@ class MainActivity :
 
     override fun onCreate(savedInstanceState: Bundle?) {
         installSplashScreen()
-        super.onCreate(savedInstanceState)
+        val bundle = if (SelectedSite.hasSelectedSiteId(this)) savedInstanceState else null
```
hichamboushaba, line 324: I suggest adding a comment explaining this logic, in the future it won't be easy to remember why we are doing it.

hichamboushaba (approved):
Nice catch @irfano, I left two comments for addressing, but I'm pre-approving.

libs/commons/src/main/java/com/woocommerce/android/tools/SelectedSite.kt
irfano: Good catch! Done: 02bbfdea042e6a1380963a979df8d3ca18b5323e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
irfano: Added a short comment here: dcea847100d

Issue #5841, opened 2022-02-11
Change currency wherever necessary to support Canada


AnirudhBhat:
Modify the logic of the currency selection to support Canada wherever necessary in our project.

- `PaymentUtils.kt`
- `CardReaderPaymentCollectibilityChecker.kt`

PR #15627, opened 2026-04-07
Improve verify-on-device skill: modular references, POS coverage, forked context

malinajirka:
I think one reviewer is enough, since we can always revert, but feel encouraged to share any feedback.

## Summary

- **Modular navigation references**: Moved screen identifiers and workflows out of the monolithic SKILL.md into 7 on-demand reference files (login, dashboard, orders, products, more menu, main navigation overview, POS). The skill instructs the agent to load only the reference files relevant to the current task, reducing context cost.
- **POS coverage**: Added `pos-navigation.md` with full feature tree, screen identifiers, test tags, and workflows (checkout, orders/refunds, bookings). POS had zero navigation guidance before.
- **Forked execution (`context: fork`)**: The skill now runs in an isolated sub-agent automatically. All heavy mobile-mcp element dumps (50-200+ elements per call) stay out of the main conversation — only a clean summary is returned.
- **New critical rules**: "Default to Main App" (prevents POS/main app confusion) and "Always Restart the App" (avoids wasted time navigating out of unknown screens).
- **Leaner reference files**: Evaluated adding logcat events as a faster navigation verification approach (66-86ms vs 3-5s for accessibility tree). However, logcat only confirms "action received" — you still need the accessibility tree to find the next tap target, so it adds a step without saving one. Decided to keep reference files lean and omit logcat events.
- **Skip-to-step shortcut**: Added note at the top of Steps section for when the app is already installed.

## Testing performed

1. **On-demand loading test**: Spawned a research agent to analyze whether SKILL.md instructions are clear enough to guide correct reference file loading. Confirmed the routing table works — agent correctly identified which files to load for an orders task and which to skip.
2. **End-to-end device test**: Ran the skill against emulator with task "verify orders list and order detail." Agent loaded SKILL.md, main-app-navigation.md, main-app-orders.md, navigated to Orders tab, opened order #5841, verified all sections present. Passed.
3. **Logcat vs accessibility tree benchmark**: Timed both approaches across Dashboard, Orders, Order Detail navigation. Logcat: 66-86ms. Accessibility tree: 3-5s. Decided against logcat since it does not eliminate the accessibility tree call.
4. **Forked context test**: Invoked skill via /verify-on-device after adding context: fork. Confirmed it runs in isolated sub-agent — main conversation received only a clean summary, no element dump noise.

## Test plan

- [ ] Run /verify-on-device with a main app task (e.g., orders) and confirm reference files load correctly
- [ ] Run /verify-on-device with a POS task and confirm POS reference loads
- [ ] Verify forked execution keeps main conversation clean


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.claude/settings.json
```diff
@@ -21,7 +21,30 @@
       "Bash(gh issue *)",
       "Bash(gh api *)",
       "Bash(adb *)",
-      "Bash(emulator *)"
+      "Bash(emulator *)",
+      "Bash(emulator *)",
```
irfano: The `"Bash(emulator *)"` entry is duplicated here.

.agents/skills/verify-on-device/SKILL.md
```diff
@@ -114,6 +134,8 @@ Plan your verification flow accordingly. If the user wants to test post-login fe
 
 ## Steps
 
+**Shortcut:** If the app is already installed and logged in, skip to step 7 (Restart and Launch the App).
```
irfano: I think it would be better to skip to step 6 instead of step 7, since that would also cover cases where a mock response is required.

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,92 @@
+# Login Flow
+
+Uses Fragment navigation inside `LoginActivity`. Multiple paths depending on login method.
+
+## Screen Detection
+
+| Screen | How to Detect |
+|--------|---------------|
+| Prologue | text: "Log in with WordPress.com" and "Log in with site address" |
```
irfano: This is incorrect. We don’t have these texts on our prologue screen.

Image

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,92 @@
...
+| WPCom Email | text: email input field, WooCommerce/Jetpack logo |
```
irfano: We don't have WooCommerce/Jetpack logo on this screen.

Image

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,92 @@
...
+| Password | text: password input field, "Forgot password?" link |
```
irfano: It would be better to use the actual text (“Reset your password”) here.

```suggestion
| Password | text: password input field, "Reset your password?" link |
```

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,92 @@
...
+| 2FA / Verification Code | text: "Enter verification code" or OTP input |
```
irfano: Image

Same here. I believe using the exact same strings would help here as well.

```suggestion
| 2FA / Verification Code | text: "Verification code" or OTP input |
```

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,92 @@
...
+| Magic Link Sent | text: "Check your email" or "Open email client" |
```
irfano: I believe using ambiguous phrases like "Check your email" or "Open email client" makes the skill less reliable. I think we should use the exact string shown on the screen instead of two similar strings joined with "or".

There seem to be more cases like this, so rather than leaving the same comment repeatedly: is there a reason not to use the exact strings?

Image

.agents/skills/verify-on-device/references/main-app-more.md
```diff
@@ -0,0 +1,135 @@
...
+- "Settings" -> Settings activity
+
+## Screen Identifiers
+
+**Reviews List** -- Fragment: `ReviewListFragment`
+
+| Key Element | Identifier | Notes |
+|-------------|-----------|-------|
+| **Primary** | `reviewsList` | Reviews RecyclerView |
+| Unread filter | `unread_filter_switch` | Toggle to filter unread |
```
irfano: `reviewsList` and `unread_filter_switch` are IDs from the product reviews list (`fragment_product_reviews_list.xml`), not from the Reviews screen opened from the menu (`ReviewListScreen`). We should fix this; otherwise, the skill may open the wrong Reviews screen.

irfano (changes requested):

.claude/settings.json
malinajirka: Fixed — removed the duplicate entry. (47b0399fbf2)

.agents/skills/verify-on-device/SKILL.md
malinajirka: Good point — updated to skip to step 6 so the mock setup step is covered. (34deb995968)

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed. The prologue shows "Log In" button and "Starting a new store?" link. Also updated the WPCom happy path workflow since the WPCom button is hidden — both flows now start with "Log In". (b61af49d633, a879b9ac17e)


It's almost guaranteed there are more mistakes, but I wouldn't worry about it too much. Most of them can AI figure out itself, when not -> that will be a good time to fix the issue any-of-us encounters. Wdyt?

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed — removed the logo reference. The email screen now uses the WooCommerce-specific string: "Log in with your WordPress.com account email address to manage your WooCommerce stores." (b61af49d633, a879b9ac17e)

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed — now uses "Reset your password". (b61af49d633)

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed — now uses "Verification code". (b61af49d633)

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Good call — went through all screens and replaced every ambiguous phrase with the exact string from strings.xml. Also fixed site address, site credentials, site picker, discovery error, Jetpack, Google SSO, and 2FA alternatives. (b61af49d633, a879b9ac17e, 60f8cd43716, 171067356a5)

.agents/skills/verify-on-device/references/main-app-more.md
malinajirka: Fixed — the Reviews screen from More Menu is Compose-based (ReviewListScreen). Now uses "Unread reviews only" text as the screen identifier instead of the XML IDs from the product reviews list. (71506df7f7a, 5f9cf6f8204)

.claude/settings.json
```diff
@@ -22,6 +22,28 @@
...
+      "mcp__mobile-mcp__mobile_save_screenshot",
+      "mcp__mobile-mcp__mobile_set_orientation",
+      "mcp__mobile-mcp__mobile_start_screen_recording",
+      "mcp__mobile-mcp__mobile_stop_screen_recording",
+      "mcp__mobile-mcp__mobile_swipe_on_screen",
+      "mcp__mobile-mcp__mobile_take_screenshot",
+      "mcp__mobile-mcp__mobile_terminate_app",
+      "mcp__mobile-mcp__mobile_type_keys",
+      "mcp__mobile-mcp__mobile_uninstall_app",
+      "Skill(verify-on-device)"
```
irfano: The last commit breaks this line. We need a comma here.

.agents/skills/verify-on-device/references/main-app-more.md
```diff
@@ -0,0 +1,134 @@
...
+| 4 | Tap to Pay | text: "Tap to Pay" |
+| 5 | Toggle Cash on Delivery | CoD toggle |
+
+### Reviews
+
+| Step | Action | Element |
+|------|--------|---------|
+| 1 | Tap "Reviews" in menu | text: "Reviews" |
+| 2 | Toggle unread filter | `unread_filter_switch` |
+| 3 | Tap a review | review row in `reviewsList` |
```
irfano: `unread_filter_switch` and `reviewsList` are IDs from the XML file. We don’t use them on the Reviews screen opened from the Menu tab.

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,93 @@
...
+| 1 | Tap "Log In" | id: `buttonLoginStore` |
+| 2 | Enter store URL | URL text field |
+| 3 | Tap Continue | continue button |
+| 4 | Enter WPCom email (if Jetpack connected) | email text field |
+| 5 | Enter password | password text field |
+| 6 | Tap Continue | continue button |
+| 7 | Select store (if multi-site) | store list row |
+| 8 | Tap Continue | text: "Continue" |
+
+### Happy Path: Site Address
```
irfano: What is` ### Happy Path: Site Address`? It looks the same as the one above (`### Happy Path: WPCom Email + Password`).

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,93 @@
...
+| Jetpack Activation | text: "Please install the free Jetpack plugin" with "Install Jetpack" or "Connect Jetpack" button |
+| Not WordPress Error | text: "We were not able to detect a WordPress site at the address you entered." |
+
+## Workflows
+
+### Happy Path: WPCom Email + Password
+
+| Step | Action | Element |
+|------|--------|---------|
+| 1 | Tap "Log In" | id: `buttonLoginStore` |
```
irfano: The actual id is `button_login_store`.

```suggestion
| 1 | Tap "Log In" | id: `button_login_store` |
```

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,93 @@
...
+| 5 | Enter password | password text field |
+| 6 | Tap Continue | continue button |
+| 7 | Select store (if multi-site) | store list row |
+| 8 | Tap Continue | text: "Continue" |
+
+### Happy Path: Site Address
+
+| Step | Action | Element |
+|------|--------|---------|
+| 1 | Tap "Log In" | id: `buttonLoginStore` |
```
irfano: The actual id is `button_login_store`.

```suggestion
| 1 | Tap "Log In" | id: `button_login_store` |
```

.agents/skills/verify-on-device/references/main-app-login.md
```diff
@@ -0,0 +1,93 @@
...
+| 4 | (Magic link opened from email) | |
+
+### Google SSO
+
+| Step | Action | Element |
+|------|--------|---------|
+| 1 | Tap "Continue with Google" | id: `continue_with_google` |
+| 2 | Complete Google auth | Google account picker |
+
+### Jetpack Activation (when Jetpack not installed)
```
irfano: `(when Jetpack not installed)` is misleading because Jetpack might be installed but not activated, and we still use this flow. The steps below are correct because they cover both cases in step 2: "Install Jetpack" or "Connect Jetpack"

.agents/skills/verify-on-device/references/main-app-login.md
irfano: > It's almost guaranteed there are more mistakes, but I wouldn't worry about it too much. Most of them can AI figure out itself, when not -> that will be a good time to fix the issue any-of-us encounters. Wdyt?

Unknown gaps are fine, but I would prefer having either no context or accurate context.

irfano (changes requested):
I reviewed all the PRs this time and added a few more comments.

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Might depend on case-by-case, but in general it's more useful for claude to have "Tap on 'Login' button" even when the button says "Continue" then not having any context - similar to humans, we don't care if the instructions are 100% precise, we just care that we know how roughly we should navigate where we need. Having said that, of course if we can make it precise with a reasonable effort, I'm all up for it.

.claude/settings.json
malinajirka: Fixed in e2e10ef8aa1890365c289c5d4ef900bbbafdc293

.agents/skills/verify-on-device/references/main-app-more.md
malinajirka: Fixed in 80b2662fbca35fafb343535451049bc1319b9997

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Good catch, fixed in 720c44bba0d4d31cf6657f31b041344c4cf64a1f

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed in 9d4bfca4c82af5be0523eeefdd3847d905d4d68d

.agents/skills/verify-on-device/references/main-app-login.md
malinajirka: Fixed in 9180590cea0a1b700b72ef9ebe893b2fdf159c14

malinajirka (commented):
Thanks for the thorough review @irfano! I've implemented the suggested changes. Let me know what you think, thanks!

irfano (approved):
Thanks for all the fixes and for improving this skill. LGTM! 👍🏻

PR #15629, opened 2026-04-08
[WOOMOB-2548] Fix back button navigating background behind POS dialog

malinajirka:
### Description
Fixes WOOMOB-2548

When the barcode scanner setup dialog (or any POS dialog using `WooPosDialogWrapper`) was visible, pressing the back button navigated the background settings flow instead of dismissing the dialog. Added a `BackHandler` to `WooPosDialogWrapper` that intercepts back presses when the dialog is visible and calls `onDismissRequest`.

### Test Steps
1. Open POS
2. Go to Settings → Hardware → Barcode scanners
3. Tap "Scanner Setup" to open the dialog
4. Press the back button
5. Verify the dialog is dismissed and the background stays on the "Barcode scanners" screen

### Images/gif
N/A


kidinov (approved):
LGTM!

PR #15633, opened 2026-04-08
[WOOMOB-2521] Read and parse order detail fulfillment data

irfano:
### Description
Fixes WOOMOB-2521
This PR adds the fulfillment data pipeline for order details by fetching `/wc/v3/orders/{id}/fulfillments`, storing the result locally, and loading it for CIAB sites.
It covers the fulfillment fields needed for order detail data persistence and keeps local records in sync with the latest server state.
This PR does not cover fulfillment method distinction because there is no confirmed `_fulfillment_method` metadata key.

### Test Steps
1. Log in to a CIAB site that has an order with fulfillment data.
2. Open the order detail screen for a fulfilled order.
3. In Android Studio, open App Inspection and then Database Inspector.
4. Inspect the `OrderFulfillmentEntity` table.
5. Verify that there is a row for the order you opened.
6. Verify that the expected fields are populated when present in the API response.
   `fulfillmentId`
   `status`
   `isFulfilled`
   `dateUpdated`
   `dateFulfilled`
   `trackingNumber`
   `shipmentProvider`
   `trackingUrl`
7. Log in to a non CIAB site.
8. Open an order detail screen.
9. Verify that no fulfillment rows are created for the non CIAB order you opened.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
@@ -279,9 +279,8 @@ class OrderDetailRepository @Inject constructor(
         orderStore.getShipmentTrackingsForOrder(selectedSite.get(), orderId).map { it.toAppModel() }
     }
 
-    fun getOrderFulfillments(orderId: Long): List<OrderFulfillment> = runBlocking {
+    suspend fun getOrderFulfillments(orderId: Long): List<OrderFulfillment> =
         orderStore.getOrderFulfillmentsForOrder(selectedSite.get(), orderId).map { it.toAppModel() }
-    }
 
     fun getOrderShippingLabels(remoteOrderId: Long) = shippingLabelStore
         .getShippingLabelsForOrder(selectedSite.get(), remoteOrderId)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
@@ -10,6 +10,7 @@ import com.woocommerce.android.analytics.AnalyticsTracker.Companion.VALUE_API_FA
 import com.woocommerce.android.analytics.AnalyticsTracker.Companion.VALUE_API_SUCCESS
 import com.woocommerce.android.model.Order
 import com.woocommerce.android.model.Order.OrderStatus
+import com.woocommerce.android.model.OrderFulfillment
 import com.woocommerce.android.model.OrderMapper
 import com.woocommerce.android.model.OrderNote
 import com.woocommerce.android.model.OrderShipmentTracking
@@ -89,6 +90,13 @@ class OrderDetailRepository @Inject constructor(
                 RequestResult.ERROR
             }
         }
+    }
+
+    suspend fun fetchOrderFulfillments(orderId: Long): Boolean = withContext(dispatchers.io) {
+        val result = withTimeoutOrNull(AppConstants.REQUEST_TIMEOUT) {
+            orderStore.fetchOrderFulfillments(orderId, selectedSite.get())
+        }
+        result?.isError == false
     }
 
     suspend fun fetchOrderRefunds(orderId: Long): List<Refund> {
@@ -269,6 +277,10 @@ class OrderDetailRepository @Inject constructor(
 
     fun getOrderShipmentTrackings(orderId: Long) = runBlocking {
         orderStore.getShipmentTrackingsForOrder(selectedSite.get(), orderId).map { it.toAppModel() }
+    }
+
+    fun getOrderFulfillments(orderId: Long): List<OrderFulfillment> = runBlocking {
+        orderStore.getOrderFulfillmentsForOrder(selectedSite.get(), orderId).map { it.toAppModel() }
     }
 
     fun getOrderShippingLabels(remoteOrderId: Long) = shippingLabelStore
--- a/libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -561,6 +561,17 @@ class MigrationTests {
         }
     }
 
+    @Test
+    fun testMigration80to81_addsOrderFulfillmentEntity() {
+        helper.createDatabase(TEST_DB, 80).close()
+
+        helper.runMigrationsAndValidate(TEST_DB, 81, true).query(
+            "SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'OrderFulfillmentEntity'"
+        ).use { cursor ->
+            assertThat(cursor.count).isEqualTo(1)
+        }
+    }
+
     companion object {
         private const val TEST_DB = "migration-test"
     }
--- a/libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -561,17 +561,6 @@ class MigrationTests {
         }
     }
 
-    @Test
-    fun testMigration80to81_addsOrderFulfillmentEntity() {
-        helper.createDatabase(TEST_DB, 80).close()
-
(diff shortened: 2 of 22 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
```diff
@@ -561,6 +561,17 @@ class MigrationTests {
         }
     }
 
+    @Test
+    fun testMigration80to81_addsOrderFulfillmentEntity() {
```
AdamGrzybkowski: I'm not sure we need to test it with auto migration, but it won't hurt either.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderFulfillmentApiResponse.kt
```diff
@@ -0,0 +1,13 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.order
+
+import com.google.gson.JsonElement
+import org.wordpress.android.fluxc.network.Response
+
+@Suppress("PropertyName", "VariableNaming")
```
AdamGrzybkowski: This won't be needed if we change the properties to:
`@SerializedName("is_fulfilled") val isFulfilled: ...`

Not using `SerializedName` annotation could lead to a crash in the release build.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -818,6 +835,29 @@ class WCOrderStore @Inject internal constructor(
...
+                    RemoteId(result.orderId)
+                )
+                val removedFulfillments = existingFulfillments.filter { existing ->
+                    result.fulfillments.none { fresh ->
+                        fresh.fulfillmentId == existing.fulfillmentId
+                    }
+                }
+
+                removedFulfillments.forEach { orderFulfillmentDao.deleteOrderFulfillment(it) }
+                result.fulfillments.forEach { orderFulfillmentDao.upsertOrderFulfillment(it) }
```
AdamGrzybkowski: Would it make sense to wrap this in transaction?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -818,6 +835,29 @@ class WCOrderStore @Inject internal constructor(
...
+            } else {
+                val existingFulfillments = orderFulfillmentDao.getOrderFulfillments(
+                    result.site.localId(),
+                    RemoteId(result.orderId)
+                )
+                val removedFulfillments = existingFulfillments.filter { existing ->
+                    result.fulfillments.none { fresh ->
+                        fresh.fulfillmentId == existing.fulfillmentId
+                    }
+                }
```
AdamGrzybkowski: Any reason for this particular logic? I wonder if we can simplify and simply remove/insert all.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
```diff
@@ -92,6 +96,100 @@ class OrderRestClientTest {
...
+                dateFulfilled = null,
+                trackingNumber = null,
+                shipmentProvider = null,
+                trackingUrl = null
+            )
+        )
+    }
+
+    @Test
+    fun `when fulfillment metadata is missing, then fields default safely`() = runTest {
```
AdamGrzybkowski: Technically, the previous test already covers that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
```diff
@@ -271,6 +279,10 @@ class OrderDetailRepository @Inject constructor(
         orderStore.getShipmentTrackingsForOrder(selectedSite.get(), orderId).map { it.toAppModel() }
     }
 
+    fun getOrderFulfillments(orderId: Long): List<OrderFulfillment> = runBlocking {
```
AdamGrzybkowski: Are you sure we can use`runBlocking`? This is only used in a test, but could easily be used later on in the app.

AdamGrzybkowski (commented):
Looks good, but I left some comments that I think are worth addressing. The one with the property naming, especially as it could lead to a crash/error in the release build.

libs/fluxc-plugin/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
irfano: I saw other tests that cover auto migration as well, so I kept this, but I agree with you that it wasn’t necessary, so I removed it: 122635ce5ff320d519671f21a6c9087671205c10

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderFulfillmentApiResponse.kt
irfano: I'm not sure if this could cause to a crash, because we use the same format in other models as well. I used this approach to follow the style in [OrderShipmentTrackingApiResponse](https://github.com/woocommerce/woocommerce-android/blob/21f75dc575e9a41f4cd43cc51072496354ca2deb/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderShipmentTrackingApiResponse.kt#L13-L12). But I agree that your suggestion is better, so I changed the code: e5a5312f3d80702261010e64b4a5f4255b21196e

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
irfano: Great advice! It’s definitely better to do this in a single transaction.
Done: 8416aa9366057652fbd920e99fa1dfd6f1661583

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
irfano: Done: 8416aa9366057652fbd920e99fa1dfd6f1661583

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
irfano: Good catch! I removed it: 923f253fbc76edb181483ea2a7c4bf2d52f6dcab

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderFulfillmentApiResponse.kt
AdamGrzybkowski: Right, I've checked, and we exclude those from obfuscation with r8.

```
-keep class org.wordpress.android.fluxc** { *; }
```

Still, I think it's best to separate the property name from the response value name, so thanks for the changes!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
irfano: I did that to follow the pattern used by the other functions that access orderStore in this class. But we don’t know yet whether we’ll need a non-suspend function in the view model. So, I applied your suggestion and converted the function to `suspend` for now: 7308c8691b4878cf273ed1f4eebd63303624ceeb

AdamGrzybkowski (approved):

PR #15635, opened 2026-04-09
[WOOMOB-2557][POS] Add enter/exit animations for refund dialogs

samiuelson:
### Description
Fixes WOOMOB-2557

The refund details dialog and issue refund dialog lacked proper enter/exit animations. Both were conditionally added/removed from the Compose tree and hardcoded `isVisible = true` to `WooPosDialogWrapper`, so `AnimatedVisibility` could never run exit transitions.

This fix keeps the dialogs in the composition tree using a `rememberRetained` utility (caches last non-null dialog state for exit animation rendering) and drives `isVisible` from the current dialog state — matching the pattern used by `WooPosExitConfirmationDialog` and `WooPosCancelBookingDialog`.

### Test Steps
1. Open a POS order that has a refund
2. Tap "View details" on a refund in the Totals section
3. Verify the refund details dialog animates in (fade + slide up)
4. Close the dialog
5. Verify the dialog animates out (fade + slide down)
6. Tap "Issue Refund" on an order
7. Verify the issue refund dialog animates in
8. Close/dismiss the dialog
9. Verify it animates out
10. Repeat steps 6-9 from the Bookings screen

### Images/gif
https://github.com/user-attachments/assets/e5b1dd73-ed89-4502-8cb3-95e9ee084a16


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -214,23 +215,27 @@ private fun WooPosOrdersScreen(
         }
 
         if (state is WooPosOrdersState.Content) {
-            when (val dialogState = state.dialogState) {
-                is WooPosOrdersState.Content.DialogState.IssueRefund -> {
+            val dialogState = state.dialogState
```
malinajirka: ❓ This is not blocking this PR, but by replacing the exhaustive `when` with `rememberRetained` + `?.let`, we lose the compile-time guarantee that new `DialogState` subclasses get handled in these screens. If someone adds a new dialog state to the sealed class, this screen would silently ignore it without a compiler warning.

malinajirka (changes requested):

malinajirka:
The "items to refund" never finish loading, it's visible also in your recording. 🤔

samiuelson:
Good catch @malinajirka! 🙇 Fixed in a8095c4394c.

**Root cause:** `rememberRetained` keeps `WooPosIssueRefundDialog` composed across the exit animation, so `LaunchedEffect(Unit)` only ran on the very first open. On dismiss, `WooPosRefundViewModel.handleDialogDismissed()` resets state to `Loading` and cancels the loading job — so on reopen the VM was stuck in `Loading` forever because `DialogOpened` was never re-fired.

**Fix:** Key the effect on `isVisible` so `DialogOpened` is dispatched on every `false → true` transition, matching the pre-animation "composed once per open" behavior:

```kotlin
LaunchedEffect(isVisible) {
    if (isVisible) {
        viewModel.onUIEvent(WooPosRefundUIEvent.DialogOpened)
    }
}
```


https://github.com/user-attachments/assets/d296a18b-c05d-4357-9f29-5ef746172bd6

malinajirka (approved):
Thanks for the fix @samiuelson!

It works as expected now. Having said that, I noticed the loading takes a looooong time. Looking at the inspector, there are 6 requests sent, each of them takes roughly 1.5s and they all run sequentially => 9s loading time. If I recall correctly, it wasn't this bad before. I'm wondering if we need all of these requests - I'd expect most of the things to be cached - and I'm also wondering if there is an opportunity for some parallelism. Wdyt? (Not related to this PR, but perhaps something we might want to create an issue for, unless it already exists)
Screenshot 2026-04-10 at 11 42 22

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: Good point @malinajirka — addressed in 179c9f7ad05.

The dialog rendering now uses a single `rememberRetained` fed by an exhaustive `when` expression, followed by a dispatch `when` (also exhaustive). Adding a new `DialogState` subtype produces **two compile errors** in the same file — one in the retain mapping and one in the dispatch — forcing the developer to explicitly decide whether the new subtype should survive exit animations and how it should be rendered.

```kotlin
val retainedDialog = rememberRetained(
    when (dialogState) {
        is DialogState.IssueRefund -> dialogState
        is DialogState.RefundDetails -> dialogState
        DialogState.Hidden -> null
    }
)

when (retainedDialog) {
    is DialogState.IssueRefund -> WooPosIssueRefundDialog(...)
    is DialogState.RefundDetails -> WooPosRefundDetailsDialog(...)
    DialogState.Hidden, null -> Unit
}
```

`rememberRetained` is still called unconditionally at a stable slot, so the exit-animation behavior is unchanged. On an `IssueRefund → Hidden` transition, the inner `when` returns `null`, the retained value stays as `IssueRefund`, the dispatch `when` stays in the `is IssueRefund` branch (same slot group), and `WooPosIssueRefundDialog` keeps its `remember`s and `hiltViewModel` while `isVisible` flips to `false` to play the exit animation.

`WooPosBookingsScreen` got the narrowed-type variant since it only retains `IssueRefund` (`CancelBooking` is rendered separately in `WooPosBookingsContent`), but the same exhaustiveness guarantee applies via the inner `when`.

samiuelson:
> Thanks for the fix @samiuelson!
> 
> It works as expected now. Having said that, I noticed the loading takes a looooong time. Looking at the inspector, there are 6 requests sent, each of them takes roughly 1.5s and they all run sequentially => 9s loading time. If I recall correctly, it wasn't this bad before. I'm wondering if we need all of these requests - I'd expect most of the things to be cached - and I'm also wondering if there is an opportunity for some parallelism. Wdyt? (Not related to this PR, but perhaps something we might want to create an issue for, unless it already exists) Screenshot 2026-04-10 at 11 42 22

We decided to get fresh data and don't rely on caching despite the speed cost because we were often dealing with incorrect refund requests related to stale data. I agree though that some of these calls may be redundant and we could try making them more efficient. The speed is expected to improve when we improve the refund endpoints in Woo core, and I'd prefer to focus on that now, and if we encounter any blockers fixing the APIs, optimizing the mobile app.

PR #15637, opened 2026-04-09
[WOOMOB-2503][POS] Make Select All row clickable in POS refund dialog

samiuelson:
### Description
Fixes WOOMOB-2503

Makes the entire "Select All" row clickable in the POS refund dialog, instead of only the checkbox. Also removes an unnecessary `.clip()` modifier from `RefundableItemRow` since there is no ripple indication to constrain.

### Test Steps
1. Open a POS order with multiple items
2. Tap "Issue Refund"
3. Verify tapping anywhere on the "Select All" row toggles the checkbox
4. Verify tapping anywhere on individual item rows still toggles their checkboxes

### Images/gif
https://github.com/user-attachments/assets/27d2fe7f-217c-44a7-b935-cf4624c81c52


kidinov (approved):
LGTM! Thanks!

PR #15639, opened 2026-04-09
[WOOMOB-2493] Add shimmer loading state to historical refunds list

samiuelson:
### Description
Fixes WOOMOB-2493

In POS > Orders > Order Details, the Totals section now shows shimmer placeholders while historical refund data loads asynchronously, instead of showing nothing until data arrives. The shimmer is only displayed when we already know refunds exist (from `order.refundTotal`).

### Test Steps
1. Open POS > Orders
2. Select an order that has historical refunds (partial or full)
3. Observe the Totals section — a shimmer placeholder should briefly appear where refund rows will load
4. Once data loads, shimmer transitions to actual refund rows (or error message)
5. Select an order without refunds and confirm no shimmer appears in the Totals section

### Images/gif
Screenshot_20260409_170758
Screenshot_20260409_170808


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
```diff
@@ -555,6 +558,38 @@ private fun RefundRowContent(
...
+
+        WooPosShimmerText(
+            text = stringResource(R.string.woopos_orders_details_refund_shimmer_date_placeholder),
+            style = WooPosTypography.BodyMedium.style,
+        )
+
+        Spacer(Modifier.height(WooPosSpacing.Small.value))
+
+        WooPosShimmerText(
+            text = stringResource(R.string.woopos_orders_details_refund_view_details),
```
kidinov: Does it need to be translatable? It just to measure width of the shimmer, right?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
```diff
@@ -113,7 +113,8 @@ sealed class WooPosOrdersState {
                     val shipping: String?,
                     val refunds: List<RefundRow>,
                     val netPayment: String?,
-                    val refundLoadError: String? = null
+                    val refundLoadError: String? = null,
```
kidinov: With this state definition, we can have states like refund loading with error and there are refunds in the list - all at the same time. So depending on the way you handle this on the UI, it can show all of that or something. IMO, potential source of bugs even for AI

kidinov (approved):
LGTM! But I left a couple of comments - one not really related to this PR, though.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetails.kt
samiuelson: Fixed in 599a134

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersState.kt
samiuelson: Fixed in c2bbe76

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15641, opened 2026-04-10
[WOOMOB-2655] Add feature flag to gate POS on phones

kidinov:
### Description
Fixes WOOMOB-2655

Add a `WOO_POS_PHONE` feature flag that controls whether POS can run on phone-sized devices. When enabled (debug builds by default), phones pass the screen size gate and POS activities lock to portrait orientation. When disabled, phones are blocked from POS as before.

**Changes:**
- New `WOO_POS_PHONE` flag in `FeatureFlag.kt` (debug-enabled, remotely controlled)
- `isWooPosPhoneLayout()` Context extension to detect phone-sized screens (shortest side < 674dp)
- `WooPosIsScreenSizeAllowed` now checks the phone flag for phone-sized devices instead of rejecting them outright (single source of truth)
- All three POS activities (`WooPosActivity`, `WooPosCardReaderActivity`, `WooPosCouponCreationActivity`) lock to portrait on phones, landscape on tablets

### Test Steps
1. On a **phone emulator** (debug build): open the app and verify the POS tab is visible
2. Tap the POS tab and verify the activity opens in **portrait**
3. On a **tablet emulator**: verify POS still works as before in **landscape**
4. Disable the `WOO_POS_PHONE` flag via feature flag overrides: verify the POS tab is **hidden** on the phone

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowed.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowed.kt
@@ -23,7 +23,7 @@ class WooPosIsScreenSizeAllowed @Inject constructor(
         val shortSize = min(screenWidthDp, screenHeightDp)
         val longSize = max(screenWidthDp, screenHeightDp)
 
-        val isTabletSize = shortSize >= MIN_SCREEN_SHORT_SIZE_DP && longSize >= MIN_SCREEN_LONG_SIZE_DP
+        val isTabletSize = shortSize >= MIN_TABLET_SHORT_SIZE_DP && longSize >= MIN_TABLET_LONG_SIZE_DP
         if (isTabletSize) return true
 
         val isPhoneWithFlagEnabled = context.isWooPosPhoneLayout() &&
@@ -33,13 +33,13 @@ class WooPosIsScreenSizeAllowed @Inject constructor(
         wooPosLog.i(
             "POS Not allowed reason: Screen size is not allowed. " +
                 "Short size: $shortSize, Long size: $longSize, " +
-                "Minimum short size: $MIN_SCREEN_SHORT_SIZE_DP, Minimum long size: $MIN_SCREEN_LONG_SIZE_DP"
+                "Minimum short size: $MIN_TABLET_SHORT_SIZE_DP, Minimum long size: $MIN_TABLET_LONG_SIZE_DP"
         )
         return false
     }
 
-    private companion object {
-        const val MIN_SCREEN_SHORT_SIZE_DP = 674
-        const val MIN_SCREEN_LONG_SIZE_DP = 800
+    companion object {
+        internal const val MIN_TABLET_SHORT_SIZE_DP = 674
+        internal const val MIN_TABLET_LONG_SIZE_DP = 800
     }
 }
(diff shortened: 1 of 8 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowed.kt
```diff
@@ -19,20 +23,23 @@ class WooPosIsScreenSizeAllowed @Inject constructor(
...
-                        "Minimum short size: $MIN_SCREEN_SHORT_SIZE_DP, Minimum long size: $MIN_SCREEN_LONG_SIZE_DP"
-                )
-            }
-        }
+        val isTabletSize = shortSize >= MIN_TABLET_SHORT_SIZE_DP && longSize >= MIN_TABLET_LONG_SIZE_DP
+        if (isTabletSize) return true
+
+        val isPhoneWithFlagEnabled = context.isWooPosPhoneLayout() &&
+            featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_PHONE)
+        if (isPhoneWithFlagEnabled) return true
```
malinajirka: ❓ Non-blocking: there's a small dead zone here where a device has `shortSize >= 674` but `longSize < 800` (think near-square foldables or a tablet in split-screen that ends up with `longSize` around 780). It isn't a tablet by the check on L26 and `isWooPosPhoneLayout()` returns `false` because `shortSize >= 674`, so the flag can never rescue it — same behavior as before the PR, so not a regression.

Given the flag's intent is "allow POS on non-tablet devices", would it be simpler to gate on `!isTabletSize` instead of `isWooPosPhoneLayout()`? That would also collapse the two short-side comparisons into one source of truth. Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowedTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        displayMetrics.widthPixels = 411
+        displayMetrics.heightPixels = 891
+        whenever(featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_PHONE)).thenReturn(false)
+
+        // WHEN
+        val result = sut()
+
+        // THEN
+        assertThat(result).isFalse()
+    }
```
malinajirka, line 74: 🔍 np: consider adding a boundary test right at `674 x 800` (should be allowed) and one at `673 x 800` (should fall through to the phone branch). Right now an off-by-one refactor from `>=` to `>` would still pass all three tests. A density != 1 case would also lock in the dp conversion in `getScreenWidthDp()`/`getScreenHeightDp()`. Not blocking.

malinajirka (approved):
LGTM overall. I left a non-blocking question about a small dead zone between the tablet and phone branches plus a couple of nits on layering and test boundaries. Feel free to address or ignore.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowed.kt
kidinov: Good catch on the dead-zone observation. Simplified the phone branch to just rely on the flag since we already return early when `isTabletSize` is true — one source of truth for the tablet short-side check, and near-square foldables / split-screen cases with `shortSize >= 674` but `longSize < 800` can now be rescued by the flag.

Fixed in fa7b677691e

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/WooPosIsScreenSizeAllowedTest.kt
kidinov: Added three tests: `674x800` (tablet boundary, allowed), `673x800` (just below, falls through to flag), and a `density=2` case where pixels look tablet-sized but dp is phone-sized (locks in the dp conversion).

Fixed in 2d14294c4d6

PR #15646, opened 2026-04-10
[WOOMOB-2586] Fallback to original product image URL when Photon fails

hichamboushaba:
### Description
Fixes WOOMOB-2586

This PR fixes an issue where product images would fail to display if Photon-hosted URLs (`i0.wp.com`) were blocked or unavailable. It introduces a robust fallback mechanism so that the app gracefully loads the original image URL whenever the Photon request fails.

For the Compose components, `ProductThumbnail` now handles the fallback logic internally using a custom Coil extension (`rememberPhotonImageRequest`). This required replacing `Modifier.size()` with a direct `imageSize` parameter across all callers so that the exact dimensions can be fetched from the server.

A similar fallback approach was applied to the View system components using new Glide extensions in `PhotonImageExt.kt`.

### Test Steps
1. Open the app and navigate to a screen with product images (e.g., Orders list, Product list, Dashboard top performers).
2. Verify that product images load correctly and appear at the correct dimensions.
3. Simulate a failure for `i0.wp.com` (Photon URL) requests using Android Studio's Network Inspector. (Note: ApiFaker cannot be used for this as it does not intercept these requests).
4. Verify that the images gracefully fall back to the original URL and load successfully despite the Photon failure.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
@@ -33,7 +33,7 @@ fun rememberPhotonImageRequest(
 
     var currentUrl by remember(originalUrl, photonUrl) { mutableStateOf(photonUrl) }
 
-    return remember(currentUrl, context, configure) {
+    return remember(currentUrl, context) {
         ImageRequest.Builder(context)
             .data(currentUrl)
             .apply(configure)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
@@ -15,6 +15,7 @@ import org.wordpress.android.util.PhotonUtils
 /**
  * Coil/Compose extension for loading a Photon image with a fallback to the original URL.
  * Generates an ImageRequest that tries the Photon URL first, and if it fails, falls back to the original URL.
+ * If the URL is already a Photon URL, no fallback is applied since the "original" would be the same URL.
  */
 @Composable
 fun rememberPhotonImageRequest(
@@ -24,11 +25,7 @@ fun rememberPhotonImageRequest(
 ): ImageRequest {
     val context = LocalContext.current
     val photonUrl = remember(originalUrl, imageSizePx) {
-        if (originalUrl.isNotEmpty()) {
-            PhotonUtils.getPhotonImageUrl(originalUrl, imageSizePx, imageSizePx)
-        } else {
-            originalUrl
-        }
+        PhotonUtils.getPhotonImageUrl(originalUrl, imageSizePx, imageSizePx)
     }
 
     var currentUrl by remember(originalUrl, photonUrl) { mutableStateOf(photonUrl) }
@@ -39,7 +36,7 @@ fun rememberPhotonImageRequest(
             .apply(configure)
             .listener(
                 onError = { _, _ ->
-                    if (currentUrl != originalUrl && originalUrl.isNotEmpty()) {
+                    if (!originalUrl.isPhotonUrl() && currentUrl != originalUrl) {
                         currentUrl = originalUrl
                     }
                 }
@@ -58,6 +55,9 @@ fun RequestManager.loadPhotonUrlWithFallback(
         return this.load(originalUrl)
     }
     val photonUrl = PhotonUtils.getPhotonImageUrl(originalUrl, width, height, quality)
+    if (originalUrl.isPhotonUrl()) {
+        return this.load(photonUrl)
+    }
     return this.load(photonUrl).error(this.load(originalUrl))
 }
 
@@ -71,5 +71,12 @@ fun <T> RequestBuilder<T>.loadPhotonUrlWithFallback(
         return this.load(originalUrl)
     }
     val photonUrl = PhotonUtils.getPhotonImageUrl(originalUrl, width, height, quality)
+    if (originalUrl.isPhotonUrl()) {
+        return this.load(photonUrl)
+    }
     return this.load(photonUrl).error(this.clone().load(originalUrl))
+}
+
+private fun String.isPhotonUrl(): Boolean {
+    return contains("i0.wp.com") || contains("i1.wp.com") || contains("i2.wp.com")
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
@@ -1,9 +1,52 @@
 package com.woocommerce.android.extensions
 
 import android.graphics.drawable.Drawable
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+import coil.request.ImageRequest
 import com.bumptech.glide.RequestBuilder
 import com.bumptech.glide.RequestManager
 import org.wordpress.android.util.PhotonUtils
+
+/**
+ * Coil/Compose extension for loading a Photon image with a fallback to the original URL.
+ * Generates an ImageRequest that tries the Photon URL first, and if it fails, falls back to the original URL.
+ */
+@Composable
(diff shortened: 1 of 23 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/compose/component/ProductThumbnail.kt
```diff
@@ -7,33 +7,40 @@ import androidx.compose.runtime.Composable
...
+    val imageSizePx = with(LocalDensity.current) { imageSize.roundToPx() }
+    val imageRequest = rememberPhotonImageRequest(
+        originalUrl = imageUrl,
+        imageSizePx = imageSizePx
+    ) {
+        crossfade(true)
+        placeholder(placeHolderDrawableId)
+        fallback(fallbackDrawableId)
+        error(errorDrawableId)
+    }
```
hichamboushaba, line 36: With this change, all consumers that use ProductThumbnail will benefit from using Photon and from the fallback logic.

AdamGrzybkowski:
Hi @hichamboushaba 

Is there an extra setup required for this fallback to work? I did add some logs and the fallback URL used is `i0.wp.com` - here's the full log:
`Falling back to original URL https://i0.wp.com/adamatomicsite.wpcomstaging.com/wp-content/...`

This is from the Compose extension, but I don't see photos loaded in other places where XML is used.

hichamboushaba:
@AdamGrzybkowski taking a look. Can you tell me the exact line where you added the logs to try to replicate what you have?

AdamGrzybkowski:
@hichamboushaba Sure, this [line](https://github.com/woocommerce/woocommerce-android/blob/issue/WOOMOB-2586-refactor-photon-url/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt#L43)

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
```diff
@@ -0,0 +1,75 @@
...
+        if (originalUrl.isNotEmpty()) {
+            PhotonUtils.getPhotonImageUrl(originalUrl, imageSizePx, imageSizePx)
+        } else {
+            originalUrl
+        }
+    }
+
+    var currentUrl by remember(originalUrl, photonUrl) { mutableStateOf(photonUrl) }
+
+    return remember(currentUrl, context, configure) {
```
AdamGrzybkowski: `configure` is a lambda, meaning it is likely unstable and may cause extra recompositions. Have you verified this?

hichamboushaba:
> Is there an extra setup required for this fallback to work? I did add some logs and the fallback URL used is i0.wp.com - here's the full log:
Falling back to original URL https://i0.wp.com/adamatomicsite.wpcomstaging.com/wp-content/...
> 
> This is from the Compose extension, but I don't see photos loaded in other places where XML is used.

I can't reproduce the issue @AdamGrzybkowski, with a simple rule to force 500 response for `i0.wp.com` in AS app inspection:
Screenshot 2026-04-20 at 10 40 54

I can confirm the fallback is working:

https://github.com/user-attachments/assets/77f8fc40-a30f-4177-be7b-c51c9cd27b8a

----
I also tried with logging:

<details>
<summary>Patch</summary>

```patch
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt	(revision 3c4a74ed8e7bf1999bfc45caf6eb5298e13c5b79)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt	(date 1776678536891)
@@ -10,6 +10,7 @@
 import coil.request.ImageRequest
 import com.bumptech.glide.RequestBuilder
 import com.bumptech.glide.RequestManager
 import org.wordpress.android.util.PhotonUtils
 
 /**
@@ -40,6 +41,7 @@
             .listener(
                 onError = { _, _ ->
                     if (currentUrl != originalUrl && originalUrl.isNotEmpty()) {
+                        WooLog.d(WooLog.T.UTILS, "Retrying with original URL: $originalUrl, photonUrl: $photonUrl")
                         currentUrl = originalUrl
                     }
                 }
@@ -58,7 +60,12 @@
         return this.load(originalUrl)
     }
     val photonUrl = PhotonUtils.getPhotonImageUrl(originalUrl, width, height, quality)
-    return this.load(photonUrl).error(this.load(originalUrl))
+    return this.load(photonUrl).error(
+        this.load(originalUrl)
+            .also {
+                WooLog.d(WooLog.T.UTILS, "Retrying with original URL: $originalUrl, photonUrl: $photonUrl")
+            }
+    )
 }
 
 fun <T> RequestBuilder<T>.loadPhotonUrlWithFallback(
@@ -71,5 +78,10 @@
         return this.load(originalUrl)
     }
     val photonUrl = PhotonUtils.getPhotonImageUrl(originalUrl, width, height, quality)
-    return this.load(photonUrl).error(this.clone().load(originalUrl))
+    return this.load(photonUrl).error(
+        this.clone().load(originalUrl)
+            .also {
+                WooLog.d(WooLog.T.UTILS, "Retrying with original URL: $originalUrl, photonUrl: $photonUrl")
+            }
+    )
 }
```

</details>

And I can see the correct logs:
```
Retrying with original URL: https://hichamwootest.blog/wp-content/uploads/2023/11/belt-2-450x450.jpg?crop=1, photonUrl: https://i0.wp.com/hichamwootest.blog/wp-content/uploads/2023/11/belt-2-450x450.jpg?strip=info&quality=65&resize=110,110&ssl=1
```

For Compose specifically, I added:
```
            .diskCachePolicy(CachePolicy.DISABLED)
            .memoryCachePolicy(CachePolicy.DISABLED)
```
To force loading from network and getting the logs all the time, but this is just for logs, the logic works with cache correctly.

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/PhotonImageExt.kt
hichamboushaba: Good point, I checked the only call site that uses it now (ProductThumbnail), and we pass only stable items to the lambda there, so Compose should be able to memoize it.
But now looking more into this, I think using `configure` as a key is not needed here, I don't think we'll have a case where the caller site might modify the `configure` lambda during runtime, I'll remove it.

AdamGrzybkowski:
Thanks for checking @hichamboushaba! I've tested this on multiple sites, and it seems to be a site-specific issue - some work, some default back to the same `i0.wp.com` URL. I have invited you to one that has an issue if you want to check it.

hichamboushaba:
@AdamGrzybkowski I think I understand what's happening.
Can you check the value of the option `Speed up image load times` in your site (in `wp-admin/admin.php?page=jetpack#/performance`):

Screenshot 2026-04-20 at 12 26 35

If it's enabled, then it could explain the issue, your site is already using photon URLs, so the fallback won't help.

I'll add a logic to avoid using the fallback when the original URL is already a photon URL, this won't help the user, but at least we won't waste more requests.

AdamGrzybkowski:
It is enabled. 

Screenshot 2026-04-20 at 13 32 58

hichamboushaba:
OK, this explains it. I added one last commit to avoid wasting time on the fallback in this situation. Ready for review now @AdamGrzybkowski.

AdamGrzybkowski (approved):
I think we could add a test case for the last commit, other than that - :shipit: 

Thanks for the changes!

hichamboushaba:
> I think we could add a test case for the last commit, other than that - :shipit:

Unit testing the behavior will be tricky, since it uses Android components heavily, and will require a way to hook into Glide/Coil's internals.
If you have some ideas please let me know, otherwise let's ship it as it's, WDYT?

hichamboushaba:
@AdamGrzybkowski I'm going to merge this now, if you have some ideas for unit tests, please let me know and I'll add them in a separate PR.

AdamGrzybkowski:
You're right 🤦 I think that's fine!

PR #15647, opened 2026-04-10
[WOOMOB-2628] Prevent bypass after failed role check

hichamboushaba:
### Description
Fixes WOOMOB-2628
This prevents the app from bypassing a failed role check after restart by only persisting the selected site once user-role verification succeeds. It also aligns WP API v2 endpoint strings with ApiFaker, updates the affected mocks/tests, and replaces the generic role-fetch snackbar with more specific copy.

### Test Steps
1. Simulate a failure for the endpoint `/wp/v2/users/me`.
2. Log in using WordPress.com.
3. Confirm that a Snackbar error is shown on the site picker.
4. Restart the app.
5. Confirm you are still blocked at the site picker level

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
@@ -542,7 +542,7 @@ class SitePickerViewModel @Inject constructor(
                             triggerEvent(SitePickerEvent.NavigateToMainActivityEvent)
                         },
                         onFailure = {
-                            triggerEvent(ShowSnackbar(R.string.error_generic))
+                            triggerEvent(ShowSnackbar(R.string.user_role_access_error_fetch_failed))
                         }
                     )
                 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
@@ -534,11 +534,11 @@ class SitePickerViewModel @Inject constructor(
                     userEligibilityFetcher.fetchUserInfo(siteVerificationModel.siteModel).fold(
                         onSuccess = {
                             selectedSite.set(siteVerificationModel.siteModel)
-                            sitePickerViewState = sitePickerViewState.copy(isProgressDiaLogVisible = false)
-
                             trackLoginEvent(currentStep = UnifiedLoginTracker.Step.SUCCESS)
                             appPrefsWrapper.removeLoginSiteAddress()
                             registerDevice(RegisterDevice.Mode.IF_NEEDED)
+
+                            sitePickerViewState = sitePickerViewState.copy(isProgressDiaLogVisible = false)
                             triggerEvent(SitePickerEvent.NavigateToMainActivityEvent)
                         },
                         onFailure = {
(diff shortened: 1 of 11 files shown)


libs/fluxc-annotations/src/main/java/org/wordpress/android/fluxc/annotations/endpoint/WPAPIEndpoint.java
```diff
@@ -2,7 +2,7 @@
 
 public class WPAPIEndpoint {
     private static final String WPCOM_REST_PREFIX = "https://public-api.wordpress.com";
-    private static final String WPAPI_PREFIX_V2 = "wp/v2";
+    private static final String WPAPI_PREFIX_V2 = "/wp/v2";
```
hichamboushaba, line 5: This change is unrelated to the PR, it's just to bring consistency to how we construct URL paths (check `WCWPAPIEndpoint` for example).
This change doesn't have any behavioral impact, given we use the extension `slashJoin` which trims any additional slashes, but it fixes an issue with how ApiFaker matches the endpoints.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -530,28 +530,28 @@ class SitePickerViewModel @Inject constructor(
...
+                    userEligibilityFetcher.fetchUserInfo(siteVerificationModel.siteModel).fold(
                         onSuccess = {
-                            sitePickerViewState = sitePickerViewState.copy(isProgressDiaLogVisible = false)
-
+                            selectedSite.set(siteVerificationModel.siteModel)
                             trackLoginEvent(currentStep = UnifiedLoginTracker.Step.SUCCESS)
                             appPrefsWrapper.removeLoginSiteAddress()
                             registerDevice(RegisterDevice.Mode.IF_NEEDED)
+
+                            sitePickerViewState = sitePickerViewState.copy(isProgressDiaLogVisible = false)
```
hichamboushaba, line 541: This change is unrelated to this fix, but I noticed it while testing PNs: we hide the loading dialog and then wait a bit before navigating to the main screen. I thought this was something I could squeeze in here.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -248,6 +248,7 @@
     <string name="login_troubleshooting_tips">Read our troubleshooting tips</string>
     <string name="login_no_wpcom_account_found">Your email isn\'t used with a WordPress.com account.</string>
     <string name="user_role_access_error_user_roles_null">This app supports only Administrator and Shop Manager user roles. We were unable to fetch user roles for your account. Please contact support.</string>
+    <string name="user_role_access_error_fetch_failed">We couldn\'t verify your role on this store. Please try again. If the problem persists, contact support.</string>
```
AdamGrzybkowski, line 251: Are we aligned on iOS with this message?

AdamGrzybkowski (approved):
Restarting the app properly showed the site picker again. LGTM 👍

WooCommerce/src/main/res/values/strings.xml
hichamboushaba: I don't think we are aligned in the handling here; they have a more generic alert that they use when signing it to a site fails, and I suspect they are planning to reuse it.

cc @itsmeichigo @adborbas if you plan on customizing the error messages when user role can't be checked, please check the wording we used here to be aligned.

PR #15648, opened 2026-04-10
[WOOMOB-2646] Add FedEx to supported shipping features

irfano:
### Description
Fixes WOOMOB-2646

This PR adds `fedex` to `features_supported_by_client` in Woo Shipping requests for:
- packages
- label rates
- label purchase

This PR intentionally does not remove the local FedEx feature flag yet. It only updates the request payload so backend gating can recognize FedEx support when the feature is enabled locally.

### Test Steps
1. Enable woo_shipping_fedex flag.
2. Log in to a site that can purchase shipping labels and is connected to the staging connect-server with FedEx enabled.
3. Open an order with shippable products.
4. Start the Create Shipping Label flow.
5. Verify the FedEx package section is shown on the package selection step.
6. Continue to the rates step.
7. Verify FedEx rates are shown.

### Images/gif

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/packages/networking/WooShippingLabelPackageRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/packages/networking/WooShippingLabelPackageRestClient.kt
@@ -14,7 +14,11 @@ class WooShippingLabelPackageRestClient @Inject constructor(
     ): WooPayload<PackageResponse> {
         return wooNetwork.executeGetGsonRequest(
             site = site,
-            path = "$URL?$UPSDAP_FEATURE_QUERY&$FEDEX_FEATURE_QUERY",
+            path = URL,
+            params = mapOf(
+                "features_supported_by_client[0]" to UPSDAP_FEATURE,
+                "features_supported_by_client[1]" to FEDEX_FEATURE,
+            ),
             clazz = PackageResponse::class.java,
         ).toWooPayload()
     }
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/packages/networking/WooShippingLabelPackageRestClient.kt
```diff
@@ -14,8 +14,7 @@ class WooShippingLabelPackageRestClient @Inject constructor(
     ): WooPayload<PackageResponse> {
         return wooNetwork.executeGetGsonRequest(
             site = site,
-            path = URL,
-            params = mapOf("features_supported_by_client[]" to UPSDAP_FEATURE),
+            path = "$URL?$UPSDAP_FEATURE_QUERY&$FEDEX_FEATURE_QUERY",
```
AdamGrzybkowski: I would suggest reverting to the previous 
```suggestion
            path = URL,
            params = mapOf(
                "features_supported_by_client[0]" to UPSDAP_FEATURE,
                "features_supported_by_client[1]" to FEDEX_FEATURE,
            ),
```
This way we don't have to hardcode the encoding and we will satisfy the unique key for the map with the indexed values.

AdamGrzybkowski (approved):
Works as described 👍 I would strongly suggest removing the hardcoded encoding though, as mentioned in the comment.

irfano:
Good suggestion! I applied it. Thanks!

AdamGrzybkowski:
FYI: The above suggested diff was not enough to make the project compile, you still need to bring back the `UPSDAP_FEATURE` and add a new one for FedEx.

PR #15662, opened 2026-04-14
[WOOMOB-2666] Adapt eligibility and splash screens for phone

kidinov:
### Description

Fixes [WOOMOB-2666](https://linear.app/a8c/issue/WOOMOB-2666).

Adapts the **POS eligibility screen** and the **splash "Syncing catalog" screen** so they render properly on a phone layout while keeping the existing tablet layout unchanged. Also pins the buttons to the bottom of the screen on both eligibility and the shared `WooPosErrorScreen` (used by Splash SyncFailed and other error surfaces), matching the pattern used elsewhere in POS.

- Eligibility screen: suggestion text uses `Modifier.adaptiveContentWidth()`. Primary/secondary buttons pinned to the bottom via `Box { Column(Center) + Column(BottomCenter) }`, using `Modifier.wooPosFullScreenActionButton()` (100% phone / 50% tablet).
- Splash **Syncing catalog**: the 160dp loading indicator is now `WooPosComponentSize.XLarge.value` so it shrinks on phone.
- Splash **SyncFailed** / `WooPosErrorScreen`: content stays vertically centered; primary/secondary buttons pinned to the bottom. Same `Box { Column(Center) + Column(BottomCenter) }` layout. Affects every error screen that uses `WooPosErrorScreen`.

> **Base branch:** `woomob-2661-woo-pos-phone-adaptive-design-system-with-breakpoint-based`. This PR depends on the button/token/breakpoint/preview work there.

### Test Steps

1. Open the app on a **tablet** (e.g. Pixel Tablet emulator).
2. Force eligibility state or open the POS → verify eligibility and splash syncing look right. Buttons should sit at the bottom of the screen.
3. Run the app on a **phone** layout (Pixel 6 emulator, or 411x891 Compose Preview now renders as Phone).
4. Verify the eligibility screen uses full-width suggestion text and full-width buttons anchored to the bottom.
5. Verify the splash syncing indicator looks smaller and proportional on phone.
6. Trigger any error flow that uses `WooPosErrorScreen` (e.g. simulate sync failure) — buttons should be anchored at the bottom, content centered above.

#### Eligibility — retryable

| Phone (before) | Phone (after) | Tablet (before) | Tablet (after) |
|---|---|---|---|
|  |  |  |  |

#### Eligibility — plan upgrade

| Phone (before) | Phone (after) | Tablet (before) | Tablet (after) |
|---|---|---|---|
|  |  |  |  |

#### Splash — loading

| Phone (before) | Phone (after) | Tablet (before) | Tablet (after) |
|---|---|---|---|
|  |  |  |  |

#### Splash — syncing

| Phone (before) | Phone (after) | Tablet (before) | Tablet (after) |
|---|---|---|---|
|  |  |  |  |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
```diff
@@ -35,17 +35,15 @@ fun WooPosErrorScreen(
...
-    Column(
+    Box(
         modifier = modifier
             .fillMaxSize()
             .clip(RoundedCornerShape(WooPosCornerRadius.Medium.value))
-            .padding(WooPosSpacing.XLarge.value),
-        horizontalAlignment = Alignment.CenterHorizontally,
-        verticalArrangement = Arrangement.Center,
+            .padding(WooPosSpacing.XLarge.value)
     ) {
```
malinajirka, line 45: 📓 `WooPosErrorScreen` is called from ~10 places beyond the splash SyncFailed screen — `WooPosTotalsScreen` (3 call sites for totals/coupon/product errors), `WooPosOrdersScreen`, `WooPosBookingsScreen`, `WooPosCouponsScreen`, `WooPosCouponsSearchScreen`, `WooPosItemsSearchScreen`, `WooPosProductsScreen`, `WooPosVariationsScreen`. Several of those render inside smaller container cards (e.g. totals panel), not a full screen.

Switching from the previous `Column(verticalArrangement = Center)` (content + buttons clustered together, centered) to `Box { Column(Center) + Column(BottomCenter) }` changes the visual rhythm on every caller, not just splash. This is a noticeable design shift, I'm personally not sure about - I think someone with a design-sensitive eye might not like it.

Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
```diff
@@ -70,7 +68,12 @@ fun WooPosErrorScreen(
                 style = WooPosTypography.BodyLarge,
                 textAlign = TextAlign.Center,
             )
-            Spacer(modifier = Modifier.height(WooPosSpacing.XXLarge.value))
+        }
+
+        Column(
+            modifier = Modifier.align(Alignment.BottomCenter),
```
malinajirka: 🔍 np: on a very short viewport (e.g. landscape phone with the IME up, or large font scale), the centered content `Column` and the `BottomCenter` button `Column` can silently overlap since neither is height-constrained. A single `Column` with a `Spacer(Modifier.weight(1f))` between the content and the buttons would let the buttons push the content up instead of drawing on top of it. Not blocking — just flagging.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
malinajirka: Screenshot_1776235517
Screenshot_1776235525
Screenshot_1776235589

malinajirka (changes requested):
LGTM overall, I've left a few comments.


The `Syncing catalog` text feels too close to the progress indicator, wdyt?
image

kidinov:
@malinajirka 

Good catch. I think this was the case even before, but I refactored the indicator + title into a centered `Column` with a `Spacer(WooPosSpacing.XLarge)` between them, so the gap is consistent regardless of indicator size. Fixed in 4d086bf.

| Tablet | Phone |
|---|---|
| tablet | phone |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
kidinov: Fair point — reverted `WooPosErrorScreen` to the trunk layout in ab520aa so other call sites are unchanged. Splash SyncFailed now matches its pre-PR look. Side benefit: also resolves the overlap risk you flagged below.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosErrorScreen.kt
kidinov: Resolved by the revert in ab520aa — single-Column layout doesn't have the unconstrained-children overlap.

malinajirka (approved):
LGTM

PR #15656, opened 2026-04-13
[WOOMOB-2661] Adaptive design system with breakpoint-based scaling for POS on phones

kidinov:
### Description

Fixes WOOMOB-2661

Adds breakpoint-based scaling to the POS design system so the UI adapts on phones while the tablet experience stays unchanged. Three tiers by `longestSide`: phone `< 880dp`, small tablet `< 1200dp`, full tablet `≥ 1200dp`. All scaling helpers return 1.0x on tablet by design.

### What changed

- **Adaptive primitives** in `WooPosSizes.kt`:
  - `Dp.toAdaptiveComponentSize()` aggressive scale (0.75x / 0.9x / 1.0x)
  - `Dp.toAdaptiveIconSize()` mild scale (0.9x / 0.95x / 1.0x)
  - `Modifier.adaptiveContentWidth()` — `fillMaxWidth` / 2/3 / 3/5 (replaces fixed 604dp panels)
  - `WooPosCornerRadius`, `WooPosElevation`, `WooPosTypography` now resolve per tier — tablet values match the prior fixed values
  - Breakpoint read via `LocalWindowInfo.current.containerSize` (API-36 safe)
- **Design-token enums**:
  - `WooPosIconSize` — XSmall 16 / Small 24 / Medium 32 / Large 40 / XLarge 48 (no scaling)
  - `WooPosComponentSize` — XSmall 56 / Small 80 / Medium 96 / Large 112 / XLarge 160 / XXLarge 256 (auto-scale)
- **Detekt rules**:
  - New `WooPosDesignSystemComponentSizeUsageRule` (Modifier.size/height/width/heightIn/widthIn/...)
  - New `WooPosDesignSystemElevationUsageRule` (shadow + `*Elevation()` calls)
  - Extended `WooPosDesignSystemSpacingUsageRule` to cover `PaddingValues(...)` and `Arrangement.spacedBy(...)`
- **Migration** — ~90 call sites across ~30 POS screens moved from raw dp to helpers/tokens. Baseline has zero `WooPosDesignSystem*` entries.

### Tablet regression is the main review concern

Sanity-check these on your tablet against `trunk`:

1. Products + cart two-pane
2. Cart with one or more items
3. Totals / checkout
4. Card payment — reader connection, cash payment, payment success (watch for collapsed shimmers)
5. Orders master-detail
6. Bookings master-detail
7. Settings

Phone check is optional: build with `WOO_POS_PHONE` feature flag enabled and confirm POS feels appropriately compact.

[...]

PR #15665, opened 2026-04-14
[WOOMOB-2707] Phone home screen NavHost skeleton

kidinov:
### Description

Closes [WOOMOB-2707](https://linear.app/a8c/issue/WOOMOB-2707). Part of [Woo POS on Phones](https://linear.app/a8c/project/woo-pos-on-phones-9a932b9546b8) (parent: [WOOMOB-2656](https://linear.app/a8c/issue/WOOMOB-2656)).

Adds the **phone home screen skeleton**: a new `WooPosHomePhoneScreen` that hosts an inner `NavHost` with three destinations — `phone_products`, `phone_cart`, `phone_totals`. The outer `homeScreen(...)` now branches on `Context.isWooPosPhoneLayout()` and picks phone or tablet.

> [!IMPORTANT]
> The most important review goal is validating there is **no regression on the tablet layout**. All POS screens on tablet should look and behave exactly as before.

### Base branch

Target: **`woomob-2661-woo-pos-phone-adaptive-design-system-with-breakpoint-based`** (PR [#15656](https://github.com/woocommerce/woocommerce-android/pull/15656)) — not `trunk`.

### What changed (6 commits)

1. **Inject POS home ViewModels via composable parameter** — `WooPosItemsScreen`, `WooPosCartScreen`, `WooPosTotalsScreen`, `WooPosFloatingToolbar` now take `viewModel: X = hiltViewModel()`. Tablet call sites unchanged (default preserves old behaviour).
2. **Add `WooPosHomePhoneScreen`** — inner NavHost, `screenPositionState` → route sync for Cart ↔ Totals, barcode listener, debug-only "Cart (debug)" FAB on Products.
3. **Route POS home to phone screen on phone layout** — branch inside `homeScreen(...)` on `isWooPosPhoneLayout()`.
4. **Share POS home dialogs** — new `WooPosHomeDialogs`, used by both tablet and phone.
5. **Extract POS home root container modifier** — `Modifier.wooPosHomeRootContainer(state, onHomeUIEvent)` wraps `fillMaxSize + background + listenForBarcodes` in one place.
6. **Share POS home product/cart/totals panes** — `WooPosHomeProductsPane`, `WooPosHomeCartPane`, `WooPosHomeTotalsPane`. Preview-mode guard baked in; phone previews no longer crash.

### Out of scope (later tickets)

- **WOOMOB-2708** — phone-adapted Products screen wrapper
- **WOOMOB-2709** — phone-adapted Cart / Totals wrappers
- **WOOMOB-2710** — back press handling + empty-cart auto-nav
- **WOOMOB-2657** — persistent bottom "Cart (n) / Checkout / Cash" button (replaces the temporary debug FAB)

### Debug FAB

Phone Products shows a **"Cart (debug)"** `ExtendedFloatingActionButton` in the bottom-right. Only compiled into debug builds (`PackageUtils.isDebugBuild()`). Temporary — lands in trunk with this PR, removed when **WOOMOB-2657** (bottom button) lands.

### Test Steps

> [!TIP]
> Tablet regression check is the priority. Phone is gated by the `WOO_POS_PHONE` feature flag (on by default in debug builds).

**Tablet (Medium Tablet API 33 or any real tablet)**

1. Build and install from this branch.
2. Open POS — products grid on the left, cart in the middle, ability to open totals. Confirm the three-pane layout is unchanged.
3. Add items to the cart, open totals, try cash payment, dismiss dialogs. Nothing should look different from `woomob-2661` baseline.

**Phone (Pixel 9 Pro XL or similar)**

1. Build and install from this branch (WOO_POS_PHONE flag is on by default in debug builds).
2. Open POS — Products is full-screen.
3. Tap the "Cart (debug)" FAB in the bottom-right → Cart full-screen.
4. Tap "Check out" inside Cart → Totals full-screen.
5. Complete cash payment → goes back to Products.

#### Tablet — no regression (before = [`woomob-2661`](https://github.com/woocommerce/woocommerce-android/pull/15656), after = this PR)

| Screen | Before | After |
| --- | --- | --- |
| Products + empty cart |  |  |
| Cart with items |  |  |
| Totals |  |  |
| Overflow menu |  |  |

#### Phone — broken 3-pane before, linear flow after

| Screen | Before (tablet layout on phone — unusable) | After (phone NavHost flow) |
| --- | --- | --- |
| POS landing |  |  |
| After adding items |  |  |
| Cart / Totals |  |   |

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosHomePhoneScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosHomePhoneScreen.kt
@@ -1,0 +1,196 @@
+package com.woocommerce.android.ui.woopos.home.phone
+
+import androidx.activity.compose.BackHandler
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.ExtendedFloatingActionButton
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
+import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.rememberNavController
+import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosExitConfirmationDialog
+import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosText
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosSpacing
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
+import com.woocommerce.android.ui.woopos.common.composeui.modifier.listenForBarcodes
+import com.woocommerce.android.ui.woopos.home.WooPosHomeState
+import com.woocommerce.android.ui.woopos.home.WooPosHomeUIEvent
+import com.woocommerce.android.ui.woopos.home.WooPosHomeViewModel
+import com.woocommerce.android.ui.woopos.home.cart.WooPosCartScreen
+import com.woocommerce.android.ui.woopos.home.cart.WooPosCartViewModel
+import com.woocommerce.android.ui.woopos.home.items.WooPosItemsScreen
+import com.woocommerce.android.ui.woopos.home.items.WooPosItemsViewModel
+import com.woocommerce.android.ui.woopos.home.totals.WooPosTotalsScreen
+import com.woocommerce.android.ui.woopos.home.totals.WooPosTotalsViewModel
+import com.woocommerce.android.ui.woopos.scanningsetup.WooPosScanningSetupDialog
+import com.woocommerce.android.util.PackageUtils
+import org.wordpress.android.util.ToastUtils
+
+private const val PHONE_PRODUCTS_ROUTE = "phone_products"
+private const val PHONE_CART_ROUTE = "phone_cart"
+private const val PHONE_TOTALS_ROUTE = "phone_totals"
+
+@Composable
+fun WooPosHomePhoneScreen(
+    isPaymentCompletedViaCash: Boolean,
+    viewModel: WooPosHomeViewModel,
+) {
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosHomePhoneScreen.kt
```diff
@@ -0,0 +1,158 @@
...
+                    }
+                }
+            }
+        }
+        previousState = state.screenPositionState
+    }
+
+    BackHandler {
+        onHomeUIEvent(WooPosHomeUIEvent.SystemBackClicked)
+    }
```
malinajirka, line 118: 💡 I think the phone `BackHandler` should pop the inner NavHost first and only fall through to `SystemBackClicked` when the NavHost stack is empty.

Walking through what happens today from `phone_products` → (debug FAB) → `phone_cart` → system back:
- The FAB calls `navController.navigate(PHONE_CART_ROUTE)` directly, so `screenPositionState` stays `Cart` the whole time.
- System back → `handleSystemBackClicked()` in the VM, which for `Cart` opens the `ExitConfirmationDialog` (`WooPosHomeViewModel.kt:116`).
- So the user ends up on the Exit dialog when they expected to land back on products.

malinajirka (commented):
Thanks @kidinov! 

I've run a side-by-side comparison of most of the screens/states (using AI) and noticed one small difference on the payment success screen (notice the different button width) - I'm not sure if it's related to this PR though:


Image


More important bug is this weird one where the POS Splash screen goes crazy in portrait mode on tablet - I haven't tested on a hw device, so might be an emulator glitch. But the production version of the app works as expected.

https://github.com/user-attachments/assets/b5e057fa-5aaa-4dbb-81e0-3bab7aab30fd

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosHomePhoneScreen.kt
kidinov: Good catch, thanks! Fixed in 7b21902 - now the `BackHandler` first tries `navController.popBackStack()` and only forwards to `SystemBackClicked` when the inner NavHost stack is empty.

malinajirka:
> Both are worth fixing, but I'd prefer to keep this PR focused on the skeleton. OK to track them as separate tickets?

💯 Sure thing.

> Splash portrait-on-tablet glitch - agreed it's weird. Production works, so likely an emulator quirk with orientation changes while the POS activity is mid-launch. I'll repro on a hardware device and file a ticket if confirmed.

The weird part was that I tested a different build on the same device and it wasn't reproducible -> that's why I assumed the issue is introduced in this PR or the parent branch. Either case, definitely worth separating into a standalone PR.

malinajirka (approved):

kidinov:
@malinajirka I actually checked it here and changed the width close to how that was 887a048c2d79ecbc502705e1cec74ba63970ab57

While validated the second issue and it looks like that it's emulator only - on a real device it works fine

PR #15666, opened 2026-04-14
[WOOMOB-2681] Fix Woo notification tap routing and cleanup

hichamboushaba:
### Description
Fixes WOOMOB-2681

This updates Woo-driven push notifications to always include the notification payload in the tap intent so they route to the correct destination even when they do not have a WP.com `remoteNoteId`.

It also fixes the remaining tray cleanup gap for Woo notifications by removing the tapped child notification and, when that child was the last active notification in its exact group (`channelType + remoteSiteId`), removing the group summary notification as well.

Changes in this PR:
- always include `FIELD_REMOTE_NOTIFICATION` in `WooNotificationBuilder.getResultIntent()`
- add summary-aware cleanup in `NotificationMessageHandler` for single tapped notifications
- remove the group summary only when the tapped child was the last active notification in its exact group
- update `MainActivityViewModel` to use the new cleanup path for single push opens
- add regression tests for last-child summary removal, remaining-sibling behavior, and exact-group scoping by store

### Test Steps
1. Prepare a test site following paaHJt-9VE-p2
2. Trigger a push notification.
3. Tap the notification.
4. Confirm the app opens the expected destination instead of falling back to My Store.
5. Confirm the tapped notification is removed from the tray.
6. Confirm the group summary is also removed when that tapped notification was the last active notification in its group.
7. Trigger two Woo notifications in the same group and tap one of them.
8. Confirm the tapped child is removed but the group summary remains because another child is still active.
9. Trigger notifications for a different store and confirm tapping one group does not incorrectly preserve or remove the other store's summary.
10. Trigger a standard notification with a non-zero `remoteNoteId` and confirm tap routing still works.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
@@ -268,6 +268,24 @@ class NotificationMessageHandler @Inject constructor(
     }
 
     @Synchronized
+    fun removeTappedNotificationAndSummaryIfNeeded(localPushId: Int, notification: Notification) {
+        with(notificationBuilder) {
+            cancelNotification(localPushId)
+
+            val hasRemainingChildrenInGroup = getActiveNotifications().any {
+                !it.isGroupSummary &&
+                    it.id != localPushId &&
+                    it.channelType == notification.channelType.name &&
+                    it.remoteSiteId == notification.remoteSiteId
+            }
+
+            if (!hasRemainingChildrenInGroup) {
+                cancelNotification(notification.getGroupPushId())
+            }
+        }
+    }
+
+    @Synchronized
     fun removeNotificationsOfTypeFromSystemsBar(type: NotificationChannelType, remoteSiteId: Long) {
         with(notificationBuilder) {
             getActiveNotifications()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/WooNotificationBuilder.kt
@@ -80,9 +80,7 @@ class WooNotificationBuilder @Inject constructor(
         flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK
         putExtra(MainActivity.FIELD_OPENED_FROM_PUSH, true)
         putExtra(MainActivity.FIELD_PUSH_ID, pushId)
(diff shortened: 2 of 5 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
```diff
@@ -267,6 +267,24 @@ class NotificationMessageHandler @Inject constructor(
...
+            val hasRemainingChildrenInGroup = getActiveNotifications().any {
+                !it.isGroupSummary &&
+                    it.id != localPushId &&
+                    it.channelType == notification.channelType.name &&
+                    it.remoteSiteId == notification.remoteSiteId
+            }
+
+            if (!hasRemainingChildrenInGroup) {
+                cancelNotification(notification.getGroupPushId())
+            }
```
hichamboushaba, line 284: This issue touches only Review notifications, and it wasn't needed for WPCom, because later in the flow, we call `removeNotificationByRemoteIdFromSystemsBar` when marking the review as read.

For order notifications, we call `removeNotificationsOfTypeFromSystemsBar`, so all order-related notifications are deleted when tapping one. I'm not sure if this is a good behavior, WDYT? But anyway, we need to address it in separate PR if needed, as I'm now focusing only on fixing issues related to the new system.

AdamGrzybkowski:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
AdamGrzybkowski: >so all order-related notifications are deleted when tapping one. I'm not sure if this is a good behavior, WDYT?

All, like all? Even from unrelated orders?

AdamGrzybkowski (commented):
All works as described :shipit:

AdamGrzybkowski (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NotificationMessageHandler.kt
hichamboushaba: [Yes](https://github.com/woocommerce/woocommerce-android/blob/dcea847100d8c38ee67cb61e344a158dab9a7677/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt#L783-L785), we can discuss that behavior async, and see if we want to modify it.

PR #15667, opened 2026-04-14
[WOOMOB-2675] Fix logout push cleanup and loading state

hichamboushaba:
### Description
Fixes WOOMOB-2675

This PR addresses three related bugs in the logout flow and adds a loading state to the App Settings screen during logout.

#### 1. Fix empty-body push delete response (`PushNotificationsRestClient`)
Deleting push tokens was failing because of a NPE, this was caused by how we parsed the response.
The WooCommerce push token DELETE endpoint returns a JSON body containing `success`. The client now decodes that response and surfaces it through the store as a normal `WooResult<Unit>` success/failure instead of treating the endpoint like a no-body delete.

#### 2. Fix logout waits for push cleanup (`PushNotificationRepository` + `AccountRepository`)
- Previously, the WpCom and App Passwords logout paths each called a different unregister method (`unregisterDeviceFromAllPushes` / `unregisterWooCoreTokensFromServer`), and App Passwords logout didn't unregister the WpCom token at all
- Unified into a single `unregisterDeviceFromPushNotifications()` called before the account type branch — both paths now await full push cleanup before credentials are torn down
- The WpCom unregistration is now guarded by `isWpComPushRegistered()` to avoid unnecessary calls
- `getWooCommerceSites()` is now dispatched on the IO dispatcher (it's a DB call that was previously run on whatever dispatcher the caller used)
- Both parallel cleanup jobs are now awaited with `awaitAll()` for correctness

 
- Added a `LogoutState` enum (`IDLE`, `IN_PROGRESS`, `PENDING_FINISH`) to the presenter to track logout across configuration changes
- `AppSettingsModule` scope changed from `ActivityScoped` → `ActivityRetainedScoped` so the presenter (and its coroutine scope + state) survives screen rotation
- `takeView()` restores UI state on re-attach: shows the loading dialog if logout is still in progress, or immediately calls `finishLogout()` if it already completed while detached
- `dropView()` now takes the view instance and only clears the reference if it matches the current view
- `logout()` is idempotent — re-tapping while in progress just re-shows the dialog
- `AppSettingsActivity` implements `showLogoutProgressDialog()` / `hideLogoutProgressDialog()` using `CustomProgressDialog`; dialog is dismissed in `onDestroy()` to avoid window leaks

### Test Steps
1. Prepare a site for Woo PN notifications by following paaHJt-9VE-p2
2. Sign in, open **App Settings → Account**, tap **Log Out** and confirm — a "Logging out…" progress dialog should appear immediately.
3. Verify the app transitions to the login screen once logout completes and the dialog is dismissed.
4. Verify `Woo Core push token deleted for site ...` was logged in logcat. 
5. Repeat steps 2–4 with an Application Passwords site.

### Images/gif

https://github.com/user-attachments/assets/b3c2e317-07f3-4be2-9e6b-6b25e18a56ff

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
@@ -31,6 +31,7 @@ import com.woocommerce.android.ui.main.AppBarStatus
 import com.woocommerce.android.ui.prefs.MainSettingsFragment.AppSettingsListener
 import com.woocommerce.android.util.AnalyticsUtils
 import com.woocommerce.android.util.parcelable
+import com.woocommerce.android.widgets.CustomProgressDialog
 import dagger.android.DispatchingAndroidInjector
 import dagger.hilt.android.AndroidEntryPoint
 import org.wordpress.android.login.LoginMode
@@ -61,6 +62,7 @@ class AppSettingsActivity :
     @Inject lateinit var statsWidgetUpdaters: WidgetUpdater.StatsWidgetUpdaters
 
     private var isBetaOptionChanged = false
+    private var progressDialog: CustomProgressDialog? = null
 
     private lateinit var binding: ActivityAppSettingsBinding
     private var toolbar: Toolbar? = null
@@ -138,7 +140,8 @@ class AppSettingsActivity :
     }
 
     override fun onDestroy() {
-        presenter.dropView()
+        hideLogoutProgressDialog()
+        presenter.dropView(this)
         super.onDestroy()
     }
 
@@ -171,6 +174,7 @@ class AppSettingsActivity :
     }
 
     override fun finishLogout() {
+        hideLogoutProgressDialog()
         notificationMessageHandler.removeAllNotificationsFromSystemsBar()
         statsWidgetUpdaters.updateTodayWidget()
 
@@ -181,6 +185,30 @@ class AppSettingsActivity :
 
         startActivity(intent)
         finish()
+    }
+
+    override fun showLogoutProgressDialog() {
+        val existingDialog = supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog
+        if (existingDialog != null) {
+            progressDialog = existingDialog
+            progressDialog?.isCancelable = false
+            return
+        }
+
+        progressDialog = CustomProgressDialog.show(
+            getString(R.string.settings_logout_dialog_title),
+            getString(R.string.settings_logout_dialog_message)
+        ).also {
+            it.show(supportFragmentManager, CustomProgressDialog.TAG)
+        }
+        progressDialog?.isCancelable = false
+    }
+
+    override fun hideLogoutProgressDialog() {
+        val dialog = progressDialog
+            ?: supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog
+        dialog?.dismissAllowingStateLoss()
+        progressDialog = null
     }
 
     override fun confirmLogout() {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
@@ -194,7 +194,7 @@ class AppSettingsActivity :
             ?.dismissAllowingStateLoss()
 
         progressDialog = CustomProgressDialog.show(
-            getString(R.string.settings_logout_dialog_title),
+            "",
             getString(R.string.settings_logout_dialog_message)
         ).also {
             it.isCancelable = false
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
@@ -188,20 +188,18 @@ class AppSettingsActivity :
     }
 
     override fun showLogoutProgressDialog() {
-        val existingDialog = supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog
-        if (existingDialog != null) {
-            progressDialog = existingDialog
-            progressDialog?.isCancelable = false
-            return
-        }
+        if (progressDialog != null) return
+
+        (supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog)
+            ?.dismissAllowingStateLoss()
 
         progressDialog = CustomProgressDialog.show(
             getString(R.string.settings_logout_dialog_title),
             getString(R.string.settings_logout_dialog_message)
         ).also {
+            it.isCancelable = false
             it.show(supportFragmentManager, CustomProgressDialog.TAG)
         }
-        progressDialog?.isCancelable = false
     }
 
     override fun hideLogoutProgressDialog() {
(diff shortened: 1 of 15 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pushnotifications/PushNotificationsRestClient.kt
```diff
@@ -35,11 +36,11 @@ class PushNotificationsRestClient @Inject constructor(private val wooNetwork: Wo
...
         site: SiteModel,
         pushTokenId: String
-    ): WooPayload<Unit> = wooNetwork.executeDeleteGsonRequest(
+    ): WooPayload<Boolean> = wooNetwork.executeDeleteGsonRequest(
         site = site,
         path = WOOCOMMERCE.push_tokens.id(pushTokenId.toLong()).pathPushNotifications,
-        clazz = Unit::class.java,
-    ).toWooPayload()
+        clazz = DeletePushTokenResponse::class.java,
+    ).toWooPayload { it.success }
```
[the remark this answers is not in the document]
hichamboushaba: The token ID is an integer in the plugin API contract.
I think it would be better if we didn't save it as a String in the first place, but this is outside of the scope of the current PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
```diff
@@ -183,6 +187,30 @@ class AppSettingsActivity :
         finish()
     }
 
+    override fun showLogoutProgressDialog() {
+        val existingDialog = supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog
+        if (existingDialog != null) {
+            progressDialog = existingDialog
+            progressDialog?.isCancelable = false
+            return
+        }
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 55b231e2f05.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
```diff
@@ -183,6 +187,30 @@ class AppSettingsActivity :
...
+
+        progressDialog = CustomProgressDialog.show(
+            getString(R.string.settings_logout_dialog_title),
+            getString(R.string.settings_logout_dialog_message)
+        ).also {
+            it.show(supportFragmentManager, CustomProgressDialog.TAG)
+        }
+        progressDialog?.isCancelable = false
+    }
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 55b231e2f05.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsModule.kt
```diff
@@ -3,13 +3,13 @@ package com.woocommerce.android.ui.prefs
...
-import dagger.hilt.android.scopes.ActivityScoped
+import dagger.hilt.android.components.ActivityRetainedComponent
+import dagger.hilt.android.scopes.ActivityRetainedScoped
 
-@InstallIn(ActivityComponent::class)
+@InstallIn(ActivityRetainedComponent::class)
 @Module
 abstract class AppSettingsModule {
-    @ActivityScoped
+    @ActivityRetainedScoped
```
hichamboushaba, line 12: This is needed to ensure we correctly handle the state during configuration changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
```diff
@@ -183,6 +187,28 @@ class AppSettingsActivity :
...
 
+    override fun showLogoutProgressDialog() {
+        if (progressDialog != null) return
+
+        (supportFragmentManager.findFragmentByTag(CustomProgressDialog.TAG) as? CustomProgressDialog)
+            ?.dismissAllowingStateLoss()
+
+        progressDialog = CustomProgressDialog.show(
+            getString(R.string.settings_logout_dialog_title),
+            getString(R.string.settings_logout_dialog_message)
```
irfano, line 198: The title and message are almost identical. Can we use a dialog without a title, similar to what we use in the login flow?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsPresenter.kt
```diff
@@ -12,23 +15,49 @@ class AppSettingsPresenter @Inject constructor(
...
+        if (appSettingsView === view) {
+            appSettingsView = null
+        }
     }
 
     override fun logout() {
+        if (logoutState == LogoutState.IN_PROGRESS) {
+            appSettingsView?.showLogoutProgressDialog()
+            return
+        }
```
irfano, line 42: If the state is `IN_PROGRESS`, it shouldn’t be possible for `logout()` to be called again, yet you still added this check. Is it necessary?

irfano (approved):
LGTM! 👍🏻 I just added a few minor comments.

But I couldn’t see the `Woo Core push token deleted for site ...` log. When I call the delete endpoint, I get a `Push token could not be found.` error from Woo Core. It’s not a blocker for this PR, but I’m wondering why it’s not working on my test site. Does it work for you?

hichamboushaba:
> But I couldn’t see the Woo Core push token deleted for site ... log. When I call the delete endpoint, I get a Push token could not be found. error from Woo Core. It’s not a blocker for this PR, but I’m wondering why it’s not working on my test site. Does it work for you?

Thankt @irfano for this catch, I'm not sure what happened yesterday, but I was sure I saw that the `/delete` endpoint returns a response with `success` param, so I made a change just before creating the PR, but it turns out, it [doesn't return](https://github.com/woocommerce/woocommerce/blame/d31253581a0cd3d2eb0623b204f83fd2988d8632/plugins/woocommerce/src/Internal/PushNotifications/Controllers/PushTokenRestController.php#L243) any content, thie meant that the delete fails, and given we ignore failures, next time you'll try logout, the token that's stored in the app is not available anymore.
You can now test it again, but make sure to delete the datastore, or start from a clean install.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsActivity.kt
hichamboushaba: Fixed in 0d0a1b0

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/AppSettingsPresenter.kt
hichamboushaba: It's just for safety, in case the button was tapped during config changes before the dialog was restored.

hichamboushaba:
@irfano please let me know when you verify the fix for the `/delete` endpoint. Thanks 🙏

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2268,6 +2268,7 @@
     <string name="settings_preferences">Preferences</string>
     <string name="settings_confirm_logout">Are you sure you want to logout from the account %s?</string>
     <string name="settings_confirm_logout_site_credentials">Are you sure you want to log out of your account?</string>
+    <string name="settings_logout_dialog_message">Logging you out.</string>
```
irfano, line 2214:
```suggestion
    <string name="settings_logout_dialog_message">Logging out</string>
```

np: Our login copy is "Logging in". For consistency, what about "Logging out"?

Image

irfano (approved):
It’s working now, Hicham! I also left a nitpick comment. 👍🏻

PR #15668, opened 2026-04-15
Display total sales instead of net revenue on dashboard

JorgeMucientes:
## Description

Closes: WOOMOB-2717

After PR #15493 switched the dashboard from `total_sales` to `net_revenue` to match wp-admin, merchants reported confusion because the revenue figure they were used to seeing (which includes taxes and shipping) dropped unexpectedly.

This PR reverts the data source back to `totalSales` and updates the label to **"Total Sales"** (previously "Net Sales") across:

- **Dashboard performance card** — ViewModel maps `totalSales` from the API instead of `netRevenue`, View displays it
- **Home screen widget** — label updated (note: widget data already used `totalSales` via `GetWidgetStats`, so the label was the only inconsistency)
- **Wear app** — `FetchStats` and `StoreStatsScreen` updated to match
- **Wearable connection** — `WearableConnectionRepository` updated
- **Localized strings** — removed obsolete "Net Sales" translations; all locales fall back to English "Total Sales" until retranslated

### Metrics reference

- `total_sales` = gross_sales − refunds − coupons **+ taxes + shipping**
- `net_revenue` = gross_sales − refunds − coupons **− taxes − shipping**

### What's NOT changed

- The `date_type` fix from #15493 (removing the hardcoded `date_created` override) is preserved — that was a correct fix
- The `GetWidgetStats.kt` data source was already using `totalSales` and is unchanged

## Testing instructions

1. Open the app → **My Store** dashboard
2. Verify the performance card shows **"Total Sales"** as the label
3. Compare the revenue value — it should include taxes and shipping (higher than net sales). Use wc-admin to compare values and ensure it matches "Total Sales" stats. 
4. Check the home screen widget also shows **"Total Sales"**

### DIFF
--- a/WooCommerce-Wear/src/main/java/com/woocommerce/android/wear/ui/stats/StoreStatsScreen.kt
+++ b/WooCommerce-Wear/src/main/java/com/woocommerce/android/wear/ui/stats/StoreStatsScreen.kt
@@ -148,7 +148,7 @@ private fun StatsContentScreen(
     Box(modifier = modifier.fillMaxSize()) {
         Column {
             Text(
-                text = stringResource(id = R.string.stats_screen_net_sales_title),
+                text = stringResource(id = R.string.stats_screen_total_sales_title),
                 textAlign = TextAlign.Center,
                 color = WooColors.woo_purple_5,
                 style = WooTypography.body2,
--- a/WooCommerce-Wear/src/main/java/com/woocommerce/android/wear/ui/stats/datasource/FetchStats.kt
+++ b/WooCommerce-Wear/src/main/java/com/woocommerce/android/wear/ui/stats/datasource/FetchStats.kt
@@ -80,7 +80,7 @@ class FetchStats @Inject constructor(
                     val totals = revenue?.parseTotal()
 
                     val formattedRevenue = wooCommerceStore.formatCurrencyForDisplay(
-                        amount = totals?.netRevenue ?: 0.0,
+                        amount = totals?.totalSales ?: 0.0,
                         site = selectedSite,
                         currencyCode = null,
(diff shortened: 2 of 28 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/285">24.5 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski (approved):
LGTM :shipit:

hichamboushaba:
@JorgeMucientes just pointing something, given this is handled as beta test, the new String won't go through the translation process (that requires code-freeze), are we OK with having it English only? Or maybe we can check if it's OK to use AI translation temporarily?

JorgeMucientes:
Good catch @hichamboushaba, what do you think about the solution I opted for here? [e0c8844](https://github.com/woocommerce/woocommerce-android/pull/15668/commits/e0c88445837e5bde222b928194ef5fcfc44fb8cc)

We can then add the proper string resources when we later implement a proper solution

hichamboushaba:
> Good catch @hichamboushaba, what do you think about the solution I opted for here? [e0c8844](https://github.com/woocommerce/woocommerce-android/pull/15668/commits/e0c88445837e5bde222b928194ef5fcfc44fb8cc)
> 
> We can then add the proper string resources when we later implement a proper solution

This is even better 👍, thanks @JorgeMucientes

PR #15670, opened 2026-04-15
[Woo PN] Refactor push notification registration into a trigger-based orchestrator

hichamboushaba:
### Description
Fixes WOOMOB-2696
Fixes WOOMOB-2682

> [!NOTE]
> **Note on diff size**: The PR diff is large, but the majority of the insertions are tests covering new and modified behaviour. Core production changes are concentrated in RegisterDevice.kt, PushNotificationRepository.kt, GetWooVisibleSites.kt, and the updated entry-point callers.
 

#### Problem

Push notification registration had several issues:

- **Scattered login-success triggers**: login-success registration depended on caller-specific side effects instead of shared success callbacks. Some flows triggered it from `LoginSiteCredentialsViewModel` or indirectly via `FETCH_ACCOUNT` in `AppInitializer`, while `LoginActivity`, `AutoLoginActivity`, and Jetpack post-login did not trigger it directly.
- **No staleness check**: `IF_NEEDED` relied on stored registration status for the selected site. Once a site had Woo registration metadata, token or locale drift was not detected during non-forced runs.
- **Spurious registrations**: `AppInitializer` registered on every `FETCH_ACCOUNT` FluxC event, not just on fresh logins.

#### Solution

`RegisterDevice` is refactored into a single-entry-point orchestrator driven by a `Trigger` enum. Each trigger encodes both the **scope** (which sites to register) and the **behaviour** (force vs. staleness-check):

| Trigger | Emitted when | Sites registered | Forces re-registration? |
|---|---|---|---|
| `LOGIN_SUCCESS` | Login completes (all flows) | All visible Woo sites | No — staleness check |
| `APP_FOREGROUND` | App returns to foreground | All visible Woo sites | No — staleness check |
| `SITE_SWITCH` | User changes the active site | Selected site only | No — staleness check, no WP.com registration/fallback |
| `TOKEN_REFRESH` | FCM token is refreshed or retrieved | All visible Woo sites | Yes — also cancels in-progress job |

**Staleness check** — `PushNotificationRepository` now persists per-site registration metadata in DataStore (token value + locale). `shouldRegisterWooPushForSite()` skips re-registration unless the FCM token value or device locale has actually changed since the last registration.

**Concurrency safety** — a `Mutex` serializes concurrent invocations. `TOKEN_REFRESH` cancels any in-progress job before acquiring the lock, ensuring the fresh token is always used.

**`GetWooVisibleSites` extension** — the use case now includes the currently selected Application Passwords site if it is not already in the WP.com site list, so these sites are never missed during a full-sweep registration.

#### Architecture

```mermaid
flowchart TD
    subgraph entryPoints["Entry Points"]
        E1["FCMMessageService / FCMRefreshWorker"]
        E2["LoginActivity / AutoLoginActivity"]
        E3["SitePickerViewModel (from login)"]
        E4["SitePickerViewModel (site changed)"]
        E5["AppInitializer (foreground)"]
        E6["WPComLoginPostLoginViewModel (Jetpack Setup)"]
    end

    E1 -->|TOKEN_REFRESH| RD["RegisterDevice"]
    E2 -->|LOGIN_SUCCESS| RD
    E3 -->|LOGIN_SUCCESS| RD
    E6 -->|LOGIN_SUCCESS| RD
    E4 -->|SITE_SWITCH| RD
    E5 -->|APP_FOREGROUND| RD

    RD --> Scope{"Trigger\nscope"}
    Scope -->|"LOGIN_SUCCESS / APP_FOREGROUND / TOKEN_REFRESH"| AllSites["GetWooVisibleSites()"]
    Scope -->|SITE_SWITCH| SelectedOnly["Selected site only"]

    AllSites --> Sweep["Per-site Woo registration sweep"]
    SelectedOnly --> Sweep

    Sweep --> Force{"TOKEN_REFRESH?"}
    Force -->|Yes| Woo["Register in Woo Core"]
    Force -->|No| Check{"Token or locale\nchanged?"}
    Check -->|Yes| Woo
    Check -->|No| SkipWoo["Skip Woo registration"]

    RD --> WpComCheck{"trigger != SITE_SWITCH\n&& hasAccessToken\n&& (forced || WP.com already registered)"}
    WpComCheck -->|Yes| WpCom["Register in WP.com"]
    WpComCheck -->|No| SkipWpCom["Skip WP.com"]
```

**Staleness check detail:**

```mermaid
flowchart LR
    A["shouldRegisterWooPushForSite(token, siteId)"]
    A --> B{"DataStore record\nexists?"}
    B -->|No| C["Register"]
    B -->|Yes| D{"Token\nchanged?"}
    D -->|Yes| C
    D -->|No| E{"Locale\nchanged?"}
    E -->|Yes| C
    E -->|No| F["Skip"]
```

### Test Steps
**For all tests, make sure PN feature flags are enabled**

#### Correctness tests
1. Make sure at least one of your sites is using the Woo PN test build (from here paaHJt-9VE-p2)
2. Open Logcat filtered to `WooCommerce-NOTIFICATIONS`.
3. Log in via WP.com credentials. Verify logs show:
   - `Starting push registration for LOGIN_SUCCESS`
   - `Registering Woo push for site <siteId> for LOGIN_SUCCESS` for each visible Woo site
   - `Registering WP.com push for LOGIN_SUCCESS`
4. Switch sites in the site picker. Verify logs show:
   - `Starting push registration for SITE_SWITCH`
   - `Registering Woo push for site <selected-site-id> for SITE_SWITCH` (only if the site's registration is stale; absent if up to date)
   - `Skipping WP.com push registration for SITE_SWITCH`
5. Log in via Application Passwords. Verify logs show `Starting push registration for LOGIN_SUCCESS` and `Registering Woo push for site <siteId> for LOGIN_SUCCESS` for the selected site.
6. **Optional:** Log in via Magic Link. Verify the same `LOGIN_SUCCESS` log sequence appears.
8. Background the app and foreground it again. Verify logs show `Starting push registration for APP_FOREGROUND`. `Registering Woo push for site <siteId> for APP_FOREGROUND` appears only for sites with a stale registration; no per-site line appears for up-to-date sites.
10. With a registration already in progress, trigger a token refresh. Verify logs show `Cancelling in-progress push registration before TOKEN_REFRESH` before the `TOKEN_REFRESH` sequence begins.

#### WOOMOB-2696 fix
1. Make sure at least one of your sites is using the Woo PN test build (from here paaHJt-9VE-p2)
2. Sign in (both login options are fine).
3. Change the device language, then background the app and foreground it again, Verify logs show `Starting push registration for APP_FOREGROUND` followed by `Registering Woo push for site <siteId> for APP_FOREGROUND`.

#### WOOMOB-2682 fix
1. Use WordPress.com login, and make sure at least one of your sites is using the Woo PN test build (from here paaHJt-9VE-p2), let's say site A.
2. Start login with a different site.
3. After login, test a PN to site A.
4. Confirm the PN is received.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -173,21 +173,17 @@ class PushNotificationRepository @Inject constructor(
         remove(getPushLocaleKeyForSite(siteId))
     }
 
-    private fun getPushTokenIdKeyForSite(siteId: Long): Preferences.Key<String> {
-        return stringPreferencesKey("$PUSH_TOKEN_KEY_PREFIX$siteId")
-    }
+    private fun getPushTokenIdKeyForSite(siteId: Long): Preferences.Key<String> =
+        stringPreferencesKey("$PUSH_TOKEN_KEY_PREFIX$siteId")
 
-    private fun getPushTokenValueKeyForSite(siteId: Long): Preferences.Key<String> {
-        return stringPreferencesKey("$PUSH_TOKEN_VALUE_KEY_PREFIX$siteId")
-    }
+    private fun getPushTokenValueKeyForSite(siteId: Long): Preferences.Key<String> =
+        stringPreferencesKey("$PUSH_TOKEN_VALUE_KEY_PREFIX$siteId")
 
-    private fun getPushLocaleKeyForSite(siteId: Long): Preferences.Key<String> {
-        return stringPreferencesKey("$PUSH_LOCALE_KEY_PREFIX$siteId")
-    }
+    private fun getPushLocaleKeyForSite(siteId: Long): Preferences.Key<String> =
+        stringPreferencesKey("$PUSH_LOCALE_KEY_PREFIX$siteId")
 
-    suspend fun isWooPushTokenRegisteredForSite(siteId: Long): Boolean {
-        return observeWooPushTokenRegisteredForSite(siteId).first()
-    }
+    suspend fun isWooPushTokenRegisteredForSite(siteId: Long): Boolean =
+        observeWooPushTokenRegisteredForSite(siteId).first()
 
     suspend fun shouldRegisterWooPushForSite(currentToken: String, siteId: Long): Boolean {
         val preferences = pushNotificationsDataStore.data.first()
@@ -198,11 +194,10 @@ class PushNotificationRepository @Inject constructor(
             registration.locale != getDeviceLocale()
     }
 
-    fun isWpComPushRegistered(): Boolean {
-        val wpComPushServerId = prefsWrapper.getFluxCPreferences()
+    fun isWpComPushRegistered(): Boolean =
+        prefsWrapper.getFluxCPreferences()
             .getString(WpComPushNotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
-        return wpComPushServerId.isNotNullOrEmpty()
-    }
+            .isNotNullOrEmpty()
 
     fun observeWooPushTokenRegisteredForSite(siteId: Long): Flow<Boolean> {
         val tokenKey = getPushTokenIdKeyForSite(siteId)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -2,6 +2,7 @@ package com.woocommerce.android.notifications.push
 
 import android.os.Build
 import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.MutablePreferences
 import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.stringPreferencesKey
@@ -50,10 +51,17 @@ class PushNotificationRepository @Inject constructor(
             tag = WooLog.T.NOTIFICATIONS,
             message = "Registering FCM token in WPCOM instance${if (BuildConfig.DEBUG) ": $token" else ""}"
         )
-        wpComPushNotificationStore.registerDevice(token, WpComPushNotificationStore.NotificationAppKey.WOOCOMMERCE)
+        wpComPushNotificationStore.registerDevice(
+            token,
+            WpComPushNotificationStore.NotificationAppKey.WOOCOMMERCE
+        )
     }
 
-    suspend fun registerPushTokenInWooCoreSystem(token: String, selectedSite: SiteModel): Result<Unit> {
+    suspend fun registerPushTokenInWooCoreSystem(
+        token: String,
+        selectedSite: SiteModel,
+        allowWpComFallback: Boolean = true
+    ): Result<Unit> {
         WooLog.d(
             tag = WooLog.T.NOTIFICATIONS,
             message = "Registering FCM token in Woo Core instance${if (BuildConfig.DEBUG) ": $token" else ""}"
@@ -70,12 +78,12 @@ class PushNotificationRepository @Inject constructor(
             metadata = metadata
         )
         return if (!result.isError) {
-            result.model?.let { pushToken ->
+            result.model?.let { tokenId ->
                 notificationAnalyticsTracker.track(
                     stat = AnalyticsEvent.WOO_PUSH_TOKEN_REGISTER_SUCCESS,
                     siteId = selectedSite.siteId
                 )
-                savePushTokenForSite(selectedSite.siteId, pushToken)
+                savePushTokenForSite(selectedSite.siteId, WooPushRegistrationData(tokenId, token, deviceLocale))
                 disableWpComNotificationsForSite(selectedSite.siteId)
                 Result.success(Unit)
             } ?: run {
@@ -101,7 +109,7 @@ class PushNotificationRepository @Inject constructor(
                 WooLog.T.NOTIFICATIONS,
                 "Woo Core push token registration failed:${result.error?.message}, fallback to WPCom"
             )
-            if (!isWpComPushRegistered()) {
+            if (allowWpComFallback && !isWpComPushRegistered()) {
                 registerPushTokenInWpComSystem(token)
             }
             Result.failure(WooException(result.error))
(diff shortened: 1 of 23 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -191,17 +236,18 @@ class PushNotificationRepository @Inject constructor(
         val sites = wooCommerceStore.getWooCommerceSites()
 
         val deleteJobs = sites.mapNotNull { site ->
-            val tokenKey = getPushTokenKeyForSite(site.siteId)
-            val pushTokenId = preferences[tokenKey] ?: return@mapNotNull null
+            val registration = preferences.getPushRegistration(site.siteId) ?: return@mapNotNull null
             async {
-                val result = wooPushNotificationsStore.deletePushToken(site, pushTokenId)
+                val result = wooPushNotificationsStore.deletePushToken(site, registration.tokenId)
                 if (result.isError) {
```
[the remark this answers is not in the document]
hichamboushaba: The feature is behind a feature flag now, so no harm in this.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -1,66 +1,132 @@
...
+            invoke(trigger)
+        }
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    suspend operator fun invoke(trigger: Trigger) {
+        if (trigger == Trigger.TOKEN_REFRESH && activeJob != null) {
+            WooLog.d(WooLog.T.NOTIFICATIONS, "Cancelling in-progress push registration before $trigger")
+            activeJob?.cancel()
+        }
```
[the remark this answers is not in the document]
hichamboushaba: The race is real but the outcome is still correct. If TOKEN_REFRESH (B) sees `activeJob == null` while A is inside the mutex (between launch and the `activeJob =` assignment), B will wait for the mutex and run sequentially after A finishes. B then registers with the fresh token, so the device ends up in the right state — A just did redundant work. Given the narrow window and benign outcome, leaving the current structure as-is.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -1,66 +1,132 @@
...
+                            )
+                            pushNotificationRepository.registerPushTokenInWooCoreSystem(
+                                token = token,
+                                selectedSite = site,
+                                allowWpComFallback = trigger != Trigger.SITE_SWITCH
+                            )
+                        }
+                    }
+                }.awaitAll()
             }
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 766a878 — switched to `supervisorScope` so one site failure does not cancel registration for the remaining sites.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -1,66 +1,132 @@
...
+
+        val sites = when (trigger) {
+            Trigger.LOGIN_SUCCESS,
+            Trigger.APP_FOREGROUND,
+            Trigger.TOKEN_REFRESH -> getWooVisibleSites()
+
+            Trigger.SITE_SWITCH -> listOfNotNull(selectedSite.getIfExists())
+        }
+
+        if (featureFlagRepository.isEnabled(FeatureFlag.WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1)) {
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 766a878 — moved `sites` resolution inside the feature flag block so `getWooVisibleSites()` is only called when the flag is enabled.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/sitevisibility/GetWooVisibleSites.kt
```diff
@@ -1,19 +1,33 @@
...
+            .plusSelectedAppPasswordWooSiteIfMissing()
+            .distinctBy { it.id }
 
     private suspend fun isSiteVisible(siteId: Long): Boolean {
         return visibleSitesDataStore.isSiteVisible(siteId).first()
     }
+
+    private fun List<SiteModel>.plusSelectedAppPasswordWooSiteIfMissing(): List<SiteModel> {
+        val currentSite = selectedSite.getOrNull()
+        if (currentSite != null && currentSite.hasWooCommerce && currentSite.connectionType == ApplicationPasswords) {
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 01b28d871d678f34dcdec1b684370be4de92f481 — added the explicit `none { it.id == currentSite.id }` check and removed `distinctBy` since it is no longer needed for correctness.

irfano (commented):
In the test steps, you say: 

> Registering Woo push for site <siteId> for APP_FOREGROUND appears only for sites with a stale registration; no per-site line appears for up-to-date sites.

But every time I bring the app from the background, I see `Registering Woo push for site 241593856 for APP_FOREGROUND` in the logs, and I see the same log for my other sites as well. Is this expected? How do we determine whether a token is stale?

hichamboushaba:
> But every time I bring the app from the background, I see Registering Woo push for site 241593856 for APP_FOREGROUND in the logs, and I see the same log for my other sites as well. Is this expected? How do we determine whether a token is stale?

Thanks @irfano, this could happen for sites that are not compatible with Woo notifications, sorry I should have mentioned this in the PR description.

But now, thinking more about this, this could be an issue worth addressing, as we may make multiple API requests needlessly. I think I may change the behavior to make sure we register only for the current site for the APP_FOREGROUND trigger, but this may mean the user will receive notifications with the wrong locale for the other sites, though. I'm not sure if this is OK as an edge case?
I'll think more about whether there is a better solution and update the PR.

irfano:
If I understand correctly, we can prevent this simply by checking the plugin version. Of course, there may be other cases, but they would be rarer than plugin version incompatibility.

hichamboushaba:
> If I understand correctly, we can prevent this simply by checking the plugin version. Of course, there may be other cases, but they would be rarer than plugin version incompatibility.

Yes, it's a possibility, but we might not have the plugin version cached, and also, we can rely on the cached version to be up-to-date for non selected sites. So to rely on plugin version, we'll need to make another request, so we are back to square 1.

I went with my idea of checking only for selected site (d706531), the locale issue is an edge case IMO, I don't imagine people changing locale often, and the other triggers will update it, just at a later point, let me know what you think?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -536,7 +536,12 @@ class SitePickerViewModel @Inject constructor(
                             selectedSite.set(siteVerificationModel.siteModel)
                             trackLoginEvent(currentStep = UnifiedLoginTracker.Step.SUCCESS)
                             appPrefsWrapper.removeLoginSiteAddress()
-                            registerDevice(RegisterDevice.Mode.IF_NEEDED)
+                            val registerDeviceTrigger = if (navArgs.openedFromLogin) {
+                                RegisterDevice.Trigger.LOGIN_SUCCESS
```
irfano, line 540: If it’s opened from login, `registerDevice.kickoff()` will already have been called in `LoginActivity`. Do we still need to call it here as well?

WooCommerce/src/test/kotlin/com/woocommerce/android/notifications/push/RegisterDeviceTest.kt
```diff
@@ -1,186 +1,294 @@
...
 
     @Test
-    fun `given user logged in and UNREGISTERED, when invoked forcefully, then registers in WPCom`() = testBlocking {
-        setupStatus(Status.UNREGISTERED)
-
-        sut(FORCEFULLY)
-
-        verify(pushNotificationRepository).registerPushTokenInWpComSystem(TEST_TOKEN)
-    }
+    fun `given token refresh trigger, when registration is otherwise unchanged, then forces Woo and WPCom registration`() =
```
irfano: I’m not sure what “otherwise unchanged” means in this test name. Would you like to update it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
hichamboushaba: Yes, this is a good question. I did it like this because in the SitePicker, we use `WooCommerceStore.fetchWooCommerceSites`, this functions does more checks for detecting Woo installation for JCP sites, this means that we might have some new sites being detected as Woo sites at this level, and the registration would be useful for them.
Triggering the registration a second time will generally be a no-op, given the `mutex` and that we save the state.

WooCommerce/src/test/kotlin/com/woocommerce/android/notifications/push/RegisterDeviceTest.kt
hichamboushaba: This just means that we don't check the registration status, I'll try to update it to be clearer.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -1,66 +1,132 @@
...
+                            )
+                        }
+                    }
+                }.awaitAll()
+            }
+        }
+
+        // For WPCom, site switching doesn't affect registration
+        val shouldEvaluateWpCom = trigger != Trigger.SITE_SWITCH &&
+            (shouldForce || pushNotificationRepository.isWpComPushRegistered())
```
irfano: For fresh logins on WP sites, we reach this line with `LOGIN_SUCCESS`. In this case, `shouldForce` will be `false`, and `pushNotificationRepository.isWpComPushRegistered()` will also be false. That means we’ll skip registration for WP sites. We need to fix this.

irfano (changes requested):

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
irfano: Sounds good.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
hichamboushaba: > and pushNotificationRepository.isWpComPushRegistered() will also be false

Why would it be false for fresh logins? `isWpComPushRegistered` checks the status of WPCom registration, so if it was already registered, then no need to call it. Am I missing something?

hichamboushaba: Sorry I misread it, now I understand, I will fix it, good catch, it should be negated.

hichamboushaba: Fixed in 25ada5c292c13b171b8547e7c2b0ea5fd3f508d3, thanks again for the catch, I added a test case to catch it.

WooCommerce/src/test/kotlin/com/woocommerce/android/notifications/push/RegisterDeviceTest.kt
hichamboushaba: Fixed in 25ada5c292c13b171b8547e7c2b0ea5fd3f508d3 — renamed to `given token refresh trigger, then forces Woo and WPCom registration regardless of current state`.

irfano: The update made it clearer for me. Thank you!

irfano (approved):
Great job refactoring this area and fixing the issue, Hicham! LGTM! 👍🏻 
Good to merge after fixing the failing unit tests.

PR #15679, opened 2026-04-16
[WOOMOB-2716] Show notifications settings for all sites when Woo PNs are available

hichamboushaba:
### Description
Fixes WOOMOB-2716

Keep the Settings notifications row available for application-password sites when `WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M2` is enabled. This preserves the existing hide behavior when M2 is off, while restoring a notification management entry point once the self-driven push notifications flow is available.

### Test Steps
1. Enable `WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M2` for a site that uses application-password authentication.
2. Open Settings.
3. Verify the `Notifications` row is visible.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/MainSettingsContract.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/MainSettingsContract.kt
@@ -29,6 +29,6 @@ interface MainSettingsContract {
         fun showLatestAnnouncementOption(announcement: FeatureAnnouncement)
         fun setEnablePushNotificationsOptionVisible(isVisible: Boolean)
         fun handleJetpackInstallOption(supportsJetpackInstallation: Boolean)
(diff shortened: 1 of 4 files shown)


irfano (commented):
Can’t we do the same thing we’re doing in #15676 here and keep the notification settings visible all the time, since they’re also useful for local notifications?

hichamboushaba:
> Can’t we do the same thing we’re doing in https://github.com/woocommerce/woocommerce-android/pull/15676 here and keep the notification settings visible all the time, since they’re also useful for local notifications?

Good question, I guess we can, my concern was that we show "orders" and "reviews" for the channels, and we have some logic specific to the cha-ching notification if the sound was disabled, so it could be confusing. But I guess eventually all sites will start supporting notifications soon, so it's OK to show it for all. I'll update the ticket.

irfano (approved):
LGTM! 👍🏻

PR #15686, opened 2026-04-17
[WOOMOB-2709] Phone Cart and Totals screen wrappers

kidinov:
### Description

Fixes WOOMOB-2709

Adds phone-adapted Cart and Totals screen wrappers for POS, reusing the existing tablet composables — no logic duplication.

Phone navigation now uses the same single source of truth as tablet: `WooPosHomeState.screenPositionState`. A new `Products` state value was added, which tablet renders the same as `Cart`. Phone-specific cart overrides (back arrow always visible, clear-all by item count) live in the phone wrapper as a pure transform, not in the cart ViewModel — the VM stays form-factor agnostic.

Smaller fixes bundled in:
- `TotalsLoading` shimmer + `TotalsPaymentSuccess` buttons get explicit horizontal padding
- Cart toolbar back+title spacing matches the standard `WooPosToolbar`
- Connect-to-reader and totals grid use `adaptiveContentWidth()`

### Test Steps

1. Run on a phone (Pixel 9 Pro XL works). The `WOO_POS_PHONE` flag is on by default in debug.
2. Open POS → add a few products.
3. Tap the **Cart (debug)** FAB → Cart screen renders full-screen with back arrow top-left (aligned with title), trash icon top-right, Check out button at bottom.
4. Tap **Check out** → Totals renders full-screen.
5. System back → lands on Cart with back+trash+Check out still visible.
6. System back → Products.
7. Open **Orders** from the menu → system back → must land on **Products**, not Totals.
8. Cash payment flow → payment success buttons have side margins.
9. **Tablet**: open POS and run through the same flows. Should look the same as before this PR.

### Images/gif

**Tablet — no regression**

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

**Phone — new screens**

| Cart | Totals | Cart after back from Totals | Payment success |
| --- | --- | --- | --- |
|  |  |  |  |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -38,7 +38,7 @@ class WooPosHomeViewModel @Inject constructor(
         scope = viewModelScope,
         key = "home_state",
         initialValue = WooPosHomeState(
-            screenPositionState = ScreenPositionState.Cart,
+            screenPositionState = ScreenPositionState.Products,
```
malinajirka, line 41: 💡 Flipping the initial state to `Products` silently breaks barcode scanning on tablet. `WooPosHomeContainer.kt`'s `isBarcodeListeningEnabled` only returns true for `Cart` / `FullScreenTotals`, so on tablet the listener is off at launch and again after every `OnNewTransactionStarted` resets back to `Products`. On tablet the cart is always visible, so `Products` should be scannable too.

Repro on a tablet:
1. Launch POS → lands in `Products` (cart panel visible, "scan a barcode" hint showing).
2. Try to scan a barcode → nothing happens.
3. Tap any product → Check out → system back → scan again → works.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosPhoneCartScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+private fun applyPhoneOverrides(state: WooPosCartState): WooPosCartState {
+    val hasItems = state.body is WooPosCartState.Body.WithItems
+    val effectiveCheckoutButton = when {
+        !hasItems -> WooPosCartState.CheckoutButtonState.Invisible
+        state.checkoutButtonState == WooPosCartState.CheckoutButtonState.Disabled ->
+            WooPosCartState.CheckoutButtonState.Disabled
+        else -> WooPosCartState.CheckoutButtonState.Enabled
```
malinajirka: ❓ I wonder if this override is a symptom of the cart's `CHECKOUT` state not fitting phone.

On tablet, during checkout the cart stays visible alongside totals, so the cart VM switches to `CHECKOUT` (read-only, clear-all hidden, Checkout button `Invisible`) to prevent edits while a payment is in progress. On phone the cart is fully replaced by totals — there's nothing to "lock" — but the VM still transitions to `CHECKOUT`, which is why this override exists: to paper over the tablet-shaped `Invisible` state.

The leak shows up on back-from-totals: phone system back from `FullScreenTotals` lands on the Cart screen without dispatching `BackFromCheckoutToCartClicked`, so the cart VM is still in `CHECKOUT`. This override then flips `Invisible` → `Enabled` and the user can re-fire checkout while a payment may still be pending.

Cleanest fix imo: on phone, back from totals should cancel the payment attempt and return the cart to `EDITABLE`. That removes both the override and the re-fire-during-payment footgun. Wdyt?

malinajirka (changes requested):
Thanks @kidinov! Overall looks good. I've found a couple issues I believe are worth investigating.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
kidinov: Good catch, fixed in https://github.com/woocommerce/woocommerce-android/pull/15686/commits/0190aa3f8509295c1b1261e844e973b4ccd683ac

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosPhoneCartScreen.kt
kidinov: Good catch, fixed in https://github.com/woocommerce/woocommerce-android/pull/15686/commits/3dba45107e66ca67060f5416754e753b200e063c

I dispatched `BackFromCheckoutToCartClicked` from the `FullScreenTotals` back path so the cart VM resets to `EDITABLE` on its own. That removed `applyPhoneOverrides` entirely - the only phone-specific tweak left is `backIconVisible = true`. The re-fire-during-payment footgun is closed because the cart status returns to `EDITABLE` before the user can tap Checkout again.

Cancelling the actual in-flight card payment is a bigger concern, tracked in [WOOMOB-2752](https://linear.app/a8c/issue/WOOMOB-2752/cancel-in-flight-card-payment-when-user-backs-out-of-totals-on-phone).

malinajirka (approved):
LGTM, thanks!

PR #11400, opened 2024-04-26
Address SQLiteBlobTooBigException crash

0nko:
This is an attempt to address some of the crash events reported in #11309. The PR doubles the size of the cursor window of the database for devices running API 28 and higher. While it probably won't fix all crashes (there can always be larger data) and it won't work for older Android versions, it should have an impact. At the same time, we don't want to make the cursor window size too big to avoid OOM errors, so 10MB seemed like a reasonable compromise.

**To test:**
There isn't much to test. I did verify that the fix worked for an absurdly large product description that was causing the app to crash previously.

PR #2627, opened 2020-07-16
Expand cursor window to 5mb for production builds running API 28+

AmandaRiu:
Partially fixes #572 by overriding the WellSql method `getCursorWindowSize()` and returning a size of 5mb. This method was [added to WellSQL in this PR](https://github.com/wordpress-mobile/wellsql/pull/18) by @nbradbury and has been in production for `DEBUG` WCAndroid builds for the last two releases. Since there haven't been any issues this PR simply makes it available to production. **Please note this fix only works on Android P and above.** 

Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15693, opened 2026-04-17
[WOOMOB-2585] Increase Room CursorWindow size to avoid SQLiteBlobTooBigException

hichamboushaba:
### Description

Fixes WOOMOB-2585

Applies the same mitigation as the WellSql side to Room-backed queries: wrap `SupportSQLiteOpenHelper.Factory` so every `query(...)` call installs a 10 MB `CursorWindow` on API 28+, avoiding `SQLiteBlobTooBigException` when a product row's combined columns exceed SQLite's default ~2 MB window.

Precedent:
- #11400 — bumped WellSql cursor window to 10 MB
- #2627 — original 5 MB bump on API 28+
- wordpress-mobile/WordPress-FluxC-Android#1586 — added `getCursorWindowSize()` to fluxc's `WellSqlConfig`

Not a structural fix — a single row over 10 MB would still crash, and the long-term answer is pagination or narrower projections on `ProductsDao.getProducts`. But product row sizes vary widely per site and matching the WellSql precedent gives immediate relief with minimal risk.

### Test Steps

Automated coverage:
- Unit tests verify the factory invokes `setWindow(...)` on all four `query(...)` overloads on API 28+ and skips it below.
- Robolectric integration test inserts a 3 MB product row into an in-memory `WCAndroidDatabase`: reproduces the crash with the default factory and confirms the row round-trips cleanly with the new factory.

Non regression tests on device (API 28+):
1. Open the Products tab and scroll — no crash in Logcat/Sentry.
2. Pull-to-refresh — still no crash.

### Images/gif

N/A

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
@@ -49,7 +49,7 @@ internal class LargeCursorWindowOpenHelperFactory(
 
         private fun Cursor.withLargerWindow(): Cursor = apply {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && this is SQLiteCursor) {
-                setWindow(CursorWindow(null, windowSizeBytes))
+                window = CursorWindow(null, windowSizeBytes)
             }
         }
     }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
@@ -1,0 +1,56 @@
+package org.wordpress.android.fluxc.persistence
+
+import android.database.Cursor
+import android.database.CursorWindow
+import android.database.sqlite.SQLiteCursor
+import android.os.Build
+import android.os.CancellationSignal
+import androidx.sqlite.db.SupportSQLiteDatabase
+import androidx.sqlite.db.SupportSQLiteOpenHelper
+import androidx.sqlite.db.SupportSQLiteQuery
+import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
+
+// Mirrors `WooWellSqlConfig.getCursorWindowSize`; only active on API 28+.
+internal class LargeCursorWindowOpenHelperFactory(
+    private val windowSizeBytes: Long,
+    private val delegate: SupportSQLiteOpenHelper.Factory = FrameworkSQLiteOpenHelperFactory()
+) : SupportSQLiteOpenHelper.Factory {
+
+    override fun create(
+        configuration: SupportSQLiteOpenHelper.Configuration
+    ): SupportSQLiteOpenHelper = OpenHelperWrapper(delegate.create(configuration), windowSizeBytes)
+
+    private class OpenHelperWrapper(
+        private val delegate: SupportSQLiteOpenHelper,
+        private val windowSizeBytes: Long
+    ) : SupportSQLiteOpenHelper by delegate {
+        override val writableDatabase: SupportSQLiteDatabase
+            get() = DatabaseWrapper(delegate.writableDatabase, windowSizeBytes)
+
+        override val readableDatabase: SupportSQLiteDatabase
+            get() = DatabaseWrapper(delegate.readableDatabase, windowSizeBytes)
+    }
+
+    private class DatabaseWrapper(
+        private val delegate: SupportSQLiteDatabase,
+        private val windowSizeBytes: Long
+    ) : SupportSQLiteDatabase by delegate {
+        override fun query(query: String): Cursor =
+            delegate.query(query).withLargerWindow()
+
+        override fun query(query: String, bindArgs: Array<out Any?>): Cursor =
+            delegate.query(query, bindArgs).withLargerWindow()
+
+        override fun query(query: SupportSQLiteQuery): Cursor =
+            delegate.query(query).withLargerWindow()
+
+        override fun query(query: SupportSQLiteQuery, cancellationSignal: CancellationSignal?): Cursor =
+            delegate.query(query, cancellationSignal).withLargerWindow()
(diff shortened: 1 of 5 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+            delegate.query(query).withLargerWindow()
+
+        override fun query(query: SupportSQLiteQuery, cancellationSignal: CancellationSignal?): Cursor =
+            delegate.query(query, cancellationSignal).withLargerWindow()
+
+        private fun Cursor.withLargerWindow(): Cursor = apply {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && this is SQLiteCursor) {
+                setWindow(CursorWindow(null, windowSizeBytes))
+            }
+        }
```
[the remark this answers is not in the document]
hichamboushaba: Not worth it IMO.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/LargeCursorWindowRoomIntegrationTest.kt
```diff
@@ -0,0 +1,74 @@
...
+@Config(sdk = [Build.VERSION_CODES.P])
+internal class LargeCursorWindowRoomIntegrationTest {
+    private val appContext = ApplicationProvider.getApplicationContext<Application>()
+
+    @get:Rule
+    val defaultDb = DatabaseTestRule(appContext)
+
+    @get:Rule
+    val enlargedDb = DatabaseTestRule(appContext, LargeCursorWindowOpenHelperFactory(TEN_MEGABYTES))
+
```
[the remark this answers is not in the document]
hichamboushaba: Not worth it IMO, in-memory Room init is a few ms, negligible next to Robolectric startup, and the symmetric `@Rule` pair keeps the tests clean with automatic close-on-failure.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        override fun query(query: SupportSQLiteQuery): Cursor =
+            delegate.query(query).withLargerWindow()
+
+        override fun query(query: SupportSQLiteQuery, cancellationSignal: CancellationSignal?): Cursor =
+            delegate.query(query, cancellationSignal).withLargerWindow()
+
+        private fun Cursor.withLargerWindow(): Cursor = apply {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && this is SQLiteCursor) {
+                setWindow(CursorWindow(null, windowSizeBytes))
```
irfano: np: Let’s apply this to address the IDE’s suggestion.

```suggestion
                window = CursorWindow(null, windowSizeBytes)
```

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+import android.os.Build
+import android.os.CancellationSignal
+import androidx.sqlite.db.SupportSQLiteDatabase
+import androidx.sqlite.db.SupportSQLiteOpenHelper
+import androidx.sqlite.db.SupportSQLiteQuery
+import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
+
+// Mirrors `WooWellSqlConfig.getCursorWindowSize`; only active on API 28+.
+internal class LargeCursorWindowOpenHelperFactory(
+    private val windowSizeBytes: Long,
```
irfano, line 15: I wonder whether taking `windowSizeBytes` as a parameter provides any benefit. Since the value is always the same, what do you think about defining it directly in `LargeCursorWindowOpenHelperFactory`?

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/LargeCursorWindowRoomIntegrationTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private fun hugeProduct() = WCProductModel(
+        localSiteId = LocalId(SITE_ID),
+        remoteId = RemoteId(1L),
+        name = "huge",
+        description = "x".repeat(HUGE_DESCRIPTION_LENGTH)
+    )
+
+    private suspend fun org.wordpress.android.fluxc.persistence.dao.ProductsDao.fetchAllProducts() = getProducts(
```
irfano: Let's import `ProductsDao`.

irfano (approved):
LGTM! Thanks for adding a general fix. I’m approving this, but please review my non-blocking comments before merging.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/LargeCursorWindowOpenHelperFactory.kt
hichamboushaba: Yes, this would work, but I think keeping the factory configurable has its benefits. We may need to apply the same logic for the `wp-android` DB, but with a smaller limit in the future. WDYT?

irfano: Then it’s okay. Let’s keep it as it is.

libs/fluxc-plugin/src/test/java/org/wordpress/android/fluxc/persistence/LargeCursorWindowRoomIntegrationTest.kt
hichamboushaba: Addressed in [85d7c1c](https://github.com/woocommerce/woocommerce-android/pull/15693/commits/85d7c1c94ddae00308d0044313a8f9fff1e84f87)

PR #15696, opened 2026-04-20
[WOOMOB-2735] Add REMOTE_TAP_TO_PAY feature flag

kidinov:
### Description

Fixes WOOMOB-2735

Adds a `REMOTE_TAP_TO_PAY` feature flag that will gate the upcoming Woo POS Remote Tap-to-Pay feature on both the phone and tablet surfaces. Default off in production, on in debug builds — same pattern as `WOO_POS_PHONE`. The flag is read independently on each surface so the phone and tablet can be flipped separately for dogfooding.

With the flag off on both surfaces, the app behaves identically to trunk. All follow-up issues in the Remote Tap-to-Pay project reference this flag for gating.

### Test Steps

1. Build a debug APK. Open Settings → Developer Options → Feature Flags and verify `REMOTE_TAP_TO_PAY` appears in the list with an override toggle.
2. Toggle the override off. Verify the repository reports the flag as disabled.
3. Toggle the override back on. Verify the repository reports the flag as enabled.
4. Run unit tests: \`./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "com.woocommerce.android.util.FeatureFlagRepositoryTest"\` — the new \`given REMOTE_TAP_TO_PAY flag, when remote key read, then matches contract\` test passes alongside the existing suite.

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 24.7
 -----
+- [Internal] Added REMOTE_TAP_TO_PAY feature flag (off in production) for the upcoming Woo POS Remote Tap-to-Pay feature [https://github.com/woocommerce/woocommerce-android/pull/15696]
 - [*] Fixed a crash when loading the product list on stores with very large product descriptions [https://github.com/woocommerce/woocommerce-android/pull/15693]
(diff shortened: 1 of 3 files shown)


samiuelson (approved):
LGTM!

PR #15697, opened 2026-04-20
[WOOMOB-2748] Fix notification navigation for App Passwords sites

hichamboushaba:
Fixes WOOMOB-2748

Both issues fixed by this PR only affect users signed in with site credentials (Application Passwords). In that flow the local site is stored with its self-hosted `siteId` rather than the WP.com one, so push notifications — which always carry the WP.com `remoteSiteId` — never match the locally stored site. This surfaced in two places:

1. The site-switch check in `onPushNotificationTapped` compared the notification's `remoteSiteId` to the selected site's `siteId` and always triggered a site switch that could never resolve.
2. The `NewOrder` branch in `onSinglePushNotificationOpened` looked up the site via `SiteStore.getSiteBySiteId` and, when it returned null, fell back to `ViewOrderList` instead of opening the order details.

The branch contains:

- **Skip site switch on push notification for ApplicationPasswords sites** — guards the site-switch path with a `SiteConnectionType` check so site-credential sites handle notifications directly against the current site.
- **Open order detail for new order notifications even for unknown sites** — removes the `getSiteBySiteId` guard so the order detail screen opens regardless of whether the notification's site can be resolved locally.
- **Remove unused `remoteNoteId` nav arg from order detail screen** — cleans up an obsolete nav argument threaded through `ViewOrderDetail`, `MainActivity.showOrderDetail`, `MainNavigationRouter`, and the orders nav graphs. It was never read by the destination.

### Test Steps
1. Sign in to a self-hosted site using site credentials (Application Passwords).
2. Trigger a new order push notification on that site.
3. Tap the notification and verify the app navigates directly to the order detail screen without attempting a site switch.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
@@ -1034,7 +1034,6 @@ class MainActivity :
         intent.data = null
         showOrderDetail(
             orderId = event.uniqueId,
-            remoteNoteId = event.remoteNoteId,
             launchedFromNotification = true
         )
     }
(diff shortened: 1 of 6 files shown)


AdamGrzybkowski (approved):
All good! Thanks for the extra fix 👍

PR #15698, opened 2026-04-20
[WOOMOB-2750] Hide unread reviews filter for Woo-registered sites

irfano:
### Description
Fixes WOOMOB-2750

This updates the Reviews screen so the \"Unread reviews only\" filter is hidden for Woo-registered sites.

Unread review state is only supported while reviews are still backed by WP.com notifications. Once a site is Woo-registered, the unread filter is not supported and the screen falls back to the regular reviews list.

### Test Steps
1. Use a site that is registered in Woo only.
2. Open Reviews.
3. Verify the \"Unread reviews only\" filter is hidden.

4. Use a site that is registered in WP.com only.
5. Open Reviews.
6. Verify the \"Unread reviews only\" filter is visible.

### Images/gif
N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the \"[Internal]\" label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
@@ -8,7 +8,9 @@ import com.woocommerce.android.model.ActionStatus
 import com.woocommerce.android.model.ProductReview
 import com.woocommerce.android.model.RequestResult
 import com.woocommerce.android.notifications.UnseenReviewsCountHandler
+import com.woocommerce.android.notifications.push.PushNotificationRegistrationStatus
 import com.woocommerce.android.tools.NetworkStatus
+import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.reviews.ReviewListViewModel.ReviewListEvent.MarkAllAsRead
 import com.woocommerce.android.ui.reviews.domain.MarkAllReviewsAsSeen
 import com.woocommerce.android.viewmodel.BaseUnitTest
@@ -27,6 +29,7 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.model.SiteModel
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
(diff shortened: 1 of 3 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
```diff
@@ -35,19 +38,36 @@ class ReviewListViewModelTest : BaseUnitTest() {
...
+    private lateinit var savedState: SavedStateHandle
     private lateinit var viewModel: ReviewListViewModel
 
     @Before
     fun setup() {
+        doReturn(true).whenever(networkStatus).isConnected()
+        testBlocking {
+            whenever(pushNotificationRegistrationStatus.invoke(SITE_ID))
+                .thenReturn(PushNotificationRegistrationStatus.Status.UNREGISTERED)
+        }
```
hichamboushaba: np, using `testBlocking` here is unusual. I would assumed `runBlocking`, but better, why not move the mocking logic above:

```kotlin
    private val pushNotificationRegistrationStatus: PushNotificationRegistrationStatus = mock {
        on { invoke(SITE_ID) } doReturn PushNotificationRegistrationStatus.Status.UNREGISTERED
    }
```

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
```diff
@@ -35,19 +38,36 @@ class ReviewListViewModelTest : BaseUnitTest() {
...
+    private val siteModel: SiteModel = mock {
+        on { siteId } doReturn SITE_ID
+    }
+    private val selectedSite: SelectedSite = mock {
+        on { getIfExists() } doReturn siteModel
+    }
+    private val pushNotificationRegistrationStatus: PushNotificationRegistrationStatus = mock()
 
     private val reviews = ProductReviewTestUtils.generateProductReviewList()
+    private lateinit var savedState: SavedStateHandle
```
hichamboushaba: np, I don't think this savedState property is used anywhere, we can remove it.

hichamboushaba (approved):
Thanks @irfano, works well.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
irfano: Fixed in ca53bff1c2b

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/reviews/ReviewListViewModelTest.kt
irfano: Fixed in a5843c9eeae

PR #15703, opened 2026-04-20
[WOOMOB-2736] Add transport layer — NSD + TLS + JSON protocol

kidinov:
### Description

Fixes WOOMOB-2736

Builds the transport foundation for Woo POS Remote Tap-to-Pay: a length-prefixed JSON protocol over a TLS socket pinned by a SHA-256 cert fingerprint, with NSD for local-network discovery. Used later by both the phone (Card Reader Mode) and the tablet (Woo POS connect dialog).

New files under `libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/`:
- `RemoteReaderMessage` — sealed class for the 5 message types.
- `RemoteReaderProtocol` — Gson framing with a `type` discriminator + 4-byte big-endian length prefix, max 64 KB.
- `RemoteReaderFingerprint` — SHA-256 helper + 4-char suffix for visual pairing safety.
- `RemoteReaderNsd` — `NsdManager` wrapper. Service type `_wooposremote._tcp.`, TXT keys `fp` and `ver`. Fail-closes on version mismatch.
- `RemoteReaderTlsServer` — phone side. EC keypair + self-signed cert via AndroidKeyStore (`KeyGenParameterSpec`); random alias per session, deleted on close.
- `RemoteReaderTlsClient` — tablet side. TrustManager pinned to a single SHA-256 fingerprint, no CA fallback.
- `RemoteReaderConnection` — thin TLS socket wrapper with `send(msg)` + `receive(): Flow<msg>`.

**Zero new third-party dependencies** — Gson (already an `api` dep via FluxC); AndroidKeyStore handles self-signed cert generation natively. No new permissions.

Target: `feature/remote-tap-to-pay` (long-lived integration branch; final merge to trunk behind the `REMOTE_TAP_TO_PAY` flag from WOOMOB-2735).

### Test Steps

1. Build the module: \`./gradlew :libs:cardreader:assembleDebug\` — should succeed.
2. Run unit tests: \`./gradlew :libs:cardreader:testDebugUnitTest\` — 5 protocol round-trip tests + 1 fingerprint fixture test must pass.
3. Detekt: \`./gradlew detektAll --auto-correct\` — no new violations.
4. No real-device TLS/NSD smoke test here — those are validated end-to-end by the downstream issues (WOOMOB-2739 on phone side, WOOMOB-2742 on tablet side).

### Images/gif

N/A

### DIFF
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
@@ -12,6 +12,7 @@ import java.io.DataInputStream
 import java.io.DataOutputStream
 import java.io.EOFException
 import java.net.Socket
+import java.net.SocketException
 
 internal class CardReaderRemoteConnection internal constructor(
     private val socket: Socket,
@@ -34,7 +35,7 @@ internal class CardReaderRemoteConnection internal constructor(
         while (!socket.isClosed) {
             val next = runCatching { protocol.read(input) }
                 .getOrElse { err ->
-                    if (err is EOFException) return@flow
+                    if (err is EOFException || err is SocketException) return@flow
                     throw err
                 }
             emit(next)
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
@@ -13,16 +13,16 @@ import java.io.DataOutputStream
 import java.io.EOFException
 import java.net.Socket
 
-class RemoteReaderConnection internal constructor(
+internal class CardReaderRemoteConnection internal constructor(
     private val socket: Socket,
-    private val protocol: RemoteReaderProtocol = RemoteReaderProtocol(),
+    private val protocol: CardReaderRemoteProtocol = CardReaderRemoteProtocol(),
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) : AutoCloseable {
     private val output = DataOutputStream(socket.getOutputStream())
     private val input = DataInputStream(socket.getInputStream())
     private val writeLock = Mutex()
 
-    suspend fun send(msg: RemoteReaderMessage) {
+    suspend fun send(msg: CardReaderRemoteMessage) {
         writeLock.withLock {
             withContext(ioDispatcher) {
                 protocol.write(output, msg)
@@ -30,7 +30,7 @@ class RemoteReaderConnection internal constructor(
         }
     }
 
-    fun receive(): Flow<RemoteReaderMessage> = flow {
+    fun receive(): Flow<CardReaderRemoteMessage> = flow {
         while (!socket.isClosed) {
             val next = runCatching { protocol.read(input) }
                 .getOrElse { err ->
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
@@ -4,7 +4,7 @@ import java.security.MessageDigest
 import java.security.cert.X509Certificate
 import java.util.Base64
 
-object RemoteReaderFingerprint {
+internal object CardReaderRemoteFingerprint {
     fun sha256(cert: X509Certificate): ByteArray {
         val digest = MessageDigest.getInstance("SHA-256")
         return digest.digest(cert.encoded)
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
@@ -12,11 +12,11 @@ internal object CardReaderRemoteFingerprint {
 
     fun sha256Base64(cert: X509Certificate): String = encodeBase64(sha256(cert))
 
-    fun suffix4(cert: X509Certificate): String = suffix4FromBytes(sha256(cert))
+    fun pairingCode(cert: X509Certificate): String = pairingCodeFromBytes(sha256(cert))
 
-    fun suffix4FromBase64(fingerprintBase64: String): String = suffix4FromBytes(decodeBase64(fingerprintBase64))
+    fun pairingCodeFromBase64(fingerprintBase64: String): String = pairingCodeFromBytes(decodeBase64(fingerprintBase64))
 
-    private fun suffix4FromBytes(fingerprint: ByteArray): String {
+    private fun pairingCodeFromBytes(fingerprint: ByteArray): String {
         require(fingerprint.size >= 2) { "Fingerprint too short" }
         return fingerprint.take(2).joinToString("") { "%02X".format(it) }
     }
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteNsd.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteNsd.kt
@@ -3,7 +3,9 @@ package com.woocommerce.android.cardreader.remote
 import android.content.Context
 import android.net.nsd.NsdManager
 import android.net.nsd.NsdServiceInfo
+import android.os.Build
 import android.util.Log
+import androidx.annotation.RequiresApi
 import kotlinx.coroutines.CancellableContinuation
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
@@ -15,11 +17,11 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
 import java.io.IOException
 import java.net.InetAddress
+import java.util.concurrent.Executor
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.random.Random
 
-@Suppress("DEPRECATION")
 internal class CardReaderRemoteNsd(
     context: Context,
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
@@ -100,7 +102,7 @@ internal class CardReaderRemoteNsd(
             }
 
             override fun onServiceFound(service: NsdServiceInfo) {
-                nsdManager.resolveService(service, buildResolveListener(this@callbackFlow::trySend))
+                resolve(service) { resolved -> trySend(resolved) }
             }
 
             override fun onServiceLost(service: NsdServiceInfo) {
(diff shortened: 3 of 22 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          5 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CardReaderRemoteConnection</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CardReaderRemoteNsd</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CardReaderRemoteNsdRegistration</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CardReaderRemoteTlsServer</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsClient.kt
```diff
@@ -0,0 +1,54 @@
...
+            init(null, arrayOf(trustManager), SecureRandom())
+        }
+        val socket = (sslContext.socketFactory.createSocket(host, port) as SSLSocket).apply {
+            soTimeout = SESSION_READ_TIMEOUT_MILLIS
+        }
+        socket.startHandshake()
+        RemoteReaderConnection(socket, ioDispatcher = ioDispatcher)
+    }
+
+    private class PinnedFingerprintTrustManager(
```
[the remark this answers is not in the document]
kidinov: False positive for this use case. This is intentional fingerprint pinning for peer-to-peer TLS — the phone uses an ephemeral self-signed cert (no CA chain exists), so CA validation is not applicable. The implementation is stricter than the default: it rejects any cert whose SHA-256 does not exactly match the pin obtained out-of-band via NSD. `checkClientTrusted` and `checkServerTrusted` both throw on any mismatch.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/RemoteReaderNsd.kt
```diff
@@ -0,0 +1,187 @@
...
+            override fun onServiceRegistered(registered: NsdServiceInfo) {
+                registrationContinuation?.takeIf { it.isActive }?.resume(registered)
+                registrationContinuation = null
+            }
+
+            override fun onRegistrationFailed(info: NsdServiceInfo?, errorCode: Int) {
+                val cont = registrationContinuation
+                registrationContinuation = null
+                if (cont != null && cont.isActive) {
+                    cont.resume(info ?: serviceInfo)
```
malinajirka: 💡 It looks like we are reporting success on failed registration, should we invoke something like resumeWithException instead to propagate the error?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServer.kt
```diff
@@ -0,0 +1,121 @@
...
+    val port: Int
+        get() = requireNotNull(serverSocket) { "Server not started" }.localPort
+
+    val fingerprint: String
+        get() = RemoteReaderFingerprint.sha256Base64(
+            requireNotNull(certificate) { "Server not started" }
+        )
+
+    suspend fun start(): Unit = withContext(ioDispatcher) {
+        sweepOrphanAliases()
```
malinajirka, line 38: 💡 Two things worth tightening here:

1. `start()` has no guard against being called twice on the same instance. If it is, the second call silently leaks the first `serverSocket` and — because `sweepOrphanAliases()` matches purely on `ALIAS_PREFIX` — wipes the first session's private key from the keystore mid-session, breaking any in-flight TLS handshake.
2. The sweep is intentionally aggressive, which is the right move for crash-recovery. But it also means two concurrent instances of this class (or anything else sharing the prefix) will nuke each other's live keys.

I'd at least add a `check(serverSocket == null) { "Server already started" }` at the top and keep the sweep as-is — that covers both the double-start footgun and the legitimate orphan-cleanup case. Wdyt?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun sha256Base64(cert: X509Certificate): String = encodeBase64(sha256(cert))
+
+    fun suffix4(cert: X509Certificate): String = suffix4FromBytes(sha256(cert))
+
+    fun suffix4FromBase64(fingerprintBase64: String): String = suffix4FromBytes(decodeBase64(fingerprintBase64))
+
+    private fun suffix4FromBytes(fingerprint: ByteArray): String {
+        require(fingerprint.size >= 2) { "Fingerprint too short" }
+        return fingerprint.take(2).joinToString("") { "%02X".format(it) }
```
malinajirka, line 21: 🔍 np: `suffix4` actually takes the *first* two bytes of the hash (a prefix), not the last two — the test at `RemoteReaderFingerprintTest.kt:14` locks in `74f81f... -> "74F8"`. Nothing broken, but the name will mislead anyone later matching the pairing UX against the tail of the fingerprint. Maybe rename to `prefix4` / `shortCode4`? Wdyt?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteNsd.kt
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            override fun onServiceLost(service: NsdServiceInfo) {
+                Log.d(TAG, "NSD service lost: ${service.serviceName}")
+            }
+        }
+
+        runCatching {
+            nsdManager.discoverServices(SERVICE_TYPE, NsdManager.PROTOCOL_DNS_SD, discoveryListener)
+        }.onFailure { Log.w(TAG, "Failed to start NSD discovery", it) }
```
malinajirka: 💡 If `discoverServices` throws synchronously (e.g. listener-already-registered, SecurityException), the `callbackFlow` stays open forever with no emissions — the collector just hangs. Worth closing the flow with the error so the UI can surface it:

```suggestion
        runCatching {
            nsdManager.discoverServices(SERVICE_TYPE, NsdManager.PROTOCOL_DNS_SD, discoveryListener)
        }.onFailure { close(it) }
```

Same shape for `onStartDiscoveryFailed` at line 91 — today it logs and the flow keeps listening to a discovery that never started; I'd `close(IOException("NSD start discovery failed: $errorCode"))` there too.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsClient.kt
```diff
@@ -0,0 +1,54 @@
...
+        private val pinnedFingerprintBase64: String,
+    ) : X509TrustManager {
+        override fun checkClientTrusted(chain: Array<out X509Certificate>, authType: String) {
+            throw CertificateException("Client authentication not supported")
+        }
+
+        override fun checkServerTrusted(chain: Array<out X509Certificate>, authType: String) {
+            val leaf = chain.firstOrNull() ?: throw CertificateException("Empty certificate chain")
+            val presented = RemoteReaderFingerprint.sha256Base64(leaf)
+            if (presented != pinnedFingerprintBase64) {
```
malinajirka, line 43: 💡Should we add some unit tests for this class? Feels like the the most important class from security perspective. Wdyt?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
```diff
@@ -0,0 +1,47 @@
...
+                protocol.write(output, msg)
+            }
+        }
+    }
+
+    fun receive(): Flow<RemoteReaderMessage> = flow {
+        while (!socket.isClosed) {
+            val next = runCatching { protocol.read(input) }
+                .getOrElse { err ->
+                    if (err is EOFException) return@flow
```
malinajirka: 📓 Calling close() throws SocketException("Socket closed") => do we need to treat both EOFException as well as SocketException as expected?

malinajirka (approved):

Tbh I'm a bit worried about introducing a relatively low-level custom communication protocol - especially considering it'll be used for payment collection. Have you considered using one of the existing solutions, such as Google's Nearby Connections API?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/RemoteReaderNsd.kt
kidinov: good catch - fixed 57b2ee8a978

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServer.kt
kidinov: good catch - fixed 9bc1312a31a

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteFingerprint.kt
kidinov: good catch - fixed 86399e94b16

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteNsd.kt
kidinov: good catch - fixed 2c8a76d264f

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsClient.kt
kidinov: good catch - fixed 0cfddf334a9

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
kidinov: good catch - fixed 190504f7bc6

kidinov:
@malinajirka thank you for the review!

> Tbh I'm a bit worried about introducing a relatively low-level custom communication protocol - especially considering it'll be used for payment collection. Have you considered using one of the existing solutions, such as Google's Nearby Connections API?

I just started directly with a similar approach that Josh took for iOS. And as iPad + Android phone can be the primary use case, this probably make sense to go the same, a bit more low level but more flexible way that can work between platforms

In any case, I hope that I will manage to abstract it the way that this level could be replaced with something else easy - without touching too much in the buisness logic

PR #15704, opened 2026-04-21
[WOOMOB-2737] Add CompositeConnectionTokenProvider for Stripe SDK

kidinov:
### Description

Fixes WOOMOB-2737

Adds a `CompositeConnectionTokenProvider` that delegates the Stripe Terminal SDK's `fetchConnectionToken` call to one of two underlying providers at runtime. The SDK singleton is never re-initialized — only the active delegate swaps atomically.

This is the single most important non-regression component of the Remote Tap-to-Pay project: it lets the new Card Reader Mode flow share the same SDK singleton with the existing self-tap TTP and BT reader flows without any changes to those flows.

**Components:**
- `CompositeConnectionTokenProvider` — atomic delegate swap via `AtomicReference`. Exposes `use(provider)` and `useDefault()` on a narrow lifecycle contract documented in the class KDoc.
- `RemoteTokenChannelProvider` — a `ConnectionTokenProvider` backed by a rendezvous `Channel<String>`. Used later by Card Reader Mode to feed tokens received over the local TLS socket. Fail-fasts on `fetchConnectionToken` after `close()` so teardown doesn't silently hang the SDK.
- Wiring: `CardReaderManagerFactory` constructs the composite wrapping the existing `TokenProvider`. `CardReaderManager.connectionTokenProvider` exposes the composite so Card Reader Mode can swap delegates from its activity lifecycle.

**Non-regression:** with the feature flag off on both surfaces, the composite always points at the default (backend-fetched) provider, and the existing self-tap TTP and BT reader flows are byte-identical to trunk.

Target: \`feature/remote-tap-to-pay\` (long-lived integration branch; final merge to trunk ships everything behind the \`REMOTE_TAP_TO_PAY\` flag from WOOMOB-2735).

### Test Steps

1. Build: \`./gradlew :libs:cardreader:assembleDebug\` — BUILD SUCCESSFUL.
2. Tests: \`./gradlew :libs:cardreader:testDebugUnitTest\` — 3 new tests on `CompositeConnectionTokenProvider` (default routes to default delegate; `use(alt)` routes to alt; `useDefault()` restores default) must pass.
3. Non-regression smoke (manual): run the existing self-tap TTP flow on a debug build — payment should work identically to trunk (the composite transparently delegates to the backend provider).

### Images/gif

N/A

### DIFF
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 
 class RemoteTokenChannelProvider : ConnectionTokenProvider, AutoCloseable {
@@ -19,6 +20,10 @@ class RemoteTokenChannelProvider : ConnectionTokenProvider, AutoCloseable {
     }
 
     override fun fetchConnectionToken(callback: ConnectionTokenCallback) {
+        if (!scope.isActive) {
+            callback.onFailure(ConnectionTokenException("Remote token provider is closed"))
+            return
+        }
         scope.launch {
             runCatching { tokens.receive() }
                 .onSuccess { callback.onSuccess(it) }
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
@@ -8,8 +8,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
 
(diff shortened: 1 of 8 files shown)


libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+    suspend fun supply(token: String) {
+        tokens.send(token)
+    }
+
+    override fun fetchConnectionToken(callback: ConnectionTokenCallback) {
+        if (!scope.isActive) {
+            callback.onFailure(ConnectionTokenException("Remote token provider is closed"))
+            return
+        }
+        scope.launch {
```
samiuelson: 💡 I didn't test it enough, but this looks like a potential place for race condition - in case scope gets canceled between `if` {} and scope.launch, the callback would never be invoked. 

  Maybe this would be safer:
```kotlin
override fun fetchConnectionToken(callback: ConnectionTokenCallback) {
        val job = scope.launch {
            runCatching { tokens.receive() }
                .onSuccess { callback.onSuccess(it) }
                .onFailure { err ->
                    callback.onFailure(
                        ConnectionTokenException(err.message.orEmpty(), err)
                    )
                }
        }
        if (!job.isActive && !job.isCompleted) {
            callback.onFailure(ConnectionTokenException("Remote token provider is closed"))
        }
    }
```

samiuelson (approved):
LGTM! I added comment about one thing I'm not sure about.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/connection/RemoteTokenChannelProvider.kt
kidinov: Good catch — switched to `invokeOnCompletion` + `AtomicBoolean` guard so the callback is always invoked exactly once, even if the scope is cancelled before the `launch` block runs. Fixed in 9036d66d878.

PR #15683, opened 2026-04-17
[WOOMOB-2470] POS Orders: Delete old VM, clean up shared state (3/3)

malinajirka:
> **PR Series**: #15681 (1/3) → #15682 (2/3) → #15683 (3/3)

### Description

Fixes WOOMOB-2470

Note: It is expected that CI is red in this branch - the 3rd PR is mergable and testable.

Final cleanup PR: deletes the now-empty `WooPosOrdersViewModel` and its test, cleans `WooPosOrdersState` to a shared-types object, simplifies the data source API, and applies code review fixes.

Known Issues 
1. Only first page is inserted into in-memory cache - fixed in https://github.com/woocommerce/woocommerce-android/pull/15705.
2. Selecting a different order after a successful refund (before the list refresh finishes), doesn't update the state of the refunded order in the list - fixed in https://github.com/woocommerce/woocommerce-android/pull/15706

#### Commits

1. **Delete WooPosOrdersViewModel** — The old VM is an empty shell after PRs 1 and 2 moved all methods out.
3. **Clean WooPosOrdersState to shared types object** — Remove old state hierarchy (`Content`, `Loading`, `Error`, `Empty`, `DialogState`), remove dead `OrderDetailsViewState.Lazy` class, convert to `object` holding only shared types used by both VMs.
4. **Code review fixes** — Make `onEmailReceiptButtonClicked`/`onIssueRefundButtonClicked` private, add `OrderAction`/`Order` imports to replace FQN usage, add 4 missing critical tests (`retryLoadOrder`, email receipt refresh, refund dialog refresh, no-op retry).
5. **Remove refund fetching from data source list/search/loadMore APIs** — Simplify `LoadOrdersResult` and `SearchOrdersResult` from `Map<Order, RefundsFetchResult>` to `List<Order>`. Remove `fetchRefundsForOrders()` from the data source entirely — refunds are now loaded on-demand by the detail VM only when an order is selected.

**Based on PR 2**: `issue/WOOMOB-2470-pos-orders-split-detail-vm`

### Test Steps

1. Open POS Orders — full regression: list, detail, search, pagination, refresh
2. Issue a partial refund — verify order updates in both list and detail
3. Issue a full refund — verify status changes to "Refunded"
4. Test single-order mode from Bookings → "View order"

### Images/gif

N/A — no visual changes, this is a refactoring PR.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15705, opened 2026-04-21
[WOOMOB-2470] POS Orders: Cache paginated orders for instant detail opens (4/4)

malinajirka:
### Description

Fixes WOOMOB-2470

Stacks on #15683.

In POS Orders, tapping a row in the first page (≤25 orders) opens the detail instantly, but tapping any row further down the list takes ~hundreds of ms on first tap. The reason: `WooPosOrdersInMemoryCache` is only populated by `WooPosOrdersDataSource.loadOrders()` (first page). Rows loaded via `loadMore()` miss the cache, so `getOrderById` falls through to `fetchSingleOrder` — a network call.

Fix: append the result of `loadMore()` into the same cache so `getOrderById` finds paginated rows too. Added `WooPosOrdersInMemoryCache.appendAll(orders)`, atomic via `AtomicReference.updateAndGet`. The append is skipped for paginated search results (`searchQuery != null`) — they're ephemeral and shouldn't leak into the main orders cache.

Freshness is preserved by the existing invalidation points: `loadOrders()` replaces the whole cache via `setAll`, so re-enter/pull-to-refresh evicts paginated entries; and `refreshOrderById` (called after a refund) updates the matching entry in place. The staleness surface is the same one first-page orders already have.

### Test Steps

1. Open POS → Orders. Scroll until pagination loads page 2 (rows 26+).
2. Tap a row from page 2 → the detail should open instantly (before this PR, it took a visible fraction of a second).
3. Pull-to-refresh and repeat — first tap fetches fresh, subsequent taps within the session are instant.
4. Search for a term, paginate within search results, tap a row — should work. Exit search — the main Orders list should not contain search results.
5. Issue a refund on a page-2 row — both list badge and detail pane should update.

### Images/gif

N/A — performance-only change, no UI differences.

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSourceTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSourceTest.kt
@@ -382,6 +382,71 @@ class WooPosOrdersDataSourceTest {
     }
 
     @Test
+    fun `when loadMore without query succeeds, then new orders appended to cache`() = runTest {
+        // GIVEN
+        whenever(ordersCache.getAll()).thenReturn(emptyList())
+        val firstPayload = WCOrderStore.FetchOrdersResponsePayload(
+            site = siteModel,
+            ordersWithMeta = emptyList(),
+            canLoadMore = true
+        )
+        whenever(orderRestClient.fetchOrders(any(), any(), eq(1), any(), any(), anyOrNull(), any(), isNull(), eq(8)))
+            .thenReturn(firstPayload)
+        sut.loadOrders().toList(mutableListOf())
+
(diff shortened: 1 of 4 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersDataSourceTest.kt
```diff
@@ -381,6 +381,71 @@ class WooPosOrdersDataSourceTest {
         assertThat(sut.hasMorePages).isFalse
     }
 
+    @Test
```
malinajirka, line 384: 📓 The tests are noisy but follow the other tests in this file.

samiuelson (approved):
LGTM!

samiuelson (approved):
LGTM!

PR #15706, opened 2026-04-21
[WOOMOB-2470] Fix stale POS list row after refund when selection changes

malinajirka:
Fixes WOOMOB-2470

### Description

Fixes a regression introduced by #15683 (the VM split).

**The bug:** after issuing a refund (or returning from the email-receipt flow), the detail VM refreshes the order in the background and updates both the detail pane and the corresponding list row. If the user tapped a different order before that refresh finished, the original order's list row was left showing pre-refund data — status, totals, etc. — until the next pull-to-refresh. Trunk did not have this problem; #15683 accidentally tied the list-row refresh to the "detail pane still shows this order" condition.

**The fix** is in two commits:

1. **Refresh list row after order update even if selection has changed.** Decouples the list-row notify from the detail-state update so the list row is refreshed whenever the cache is updated, regardless of what the detail pane is showing.
2. **Skip stale detail-pane work when selection changes during refresh.** The fetch + list-row notify are now wrapped in `withContext(NonCancellable)`, guaranteeing they run to completion even when the user moves on. The expensive detail-pane work that follows (refund re-fetch, view-state mapping) is cancellable, so it's skipped when it would be discarded anyway. `notifyOrderRefreshed` switches to a non-suspending `tryEmit` on a buffered `SharedFlow`, which also removes a latent risk of events being dropped if a collector wasn't immediately ready.

### Test Steps

1. Install on a tablet and open POS → historical orders.
2. Pick an order that shows **Issue refund** (typically a Completed order with line items). Issue a refund and dismiss the dialog.
3. Before the list row visibly updates, tap a different order in the list.
4. Confirm the first order's row now reflects the refund (status / total updated). Without the fix it stays on the pre-refund values until pull-to-refresh.
5. Repeat the same pattern with **Email receipt**: send a receipt, return to the screen, then quickly select a different order. The original order's row should still reflect any changes.

### Images/gif

N/A

- [ ] I have considered if this change warrants release notes and have added them to \`RELEASE-NOTES.txt\` if necessary. Use the "[Internal]" label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersCoordinator.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersCoordinator.kt
@@ -1,6 +1,7 @@
 package com.woocommerce.android.ui.woopos.orders
 
 import dagger.hilt.android.scopes.ActivityRetainedScoped
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharedFlow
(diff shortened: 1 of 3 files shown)


samiuelson (approved):
LGTM!

💡 One thing I noticed, not related to changes from this PR, is that `loadOrder()` is public. Is it going to be called from outside the VM?

PR #15708, opened 2026-04-21
[WOOMOB-2738] Add ReadyToPair + WaitingForPayment states to card reader payment dialog

kidinov:
### Description

Fixes WOOMOB-2738

Part of the Woo POS **Remote Tap-to-Pay** feature ([Linear project](https://linear.app/a8c/project/woo-pos-remote-tap-to-pay-4720ed488602)). Extends the existing TTP payment dialog so it can also render phone-as-reader states, without adding a new ViewModel or a new UI surface.

**What changes:**

- Two new view states `RemoteTapToPayReadyToPair(deviceName, fingerprintSuffix)` and `RemoteTapToPayWaitingForPayment(tabletName)` extend the existing `ViewState` sealed class. They render in the same dialog slots (icon, header, subtitle, footer button) as the other states.
- `onCancelClicked()` on `CardReaderPaymentViewModel` branches on current state: the two new subtypes emit a new `ExitCardReaderMode` event via the bridge; everything else keeps its current per-state cancel behavior.
- New `RemoteTapToPayReaderStateBridge` (`@Singleton`, Hilt-injected) owns the state-override `LiveData`, the VM-originated events, and the `push` / `clear` / `emitEvent` API. The VM injects the bridge and composes its `LiveData` sources inside the existing `MediatorLiveData` for `viewStateData` and `event`. WOOMOB-2739's `CardReaderModeSession` will inject the same bridge and call `push` / `clear` directly - no VM-mediated push API needed.

**Package layout:**

- `ui/payments/cardreader/payment/RemoteTapToPayPaymentViewState.kt` - the two state classes. Stays in the same package as `ViewState` because Kotlin requires sealed-class subtypes to live in the same package as the parent.
- `ui/payments/cardreader/payment/remote/ExitCardReaderMode.kt` - the new event.
- `ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt` - the bridge. Future Remote TTP code lands in the same `remote/` subpackage.

**Non-regression:**

With no `push()` call on the bridge, `viewStateData` and `event` pass through the existing `paymentController`-derived values unchanged. Existing self-tap TTP and BT reader flows are not touched.

**Scope:**

This PR is **scaffolding** - no production caller uses the bridge yet. Wiring happens in WOOMOB-2739 (`CardReaderModeSession` + `CardReaderModeActivity`), which will inject the same `@Singleton` bridge and push / clear states as messages arrive over the TLS socket. The PR targets the long-lived `feature/remote-tap-to-pay` branch and the feature ships behind the `REMOTE_TAP_TO_PAY` flag (off by default).

### Test Steps

1. Check out this branch, run the app with `REMOTE_TAP_TO_PAY` off, and pay an order via self-tap TTP on a phone. Dialog should render exactly as trunk - no new states appear.
2. If a Bluetooth card reader is handy, connect and pay an order with it. Same thing - no change.
3. Read `CardReaderPaymentViewModel.kt`: confirm that with no `push` call on the bridge, `viewStateData` simply forwards whatever `paymentController.paymentState` emits, and `event` forwards only mapped controller events.

### Images/gif

N/A - no visible UI change in this PR (states only render when the bridge is pushed, which lands in WOOMOB-2739).

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
@@ -28,6 +28,6 @@ class RemoteTapToPayReaderStateBridge {
 
     fun emitEvent(event: MultiLiveEvent.Event) {
         event.isHandled = false
-        _events.postValue(event)
+        _events.value = event
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
@@ -31,6 +31,6 @@ class RemoteTapToPayReaderStateBridge @Inject constructor() {
 
     fun emitEvent(event: MultiLiveEvent.Event) {
         event.isHandled = false
-        _events.value = event
+        _events.postValue(event)
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
@@ -1,0 +1,33 @@
+package com.woocommerce.android.ui.payments.cardreader.payment.remote
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayReadyToPair
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayWaitingForPayment
+import com.woocommerce.android.ui.payments.cardreader.payment.ViewState
+import com.woocommerce.android.viewmodel.MultiLiveEvent
+
+class RemoteTapToPayReaderStateBridge {
+
+    private val _stateOverride = MutableLiveData<ViewState?>(null)
+    val stateOverride: LiveData<ViewState?> = _stateOverride
+
+    private val _events = MutableLiveData<MultiLiveEvent.Event>()
+    val events: LiveData<MultiLiveEvent.Event> = _events
+
+    fun push(state: ViewState) {
+        require(state is RemoteTapToPayReadyToPair || state is RemoteTapToPayWaitingForPayment) {
+            "push only accepts RemoteTapToPayReadyToPair or RemoteTapToPayWaitingForPayment"
+        }
+        _stateOverride.postValue(state)
+    }
+
+    fun clear() {
+        _stateOverride.postValue(null)
+    }
+
+    fun emitEvent(event: MultiLiveEvent.Event) {
+        event.isHandled = false
+        _events.postValue(event)
+    }
+}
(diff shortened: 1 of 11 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
```diff
@@ -0,0 +1,36 @@
...
+    val events: LiveData<MultiLiveEvent.Event> = _events
+
+    fun push(state: ViewState) {
+        require(state is RemoteTapToPayReadyToPair || state is RemoteTapToPayWaitingForPayment) {
+            "push only accepts RemoteTapToPayReadyToPair or RemoteTapToPayWaitingForPayment"
+        }
+        _stateOverride.postValue(state)
+    }
+
+    fun clear() {
```
samiuelson, line 28: ❓ I think clear should be called in CardReaderPaymentViewModel::onCleared, otherwise the singleton will leak stale view state.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.lifecycle.MutableLiveData
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayReadyToPair
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayWaitingForPayment
+import com.woocommerce.android.ui.payments.cardreader.payment.ViewState
+import com.woocommerce.android.viewmodel.MultiLiveEvent
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class RemoteTapToPayReaderStateBridge @Inject constructor() {
```
samiuelson, line 13: 💡 There seems to be thread-safety inconsistency in this class. push()/clear() use postValue (any-thread safe), but emitEvent() uses setValue (main-thread only). Calling emitEvent() from a background thread will crash the app mid-payment 🤔

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayReadyToPair
+import com.woocommerce.android.ui.payments.cardreader.payment.RemoteTapToPayWaitingForPayment
+import com.woocommerce.android.ui.payments.cardreader.payment.ViewState
+import com.woocommerce.android.viewmodel.MultiLiveEvent
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
samiuelson, line 12: What is Singleton needed for?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
kidinov: Good catch — calling `clear()` in `onCleared` so a pushed override doesn't survive into the next VM. Fixed in a6a292bf785.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
kidinov: Good catch — switched `emitEvent` to `postValue` so the whole class is uniformly any-thread-safe. Fixed in 54f3d3b9f98.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/remote/RemoteTapToPayReaderStateBridge.kt
kidinov: It's a shared touchpoint so a producer outside the VM graph can push state/events into the payment VM. In this PR only the VM reads from it, but the producer (card-reader-mode state machine in WOOMOB-2739) lives on the follow-up branch. Both sides need the same instance, hence `@Singleton`.

samiuelson (approved):
LGTM

PR #15709, opened 2026-04-21
[WOOMOB-2741] Add Woo POS discoverability for phone-as-card-reader

kidinov:
### Description

Fixes WOOMOB-2741

Part of the Woo POS **Remote Tap-to-Pay** feature ([Linear project](https://linear.app/a8c/project/woo-pos-remote-tap-to-pay-4720ed488602)). Adds a discoverability layer inside the POS card-reader scanning dialog so merchants know a phone can also act as a card reader, with a short explainer they can open on demand.

**What changes:**

- `WooPosRemoteReaderTipStrip` - a short, centered, underlined link rendered at the **bottom** of the scanning content: _"Use an Android phone as a card reader"_. Tapping it opens the explainer.
- `WooPosRemoteReaderExplainerContent` - a new instructional block that explains the setup + requirements. Renders **inside the existing `WooPosCardReaderConnectionDialog`** (same `WooPosDialogWrapper`), not as a separate modal.
- Added a new `WooPosCardReaderConnectionState.RemoteTapToPayExplainer(onDismissClicked)` to the sealed hierarchy. The controller transitions Scanning ↔ Explainer via `showRemoteTapToPayExplainer()` / `hideRemoteTapToPayExplainer()`. The dialog composable picks up the state through the existing `AnimatedContent` slot.
- `WooPosCardReaderConnectionController` reads `FeatureFlagRepository.isEnabled(REMOTE_TAP_TO_PAY)` once at init and exposes the result via `Scanning.isRemoteTapToPaySupported` - VM stays lean, no flag injected there.
- `ScanningContent` hides the tip when `currentWooPosBreakpoint() == WooPosBreakpoint.Phone` so the phone POS prototype doesn't show it.
- The controller guards all Scanning transitions (`enterScanningState()`) so that discovery events re-emitted by the Stripe SDK don't yank the user out of the explainer.

**Explainer design (per codex consult):**

- Title `Heading` bold, centered: _"No card reader? No problem"_
- Intro `BodyLarge` centered, `onSurfaceVariant`: _"Use the Woo app on your phone to take card payments."_
- Two left-aligned sections: _"Set up your phone"_ / _"Check compatibility"_ - heading is `BodyLarge` `SemiBold`, body is `BodyMedium` `onSurfaceVariant`
- Button `Got it` full-width, centered horizontally
- No X close button (button is the dismissal)
- Back press dismisses the explainer first, then normal back behavior

**Non-regression:**

With `REMOTE_TAP_TO_PAY` off, `Scanning.isRemoteTapToPaySupported` is `false` - no tip, no explainer path, `ScanningContent` falls through to the original body. Byte-identical to trunk.

### Test Steps

1. Enable the `REMOTE_TAP_TO_PAY` flag in the debug menu. On a tablet, open Woo POS and trigger the card reader connect dialog from any of the three entry points (Settings → Hardware → Card Readers → Connect, Home screen Connect shortcut, Totals _Connect to collect_). Verify the tip appears below the scanning illustration.
2. Tap the tip. Verify the explainer renders inside the same dialog (no separate modal pops). Verify there is no X close button and `Got it` dismisses it.
3. While in the explainer, wait a moment for discovery events / background emissions. Verify the explainer stays open.
4. Press back while the explainer is open - verify it dismisses back to scanning.
5. Disable the flag. Open the same dialog. Verify the tip is gone and layout matches trunk.
6. On a phone-sized POS window, with the flag on, verify the tip is hidden.

### Images/gif

https://github.com/user-attachments/assets/eefa73f7-2b87-4f91-a36b-f7b08a8bc16b


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -71,13 +77,34 @@ class WooPosCardReaderConnectionController(
...
+
+    fun hideRemoteTapToPayExplainer() {
+        if (_state.value !is WooPosCardReaderConnectionState.RemoteTapToPayExplainer) return
+        _state.value = WooPosCardReaderConnectionState.Scanning(
+            isRemoteTapToPaySupported = isRemoteTapToPayEnabled,
+        )
+    }
+
+    private fun enterScanningState() {
+        if (_state.value is WooPosCardReaderConnectionState.RemoteTapToPayExplainer) return
```
malinajirka, line 97: 💡 The guard here prevents re-entering `Scanning` while the Explainer is open, but it doesn't protect the other state transitions. While the user is reading the Explainer, `discoveryJob` is still running in the background, and `handleReadersFound` / `handleDiscoveryEvent.Failed` write directly to `_state.value` with `ReaderFound` / `MultipleReadersFound` / `Connecting` / `ScanningFailed` without any `is Explainer` check. So if a reader shows up (or discovery fails) during the seconds the user spends reading the tip, the Explainer disappears — which contradicts the PR description's claim that discovery events "don't yank the user out of the explainer." The Stripe SDK re-emitting `Started` is guarded, but `ReadersFound` is equally likely in practice.

Two options I can think of:
1. Cancel `discoveryJob` on `showRemoteTapToPayExplainer()` and restart it on `hideRemoteTapToPayExplainer()`. Makes the invariant obvious.
2. Add the same Explainer early-return in `handleReadersFound` / the `Failed` branch and buffer the latest result so it surfaces on dismiss.

Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/remote/WooPosRemoteReaderTipStrip.kt
```diff
@@ -0,0 +1,46 @@
...
+) {
+    WooPosText(
+        text = stringResource(R.string.woopos_remote_ttp_tip_strip_text),
+        style = WooPosTypography.BodyMedium,
+        color = MaterialTheme.colorScheme.primary,
+        textAlign = TextAlign.Center,
+        textDecoration = TextDecoration.Underline,
+        modifier = modifier
+            .fillMaxWidth()
+            .clickable(onClick = onClick)
```
malinajirka: 🔍 np: `.clickable` has no role, so TalkBack announces the strip as plain text. Something like `.clickable(role = Role.Button, onClick = onClick)` would make its purpose discoverable. Not blocking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
```diff
@@ -449,7 +461,25 @@ private fun CardReaderDialogContent(
 }
 
 @Composable
-private fun ScanningContent() {
+private fun ScanningContent(
+    isRemoteTapToPayEnabled: Boolean,
+    onTipClick: () -> Unit,
+) {
+    val showTip = isRemoteTapToPayEnabled && currentWooPosBreakpoint() != WooPosBreakpoint.Phone
```
malinajirka: np: I'd call it `hint` instead of `tip` just to avoid collision with tipping.

malinajirka (approved):
Thanks @kidinov! Overall looks good - I'd consider adding a bit more unit tests.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
kidinov: Good idea, cancelling `discoveryJob` on show and restarting on hide makes the invariant explicit. Fixed in 1d730b4a89a.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/remote/WooPosRemoteReaderTipStrip.kt
kidinov: Good catch, setting `role = Role.Button` so TalkBack announces the strip as a button. Fixed in cc1667b68fd.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
kidinov: Good idea, renamed `tip` → `hint` across the feature (file, composable, string key, VM method, dialog params) so it no longer collides with the tipping flow. Fixed in 81c61b13072.

PR #15717, opened 2026-04-22
[WOOMOB-2739] Card Reader Mode activity, UI and Payments Hub entry

kidinov:
### Description

Fixes WOOMOB-2739, also fixes WOOMOB-2791 (dark mode).

Adds the UI for Card Reader Mode: a `CardReaderModeActivity` that renders session state via `CardReaderModeViewModel` (Starting → Ready to pair → Waiting for payment), and a flag-gated **Card Reader Mode** row in Payments Hub that launches it.

Also removes the `CardReaderModeStateBridge` singleton introduced in #15708. With the dedicated activity, the bridge had no second consumer left - `CardReaderModeViewModel` collects `session.state` and exposes the view state directly, which also reverts the `CardReaderPaymentViewModel` wiring from #15708 that was never reached at runtime.

**Base:** #15716 (session).

**Known gaps:**

- `CollectPaymentRequest` on the phone side is still stubbed - real payment collection deferred to **WOOMOB-2739-b**.
- Foreground service replacing the dedicated activity - **WOOMOB-2761**.
- Row is gated on `TapToPayAvailabilityStatus().isAvailable && REMOTE_TAP_TO_PAY`, so non-TTP phones never see it.

### Test Steps

1. Debug builds have `REMOTE_TAP_TO_PAY` on. Install on a TTP-capable phone.
2. **My Store → Menu → Payments** - **Card Reader Mode** row appears under "Manage Card Reader".
3. Tap the row → starting spinner, then "Ready to pair" with device name + 4-char pairing code. Screen stays on.
4. From the companion script on the same Wi-Fi, pair using the code and send `ConnectRequest` → phone flips to **Waiting for payment**.
5. Send `CollectPaymentRequest` → `ErrorMessage(code="not_implemented")`, stays on **Waiting for payment**.
6. Cancel or close the socket → activity finishes, `CARD_READER` logcat shows cleanup + NSD unregister.
7. Toggle the system to dark mode and repeat - the screen renders with the correct dark background and text colors.

### Images/gif

https://github.com/user-attachments/assets/112afe8f-cf46-473a-bddc-9416f1b34ea8


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt
```diff
@@ -238,13 +246,29 @@ class PaymentsHubViewModel @Inject constructor(
...
+                label = UiStringRes(R.string.card_reader_mode_settings_row_label),
+                index = 10,
+                onClick = ::onCardReaderModeClicked,
+            )
+        )
+    }
+
+    private fun onCardReaderModeClicked() {
+        triggerEvent(PaymentsHubEvents.NavigateToCardReaderMode)
+    }
```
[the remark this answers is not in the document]
kidinov: @copilot apply changes based on [this feedback](https://github.com/woocommerce/woocommerce-android/pull/15717#discussion_r3124687326)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+@HiltViewModel
+class CardReaderModeViewModel @Inject constructor(
+    private val session: CardReaderRemoteSession,
+) : ViewModel() {
+
+    private val _stateOverride = MutableLiveData<ViewState?>(null)
+    val stateOverride: LiveData<ViewState?> = _stateOverride
+
+    private val _events = MutableLiveData<MultiLiveEvent.Event>()
+    val events: LiveData<MultiLiveEvent.Event> = _events
```
malinajirka: `_events` is plain `MutableLiveData`, which is sticky — on Activity recreation a new observer re-reads the last value and `finish()` fires again. The `CardReaderModeExit.isHandled = false` reset in `exit()` doesn't guard against this since the observer never checks `isHandled`.

Two options:
- Extend `ScopedViewModel(savedState)` and use `triggerEvent(CardReaderModeExit)` like the rest of `ui/payments/`.
- Or, if you want to keep this portable for the POS move, expose events via a `Channel<…>.receiveAsFlow()` (consume-once, no sticky redelivery, no app-framework dep).

Either fixes the bug and lets the `isHandled` workaround go away.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    init {
+        viewModelScope.launch {
+            session.state.collect { sessionState ->
+                _stateOverride.postValue(mapToViewState(sessionState))
+            }
+        }
+        session.start(viewModelScope)
```
malinajirka, line 39: 💡 There's no error path off `session.state`. If `session.start()` fails internally (NSD registration blocked on the current Wi-Fi, TLS key generation fails on a device without a secure lockscreen, socket bind fails, …), `CardReaderRemoteSession` catches it, logs via `logWrapper.e`, and resets `_state` to `Idle`. Here `Idle` maps to `null`, which the screen renders as the "Starting Card Reader Mode" spinner indefinitely — the user has no way to know anything went wrong, and we don't send a Sentry event.

I think at minimum we need a terminal `Error` state in `CardReaderRemoteSessionState` that the VM can map to a failure screen with a retry action (plus a `logError`). Could also be a follow-up if we want to keep this PR tight, but worth tracking before we ship the flag.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class CardReaderModeViewModel @Inject constructor(
+    private val session: CardReaderRemoteSession,
+) : ViewModel() {
+
+    private val _stateOverride = MutableLiveData<ViewState?>(null)
+    val stateOverride: LiveData<ViewState?> = _stateOverride
```
malinajirka: 🔍 np: Looks like `stateOverride` might be just a leftover from the removed bridge — nothing is being overridden anymore, right? Would rename to `viewState` be perhaps better for clarity. Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModelTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun `given starting view state, when cancel clicked, then exit event is emitted`() = testBlocking {
+        sessionState.value = CardReaderRemoteSessionState.Starting
+
+        advanceUntilIdle()
+
+        (viewModel.stateOverride.getOrAwaitValue() as RemoteTapToPayStarting).onPrimaryActionClicked.invoke()
+
+        assertThat(viewModel.events.getOrAwaitValue()).isEqualTo(CardReaderModeExit)
+    }
+}
```
malinajirka, line 120: 🔍 np: missing a quick test that `onCleared()` calls `session.stop()` — easy to regress into a leaked NSD registration / server socket if someone refactors the lifecycle later. Wdyt?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -1525,6 +1525,10 @@
     <string name="card_reader_mode_waiting_header">Paired with tablet</string>
     <string name="card_reader_mode_waiting_subtitle">Waiting for payment…</string>
     <string name="card_reader_mode_cancel">Cancel</string>
+    <string name="card_reader_mode_settings_row_label">Card Reader Mode</string>
+    <string name="card_reader_mode_device_identifier">%1$s · %2$s</string>
```
malinajirka: ❓ I'm not sure, but feels like we might want to keep as translatable=false, wdyt? I'm not 100% sure since in RTL it might need to be swapped.

malinajirka (approved):
Thanks @kidinov ! I haven't tested this since I wasn't sure what you mean by `From the companion script on the same Wi-Fi`. Either case, since this goes into a feature branch, I think that's fine for now. I've left a couple inline comments.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
kidinov: Switched events to `Channel<CardReaderModeExit>` exposed as a `Flow` via `receiveAsFlow()`, and the activity collects with `repeatOnLifecycle(STARTED)`. Dropped the `isHandled` workaround and the `MultiLiveEvent.Event` parent. In commit 05fcc339a90.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
kidinov: Added `CardReaderRemoteSessionState.Error(message)` and a `RemoteTapToPayError` UI state mapped from the session catch branch. The exception `toString()` (class + message) is shown in the subtitle so it is easy to debug from a screenshot. `LogWrapper.e` was already in place. In commit 463f9c7702b.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModel.kt
kidinov: Renamed to `viewState` (also switched the underlying `LiveData` to `StateFlow` since I was touching it). In commit 05fcc339a90.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeViewModelTest.kt
kidinov: Added without reflection - construct the VM via `ViewModelProvider(ViewModelStore(), factory)` and call `store.clear()` to trigger the real `onCleared` path. In commit 73a13946240.

WooCommerce/src/main/res/values/strings.xml
kidinov: Marked `translatable="false"` since it is a code-formatting string. In commit a263a7874c6.

PR #15718, opened 2026-04-22
[Woo POS] [Remote TTP] Unified BT + NSD discovery stream

kidinov:
Fixes WOOMOB-2740.

### Description

Adds a unified discovery stream for the Woo POS connect dialog that merges the existing Bluetooth reader scan with a new NSD-based scan that finds phones running Card Reader Mode. The NSD branch is gated on `FeatureFlag.REMOTE_TAP_TO_PAY` — with the flag off, the stream is byte-identical to today's BT-only discovery.

- Introduces `CardReaderRemoteDiscovery` (public) in `libs/cardreader` so the main app can browse NSD without leaking Stripe or internal NSD types.
- Adds `WooPosRemoteReaderDiscovery` adapter and `WooPosUnifiedDiscoveryStream` merger in `ui/woopos/cardreader/remote/`.
- Rewires `WooPosCardReaderConnectionController.startDiscovery()` onto the unified stream. Phone rows are present in the event payload but not rendered yet — that lands with WOOMOB-2742 (Task 4.4).
- Phones are keyed by TLS fingerprint (stable, cert-bound) and the map is capped at 32 entries to bound memory against LAN floods. NSD failures are absorbed so a broken listener cannot tear down Bluetooth discovery.

### Test Steps

1. With `REMOTE_TAP_TO_PAY` **off**, open Woo POS → Settings → Card Readers → Connect. Bluetooth discovery behaves exactly as on trunk.
2. With the flag **on**, open the same dialog. No visible difference in the list yet — this PR only wires the stream. Verify no crash, no new permission prompts, existing BT readers still appear and connect normally.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15721, opened 2026-04-23
[WOOMOB-2742] Tablet session, and phone rows in discovery list

kidinov:
### Description

Fixes WOOMOB-2742.

Wires the tablet side of Woo POS Remote Tap-to-Pay behind `REMOTE_TAP_TO_PAY`. Phones from the unified discovery stream render as tappable rows in the Connect Reader dialog with a transport icon + Wi-Fi label + 4-char pairing code; tapping a phone opens a pinned TLS connection and runs the `ConnectRequest` / `ConnectAck` exchange via a new `CardReaderRemoteTabletClient` and a `@Singleton` `WooPosRemoteReaderSession`.

- `WooPosRemoteReaderPaymentFlow` is **not** in this PR — it'll ship with the follow-up that integrates remote collect into `WooPosCardPaymentViewModel`, so the result shape is driven by a real caller.
- BT/Location permission gate still runs for phone-only users — tracked in WOOMOB-2784.
- Stacked on #15718.

### Test Steps

1. Enable `REMOTE_TAP_TO_PAY` + simulated card reader. Open Connect Reader, confirm phone rows appear with icon + Wi-Fi label + 4-char pairing code.
2. Tap a phone row → Connecting → Connected (or retry/cancel on failure).
3. Cancel mid-connect, reopen, connect again — no stuck state.
4. Tap a BT reader — existing flow still works, auto-connect to last-known still fires.
5. Disable the flag — dialog behaves as trunk.
6. Against a real phone in Card Reader Mode, confirm the pairing code matches what the phone shows.

### Images/gif

https://github.com/user-attachments/assets/8ce65088-2e1d-4c67-b85b-d05e8d903d31


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTabletClient.kt
```diff
@@ -0,0 +1,125 @@
...
+        connectionToken: String,
+        locationId: String,
+    ): ConnectOutcome {
+        disconnect()
+        return try {
+            val opened = tlsClient.connect(reader.host, reader.port, reader.fingerprintBase64)
+            connection = opened
+            val requestId = UUID.randomUUID().toString()
+            opened.send(ConnectRequest(requestId, connectionToken, locationId))
+            when (val reply = opened.receive().first { it.requestId == requestId }) {
```
malinajirka: 💡 Have you considered adding a timeout here to ensure we don't end up with an infinite loading?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
```diff
@@ -23,6 +29,26 @@ internal class CardReaderRemoteConnection internal constructor(
...
 
+    private val readerScope = CoroutineScope(SupervisorJob() + ioDispatcher)
+    private val messages = Channel<CardReaderRemoteMessage>(capacity = MESSAGE_BUFFER_CAPACITY)
+
+    private val readerJob: Job = readerScope.launch {
+        var fatalError: Throwable? = null
+        try {
+            while (!socket.isClosed) {
+                val next = runCatching { protocol.read(input) }.getOrElse { err ->
+                    if (err !is EOFException && err !is SocketException && err !is SSLException) {
```
malinajirka: 💡 `SSLException` thrown mid-stream is treated identically to a clean disconnect: `fatalError` stays null and the channel closes "normally". Should we at least somehow log the error?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -271,13 +278,16 @@ class WooPosCardReaderConnectionController(
         discoveryJob?.cancel()
         connectionStatusJob?.cancel()
         softwareUpdateJob?.cancel()
+        remoteConnectionJob?.cancel()
         selectedReader = null
+        scope.launch { remoteReaderSession.disconnect() }
```
malinajirka, line 283: 💡 `cancel()` advances state to `Scanning` immediately while `remoteReaderSession.disconnect()` runs in a fire-and-forget `scope.launch`. Two things bite here:

1. `disconnect()` is `mutex.withLock { ... }` on a `@Singleton`. If `connect()` is currently inside the mutex on a slow TLS handshake (no timeout — see TabletClient comment), this `disconnect()` queues behind it. If the user immediately taps another phone, the new `connect()` queues too, so the second attempt is stuck "Connecting" until the original network call eventually unwinds.
2. If `disconnect()` ever throws, the launch dies silently — no log, no Sentry.

I think `disconnectInternal()` could close `client` synchronously without the mutex (`CardReaderRemoteConnection.close()` is non-blocking + idempotent), which would also unblock the suspended `receive().first { ... }` in any in-flight `connect()` and free the mutex for the retry. Wdyt?

Separately: after a phone connect failure, `onCancelClicked = { cancel() }` here — but `cancel()` doesn't restart `startDiscovery()` (compare the BT path's `handleConnectionFailed` which calls `enterScanningState(); startDiscovery()`). So the user lands in Scanning with no active discovery job and "Searching…" forever.

malinajirka (approved):
Thanks @kidinov! 

I think it'd be nice to add more unit tests, e.g. WooPosRemoteReaderSession and DefaultCardReaderRemoteTabletClient dont' have any AFAICS.

Regarding security of this solution - I'd personally consider verifying that the client (phone) has access to the site (is logged in). That would significantly increase "chances" the tablet is connected to the correct phone with proper authorization. AFAIU, we are sending secrets to the connected device and the only verification are 4 digits during connection - not unprecedented, but worth considering.

I have done just shallow testing as I assume a thorough testing will be done when the feature branch is complete.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTabletClient.kt
kidinov: Fixed in b337694e69e — wrapped the connect block in `withTimeout(DEFAULT_CONNECT_TIMEOUT_MILLIS = 30s)`.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteConnection.kt
kidinov: Fixed in 842b43d3f8b — `SSLException` is now logged at warn level so TLS issues are visible.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
kidinov: Both fixed:

- 9b35343cc8f — `WooPosRemoteReaderSession.disconnect()` now closes the client synchronously outside the mutex, so an in-flight `connect()` is unblocked and the mutex frees up.
- 6a3cec694e2 — `cancel()` now calls `startDiscovery()` after entering Scanning, matching the BT retry pattern, so the user is no longer stuck on "Searching…".

kidinov:
Thanks for the review!

**Tests:** I added basic unit tests for `DefaultCardReaderRemoteTabletClient` and `WooPosRemoteReaderSession` in 3fe83736ec7. Just main happy and failure paths, nothing wide. We can remove the `unit-tests-exemption` label from these classes now.

**Security (phone site auth):** Good point. Today pairing uses the cert fingerprint, the 4-char code (which is just the same fingerprint shown in a readable way), and the operator picking the right phone from the list. In practice it is similar to BT pairing, but Wi-Fi LAN reach is much wider than BT, so site auth would be a real improvement.

Also, the phone already does more of the Stripe work over time. In #15732 the phone creates the `PaymentIntent` instead of the tablet. Adding "phone proves it has a token for site X" fits on top of that.

I filed [WOOMOB-2853](https://linear.app/a8c/issue/WOOMOB-2853/remote-ttp-verify-phone-is-logged-into-the-same-wp-site-during-pairing) to track it. We can decide if we want to land it before the feature branch merges or after.

PR #15723, opened 2026-04-23
[RSM] Add phone layout support for POS Orders screen

samiuelson:
### Description

The POS Orders screen used a fixed dual-pane layout (30% list / 70% details) designed for tablets. On phone-sized screens this was too cramped to be usable.

This PR adapts the orders screen for phones following the same `isWooPosPhoneLayout()` pattern already established by the home screen:

- **Phone**: the orders list is shown full-screen. Tapping an order navigates to a full-screen order detail view (reusing the existing `ORDER_DETAILS_ROUTE` and `isSingleOrderMode` path). Back returns to the list.
- **Tablet**: the existing side-by-side dual-pane layout is unchanged.

Additional fixes included:
- **No selection border on phone** — on tablet, the selected order card gets a 2dp border to indicate which order's details are shown in the right pane. On phone there is no detail pane, so the selection border is suppressed via a `withoutSelection()` helper.
- **Vertical alignment fix in order details** — the detail header row used a hardcoded `56.dp` min height (`WOO_POS_ORDERS_TOOLBAR_HEIGHT`) while the toolbar uses the adaptive `WooPosComponentSize.XSmall.value` (which scales down on phones). This caused the back arrow, title, action buttons, and overflow menu to sit at different vertical positions. Switched the detail row to use the same adaptive size so they align.

No ViewModel or navigation changes were needed — the existing `OpenOrderDetails` navigation event and `ORDER_DETAILS_ROUTE` were already fully wired.

### Test Steps

1. Open POS on a **phone** device or emulator
2. Navigate to Orders — the list should appear full-screen with no detail pane
3. Tap an order — it should navigate to a full-screen detail view with back arrow, title ("Order #NNN"), and action buttons all vertically aligned
4. Verify no order card shows a selection border (black outline) in the list
5. Press back — should return to the orders list
6. Verify search, pull-to-refresh, and pagination work on the phone list
7. Open POS on a **tablet** — verify the dual-pane layout still works as before (list on left, details on right, selected order has border)

### Images/gif

https://github.com/user-attachments/assets/18632798-a33c-4c46-b3af-f49f7462c2ff


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -103,10 +106,15 @@ fun WooPosOrdersScreen(
         listState = listState,
         detailState = detailState,
         isSingleOrderMode = detailViewModel.isSingleOrderMode,
+        isPhoneLayout = isPhoneLayout,
         scrollToTopEvent = listViewModel.scrollToTopEvent,
         onBackClicked = { onNavigationEvent(WooPosNavigationEvent.GoBack) },
         onRefresh = listViewModel::onRefresh,
-        onOrderSelected = listViewModel::onOrderSelected,
+        onOrderSelected = if (isPhoneLayout && !detailViewModel.isSingleOrderMode) {
```
toupper: On tablet, `listViewModel.onOrderSelected` fires `ordersAnalyticsTracker.trackOrdersListRowTapped(...)`. On phone we bypass it entirely and go straight to navigation, so that tracking event is never emitted. Worth invoking the tracker on the phone path too (or routing both layouts through the ViewModel and letting it decide).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -668,6 +700,16 @@ private fun OrdersError(
     )
 }
 
+private fun WooPosOrdersListState.Content.withoutSelection(): WooPosOrdersListState.Content {
```
toupper: This allocates a new Content + a new Loaded + a new list of `OrderItemViewState` on every recomposition while scrolling/refreshing on phone. Preferable to push an `isSelectable: Boolean` (or `showSelection`) into OrdersListPane / the row composable and have the card ignore `isSelected` on phone, no state cloning, and lazy-list keys stay stable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsMapper.kt
```diff
@@ -110,7 +110,9 @@ class WooPosOrderDetailsMapper @Inject constructor(
                 formatPrice(BigDecimal.ZERO)
             },
             paymentMethodTitle = order.paymentMethodTitle.takeIf { it.isNotBlank() },
-            actionsState = WooPosOrdersState.OrderActionsState.Loading
+            actionsState = WooPosOrdersState.OrderActionsState.Loaded(
```
toupper: The name of the function promises `WithoutActions` but the body now eagerly computes them. Either rename (e.g. mapOrderDetailsFast / mapOrderDetailsWithoutRefundData) or keep the loading semantics,  current state is misleading for the next reader.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsMapper.kt
```diff
@@ -110,7 +110,9 @@ class WooPosOrderDetailsMapper @Inject constructor(
                 formatPrice(BigDecimal.ZERO)
             },
             paymentMethodTitle = order.paymentMethodTitle.takeIf { it.isNotBlank() },
-            actionsState = WooPosOrdersState.OrderActionsState.Loading
```
toupper, line 113: After this PR both mappers always emit Loaded, and WooPosOrderDetailsViewModel no longer transitions actions from Loading → Loaded. That makes the is OrderActionsState.Loading branch in WooPosOrderDetails.kt (the shimmer placeholder) dead code, and the Loading subtype itself unused. Either delete OrderActionsState.Loading and the shimmer branch, or keep the original deferred-loading flow — having the state but no producer is confusing.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
```diff
@@ -44,26 +44,38 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.toAdaptiv
...
-fun WooPosOrdersLoadingScreen(modifier: Modifier = Modifier) {
-    Row(
-        modifier = modifier.fillMaxSize()
-    ) {
+fun WooPosOrdersLoadingScreen(
+    isPhoneLayout: Boolean = false,
+    modifier: Modifier = Modifier,
+) {
+    if (isPhoneLayout) {
         WooPosOrdersListLoadingPane(
```
toupper, line 52: Minor: I see we are duplicating the `WooPosOrdersListLoadingPane`. Can we extract the list-loading modifier (or the pane itself) into a single source of truth? That would keep the two paths from drifting on top-padding/background changes. Not blocking, just a smell now that there are two layouts to keep in sync.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
toupper: In single-order mode there's no list rendered at all, so onOrderSelected is unreachable. The extra clause is harmless but confusing — consider dropping it (or adding a comment explaining it's belt-and-braces).

samiuelson: Extracted `trackOrderTapped()` from `onOrderSelected()` so the phone path fires the analytics event before navigating. Also removed the unreachable `isSingleOrderMode` guard (see comment below).

See 359c0953535

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: Replaced `withoutSelection()` with an `isSelectable` flag threaded through `OrdersListPane` → `OrdersList` → `LoadedOrdersList`. The card now uses `isSelected = isSelectable && item.isSelected` — no allocations during scroll.

See 0dc26c14b96

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsMapper.kt
samiuelson: Renamed to `mapOrderDetailsWithoutRefunds` to reflect its actual purpose (everything except refund-dependent data).

See fd4a1a1f0dc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsMapper.kt
samiuelson: Collapsed `OrderActionsState` from a sealed class into a plain data class — removed the `Loading` subtype and the shimmer branch in `OrderActions`.

See fd4a1a1f0dc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
samiuelson: `WooPosOrdersListLoadingPane` is already a single shared composable used in both places. The phone and tablet branches of `WooPosOrdersLoadingScreen` only differ in the outer container (phone = full-screen, tablet = side-by-side with details pane), which is intentional for the different layouts.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: Removed the guard — in single-order mode the list is never rendered, so `onOrderSelected` is unreachable anyway. The condition is now just `if (isPhoneLayout)`.

See 359c0953535

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersLoadingScreen.kt
```diff
@@ -44,26 +44,38 @@
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.toAdaptiveIconSize
 
 @Composable
-fun WooPosOrdersLoadingScreen(modifier: Modifier = Modifier) {
-    Row(
-        modifier = modifier.fillMaxSize()
-    ) {
+fun WooPosOrdersLoadingScreen(
+    isPhoneLayout: Boolean = false,
+    modifier: Modifier = Modifier,
```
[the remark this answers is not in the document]
samiuelson: Fixed — `modifier` is now the first optional parameter.

See 490a261c949

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
```diff
@@ -136,7 +135,7 @@ class WooPosOrderDetailsViewModelTest {
 
         // THEN
         val loaded = viewModel.state.value as WooPosOrderDetailsState.Loaded
-        assertThat(loaded.details.actionsState).isInstanceOf(OrderActionsState.Loaded::class.java)
+        assertThat(loaded.details.actionsState.actions).isNotNull()
```
toupper: After the sealed-class → data-class collapse, actions is a List<OrderAction> — non-null by Kotlin's type system, so this assertion can never fail. It compiled cleanly because the previous assertion (isInstanceOf(OrderActionsState.Loaded::class.java)) lost its meaning. Replace with something content-bearing, e.g. isNotEmpty() or containsExactlyInAnyOrder(IssueRefund(1L), EmailReceipt(1L)).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/list/WooPosOrdersListViewModel.kt
```diff
@@ -88,6 +88,18 @@ class WooPosOrdersListViewModel @Inject constructor(
...
+        trackOrderTapped(orderId)
+
+        val updatedItems = loadedItems.items.map { it.copy(isSelected = it.id == orderId) }
+        _state.value = current.copy(
+            items = WooPosOrdersListState.Content.Items.Loaded(updatedItems)
+        )
+        coordinator.selectOrder(orderId)
+    }
+
+    fun trackOrderTapped(orderId: Long) {
```
toupper: Now a public ViewModel method that fires analytics without doing the rest of onOrderSelected (state update + coordinator). Future maintainer may call this thinking it's a generic "track tap" helper. Consider:

internal visibility, or
Renaming to make the pairing explicit (e.g. trackOrderTappedOnPhone), or
Wrapping the phone path in a dedicated VM method like onPhoneOrderSelected(orderId) that fires the analytics and exposes a navigation event — keeps the composable from doing two-step business logic inline.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -103,10 +106,18 @@ fun WooPosOrdersScreen(
         listState = listState,
         detailState = detailState,
         isSingleOrderMode = detailViewModel.isSingleOrderMode,
+        isPhoneLayout = isPhoneLayout,
         scrollToTopEvent = listViewModel.scrollToTopEvent,
         onBackClicked = { onNavigationEvent(WooPosNavigationEvent.GoBack) },
         onRefresh = listViewModel::onRefresh,
-        onOrderSelected = listViewModel::onOrderSelected,
+        onOrderSelected = if (isPhoneLayout) {
```
toupper, line 114: Same observation as before — the tablet path is a clean method ref into the VM, the phone path mixes two concerns at the call site. Folding these into a single onPhoneOrderSelected (or having onOrderSelected itself check a mode and emit a SharedFlow event for nav) would make the two layouts symmetric and keep the analytics rule co-located with the selection logic.

toupper (approved):
Thanks for the fixes @samiuelson! I added a few more comments that can be addressed before merging.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
samiuelson: Replaced with `isNotEmpty()` — verifies actions were actually computed for the completed order.

See 569f637810f

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/list/WooPosOrdersListViewModel.kt
samiuelson: Introduced `WooPosScreenType` (SinglePane/DualPane) param on `onOrderSelected` so both paths call the same method. `trackOrderTapped` is now `private`. On SinglePane the duplicate-tap guard, selection update, and coordinator notify are skipped since the phone navigates to a separate detail screen.

See 8ea7ae5aca2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: Both paths now call `listViewModel.onOrderSelected(orderId, screenType)` — phone adds the navigation call after, which is inherent since tablet doesn't navigate.

See 8ea7ae5aca2

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15726, opened 2026-04-23
Bump push notifications min WC version to 10.8.0

irfano:
### Description

Updates `PUSH_NOTIFICATIONS_MIN_WC_VERSION` from `10.6.0` to `10.8.0` to align with the final Woo Core release that ships the self-driven push notifications endpoint.

Context: the `POST /wp-json/wc-push-notifications/push-tokens` endpoint (core send path) is merged in Woo Core and scheduled for the 10.8.0 release. The previous `10.6.0` bound was a placeholder from an earlier milestone and would let incompatible stores through the `CheckWooPluginPushNotificationsSupport` gate.

### Test Steps

1. Install the app on a site running WooCommerce **< 10.8.0**.
2. Enable the `WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M2` feature flag override (Settings → Beta Features → Override Feature Flags).
3. On the dashboard, tap the "Never miss a new order" card.
4. **Verify:** the "Check plugin compatibility" step fails with the outdated-plugin error and the action button shows "Update plugin".
5. Upgrade the site to Woo **10.8.0** (or a `10.8.0-dev` build) and repeat steps 3–4.
6. **Verify:** the "Check plugin compatibility" step succeeds and the flow continues to the next step.

### Images/gif

N/A

- [ ] I have considered if this change warrants release notes and have added them to \`RELEASE-NOTES.txt\` if necessary. Use the "[Internal]" label for non-user-facing changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/CheckWooPluginPushNotificationsSupport.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/CheckWooPluginPushNotificationsSupport.kt
@@ -10,7 +10,7 @@ class CheckWooPluginPushNotificationsSupport @Inject constructor(
     private val getWooCorePluginCachedVersion: GetWooCorePluginCachedVersion
 ) {
     companion object {
-        const val PUSH_NOTIFICATIONS_MIN_WC_VERSION = "10.6.0"
+        const val PUSH_NOTIFICATIONS_MIN_WC_VERSION = "10.8.0"
     }
 
     suspend operator fun invoke(forceRefresh: Boolean): Result {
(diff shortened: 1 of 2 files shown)


hichamboushaba:
Thanks @irfano for handling this, but did we confirm the correct version with Hannah first? Currently all work of PNs is merged but it's behind a feature flag, so we need to confirm the correct version where the feature flag will be enabled.
Also, Hannah is working on a new status endpoint, this might end up delaying the merge until after 10.8.

WDYT about the approach I suggested to the iOS team here p1776684243082869/1776683122.262339-slack-C03L1NF1EA3? Essentially, it involves waiting and creating a final PR to activate the feature once everything is prepared. This PR should update the version and turn on our local feature flag.

irfano:
Thanks for the extra information, @hichamboushaba! Your suggestion sounds good.

I updated this to fix the failing steps in the test plan, because the plugin compatibility step was failing due to the incorrect hardcoded version. Let’s merge this for now, then we can go with your plan once we know the correct version.

hichamboushaba (approved):

PR #15727, opened 2026-04-24
[RSM] Convert POS Issue Refund from dialog to full-screen flow | Part 1

samiuelson:
### Description

Converts the POS Issue Refund flow from a dialog overlay to a full-screen Compose Navigation route. This is a follow-up to #15723 (phone layout for orders).

**What changed:**

The Issue Refund flow was previously rendered inside a `WooPosDialogWrapper` overlay on the orders screen. This PR converts it to a dedicated navigation route (`home/orders/issue_refund/{orderId}`) that slides in as a full-screen composable — consistent with other POS flows like payment, email receipt, and refund reason.

Key architectural changes:
- **New route**: `WooPosIssueRefundNavigation.kt` defines the route and reads `REFUND_REASON_RESULT_KEY` directly from its own `savedStateHandle`, eliminating the prop-drilling chain through the orders screen.
- **Screen layout**: Toolbar at top with a dynamic title that adapts to the current refund step (e.g., "Select items to refund" → "Review Refund" → "Refund $X.XX?"). Action buttons are pinned to the bottom of the screen, matching the pattern used in checkout and payment flows.
- **Navigation event**: `OpenIssueRefund(orderId)` is intercepted at the composable level in both the orders screen and bookings screen, triggering navigation instead of setting dialog state.
- **Result handling**: When the refund screen is dismissed, it passes `ISSUE_REFUND_DISMISSED_KEY` back via `GoBackWithResult`, and the calling screen triggers an order refresh.
- **Cleanup**: Removed `DialogState.IssueRefund` from `WooPosOrderDetailsState`, removed dialog-related methods from the detail ViewModel, and removed the `WooPosDialogWrapper` / `rememberRetained` wiring from both the orders and bookings screens.

The Refund Details dialog (read-only historical view) remains as a dialog for now.

### Test Steps

1. Open POS and create an order with at least one item
2. Navigate to Orders, select a completed order
3. Tap "Issue Refund" — verify it opens as a full-screen view (not a dialog)
4. Verify the toolbar title says "Select items to refund"
5. Select items and tap Continue — toolbar should update to "Review Refund"
6. Tap Continue — toolbar should update to "Refund $X.XX?"
7. Tap "Yes, Proceed" — verify the refund processes and shows success screen with "Refund Complete" title
8. Tap Done — verify you return to the order details with the refund reflected
9. Verify the back button works at each step

### Images/gif

https://github.com/user-attachments/assets/ca3584b8-2cd0-4e85-85df-e5bb84f1d94f


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
```diff
@@ -200,23 +205,12 @@ private fun WooPosBookingsScreen(
         if (state is WooPosBookingsState.Content) {
             val dialogState = state.dialogState
 
-            val retainedIssueRefund: WooPosBookingsState.Content.DialogState.IssueRefund? = rememberRetained(
-                when (dialogState) {
-                    is WooPosBookingsState.Content.DialogState.IssueRefund -> dialogState
-                    is WooPosBookingsState.Content.DialogState.CancelBooking -> null
-                    WooPosBookingsState.Content.DialogState.Hidden -> null
+            if (dialogState is WooPosBookingsState.Content.DialogState.IssueRefund) {
```
toupper: When the refund screen pops back, the bookings entry recomposes. Two LaunchedEffects are queued: the dismissal one (which calls viewModel.onIssueRefundDialogDismissed() to flip dialog state to Hidden) and this navigation one. They both launch from a composition where dialogState is still IssueRefund, and navigateOnce won't suppress this because the current route is the bookings route. End result: refund screen re-opens immediately on pop.

The orders screen sidesteps this by intercepting at the click callback (synchronous, fire-once). Suggest doing the same for bookings — emit OpenIssueRefund from the click handler / a one-shot SharedFlow, not from observing dialogState. Worth a manual repro: "Issue Refund → close → confirm you stay on bookings".

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -121,12 +126,18 @@ fun WooPosOrdersScreen(
         onSearchErrorRetry = listViewModel::onSearchErrorRetry,
         onOrdersEmptyActionClicked = listViewModel::onOrdersEmptyActionClicked,
         onOrdersLoadingErrorRetryButtonClicked = listViewModel::onOrdersLoadingErrorRetryButtonClicked,
-        onUIEvent = detailViewModel::onUIEvent,
+        onUIEvent = { event ->
```
toupper: The same UI event has two handlers — one in the orders composable, one in the bookings VM (via DialogState.IssueRefund). And the detail VM keeps a IssueRefund -> { /* handled at composable level */ } no-op branch. If anything new ever wires up to OrderActionClicked (e.g., a tracking call inside handleActionClicked), it'll silently miss the refund path. Pick one pattern and use it in both screens — preferably the click-side interception, since it avoids the previous issue

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModel.kt
```diff
@@ -142,7 +142,9 @@ class WooPosOrderDetailsViewModel @Inject constructor(
     private fun handleActionClicked(action: OrderAction) {
         when (action) {
             is OrderAction.EmailReceipt -> onEmailReceiptButtonClicked(action.orderId)
-            is OrderAction.IssueRefund -> onIssueRefundButtonClicked(action.orderId)
+            is OrderAction.IssueRefund -> {
```
toupper: If the VM no longer cares about IssueRefund clicks, removing the branch (and letting it fall through when exhaustiveness) is cleaner. Right now the comment is the only thing telling a reader the click is intercepted somewhere else.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFullScreenInputLayout.kt
```diff
@@ -35,7 +36,9 @@ fun WooPosFullScreenInputLayout(
     BackHandler { onBackClicked() }
 
     Column(
-        modifier = modifier.fillMaxSize()
+        modifier = modifier
+            .fillMaxSize()
+            .navigationBarsPadding()
```
toupper: This file isn't part of the refund flow rewrite. Adding `.navigationBarsPadding()` here changes every other screen that uses the layout (refund reason, etc.) I think being in its own PR would be cleaner, perhaps even outside RSM.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundDialog.kt
```diff
@@ -99,69 +95,302 @@ fun WooPosIssueRefundDialog(
...
+                .fillMaxSize()
+                .statusBarsPadding()
+                .navigationBarsPadding()
+                .padding(
+                    start = WooPosSpacing.Medium.value,
+                    end = WooPosSpacing.Medium.value,
+                    bottom = WooPosSpacing.XLarge.value,
+                ),
+        ) {
+            Spacer(modifier = Modifier.height(WooPosComponentSize.XSmall.value))
```
toupper: The inner Column makes room for the floating header by hard-coding a Spacer(height = WooPosComponentSize.XSmall.value) that mirrors the header's height. If either value changes, content slips under the header. Suggest using a Column { Header(); Content() } (no overlay), or extracting the offset value as a single source of truth — the WooPosFullScreenInputLayout already does the former, would the refund screen fit it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundDialog.kt
```diff
@@ -71,23 +72,18 @@ import com.woocommerce.android.ui.woopos.root.navigation.WooPosNavigationEvent
...
     orderId: Long,
-    isVisible: Boolean,
-    onDismissRequest: () -> Unit,
     onNavigationEvent: (WooPosNavigationEvent) -> Unit,
     refundReasonUpdate: String? = null,
-    disablePartialRefund: Boolean = false
 ) {
     val viewModel: WooPosRefundViewModel =
-        hiltViewModel<WooPosRefundViewModel, WooPosRefundViewModel.Factory>(key = "refund_$orderId") { factory ->
+        hiltViewModel<WooPosRefundViewModel, WooPosRefundViewModel.Factory> { factory ->
```
toupper, line 99: Old code passed key = "refund_$orderId" to ensure a fresh VM per order. The new dedicated route already scopes the VM per backstack entry (different orderId arg ⇒ different entry), so this is probably fine, but worth confirming behavior when re-navigating to the same orderId after dismissing — does factory.create(orderId) see the previous VM's state, or always start fresh?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
```diff
@@ -225,7 +225,7 @@ class WooPosOrderDetailsViewModelTest {
     }
 
     @Test
-    fun `given loaded order, when issue refund action clicked, then dialog state is IssueRefund`() = runTest {
+    fun `given loaded order, when issue refund action clicked, then dialog state is unchanged`() = runTest {
```
toupper: The test now feeds an event into the VM and verifies nothing happens. Either delete it (covered implicitly by exhaustive when) or replace it with one that exercises the screen-side navigation interception. The current name + body don't add coverage.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
```diff
@@ -303,21 +299,19 @@ class WooPosOrderDetailsViewModelTest {
...
         // GIVEN
         viewModel = createViewModel()
         advanceUntilIdle()
         coordinator.selectOrder(1L)
         advanceUntilIdle()
-        viewModel.onUIEvent(WooPosOrdersUIEvent.OrderActionClicked(OrderAction.IssueRefund(orderId = 1L)))
 
         // WHEN
-        viewModel.onIssueRefundDialogDismissed()
+        viewModel.onBackFromIssueRefund()
```
toupper, line 320: Orders VM renamed to onBackFromIssueRefund(). Bookings VM kept onIssueRefundDialogDismissed(). Both are called from the same ISSUE_REFUND_DISMISSED_KEY handler. Aligning names would make the symmetry obvious — onBackFromIssueRefund() reads better now that there's no dialog.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundReasonNavigation.kt
```diff
@@ -44,16 +44,10 @@ fun NavGraphBuilder.refundReasonScreen(
...
-            slideInHorizontally(
-                initialOffsetX = { fullWidth -> fullWidth },
-            )
-        },
-        exitTransition = {
-            slideOutHorizontally(
-                targetOffsetX = { fullWidth -> fullWidth },
-            )
-        },
+        enterTransition = { fadeIn() },
```
toupper: Switched from slide-horizontal to fade — likely deliberate to match the new refund-flow style — but the PR description doesn't call this out. Worth a line in the description so reviewers don't flag it as drift, and worth confirming the refund-reason screen still feels consistent with surrounding screens (booking note, email receipt, etc.) that may still use slide.

toupper:
Thanks for the PR @samiuelson! Please notice the size of this PR is quite big, it would be better to break it in smaller ones the next time. Thanks a lot!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/bookings/WooPosBookingsScreen.kt
samiuelson: Fixed in 2d6ed6f. Bookings now emits `OpenIssueRefund` directly from `handleBookingAction` via `_navigationEvent`, matching the orders screen click-side pattern. No more `LaunchedEffect` observing `dialogState`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: Unified in ac18e59. Both orders and bookings now use the same pattern: the detail VM emits `OpenIssueRefund` via a `SharedFlow`, collected at the composable level. The composable-side interception is gone.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModel.kt
samiuelson: Initially removed in 321df83. Then ac18e59 routed `IssueRefund` back through the detail VM's `handleActionClicked` → `_navigationEvent.tryEmit(OpenIssueRefund)`, so it's no longer a no-op — it's the single source of navigation for both screens.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosFullScreenInputLayout.kt
samiuelson: Reverted from the shared layout in 2afdd30. Re-applied locally to the refund reason screen only in f9f4c3d.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundDialog.kt
samiuelson: Fixed in 940bf2e. Replaced the `Box` overlay with a sequential `Column { Header(); Content() }` layout.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundDialog.kt
samiuelson: Confirmed manually. Each navigation to the route creates a new backstack entry, so `hiltViewModel` returns a fresh VM instance even for the same `orderId`. The old `key = "refund_$orderId"` was needed because `rememberRetained` shared a single composition scope — that's no longer the case with dedicated route navigation.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
samiuelson: Removed in 296b635.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/WooPosOrderDetailsViewModelTest.kt
samiuelson: Renamed in 492da28. Both VMs now use `onBackFromIssueRefund()`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundReasonNavigation.kt
samiuelson: Reverted to slide-horizontal in d513d65. Consistent with other POS flows now.

samiuelson:
Thanks for review, @toupper! I added improvements, let me know what you think. To avoid growing this PR further, I raised 2 more PRs: 
* https://github.com/woocommerce/woocommerce-android/pull/15794
* https://github.com/woocommerce/woocommerce-android/pull/15795
that contain additional code style-related improvements.

toupper (approved):
All review comments addressed cleanly. Both screens now use the click-side SharedFlow pattern via the detail VM, which removes the LaunchedEffect-recompose bug and unifies the two paths. The follow-ups (`emitNav` helper, non-Content IssueRefund guard, extra test coverage for the SharedFlow emission and the not-Loaded back path, sequential header layout, slide transitions restored) are nice cleanups on top. LGTM.

PR #15731, opened 2026-04-24
[RSM] Adapt POS Settings to phone

samiuelson:
### Description

Adds phone layout for POS Settings.

### Test Steps

1. Open POS settings on a **phone-sized device**
2. Verify "Get help and support" text at the bottom is fully visible (not clipped)
3. Verify category items do **not** show a black selection border
4. Tap a category — verify it navigates to the detail pane with a fade transition and a back button
5. Tap back — verify it returns to the categories list
6. Open POS settings on a **tablet-sized device**
7. Verify category items **do** show a black selection border when selected
8. Verify the two-column layout works as before

### Images/gif

https://github.com/user-attachments/assets/dddeb328-caad-4729-9847-894a8895823e


toupper:
`WooPosOrdersScreen.kt `and `WooPosOrderDetails.kt` in this PR are the same orders phone-layout work that's already open in [#15723](https://github.com/woocommerce/woocommerce-android/pull/15723) — same isPhoneLayout plumbing, same withoutSelection() helper, same WOO_POS_ORDERS_TOOLBAR_HEIGHT → WooPosComponentSize.XSmall.value swap. Both PRs target trunk. Whichever lands second will conflict.

This PR's version adds one improvement worth porting back to [#15723](https://github.com/woocommerce/woocommerce-android/issues/15723): `val deselectedState = remember(listState) { listState.withoutSelection() } `— caches the no-selection copy across static recompositions (addresses the per-recomposition allocation comment from [#15723](https://github.com/woocommerce/woocommerce-android/issues/15723)'s review). Suggest either: (a) drop the orders changes from this PR and rebase on top of [#15723](https://github.com/woocommerce/woocommerce-android/issues/15723), or (b) close [#15723](https://github.com/woocommerce/woocommerce-android/issues/15723) and keep this. Don't merge both.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/WooPosSettingsScreen.kt
```diff
@@ -50,24 +60,116 @@ fun WooPosSettingsScreen(onNavigationEvent: (WooPosNavigationEvent) -> Unit) {
...
+    onCategorySelected: (WooPosSettingsCategory) -> Unit,
+    onNavigate: (WooPosSettingsDetailDestination) -> Unit,
+    onBack: () -> Unit,
+    onShowProductInfoDialog: () -> Unit,
+    onShowScanningSetupDialog: () -> Unit,
+    onRetrySync: () -> Unit,
+    onDismissDialog: () -> Unit,
+    onNavigationEvent: (WooPosNavigationEvent) -> Unit
+) {
+    var isShowingDetail by remember { mutableStateOf(false) }
```
toupper: The phone navigation state is local Compose state, not persisted. If the user is mid-detail and the process is killed (or even just a config change that nukes composition), they jump back to the categories list while the VM still holds a selectedCategory. Two suggestions:

- Use rememberSaveable to survive process death.
- Better: drive this from VM state (e.g. an isDetailVisible field, or derive from selectedCategory != null) so phone and tablet share the same source of truth. Right now phone uses local state, tablet uses VM nav.
Let me know what you think

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/WooPosSettingsScreen.kt
```diff
@@ -50,24 +60,116 @@ fun WooPosSettingsScreen(onNavigationEvent: (WooPosNavigationEvent) -> Unit) {
...
+                    },
+                    modifier = Modifier.fillMaxSize()
+                )
+            }
+        } else {
+            WooPosSettingsDetailPaneScreen(
+                state = state,
+                onNavigate = onNavigate,
+                onBack = {
+                    if (state.canGoBack) {
```
toupper: Phone: local isShowingDetail flips visibility.
Tablet: state.canGoBack + containerViewModel.navigateBack().

The two paths can disagree. Example: user on phone selects a category (VM has selection, local flag goes true), backs out (flag goes false, but VM selectedCategory is unchanged). Rotate / config change → tablet UI now loads with that selection and the right pane populated even though the user "left" it. Easier to maintain if both layouts read from the same VM-driven isShowingDetail/selectedCategoryOrNull and the screen just renders one or two panes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailPaneScreen.kt
```diff
@@ -37,11 +37,13 @@ fun WooPosSettingsDetailPaneScreen(
     onShowProductInfoDialog: () -> Unit,
     onShowScanningSetupDialog: () -> Unit,
     onNavigationEvent: (WooPosNavigationEvent) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    alwaysShowBackButton: Boolean = false,
```
toupper: This flag exists only because phone needs to show the back arrow at the top of the detail pane even when state.canGoBack == false. If the VM exposed phone-aware nav state as suggested above, state.canGoBack would be true on phone whenever a category is selected, and this flag goes away.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/categories/WooPosSettingsCategoriesPaneScreen.kt
```diff
@@ -84,6 +89,8 @@ internal fun WooPosSettingsCategoriesPaneScreenContent(
                 onClick = { onCategorySelected(item) }
             )
         }
+
+        Spacer(modifier = Modifier.size(WooPosSpacing.Medium.value))
```
toupper: The PR description says this was added because "Get help and support" was clipped on phone. But the spacer is now applied unconditionally — tablet gets it too. If it's a phone-only fix, gate it behind isPhoneLayout (or apply via the screen's outer padding); otherwise call out in the description that tablet's bottom spacing intentionally changed too, and verify visually.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/categories/WooPosSettingsCategoriesPaneScreen.kt
```diff
@@ -38,13 +40,15 @@ fun WooPosSettingsCategoriesPaneScreen(
     viewModel: WooPosSettingsCategoriesViewModel = hiltViewModel()
 ) {
     val state by viewModel.state.collectAsState()
+    val isPhoneLayout = LocalContext.current.isWooPosPhoneLayout()
```
toupper: isWooPosPhoneLayout() reads resources.displayMetrics each call. Cheap, but happens on every recomposition. Wrap in remember(LocalConfiguration.current) { ... } so it only re-runs on config changes. Same comment applies to WooPosOrdersScreen and WooPosSettingsScreen if you want consistency.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/WooPosSettingsScreen.kt
samiuelson: Went with the second option and hoisted the pane state into the VM — added `isDetailVisible: Boolean = false` to `WooPosSettingsState`, set by `onCategorySelectedFromPhoneList(...)` and reset by `navigateBack()` when at the root destination. Phone and tablet now share a single source of truth and the pane survives configuration changes. (5410c67)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/WooPosSettingsScreen.kt
samiuelson: Same hoist resolves this. Tablet never sets `isDetailVisible` so its `canGoBack` semantics are unchanged; phone uses it via `canGoBack = currentDestination.parentDestination != null || isDetailVisible`. The "rotate after backing out and the tablet UI shows a stale selection" scenario doesn't reproduce — `navigateBack()` clears `isDetailVisible` when leaving the detail pane on phone. (5410c67)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailPaneScreen.kt
samiuelson: Removed. `state.canGoBack` is now phone-aware, so the parameter isn't needed. (5410c67)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/categories/WooPosSettingsCategoriesPaneScreen.kt
samiuelson: Good catch. Gated behind a new `showBottomSpacer: Boolean = false` parameter on `WooPosSettingsCategoriesPaneScreenContent`, set to `true` only from the phone wrapper. Tablet bottom spacing is unchanged. (5410c67)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/categories/WooPosSettingsCategoriesPaneScreen.kt
samiuelson: Memoized at both settings sites with `remember(context, configuration) { context.isWooPosPhoneLayout() }` (5410c67). The `WooPosOrdersScreen` site is no longer relevant — see the scope-overlap reply.

samiuelson:
#15723 has since landed on trunk. I merged trunk into this branch (081d9107) and took the trunk version of `WooPosOrdersScreen.kt` and `WooPosOrderDetails.kt`, so the orders work is no longer in this PR's diff. #15731 is now scoped to settings only.

samiuelson:
Thanks for review, @toupper! I added improvements, let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/details/WooPosSettingsDetailPaneScreen.kt
toupper: Just to flag for the record (not a blocker): the `alwaysShowBackButton` param did get removed in 5410c67, but then got reintroduced as `showBackOnRoot` (plus `registerBackHandler`) in 93de191 / 870aa11. So the original concern still technically stands — `state.canGoBack` isn't phone-aware, and the detail pane needs the call-site flag to render the back arrow at root on phone. Naming is clearer now though, so I think it's fine to ship as-is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/settings/categories/WooPosSettingsCategoriesPaneScreen.kt
toupper: Same kind of flag — the `remember(context, configuration)` you added in 5410c67 was later dropped in 66adf5af ("drop redundant remember"), so the call site at `WooPosSettingsScreen.kt:59` is back to `LocalContext.current.isWooPosPhoneLayout()` recomputing per recomposition. In practice it's a primitive read on a stable Context so the cost is negligible — happy to leave as-is, just noting the reply text doesn't match the final code.

toupper (approved):
Hoist into the VM looks good — phone and tablet now share `selectedCategory` / `currentDestination` / `showingDetail`, and the test coverage in WooPosSettingsViewModelTest for the new flow is solid. Bottom-spacer scoping via the call-site modifier is cleaner than a parameter, agreed.

Two small reply-vs-code mismatches I noted inline (not blockers, just to keep the threads accurate): `showBackOnRoot` was reintroduced after the initial fix, and the `remember` was later dropped as redundant. Happy with where the PR landed otherwise.

PR #15733, opened 2026-04-26
[WOOMOB-2842] Step 1: QR login — foundation: payload type, parser, and analytics events

malinajirka:
Fixes WOOMOB-2842

### Description

Foundation layer for the QR login feature: a typed `QrLoginPayload` (Application-Password and OAuth ticket variants), the shared `QrLoginPayloadParser` for the URL scheme that the desktop QR encodes, and the `LOGIN_QR_*` analytics events / `UnifiedLoginTracker` steps that the rest of the flow will emit against. No UI yet — purely model + parsing + telemetry plumbing the later steps depend on.

This is **Step 1 of 10** in a stacked PR chain. The `status: do not merge` label is applied because later steps depend on this one.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events ← **this PR**

### Test Steps

### Images/gif

N/A — no UI in this slice.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
```diff
@@ -118,6 +118,11 @@ enum class AnalyticsEvent(override val siteless: Boolean = false) : IAnalyticsEv
     LOGIN_USE_SECURITY_KEY_CLICKED(siteless = true),
     LOGIN_SECURITY_KEY_FAILURE(siteless = true),
     LOGIN_SECURITY_KEY_SUCCESS(siteless = true),
+    LOGIN_QR_PROLOGUE_SHOWN(siteless = true),
+    LOGIN_QR_PROLOGUE_SCAN_TAPPED(siteless = true),
+    LOGIN_QR_PROLOGUE_FALLBACK_TAPPED(siteless = true),
+    LOGIN_QR_SCAN_FAILED(siteless = true),
+    LOGIN_QR_SUCCESS(siteless = true),
```
JorgeMucientes, line 125: We need to make sure this events are the same when added to iOS. Perhaps a good case to try out EventHorizon: pe5sF9-5fP-p2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayload.kt
```diff
@@ -0,0 +1,23 @@
...
+ * ```
+ * woocommerce://qr-login?token=<64-byte hex>&siteUrl=<URL-encoded site URL>
+ * ```
+ *
+ * The [Ticket.token] is a single-use 5-minute bearer ticket — not a credential. The app exchanges
+ * it for an Application Password by POSTing to
+ * `{siteUrl}/wp-json/wc-admin/mobile-app/qr-login-exchange`.
+ */
+sealed interface QrLoginPayload {
+    data class Ticket(
```
JorgeMucientes, line 17: Hmm, I'm curious behind the reasoning for this "Ticket" naming. Is this a convention or something?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayloadParser.kt
```diff
@@ -0,0 +1,75 @@
...
+ * `siteUrl` returns [QrLoginPayload.Invalid]. The parser does not validate the token format
+ * beyond non-blank — that's the server's job during exchange. `siteUrl` is parsed via OkHttp's
+ * [okhttp3.HttpUrl] and rejected if it carries userinfo, query, or fragment components — those
+ * have no role in a Woo site root and are classic spoofing surfaces in the confirmation prompt.
+ */
+class QrLoginPayloadParser @Inject constructor() {
+
+    fun parse(raw: String?): QrLoginPayload {
+        val uri = parseDeepLink(raw) ?: return QrLoginPayload.Invalid
+        val token = uri.queryParam(PARAM_TOKEN)?.takeIf { it.isNotBlank() } ?: return QrLoginPayload.Invalid
```
JorgeMucientes: Could we validate the token shape here instead of accepting any non-blank value?

Right now something like `woocommerce://qr-login?token=x&siteUrl=https%3A%2F%2Fstore.example` is parsed as `QrLoginPayload.Ticket`, so a malformed Woo-looking QR can advance into the confirmation/exchange flow instead of being treated as `Invalid`.

The backend currently generates the token with `wp_generate_password(64, false)`, so the expected shape should be 64 alphanumeric chars, not hex. Related: the KDoc says `<64-byte hex>`, but that seems inaccurate for the current backend contract.

Maybe something like:

```kotlin
private val TOKEN_REGEX = Regex("^[A-Za-z0-9]{64}$")

private fun isValidToken(token: String) = TOKEN_REGEX.matches(token)
```
and then
```kotlin
val token = uri.queryParam(PARAM_TOKEN)?.takeIf(::isValidToken) ?: return QrLoginPayload.Invalid
```

JorgeMucientes (approved):
Overall LGTM, pre-approving but left a couple minor nits

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
malinajirka: Tbh with AI, I'd rely on AI keeping this in sync instead of introducing a new process that will likely slow down development. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayload.kt
malinajirka: I couldn't come up with a good name for it - I wanted to convey the idea that it's not just a token but at the same time that it's a single use => ticket felt like a fitting name. Having said that, I'm up for renaming it if you prefer - we can just create a ticket and do it in the last PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayloadParser.kt
malinajirka: Good catch on both — KDoc was wrong (`wp_generate_password(64, false)` returns alphanumerics, not hex) and accepting any non-blank token does let malformed Woo-looking QRs slip through to the confirmation prompt.

Went with `^[A-Za-z0-9]{64,512}$` rather than a fixed 64 — keeps the sanity gate but leaves headroom if the server lengthens the token without forcing a client release in lockstep. Server stays the authority on whether the token is actually valid; this is just a shape check so obviously-malformed payloads get rejected at parse time. Fixed the KDoc and added tests for the short / too-long / non-alphanumeric cases.

PR #15734, opened 2026-04-26
[WOOMOB-2843] Step 2: QR login — token exchange REST client

malinajirka:
Fixes WOOMOB-2843

### Description

Adds `QrLoginRestClient`, a thin networking layer that exchanges a one-time ticket (parsed from the QR payload in step 1) for an Application Password against the merchant's site. Implements every error branch the UI later distinguishes (network, server, rate-limit, token-rejected, endpoint-missing, invalid-payload) so the ViewModel can route them to the right error screen. Fully unit-tested.

This is **Step 2 of 10** in a stacked PR chain — depends on `step-1-qr-login-foundation`.

### Stacked PR chain
- Step 2: #15734 — token exchange REST client ← **this PR**

### Test Steps

### Images/gif

N/A — no UI in this slice.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -1,6 +1,110 @@
 package com.woocommerce.android.network.qrlogin
 
+import com.google.gson.Gson
+import com.google.gson.annotations.SerializedName
 import com.woocommerce.android.ui.login.qrlogin.Secret
+import com.woocommerce.android.util.CoroutineDispatchers
+import com.woocommerce.android.util.WooLog
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.withContext
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import javax.inject.Inject
+import javax.inject.Named
+
+/**
+ * Calls the QR login token-exchange endpoint on the merchant's WordPress site.
+ *
+ * The endpoint is unauthenticated — possession of a valid, unexpired, single-use token is the
+ * authorization.
+ */
+class QrLoginRestClient @Inject constructor(
+    @Named("custom-ssl") private val okHttpClient: OkHttpClient,
+    private val gson: Gson,
+    private val dispatchers: CoroutineDispatchers
+) {
+
+    suspend fun exchange(siteUrl: String, token: String): Result<QrLoginCredentials> =
+        withContext(dispatchers.io) {
+            try {
+                Result.success(performExchange(siteUrl, token))
+            } catch (ce: CancellationException) {
+                throw ce
+            } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
+                Result.failure(t)
+            }
+        }
+
+    private fun performExchange(siteUrl: String, token: String): QrLoginCredentials {
+        val request = buildExchangeRequest(siteUrl, token)
+        okHttpClient.newCall(request).execute().use { response ->
+            if (!response.isSuccessful) throw QrLoginExchangeException.HttpError(response.code)
+            return parseExchangeBody(response.body.string())
+                ?: throw QrLoginExchangeException.MalformedResponse
+        }
+    }
+
+    private fun buildExchangeRequest(siteUrl: String, token: String): Request {
+        // The parser only hands us URLs that survived HttpUrl validation, so this should never
+        // be null in practice — fall through to the generic catch if it ever is.
+        val baseUrl = siteUrl.toHttpUrlOrNull()
+            ?: throw IllegalArgumentException("siteUrl could not be parsed by HttpUrl")
+        val exchangeUrl = baseUrl.newBuilder()
+            .addPathSegments(EXCHANGE_PATH_SEGMENTS)
+            .build()
+        val payload = gson.toJson(ExchangeRequest(token))
+        return Request.Builder()
+            .url(exchangeUrl)
+            .post(payload.toRequestBody(JSON_MEDIA_TYPE))
+            .build()
+    }
+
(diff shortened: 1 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -0,0 +1,153 @@
...
+        }
+        return credentials
+    }
+
+    private fun mapHttpStatus(code: Int): QrLoginExchangeException = when (code) {
+        HTTP_UNAUTHORIZED, HTTP_FORBIDDEN -> QrLoginExchangeException.TokenRejected
+        HTTP_NOT_FOUND -> QrLoginExchangeException.EndpointMissing
+        HTTP_TOO_MANY_REQUESTS -> QrLoginExchangeException.RateLimited
+        else -> QrLoginExchangeException.HttpError(code)
+    }
```
JorgeMucientes, line 87: I think this status-only mapping might be a bit too broad. The backend can return the same status codes for different cases, for example invalid_token / token_expired are 401, but ssl_required is 403, and user_not_found is 404 while a missing route can also be 404. If we only map by status, we may show the wrong UI state, like treating an SSL/config issue as an expired QR code, or treating user_not_found as an unsupported plugin. Could we parse the WP error code from the response body first and only fall back to status mapping when the code is missing or unknown?
Something like: 
```kotlin
private fun mapHttpError(code: Int, body: String?): QrLoginExchangeException {
    val errorCode = body
        ?.runCatching { gson.fromJson(this, ErrorResponse::class.java).code }
        ?.getOrNull()

    return when (errorCode) {
        "invalid_token",
        "token_expired" -> QrLoginExchangeException.TokenRejected

        "rate_limit_exceeded" -> QrLoginExchangeException.RateLimited

        "rest_no_route" -> QrLoginExchangeException.EndpointMissing

        "ssl_required",
        "insecure_site_url",
        "user_not_found",
        "application_passwords_unavailable",
        "application_password_failed" -> QrLoginExchangeException.HttpError(code)

        else -> mapHttpStatus(code)
    }
}
```

And change this line:

```if (!response.isSuccessful) throw mapHttpStatus(response.code)```

to read the error body:
```
if (!response.isSuccessful) {
    throw mapHttpError(response.code, response.body.string())
}
```

File: WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt [hunk 2]
```diff
@@ -0,0 +1,153 @@
...
+    private fun parseExchangeBody(body: String): QrLoginCredentials? {
+        // Let JsonSyntaxException propagate so mapException converts it to MalformedResponse.
+        val response = gson.fromJson(body, ExchangeResponse::class.java)
+        val credentials = response?.toCredentials()
+        if (credentials == null) {
+            WooLog.w(
+                WooLog.T.LOGIN,
+                "QR login exchange response missing required fields: ${response.missingFields()}"
+            )
+        }
```
JorgeMucientes, line 78: Small edge case: if the server returns null as the response body, response will be null. The code handles that at first with response?.toCredentials(), but then the log line calls response.missingFields() without the ?, so it would crash there. That crash would be reported as Unknown instead of MalformedResponse. Could we handle the null response before calling missingFields()?

JorgeMucientes (changes requested):

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
malinajirka: I have spent some time thinking about this and exploring how it could look like. But I always end up with exactly the same buckets as we have now. Let me share my reasoning.

| Backend code | Status | Today's bucket | After body-code mapping | Reachable? |
  |---|---|---|---|---|
  | `invalid_token` | 401 | TokenRejected ✓ | TokenRejected | yes (common) |
  | `token_expired` | 401 | TokenRejected ✓ | TokenRejected | yes (common) |
  | `rate_limit_exceeded` | 429 | RateLimited ✓ | RateLimited | yes |
  | `application_password_failed` | 500 | ServerError ✓ | ServerError | yes (rare) |
  | `rest_no_route` | 404 | EndpointMissing ✓ | EndpointMissing | **no**  |
  | `user_not_found` | 404 | EndpointMissing ✗ | ServerError | **no** |
  | `ssl_required` | 403 | TokenRejected ✗ | ServerError | **no** |
  | `insecure_site_url` | 500 | ServerError ✓ | ServerError | **no** |
  | `application_passwords_unavailable` | 501 | ServerError ✓ | ServerError | **no** |

All the "new" states are actually not reachable -> e.g. if woo is outdated, they couldn't have generated the qr code... I assume we don't generate invalid QR code for sites without SSL, right? etc.. 

Bottom line is all the "new" states should be fixed on the backend side (if they are not already) and QR could never be generated for them. If we receive such QR code in the app, that kind of suggest someone is trying to create an artificial one and the less info we share the more secure the solution is.

So all in all, I think the current approach is actually more secure and more specific mapping doesn't bring any value AFAICT. Wdyt @JorgeMucientes, am I missing something?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
malinajirka: Good catch — fixed in 0ac8918.

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
JorgeMucientes: Thanks for looking deeper into this @malinajirka 

> All the "new" states are actually not reachable -> e.g. if woo is outdated, they couldn't have generated the qr code... I assume we don't generate invalid QR code for sites without SSL, right? etc..

Yeah, given the finer grained mapping would only impact "unreachable" flows, I agree it doesn't add any value. Fine with keeping the existing implementation

JorgeMucientes (approved):

PR #15735, opened 2026-04-26
[WOOMOB-2844] Step 3: QR login — authenticator orchestrating exchange + site fetch + eligibility

malinajirka:
Fixes WOOMOB-2844

### Description

Adds `QrLoginAuthenticator`, which sequences the moving parts behind a single `Result<LocalSiteId>`: ticket exchange (via the rest client from step 2) → site lookup → eligibility check → Application Password persistence. On any post-exchange failure the AP is revoked so the merchant isn't left with a stranded credential on their site. Unit-tested including the AP-revocation paths.

This is **Step 3 of 10** in a stacked PR chain — depends on `step-2-qr-login-exchange-client`.

### Stacked PR chain
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility ← **this PR**

### Test Steps

### Images/gif

N/A — no UI in this slice.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
@@ -48,7 +48,15 @@ class QrLoginAuthenticator @Inject constructor(
             username = credentials.userLogin,
             password = credentials.applicationPassword.reveal()
         )
-        ensureUserEligible(site)
+        try {
+            ensureUserEligible(site)
+        } catch (ce: CancellationException) {
+            wpApiSiteRepository.deleteApplicationPassword(site.id)
+            throw ce
+        } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
+            wpApiSiteRepository.deleteApplicationPassword(site.id)
+            throw t
+        }
         selectedSite.set(site)
         return site.id
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
@@ -1,0 +1,51 @@
+package com.woocommerce.android.ui.login.qrlogin
+
+import com.woocommerce.android.network.qrlogin.QrLoginCredentials
+import com.woocommerce.android.network.qrlogin.QrLoginRestClient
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
```diff
@@ -0,0 +1,95 @@
...
+            throw ce
+        } catch (e: QrLoginAuthenticationException) {
+            Result.failure(e)
+        } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
+            WooLog.e(WooLog.T.LOGIN, "QR login authentication failed", t)
+            Result.failure(t)
+        }
+    }
+
+    private suspend fun authenticateWithCredentials(siteUrl: String, credentials: QrLoginCredentials): Int {
```
malinajirka, line 48: 📓 This logic is kind of copied from the pre-existing login flow - I wanted to avoid modifying the existing flows and risking regressions, until we have data that this flow will stay long term. I'm doubting myself now though, wdyt?

JorgeMucientes: I see this is mostly following the existing AP login flow, which imho makes sense. The part I’m unsure about is the new QR-specific cleanup path: after the exchange succeeds, the server has already created a real Application Password, but `deleteApplicationPassword(localSiteId)` only removes the local stored credentials. Since the PR description says post-exchange failures revoke the AP, should this cleanup use the existing `SiteStore.deleteApplicationPassword(site)` path instead, so the server-side AP is revoked too?

JorgeMucientes (approved):
Pre-approving but I think it's worth cleaning the created AP on the server when the login flow fails.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginAuthenticator.kt
malinajirka: Nice catch! Fixed in https://github.com/woocommerce/woocommerce-android/pull/15735/commits/db5d4ecc8c23f04abb2dfc4594d85dd1d83b0205

malinajirka: Actually, I'm still working on one more iteration - we should remove everything on a failed login attempt, not just the app password.

malinajirka:
For some reason, I can't follow up on your comment - either case, I have added the AP deletion, good call!

PR #15736, opened 2026-04-26
[WOOMOB-2845] Step 4: QR login — scanner ViewModel state machine and feature-flag gate

malinajirka:
Fixes WOOMOB-2845

### Description

Adds `QrLoginScannerViewModel` (state for authenticating / endpoint-missing / pending-confirmation / current-error, plus `onScanResult`, `onConfirmSite`, `onStartOver`, `onRetryExchange`) and a `QrLoginAvailability` gate that hides QR entirely when the device has no camera or the feature flag is off. The VM holds the last successfully-parsed ticket so transient errors can replay the same exchange via `onRetryExchange` instead of bouncing the user back to the scanner.

This is **Step 4 of 10** in a stacked PR chain — depends on `step-3-qr-login-authenticator`.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events
- Step 2: #15734 — token exchange REST client
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility
- Step 4: #15736 — scanner ViewModel state machine and feature-flag gate ← **this PR**
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate
- Step 6: #15738 — camera viewfinder, scanner screen, and endpoint-missing fallback
- Step 7: #15739 — fullscreen confirm + error screens with retry-eligible mapping
- Step 8: #15740 — prologue and scanner fragments hosting the Compose surfaces
- Step 9: #15741 — activity wiring and `woocommerce://qr-login` deep link
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR

### Test Steps

N/A — see the PR for `step-10-qr-login-install-qr-detection` for the full feature test plan. This PR is part of a stacked chain (`status: do not merge`) and is unit-test-covered where applicable.

### Images/gif

N/A — no UI in this slice.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15737, opened 2026-04-26
[WOOMOB-2846] Step 5: QR login — prologue screen with QR icon, URL pill, and camera-permission gate

malinajirka:
Fixes WOOMOB-2846

### Description

Adds `QrLoginPrologueScreen` — the first screen the merchant sees when they tap "Log In" on a fresh install: a QR icon, hero copy explaining what's about to happen, the `woo.com/mobilelogin` URL pill so they know where to find the QR on desktop, and a "Sign in with site address instead" fallback link. The camera permission is requested *here*, while the prologue is on screen, so the system permission prompt overlays the purple background instead of the black scanner surface that briefly appeared in earlier iterations. The hero is wrapped in a scrollable column so the fallback link stays visible in landscape on phones.

This is **Step 5 of 10** in a stacked PR chain — depends on `step-4-qr-login-availability-and-viewmodel`.

### Stacked PR chain
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate ← **this PR**

### Test Steps

### Images/gif

See test plan in step-10 PR — prologue screen is exercised end-to-end there.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueScreen.kt
@@ -5,6 +5,7 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.asPaddingValues
@@ -13,12 +14,17 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.systemBars
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.colorResource
@@ -28,6 +34,10 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import com.woocommerce.android.R
+import com.woocommerce.android.ui.compose.component.WCColoredButton
+import com.woocommerce.android.ui.compose.component.WCTextButton
+import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
+import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
 
 @Composable
 fun QrLoginPrologueScreen(
@@ -62,6 +72,7 @@ fun QrLoginPrologueScreen(
             verticalArrangement = Arrangement.SpaceBetween
         ) {
             Hero()
+            Buttons(onScanClicked = onScanClicked, onFallbackClicked = onFallbackClicked)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/res/values/strings.xml
```diff
@@ -212,6 +212,14 @@
 
     <string name="login_prologue_start_new_store">Starting a new store?</string>
 
+    <!-- QR login prologue -->
+    <string name="login_qr_prologue_title">Scan to sign in</string>
+    <string name="login_qr_prologue_subtitle">On a device where you\'re already signed in, visit:</string>
```
malinajirka, line 217: 📓  heads-up — this copy is reworded to **"On your computer, visit:"** in the chain's final PR (#15742) to remove the ambiguity around "device".

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueScreen.kt
```diff
@@ -0,0 +1,230 @@
...
+@Composable
+private fun Buttons(
+    onScanClicked: () -> Unit,
+    onFallbackClicked: () -> Unit
+) {
+    Column(
+        modifier = Modifier.fillMaxWidth(),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        WCColoredButton(
```
malinajirka, line 204: :note: heads-up — in light mode this button's default purple container blends into the purple prologue background. Fixed in #15742 (final PR of the chain) by passing explicit white-background `colors` to `WCColoredButton`, mirroring the existing `Woo.Button.Colored.White` style used by the legacy prologue. Don't bother flagging it here.

WooCommerce/src/main/res/values/colors_base.xml
```diff
@@ -108,6 +108,10 @@
     <color name="prologue_login_background_color">@color/woo_purple_70</color>
     <color name="prologue_login_shape_color">@color/woo_purple_40</color>
     <color name="prologue_login_button_color">@color/white</color>
+    <color name="prologue_login_on_background">@color/woo_white</color>
+    <color name="prologue_login_on_background_secondary">@color/woo_white_alpha_087</color>
+    <color name="prologue_login_on_background_tertiary">@color/woo_white_alpha_060</color>
+    <color name="prologue_login_url_badge_background">#2EFFFFFF</color>
```
malinajirka, line 114: :note: heads-up — this hardcoded `#2EFFFFFF` (~18% white-alpha) is replaced in #15742 (final PR of the chain) with the existing `@color/woo_white_alpha_012` token from the established alpha scale, accepting a marginal opacity shift to avoid introducing a one-off token. Same swap is applied in `values-night/colors_base.xml`. Don't bother flagging it here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueScreen.kt
```diff
@@ -0,0 +1,230 @@
...
+        contract = ActivityResultContracts.RequestPermission(),
+        onResult = { granted ->
+            // The scanner asks for the permission again on first composition; if the user just
+            // granted it here, that follow-up returns immediately with no extra dialog.
+            // If the user denied, we stay on the prologue — they can tap again to retry, or
+            // pick the site-URL fallback. We don't navigate to a black scanner-with-permission-
+            // prompt screen.
+            if (granted) onScanClicked()
+        }
+    )
```
JorgeMucientes, line 65: Haven't seen this handled in other PRs so will report as part of this one. I think we can improve how we handle the permission denial here, because right now we are just dismissing the dialog and don't provide any clear fallback instructions. 
This is specially "painful" when you deny the permission a couple times and then you just can't move forward with the "Scan QR code" button because it doesn't have any effect: 

https://github.com/user-attachments/assets/91462011-2046-445a-afe3-c41477b9912e

If the user denies permission first time we could show something like: 
```
Title: Camera access is needed to scan your QR code
Body: Without camera access, you can still log in by entering your store address.
Primary CTA: Allow camera access
Secondary CTA: Enter store address instead
```

If the user denies permanently when they tap again in "scan QR code" button we could show: 
```
Title: Camera access is turned off
Body: Enable camera access in Settings to scan your QR code, or enter your store address to log in another way.
Primary CTA: Open Permissions Settings
Secondary CTA: Enter store address instead
```

JorgeMucientes (approved):
I'm approving, however we need to improve the Camera permission handling as shared in the comment. No need to do it as part of this PR though.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueScreen.kt
malinajirka: Nice catch! I thought of this when I was implementing it but forgot about it. I created an issue here - #WOOMOB-3008

PR #15741, opened 2026-04-26
[WOOMOB-2850] Step 9: QR login — activity wiring and woocommerce://qr-login deep link

malinajirka:
Fixes WOOMOB-2850

### Description

Wires the QR flow into `LoginActivity`: gates entry on `QrLoginAvailability` (camera + feature flag), hooks up the prologue's "Scan QR code" / fallback listeners, and registers the `woocommerce://qr-login?token=…&siteUrl=…` deep link so wp-admin's QR can launch the merchant straight into the confirm screen on a freshly installed app. With this PR, the feature is wired together end-to-end.

This is **Step 9 of 10** in the QR login stacked PR chain. The follow-up step (`step-10-qr-login-install-qr-detection`) detects the wp-admin onboarding *install* QR and is the final step of the chain — its PR carries the full end-to-end test plan. The `status: do not merge` label is applied so this lands only after PRs 1–8 are reviewed and ready.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events
- Step 2: #15734 — token exchange REST client
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility
- Step 4: #15736 — scanner ViewModel state machine and feature-flag gate
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate
- Step 6: #15738 — camera viewfinder, scanner screen, and endpoint-missing fallback
- Step 7: #15739 — fullscreen confirm + error screens with retry-eligible mapping
- Step 8: #15740 — prologue and scanner fragments hosting the Compose surfaces
- Step 9: #15741 — activity wiring and `woocommerce://qr-login` deep link ← **this PR**
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR

### Test Steps

N/A — see the PR for `step-10-qr-login-install-qr-detection` for the full feature test plan. This PR is part of a stacked chain (`status: do not merge`) and is unit-test-covered where applicable.

### Images/gif

See test plan in step-10 PR — the full QR login flow is exercised end-to-end there.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15740, opened 2026-04-26
[WOOMOB-2849] Step 8: QR login — prologue and scanner fragments hosting the Compose surfaces

malinajirka:
Fixes WOOMOB-2849

### Description

Adds `QrLoginPrologueFragment` and `QrLoginScannerFragment` — the Fragment hosts that wire the Compose screens (steps 5–7) into the existing login navigation. Plumbs the post-login handoff back to the listener, handles deep-link entry vs camera entry differently for back-press / cancel (deep-link entry exits the QR flow on cancel; camera entry returns to the prologue), and passes the ViewModel callbacks into the screens including the `onRetryExchange` retry-with-same-ticket wiring.

This is **Step 8 of 10** in a stacked PR chain — depends on `step-7-qr-login-confirm-and-error-screens`.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events
- Step 2: #15734 — token exchange REST client
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility
- Step 4: #15736 — scanner ViewModel state machine and feature-flag gate
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate
- Step 6: #15738 — camera viewfinder, scanner screen, and endpoint-missing fallback
- Step 7: #15739 — fullscreen confirm + error screens with retry-eligible mapping
- Step 8: #15740 — prologue and scanner fragments hosting the Compose surfaces ← **this PR**
- Step 9: #15741 — activity wiring and `woocommerce://qr-login` deep link
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR

### Test Steps

N/A — see the PR for `step-10-qr-login-install-qr-detection` for the full feature test plan. This PR is part of a stacked chain (`status: do not merge`) and is unit-test-covered where applicable.

### Images/gif

See test plan in step-10 PR — fragments are exercised end-to-end there.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15739, opened 2026-04-26
[WOOMOB-2848] Step 7: QR login — fullscreen confirm + error screens with retry-eligible mapping

malinajirka:
Fixes WOOMOB-2848

### Description

Adds the post-scan fullscreen routing: `QrLoginConfirmSiteScreen` ("Sign in to this store?" with the host rendered in a purple pill — hero is now scrollable so Connect/Cancel stay visible in landscape), `QrLoginErrorScreen` (titled message + primary/secondary actions, with copy that varies per `ErrorReason`), `QrLoginAuthenticatingScreen` (fullscreen "Signing you in…" with a centered spinner — replaces the centered ProgressDialog on a black scrim that earlier iterations used), and the `isRetryEligible()` mapper so transient failures (network / server / rate-limited) replay the same ticket via `onRetryExchange` while terminal failures stay on the "scan a new code" / start-over path.

This is **Step 7 of 10** in a stacked PR chain — depends on `step-6-qr-login-scanner-screen`.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events
- Step 2: #15734 — token exchange REST client
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility
- Step 4: #15736 — scanner ViewModel state machine and feature-flag gate
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate
- Step 6: #15738 — camera viewfinder, scanner screen, and endpoint-missing fallback
- Step 7: #15739 — fullscreen confirm + error screens with retry-eligible mapping ← **this PR**
- Step 8: #15740 — prologue and scanner fragments hosting the Compose surfaces
- Step 9: #15741 — activity wiring and `woocommerce://qr-login` deep link
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR

### Test Steps

N/A — see the PR for `step-10-qr-login-install-qr-detection` for the full feature test plan. This PR is part of a stacked chain (`status: do not merge`) and is unit-test-covered where applicable.

### Images/gif

See test plan in step-10 PR — all confirm + error screens are exercised end-to-end there.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15738, opened 2026-04-26
[WOOMOB-2847] Step 6: QR login — camera viewfinder, scanner screen, and endpoint-missing fallback

malinajirka:
Fixes WOOMOB-2847

### Description

Adds the camera-side UI: `QrLoginViewfinder` (centered cutout, corner brackets, "Visit woo.com/mobilelogin" pill rendered above the scrim with high-contrast styling so it's readable on light camera previews), `QrLoginScannerScreen` routing camera + endpoint-missing + signing-in states inside a single Box so a non-black background is always visible, and the fullscreen "This store can't complete QR login" screen for stores whose WooCommerce plugin doesn't expose the QR endpoint.

This is **Step 6 of 10** in a stacked PR chain — depends on `step-5-qr-login-prologue-screen`.

### Stacked PR chain
- Step 1: #15733 — foundation: payload type, parser, and analytics events
- Step 2: #15734 — token exchange REST client
- Step 3: #15735 — authenticator orchestrating exchange + site fetch + eligibility
- Step 4: #15736 — scanner ViewModel state machine and feature-flag gate
- Step 5: #15737 — prologue screen with QR icon, URL pill, and camera-permission gate
- Step 6: #15738 — camera viewfinder, scanner screen, and endpoint-missing fallback ← **this PR**
- Step 7: #15739 — fullscreen confirm + error screens with retry-eligible mapping
- Step 8: #15740 — prologue and scanner fragments hosting the Compose surfaces
- Step 9: #15741 — activity wiring and `woocommerce://qr-login` deep link
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR

### Test Steps

N/A — see the PR for `step-10-qr-login-install-qr-detection` for the full feature test plan. This PR is part of a stacked chain (`status: do not merge`) and is unit-test-covered where applicable.

### Images/gif

See test plan in step-10 PR — viewfinder + scanner screen are exercised end-to-end there.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15742, opened 2026-04-27
[WOOMOB-2854] Step 10: QR login — detect wp-admin install QR and guide users to the sign-in QR

malinajirka:
Fixes WOOMOB-2854

### Description

Final step of the stacked QR login PR series. The wp-admin onboarding flow displays a QR that links to `https://woocommerce.com/mobile/?utm_source=wc_onboarding_mobile_task` so users can install the app from the App Store / Play Store. If a merchant who already has the app installed scans that install QR, the chain (steps 1–9) currently surfaces the generic **"Not a WooCommerce code"** error, which doesn't explain why or what to do next.

This PR detects the install QR specifically and shows a dedicated, actionable error: **"That's the install QR — the app is already installed. Tap _App is installed_ in your wp-admin onboarding to reveal the sign-in QR. Don't see that button? Visit *woo.com/mobilelogin* on your computer."**

This PR also folds in three small follow-ups across the chain that surfaced during review on the previous round:
- Detekt fix for `QrLoginPayloadParser.queryParam` (loop-with-too-many-jumps refactor).
- Landscape scrolling fixes on `QrLoginPrologueScreen` and `QrLoginConfirmSiteScreen` so the bottom CTAs stay reachable on phones.
- Light-mode polish on the prologue: the **Scan QR code** button now uses an explicit white-on-purple style instead of the default purple-on-purple that blended into the background; the URL pill's hardcoded `#2EFFFFFF` is replaced with the existing `@color/woo_white_alpha_012` token. Subtitle reworded from *"On a device where you're already signed in, visit:"* → *"On your computer, visit:"* to remove the phone-vs-computer ambiguity.

Implementation:
- Adds `QrLoginPayload.InstallQrCode` to the parser sealed hierarchy.
- Parser detects `https://woocommerce.com/mobile[/...]` (case-insensitive host, https only) — utm parameters are accepted, unrelated paths are rejected.
- New `ErrorReason.InstallQrCode` routed to a dedicated `QrLoginErrorContent` mapping with install-QR-specific copy.
- New strings: `login_qr_scanner_error_install_qr_title` / `_body`.
- Unit tests cover the install URL with utm parameters, without query, mixed-case host, and rejection of unrelated paths / http URLs.

This is **Step 10 of 10** in the QR login stacked PR chain — depends on `step-9-qr-login-activity-wiring`. Test plan below covers the full feature end-to-end since this is the final PR before the chain is ready to merge.

### Stacked PR chain
- Step 10: #15742 — detect wp-admin install QR and guide users to the sign-in QR ← **this PR**

https://github.com/user-attachments/assets/2ceffa07-0cb2-49ed-9023-133750d453b3

#### Manual test steps (for a human reviewer)

These cover the main paths and don't require apifaker / mock-server setup.

- [x] **Happy path** — open the app on a fresh install, tap *Log In* → *Scan QR code* → grant camera → scan a real sign-in QR from your own wp-admin → the *Sign in to this store?* screen appears with your store host in a purple pill → tap *Continue* → *Signing you in…* → land in the store dashboard.
- [x] **Cancel** — repeat the above but tap *Cancel* on the confirm screen → returns to the prologue (NOT a black screen).
- [x] **Install QR (this PR's main scenario)** — point the camera at the wp-admin onboarding install QR, OR open `https://woocommerce.com/mobile/?utm_source=wc_onboarding_mobile_task` from a desktop browser as a printed QR → fullscreen *"That's the install QR"* with body explaining how to find the sign-in QR (the *App is installed* button) and the `woo.com/mobilelogin` fallback. Tap *Scan a new code* → returns to scanner cleanly.
- [x] **Wrong / non-Woo QR** — scan any random QR (e.g. a URL QR for `google.com`) → fullscreen *"Not a WooCommerce code"* → *Scan a new code* → exits cleanly.
- [x] **No camera / feature flag off** — the QR_LOGIN feature flag is debug-only in this build; on a release build (or with the flag flipped off) tap *Log In* → it should route directly to the legacy site-URL flow, skipping the QR prologue. (You can simulate by flipping the flag locally.)
- [x] **Landscape on phone** — rotate the device on the prologue and on the confirm screen; the hero scrolls and the fallback / Cancel CTAs stay visible at the bottom edge.
- [x] **Light + dark mode chrome** — toggle the system theme on the prologue; the *Scan QR code* button is clearly visible in both themes (NOT purple-on-purple), and the *Visit `woo.com/mobilelogin`* pill is readable in both.

#### AI testing (the error-state coverage)

The exchange-error states (`401`, `5xx`, `429`, `404`, network unreachable, malformed deep link, retry-with-same-ticket) need apifaker rules and an emulator with the `verify-on-device` skill to exercise reliably. Paste the prompt below into Claude (or your AI agent of choice) — it should pick up from there.

<details>
<summary><strong>Prompt for the AI agent</strong></summary>

```
Verify the QR login error-state coverage for PR #15742 on the running emulator.

Setup:
- App is already installed and on the login prologue. Don't rebuild if it's already running.
- Use the apifaker library (libs/apifaker) to mock responses for the
  /wp-json/wc-admin/mobile-app/qr-login-exchange endpoint.
- For each scenario below: trigger via the deep link template
  `adb shell 'am start -a android.intent.action.VIEW -p com.woocommerce.android.dev     -d "woocommerce://qr-login?token=tok&siteUrl=https%3A%2F%2Fmystore.example"'`
  unless otherwise specified, then use the verify-on-device skill (element list +
  logcat — no screenshots needed for the assertion itself) to confirm each expected
  state, and finally take ONE screenshot per scenario for the PR write-up.

Scenarios:
1. 401 token rejected → fullscreen "Code expired" → tap "Scan a new code" exits
   cleanly. Verify the error reason is `TokenRejected` (NOT retry-eligible — there
   should be no "Try again" path back to the same exchange).
2. 5xx server error → fullscreen "Something went wrong" → tap "Try again" replays
   the exchange with the same ticket (NOT the scanner). Confirm the second call
   uses the same token (apifaker request log).
3. 429 rate limited → fullscreen "Too many attempts" → "Try again" retries.
4. 404 endpoint missing → fullscreen "This store can't complete QR login" → tap
   "Try scanning again" exits cleanly. "Enter site URL instead" switches to the
   legacy URL flow.
5. Network unreachable host (siteUrl that won't resolve, no apifaker rule needed)
   → fullscreen "Couldn't reach your store" → "Try again" retries.
6. Malformed deep link (`woocommerce://qr-login?invalid=true`) → "Not a WooCommerce
   code" → "Scan a new code" exits cleanly.
7. Process restart from recents AFTER deep-link entry → confirm that re-launching
   from recents does NOT replay the single-use token (the scanner should NOT
   restart the exchange on cold start; verify via logcat that no second exchange
   request fires).
8. Authenticating state UX → trigger any successful exchange and verify the
   "Signing you in…" view fills the screen (centered spinner on the surface
   color, NOT a small ProgressDialog on a black scrim).

Report: a table with one row per scenario (status, observed state, screenshot
path, any deviations).
```

</details>

### Images/gif

Screenshots captured during local verification on emulator-5554 are attached to the Linear issue. The AI run above (when it lands) will produce per-error-state screenshots.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
@@ -28,7 +28,7 @@ import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import com.woocommerce.android.R
 import com.woocommerce.android.ui.compose.component.WCColoredButton
-import com.woocommerce.android.ui.compose.component.WCOutlinedButton
+import com.woocommerce.android.ui.compose.component.WCTextButton
 import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
 import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
 
@@ -74,9 +74,10 @@ fun QrLoginErrorScreen(
             modifier = Modifier.fillMaxWidth()
         )
         Spacer(Modifier.height(dimensionResource(id = R.dimen.major_75)))
-        WCOutlinedButton(
+        WCTextButton(
             onClick = onSecondaryClicked,
             text = stringResource(id = content.secondaryAction),
+            allCaps = false,
             modifier = Modifier.fillMaxWidth()
         )
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
@@ -19,7 +19,11 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.dimensionResource
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.withStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import com.woocommerce.android.R
@@ -58,7 +62,7 @@ fun QrLoginErrorScreen(
         )
         Spacer(Modifier.height(dimensionResource(id = R.dimen.major_100)))
         Text(
-            text = stringResource(id = content.body),
+            text = bodyAnnotatedString(content),
             style = MaterialTheme.typography.bodyLarge,
             color = MaterialTheme.colorScheme.onSurfaceVariant,
             textAlign = TextAlign.Center
@@ -91,5 +95,31 @@ private fun QrLoginErrorScreenPreview() {
             onPrimaryClicked = {},
             onSecondaryClicked = {},
         )
+    }
+}
+
+/**
+ * Resolves the body text for an error. When [QrLoginErrorContent.bodyHighlightedArgs] is empty
+ * we just return the body string as-is. Otherwise we treat the body as a `%1$s, %2$s, …`
+ * template, substitute each arg in order, and apply a SemiBold span to each substituted run.
+ * This pattern keeps `<b>…</b>` markup out of strings.xml so translators can't accidentally
+ * drop or break it.
+ */
+@Composable
+private fun bodyAnnotatedString(content: QrLoginErrorContent): AnnotatedString {
+    val template = stringResource(id = content.body)
+    if (content.bodyHighlightedArgs.isEmpty()) return AnnotatedString(template)
+    val args = content.bodyHighlightedArgs.map { stringResource(id = it) }
+    return buildAnnotatedString {
+        var cursor = 0
+        args.forEachIndexed { index, value ->
+            val placeholder = "%${index + 1}\$s"
+            val placeholderStart = template.indexOf(placeholder, startIndex = cursor)
+            if (placeholderStart < 0) return@forEachIndexed
+            append(template.substring(cursor, placeholderStart))
+            withStyle(SpanStyle(fontWeight = FontWeight.SemiBold)) { append(value) }
+            cursor = placeholderStart + placeholder.length
+        }
+        append(template.substring(cursor))
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
@@ -1,6 +1,7 @@
 package com.woocommerce.android.ui.login.qrlogin
 
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
@@ -10,7 +11,6 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.systemBarsPadding
-import androidx.compose.foundation.background
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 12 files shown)


JorgeMucientes:
Excellent job @malinajirka 👏🏼 🥇 

Given this is the PR where I'm manually testing all the UX and different paths,  I'm leaving here any UX feedback that I come up with from the manual testing. 

#### Inconsistent secondary button CTA

During login all secondary CTAs are "Text Buttons" however in the code expired screen we are using an outlined button: 

https://github.com/user-attachments/assets/74d1453a-b88f-41bc-a1af-e5d1780c1108

#### Canceling the flow returns to QR code scanner, not login prologue
When tapping cancel it should return to login prologue instead of going back to opening the QR code scanner, which may lead to a trap flow, if the QR code is still visible to the camera. See the screen recording. 


https://github.com/user-attachments/assets/89eb3a8a-c3bf-4f4f-afee-f777471fb609

#### Feature Flag disabling not enough
Even with the local FF disabled, I can still scan a QR code and log in using the camera or a third part QR code scanner.

#### Landscape UI can be polished
Screenshot_20260428_172315


It requires scrolling to see the login url. We can perhaps have a landscape layout displaying the big url to the right of “On your computer, visit:” text. 

#### Other
Screenshot 2026-04-28 at 17 30 11
I’d make the woo.com-mobilelogin  available to copy on tapping. Specially handy for users sharing clipboard between phone and desktop (very common in iOS)

malinajirka:
Thanks for the testing @JorgeMucientes, some nice findings!

> Inconsistent secondary button CTA

Fixed in b1eb4da83ed93590d52a4bbd78111c3e20571958

04-error-screen-text-button

> Canceling the flow returns to QR code scanner, not login prologue

Fixed in a57cf7a1829d75ff421c2342e0a5d61e91b98274

> Feature Flag disabling not enough

Hmm, I was aware of this one but it's tricky, since AFAIK we can't modify the manifest. Either case, I added a guard to display the regular login flow when the QR code is scanned and FF disabled in 9b5bf8cd005e29005c72ab78caa5d2714976c671. It's still technically a change compared to production, since there the QR would be ignored, but I think that's fine, wdyt?

> Landscape UI can be polished

Fixed in 21d47019450c924457703a52ec6e6acf948d37c9 and 123a380ac35936ce5e97aa11edac2ef95d9fe961 and 1e3d9bc0a2e49664986e76cbac8eedf872b9b111

07-phone-landscape-fixed

09-tablet-landscape-unchanged

> I’d make the woo.com-mobilelogin available to copy on tapping. 

Great idea! Fixed in 0225087fd0e270c24028590f2e4f1e2b7bed6228

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
```diff
@@ -93,3 +98,29 @@ private fun QrLoginErrorScreenPreview() {
...
+
+/**
+ * Resolves the body text for an error. When [QrLoginErrorContent.bodyHighlightedArgs] is empty
+ * we just return the body string as-is. Otherwise we treat the body as a `%1$s, %2$s, …`
+ * template, substitute each arg in order, and apply a SemiBold span to each substituted run.
+ * This pattern keeps `<b>…</b>` markup out of strings.xml so translators can't accidentally
+ * drop or break it.
+ */
+@Composable
+private fun bodyAnnotatedString(content: QrLoginErrorContent): AnnotatedString {
```
JorgeMucientes: Np, can we not just add this bold in the string res definition: 
```xml
<string name="login_qr_scanner_error_install_qr_body">
    The app's already installed — this QR installs it. In wp-admin, tap the &lt;b&gt;%1$s&lt;/b&gt; button to reveal the sign-in QR. Or visit &lt;b&gt;%2$s&lt;/b&gt; on your computer.
</string>
```
and then remove this function completely and rely on: 
```kotlin
Text(
    text = annotatedStringRes(
        content.body,
        content.bodyHighlightedArgs.map { stringResource(id = it) }.toTypedArray()
    ),
    
)
```

I mean the function comment states `This pattern keeps `<b>…</b>` markup out of strings.xml so translators can't accidentally drop or break it.` 
But I feel this is how we've been doing things so far. Anyway its just a nit no need to change.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayloadParserTest.kt
```diff
@@ -112,6 +112,35 @@ class QrLoginPayloadParserTest : BaseUnitTest() {
...
+    @Test
+    fun `given other woocommerce com path, when parsed, then returns Invalid`() {
+        assertThat(parser.parse("https://woocommerce.com/")).isEqualTo(QrLoginPayload.Invalid)
+        assertThat(parser.parse("https://woocommerce.com/blog")).isEqualTo(QrLoginPayload.Invalid)
+    }
+
+    @Test
+    fun `given http install QR url, when parsed, then returns Invalid`() {
+        assertThat(parser.parse("http://woocommerce.com/mobile/")).isEqualTo(QrLoginPayload.Invalid)
+    }
```
JorgeMucientes: Shouldn't this case display the `QrLoginPayload.InstallQrCode` error?

JorgeMucientes (approved):
Works great! Thanks for addressing the feedback. I just left a couple more nitpicks but no need to apply them.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginErrorScreen.kt
malinajirka: Hmm for some reason I thought we were avoiding having html like this in strings.xml - in the past, I'm sure we were, but we might have changed it at some point and I missed it. Feels quite risky, and kind of impossible to catch AFAICT, but since we are doing it all over the place, I agree this is unnecessarily complex.

malinajirka: Fixed in 9c14b1df5062a4c833bcb4c82c3712f685ff8380

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPayloadParserTest.kt
malinajirka: Good catch! Technically, the QR should always include only https, but I think there is nothing wrong in accepting also http, so fixed in 2ef00e512b828af5ecf23a0e8818b709875bde0a.

PR #15743, opened 2026-04-27
[WOOMOB-2787] cardreader library: payment APIs, protocol, transport

kidinov:
Fixes WOOMOB-2787

Adds the cardreader library plumbing for Woo POS Remote Tap-to-Pay: the phone runs as a Stripe Tap-to-Pay reader, the tablet drives the payment over a TLS socket. This PR is library-only — the activity, ViewModels and POS UI come in follow-up PRs.

**What it adds**
- `CardReaderManager.createPaymentIntent(...)` and `retrieveAndCollectPayment(...)` so the phone can issue and collect a Stripe `PaymentIntent` locally. Backed by `PaymentManager` with the existing actions split.
- `CardReaderRemoteSession.handleCollectPaymentRequest`: phone receives the full `PaymentInfo` over the socket, creates the intent, collects, replies with `PaymentIntentResult`.
- `CardReaderRemoteConnection.closed: StateFlow<Boolean>` so the tablet can detect when the phone drops the session.
- `CardReaderRemoteHeartbeat` (`Ping` every 30s in both directions) to keep the TLS socket alive past its 90s read timeout while the pair is idle.
- `CardReaderRemoteTabletClient.connect(timeoutMillis = 30s)` so a stalled handshake fails instead of hanging.
- TLS server moved off `AndroidKeyStore` to `okhttp3.tls.HeldCertificate` — some ColorOS builds reject the keystore's ECDSA digest combinations.
- NSD TXT records carry device name + 4-char pairing-code suffix so the tablet can label phones (e.g. `CPH2747 · 2846`).
- `LogWrapper` is used everywhere in the remote package; no `android.util.Log` and no `returnDefaultValues = true` in `build.gradle`.

### Test Steps

This PR has no UI — the user-visible flow is exercised by the follow-up PRs. To verify the library in isolation:

1. Build the cardreader module: it must compile under Wasabi/Jalapeno debug.
2. Build the main app — no API breakage in callers.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/cardreader/src/test/java/com/woocommerce/android/cardreader/remote/DefaultCardReaderRemoteTabletClientTest.kt
```diff
@@ -31,7 +38,7 @@
     @Test
     fun `given no active connection, when collectPayment, then returns Failed`() = testBlocking {
         // WHEN
-        val outcome = client.collectPayment(paymentIntentClientSecret = "pi_secret")
+        val outcome = client.collectPayment(paymentInfo = mock<PaymentInfo>())
```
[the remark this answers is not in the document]
kidinov: Fixed in 3661867f8b6 — replaced the `mock<PaymentInfo>()` with a real `PaymentInfo` constructed from the data class.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTabletClient.kt
```diff
@@ -89,16 +127,30 @@ internal class DefaultCardReaderRemoteTabletClient(
...
     }
 
+    private fun bridgeClosedSignal(connection: CardReaderRemoteConnection) {
+        closedBridgeJob?.cancel()
+        closedBridgeJob = scope.launch {
+            connection.closed.collect { isClosed ->
+                _connectionClosed.value = isClosed
+            }
+        }
+    }
```
malinajirka, line 142: ❓ When the underlying connection closes remotely, `_connectionClosed` flips to `true` here, but the `connection` field still points at the dead instance. So the next `collectPayment(...)` call passes the `connection ?: return Failed("Not connected to a remote reader")` guard and then throws an `IOException` from `active.send(...)`, which falls into the generic `Throwable` branch (line 174) and returns `Failed(IOException)` without calling `disconnect()`. End result is a slightly confusing error path and a lingering `connection` reference until the next explicit `disconnect()`.

Wdyt about nulling out `connection` (and cancelling the heartbeat) here when `isClosed == true`, so a follow-up `collectPayment` fails the guard cleanly?

```suggestion
    private fun bridgeClosedSignal(connection: CardReaderRemoteConnection) {
        closedBridgeJob?.cancel()
        closedBridgeJob = scope.launch {
            connection.closed.collect { isClosed ->
                _connectionClosed.value = isClosed
                if (isClosed) {
                    heartbeatJob?.cancel()
                    heartbeatJob = null
                    this@DefaultCardReaderRemoteTabletClient.connection = null
                }
            }
        }
    }
```

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
```diff
@@ -53,6 +56,52 @@ internal class PaymentManager(
...
+            when (val result = createPaymentAction.createPaymentIntent(paymentInfo)) {
+                is Success -> CreatePaymentIntentResult.Success(
+                    paymentIntentId = result.paymentIntent.id.orEmpty(),
+                    clientSecret = result.paymentIntent.clientSecret.orEmpty(),
+                )
+                is Failure -> CreatePaymentIntentResult.Failed(result.exception)
+            }
+        } catch (cancel: CancellationException) {
+            throw cancel
+        } catch (cause: Throwable) {
```
malinajirka: 🔍 np: `catch (cause: Throwable)` catches `OutOfMemoryError`, `LinkageError`, `StackOverflowError`, etc. and turns them into a `Failed` outcome, which masks JVM-level errors that should crash the process. Same pattern shows up across this PR (here, `retrieveAndCollectPayment` line 88, `CardReaderRemoteTabletClient.connect` line 124 and `collectPayment` line 173). Probably safer to narrow them all to at least `Exception`. Wdyt?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
```diff
@@ -53,6 +56,52 @@ internal class PaymentManager(
...
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    suspend fun retrieveAndCollectPayment(clientSecret: String): RetrieveAndCollectResult =
+        try {
+            val retrieved = terminalWrapper.retrievePaymentIntent(clientSecret)
+            val processed = terminalWrapper.processPaymentIntent(retrieved)
+            RetrieveAndCollectResult.Success(
+                paymentIntentId = processed.id.orEmpty(),
+                status = processed.status?.name?.lowercase() ?: "unknown",
```
malinajirka: ❓ With `.orEmpty()` / `?: "unknown"` we report a `Success` to the tablet even when Stripe returned a `PaymentIntent` with a null id or null status. Empty `paymentIntentId` is hard to recover from on the tablet (no way to capture later) and `"unknown"` is indistinguishable from a real status. Wdyt about treating null id/status as `Failed` instead so the tablet sees a real error?

```suggestion
    suspend fun retrieveAndCollectPayment(clientSecret: String): RetrieveAndCollectResult =
        try {
            val retrieved = terminalWrapper.retrievePaymentIntent(clientSecret)
            val processed = terminalWrapper.processPaymentIntent(retrieved)
            val id = processed.id
            val status = processed.status?.name?.lowercase()
            if (id == null || status == null) {
                RetrieveAndCollectResult.Failed(
                    IllegalStateException("processPaymentIntent returned null id or status"),
                )
            } else {
                RetrieveAndCollectResult.Success(paymentIntentId = id, status = status)
            }
        } catch (cancel: CancellationException) {
            throw cancel
        } catch (cause: Exception) {
            RetrieveAndCollectResult.Failed(cause)
        }
```

Same thing on `createPaymentIntentOnly` line 65–66 — an empty `clientSecret` shipped as `Success` will end up failing somewhere downstream on the tablet.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServer.kt
```diff
@@ -35,87 +29,57 @@ internal class CardReaderRemoteTlsServer(
...
 
-        val keyManagerFactory = KeyManagerFactory.getInstance(
-            KeyManagerFactory.getDefaultAlgorithm()
-        ).apply { init(keyStore, null) }
+        val keyManagerFactory = KeyManagerFactory
+            .getInstance(KeyManagerFactory.getDefaultAlgorithm())
+            .apply { init(keyStore, KEYSTORE_PASSWORD) }
 
-        val sslContext = SSLContext.getInstance("TLSv1.3").apply {
+        val sslContext = SSLContext.getInstance("TLS").apply {
```
malinajirka, line 53: ❓ Both server and client moved from `SSLContext.getInstance("TLSv1.3")` to `"TLS"` in this PR, is there any reason for it?

malinajirka (approved):
Thanks @kidinov ! I've left a couple comments/questions.

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTabletClient.kt
kidinov: Fixed in 1a1361228df

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
kidinov: Fixed in 0a6ae07fd7c

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
kidinov: Fixed in e617a8e1349

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServer.kt
kidinov: During my testing I could not get TLSv1.3 to work on my real device, so I relaxed it to `TLS`. Not sure why 1.3 fails while the negotiated version works.

PR #15744, opened 2026-04-27
[WOOMOB-2787] Phone-side Card Reader Mode: location permission and simulated reader

kidinov:
Fixes WOOMOB-2787

The phone runs as a Stripe Tap-to-Pay reader for the Woo POS Remote Tap-to-Pay flow. Stripe Terminal needs `ACCESS_FINE_LOCATION` to discover the built-in TTP reader. The activity now asks for the permission before starting the session, and the ViewModel only kicks off the Stripe SDK once the permission is granted.

**Notable**

- Permission flow: explainer state while the permission is missing → system dialog → either start the session, or show an error if denied.
- `CardReaderModeExit` (object) is replaced by `CardReaderModeEvent` sealed class so the ViewModel can emit both `Exit` and `RequestLocationPermission`.
- Simulated card reader: dev option threaded through `session.start(isSimulated)`. With it enabled, Stripe Terminal uses its simulator instead of real hardware.
- `MockCardReaderManagerModule` (androidTest) gets fakes for the new `createPaymentIntent` / `retrieveAndCollectPayment` API.

### Test Steps

1. Open **Payments → Card Reader Mode** on a TTP-capable phone.
2. If location permission is missing: explainer screen appears with a Continue button. Tap Continue → system permission dialog appears. Deny → error screen with "location required" message. Grant → session starts and Card Reader Mode shows the existing pairing/waiting states.
3. Re-open the screen with permission already granted → session starts immediately.
4. Toggle **Simulated card reader** in developer options → reopen Card Reader Mode → session starts in simulated mode (no real reader needed).

### DIFF
--- a/WooCommerce/src/androidTest/kotlin/com/woocommerce/android/di/MockCardReaderManagerModule.kt
+++ b/WooCommerce/src/androidTest/kotlin/com/woocommerce/android/di/MockCardReaderManagerModule.kt
@@ -15,10 +15,12 @@ import com.woocommerce.android.cardreader.connection.event.SoftwareUpdateAvailab
 import com.woocommerce.android.cardreader.connection.event.SoftwareUpdateStatus
 import com.woocommerce.android.cardreader.payments.CardInteracRefundStatus
 import com.woocommerce.android.cardreader.payments.CardPaymentStatus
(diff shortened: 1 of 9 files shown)


samiuelson (approved):
LGTM!

PR #15748, opened 2026-04-27
[WOOMOB-2861] Add dashboard revenue type selector

JorgeMucientes:
⚠️ Do not merge until base branch is trunk

### Description
Fixes WOOMOB-2861

Adds the Material 3 segmented control for switching the Performance card revenue metric between Gross, Net, and Total. The selected metric updates the displayed card total and chart values from the already-loaded revenue stats, clears any selected chart point when the metric changes, and tracks selector usage.

Stacked on the WOOMOB-2859 data-layer PR.

### Test Steps
Use `wc-admin/analytics` section to ensure and compare data matches with the one we display in the app. 

1. Launch the app and open the Dashboard.
2. Find the Performance card and switch between Gross, Net, and Total.
3. Confirm the displayed revenue total and chart update when each segment is selected.
4. Change dashboard date ranges and confirm the selected metric continues to render correctly.

https://github.com/user-attachments/assets/e877feb7-149c-4f9c-bbe4-b190a527321d


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
```diff
@@ -162,6 +162,7 @@ enum class AnalyticsEvent(override val siteless: Boolean = false) : IAnalyticsEv
     DASHBOARD_STATS_CUSTOM_RANGE_ADD_BUTTON_TAPPED,
     DASHBOARD_STATS_CUSTOM_RANGE_CONFIRMED,
     DASHBOARD_STATS_CUSTOM_RANGE_EDIT_BUTTON_TAPPED,
+    DASHBOARD_STATS_REVENUE_TYPE_SELECTED,
```
AdamGrzybkowski, line 165: I know we are in a hurry, but it would be great to not add new events here and use the EvenHorizon - pe5sF9-5fP-p2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -240,6 +264,53 @@ private fun StatsChart(
...
+            id = R.string.dashboard_stats_revenue_type_net
+        )
+    )
+    val segmentedButtonColors = SegmentedButtonDefaults.colors(
+        activeContainerColor = colorResource(id = R.color.color_primary),
+        activeContentColor = colorResource(id = R.color.woo_white),
+    )
+
+    SingleChoiceSegmentedButtonRow(
+        modifier = modifier.padding(start = 16.dp, top = 8.dp, end = 16.dp, bottom = 4.dp)
```
AdamGrzybkowski, line 279: ⛏️ Most of the time, we use dimens definitions in the compose like `dimensionResource(id = dimen.minor_100)` to reuse existing values.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -240,6 +264,53 @@ private fun StatsChart(
...
+        options.forEachIndexed { index, (type, label) ->
+            SegmentedButton(
+                selected = selectedType == type,
+                onClick = { onTypeSelected(type) },
+                shape = SegmentedButtonDefaults.itemShape(index = index, count = options.size),
+                colors = segmentedButtonColors,
+                icon = {},
+                modifier = Modifier
+                    .weight(1f)
+                    .height(36.dp)
```
AdamGrzybkowski: A hardcoded height won't scale well with font scaling.

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -240,6 +264,53 @@ private fun StatsChart(
...
+        DashboardStatsViewModel.RevenueStatsType.TOTAL to stringResource(
+            id = R.string.dashboard_stats_revenue_type_total
+        ),
+        DashboardStatsViewModel.RevenueStatsType.GROSS to stringResource(
+            id = R.string.dashboard_stats_revenue_type_gross
+        ),
+        DashboardStatsViewModel.RevenueStatsType.NET to stringResource(
+            id = R.string.dashboard_stats_revenue_type_net
+        )
+    )
```
AdamGrzybkowski: I think this is okay for now, but I think it is better to keep the list of available options in the VM.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsViewModel.kt
```diff
@@ -108,6 +109,11 @@ class DashboardStatsViewModel @AssistedInject constructor(
     private var _lastUpdateStats = MutableLiveData<Long?>()
     val lastUpdateStats: LiveData<Long?> = _lastUpdateStats
 
+    private val _selectedRevenueStatsType = MutableStateFlow(
+        RevenueStatsType.fromName(appPrefsWrapper.getDashboardRevenueStatsType())
+    )
+    val selectedRevenueStatsType = _selectedRevenueStatsType.asStateFlow()
```
AdamGrzybkowski: To be consistent with the code base, I think we should expose LiveData here.

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
JorgeMucientes: I have mixed feelings about EvenHorizon approach. I think we should further discuss if we want it to be the default mechanism to add new events, and if that is the case everyone should be aligned to ensure we follow that process always. Right now I think we have Adam's POC from recent hackweek but no more than that. 
For this particular 3 events, given we are in the RSM context, PRs are already in flight, both for iOS and Android and we want to have this merged tomorrow EOD I think the overhead is too much. In particular because of the coordination needed now between platforms. 
So, for this time I'd prefer to keep this as is. 
I'm ok with creating a Lnear migration task for this, to tackle when we have more time.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
JorgeMucientes: > Most of the time, we use dimens definitions in the compose like dimensionResource(id = dimen.minor_100) to reuse existing values.

There has been several debates on the past around this: p1713854561214809/1713853806.248619-slack-CGPNUU63E
TLDR: 
> - The main benefit of reusing defined dimens instead of hardcoded values is that in the event of a large UI migration, changing a single value in the dimens file would automatically update it everywhere.  However, the currently defined dimens in XML are not semantic and the same dimen is used for many different purposes in the UI. In case of a migration, all usages will have to be revisited manually, losing the mentioned advantage. 
> - Given there’s not a clear benefit of using defined dimensions anymore, using dimensionresource() in compose code seems like extra boilerplate and makes it less clear when reading compose code which value is being used. Many times we end up clicking on the dimension name (like major_275) to see what is the real value being used.

We use to have some compose-guidelines.md in the repo reflecting this, however, I've just checked and they [have recently been removed in favor of a agen Skill.md](https://github.com/woocommerce/woocommerce-android/commit/afc00ff6995c74f63398b54083594a393647167f#diff-13411a91b1a348eab8a65b80e6d6709ca1530ab8217d53df8b90fbb07552eb8c) and there's no mention to the dimensions usage. 

With the above said I lean towards using dimensions directly in compose code for simplicity.

JorgeMucientes:
Thanks for the careful review @AdamGrzybkowski . You surfaced good improvements and bugs there 🙏🏼 \
Applied most of your feedback and replied to a couple of comments. 
Ready for another round :)

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -65,7 +64,8 @@ fun DashboardStatsCard(
     val revenueStatsState by viewModel.revenueStatsState.observeAsState()
     val visitorsStatsState by viewModel.visitorStatsState.observeAsState()
     val lastUpdateState by viewModel.lastUpdateStats.observeAsState()
-    val selectedRevenueStatsType by viewModel.selectedRevenueStatsType.collectAsStateWithLifecycle()
+    val selectedRevenueStatsType by viewModel.selectedRevenueStatsType
+        .observeAsState(DashboardStatsViewModel.RevenueStatsType.DEFAULT)
```
AdamGrzybkowski: The state should come from the VM, the default is not needed here as with other states in this class.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsViewModel.kt
```diff
@@ -109,10 +108,10 @@ class DashboardStatsViewModel @AssistedInject constructor(
     private var _lastUpdateStats = MutableLiveData<Long?>()
     val lastUpdateStats: LiveData<Long?> = _lastUpdateStats
 
-    private val _selectedRevenueStatsType = MutableStateFlow(
+    private val _selectedRevenueStatsType = MutableLiveData(
```
AdamGrzybkowski, line 113: You can keep the private property here as StateFlow - the comment was only about the public one. Flows are easier to work with in the VM.

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
AdamGrzybkowski: It would be great if you could voice those feelings on a p2. I don't think it makes sense to merge new solutions into the codebase if they won't be used in the future.

AdamGrzybkowski (approved):
LGTM :shipit: Left some comments, but they are not blockers, so approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/analytics/AnalyticsEvent.kt
JorgeMucientes: Will do :) 
Tbh I had a long pending reply to add to Adam's P2 post but never got the proper time and chance to do so with all the many changes going on recently. By the way I'm not against it I just think we haven't discussed team wide all the pros and cons and committed to follow one or another approach, and I think we need to do that before assuming EventHorizon is the new standard.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsViewModel.kt
JorgeMucientes: Yeah, thought about it, but after looking at the rest of the private var in the VM using `MutableLiveData` I lean towards using the same approach for consistency. 

> Flows are easier to work with in the VM.

Agree although for this particular case using `MutableLiveData` didn't make any difference in the VM usage.

PR #15751, opened 2026-04-27
[WOOMOB-2924] Add local feature flag for AI Assistant work

hichamboushaba:
Fixes WOOMOB-2924

### Description

Adds the `AI_ASSISTANT` entry to the `FeatureFlag` enum so that all subsequent AI Assistant work can land behind a feature gate from day one.

The flag uses `localValue = PackageUtils.isDebugBuild()`, so it is on in debug builds and off in release. No production code reads the flag yet — the entry point and DI wiring land in a follow-up ticket (WOOMOB-2918).

> [!NOTE]
> The behavior will be tested later when we add the initial entry point for the feature. 

### Test Steps
Green CI is enough for now.

### Images/gif

N/A


staskus (approved):
👍

PR #15752, opened 2026-04-27
[RSM] Adapt POS dialogs for phone screens

samiuelson:
### Description

POS dialogs (exit confirmation, card reader connection, scanning setup, refund dialogs, etc.) use a fixed width fraction via `WooPosDialogWrapper` — 75% by default, 55% for the card reader dialog. This looks great on tablets but results in narrow, cramped dialogs on phones where titles get truncated and content doesn't use enough screen space.

This PR makes `WooPosDialogWrapper` breakpoint-aware using the existing `WooPosBreakpoint` system:
- **Phone**: dialogs use 95% screen width and are capped at 90% screen height to prevent overflow
- **SmallTablet / Tablet**: behavior is unchanged — the caller's `widthFraction` is used as before

Additionally, the card reader connection dialog title is changed from `maxLines = 1` to `maxLines = 2` with ellipsis, so longer titles (like reader serial numbers) can wrap instead of being truncated on a single line.

All 7+ POS dialogs benefit automatically since they all go through `WooPosDialogWrapper`.

### Test Steps

1. Open POS on a **phone-sized device** (or fold a foldable)
2. Press back to trigger the **exit confirmation dialog** — verify the title is fully visible and the dialog takes up most of the screen width
3. Add a product, check out, and tap "Connect to reader" to see the **card reader connection dialog** — verify the title is not truncated
4. Repeat on a **tablet** and verify dialogs look the same as before (unchanged width)

### Images/gif

image
Screenshot_20260429_144055
Screenshot_20260429_144108
Screenshot_20260429_144113

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
@@ -434,7 +434,7 @@ private fun CardReaderDialogContent(
             color = MaterialTheme.colorScheme.onSurface,
             fontWeight = FontWeight.Bold,
             textAlign = TextAlign.Center,
-            maxLines = 1,
+            maxLines = 2,
             overflow = TextOverflow.Ellipsis,
         )
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
@@ -27,12 +27,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.vector.ImageVector
-import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalWindowInfo
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.res.vectorResource
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
-import androidx.compose.ui.unit.dp
 import com.woocommerce.android.R
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosBreakpoint
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosCornerRadius
@@ -63,8 +63,10 @@ fun WooPosDialogWrapper(
         WooPosBreakpoint.SmallTablet,
         WooPosBreakpoint.Tablet -> widthFraction
     }
+    val density = LocalDensity.current
+    val containerHeight = with(density) { LocalWindowInfo.current.containerSize.height.toDp() }
     val phoneMaxHeight = if (breakpoint == WooPosBreakpoint.Phone) {
-        (LocalConfiguration.current.screenHeightDp * PHONE_MAX_HEIGHT_FRACTION).dp
+        containerHeight * PHONE_MAX_HEIGHT_FRACTION
     } else {
         null
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
@@ -63,9 +63,9 @@ fun WooPosDialogWrapper(
         WooPosBreakpoint.SmallTablet,
         WooPosBreakpoint.Tablet -> widthFraction
     }
-    val density = LocalDensity.current
-    val containerHeight = with(density) { LocalWindowInfo.current.containerSize.height.toDp() }
     val phoneMaxHeight = if (breakpoint == WooPosBreakpoint.Phone) {
+        val density = LocalDensity.current
+        val containerHeight = with(density) { LocalWindowInfo.current.containerSize.height.toDp() }
         containerHeight * PHONE_MAX_HEIGHT_FRACTION
     } else {
         null
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
@@ -14,6 +14,7 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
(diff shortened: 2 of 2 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
```diff
@@ -50,6 +57,17 @@ fun WooPosDialogWrapper(
...
+    val breakpoint = currentWooPosBreakpoint()
+    val adaptiveWidthFraction = when (breakpoint) {
+        WooPosBreakpoint.Phone -> PHONE_WIDTH_FRACTION
+        WooPosBreakpoint.SmallTablet,
+        WooPosBreakpoint.Tablet -> widthFraction
+    }
+    val phoneMaxHeight = if (breakpoint == WooPosBreakpoint.Phone) {
+        (LocalConfiguration.current.screenHeightDp * PHONE_MAX_HEIGHT_FRACTION).dp
+    } else {
+        null
```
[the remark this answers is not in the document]
samiuelson: Good catch! Fixed in f425d81da88 — now using `LocalWindowInfo.current.containerSize.height` with `LocalDensity` to stay consistent with `currentWooPosBreakpoint()` and correctly handle foldables/multi-window.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
```diff
@@ -50,6 +57,19 @@ fun WooPosDialogWrapper(
...
+        WooPosBreakpoint.SmallTablet,
+        WooPosBreakpoint.Tablet -> widthFraction
+    }
+    val density = LocalDensity.current
+    val containerHeight = with(density) { LocalWindowInfo.current.containerSize.height.toDp() }
+    val phoneMaxHeight = if (breakpoint == WooPosBreakpoint.Phone) {
+        containerHeight * PHONE_MAX_HEIGHT_FRACTION
+    } else {
+        null
+    }
```
toupper, line 72: Nice use of the existing `WooPosBreakpoint` abstraction. The `when` is exhaustive so a future breakpoint will fail compilation rather than silently fall through, and the `Modifier.then(...)` pattern below is the idiomatic way to apply a conditional modifier.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
```diff
@@ -50,6 +57,19 @@ fun WooPosDialogWrapper(
...
 ) {
     val closeContentDescription = stringResource(R.string.close)
+    val breakpoint = currentWooPosBreakpoint()
+    val adaptiveWidthFraction = when (breakpoint) {
+        WooPosBreakpoint.Phone -> PHONE_WIDTH_FRACTION
+        WooPosBreakpoint.SmallTablet,
+        WooPosBreakpoint.Tablet -> widthFraction
+    }
+    val density = LocalDensity.current
+    val containerHeight = with(density) { LocalWindowInfo.current.containerSize.height.toDp() }
```
toupper: Minor nit: `containerHeight` is computed on every composition regardless of breakpoint. You could move both `density` and `containerHeight` inside the `if (breakpoint == WooPosBreakpoint.Phone)` branch so the tablet path skips the lookup entirely. Trivial cost as-is, just a tidiness thing.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
```diff
@@ -87,7 +107,14 @@ fun WooPosDialogWrapper(
                 modifier = modifier
                     .statusBarsPadding()
                     .navigationBarsPadding()
-                    .fillMaxWidth(widthFraction),
+                    .fillMaxWidth(adaptiveWidthFraction)
+                    .then(
+                        if (phoneMaxHeight != null) {
+                            Modifier.heightIn(max = phoneMaxHeight)
```
toupper, line 113: 90% of container height is plenty in portrait but pretty tight on a phone in landscape. Worth a manual check on a phone in landscape with a content-heavier dialog (card reader scanning, refunds) to confirm nothing clips awkwardly.

Also worth a quick sanity check that a foldable transitioning folded → unfolded recomposes correctly — `LocalWindowInfo.current.containerSize` should update, but with 7+ dialogs depending on this it's worth eyeballing.

toupper, line 113: Capping `heightIn(max = ...)` means content exceeding the cap will be cut off unless the inner `Column` is scrollable. Worth confirming that dialogs with the most content (refund, scanning setup) scroll rather than truncate, and adding that to the test plan.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
```diff
@@ -434,7 +434,7 @@ private fun CardReaderDialogContent(
             color = MaterialTheme.colorScheme.onSurface,
             fontWeight = FontWeight.Bold,
             textAlign = TextAlign.Center,
-            maxLines = 1,
+            maxLines = 2,
```
toupper, line 437: The change isn't gated on breakpoint, so a long tablet title (0.55 width fraction) will now wrap to 2 lines too instead of ellipsizing on 1. That's almost certainly the better behavior, but worth confirming the dialog's vertical layout still looks balanced (icon/spacing alignment) when a tablet title does wrap.

toupper (commented):
Drive-by review from the iOS side for parity awareness. Approach looks clean — leaving a few small thoughts inline.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
samiuelson: Fixed in a79f2e18d4d

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
samiuelson: Thanks! 🙌

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
samiuelson: Good call — will eyeball card reader scanning and refund dialogs on a phone in landscape, plus a folded → unfolded transition on a foldable, before merging.

samiuelson: Fair point — will verify the content-heavy dialogs (refund, scanning) scroll rather than truncate when the cap kicks in, and add it to the test plan.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionDialog.kt
samiuelson: Good catch — will visually confirm the icon/spacing alignment still looks balanced on tablet when a long title wraps to 2 lines.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosDialogWrapper.kt
samiuelson: POS is locked to portrait on phones and landscape on tablets, so this is out of scope for this PR for now.

samiuelson: I verified all the existing dialogs and didn't find any problems.

toupper (approved):
All comments addressed. `LocalWindowInfo.containerSize` migration is correct and matches the breakpoint system, the tablet branch now skips the density/container lookup, and `maxLines = 2` is in place on the card reader title. Verified the orientation-lock claim in `Activity.lockWooPosOrientation()` (libs/pos util) — phones are pinned to portrait and tablets to landscape, so the phone-landscape concern doesn't apply here. LGTM.

PR #15754, opened 2026-04-27
[WOOMOB-2886] POS card reader: fixes from remote TTP integration bug-bash

kidinov:
### Description
Fixes WOOMOB-2886.

Bug-fix-only PR on top of #15745. Issues surfaced while integration-testing the remote Tap-to-Pay stack (lib + phone + tablet PRs merged together) — most affect the regular BT card reader flow as well, so they're worth fixing before the stack lands.

Grouped by area:

**POS card reader connect / disconnect / discovery**
- `disconnect()` now runs on `Dispatchers.IO` (was on Main → `StrictMode NetworkViolation` from SSL close) and survives a failing `remoteReaderSession.disconnect()` so the BT disconnect always reaches the SDK.
- `disconnectReader()` in the lib now syncs `readerStatus` to `NotConnected` only when our cached state is `Connected` (don't clobber `Reconnecting`, where Stripe SDK's `connectedReader` is null by design).
- `startDiscovery()` cancels Stripe's auto-reconnect (`cancelReconnection()`) when status is `Reconnecting`; defensively disconnects only when `Connecting`/`NotConnected`; no-op when `Connected` (preserves a live connection through dialog dismiss).
- `cancel()` no longer restarts discovery on dismiss (matches trunk) — the leftover discovery job was getting cancelled mid-flight by the next discovery start, leaving Stripe in a bad state where it never reported the BT reader.

**Remote TTP session & state**
- Phone-side `handleConnectRequest` defensively disconnects any leftover reader before TTP discovery (Stripe SDK refuses discovery while a reader is bound).
- `WooPosRemoteReaderSession.fetchToken` now rethrows `CancellationException` instead of misreporting it as a token fetch failure.
- `sendCollectPayment` checks state under the mutex and returns `Failed` if not `Connected`, so we don't run a payment on a stale client.
- `LAST_CONNECTED_PHONE_NAME` is cleared on remote connect failure, so auto-reconnect doesn't keep targeting an unreachable phone.

**Payment UX**
- `CardReaderRemoteTabletClient` now throws a typed connection-lost error when the receive flow ends without a matching reply (was leaking `NoSuchElementException("Expected at least one element matching the predicate")` to the user).
- `WooPosRemoteReaderPaymentFlow.collect` now exposes friendly error strings and an `onCaptureStarting` callback.
- Tablet now transitions to the same full-screen "Processing payment" state as BT once the phone returns from collect (used to stay on totals + "Ready for payment" until success).
- `WooPosCardPaymentViewModel` no longer cancels in-flight BT payments when the reader briefly drops to `Reconnecting` (was a regression vs trunk).

**Diagnostics**
- One log line in `CompositeConnectionTokenProvider.fetchConnectionToken` printing the active delegate, to make root-causing the BT-vs-remote token routing easier next time.

### Test Steps
1. **POS BT reader basic flow**: connect a BT reader, take a payment, disconnect from the toolbar chip and from Settings → Hardware → Card Reader. All three should work end-to-end.
2. **POS BT reconnect during payment**: start a card payment, briefly toggle Bluetooth/move the reader to force a transient disconnect. Payment should keep going (no "Reader disconnected" cancel).
3. **POS BT reconnect after a stuck Stripe state**: disconnect from a BT reader, immediately try to connect again. Discovery should find the reader (was failing with "Unexpected operation in progress: ReconnectBluetoothReaderOperation").
4. **Remote TTP happy path**: pair tablet ↔ phone, take a payment from the tablet by tapping a (simulated) card on the phone. Tablet should briefly show "Ready for payment" then transition to full-screen "Processing payment", then success.
5. **Remote TTP error UX**: drop the phone connection mid-payment (turn screen off / close Card Reader Mode). Tablet should show a friendly "Connection to the phone reader was lost" message instead of the raw Kotlin exception.
6. **Remote TTP retry after connectivity issue**: trigger a phone connect failure, then dismiss + reopen the connect dialog. Should not auto-reconnect to the same unreachable phone forever.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
@@ -119,6 +119,8 @@ class WooPosCardPaymentViewModel @Inject constructor(
                             _state.value = buildPreparingState()
                             collectPayment()
                         }
+                        // Keep current state during a transient reconnect so an in-flight BT
+                        // payment is not cancelled when the reader briefly drops to Reconnecting.
                         WooPosEffectiveReaderStatus.Reconnecting -> Unit
                         WooPosEffectiveReaderStatus.Connecting,
                         WooPosEffectiveReaderStatus.Disconnected -> {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
@@ -119,8 +119,8 @@ class WooPosCardPaymentViewModel @Inject constructor(
                             _state.value = buildPreparingState()
                             collectPayment()
                         }
+                        WooPosEffectiveReaderStatus.Reconnecting -> Unit
                         WooPosEffectiveReaderStatus.Connecting,
-                        WooPosEffectiveReaderStatus.Reconnecting,
                         WooPosEffectiveReaderStatus.Disconnected -> {
                             _state.value = buildReaderDisconnectedState()
                             cancelPayment()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
@@ -322,9 +322,11 @@ class WooPosCardReaderConnectionController(
                 is CardReaderStatus.Connected -> Unit
                 is CardReaderStatus.Reconnecting ->
                     runCatching { cardReaderManager.cancelReconnection() }
+                        .onFailure { logger.e("startDiscovery(): cancelReconnection() failed - ${it.message}") }
                 is CardReaderStatus.Connecting,
                 is CardReaderStatus.NotConnected ->
                     runCatching { cardReaderManager.disconnectReader() }
+                        .onFailure { logger.e("startDiscovery(): disconnectReader() failed - ${it.message}") }
             }
             unifiedDiscoveryStream
                 .discover(
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
@@ -319,6 +319,7 @@ class WooPosCardReaderConnectionController(
     private fun startDiscovery() {
         discoveryJob?.cancel()
         discoveryJob = scope.launch {
+            runCatching { cardReaderManager.disconnectReader() }
             unifiedDiscoveryStream
                 .discover(
                     isSimulated = developerOptionsRepository.isSimulatedCardReaderEnabled(),
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
@@ -46,6 +46,7 @@ import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 @Suppress("LongParameterList")
 class WooPosCardReaderConnectionController(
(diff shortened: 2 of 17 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/remote/WooPosRemoteReaderPaymentFlow.kt
```diff
@@ -57,15 +63,28 @@ class WooPosRemoteReaderPaymentFlow @Inject constructor(
...
-            CollectPaymentOutcome.TimedOut -> Result.Failed("Timed out waiting for phone reader")
-            is CollectPaymentOutcome.Failed -> Result.Failed(
-                outcome.cause.message ?: "Remote collection failed"
-            )
+            is CollectPaymentOutcome.Success -> {
+                onCaptureStarting()
+                capture(order.id, outcome.paymentIntentId)
+            }
+            is CollectPaymentOutcome.Rejected -> Result.Failed(genericFailureMessage())
+            CollectPaymentOutcome.TimedOut -> Result.Failed(genericFailureMessage())
```
malinajirka: 💡 We might want to cosnider logging the original errors before mapping them?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -309,6 +318,14 @@ class WooPosCardReaderConnectionController(
     private fun startDiscovery() {
         discoveryJob?.cancel()
         discoveryJob = scope.launch {
+            when (cardReaderManager.readerStatus.value) {
+                is CardReaderStatus.Connected -> Unit
+                is CardReaderStatus.Reconnecting ->
+                    runCatching { cardReaderManager.cancelReconnection() }
```
malinajirka, line 324: 💡 These two `runCatching {…}` calls (here and on line 327) have no `onFailure`. If `cancelReconnection()` or `disconnectReader()` throws here (e.g. `IllegalStateException("Terminal not initialized")` from `CardReaderManagerImpl`, or a Stripe SDK error), discovery will silently start in a bad state and the next failure looks like "discovery just doesn't work" with no breadcrumb.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
```diff
@@ -119,8 +119,8 @@ class WooPosCardPaymentViewModel @Inject constructor(
                             _state.value = buildPreparingState()
                             collectPayment()
                         }
+                        WooPosEffectiveReaderStatus.Reconnecting -> Unit
```
malinajirka, line 124: ❓ I would consider adding a comment and/or a test that provides reasoning why we want to handle reconnecting this way. Otherwise, it's quite likely we'll re-introduce the issue in the future.

malinajirka (approved):
Thanks @kidinov!

I noticed that the phone remains in reader phone even if I cancel the reader mode on it or when I leave the reader-mode screen. I can even start multiple reader modes on a single phone by going back and forth and the tablet keeps finding them until I kill the phone app.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/cardreader/remote/WooPosRemoteReaderSessionTest.kt
```diff
@@ -92,6 +97,37 @@
...
+                assertThat(e.message).isEqualTo("cancelled")
+            }
+        }
+
+    @Test
+    fun `given session is not connected, when sendCollectPayment, then returns Failed without calling client`() =
+        runTest {
+            // GIVEN
+            val session = createSession()
+            val paymentInfo: PaymentInfo = mock()
```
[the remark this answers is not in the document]
kidinov: Addressed in 18227ee5e8a — replaced the mock with a real `PaymentInfo` constructor.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/remote/WooPosRemoteReaderPaymentFlow.kt
kidinov: Good call — added error logging before mapping in 7eac3ae1c43.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
kidinov: Added `onFailure` logging to both blocks in 7f00ee853dc.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardpayment/WooPosCardPaymentViewModel.kt
kidinov: Added a short comment explaining the rationale in 5eaa3e1da5d.

kidinov:
@malinajirka 

> I noticed that the phone remains in reader phone even if I cancel the reader mode on it or when I leave the reader-mode screen. I can even start multiple reader modes on a single phone by going back and forth and the tablet keeps finding them until I kill the phone app.

Good catch! Fixed that in e8242ec63bb85c61f5ac770e872c903572173578 and d362e2b747a759cc6e6530d41e3151690f291b77

PR #15753, opened 2026-04-27
[WOOMOB-2887] Create skeleton modules for the AI assistant feature

hichamboushaba:
Fixes WOOMOB-2887

### Description

Sets up the module home for the AI assistant feature. All assistant code will live here; `:WooCommerce` keeps only host integration (entry points, navigation, DI bootstrap).

Two nested modules under `:libs:ai-assistant`:

- **`:feature`** — Android library, Hilt-wired. Will own chat UI, ViewModels, cards, and tool adapters. Carries the new `okhttp-sse` and `kotlinx-serialization-json` dependencies and the `kotlin-serialization` plugin.
- **`:core`** — pure-Kotlin (`org.jetbrains.kotlin.jvm`) skeleton, no Android / Hilt / Compose. Reserved for runtime/domain code (`ChatService`, agentic loop, history budgeter). Empty for now.

Hilt scaffolding (`AiAssistantModule`) provides a `@AiAssistantJson Json` instance with defaults tuned for streaming AI payloads: `ignoreUnknownKeys`, `coerceInputValues`, `encodeDefaults = false`.

**Why kotlinx-serialization:** AI payloads are polymorphic and incremental — sealed hierarchies like `ChatStreamEvent`, `ToolCall`, and `ToolResult` benefit from compile-time exhaustiveness so a missed variant fails the build instead of crashing mid-stream. The plugin is scoped to the assistant modules only; FluxC, `:WooCommerce`, and the rest of the app continue using Gson. `kotlinx-serialization-json` is set to 1.11.0 (latest stable), upgrading the project's existing transitive consumers from 1.7.3 via Gradle conflict resolution.

### Test Steps

Infrastructure-only — no user-facing surface yet. Reviewers can validate via:

1. `./gradlew :libs:ai-assistant:feature:assembleDebug` succeeds.
2. `./gradlew :libs:ai-assistant:core:check` succeeds.
3. `./gradlew :WooCommerce:assembleWasabiDebug` still succeeds.
4. Confirm in the two `build.gradle` files that the dependency direction holds: `:WooCommerce` → `:libs:ai-assistant:feature` → `:libs:ai-assistant:core`, never the reverse.

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15755, opened 2026-04-27
[WOOMOB-2889] Define :ai-assistant:core contracts for chat service

hichamboushaba:
### Description

Part of WOOMOB-2889 (part 1 of 3)

This PR adds the pure-JVM contracts that the AI assistant chat layer is built on. It introduces the `:libs:ai-assistant:core` source set with no Android, Hilt, or transport dependencies — only the types the agentic loop and any `ChatService` adapter must agree on.

This is the first of three stacked PRs for WOOMOB-2889:
- **PR 1 (this one)**: contracts + pinned config + module README
- **PR 2 ([#15757](https://github.com/woocommerce/woocommerce-android/pull/15757))**: `ChatStreamParser` + `WpComJetpackAiTokenProvider`
- **PR 3 ([#15756](https://github.com/woocommerce/woocommerce-android/pull/15756))**: `JetpackAiChatService` + Hilt wiring

All three stack on top of [#15753](https://github.com/woocommerce/woocommerce-android/pull/15753) (WOOMOB-2887, skeleton modules). The split is along the `:core` / `:feature` module boundary so the contract is reviewable independently of the SSE adapter that implements it.

#### What's in this PR

- **`AssistantConfig`** — pinned triple (model id, prompt version, tool-catalog version) plus the `feature` billing tag. Bumping any of these is release-blocking and gated on the regression suite (WOOMOB-2922). Collapsing the triple into one object makes a version bump reviewable as a single diff and makes telemetry (`completionStack`) consistent across call sites.
- **`ChatService`** — single seam between the agentic loop and the completion backend. Concrete adapters own JWT auth, transport, and JSON shape; the loop only sees `Flow<AssistantEvent>`.
- **`AssistantEvent`** sealed hierarchy — `TextDelta`, `ToolCallDelta`, `Finish` (with `FinishReason`), `Failed` (with `AssistantErrorKind`). Transport-level details (HTTP status, SSE framing) never escape the adapter.
- **`AssistantMessage`** sealed hierarchy — `System`, `User`, `Assistant`, `Tool`. Pure-Kotlin so the loop can append turns without depending on transport-level JSON shapes. Mapping to the on-the-wire DTO (OpenAI-style for now, anything OpenAI-compatible later) is the adapter's job, captured forward-looking in WOOMOB-2890.
- **`ChatRequest`** — the loop's input to a turn: messages + optional tools. Deliberately minimal (no `model`/`feature`/`temperature`/`stream` overrides) — Android's posture is to pin the completion stack in code and bump it via release-gated edits to `AssistantConfig`, not per-call config.
- **`ToolDefinition`** — name, description, and JSON-schema parameters block, kept transport-agnostic.
- **`JwtTokenProvider`** + **`AssistantAuthException`** — the auth port that adapters depend on (the WP.com implementation lands in PR 2).
- **Module README** — documents the intended module layout and ownership boundaries for the assistant module family.

#### Why pure JVM

`:core` has zero Android / Hilt / OkHttp dependencies. That keeps the contract reviewable in isolation, lets its tests run as fast pure-JVM tests, and prevents transport details from leaking into the seam the loop consumes.

### Test Steps

1. Pull the branch and confirm `:libs:ai-assistant:core:test` passes locally.
2. Skim `AssistantConfig.kt` and confirm the pinned triple matches what telemetry should report (`completionStack`).
3. Confirm `:core` has no `android.*`, `dagger.*`, or `okhttp3.*` imports anywhere — this is a hard property of the module, not just a convention.

### Images/gif

N/A — no UI changes. The feature is behind the `AI_ASSISTANT` local flag added in a prior PR.

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantErrorKind.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantErrorKind.kt
@@ -1,0 +1,22 @@
+package com.woocommerce.android.aiassistant.core.chat
+
+/**
+ * Closed vocabulary for transport-level errors surfaced by [ChatService].
+ * Transport-specific exceptions (IOException, HTTP status codes, JSON parse
+ * failures) are normalized to one of these values at the service boundary so
+ * the loop and UI speak a single error vocabulary.
+ *
+ * The full assistant error model (loop / tool / safety failures) lives
+ * elsewhere; this enum only covers the failure modes a chat stream can
+ * produce.
+ */
+enum class AssistantErrorKind {
+    NETWORK,
+    TIMEOUT,
+    AUTH,
+    RATE_LIMIT,
+    UPSTREAM_FAILURE,
+    INVALID_STREAM,
+    CANCELLED,
+    UNKNOWN,
+}
--- a/config/gradle/jacoco.gradle
+++ b/config/gradle/jacoco.gradle
@@ -19,6 +19,7 @@ rootProject.afterEvaluate {
                 ':libs:apifaker:testDebugUnitTest',
                 ':libs:commons:testDebugUnitTest',
                 ':libs:pos:testDebugUnitTest',
+                ':libs:ai-assistant:core:test',
                 ':libs:ai-assistant:feature:testDebugUnitTest',
         )
 
--- a/libs/ai-assistant/README.md
+++ b/libs/ai-assistant/README.md
@@ -1,0 +1,48 @@
+# `:libs:ai-assistant`
+
+Assistant feature modules for WooCommerce Android. All assistant implementation lives here;
+`:WooCommerce` owns only host integration (entry point, navigation, DI bootstrap).
+
+## Module layout
+
+```
+:WooCommerce  ──depends on──►  :libs:ai-assistant:feature  ──depends on──►  :libs:ai-assistant:core
+```
+
+### `:libs:ai-assistant:feature`
+
+Android library. The implementation home for the assistant:
+
+- Chat UI and screen-level state (Compose)
+- ViewModel and agentic loop orchestration
+- `ChatService` implementation (`JetpackAiChatService` over OkHttp SSE)
+- JWT token provider (WP.com Jetpack AI)
+- Tool adapters and assistant-specific repository/wiring
+- Result cards and assistant UI components
+- Telemetry hooks
+- Navigation targets or host callbacks emitted outward to `:WooCommerce`
+
+### `:libs:ai-assistant:core`
+
+Pure Kotlin/JVM library. The contract layer consumed by `:feature`:
+
+- `ChatService` and completion-facing interfaces
+- `AssistantEvent`, `AssistantMessage`, `ChatRequest`, `AssistantErrorKind`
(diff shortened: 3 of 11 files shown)


hichamboushaba:
@staskus reminder about this PR 🙂. No pressure, just I saw that you are reviewing later PRs, while this is the first PR of the stack 😄

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantErrorKind.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Closed vocabulary for transport-level errors surfaced by [ChatService].
+ * Transport-specific exceptions (IOException, HTTP status codes, JSON parse
+ * failures) are normalized to one of these values at the service boundary so
+ * the loop and UI speak a single error vocabulary.
+ *
+ * The full assistant error model (loop / tool / safety failures) lives
+ * elsewhere; this enum only covers the failure modes a chat stream can
+ * produce.
+ */
+enum class AssistantErrorKind {
```
staskus, line 13: Not sure if we need to unify 100%, just pointing the difference with iOS:

```
 - OUTCOME_UNKNOWN — the write may have committed but the response was lost. Without this, the eventual write-tool layer cannot safely retry mutations without risking double-writes (double-charges, double-orders). This is the single most important
  addition.
  - INVALID_TOOL_CALL — model emitted a malformed tool call; recoverable via re-prompt.
  - TOOL_FAILED — tool ran but returned a logical failure (4xx, validation error). Distinct from transport failure.
  - REJECTED_BY_SAFETY — model refused. UI messaging differs from UPSTREAM_FAILURE.
  - ```

staskus (approved):
@hichamboushaba you're right to remind! I reviewed them both at the same time but didn't conclude this one.

The contract looks good to me. This one is a close match to iOS 👍

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantErrorKind.kt
hichamboushaba: Thanks @staskus, noted, I'll see if I need to update this in https://github.com/woocommerce/woocommerce-android/pull/15762, as the other PR is what would return the errors we actually expose to UI.

staskus: Yes, I see! I didn't immediately understand that Android distinguishes between two error types!

staskus: But yes, it looks more or less identical to iOS.

hichamboushaba: > `REJECTED_BY_SAFETY`

@staskus we are completely aligned with https://github.com/woocommerce/woocommerce-android/pull/15762, the only difference is `REJECTED_BY_SAFETY`, let's defer discussing it until you add it to iOS.

PR #15762, opened 2026-04-28
[WOOMOB-2891] Define AssistantError sealed class and normalized error model

hichamboushaba:
Fixes WOOMOB-2891

### Description

The agentic loop, retry policy, and any UI that consumes the AI assistant need a single, finite error vocabulary. Until now the codebase had only `AssistantErrorKind` — a stream-level enum scoped to transport errors surfaced by `ChatService`. This PR introduces the loop-level counterpart: `AssistantError`.

`AssistantError` is a sealed class with 10 variants covering the full error surface:

| Variant | Meaning |
|---|---|
| `Network` | Connectivity / IO failure |
| `Auth` | JWT acquisition or HTTP 401/403 |
| `RateLimit` | HTTP 429 |
| `Timeout` | Request timed out (SocketTimeoutException or HTTP 408) |
| `UpstreamFailure` | HTTP 5xx from the backend |
| `ToolFailed(toolName, cause?)` | Tool execution threw an exception |
| `InvalidToolCall(toolName)` | Model emitted an unparseable or unknown tool call |
| `OutcomeUnknown` | Write tool was dispatched but no response was ever received |
| `Cancelled` | Coroutine was cancelled |
| `Unknown(cause?)` | Catch-all for unclassified failures |

`OutcomeUnknown` is deliberately separate from `Network` and `Timeout`. It means a write tool was sent but the response was lost, so the server may or may not have applied it — that ambiguity requires different retry and UX treatment than a clean transport failure.

Two top-level mapping helpers normalise raw inputs into `AssistantError`:

- **`Throwable.toAssistantError()`** — maps `AssistantAuthException`, `SocketTimeoutException`, `UnknownHostException`, `ConnectException`, `IOException`, and `CancellationException`. Branch order in the `when` expression is load-bearing (`SocketTimeoutException` must precede its `IOException` supertype); this is documented in the KDoc and guarded by a dedicated test.
- **`assistantErrorFromHttpCode(code: Int)`** — maps 401/403 → `Auth`, 408 → `Timeout`, 429 → `RateLimit`, 500–599 → `UpstreamFailure`, anything else → `Unknown`.

Both live in `libs/ai-assistant/core` (pure Kotlin/JVM). `AssistantErrorKind` is intentionally unchanged — it stays scoped to `AssistantEvent.Failed` for stream events. The two types coexist by design: the agentic loop (WOOMOB-2893) will use these helpers at the service boundary so callers above it never observe raw exception types.

> [!IMPORTANT]
>  **Note on apparent duplication with `JetpackAiChatService`:** reviewers familiar with [#15756](https://github.com/woocommerce/woocommerce-android/pull/15756) will notice that `JetpackAiChatService` already contains similar inline exception-to-`AssistantErrorKind` mapping (`toAssistantException` / `mapError`). That mapping operates at the stream layer — it converts OkHttp/SSE failures into `AssistantEvent.Failed` events. The helpers introduced here operate at the loop layer — they convert those same raw inputs into `AssistantError`, the type the loop and UI will consume. The mapping rules are intentionally parallel (same exception → same semantic outcome), but the output types are different because the two layers serve different consumers. The inline mapping in `JetpackAiChatService` will be cleaned up or delegated to these helpers once the agentic loop (WOOMOB-2893) is in place and wires everything together.

22 unit tests written TDD (full branch coverage on both helpers).

### Test Steps

This is a pure library change in `libs/ai-assistant/core` with no user-visible surface yet.

1. Run `./gradlew :libs:ai-assistant:core:test` — 22 tests should pass.
2. Open `AssistantError.kt` and confirm it has exactly 10 sealed subclasses.
3. Open `AssistantErrorKind.kt` and confirm it is unchanged (stream-level enum stays as-is).

### Images/gif

N/A

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
@@ -1,0 +1,25 @@
+package com.woocommerce.android.aiassistant.core.chat
+
+/**
+ * Loop-level normalized error vocabulary surfaced to the agentic loop and any
+ * UI layer that consumes the AI assistant. Service-boundary code maps raw
+ * exceptions and HTTP status codes into one of these variants so callers never
+ * branch on transport types.
+ *
+ * [OutcomeUnknown] is intentionally separate from [Network] and [Timeout]: it
+ * means a write tool was dispatched but the response was lost, so the server
+ * may or may not have applied it. That ambiguity drives different retry and
+ * UX decisions than a clean transport failure.
+ */
+sealed class AssistantError {
+    data object Network : AssistantError()
+    data object Auth : AssistantError()
+    data object RateLimit : AssistantError()
+    data object Timeout : AssistantError()
+    data object UpstreamFailure : AssistantError()
+    data class ToolFailed(val toolName: String, val cause: Throwable? = null) : AssistantError()
+    data class InvalidToolCall(val toolName: String) : AssistantError()
+    data object OutcomeUnknown : AssistantError()
+    data object Cancelled : AssistantError()
+    data class Unknown(val cause: Throwable? = null) : AssistantError()
+}
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
@@ -19,7 +19,7 @@ sealed class AssistantError {
     data object UpstreamFailure : AssistantError()
     data class ToolFailed(val toolName: String, val cause: Throwable? = null) : AssistantError()
     data class InvalidToolCall(val toolName: String) : AssistantError()
-    data object OutcomeUnknown : AssistantError()
+    data class OutcomeUnknown(val toolName: String) : AssistantError()
     data object Cancelled : AssistantError()
     data class Unknown(val cause: Throwable? = null) : AssistantError()
 }
--- a/config/gradle/jacoco.gradle
+++ b/config/gradle/jacoco.gradle
@@ -19,7 +19,7 @@ rootProject.afterEvaluate {
                 ':libs:apifaker:testDebugUnitTest',
                 ':libs:commons:testDebugUnitTest',
                 ':libs:pos:testDebugUnitTest',
-                ':libs:ai-assistant:core:test',
+                ':libs:ai-assistant:core:jacocoTestReport',
                 ':libs:ai-assistant:feature:testDebugUnitTest',
         )
 
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/auth/JwtTokenProvider.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/auth/JwtTokenProvider.kt
@@ -16,9 +16,9 @@ interface JwtTokenProvider {
 
 /**
  * Raised by [JwtTokenProvider] implementations when a JWT cannot be obtained.
- * [com.woocommerce.android.aiassistant.core.chat.ChatService] maps this to
- * [com.woocommerce.android.aiassistant.core.chat.AssistantErrorKind.AUTH] before
- * surfacing it to the loop.
+ * The chat service maps this to [com.woocommerce.android.aiassistant.core.chat.AssistantErrorKind.AUTH]
+ * at the stream layer; [com.woocommerce.android.aiassistant.core.chat.toAssistantError] maps it to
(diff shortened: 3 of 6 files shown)


libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+sealed class AssistantError {
+    data object Network : AssistantError()
+    data object Auth : AssistantError()
+    data object RateLimit : AssistantError()
+    data object Timeout : AssistantError()
+    data object UpstreamFailure : AssistantError()
+    data class ToolFailed(val toolName: String, val cause: Throwable? = null) : AssistantError()
+    data class InvalidToolCall(val toolName: String) : AssistantError()
+    data object OutcomeUnknown : AssistantError()
```
staskus: Since:

> [OutcomeUnknown] <...> means a write tool was dispatched but the response was lost, so the server

We should put toolName here as well.

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
```diff
@@ -0,0 +1,25 @@
...
+ * UI layer that consumes the AI assistant. Service-boundary code maps raw
+ * exceptions and HTTP status codes into one of these variants so callers never
+ * branch on transport types.
+ *
+ * [OutcomeUnknown] is intentionally separate from [Network] and [Timeout]: it
+ * means a write tool was dispatched but the response was lost, so the server
+ * may or may not have applied it. That ambiguity drives different retry and
+ * UX decisions than a clean transport failure.
+ */
+sealed class AssistantError {
```
staskus, line 14: From Claude (I double-checked and it looks to be the pattern):

AssistantError.kt:15-19 — these data objects force the mapper to drop the originating Throwable. AssistantEvent.Failed in this same module already keeps it (KDoc: "preserved for telemetry but never inspected by UI code"), and Unknown(cause?) / ToolFailed(cause?) follow the same pattern here. Once the loop reports to Sentry, every Network/Auth/Timeout collapses into one bucket with no stack trace.

Could they carry the cause too?
```
data class Network(val cause: Throwable? = null) : AssistantError()
data class Auth(val cause: Throwable? = null) : AssistantError()
data class RateLimit(val cause: Throwable? = null) : AssistantError()
data class Timeout(val cause: Throwable? = null) : AssistantError()
data class UpstreamFailure(val cause: Throwable? = null) : AssistantError()
```
Mapper: is SocketTimeoutException -> AssistantError.Timeout(cause = this) etc. The null default keeps assistantErrorFromHttpCode and existing tests untouched.

staskus (approved):
👍 LGTM, left 2 comments

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
hichamboushaba: Fixed — `OutcomeUnknown` is now `data class OutcomeUnknown(val toolName: String)`. Added a test for it too.

staskus: Thanks!

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/AssistantError.kt
hichamboushaba: Thanks @staskus 

Thought about this, and I'd push back:

**1. The `AssistantEvent.Failed` where the KDoc you mentioned exists.** That type sits at the stream boundary, its `cause` is meant to be consumed by the loop for telemetry *before* it maps to `AssistantError`. By the time we're holding an `AssistantError`, the throwable has already been logged/breadcrumbed one step upstream. Duplicating `cause` into the post-mapping vocabulary won't be needed.

**2. Equality semantics break.** `AssistantError` is a classification, used for `==` comparisons in tests, `when` exhaustiveness, and UI branching. Switching the five transport variants to `data class(cause)` means `Network(exA) != Network(exB)` even though they mean the same thing to every consumer.

`Unknown(cause)` and `ToolFailed(cause)` are the exceptions: `Unknown` carries cause because classification *failed* (the throwable is the only remaining diagnostic signal), and `ToolFailed` is already a discriminated record carrying `toolName`. The transport kinds are pure classifications where cause would be redundant noise.

Until a real need for carrying more data is present, I'll keep it like this for now.

staskus: Thanks for the thorough answer!

PR #15766, opened 2026-04-28
Step 14: QR login — send device metadata on exchange

JorgeMucientes:
**Do not merge - Parent needs to be merged first.**

### Description

Adds device metadata to the QR login exchange request.

The app now sends a `device` object with the QR login token so WooCommerce Core can show a clearer device name for the Application Password and related sign-in UI. The payload includes:

- `os`
- `os_version`
- `model`
- `brand`
- `app_version`

The values come from `QrLoginDeviceInfoProvider`, which keeps the Android `Build` and `BuildConfig` lookups injectable and testable.

Server-side counterpart: [woocommerce/woocommerce#64456](https://github.com/woocommerce/woocommerce/pull/64456). 

### Expected Behavior

- QR login still works as before.
- On supported WooCommerce Core builds, the issued Application Password and related wp-admin UI can identify the signing-in device.
- On older WooCommerce Core builds, the extra metadata is ignored and the login still succeeds.

### Test Steps

1. Build and install the app: `./gradlew :WooCommerce:installWasabiDebug`.
2. Use a merchant site running the WooCommerce Core counterpart.
3. Generate a QR login code in wp-admin.
4. Scan the QR code with the Android app and complete sign-in.
5. In wp-admin, open `Users > Profile > Application Passwords`.
6. Confirm the new Application Password uses the device-aware name/details when supported by the server.
7. Optional: inspect the QR login request and confirm the JSON body contains `device` alongside the token.
8. Regression check: repeat on an older WooCommerce Core build and confirm QR login still succeeds.

### Stack

Builds on #15772. This is step 14 in the QR login stack.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginDeviceInfoProvider.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginDeviceInfoProvider.kt
@@ -1,0 +1,50 @@
+package com.woocommerce.android.network.qrlogin
+
+import android.os.Build
+import com.google.gson.annotations.SerializedName
+import com.woocommerce.android.BuildConfig
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Wraps the Android `Build` and `BuildConfig` lookups behind an injectable seam so the
+ * QR exchange request can include device metadata without making `QrLoginRestClient`
+ * untestable.
+ *
+ * Production reads `Build.MODEL` / `Build.BRAND` / `Build.VERSION.RELEASE` / `BuildConfig.VERSION_NAME`;
+ * tests pass a fake instance with deterministic values.
+ *
+ * The fields here are mirrored on the server side under the `device` whitelist in
+ * `MobileAppQRLogin::DEVICE_PAYLOAD_KEYS` (WooCommerce Core). Each field is length-capped
+ * server-side, so we don't need to truncate locally.
+ */
+@Singleton
+open class QrLoginDeviceInfoProvider @Inject constructor() {
+    open fun get(): QrLoginDeviceInfo = QrLoginDeviceInfo(
+        os = OS_NAME,
+        osVersion = Build.VERSION.RELEASE.orEmpty(),
+        model = Build.MODEL.orEmpty(),
+        brand = Build.BRAND.orEmpty(),
+        appVersion = BuildConfig.VERSION_NAME.orEmpty(),
+    )
+
+    private companion object {
+        const val OS_NAME = "Android"
+    }
+}
+
+/**
+ * JSON-serializable device metadata sent on the QR login exchange request.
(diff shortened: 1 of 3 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginDeviceInfoProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Production reads `Build.MODEL` / `Build.BRAND` / `Build.VERSION.RELEASE` / `BuildConfig.VERSION_NAME`;
+ * tests pass a fake instance with deterministic values.
+ *
+ * The fields here are mirrored on the server side under the `device` whitelist in
+ * `MobileAppQRLogin::DEVICE_PAYLOAD_KEYS` (WooCommerce Core). Each field is length-capped
+ * server-side, so we don't need to truncate locally.
+ */
+@Singleton
+open class QrLoginDeviceInfoProvider @Inject constructor() {
```
malinajirka: 💡 The class and `get()` are `open` purely so the test can subclass them — but the provider abstraction itself is already the seam, so we don't really need to weaken the production class to make `Build`/`BuildConfig` mockable. We can just `mock<QrLoginDeviceInfoProvider> { on { get() } doReturn QrLoginDeviceInfo(...) }` in the test and drop the `open`s. Wdyt?

```suggestion
@Singleton
class QrLoginDeviceInfoProvider @Inject constructor() {
    fun get(): QrLoginDeviceInfo = QrLoginDeviceInfo(
```

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginDeviceInfoProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+ * untestable.
+ *
+ * Production reads `Build.MODEL` / `Build.BRAND` / `Build.VERSION.RELEASE` / `BuildConfig.VERSION_NAME`;
+ * tests pass a fake instance with deterministic values.
+ *
+ * The fields here are mirrored on the server side under the `device` whitelist in
+ * `MobileAppQRLogin::DEVICE_PAYLOAD_KEYS` (WooCommerce Core). Each field is length-capped
+ * server-side, so we don't need to truncate locally.
+ */
+@Singleton
```
malinajirka: 💡 I think a singleton is an overkill here - the class is stateless so there is no reason to keep it in memory forever. I'd suggest dropping it since the class is cheap to create or at least replacing it with `@Reusable`.

malinajirka (approved):
Thanks @JorgeMucientes! I've left a couple suggestions, but they are not blocking.

JorgeMucientes:
Hey @malinajirka thanks for the feedback! Applied your suggestions. Merging!

PR #15764, opened 2026-04-28
[WOOMOB-2893] Implement AgenticLoop with iteration cap, tool-call assembly, and retry policy

hichamboushaba:
### Description

Fixes WOOMOB-2893

Implements the client-side bounded agentic loop on top of the existing `ChatService` / `AssistantMessage` contracts. This is the runtime that drives the message → tool → message cycle for one user turn.

**What this branch adds**

`:libs:ai-assistant:core` (pure-JVM, no Android / Hilt / `javax.inject`):

- `AgenticLoop` / `LoopEvent` / `LoopOutcome` / `SessionContext` — loop contract
- `AgenticLoopImpl` — the bounded loop: streams from `ChatService`, assembles tool-call deltas, checks tool safety, executes safe tools via `ToolRegistry`, retries transient stream failures under `RetryPolicy`; capped at 5 iterations per turn
- `ToolCallAssembler` — groups streamed `ToolCallDelta` fragments by index, concatenates `argumentsDelta`, parses as `JsonObject`
- `RetryPolicy` / `ConservativeRetryPolicy` — retry decision model with `RetryNow` / `ShowManualRetry` / `DoNotRetry`
- `NoOpToolRegistry` — safe default for hosts with no tools registered
- Tool catalog primitives needed by the loop: `ToolRegistry`, `ToolDescriptor`, `ToolResult`, `ToolSafetyLevel`

`:libs:ai-assistant:feature`:

- `AiAssistantModule` — Hilt `@Provides` bindings that wire the `:core` plain classes (`AgenticLoopImpl`, `NoOpToolRegistry`, `ConservativeRetryPolicy`) into the DI graph

The loop implementation lives in `:core` so it can be exercised in headless JVM unit tests via plain constructor injection. `:feature` owns all Hilt wiring.

**Key design decisions aligned with cross-platform research**

- `LoopOutcome` stays at exactly 4 values (`COMPLETED`, `FAILED`, `MAX_ITERATIONS`, `STOPPED`) per the cross-platform alignment doc
- `LoopEvent.Finished` carries `retryAvailable: Boolean` so the caller can distinguish a recoverable failure (`ShowManualRetry`) from a terminal one (`DoNotRetry`) without adding a fifth outcome
- Tool safety is two-level: `SAFE` / `UNSAFE` (no `DESTRUCTIVE` tier in v1 per cross-platform decision)
- On stream failure, any partial text already streamed to the consumer is preserved in `updatedHistory`
[...]

PR #15767, opened 2026-04-28
[WOOMOB-2894] Add tool registry and catalog selector wiring

hichamboushaba:
### Description
Fixes WOOMOB-2894

This stacks on top of #15764 and wires the assistant tool catalog so conversations can use a pinned per-scope snapshot instead of re-reading the full registry on every turn.

It adds the catalog contracts and feature-side wiring needed for tool scoping:
- introduces `ToolScope`, `CatalogSnapshot`, and `ToolCatalogSelector`
- updates `SessionContext` and `AgenticLoopImpl` to consume the pinned catalog snapshot for tool definitions and lookup
- adds `WooCommerceToolRegistry` (Set-based, driven by handlers) and `DefaultToolCatalogSelector` for GLOBAL, ORDERS, PRODUCTS, and ANALYTICS scopes
- adds `AssistantToolHandler` as an interface with a `descriptor` property, so each handler carries its own `ToolDescriptor` alongside its `execute()` logic
- adds a `StubToolHandler` abstract base class and 13 concrete stub handler classes (one per tool), each declaring its descriptor and returning a "not yet implemented" error until the real data-layer call is wired
- wires all 13 handlers into `AiAssistantModule` via Hilt `@Binds @IntoSet` multibinding

**Tool implementations are not part of this PR.** Each stub handler carries the stable descriptor (name, description, safety level, field constraints) for its tool and returns a `ValidationError` from `execute()`. The actual data-layer implementations (REST calls, FluxC store access) will be added in follow-up tickets — each one replaces the stub's `execute()` body. Because the registry is entirely driven by the injected `Set<AssistantToolHandler>`, adding a real handler requires no registry or DI changes.

The design also makes phantom-tool exposure structurally impossible: `descriptors()` is derived from the injected handler set, so a tool cannot appear in the catalog without a corresponding handler already wired in DI.

### Test Steps
1. Review `AssistantToolHandler` and confirm it is now a regular interface with a `val descriptor: ToolDescriptor` property.
2. Review `WooCommerceToolRegistry` and confirm it takes `Set<AssistantToolHandler>`, derives descriptors from handlers (no hardcoded CATALOG), and guards against duplicate tool names at construction time.
3. Review `DefaultToolCatalogSelector` and confirm each scope returns the expected subset.
4. Review the `handlers/` package and confirm all 13 stub handlers are present with correct tool names, safety levels, and write-constraint descriptions.
5. Review `AiAssistantModule` and confirm the `@Binds @IntoSet` bindings wire all 13 handlers.
6. Confirm `AgenticLoopImpl` reads tool descriptors from `SessionContext.catalogSnapshot.tools`.
7. Run `WooCommerceToolCatalogTest` and confirm all 3 invariant tests pass (13 names, SAFE/UNSAFE classification, write-constraint text).

### Images/gif
N/A

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -167,7 +167,9 @@ class AgenticLoopImpl(
                 else -> toolRegistry.execute(call)
             }
             toolResults += result
-            emit(LoopEvent.ToolCallFinished(result))
+            if (result !is ToolResult.RejectedBySafety) {
+                emit(LoopEvent.ToolCallFinished(result))
+            }
         }
         return toolResults
     }
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -32,7 +32,7 @@ class AgenticLoopImpl(
         history: List<AssistantMessage>,
         context: SessionContext,
     ): Flow<LoopEvent> = flow {
-        val toolDescriptors = toolRegistry.descriptors()
+        val toolDescriptors = context.catalogSnapshot.tools
         val toolDefs = toolDescriptors.map { it.toToolDefinition() }
         val assembler = ToolCallAssembler(json)
 
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -23,6 +23,7 @@ class AgenticLoopImpl(
     private val chatService: ChatService,
     private val toolRegistry: ToolRegistry,
     private val retryPolicy: RetryPolicy,
+    private val historyBudgeter: HistoryBudgeter,
     private val json: Json,
 ) : AgenticLoop {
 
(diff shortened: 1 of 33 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/ToolCatalogSelector.kt
```diff
@@ -0,0 +1,7 @@
+package com.woocommerce.android.aiassistant.core.loop
+
+import com.woocommerce.android.aiassistant.core.chat.ToolDescriptor
+
+fun interface ToolCatalogSelector {
+    fun select(scope: ToolScope, fullRegistry: List<ToolDescriptor>): CatalogSnapshot
```
hichamboushaba, line 6: I don't know whether we discussed this, but this is a part that Claude suggested when I was researching and mentioned the various entry points to the chat, the idea is that when opening the assistant from specific screens, we might want to reduce the number of tools for that context.

I'm adding it now, and we can decide later if it's a good idea or not.

staskus: Thanks for the context. Yes, this is the most interesting part I got interested in before I saw your comment.

I will leave it out of scope for now for iOS. I could see how entering the chat from orders the user may still have questions on products and analytics 🤔

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
```diff
@@ -167,7 +167,9 @@ class AgenticLoopImpl(
                 else -> toolRegistry.execute(call)
             }
             toolResults += result
-            emit(LoopEvent.ToolCallFinished(result))
+            if (result !is ToolResult.RejectedBySafety) {
```
staskus, line 182: 👍

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/ProductVariationsListToolHandler.kt
```diff
@@ -0,0 +1,15 @@
+package com.woocommerce.android.aiassistant.tools.handlers
+
+import com.woocommerce.android.aiassistant.core.chat.ToolDescriptor
+import com.woocommerce.android.aiassistant.core.chat.ToolSafetyLevel
+import kotlinx.serialization.json.buildJsonObject
+import javax.inject.Inject
+
+class ProductVariationsListToolHandler @Inject constructor() : StubToolHandler() {
+    override val descriptor = ToolDescriptor(
+        name = "product_variations_list",
```
staskus, line 10: product_variations_update is the one missing from https://github.com/hichamboushaba/woomobile-ai-assistant-research/blob/main/android/docs/research/13-tool-catalog-context-management.md

staskus (approved):
All the changes make sense 👍

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/ProductVariationsListToolHandler.kt
hichamboushaba: Good eye, I trusted the AI here, but I see it missed one, I'll make sure to add it when I work on the handlers implementations.

libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/ToolCatalogSelector.kt
hichamboushaba: > I will leave it out of scope for now for iOS. I could see how entering the chat from orders the user may still have questions on products and analytics 🤔

Yes, that's my thought too, but depending on the number of tools, we might want to do some filtering. For now with 13 tools only, it's not needed, I agree.

PR #15769, opened 2026-04-28
[WOOMOB-2895] Implement sliding-window HistoryBudgeter

hichamboushaba:
Fixes WOOMOB-2895

### Description

Implements the `HistoryBudgeter` interface and its production implementation, `SlidingWindowHistoryBudgeter`, and wires it into `AgenticLoopImpl` and the Hilt DI graph.

**Why this is needed:**
Without a history budget, `AgenticLoopImpl` resends the entire conversation transcript to the model on every turn. As sessions grow, this inflates token usage and risks hitting context limits. The agreed v1 policy (session-only sliding window) trims the model's view to the most recent N messages while leaving the UI transcript untouched.

**What changed:**

*New types (`:libs:ai-assistant:core`):*
- `CardRef(entityType: String, entityId: Long)` — a lightweight reference to an active-thread entity (order, product, customer) for follow-up context. `retainedEntityRefs` carries these through the budget seam so a future ticket can inject them into the model's prompt for follow-ups like "open the second one".
- `BudgetedHistory` — the budgeter's output: a trimmed message list for the model, plus `retainedEntityRefs`.
- `HistoryBudgeter` — a `fun interface` with a single `build(systemPrompt, rawTranscript, currentUserTurn)` method. Separating the three roles makes the contract unambiguous: the implementer decides which historical messages to include; the caller always provides the full transcript and gets the full transcript back.
- `SlidingWindowHistoryBudgeter` — the production implementation: `rawTranscript.takeLast(windowSize)`, with a configurable window (default 20 messages).

*Modified:*
- `AgenticLoopImpl` — injects `HistoryBudgeter`. On each `runTurn` call it splits the caller's `history` into the system prompt and raw transcript, calls the budgeter, and tracks two separate lists: `modelMessages` (budgeted, sent to `ChatRequest`) and `newTurnMessages` (messages generated this turn). `LoopEvent.Finished.updatedHistory` is always `history + newTurnMessages`, so the caller's full transcript is preserved regardless of how aggressively the budgeter trims.
- `AiAssistantModule` — provides `SlidingWindowHistoryBudgeter` as `HistoryBudgeter` and passes it into `AgenticLoopImpl`.

**Stacks on:** [WOOMOB-2894 PR #15767](https://github.com/woocommerce/woocommerce-android/pull/15767)

### Test Steps

This PR adds no user-visible behaviour — the budgeting is internal to the AI assistant runtime and the assistant UI is not yet connected to the runtime in a releasable build. Verification is through the unit tests added in this PR:

1. Run `:libs:ai-assistant:core:test` — all `SlidingWindowHistoryBudgeterTest` and `AgenticLoopImplTest` cases should pass.
2. In particular, `given budgeter trims history, when running turn, then model receives only budgeted messages` verifies that the model's `ChatRequest` contains only the windowed messages, and `given budgeter trims history, when running turn, then updatedHistory contains full prior history plus new turn messages` verifies that the full transcript is returned to the caller.

### Images/gif

N/A

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/CardRef.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/CardRef.kt
@@ -1,0 +1,6 @@
+package com.woocommerce.android.aiassistant.core.chat
+
+data class CardRef(
+    val entityType: String,
+    val entityId: String,
+)
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/CardRef.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/chat/CardRef.kt
@@ -2,5 +2,5 @@ package com.woocommerce.android.aiassistant.core.chat
 
 data class CardRef(
     val entityType: String,
-    val entityId: String,
+    val entityId: Long,
 )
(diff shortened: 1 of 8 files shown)


staskus (approved):
✅ Looks good and clear.

Something I haven't tackled yet on iOS side but I should, I'll create the task to mirror the behavior.

After i1, we may need to research to see if sliding window is what makes the most sense. Maybe using a "compact" strategy from the coding agents would also make sense or some hybrid strategy. I personally don't like when the agent suddenly "forgets" what we talked a few messages ago.

PR #15772, opened 2026-04-29
[WOOMOB-2806] Step 13: QR login — warn before replacing an active session

malinajirka:
### Description

Fixes WOOMOB-2806

When a user is already signed in and a QR-login deep link arrives (Ticket / wp.com magic link / site-URL prefill), the scanner now interrupts the hand-off with a fullscreen warning that "you're already signed in, continuing will sign you out." On confirm, `AccountRepository.logout()` runs (covering both wp.com OAuth and per-site app-password sessions, plus prefs / Zendesk / DataStore / selected-site cleanup) and the original payload action resumes from a clean state. On cancel, the existing session is left intact and the activity exits.

The warning copy is intentionally generic — the user-facing wp.com vs. app-password distinction isn't surfaced anywhere else in the QR flow, and the canonical logout handles both shapes uniformly. In practice the warning is only reachable via deep link (the in-app QR scanner sits behind a logged-out `LoginActivity`), but the `AccountRepository.isUserLoggedIn()` gate in the ViewModel means we're correct in either case.

The work is split into 6 atomic commits — refactor → scaffolding (events, strings, screen) → feature wiring → BackHandler polish — each one focused and individually reviewable.

### Test Steps

1. Sign in to a wp.com account.
2. Trigger a Ticket QR deep link via `adb shell am start -a android.intent.action.VIEW -d "woocommerce://qr-login?ticket=…&siteUrl=…"`. Expect the warning → tap **Continue and sign out** → existing session is logged out → **Sign in to this store?** confirmation appears → tap Continue → app authenticates with the new credentials.
3. With a wp.com session still active, trigger a wp.com magic-link QR. Expect warning → Continue → logout → browser opens for magic-link sign-in.
4. Sign in via app-password to site A; trigger a site-URL-only QR for site B. Expect warning → Continue → app-password for site A is cleared (check logcat for `Application password deleted for site Id…`) → land on the prefilled site-address login screen.
5. Reach the warning, tap **Cancel** (or system back). Expect to return to wherever the deep link interrupted, with the existing session intact, and `LOGIN_QR_SESSION_REPLACE_DISMISSED` fired.
6. Sign **out**. Trigger any of the three QR types — the warning should not appear (regression check).

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
@@ -6,11 +6,13 @@ import com.woocommerce.android.analytics.AnalyticsEvent
 import com.woocommerce.android.analytics.AnalyticsTracker
 import com.woocommerce.android.analytics.AnalyticsTrackerWrapper
 import com.woocommerce.android.network.qrlogin.QrLoginExchangeException
+import com.woocommerce.android.ui.login.AccountRepository
 import com.woocommerce.android.ui.login.WPApiSiteRepository.CookieNonceAuthenticationException
 import com.woocommerce.android.ui.login.qrlogin.QrLoginScannerViewModel.UiState.Authenticating
 import com.woocommerce.android.ui.login.qrlogin.QrLoginScannerViewModel.UiState.Confirming
 import com.woocommerce.android.ui.login.qrlogin.QrLoginScannerViewModel.UiState.Error
 import com.woocommerce.android.ui.login.qrlogin.QrLoginScannerViewModel.UiState.Idle
+import com.woocommerce.android.ui.login.qrlogin.QrLoginScannerViewModel.UiState.WarningSessionReplace
 import com.woocommerce.android.ui.orders.creation.CodeScannerStatus
 import com.woocommerce.android.util.WooLog
 import com.woocommerce.android.viewmodel.MultiLiveEvent.Event
@@ -42,6 +44,7 @@ class QrLoginScannerViewModel @Inject constructor(
     savedState: SavedStateHandle,
     private val parser: QrLoginPayloadParser,
     private val authenticator: QrLoginAuthenticator,
+    private val accountRepository: AccountRepository,
     private val analyticsTracker: AnalyticsTrackerWrapper
 ) : ScopedViewModel(savedState) {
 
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
```diff
@@ -176,6 +208,32 @@ class QrLoginScannerViewModel @Inject constructor(
...
+     * We flip into [Authenticating] for the duration of the logout so the UI gives feedback
+     * while [AccountRepository.logout] talks to the server — for the Ticket path this is then
+     * overwritten by [Confirming] so the merchant still sees the host-confirmation step.
+     */
+    fun onConfirmSessionReplace() {
+        val pending = (_uiState.value as? WarningSessionReplace)?.pending ?: return
+        analyticsTracker.track(AnalyticsEvent.LOGIN_QR_SESSION_REPLACE_CONFIRMED)
+        _uiState.value = Authenticating
+        launch {
+            accountRepository.logout()
```
JorgeMucientes: Edge case handling: perhaps we should not `resumePending(pending)` unless logout is successful (returns true) ? What I'm not sure is how to handle the failure, whether we should add a retry or something. 
In any case I don't think this is a blocker.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -258,6 +258,10 @@
     <string name="login_qr_scanner_error_user_role_body">Your account doesn\'t have permission to use the app for this store.</string>
     <string name="login_qr_error_primary_retry">Try again</string>
     <string name="login_qr_error_primary_scan">Scan a new code</string>
+    <string name="login_qr_session_replace_title">You\'re already signed in</string>
+    <string name="login_qr_session_replace_body">Continuing will sign you out of your current session and start a new sign-in. Anything you haven\'t saved may be lost.</string>
```
JorgeMucientes: NP. This part `Anything you haven\'t saved may be lost.` sounds a bit weird. I think the only possible "lost" state is if this happens in the middle of product or order creation which I think would be very corner case. Imho I'd remove it, as it may lead to thinking you can actually loose some store data. 
But just a suggestion feel free to ignore it.

JorgeMucientes (approved):
Well done! 
Works as expected and code looks good. Just left a minor suggestion about handling logout failures with retry. Non-blocking though. Lmk your thoughts :)

WooCommerce/src/main/res/values/strings.xml
malinajirka: Good call, updated in cfa8eed80877d80d30c318e7ffa18aebfd580c5c

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
malinajirka: Fixed in b57ad7ac1d73efd5a1b1a987f34790efe297daae

PR #15777, opened 2026-04-29
[RSM-1625] Add new orders notification settings detail UI

irfano:
### Description
Fixes RSM-1625

Adds the New orders notification settings detail screen behind the smarter notifications flow:
- lets the New orders row open the native detail screen
- adds local UI state for the enable switch and order filter selection
- shows a store-formatted threshold input when high-value orders is selected
- reuses the settings section header style and shares full-format currency input cleaning through `WCCurrencyUtils`

### Test Steps
1. Enable the `smarter_notifications` feature flag for a Woo-driven store.
2. Open Settings > Notifications.
3. Tap New orders.
4. Confirm the New orders screen opens with the enable switch and order filter options.
5. Select Only high-value orders and confirm the threshold input appears with the store currency symbol and defaults to `$100`.
6. Turn the enable switch off and confirm the order filter options are disabled.

### Images/gif
image


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+        data object AllOrders : NotificationPreference
+
+        data class HighValueOrders(
+            val thresholdAmount: BigDecimal = BigDecimal(DEFAULT_THRESHOLD_AMOUNT)
+        ) : NotificationPreference
+    }
+
+    companion object {
+        private const val DEFAULT_DECIMAL_NUMBER = 2
+        private const val DEFAULT_THRESHOLD_AMOUNT = 100
```
irfano, line 52: I used **$100** as the default threshold, but I’m considering fetching all product prices and using the most expensive product price as the default instead. That said, the current **$100** threshold is simple and looks good enough.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/EnableNotificationsCard.kt
```diff
@@ -0,0 +1,60 @@
...
+        Row(
+            modifier = Modifier.padding(horizontal = 16.dp, vertical = 20.dp),
+            horizontalArrangement = Arrangement.spacedBy(16.dp),
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            Column(modifier = Modifier.weight(1f)) {
+                Text(
+                    text = title,
+                    style = MaterialTheme.typography.titleMedium,
+                    color = colorResource(id = R.color.color_on_surface_high)
```
AdamGrzybkowski: I think it would be more future-proof to use `MaterialTheme.colorScheme` - this way, when we introduce a new design system, it will be easier to transition. Any reasons for not using it? 

This title could use `onSurface` and the body below `onSurfaceVariant`. That would match the semantics from Material 3.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/EnableNotificationsCard.kt
```diff
@@ -0,0 +1,60 @@
...
+    description: String,
+    isEnabled: Boolean,
+    onEnabledChanged: (Boolean) -> Unit
+) {
+    Surface(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp, vertical = 16.dp),
+        shape = RoundedCornerShape(8.dp),
+        border = BorderStroke(1.dp, colorResource(id = R.color.woo_gray_5)),
```
AdamGrzybkowski: This could be replaced with `MaterialTheme.colorScheme.outlineVariant` (default divider color in material 3) or `MaterialTheme.colorScheme.outline` - a slightly lighter version, which is probably closer to woo_gray_5.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/NotificationPreferenceOption.kt
```diff
@@ -0,0 +1,67 @@
...
+    ) {
+        RadioButton(
+            selected = selected,
+            onClick = onClick,
+            enabled = enabled,
+            colors = RadioButtonDefaults.colors(
+                selectedColor = MaterialTheme.colorScheme.primary,
+                unselectedColor = colorResource(id = R.color.color_on_surface_medium),
+                disabledSelectedColor = colorResource(id = R.color.color_on_surface_medium).copy(alpha = 0.38f),
+                disabledUnselectedColor = colorResource(id = R.color.color_on_surface_medium).copy(alpha = 0.38f)
```
AdamGrzybkowski: Same comment about colors here - let's try and use Material 3 theming system.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/NotificationPreferenceOption.kt
```diff
@@ -0,0 +1,67 @@
...
+            modifier = Modifier
+                .weight(1f)
+                .padding(top = 4.dp, start = 8.dp)
+        ) {
+            Text(
+                text = title,
+                style = MaterialTheme.typography.titleMedium,
+                color = colorResource(id = R.color.color_on_surface_high).let {
+                    if (!enabled) it.copy(alpha = 0.38f) else it
+                }
```
AdamGrzybkowski: Here and in all the other places as well - I will skip adding comments for all those places.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val _viewState = MutableStateFlow(
+        ViewState(
+            currencySymbol = currencyParameters.currencySymbol.orEmpty(),
+            currencyPosition = currencyFormattingParameters?.currencyPosition ?: LEFT,
+            currencyDecimalSeparator = currencyFormattingParameters?.currencyDecimalSeparator ?: ".",
+            currencyThousandSeparator = currencyFormattingParameters?.currencyThousandSeparator ?: ",",
+            currencyDecimalNumber = currencyFormattingParameters?.currencyDecimalNumber ?: DEFAULT_DECIMAL_NUMBER
+        )
+    )
+    val viewState = _viewState.asStateFlow()
```
AdamGrzybkowski: I believe we should be exposing LiveData as the public field to be consistent across the code base.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
```diff
@@ -0,0 +1,248 @@
...
+                        }
+                    )
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun ThresholdAmountField(
```
AdamGrzybkowski, line 100: ❓ Curious, but do we really need decimal values here? Do you think anyone would want 100.50 over 100? 

Assuming an empty field, to achieve 200, I have to type 20000 - extra to taps, because of the two decimal places.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+        val notificationsEnabled: Boolean = true,
+        val notificationPreference: NotificationPreference = NotificationPreference.AllOrders,
+        val currencySymbol: String,
+        val currencyPosition: CurrencyPosition,
+        val currencyDecimalSeparator: String,
+        val currencyThousandSeparator: String,
+        val currencyDecimalNumber: Int
+    )
+
+    sealed interface NotificationPreference {
```
AdamGrzybkowski: Having a sealed class means that the user will lose the custom amount if they accidentally tap on the `All new orders`. If we kept the type and the value separately, we could restore whatever the user has entered.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
AdamGrzybkowski: I think this is ok for now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/EnableNotificationsCard.kt
irfano: Done: b16099981b4e2303bdd28772bbd411ab32986b95

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/EnableNotificationsCard.kt
irfano: Done: b16099981b4e2303bdd28772bbd411ab32986b95

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/NotificationPreferenceOption.kt
irfano: Done: b16099981b4e2303bdd28772bbd411ab32986b95

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/compose/NotificationPreferenceOption.kt
irfano: Updated this and the other `R.color` usages as well: b16099981b4e2303bdd28772bbd411ab32986b95

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Good catch! Done: 578d7eb384a7194cbe8a18c73b805b2e1da6a621

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
irfano: I used the same view in the Custom Amount screen, but your point is good. Still, I’m not sure about removing decimal support. It may be useful for some currencies and some stores.

I changed the design and used the Amount field that we already use in “Fixed Cart Discount.”



I updated the screen using the same field (a0cfed566948e88bf601a5a37449164fb80ba991). How does it look now?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Agreed, separating the threshold amount into its own property is cleaner. Updated in c08570ce5662339ccd58d4cef718e926ab2dc8dc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
AdamGrzybkowski: I like it!

AdamGrzybkowski (approved):
Looks good, thanks for the changes! 

One small thing: When I delete the min value threshold, the value defaults to 0. The server won't accept this, so maybe defaulting to 1 would be better, or back to the default 100.

irfano:
> One small thing: When I delete the min value threshold, the value defaults to 0. The server won't accept this, so maybe defaulting to 1 would be better, or back to the default 100.

I used 1 as the minimum threshold: f3abe4db121d478bdf1185e900db432f27b1aec7

PR #15775, opened 2026-04-29
[WOOMOB-2899][WOOMOB-2900] Add orders_update, products_list, and products_get AI tools

hichamboushaba:
Fixes WOOMOB-2899
Fixes WOOMOB-2900

**Depends on:** #15774 (WOOMOB-2898, `orders_list` / `orders_get` infrastructure). This PR targets that branch — merge order is WOOMOB-2898 first, then this one.

### Description

Implements two AI assistant tool groups on top of the `AssistantToolHandler` infrastructure introduced in #15774.

**WOOMOB-2899 — `orders_update` (write, UNSAFE)**

Adds a status-only update tool that lets the AI assistant advance an order through a restricted set of transitions: `on-hold → processing` and `processing → completed`. Other transitions (cancellation, refund) are deliberately excluded.

- `AIOrdersDataSource.updateOrderStatus` wraps `WCOrderStore.updateOrderStatus`, which emits a `Flow<UpdateOrderResult>`. `.last()` is used to collect the final result: if the order exists in Room two events are emitted (optimistic + remote) and we take the remote one; if the order is not in Room only the optimistic error event is emitted and we surface that as a failure.
- `OrdersUpdateToolHandler` fetches the current order status, validates the requested transition against an allowlist map, then calls `updateOrderStatus`. Returns `ValidationError` with the current and target statuses in the message for disallowed transitions; `TransportError(retryable=true)` for network failures.
- Classified `UNSAFE` so the agentic loop requires explicit user confirmation before execution.

**WOOMOB-2900 — `products_list` + `products_get` (read, SAFE)**

Adds two read-only product tools backed by a new `AIProductsDataSource` that wraps `WCProductStore`.

`products_list`:
- Optional `search` param (free-text, forwarded to `WCProductStore.searchProductsByNameAndSku`); when absent, calls `WCProductStore.fetchProducts`.
- Blank/whitespace search is normalised to `null` so the plain-fetch path is used.
- Optional `page` + `per_page` (clamped 1–50, default 20); offset computed as `(page - 1) * perPage`.
- Response: `{ count, ids, status_counts }` — compact enough for the model to reason over without blowing the context window.

`products_get`:
- Required `id` param; cache-first via `WCProductStore.getProductByRemoteId`, falls back to `fetchSingleProduct` on cache miss.
- Response: 12-field product detail object with all pricing, stock, and status fields the model is likely to need.

[...]

PR #15778, opened 2026-04-29
[WOOMOB-2901, WOOMOB-2902] Add AI product variation and update tools

hichamboushaba:
Fixes WOOMOB-2901
Fixes WOOMOB-2902

**Depends on:** #15775 (WOOMOB-2899 / WOOMOB-2900, product tool data-source and read-tool infrastructure). Merge order is #15775 first, then this one.

### Description

Implements the remaining product AI assistant tools on top of the `AssistantToolHandler` and `AIProductsDataSource` patterns introduced in #15775.

**WOOMOB-2901 — `product_variations_list` (read, SAFE)**

Replaces the previous variation-list stub with a real handler while intentionally keeping the existing `product_variations_list` tool name.

The tool supports two modes:
- List mode: required `product_id`, optional `page` and `per_page` (default 20, clamped 1-50). Returns `{ product_id, count, ids, stock_status_counts, price_range }`.
- Detail mode: required `product_id` plus optional `variation_id`. Returns variation detail fields needed for stock and pricing questions: status, SKU, regular/sale price, on-sale flag, stock management, stock quantity, and stock status.

A new `AIVariationsDataSource` wraps `WCProductStore.fetchProductVariations`, `fetchSingleVariation`, and the local variation cache helpers. `ProductsGetToolHandler` now points variable-product follow-up questions to `product_variations_list` for either list or single-variation lookup.

**WOOMOB-2902 — `products_update` (write, UNSAFE)**

Replaces the previous `products_update` stub with a real write handler under `tools/products/`.

The update schema is deliberately narrow:
- Required: `id`
- Optional: `name`, `regular_price`, `sale_price`, `stock_quantity`, `status`
- Allowed statuses: `draft`, `pending`, `private`, `publish`

The handler validates that at least one optional field is provided and keeps the tool classified as `UNSAFE`, so the agentic loop requires merchant confirmation before execution.

`AIProductsDataSource.updateProduct` fetches the existing product, applies only the provided fields, and sends the updated copy through `WCProductStore.batchUpdateProducts`. Setting `stock_quantity` also enables `manageStock` automatically, matching the FluxC update contract. The tool is intentionally limited to simple products; variable parent products should be updated through individual variations instead.

Both tools are wired through Hilt `@Binds @IntoSet` multibinding, and the old stubs in `tools/handlers/` are removed.

### Test Steps

These tools are backend infrastructure with no dedicated UI in this PR. To verify through the assistant flow:

1. Build and install a debug APK.
2. Open the AI Assistant chat on a store with at least one variable product.
3. Ask the assistant to list variations for the variable product ID — confirm it returns variation IDs, stock status counts, and a price range.
4. Ask for one specific variation by product ID and variation ID — confirm it returns pricing, SKU, status, and stock fields.
5. Ask the assistant to update a simple product field such as name, price, stock quantity, or status — confirm it requests confirmation before executing the unsafe write.
6. Try to update a variable parent product through `products_update` — confirm the assistant directs the update through individual variations instead.

### Images/gif

N/A — no UI changes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/products/ProductsUpdateToolHandler.kt
```diff
@@ -0,0 +1,117 @@
+package com.woocommerce.android.aiassistant.tools.products
+
```
staskus, line 2: There are these little differences the cross-referencing AI catches:

1. product_variations_list schema asymmetry vs iOS. Android adds an optional variation_id field to make this tool dual-mode (list when absent, detail when present). iOS keeps it list-only. Same tool name, different schemas → cross-platform prompt design gets harder. If iOS plans a future product_variations_get (single-variation lookup), Android should rename its detail mode into that sibling tool now to match.

2. Variable-product update rejection is broader on Android. Android's AIProductsDataSource.updateProduct rejects any non-simple-product update via UnsupportedProductTypeException (any field). iOS only refuses when prices are in the body — name/status/stock_quantity updates on a variable parent succeed on iOS. Merchants asking the AI to rename or unpublish a variable parent will succeed on iOS, fail on Android. Either tighten iOS or loosen Android, then keep the descriptions consistent.

3. products_update output shape differs. iOS returns a full ProductSummary (canonical post-update state). Android returns a sparse object with only the fields the caller wrote, keyed by product_id instead of id. The model on Android has to call products_get again to re-confirm state after every update — different shape per platform, harder shared prompt design.

----

Quite hard to align now on every single detail. After we finish, we could have a sync session where we list all of these differences and iron what we want in one go.

staskus (approved):

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/products/ProductsUpdateToolHandler.kt
hichamboushaba: Good points, I'll address the point 3 here.

For the rest, I think each approach has its own merits, so as you said, we can align later.

hichamboushaba: Updated `products_update` to return the canonical product detail response shape, including `id`, prices, stock, status, and other post-update fields, instead of the sparse changed-fields payload with `product_id`. I kept the other two differences as-is for this PR because Android intentionally keeps `product_variations_list` dual-mode here, and `products_update` is scoped to simple products only; those remaining cross-platform differences are better handled in the follow-up alignment discussion you suggested.

PR #15781, opened 2026-04-30
Remove REMOTE_TAP_TO_PAY feature flag

kidinov:
### Description

Remove the `REMOTE_TAP_TO_PAY` feature flag and its gates so the Woo POS remote tap-to-pay flow ships always-on with the rest of `feature/remote-tap-to-pay`.

- Drop the enum entry from `FeatureFlag`
- Inline the flag as always-true in:
  - `PaymentsHubViewModel` — phone-eligibility check now only depends on TTP availability, so the **Card Reader Mode** row shows whenever TTP is available
  - `WooPosUnifiedDiscoveryStream` — NSD branch always runs alongside Bluetooth discovery
  - `WooPosCardReaderConnectionController` + `Factory` — removed the now-redundant `isRemoteTapToPayEnabled` field
- Simplify `WooPosCardReaderConnectionState.Scanning` to a `data object` (the `isRemoteTapToPaySupported` flag is always true now); the dialog shows the Wi-Fi hint strip on non-phone breakpoints
- Update tests and remove the FF-specific contract test
- Replace the 24.7 internal flag note in `RELEASE-NOTES.txt` with a user-facing 24.8 entry for the feature

### Test Steps

1. Open **Settings → Payments** on a TTP-eligible store on the phone — the **Card Reader Mode** row appears.
2. Tap **Card Reader Mode** — the Card Reader Mode screen launches.
3. On a tablet, open Woo POS, start a card reader connection — both Bluetooth readers and discovered phones (over Wi-Fi LAN) appear in the scanning list, and the hint strip is visible.

### Images/gif

N/A


samiuelson (approved):
LGTM

samiuelson:
Adding "do not merge" label to make sure it doesn't land in milestone 24.7.

malinajirka:
Seems like this is safe to merge now.

PR #15783, opened 2026-04-30
Fix remote-phone discovery state mismatch and ambiguous auto-reconnect

kidinov:
### Description

Two related bugs found while reviewing the remote tap-to-pay branch. Both have to do with how the tablet identifies a discovered phone reader.

**1. Discovery add/remove key mismatch.** `Added` events keyed phones in the in-memory map by display name (`Build.MODEL`, e.g. `"Pixel 8"`), but `Removed` events emit the NSD service name (`"woopos-remote-a3f4"`). The names never matched, so phones that went offline stayed in the discovered list forever. Fixed by adding a `serviceName` field to `WooPosDiscoveredReader.Phone` and keying the discovery state by it.

**2. Auto-reconnect could pair to the wrong same-model phone.** `setLastConnectedPhoneName(phone.name)` persists `Build.MODEL`, which is not unique across same-model devices. With two same-model phones on the LAN, auto-reconnect would silently pair to whichever NSD resolved first. Auto-reconnect is preserved — but `findLastKnownPhone` now returns null when more than one discovered phone matches the persisted name, so the user picks manually in that case. A proper fix (per-install device UUID broadcast in NSD TXT) is left as a follow-up.

Each fix is in its own commit.

### Steps to reproduce (before this PR)

**Bug 1 — stale phone never removed**
1. On the tablet, open Woo POS → start a card-reader connection.
2. On a paired phone, open **Card Reader Mode** so it appears in the tablet's list.
3. Close **Card Reader Mode** on the phone (or kill the app).
4. **Expected:** the phone disappears from the tablet's list within a few seconds.
   **Before:** the phone stays in the list until the tablet's discovery is restarted.

**Bug 2 — auto-pair to wrong phone of the same model**
Probably not possible to test that without code changes

The persisted "last connected phone" name only survives a successful connect — an explicit disconnect from the tablet clears it. So the bug is reachable when the tablet did NOT explicitly disconnect (phone went offline, tablet app was killed, network dropped):

1. Two phones of the same model on the same Wi-Fi (e.g. two Pixel 8s — call them **A** and **B**).
2. On the tablet: open Woo POS → connection screen → pair to **A** → take a payment so the connection succeeds (this persists `LastConnectedPhoneName = "Pixel 8"`).
3. Do **not** disconnect from the tablet UI. Instead, close **Card Reader Mode** on **A** (or kill the app on A so its NSD broadcast stops).
4. Start **Card Reader Mode** on **B**. **A** stays off.
5. On the tablet: leave the connection screen and come back (or kill and re-open the POS app).
6. **Expected:** the tablet should not silently auto-connect to a different physical device than the one previously paired.
   **Before:** the tablet auto-connects to **B** because the persisted `\"Pixel 8\"` matches its display name.
   **After:** if both **A** and **B** are visible, the tablet skips auto-connect and shows them in the list for the user to pick. If only **B** is visible, auto-connect still fires (single match) — accepting that this corner case still mispairs; the proper fix needs a stable per-device ID, tracked separately.

### Test Steps

1. Repeat Bug 1 reproduction — the phone now disappears from the list shortly after Card Reader Mode is closed.
2. Bug 2 — collision case: with two same-model phones both running Card Reader Mode, after pairing to one and putting it offline (without disconnecting), open the connection screen with both phones visible — the tablet does not auto-connect; both phones appear in the list for the user to pick.
3. Bug 2 — happy path: with a single phone, pair → take a payment → close Card Reader Mode without disconnecting → re-open it → tablet still auto-connects to it.
4. Smoke: a normal connect → pay → disconnect cycle with one phone still works.
5. Smoke: Bluetooth reader auto-reconnect (different code path) still works.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
@@ -300,7 +300,7 @@ class WooPosCardReaderConnectionController(
         withContext(dispatchers.io) {
             logger.d("disconnect(): clearing prefs")
             appPrefsWrapper.removeLastConnectedCardReaderId()
-            appPrefsWrapper.removeLastConnectedPhoneName()
+            appPrefsWrapper.removeLastConnectedPhoneDeviceId()
             cardReaderTrackingInfoKeeper.setTransport(null)
 
             logger.d("disconnect(): stopping remote session")
@@ -454,21 +454,8 @@ class WooPosCardReaderConnectionController(
     }
 
     private fun findLastKnownPhone(phones: List<WooPosDiscoveredReader.Phone>): WooPosDiscoveredReader.Phone? {
-        val lastConnectedName = appPrefsWrapper.getLastConnectedPhoneName() ?: return null
-        // The persisted identifier is Build.MODEL, which is not unique across same-model devices.
-        // If multiple discovered phones match, refuse to auto-connect and let the user pick — silently
-        // pairing to the wrong physical device is worse UX than asking once.
-        val matches = phones.filter { it.name == lastConnectedName }
-        return when (matches.size) {
-            0 -> null
-            1 -> matches.first()
-            else -> {
-                logger.d(
-                    "Multiple discovered phones match last-connected name '$lastConnectedName'; skipping auto-connect"
-                )
-                null
-            }
-        }
+        val lastDeviceId = appPrefsWrapper.getLastConnectedPhoneDeviceId() ?: return null
+        return phones.find { it.deviceId == lastDeviceId }
     }
 
     private fun continueSearching() {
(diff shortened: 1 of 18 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -455,7 +455,20 @@ class WooPosCardReaderConnectionController(
 
     private fun findLastKnownPhone(phones: List<WooPosDiscoveredReader.Phone>): WooPosDiscoveredReader.Phone? {
         val lastConnectedName = appPrefsWrapper.getLastConnectedPhoneName() ?: return null
-        return phones.find { it.name == lastConnectedName }
+        // The persisted identifier is Build.MODEL, which is not unique across same-model devices.
+        // If multiple discovered phones match, refuse to auto-connect and let the user pick — silently
+        // pairing to the wrong physical device is worse UX than asking once.
+        val matches = phones.filter { it.name == lastConnectedName }
```
malinajirka: 💡The new `when (matches.size)` only protects when both same-model phones happen to be in the *same* discovery snapshot at the moment `handleReadersFound` runs. NSD discovery is incremental, though: if phone A is discovered a few hundred ms before B, `matches.size == 1` and we auto-connect to A even if B is the one that was originally paired. The original mispair just got rarer. 

Ideally, we'd implement the proper fix you suggested right away. However, if that's too complex, I'd at least consider adding some sort of `delayed connection` so we give all the phones lets say at least a second to appear during the discovery phase. Wdyt?

malinajirka (commented):
Thanks @kidinov! I haven't tested the changes yet, but I think I found one issue with the approach. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
kidinov: You are right. Race is real even with `when (matches.size)` - NSD is incremental, phone B can come later and still mispair.

I pushed a fix to this PR (`266af3570de`):

- Phone broadcasts a stable UUID in NSD TXT (`did=...`), persisted in `AppPrefs`
- Tablet matches by `deviceId`, not display name
- The `when (matches.size)` block is gone

Users will re-pair once after this ships.

malinajirka (approved):
Nice, thanks!

PR #15784, opened 2026-04-30
Step 15: QR login — add number-matching approval step

JorgeMucientes:
**Do not merge - Parent needs to be merged first.**

### Description

Fixes WOOMOB-2947.

Adds a number-matching approval step to QR login. After the app scans a QR code, it shows a 3-digit number and waits for the merchant to tap the matching number in wp-admin. The app only exchanges the QR token for credentials after the server reports that the correct number was approved.

This PR also:

- Adds `/qr-login-scan` and `/qr-login-session-status` calls.
- Sends `supports_number_matching` in `/qr-login-scan` so WooCommerce Core can identify number-matching-capable clients during rollout.
- Polls session status immediately, then every 2 seconds while waiting for approval.
- Sends `Cache-Control: no-cache, no-store` on status polling requests.
- Logs fail-closed protocol anomalies for approved statuses without grants and unknown session states.
- Keeps unknown poll failures out of the network-error bucket.
- Moves the final credential exchange orchestration into `QrLoginScannerViewModel`.
- Extracts QR login error mapping into `QrLoginErrorMapper`.
- Shows terminal UI for rejected, expired, invalid-grant, and retryable error states.

This is a protocol change and needs the matching WooCommerce Core implementation before end-to-end testing.

### Expected Behavior

- The phone shows the store host, a 3-digit number, and a countdown after scanning.
- wp-admin shows multiple number options for the same QR session.
- Tapping the matching number signs the merchant in.
- Tapping the wrong number or waiting for expiry shows a terminal error in the app.
- Cancelling on the phone returns to the scanner without completing the login.
- Cached `scanned` responses should not keep the app stuck after approval.

### Test Steps

1. Build and install the app: `./gradlew :WooCommerce:installWasabiDebug`.
2. Use a merchant site running the WooCommerce Core number-matching counterpart.
3. Open the QR login screen in wp-admin and render a QR code.
4. In the Android app, open QR login and scan the QR code.
5. Confirm the phone shows a 3-digit number and countdown.
6. In wp-admin, tap the matching number.
7. Confirm the app advances to signing in and lands in the store.
8. Repeat the flow and tap a wrong number in wp-admin. Confirm the app shows the rejected/denied state.
9. Repeat the flow and let the session expire. Confirm the app shows the timeout state.
10. Repeat the flow and tap Cancel on the phone. Confirm the app returns to the scanner.
11. Optional: inspect network traffic and confirm session-status requests include `Cache-Control: no-cache, no-store`.

### Stack

Builds on #15766. This is step 15 in the QR login stack.

Release coordination: Android with this PR must ship before the matching WooCommerce Core protocol change reaches merchant sites.

### Images/gif

To be added before moving out of draft.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -168,11 +168,10 @@ class QrLoginRestClient @Inject constructor(
         return Request.Builder()
             .url(statusUrl)
             .get()
-            // Force-bust any intermediary HTTP cache (OkHttp's shared cache, edge proxy,
-            // CDN). Polling responses are state-bearing — pinning the first response and
-            // serving it forever means the app sees `scanned` indefinitely even after the
-            // merchant approves on wc-admin.
-            .cacheControl(CacheControl.FORCE_NETWORK)
+            // Force-bust any intermediary HTTP cache (OkHttp's shared cache, edge proxy, CDN).
+            // Polling responses are state-bearing, so pinning the first response means the app
+            // sees `scanned` indefinitely even after the merchant approves on wc-admin.
+            .cacheControl(POLL_CACHE_CONTROL)
             .build()
     }
 
@@ -359,6 +358,10 @@ class QrLoginRestClient @Inject constructor(
         const val HTTP_BAD_REQUEST = 400
         const val HTTP_CONFLICT = 409
         const val HTTP_UPGRADE_REQUIRED = 426
+        val POLL_CACHE_CONTROL = CacheControl.Builder()
+            .noCache()
+            .noStore()
+            .build()
         val JSON_MEDIA_TYPE = "application/json; charset=utf-8".toMediaType()
     }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -7,6 +7,7 @@ import com.woocommerce.android.util.CoroutineDispatchers
 import com.woocommerce.android.util.WooLog
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.withContext
+import okhttp3.CacheControl
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.OkHttpClient
@@ -167,6 +168,11 @@ class QrLoginRestClient @Inject constructor(
         return Request.Builder()
             .url(statusUrl)
             .get()
+            // Force-bust any intermediary HTTP cache (OkHttp's shared cache, edge proxy,
+            // CDN). Polling responses are state-bearing — pinning the first response and
+            // serving it forever means the app sees `scanned` indefinitely even after the
+            // merchant approves on wc-admin.
+            .cacheControl(CacheControl.FORCE_NETWORK)
             .build()
     }
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -318,7 +318,13 @@ class QrLoginRestClient @Inject constructor(
             exchangeGrant
                 ?.takeIf { it.isNotBlank() }
                 ?.let { QrLoginSessionStatus.Approved(it) }
-                ?: QrLoginSessionStatus.Expired
+                ?: run {
+                    WooLog.w(
+                        WooLog.T.LOGIN,
+                        "$QR_LOGIN_APPROVED_NO_GRANT_LOG_ID: approved session-status missing exchange_grant"
+                    )
+                    QrLoginSessionStatus.Expired
+                }
     }
 
     private data class ExchangeRequest(
@@ -358,6 +364,7 @@ class QrLoginRestClient @Inject constructor(
         const val HTTP_BAD_REQUEST = 400
         const val HTTP_CONFLICT = 409
         const val HTTP_UPGRADE_REQUIRED = 426
+        const val QR_LOGIN_APPROVED_NO_GRANT_LOG_ID = "QR_LOGIN_APPROVED_NO_GRANT"
         val POLL_CACHE_CONTROL = CacheControl.Builder()
             .noCache()
             .noStore()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -279,6 +279,8 @@ class QrLoginRestClient @Inject constructor(
     private data class ScanRequest(
         val token: String,
         val device: QrLoginDeviceInfo,
+        // Capability flag for the Core number-matching cutover. Servers that require
+        // number matching reject legacy clients missing this value before mutating token state.
         @SerializedName("supports_number_matching") val supportsNumberMatching: Boolean,
     )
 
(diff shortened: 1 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
```diff
@@ -269,14 +429,45 @@ class QrLoginScannerViewModel @Inject constructor(
...
+    }
+
+    private fun Throwable.toPollReason(): ErrorReason = when (this) {
+        QrLoginSessionStatusException.EndpointMissing -> ErrorReason.EndpointMissing
+        QrLoginSessionStatusException.RateLimited -> ErrorReason.RateLimited
+        QrLoginSessionStatusException.Network -> ErrorReason.Network
+        QrLoginSessionStatusException.MalformedResponse -> ErrorReason.ServerError
+        is QrLoginSessionStatusException.HttpError -> ErrorReason.ServerError
+        is QrLoginSessionStatusException.Unknown -> ErrorReason.Unknown
+        else -> ErrorReason.Network
```
malinajirka: 💡 `else -> ErrorReason.Network` makes any unmapped exception (including programming errors, future SDK exceptions, etc.) look like a network blip, should we at least log an error to the log?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -108,9 +272,58 @@ class QrLoginRestClient @Inject constructor(
...
+            "approved" -> approvedOrFailClosed()
+            "rejected" -> QrLoginSessionStatus.Rejected
+            // Treat unknown states defensively as Expired so the UI shows a terminal screen
+            // rather than spinning indefinitely on a state the app doesn't understand.
+            else -> QrLoginSessionStatus.Expired
+        }
+
+        // Approved without a grant shouldn't be possible against a Task-7 server, but if
+        // it ever happens we fail closed rather than continuing to poll forever.
+        private fun approvedOrFailClosed(): QrLoginSessionStatus =
```
malinajirka, line 326: 💡 If the server returns `state=approved` but no `exchange_grant`, that's a real protocol violation, not an expiry. Failing closed in the UI is exactly right, but silently mapping to `Expired` means we'd never notice this from telemetry — we'd just see a phantom uptick in expirations. A `WooLog.w` (or, better, a Sentry-grade `logError` with a dedicated ID like `QR_LOGIN_APPROVED_NO_GRANT`) before returning `Expired` would catch a server regression we'd otherwise miss. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -108,9 +272,58 @@ class QrLoginRestClient @Inject constructor(
...
+        val state: String?,
+        @SerializedName("exchange_grant") val exchangeGrant: String?,
+    ) {
+        fun toStatus(): QrLoginSessionStatus = when (state) {
+            "scanned" -> QrLoginSessionStatus.Scanned
+            "approved" -> approvedOrFailClosed()
+            "rejected" -> QrLoginSessionStatus.Rejected
+            // Treat unknown states defensively as Expired so the UI shows a terminal screen
+            // rather than spinning indefinitely on a state the app doesn't understand.
+            else -> QrLoginSessionStatus.Expired
```
malinajirka: 🔍 Same shape as the `approvedOrFailClosed` comment — any future server state we don't know about (`pending_2fa`, `locked`, …) silently becomes `Expired`. I'd consider adding a log here, potentially even a Sentry log. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -32,37 +41,173 @@ class QrLoginRestClient @Inject constructor(
...
 ) {
 
-    suspend fun exchange(siteUrl: String, token: String): Result<QrLoginCredentials> =
+    suspend fun scan(siteUrl: String, token: String): Result<QrLoginScanResult> =
+        withContext(dispatchers.io) {
+            try {
+                Result.success(performScan(siteUrl, token))
+            } catch (ce: CancellationException) {
+                throw ce
+            } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
```
malinajirka: 📓 This will catch all errors not just exceptions (e.g. OutOfMemoryError) => I'd consider replacing Throwable with Exception so Errors are handled properly. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -32,37 +41,173 @@ class QrLoginRestClient @Inject constructor(
...
+        siteUrl: String,
+        token: String,
+        exchangeGrant: String,
+    ): Result<QrLoginCredentials> =
+        withContext(dispatchers.io) {
+            try {
+                Result.success(performExchange(siteUrl, token, exchangeGrant))
+            } catch (ce: CancellationException) {
+                throw ce
+            } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
```
malinajirka: Same as above.

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -108,9 +272,58 @@ class QrLoginRestClient @Inject constructor(
...
 
-    private data class ExchangeRequest(
+    // endregion
+
+    // region request / response shapes
+
+    private data class ScanRequest(
         val token: String,
         val device: QrLoginDeviceInfo,
+        @SerializedName("supports_number_matching") val supportsNumberMatching: Boolean,
```
malinajirka, line 284: ❓ This is not about this PR but I'm wondering what is this parameter for - AFIAU we didn't release the backend or the app before number matching supports was added, so the parameter feels useless. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
```diff
@@ -147,12 +374,65 @@ data class QrLoginCredentials(
...
+
 sealed class QrLoginExchangeException(message: String) : Exception(message) {
     data object TokenRejected : QrLoginExchangeException("Token was rejected by the site")
     data object EndpointMissing : QrLoginExchangeException("Exchange endpoint not available on the site")
     data object RateLimited : QrLoginExchangeException("Rate limit hit on the exchange endpoint")
     data object Network : QrLoginExchangeException("Network failure during exchange")
     data object MalformedResponse : QrLoginExchangeException("Exchange response was malformed")
+    data object NotApproved :
+        QrLoginExchangeException("Exchange called before merchant approved the number match")
+    data object InvalidExchangeGrant :
```
malinajirka, line 451: I'm wondering if we should tell the user this level of detail - usually telling more makes the code more vulnerable to attack. But maybe it's fine here, I'm not sure. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
```diff
@@ -5,20 +5,29 @@ import com.woocommerce.android.OnChangedException
 import com.woocommerce.android.analytics.AnalyticsEvent
```
malinajirka, line 4: 📓 We can do that in a follow up - but this VM is too large now and does too many things. I think we should split it into subcomponents, wdyt?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -228,11 +228,13 @@
...
-    <string name="login_qr_confirm_title">Sign in to this store?</string>
-    <string name="login_qr_confirm_subtitle">Only continue if this is your store.</string>
-    <string name="login_qr_confirm_security_note">You\'re signing in with a one-time code from this store.</string>
-    <string name="login_qr_confirm_connect">Continue</string>
-    <string name="login_qr_confirm_cancel">Cancel</string>
+    <!-- Number-matching approval step -->
+    <string name="login_qr_match_title">Confirm sign-in</string>
+    <string name="login_qr_match_host_label">You\'re signing in to</string>
+    <string name="login_qr_match_subtitle">Tap this number on your computer to finish.</string>
+    <string name="login_qr_match_countdown" comment="%1$d is seconds remaining before the challenge expires.">Expires in %1$ds</string>
```
malinajirka: I'm not familiar with this `comment` attribute, do we truly support it?

malinajirka (approved):
Thanks @JorgeMucientes! Looks good overall, I've left a few non-blocking comments.

PR #15788, opened 2026-04-30
POS settings: add Reader connected indicator and Disconnect button to card reader screen

samiuelson:
### Description

Adds a status row at the bottom of the POS Settings → Hardware → Card Readers connected card containing a green dot + "Reader connected" indicator on the left and a Disconnect button on the right. Inspired by the main app's Menu → Payments card reader detail screen, which exposes a Disconnect action when a reader is connected and consistent with the overflow menu in POS home showing the reader status indicator.

This change is needed to allow disconnecting the reader on phone device where the reader status overflow indicator is not visible.

⚠️ "Do not merge" label - don't merge untill the parent is `trunk`

### Test Steps

1. Have a Woo store with a card reader configured (Stripe M2 or BBPOS).
2. Launch POS, open Settings → Hardware → Card Readers.
3. While **disconnected**: verify the screen shows the existing "Connect your reader" entry; no Disconnect button or green dot.
4. Connect a reader.
5. While **connected**: verify the card shows Device name, Battery (if reported), Firmware, then a row at the bottom containing a green dot + "Reader connected" on the left and a "Disconnect" button on the right.
6. Tap Disconnect → reader disconnects and the screen returns to the disconnected state.
7. Verify on both phone (POS phone layout) and tablet — the bottom row should remain on a single line with the disconnect button visible (not clipped).
8. Verify in light and dark theme — the dot should be green in both (ForestGreen on light, VividGreen on dark).

### Images/gif

image
image


toupper (commented):
Looks good overall — the new "Reader connected" row at the bottom of the card with the dot indicator + Disconnect button is a nice phone-friendly addition, and reusing the existing `woopos_reader_connected` string is the right call.

A few small things worth a look — none blocking:

**Pattern duplication with the floating toolbar.** `WooPosHomeFloatingToolbar.kt` already has private `Circle(size, color)` and `ReaderStatusText(modifier, title)` helpers that compose to exactly the same visual (dot + "Reader connected" text), with the same `Spacer(width=Medium)`, `Spacer(width=XSmall)`, and `padding(horizontal=Small)` rhythm. The PR description even references the toolbar's overflow indicator as inspiration. Worth promoting those two helpers (or a combined indicator) to a shared component so the settings card and the toolbar can't drift visually.

**`14.dp.toAdaptiveComponentSize()` vs the toolbar's bare `14.dp`.** The toolbar uses raw `14.dp` for its identical-looking dot, but here it's wrapped in `toAdaptiveComponentSize()`, which scales it down to ~10.5dp on phone and ~12.6dp on small tablet. Either is defensible, but the inconsistency between the two reader indicators feels accidental rather than designed — happy to defer to whatever you prefer, just flagging.

**Redundant nested `Row` with `verticalAlignment`.** The call site is already a `Row(verticalAlignment = CenterVertically) { ReaderConnectedIndicator(); Spacer.weight(1f); WooPosOutlinedButtonSmall(...) }`, and `ReaderConnectedIndicator()` is itself a `Row(verticalAlignment = CenterVertically) { Spacer; Box; Spacer; Text }`. The leading `Spacer(width=Medium)` inside the indicator is just nudging content right. Could either flatten the indicator into its three composables at the call site (with a `modifier.padding(start = Medium)`) or have the indicator take a `modifier` parameter and apply the start padding there — drops one Row layer.

samiuelson:
Thanks for the review @toupper! Addressed all three points in 04612d4 — none were blocking but they were good catches.

**1. Pattern duplication.** Promoted the toolbar's private `Circle` and `ReaderStatusText` to a shared `WooPosReaderStatusIndicator.kt` in `ui/woopos/common/composeui/component/`, exposing `WooPosReaderIndicatorDot(size, color, modifier)` and `WooPosReaderStatusText(title, modifier)`. Both the toolbar and the settings card now use them. Kept them as separate primitives (rather than a combined indicator) because the toolbar's dot color animates between `success`/`alert` based on reader state while the settings row is fixed `success` — a single combined component would have needed a color param anyway and obscured the toolbar's animation.

**2. `14.dp.toAdaptiveComponentSize()` vs raw `14.dp`.** Dropped the adaptive wrapper in the settings card so both indicators now render the same `14.dp` dot.

**3. Redundant nested `Row`.** Flattened — the call site is now a single `Row(verticalAlignment = CenterVertically)` containing the dot, text, weighted spacer, and the disconnect button. The leading start padding moved onto the dot's `Modifier` instead of an inner `Spacer`. Net: one Row layer dropped.

Mind taking another look?

toupper (approved):
All three points addressed cleanly:

- Shared `WooPosReaderIndicatorDot` + `WooPosReaderStatusText` in their own file, with both the toolbar and settings card now using them — the parallel implementation is gone.
- Bare `14.dp` on both call sites, so the toolbar and settings indicators are visually consistent.
- The nested `Row` is flattened: dot + text live directly inside the outer row at the call site, with the leading offset moved onto the dot's modifier.

LGTM.

PR #15789, opened 2026-04-30
[WOOMOB-2896] Add assistant confirmation safety handoff

hichamboushaba:
### Description

Fixes WOOMOB-2896

**Adds the in-code safety handoff for unsafe AI assistant tool calls and the localized confirmation-preview contract that future UI will render.** Internal-only — the feature is gated behind the `AI_ASSISTANT` local flag and has no merchant-visible surface in this PR.

#### Core safety runtime (`:libs:ai-assistant:core`)

Safety decisions are now enforced in code instead of relying on prompt text:

- safe tools execute immediately
- unsafe tools pause the agentic loop and emit `LoopEvent.ConfirmationRequested`
- the loop awaits `SafetyOrchestrator.confirm(requestId)` / `cancel(requestId)`
- confirmed requests resume normal tool execution
- cancelled requests stop the loop with `AssistantError.Cancelled` / `LoopOutcome.STOPPED`

The core confirmation request stays preview-free and Android-free. It carries only the metadata needed to identify the pending call: confirmation request ID, tool call ID, tool name, raw `JsonObject` arguments, and `ToolSafetyLevel`. This keeps `:core` a pure-Kotlin module — no `Context`, no `R`, no `UiString`, no WooCommerce copy.

The loop also preserves model-history consistency: when a model emits multiple tool calls and a later unsafe call is cancelled, earlier completed calls keep matched assistant tool-call/tool-result pairs, while the cancelled call is not persisted as an unmatched call or a synthetic rejected result.

**Decline behavior diverges from iOS by design.** iOS replays a declined proposal as a `user_cancelled` tool result so the model can produce one more turn. Android ends the turn: the confirmation card is the answer, the unsafe tool is never dispatched, no cancelled call is written to history, and the loop finishes with `LoopOutcome.STOPPED`. This prevents declined writes from being retried or paraphrased. We can revisit if cross-platform parity becomes required.

#### Safety catalog constraints

The catalog stays deliberately narrow this iteration, and tests guard each rule so new unsafe tools cannot be added silently:

- only `SAFE` and `UNSAFE` levels are used (no destructive/deletion tier)
- write and bulk-write tools remain `UNSAFE`
- deletion/destructive tool names are asserted absent
- every unsafe tool must have a dedicated confirmation-preview path

#### Feature-owned localized previews (`:libs:ai-assistant:feature`)

All preview surface lives in `:feature`, where WooCommerce-specific copy belongs:

- `ConfirmationPreview`, `ConfirmationPreviewField`, `ConfirmationPreviewText`
- `ConfirmationPreviewRenderer` — non-Compose, not injectable, callers provide the `Context`. Keeps it testable and avoids baking app-level context into a future UI path. Compose UI can later resolve the same nodes via `LocalContext`.
- `WooCommerceConfirmationPreviewBuilder`
- feature-local `R.string` resources for confirmation messages, field labels, summaries, and values

These types are `internal` so they don't leak to the host before there's a real UI integration point. The preview text model uses Android resources directly and supports raw merchant/API values, string resources with nested preview args, and explicit singular/multiple resource pairs (no `<plurals>` — unsupported by the project's GlotPress flow).

The WooCommerce builder mirrors the existing handler argument contracts:

- single order updates preview supported fields such as `status`
- bulk order updates preview `status`, `customer_note`, `billing_email`
- product updates preview supported fields and ignore unsupported extras (e.g. `stock_status`)
- bulk product updates summarize one or many product IDs
- product variation updates preview `product_id`, variation `id`, and supported variation fields
- numeric IDs are parsed as `Long`, matching handlers; malformed IDs fall back instead of producing misleading copy
- unknown tools use a redacted generic fallback and never expose raw JSON arguments

#### Out of scope

No bottom sheet, no Compose UI, no `:WooCommerce` host integration. Future UI work will consume `LoopEvent.ConfirmationRequested`, build a feature-owned preview from the request, render it, and resolve the request through the same `SafetyOrchestrator`.

### How to verify

Internal infrastructure — no merchant-visible flow yet. Reviewers can validate via:

1. `./gradlew :libs:ai-assistant:core:testDebugUnitTest :libs:ai-assistant:feature:testDebugUnitTest` passes.
2. `:core` has no `android.*`, `dagger.*`, or WooCommerce imports (module remains pure Kotlin).
3. Skim the catalog tests to confirm the safety/preview-coverage invariants.
4. Skim the loop tests for safe-execute, unsafe-pause/resume, cancel-stop, and mixed-history paths.
5. Confirm the diff adds no UI, bottom-sheet, POS, or `:WooCommerce` host code.

### Images/gif

N/A — no UI changes.

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -90,19 +90,7 @@ class AgenticLoopImpl(
 
             val toolExecution = executeTools(assembledResults, validCalls, toolDescriptors)
             if (toolExecution is ToolExecutionOutcome.Cancelled) {
-                val completedCalls = toolExecution.completed.map { it.historyToolCall }
-                val cancelledAssistantMessage = AssistantMessage.Assistant(
-                    content = stream.assistantText.takeIf { it.isNotEmpty() },
-                    toolCalls = completedCalls,
-                )
-                if (cancelledAssistantMessage.content != null || completedCalls.isNotEmpty()) {
-                    newTurnMessages.add(cancelledAssistantMessage)
-                }
-                toolExecution.completed.forEach { completed ->
-                    newTurnMessages.add(completed.toToolMessage())
-                }
-                emit(LoopEvent.Failed(AssistantError.Cancelled))
-                emit(LoopEvent.Finished(LoopOutcome.STOPPED, history + newTurnMessages, retryAvailable = false))
+                emitCancelledToolExecution(toolExecution, stream.assistantText, history, newTurnMessages)
                 return@flow
             }
 
@@ -193,21 +181,10 @@ class AgenticLoopImpl(
         }
         for (call in validCalls) {
             val descriptor = toolDescriptors.find { it.name == call.name }
-            val result = when {
-                descriptor == null ->
-                    ToolResult.ValidationError(call.id, "Unknown tool: ${call.name}")
-                else -> when (val decision = safetyOrchestrator.evaluate(call, descriptor)) {
-                    SafetyDecision.Execute -> executeApprovedTool(call)
-                    is SafetyDecision.RequireConfirmation -> {
-                        emit(LoopEvent.ConfirmationRequested(decision.request))
-                        val confirmationResult = safetyOrchestrator.awaitResult(decision.request.id)
-                        emit(LoopEvent.ConfirmationResolved(confirmationResult))
-                        when (confirmationResult.decision) {
-                            ConfirmationDecision.CONFIRMED -> executeApprovedTool(call)
-                            ConfirmationDecision.CANCELLED -> return ToolExecutionOutcome.Cancelled(completedTools)
-                        }
-                    }
-                }
+            val result = if (descriptor == null) {
+                ToolResult.ValidationError(call.id, "Unknown tool: ${call.name}")
+            } else {
+                executeConfirmedTool(call, descriptor) ?: return ToolExecutionOutcome.Cancelled(completedTools)
             }
             completedTools += CompletedToolCall(call, result)
             emit(LoopEvent.ToolCallFinished(result))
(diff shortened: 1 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
✅ 

> Decline behavior diverges from iOS by design. iOS replays a declined proposal as a user_cancelled tool result so the model can produce one more turn. Android ends the turn: the confirmation card is the answer, the unsafe tool is never dispatched, no cancelled call is written to history, and the loop finishes with LoopOutcome.STOPPED. This prevents declined writes from being retried or paraphrased. We can revisit if cross-platform parity becomes required.

This particular decision wasn't conscious on my part 🤔 I need to test how it feels in the real world. Yes the model likely can paraphrase, acknowledge cancelation, or ask if something else is needed. Needs testing.

**Claude comment, maybe some additional cancelation handling could be added:**

> Pending-request leak on outer flow cancellation. SafetyOrchestratorImpl.awaitResult cleans up via try { deferred.await() } finally { pending.remove(...) }. But if the outer Flow is cancelled (user cancels the turn from the UI) while parked on await, the await never returns and the pending entry stays in the ConcurrentHashMap forever. iOS handles this via cancelAllPending() on stream termination. Fix: add a try { ... } finally { safetyOrchestrator.cancelPending(turnId) } around the loop body, or move to a Channel/select-based design.

hichamboushaba:
> **Claude comment, maybe some additional cancelation handling could be added:**
> 
> > Pending-request leak on outer flow cancellation. SafetyOrchestratorImpl.awaitResult cleans up via try { deferred.await() } finally { pending.remove(...) }. But if the outer Flow is cancelled (user cancels the turn from the UI) while parked on await, the await never returns and the pending entry stays in the ConcurrentHashMap forever. iOS handles this via cancelAllPending() on stream termination. Fix: add a try { ... } finally { safetyOrchestrator.cancelPending(turnId) } around the loop body, or move to a Channel/select-based design.

Thanks @staskus.
I submitted a fix, but for future readers, The leak is not from cancellation while suspended inside `deferred.await()` because `awaitResult()` has a finally and coroutine cancellation should run it. There is still a risk of a cancellation window after `ConfirmationRequested` is emitted and before `awaitResult()` starts. I added an outer try/finally around the confirmation branch and call `safetyOrchestrator. cancelPending(requestId)` as a safety cleanup path. That keeps the current design and matches the intent of iOS’ stream-termination cleanup without introducing a channel/select redesign.

PR #15790, opened 2026-04-30
[WOOMOB-2911, WOOMOB-2912] Add assistant chat UI runtime

hichamboushaba:
### Description
Fixes WOOMOB-2911
Fixes WOOMOB-2912

This PR adds the Android AI Assistant chat foundation. It introduces the Compose chat surface for user and assistant messages, streaming/loading/error/retry states, cancellation, and the pending-confirmation UI needed when the agentic loop requests confirmation for unsafe tool execution.

It also adds the feature-layer `AssistantRuntime` bridge between `AssistantViewModel` and the core `AgenticLoop`. The runtime selects the active tool catalog, registers tools for a request, translates loop events into UI-facing results, and coordinates confirmation handoff with the safety contract from the base branch.

The feature is intentionally not wired into app navigation or a production entry point yet. This keeps WOOMOB-2911 and WOOMOB-2912 scoped to the reusable UI/runtime foundation while leaving entry point integration and final product design iteration for follow-up work.

### Test Steps
1. This feature is not wired to an in-app entry point yet.
2. Validate by code review and green CI.

### Images/gif
Tentative UI done by GPT, we'll iterate on this later when we have a final design.
assistant-chat-screen-wootheme-phone-refined


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantUiState.kt
```diff
@@ -0,0 +1,64 @@
+package com.woocommerce.android.aiassistant.ui
+
+import com.woocommerce.android.aiassistant.core.chat.AssistantError
+import com.woocommerce.android.aiassistant.runtime.AssistantPendingConfirmation
+
+data class AssistantUiState(
+    val messages: List<AssistantUiMessage> = emptyList(),
```
staskus, line 7: Heads up: On iOS we made each chat message a list of pieces (text, tool call, tool result, card, confirmation), and the UI draws a different view for each piece. That way one assistant turn can show text, then a card, then an inline "are you sure?" prompt, all in the same bubble flow.

```swift
public struct ChatMessage {
    public let id: UUID
    public var segments: [MessageSegment]
}
```

Android's AssistantUiMessage is just id + role + text right now, and the screen draws one text bubble per item.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantViewModel.kt
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private fun LoopOutcome.toAssistantUiStatus(): AssistantUiStatus = when (this) {
+        LoopOutcome.COMPLETED,
+        LoopOutcome.STOPPED -> AssistantUiStatus.IDLE
+        LoopOutcome.FAILED,
+        LoopOutcome.MAX_ITERATIONS -> AssistantUiStatus.ERROR
+    }
+
+    private fun AssistantRuntimeEvent.Finished.toAssistantUiError(): AssistantUiError? = when (outcome) {
```
staskus, line 203: Claude flagged:


> This is the only place in the ViewModel where errors are mapped to UI errors. It runs only on Finished — when the loop has fully terminated. So AssistantError.OutcomeUnknown (from AssistantUiState.kt:55 enum) is reachable only at terminal Finished.
> 
> The scenario it misses: model emits orders_update write tool → tool dispatches → loop receives outcomeUnknown (transport drop after the PUT was sent) → the loop's retry policy decides to keep going (because the model's next turn might want to read the order back to verify) → loop continues → eventually finishes COMPLETED with a benign final assistant message. The user never sees that the write may have applied.
> 
> iOS puts outcomeUnknown on a streamingState: .outcomeUnknown("This may have applied; verify before retrying") field that the controller sets immediately on observing the failed event, not just at Finished. Subsequent text deltas can't overwrite that surface — the warning sticks until the user explicitly dismisses or restarts.
> 
> The Android fix would be: detect AssistantError.OutcomeUnknown mid-loop in reduceRuntimeEvent, set it in _uiState.error, and gate the Finished handler from clearing error if it's currently OUTCOME_UNKNOWN.

staskus (approved):
✅ LGTM. Left a comment to check

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantUiState.kt
hichamboushaba: Yes, this will be added later for Android too. For now I just wanted to have a screen quickly for testing the PRs.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantViewModel.kt
hichamboushaba: Good catch, I'll make sure to fix this in next PRs.

PR #15750, opened 2026-04-27
[WOOMOB-2862] Add dashboard order type bottom sheet

JorgeMucientes:
### Description
Fixes WOOMOB-2862

Adds the Performance card order type bottom sheet UI. The order label now shows the selected order date type with a chevron affordance, opens a bottom sheet with Paid orders, All orders, and Completed orders, saves selection changes through the analytics setting API, refreshes dashboard stats after a successful save, and tracks both opening and selection events.

Stacked on the WOOMOB-2860 order date-type API PR.

### Test Steps
Use `wc-admin/analytics` section to ensure and compare data matches with the one we display in the app. Careful with the caching of wc-admin. As sometimes it takes awhile to refresh the order count after changing the `date_type` setting. The easiest way to compare fresh results with `wc-admin` is to change the date range after the `date_type` has been updated in the app. 

1. Launch the app and open the Dashboard.
2. In the Performance card, tap the order type label below the order count.
3. Confirm the bottom sheet shows Paid orders, Placed orders, and Completed orders with descriptions and the current selection.
4. Select a different date type and confirm the sheet closes, the card label updates, and the Performance card refreshes.
5. Confirm the setting is updated in wc-admin/analytcis/settings. 
6. Reopen the sheet and confirm the new selection is shown.

https://github.com/user-attachments/assets/945ba04f-e32b-4d7a-98d4-fefbbfce9ec0

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15791, opened 2026-04-30
Fix Performance card stat label alignment when Jetpack badge is shown

JorgeMucientes:
### Description

Fixes a regression from [#15750](https://github.com/woocommerce/woocommerce-android/pull/15750): the `Paid orders` / `Visitors` / `Conversion` labels in the Performance card no longer line up when the visitors column shows the Jetpack badge.

That PR flipped the stat row's `gravity` from `bottom` to `top`. Because the visitors column is taller in the badge state (Jetpack icon + skeleton stacked above a bottom-anchored label), top alignment leaves its label sitting ~24 dp below the others. Reverting to `gravity="bottom"` restores a shared baseline. There's a small tradeoff, when selecting `completed orders` the label will be 2 lines and thus may seem miss aligned with the rest. However, when Jetpack is not enable visitors and conversion numbers will never be displayed so it doesn't look that off: 

Screenshot 2026-04-30 at 17 45 30

### Test Steps

1. Open a site whose visitor stats are unavailable (Jetpack-disconnected or self-hosted/JCP store) so the Performance card shows the Jetpack badge.
2. On the **My Store** tab, confirm `Paid orders`, `Visitors`, and `Conversion` share the same baseline.
3. Switch to a site with visitor stats and confirm labels still align (regression check).

### Images/gif

| Before | After |
| --- | --- |
| Screenshot 2026-04-30 at 17 11 46 | Screenshot 2026-04-30 at 17 37 32 |

### DIFF
--- a/WooCommerce/src/main/res/layout/dashboard_main_stats_row.xml
+++ b/WooCommerce/src/main/res/layout/dashboard_main_stats_row.xml
@@ -19,7 +19,7 @@
         android:layout_height="wrap_content"
         android:layout_marginTop="@dimen/major_100"
         android:baselineAligned="false"
-        android:gravity="top"
+        android:gravity="bottom"
         android:orientation="horizontal">
 
         <LinearLayout


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

irfano (commented):
I think `center` works better here. Looking at the screenshots without the Jetpack icon, center looks a bit better. Wdyt?

|`bottom`|`center`|
|-|-|
|||
|||

irfano (approved):
I checked the screenshots I shared again. When using `center`, “Conversion” and “Visitors” become less aligned. They’re not too bad, though.

I would still use `center` to make it better for the normal case, since `bottom` works better for the error case. It’s up to you. I’m approving.

JorgeMucientes:
Hey @irfano thanks for the review. I think it's a matter of preference. But tbh I'm not sure about `center` either. Because with center there's misalignment in all cases. 
Let's merge this and I'll revisit this again after RSM. Moreover, we need to remove the Jetpack visitor stat as we know it's completely wrong. So likely we probably need to redesign the card a bit more.

PR #15792, opened 2026-04-30
Add assistant headless test harness

hichamboushaba:
### Description
Part of WOOMOB-2896 follow-up work.

This adds a reusable headless test harness for the AI assistant core test fixtures, modeled after the iOS headless harness approach but adapted to the Android JVM/Robolectric test stack.

The initial implementation intentionally uses scripted model stream events and scripted tool results. That keeps the first version simple and deterministic while still validating the agentic loop end to end: assistant text streaming, tool-call assembly, tool execution, follow-up model turns, safety confirmation handoff events, loop outcomes, and hard-check evaluation.

This does not try to validate real LLM behavior or live WooCommerce API responses yet. Instead, it establishes the reusable fixture surface in `:libs:ai-assistant:core` so a separate PR can build live demo-store headless tests on top of the same harness by swapping in higher-level tool/data wiring.

The harness can run assistant scenarios without launching UI, record assistant text, tool calls, tool result kinds, confirmation requests/results, and loop errors, and evaluate hard checks from JSON baselines. It also adds scripted chat, tool registry, and safety orchestrator fixtures so future live-demo-store smoke tests can reuse the same core harness without depending on production UI wiring.

The safety coverage follows the current trunk behavior: confirmed unsafe tools execute, while cancelled unsafe tools finish the turn with `LoopOutcome.STOPPED` and no tool execution.

### Test Steps
1. Review the new `:libs:ai-assistant:core` test fixtures under `core/src/testFixtures/kotlin/com/woocommerce/android/aiassistant/core/headless`.
2. Verify the headless tests cover safe tool execution, confirmed unsafe tool execution, default-cancelled unsafe tool execution, baseline parsing, and hard-check evaluation.
3. Verify the feature test demonstrates consuming the core headless fixtures from `:libs:ai-assistant:feature`.

### Images/gif
N/A

### DIFF
--- a/libs/ai-assistant/core/build.gradle
+++ b/libs/ai-assistant/core/build.gradle
@@ -1,5 +1,6 @@
 plugins {
     id("org.jetbrains.kotlin.jvm")
+    id("java-test-fixtures")
     alias(libs.plugins.kotlin.serialization)
     alias(libs.plugins.dependency.analysis)
 }
@@ -16,4 +17,7 @@ dependencies {
     testImplementation(libs.junit)
     testImplementation(libs.assertj.core)
     testImplementation(libs.kotlinx.coroutines.test)
+
+    testFixturesApi(libs.kotlinx.coroutines.core)
+    testFixturesApi(libs.kotlinx.serialization.json)
 }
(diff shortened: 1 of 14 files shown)


staskus (approved):
✅ LGTM

PR #15794, opened 2026-04-30
[RSM] Convert POS Issue Refund from dialog to full-screen flow | Part 2

samiuelson:
## File rename

⚠️ Do not merge label - don't merge until the parent branch is set to `trunk`.

### Description

Follow-up to the POS Issue Refund fullscreen-flow PR: https://github.com/woocommerce/woocommerce-android/pull/15727. Now that the refund UI is a fullscreen Compose Navigation route rather than a dialog, the leftover `Dialog`-prefixed identifiers have become misleading. This PR is a pure rename pass — no behavior change.

Renames:
- File `WooPosIssueRefundDialog.kt` → `WooPosIssueRefundScreen.kt` (the function inside was already named `WooPosIssueRefundScreen`).
- `WooPosRefundUIEvent.DialogOpened` → `RefundFlowOpened`.
- `WooPosRefundUIEvent.DialogDismissed` → `RefundFlowDismissed`.
- `WooPosRefundViewModel.handleDialogDismissed()` → `handleRefundFlowDismissed()` (and the matching `error()` message inside it).
- All usages in `WooPosRefundViewModelTest.kt` updated, including three test names that referenced `DialogDismissed`.

`WooPosRefundDetailsDialog` is intentionally left alone since it is still a real dialog (uses `WooPosDialogWrapper`).

This PR is stacked on top of the fullscreen-flow PR (`issue/pos-refund-fullscreen-flow`). The base will roll forward to `trunk` once that one merges.

### Test Steps

Since this is a pure rename refactor with no behavioral change, a quick smoke test is enough:
1. Open POS, go to Orders, pick a refundable order, tap "Issue refund".
2. Confirm the refund screen opens fullscreen as before, the close button works, and the back-to-orders refresh still happens after dismissing.
3. From a Booking, tap the IssueRefund overflow action and confirm the refund screen still opens with full-refund-only mode (no item selection, no Edit refund button).

### Images/gif

N/A


toupper (approved):
Pure rename pass, exactly as described. File rename, two UIEvent renames, the matching VM method + `error()` message, and three test names — all the call sites and test usages are updated, and `WooPosRefundDetailsDialog` is correctly left alone since it's still a real dialog. No behavior change. LGTM.

PR #15801, opened 2026-05-01
[WOOMOB-2907] Implement show_cards structured card contract

hichamboushaba:
### Description
Fixes WOOMOB-2907

Implements the Android v1 `show_cards` tool contract for AI Assistant order and product references. The tool now accepts `references: [{ family, id }]`, validates each reference deterministically, dedupes first-seen refs by `(family, id)`, enforces the max reference count, and reports rejected or missing refs with stable lower-snake-case reasons.

The result is split into the two intended surfaces:
- `structured`: compact model-visible data only, including counts, resolved refs, missing refs, and rejected refs.
- `uiStructured.cards`: app-owned card payloads for local UI rendering, kept out of model history.

This also adds a resolver seam for future real order/product resolution. The default resolver reports `not_found`, so this PR does not introduce a new direct FluxC/store/auth path and leaves production data lookup to the follow-up resolver ticket.

### Android/iOS Alignment
| Aspect | Status | Notes |
|---|---|---|
| Contract shape | Mostly aligned: both use `show_cards(references)` and the same six `structured` keys. | This PR keeps the shared contract shape. |
| Supported families | Android currently supports `order` and `product`. iOS also supports `customer` and `product_variation`, with `parent_id` for variation refs. | This is an expected Android v1 gap. Extra families can be added separately when Android expands card coverage. |
| Model/UI split | Aligned: both keep render payloads out of model history and use `uiStructured` for UI-only card data. | This is the main boundary this PR protects. |
| Resolver/runtime parity | Not aligned yet: Android has only the contract seam and default `not_found`; iOS resolves real entities and has card-render bridging. | This is expected for this slice; resolver and UI/card follow-up tickets cover the runtime behavior. |
| Summary fields | Android has compact order/product summaries; iOS exposes a little more, such as `customer_name`. | Summary fields are intentionally compact here. Since Android plans to reuse existing app cards, any summary gaps should be revisited during UI/card integration rather than expanding model-visible `structured` now. |
| Validation/rejections | Functionally aligned, but reason strings differ: Android uses ticket-required lower-snake-case values; iOS uses its own enum raw values. | Shared tooling should not assume identical reason strings across platforms. |
| Tests | Android covers the contract/seam; iOS also covers resolver/data behavior. | Resolver/data coverage belongs with the Android real-resolution follow-up. |

### Test Steps
1. N/A - this is an internal AI Assistant tool contract with no manual UI surface in this ticket.

### Images/gif
N/A

[...]

PR #15805, opened 2026-05-01
[WOOMOB-2966] Implement show_cards reference resolver

hichamboushaba:
### Description
Fixes WOOMOB-2966

This PR implements the Android AI Assistant `show_cards` reference resolver on top of the structured card contract added in #15801. That earlier PR defined the tool input/output shape and protected the model/UI split, but its default resolver intentionally returned `not_found` for every valid ref. This PR replaces that seam with real order/product resolution while preserving the same `show_cards(references)` contract.

#### What changed

`show_cards` now resolves validated `order` and `product` references into app-owned card payloads:

- refs are still validated and deduped before resolution
- unsupported families remain rejected before any data-source call
- order refs are resolved through `AIOrdersDataSource`
- product refs are resolved through `AIProductsDataSource`
- refs are batched once per supported family
- the final output preserves the input display order after validation/dedupe
- resolved refs include compact model-visible summaries
- unresolved refs are returned in `missing_refs` with stable reasons

The resolver builds two separate result surfaces:

- `structured`: compact model-visible data only — counts, resolved refs, missing refs, rejected refs, and small summaries
- `uiStructured.cards`: app-owned render payloads for local UI rendering, kept out of model history

This keeps raw FluxC/store entities and full render payloads out of model-visible history.

#### Cache-first resolver behavior

The resolver path now prefers locally cached FluxC data before fetching. The new batch helpers first read the local store cache, then fetch only IDs that were not already cached:

- `AIOrdersDataSource.getOrders()` reads `WCOrderStore.getOrdersByIdsAndSite()` first, then fetches uncached misses via the existing include-backed order fetch path.
- `AIProductsDataSource.getProducts()` reads `WCProductStore.getProductsByRemoteIds()` first, then fetches uncached misses with `includedProductIds` and `forceRefresh = false`.

The helper return type, `CachedLookupResult<T>`, carries resolved items plus cache hit/miss and fetch status metadata. This lets cached cards still render if fetching uncached misses fails. In that partial-failure case, only refs that were not cached and still cannot be resolved are marked `fetch_failed`. If the fetch succeeds but an ID is still absent from storage, that ref remains `not_found`.

True in-memory same-turn tool-result caching is intentionally deferred. The current tool registry and agent loop do not expose a turn-scoped entity cache to tool handlers without broader orchestration changes, so this PR implements the safe local-cache-first slice only.

#### Data-source and module boundaries

The resolver stays behind the existing AI data-source boundary. It does not import or call `WCOrderStore`, `WCProductStore`, `SelectedSite`, FluxC error types, auth, or REST/cache APIs directly. Selected-site lookup, FluxC access, fetch behavior, and error mapping remain owned by `AIOrdersDataSource` and `AIProductsDataSource`.

All production changes are contained in `:libs:ai-assistant:feature`. There are no `:WooCommerce` host changes, no new auth path, no customer/product-variation card support, and no UI rendering changes in this PR.

#### Tests added or updated

- resolver batching and display-order preservation across mixed order/product refs
- order/product summary and card payload mapping from app-owned data
- positive numeric ID validation before resolver lookup
- missing refs after successful lookup returning `not_found`
- per-family fetch failure isolation
- cached order/product hits short-circuiting fetches
- partial cache hits still being returned when fetching misses fails
- unresolved miss refs becoming `fetch_failed` only when the miss fetch failed
- compact `structured` output excluding private render payload fields
- catalog/schema coverage for the `show_cards` tool descriptor

Validation performed locally after rebasing onto current `origin/trunk`:

- `ANDROID_HOME=/Users/hicham/Library/Android/sdk ./gradlew :libs:ai-assistant:feature:testDebugUnitTest --tests "*.ShowCardsToolHandlerTest" --tests "*.ShowCardsResolverTest" --tests "*.AIOrdersDataSourceTest" --tests "*.AIProductsDataSourceTest" --tests "*.WooCommerceToolCatalogTest"`
- `ANDROID_HOME=/Users/hicham/Library/Android/sdk ./gradlew detektAll --auto-correct`
- `git diff --check origin/trunk...HEAD`
- checked changed Kotlin files for `FIXME`, force unwraps, and wildcard imports

### Test Steps
Rich cards are not surfaced yet, so green CI is enough.

### Images/gif
N/A

### DIFF
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/cards/ShowCardsResolver.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/cards/ShowCardsResolver.kt
@@ -1,6 +1,15 @@
 package com.woocommerce.android.aiassistant.tools.handlers.cards
 
+import com.woocommerce.android.aiassistant.di.AiAssistantJson
+import com.woocommerce.android.aiassistant.tools.orders.AIOrdersDataSource
+import com.woocommerce.android.aiassistant.tools.products.AIProductsDataSource
+import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.encodeToJsonElement
+import kotlinx.serialization.json.jsonObject
+import org.wordpress.android.fluxc.model.WCProductModel
+import org.wordpress.android.fluxc.persistence.entity.OrderEntity
+import javax.inject.Inject
 
 internal interface ShowCardsResolver {
     suspend fun resolve(refs: List<ValidatedRef>): List<ShowCardsResolution>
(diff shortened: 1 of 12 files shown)


libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/products/AIProductsDataSource.kt
```diff
@@ -84,6 +85,66 @@ internal class AIProductsDataSource @Inject constructor(
...
+                    cacheHitCount = 0,
+                    cacheMissCount = 0,
+                    fetchAttempted = false,
+                    fetchFailed = false,
+                )
+            )
+        }
+
+        val site = selectedSite.get()
+        val cachedProducts = productStore.getProductsByRemoteIds(site, ids)
```
staskus, line 103: Claude flagged one thing:

> The products data source re-reads getProductsByRemoteIds after a successful fetch, but orders concatenate cachedOrders + page.orders from the fetch response. Worth mirroring the products pattern here for symmetry and durability against potential FluxC dispatcher races.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/cards/ShowCardsResolver.kt
```diff
@@ -21,12 +31,125 @@ internal sealed interface ShowCardsResolution {
...
+            OrderSummary(
+                id = ref.id,
+                number = number,
+                status = status,
+                total = total,
+                currency = currency,
+                dateCreated = dateCreated,
+            )
+        ),
+        card = ShowCardPayload(
```
staskus, line 99: So ShowCardPayload is very specific on Android with titles, subtitles, badges. On iOS for now, it all basically is like "attributes". I'm not sure which makes the most sense yet, depends on what and how we display in UI.

staskus (approved):

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/cards/ShowCardsResolver.kt
hichamboushaba: Adjusted the planned shape so `ShowCardPayload` stays as the app-only UI envelope while family-specific facts move into typed `ShowCardDetails` subclasses instead of `subtitle`/`badges` plus a generic attributes map. `structured` remains compact and unchanged for model history; the typed details serialize only in `uiStructured.cards` with a local `kind` discriminator.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/handlers/cards/ShowCardsResolver.kt
hichamboushaba: Good point, I think Android's structure was picked by AI because we have only two families now, and it didn't think of scalability later.
I updated our logic to be similar to iOS, but with a difference: we use typed items, not just a map of attributes.

Check [3c7e488](https://github.com/woocommerce/woocommerce-android/pull/15805/commits/3c7e48824cef86cb6b4f64e86869051a2ed657e7)

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/products/AIProductsDataSource.kt
hichamboushaba: Fixed in f1811bd9eac. Orders now re-read the local store after the include fetch, matching the product path.

PR #15807, opened 2026-05-01
[WOOMOB-2915] Show unsafe tool confirmations inline

hichamboushaba:
### Description
Fixes WOOMOB-2915

This PR implements inline confirmation cards for unsafe AI Assistant write tools. Instead of showing confirmation in a separate panel, the assistant now keeps the confirmation request and its final state inside the transcript. The interaction is intentionally close to the iOS direction in https://github.com/woocommerce/woocommerce-ios/pull/17027, while staying within the Android assistant runtime and Compose UI model.

#### What changes for users

When the assistant needs approval before running an unsafe write, merchants now see a durable card in the conversation with a `Pending change` state, a readable summary, and Cancel / Confirm actions. After the merchant decides, the same card remains in the transcript as either `Confirmed` or `Cancelled`, so the conversation keeps a record of what happened.

For single-record order, product, and variation updates, the card can show the current value under **Now** and the proposed value under **After**. Bulk updates intentionally show the proposed **After** values only. While a confirmation card is pending, the composer stop control is hidden so the card actions are the clear decision point.

The card UI was also polished to better match the iOS feel without pixel matching: it uses Material 3 / WooTheme color tokens, a bordered card surface, clearer field rows, state icons, and stable equal-width Cancel / Confirm actions.

#### Runtime and UI behavior

The assistant transcript now supports segmented messages, so one assistant message can contain streamed text, an inline confirmation card, and follow-up text that arrives after the confirmation resolves. Confirmation cards carry their own `Pending`, `Confirmed`, and `Cancelled` state instead of relying on a separate top-level pending panel.

Confirm and Cancel both go through the core `ConfirmationResult` path. Cancelling from the card, stopping while a confirmation is open, or resetting an open gate resolves the pending confirmation as `Cancelled` without running the write. Confirmed cards stay in place and later assistant text is appended after the card.

Confirmation previews now build localized summaries and field rows for order, product, product variation, and bulk update tools. Single-record snapshots fetch fresh order data for current values, normalize `wc-` order status prefixes, and preserve fractional stock quantities.

This PR also normalizes assistant replay payloads for Jetpack AI by sending empty assistant content for tool-call-only replay messages, which matches the backend contract while preserving normal assistant text content.

#### Dependency notes

This PR is stacked on `issue/WOOMOB-2966-implement-show-cards-reference-resolver`, so the GitHub diff shows the WOOMOB-2915 confirmation-card work on top of that branch.

#### Tests added or updated

- core loop confirmation request / resolution behavior
- runtime confirmation dispatch and cancellation handling
- preview building and rendering for single-record and bulk updates
- snapshot resolution for current order, product, and variation values
- fresh order reads for confirmation snapshots
- transcript segment ordering around pending, confirmed, and cancelled cards
- hiding the stop control while a confirmation card is pending
- Jetpack AI / OpenAI request mapping for tool-call-only assistant replay messages

### Test Steps
1. Build and install the app from this branch, then open the AI Assistant on a store that can use write-capable assistant tools.
2. Ask the assistant to update a single order, for example changing an order status.
3. Verify the confirmation appears inline in the assistant transcript with `Pending change`, a state icon, a clear summary, Now / After rows, and Cancel / Confirm actions.
4. Verify there is no modal, bottom sheet, or separate confirmation panel, and that the stop control is hidden while the confirmation card is pending.
5. Tap Cancel and verify the card remains in the transcript as `Cancelled`, the turn ends cleanly, and the order is not changed.
6. Repeat the single-order update and tap Confirm. Verify the card changes to `Confirmed`, the assistant continues after the card, and the order change is applied.
7. Ask for a single product or variation update that changes price, stock, SKU, or status. Verify the card shows current values under Now and proposed values under After.
8. Ask for a bulk order or product update. Verify the card uses a bulk summary and only shows the proposed After values.
9. Start another unsafe update and use the conversation cancel/reset path while the card is pending. Verify the pending card resolves to `Cancelled` and no write occurs.

### Images/gif

https://github.com/user-attachments/assets/6993be7d-3276-47f4-a8e6-77154f0b36ee

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -271,7 +271,6 @@ class AgenticLoopImpl(
         outcome.completed.forEach { completed ->
             newTurnMessages.add(completed.toToolMessage())
         }
-        emit(LoopEvent.Failed(AssistantError.Cancelled))
         emit(LoopEvent.Finished(LoopOutcome.STOPPED, history + newTurnMessages, retryAvailable = false))
     }
 
--- a/libs/ai-assistant/core/src/test/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImplTest.kt
+++ b/libs/ai-assistant/core/src/test/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImplTest.kt
@@ -663,7 +663,8 @@ class AgenticLoopImplTest {
         }
 
     @Test
-    fun `given UNSAFE tool is cancelled, when loop awaits confirmation, then cancelled error is emitted and history is clean`() = runTest {
+    fun `given UNSAFE tool is cancelled, when loop awaits confirmation, then turn stops cleanly and history is clean`() =
+        runTest {
         val unsafeDescriptor = ToolDescriptor(
             name = "orders_update",
             description = "Updates an order",
@@ -724,8 +725,7 @@ class AgenticLoopImplTest {
 
         assertThat(events.filterIsInstance<LoopEvent.ConfirmationResolved>().single().result.decision)
             .isEqualTo(ConfirmationDecision.CANCELLED)
-        assertThat(events.filterIsInstance<LoopEvent.Failed>().single().error)
-            .isEqualTo(AssistantError.Cancelled)
+        assertThat(events.filterIsInstance<LoopEvent.Failed>()).isEmpty()
         assertThat(events.filterIsInstance<LoopEvent.ToolCallStarted>()).isEmpty()
         assertThat(events.filterIsInstance<LoopEvent.ToolCallFinished>()).isEmpty()
         assertThat(registryExecuted).isFalse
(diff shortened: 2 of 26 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
The happy path works.

I found it to be less stable in a more custom scenarios:

#### ⚠️  Bulk changes

 > Ask for a bulk order or product update. Verify the card uses a bulk summary and only shows the proposed After values.

Asking to "Change last 5 orders to failed" asks for 5 different confirmations rather than a single confirmation for a bulk change. 

---

For iOS, I haven't yet wired before and after values. I recognize it can require another tool call. It probably works better on Android if it takes a cached value. Something I should consider plugging on iOS as well.

hichamboushaba:
> Asking to "Change last 5 orders to failed" asks for 5 different confirmations rather than a single confirmation for a bulk change.

This shouln't happen, unless the model decided to make 5 API requests to update them.
Currently, we don't set any system prompt on Android, so the model is quite unreliable, and it could explain what happened to you.
In my testing, it behaved correctly:
Screenshot_20260504_140931

> For iOS, I haven't yet wired before and after values. I recognize it can require another tool call. It probably works better on Android if it takes a cached value. Something I should consider plugging on iOS as well.

I think a last edit deleted the screenshots from the comment, I can't see anything there to understand the issue well.
But anyway, I think I'll merge this as is, for UI itself, we can align later, all of my work right now is just tentative. Because we need full alignment on UI.

PR #15808, opened 2026-05-01
[RSM-1628] Add disabled system notifications warning

irfano:
### Description

Fixes RSM-1628

Adds an inline warning to the smarter notification settings screen when app-level notifications are disabled in Android system settings.

The warning appears below the notification type rows and opens the app notification settings when tapped, so merchants can re-enable notifications from the device settings screen.

### Test Steps

1. Use a Woo-driven site with smarter notifications enabled.
2. Disable WooCommerce app notifications from Android system settings.
3. Open Settings > Notifications.
4. Verify the warning appears below the notification type rows.
5. Tap the warning.
6. Verify Android opens the WooCommerce app notification settings.
7. Enable WooCommerce app notifications from Android system settings.
8. Return to the WooCommerce app and verify the warning is no longer shown.

### Images/gif


AdamGrzybkowski (approved):
Looks good and works as described!

One thing, though: We should also check whether each channel is enabled and add a similar warning to each push type mapped to that channel.

irfano:
> One thing, though: We should also check whether each channel is enabled and add a similar warning to each push type mapped to that channel.

👍🏻 I’ll handle it in a separate PR.

PR #15809, opened 2026-05-01
[RSM-1628] Add cha-ching restore action to New Orders settings

irfano:
### Description
Fixes RSM-1628

Stacked on #15808.

Adds the existing Cha-Ching sound restore action to the New Orders notification settings screen. The screen now reads and refreshes the New Order notification channel sound state, then shows the action above the “Notify me for” section when the sound is disabled or changed from the default.

### Test Steps
1. Open Settings > Notifications > New Orders on a Woo-driven store.
2. Change the New Orders notification channel sound in Android system settings so it is disabled or no longer uses the default sound.
3. Return to the New Orders notification settings screen.
4. Confirm the Cha-Ching restore action appears above the “Notify me for” section.
5. Tap the action and confirm the success snackbar appears.
6. Return to the screen and confirm the action is no longer shown once the channel uses the default sound again.

### Images/gif
Screenshot_20260504_160556 Screenshot_20260504_160604

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
@@ -2,6 +2,7 @@ package com.woocommerce.android.ui.prefs.notifications
 
 import android.content.res.Configuration
 import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -11,6 +12,7 @@ import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -36,11 +53,45 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
         _viewState.update { it.copy(thresholdAmount = amount) }
     }
 
+    fun refreshNotificationSettings() {
+        _viewState.update {
+            it.copy(newOrderNotificationSoundStatus = notificationChannelsHandler.checkNewOrderNotificationSound())
+        }
+    }
+
+    fun onEnableChaChingSoundClicked() {
```
irfano, line 62: This intentionally mirrors the existing `NotificationSettingsViewModel.onEnableChaChingSoundClicked()` implementation. I considered extracting the shared logic, but the duplicated block is small and adding another abstraction/file felt unnecessary for this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
```diff
@@ -116,6 +137,30 @@ private fun ThresholdAmountField(
...
+            text = title,
+            style = MaterialTheme.typography.titleMedium
+        )
+        Text(
+            text = subtitle,
+            style = MaterialTheme.typography.bodyMedium
+        )
+    }
+}
+
```
AdamGrzybkowski, line 176: I know this is an old UI, but can we make it look more like an actual button? 

This may not be a proper button, but at least it matches the screen UI a bit better and has clearly visible boundaries. 
Image

I would also consider moving this above the `enable new order notification` switch, since the switch doesn't affect whether this button is enabled.

AdamGrzybkowski (approved):
Works well, but I left a suggestion to improve the UI a bit. 

I will approve this anyway to not block the PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsScreen.kt
irfano: Sounds good! I made the change: 4e41b5215fb8f488074b0dff75c95dfb2a909c1e

PR #15810, opened 2026-05-01
Add explicit orderby date desc to order list WC REST requests

zain0ul0abiddin:
### Description

Investigated orders list showing an outdated “newest” order on Android, while iOS / REST looked correct.
The orders list loads summaries through `OrderRestClient.fetchOrderListSummaries`, which called `GET /wc/v3/orders` with `per_page`, `offset`, and `_fields` but did **not** send `orderby` or `order`. On some stores (Jetpack / Woo), the effective sort was not newest-first, so the first page could start at an older order even when newer orders existed.

Closes: WOOMOB-2971

**Findings:** The behavior tracked to **Experimental → Application passwords for Jetpack**. When enabled, Woo Android can use **direct site REST** with application passwords; when disabled, it uses the **Jetpack tunnel**. Repro: **wrong/incomplete list with app passwords on**, **correct with Jetpack-only**. Same `wc/v3/orders` request **without** `orderby`/`order` via **Jetpack** returned correct newest-first data for the tested store—so the issue is likely **path / host / cache / security** (e.g. LiteSpeed Cache, CleanTalk), not the order list UI alone. **iOS** also omits explicit `orderby`/`order` for `loadAllOrders` but uses **page**-based pagination and a different `_fields` set than Android summaries.

**Change:** Add explicit **`orderby=date`** and **`order=desc`** to `fetchOrderListSummaries` and `fetchOrdersListFirstPage` in `OrderRestClient` (already used by `fetchOrders`). Harmless defensive alignment with WC defaults and our other client code; filters (`after`, `before`, `status`, `search`, etc.) are unchanged.

**Note:** Does not replace site-side investigation for tunnel vs app-password / cache; consider linking related Linear reports that mention missing or inconsistent orders in the app.

### Test Steps

1. On a store showing this issue, toggle **Experimental → Application passwords** off → Orders list and pull-to-refresh; confirm newest orders.
2. Toggle on → repeat; The order list should remain the same


samiuelson (approved):
LGTM!

PR #15811, opened 2026-05-01
Replace POS reader-disconnected fullscreen error with payment-method row

samiuelson:
### Description

Reshapes the bottom of the POS checkout (`WooPosTotalsScreen`) from a single full-width action into a two-button payment-method row. With this change:

- When the card reader is **connected**, the existing single `Cash` button is replaced with a row hosted by a new `WooPosCheckoutPaymentButtons` composable (today: `Cash` only — but the row is now the seam every method plugs into).
- When the card reader is **disconnected**, the full-screen "reader not connected" error UI is removed entirely. The totals stay visible and a `Card reader` / `Cash` button row sits beneath them, so the merchant can choose to connect a reader or take cash without leaving the totals view.

This is a foundation change: it unblocks the gated Tap to Pay entry point that ships in the stacked follow-up PR. No payment-collection logic moves; existing UI events (`OnCashPaymentClicked`, `ConnectReaderClicked`) are reused unchanged.

### Test Steps

1. Install a debug build on a phone or tablet: `./gradlew :WooCommerce:installWasabiDebug`.
2. Open POS, add an item, navigate to Checkout.
3. **Reader disconnected** (no card reader paired): verify totals are visible and that a button row below them shows `Card reader` (primary on tablet / secondary on phone) and `Cash`. Tap `Card reader` → existing connect-reader flow opens. Tap `Cash` → existing cash-payment flow opens. Confirm the previous full-screen "reader not connected" error is no longer shown.
4. **Reader connected** (Bluetooth reader paired): verify the totals + the existing "ready for payment" UI are unchanged, and that the bottom action is the same `Cash` button as before (still single full-width).
5. Verify Compose previews in `WooPosTotalsScreen.kt` render correctly in Android Studio.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCheckoutPaymentButtons.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCheckoutPaymentButtons.kt
@@ -1,0 +1,47 @@
+package com.woocommerce.android.ui.woopos.home.totals
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.navigationBarsPadding
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #13654, opened 2025-03-04
[Jetpack Setup] Add networking layer for the Connection API 

hichamboushaba:
Closes: #13629

> [!NOTE]
> Depends on https://github.com/woocommerce/woocommerce-android/pull/13645 

### Description
This PR adds the networking integration with the Jetpack Connection API.

### Steps to reproduce
Code review should be enough, the integration will be tested in the next PR.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

## Reviewer (or Author, in the case of optional code reviews):

Please make sure these conditions are met before approving the PR, or request changes if the PR needs improvement:

- [ ] The PR is small and has a clear, single focus, or a valid explanation is provided in the description. If needed, please request to split it into smaller PRs.
- [ ] Ensure Adequate Unit Test Coverage: The changes are reasonably covered by unit tests or an explanation is provided in the PR description.
- [ ] Manual Testing: The author listed all the tests they ran, including smoke tests when needed (e.g., for refactorings). The reviewer confirmed that the PR works as expected on big (tablet) and small (phone) in case of UI changes, and no regressions are added.

PR #15815, opened 2026-05-04
[AI Support Chat] Network layer for Odie chat endpoint

iamgabrielma:
Closes WOOMOB-2951. Stacked on #15814 , which needs to be reviewed/merged first.

Sorry for the 500+ LOC PR, mostly is networking scaffolding + tests.

### Description

Adds the transport layer for the wpcom Odie chat endpoint (`POST/GET/wpcom/v2/odie/chat/{bot_slug}[/{chat_id}]`) used by the AI Troubleshooting feature. Mirrors iOS woocommerce/woocommerce-ios#16975

No UI / repository / persistence yet.

Adds a transport layer for the WordPress.com Odie chat endpoint:

 - `POST /wpcom/v2/odie/chat/{bot_slug}`
 - `POST /wpcom/v2/odie/chat/{bot_slug}/{chat_id}`
 - `GET  /wpcom/v2/odie/chat/{bot_slug}/{chat_id}`

## Test steps

Manual smoke testing isn't possible in this PR, there's no UI consumer yet.

- [ ] `./gradlew :WooCommerce:assembleWasabiDebug` green.
- [ ] `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*SupportChat*"` 16 tests pass (response decode  + rest client + role deserializer).
- [ ] `./gradlew detektAll` clean.

cc: @itsmeichigo


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClient.kt
```diff
@@ -0,0 +1,67 @@
...
+    dispatcher: Dispatcher,
+    @Named("regular") requestQueue: RequestQueue,
+    accessToken: AccessToken,
+    userAgent: UserAgent
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+
+    suspend fun sendMessage(
+        botSlug: String,
+        message: String,
+        context: Map<String, Any>
```
malinajirka: ❓ Using `Any` here feels a bit off. What values are we expecting here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClient.kt
```diff
@@ -0,0 +1,67 @@
...
+        params = emptyMap(),
+        clazz = SupportChatResponse::class.java
+    )
+
+    private fun chatUrl(botSlug: String): String = "$ODIE_CHAT_BASE_URL/$botSlug"
+
+    private fun chatUrl(botSlug: String, chatId: Long): String = "${chatUrl(botSlug)}/$chatId"
+
+    companion object {
+        private const val ODIE_CHAT_BASE_URL = "https://public-api.wordpress.com/wpcom/v2/odie/chat"
```
malinajirka: ⚠️ We usually define endpoints in text-files and let the framework generate Kotlin/Java for us - in your case, your endpoint should be defined in https://github.com/woocommerce/woocommerce-android/blob/10bef492cc6dc383d71a08e1bf5a7cad5e571d75/libs/fluxc-processor/src/main/resources/wp-com-v2-endpoints.txt => if you add it there and rebuild the project, it'll generate the corresponding Java/Kotlin classes (`org.wordpress.android.fluxc.generated.endpoint.WPCOMV2`) => then you can use it from code by referencing for example `WPCOMV2.odie...`

You can look for example at this PR which introduces a new endpoint consumer https://github.com/woocommerce/woocommerce-android/pull/13654/changes

malinajirka (changes requested):
Thanks @iamgabrielma! I haven't done a full review yet, but I dropped a comment which will likely result in changes, so I'll do a full review then.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClient.kt
iamgabrielma: Updated to JsonObject 30d6af21fc9b53b59064a3d0632af5e829fd25d5

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClient.kt
iamgabrielma: Ah, interesting. Updated on c10b90e0f4f0cd18ecfc43fb4422997e42c52ffb

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClientTest.kt
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private suspend fun stubPostResponse(
+        data: SupportChatResponse? = null,
+        error: WPComGsonNetworkError? = null
+    ) {
+        val response: Response<SupportChatResponse> = if (error != null) {
+            Response.Error(error)
+        } else {
+            Response.Success(data ?: mock(), emptyList())
```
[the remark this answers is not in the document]
iamgabrielma: addressed on 5224252de9081fe817660accd34b66e6ed30001c

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/networking/SupportChatRestClientTest.kt
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private suspend fun stubGetResponse(
+        data: SupportChatResponse? = null,
+        error: WPComGsonNetworkError? = null
+    ) {
+        val response: Response<SupportChatResponse> = if (error != null) {
+            Response.Error(error)
+        } else {
+            Response.Success(data ?: mock(), emptyList())
```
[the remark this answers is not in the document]
iamgabrielma: addressed on 5224252de9081fe817660accd34b66e6ed30001c

iamgabrielma:
Thanks for the review @malinajirka , this is ready for another go 🙇

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatResponse.kt
```diff
@@ -0,0 +1,11 @@
+package com.woocommerce.android.ui.aisupportchat.networking.model
+
+import com.google.gson.annotations.SerializedName
+
+data class SupportChatResponse(
+    @SerializedName("chat_id") val chatId: Long,
```
samiuelson, line 6: `chatId: Long` is non-null with no default — Gson silently coerces a missing `chat_id` (or JSON `null`) to `0L`. iOS PR #16975 also treats this as required, so keep the type, but consider a fixture-based test that decodes JSON without `chat_id` and asserts the intended behavior. Same risk applies to `SupportChatMessage.messageId` and `content`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatResponse.kt
```diff
@@ -0,0 +1,11 @@
...
+    @SerializedName("session_id") val sessionId: String?,
+    @SerializedName("bot_slug") val botSlug: String?,
+    @SerializedName("bot_version") val botVersion: String?,
```
samiuelson: iOS PR #16975 treats these three as required non-optional. Consider matching to keep the response envelope symmetric across platforms:

```suggestion
    @SerializedName("session_id") val sessionId: String,
    @SerializedName("bot_slug") val botSlug: String,
    @SerializedName("bot_version") val botVersion: String,
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatMessageContext.kt
```diff
@@ -0,0 +1,8 @@
+package com.woocommerce.android.ui.aisupportchat.networking.model
+
+import com.google.gson.annotations.SerializedName
+
+data class SupportChatMessageContext(
+    @SerializedName("sources") val sources: List<SupportChatSource>? = null,
```
samiuelson: iOS defaults `sources` to `[]` (non-optional after decode). Keeping it nullable on Android forces every consumer to null-check; defaulting to `emptyList()` matches iOS and makes the call site simpler:

```suggestion
    @SerializedName("sources") val sources: List<SupportChatSource> = emptyList(),
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatSource.kt
```diff
@@ -0,0 +1,10 @@
+package com.woocommerce.android.ui.aisupportchat.networking.model
+
+import com.google.gson.annotations.SerializedName
+
+data class SupportChatSource(
+    @SerializedName("title") val title: String?,
+    @SerializedName("url") val url: String?,
```
samiuelson: iOS PR #16975 requires `title` and `url`; an all-null `SupportChatSource` is semantically meaningless anyway. Consider tightening — `heading`/`content` can stay nullable.

```suggestion
    @SerializedName("title") val title: String,
    @SerializedName("url") val url: String,
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatFlags.kt
```diff
@@ -0,0 +1,10 @@
+package com.woocommerce.android.ui.aisupportchat.networking.model
+
+import com.google.gson.annotations.SerializedName
+
+data class SupportChatFlags(
```
samiuelson, line 7: Heads-up: the defaults here are load-bearing and rely on a subtle Gson rule. When **every** parameter has a default, kotlinc emits a synthetic no-arg constructor that Gson's `ConstructorConstructor` finds and uses, so missing fields keep their declared defaults — that's why the `given missing optional flags, when decoded, then defaults to false` test passes.

The instant any future field becomes non-defaulted, kotlinc no longer emits that synthetic ctor and Gson falls back to `sun.misc.Unsafe.allocateInstance()`, which bypasses the constructor entirely. Every Boolean default silently stops working — they all coerce to `false` (the JVM zero-value). The existing test would still pass because it only asserts `false`.

Worth either (a) a one-line comment on the class warning future maintainers, or (b) a regression test that defaults a flag to `true` and asserts a round-trip preserves it — that would catch any future Unsafe-fallback regression.

samiuelson (approved):
Approving — no blockers, but a few suggestions worth folding into this PR or a follow-up. Inline notes anchor the bigger items.

## Headline items

1. **Align response-envelope nullability with iOS PR #16975.** iOS treats `sessionId`, `botSlug`, `botVersion`, `SupportChatMessageContext.sources`, and `SupportChatSource.title`/`url` as required non-optional. The two platforms already agree on `SupportChatFlags` (boolean defaults to `false`); tightening the envelope keeps callers symmetric. Inline suggestions below.
2. **Document the Gson default-handling contract in `SupportChatFlags`.** The defaults work today only because every field has a default (kotlinc emits a synthetic no-arg ctor that Gson picks up). Adding any non-defaulted field silently switches Gson to `Unsafe.allocateInstance()`, zeroing every field — and the existing missing-flags test would still pass since it only asserts `false`. Inline note below.
3. **Pin the `chatId` contract.** Non-null `Long` with no default → Gson silently coerces a missing `chat_id` to `0L`. iOS made it required too, so keep the type, but a fixture-based test decoding JSON without `chat_id` would document whatever behavior is intended.

## Test coverage gaps

- `sendFollowUpMessage` has no error-path test (sendMessage and fetchChat both do).
- `fetchChat` and `sendFollowUpMessage` happy-path tests check only the URL — no `Response.Success` / `data.chatId` assertion. A future refactor returning `Response.Error` would still pass.
- `sendMessage` body assertion doesn't verify the exact key set — mirror the `containsOnlyKeys("message")` pattern already used in the follow-up test.
- `stubGetResponse` uses positional mock args while `stubPostResponse` uses named args. Convert for parity — any reordering of `WPComGsonRequestBuilder.syncGetRequest` parameters would silently match the wrong slot.
- No test exercises an empty `messages: []` response or `SupportChatSource` with all-null fields.

## Minor

- The six new DTO classes are `public`; consider `internal` since they live under `ui/aisupportchat/networking/model/`.
- Stray blank line between the two new Odie entries and the next group in `wp-com-v2-endpoints.txt`.

## Positives

- `SupportChatRole` forward-compat via `UNKNOWN` + custom `JsonDeserializer` — covered by both unknown-wire-value and malformed-JSON tests.
- DI signature, `@Named("regular")` queue, and `BaseWPComRestClient` superclass call exactly mirror `SitePlanRestClient` — clean reuse of the established transport pattern.
- BDD test naming, `BaseUnitTest`/`testBlocking`, AssertJ usage, and fixture organization under `support-chat/` are all in line with repo conventions.
- Endpoint definition (`#String` on `bot_slug`, untyped `chat_id` defaulting to `long`) is correct against `RESTPoet` and matches the Kotlin call sites.

Nice work on the foundational layer.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatResponse.kt
iamgabrielma: https://github.com/woocommerce/woocommerce-android/pull/15815/changes/0b415536dfbfb736a2e6a4a04d98b6e42b632a06

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatResponse.kt
iamgabrielma: Updated! https://github.com/woocommerce/woocommerce-android/pull/15815/changes/0b415536dfbfb736a2e6a4a04d98b6e42b632a06

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatMessageContext.kt
iamgabrielma: Updated https://github.com/woocommerce/woocommerce-android/pull/15815/changes/0b415536dfbfb736a2e6a4a04d98b6e42b632a06

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatSource.kt
iamgabrielma: Updated https://github.com/woocommerce/woocommerce-android/pull/15815/changes/0b415536dfbfb736a2e6a4a04d98b6e42b632a06

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatFlags.kt
iamgabrielma: https://github.com/woocommerce/woocommerce-android/pull/15815/changes/0534afdff15efb6626b6c4829f8b1905ac34e157

iamgabrielma:
Thanks for the review @malinajirka @samiuelson , this one is ready for re-review, it seems 1 approval is not enough to merge?

malinajirka (approved):

PR #15819, opened 2026-05-04
[WOOMOB-2970] Add typing and tool activity indicators to assistant chat

hichamboushaba:
### Description
Fixes WOOMOB-2970

This PR adds in-flight feedback to the Android AI Assistant chat surface. It builds on the inline transcript model from WOOMOB-2915 so merchants can see when the assistant is preparing a response, when it is working with store tools, and which tool steps have completed, without exposing internal tool arguments or raw tool results.

#### What changes for users

After sending a prompt, the chat now shows a standalone typing indicator at the bottom of the transcript while the assistant is working and before any response text has streamed. The indicator uses three pulsing dots in a compact rounded surface, matching the same in-flight language used by the tool activity UI.

When the assistant starts a store tool, the transcript shows a concise activity pill such as `Checking orders`, `Updating products`, or `Checking analytics`. Running tool calls use pulsing dots. Finished tool calls stay in the transcript with a checkmark, giving the conversation a durable step history similar to the completed confirmation cards. If a turn is stopped or fails while a tool is still running, the abandoned running pill is cleared so stale activity is not left behind.

The activity labels are intentionally human-readable and sanitized. The UI does not show scratchpad text, raw tool arguments, private JSON, structured tool results, or low-level tool names for unknown tools; unknown activity falls back to generic store-checking copy.

#### Runtime and UI behavior

The assistant runtime now forwards sanitized tool lifecycle events from the existing core loop into feature UI state. `ToolCallStarted` carries only the tool call ID and tool name; `ToolCallFinished` carries only the tool call ID. The ViewModel maps those events into assistant transcript segments, derives typing state from the existing active turn state, and clears or marks tool activity based on the terminal path.

The chat UI now renders assistant message segments as first-class rows:

- text remains in assistant/user message bubbles
- typing is a standalone bottom-row indicator
- running tool activity is a standalone pill with pulsing dots
- completed tool activity is the same pill with a checkmark
- confirmation cards continue to use their own card chrome instead of being wrapped in an extra assistant bubble

This avoids nested surfaces, keeps mixed assistant text/tool/card messages readable, and preserves the transcript as the source of truth for the visible assistant flow.

Accessibility semantics are kept on the visible activity surfaces. The typing indicator exposes `Assistant is thinking`, and tool activity exposes `Assistant is working: <activity>`. The global status pill and existing Send/Stop behavior are unchanged.

#### Dependency and safety notes

[...]

PR #15820, opened 2026-05-04
[WOOMOB-2967] Wire show cards UI payload into assistant messages

hichamboushaba:
### Description
Fixes WOOMOB-2967

This PR wires successful `show_cards` UI payloads into the visible Android AI Assistant transcript. It builds on the typed order/product card rendering work and the typing/tool-activity transcript model from #15819 so cards can be shown as first-class assistant message segments instead of remaining only in tool results.

#### What changes for users

When the assistant resolves order or product references through `show_cards`, the final assistant reply can now include native WooCommerce cards directly in the conversation. Those cards use the existing host-owned order/product row UI, support the existing open-order/open-product actions, and now render inside a rounded card surface with a visible background and subtle outline so the card content no longer floats on the chat background.

If the tool resolves only some requested references, the available cards are still shown. If no cards resolve, no empty card UI is displayed. Card duplicates are filtered by the first-seen `(family, id)` key for the active assistant turn.

#### Runtime and UI behavior

The assistant runtime now tracks tool names by tool-call ID while preserving the tool activity events from #15819. On `ToolCallFinished`, it emits `CardsResolved` only when the result is a successful `show_cards` call with parseable `uiStructured.cards`.

The ViewModel appends resolved cards as `AssistantUiSegment.Card` entries on the active assistant message. Cards that arrive between streamed text deltas stay attached to the same final visible assistant reply. Malformed, empty, non-success, non-`show_cards`, and arbitrary card-shaped JSON payloads do not create card segments.

The card parser uses `kotlinx-serialization-json` for AI payloads and keeps raw `(family, id)` keys alongside parsed render models so ViewModel-level dedupe follows the cross-platform first-seen rule.

#### Card rendering polish

Card chrome is owned by `AssistantCardSegment`, so all current and future assistant cards share the same transcript surface. The segment wraps renderer content in a Material 3 `Surface` with rounded corners, `surfaceContainerHighest`, and an `outlineVariant` border. Host renderers stay focused on card content and navigation rather than transcript container styling.

The host order card mapper also formats ISO order dates into the same short month/day style used by existing order rows, with a raw-string fallback for unparseable dates.

This PR is stacked on #15819 (`issue/woomob-2970-typing-indicator-pill`). The PR should be reviewed as the `show_cards` delivery and card-surface slice on top of that base.

The assistant feature module remains host-agnostic. `:libs:ai-assistant:feature` owns runtime delivery, UI state, parsing, and segment chrome. `:WooCommerce` owns WooCommerce-specific row mapping and navigation. There are no new auth paths, Jetpack AI JWT changes, Gson usage for AI payloads, or dependency-direction reversals.

#### Tests added or updated

- successful `show_cards` results emitting card events
- non-success `show_cards` results not emitting card events
- card-shaped `uiStructured` from non-`show_cards` tools being ignored
- malformed `uiStructured` producing no cards and no crash
- empty card payloads producing no card UI
- partial success preserving available cards
- arbitrary tool-result JSON not becoming fallback card UI
- appending resolved cards into assistant message state
- preserving cards on the final visible assistant message
- first-seen `(family, id)` dedupe, including same ID across different families
- parser raw key preservation and malformed payload handling
- host order-date formatting fallback behavior

### Test Steps
1. Build and install the Wasabi debug app from this branch.
2. Open the AI Assistant from the More menu on a store with recent orders or products.
3. Ask for a recent order or product in a card. **Note:** Given there is no system prompt, you need to explicitly ask for cards.
4. Verify the final assistant reply includes a native card attached to the assistant message, with a visible rounded background and subtle border.
5. Tap the card and verify it opens the corresponding order or product screen.
6. Ask for references that do not resolve to cards and verify no empty card UI is shown.
7. Ask a non-card question, such as revenue for today, and verify the response remains plain assistant text with no JSON or fallback card rendering.

### Images/gif
| Orders | Products |
| --- | --- |
| Screenshot_20260504_212428 |Screenshot_20260504_212337 |

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
@@ -10,6 +10,7 @@ import org.mockito.kotlin.mock
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import java.math.BigDecimal
+import java.util.Locale
 
 class WooAssistantCardRendererTest {
     private val currencyFormatter: CurrencyFormatter = mock()
@@ -17,9 +18,15 @@ class WooAssistantCardRendererTest {
 
     @Test
     fun `given assistant order card, when mapped, then host row model formats total with order currency`() {
+        val originalLocale = Locale.getDefault()
+        Locale.setDefault(Locale.US)
         whenever(currencyFormatter.formatCurrency("12.34", "USD")).thenReturn("$12.34")
 
-        val model = orderCard().toOrderSummaryRowModel(context, currencyFormatter)
+        val model = try {
+            orderCard().toOrderSummaryRowModel(context, currencyFormatter)
+        } finally {
+            Locale.setDefault(originalLocale)
+        }
 
         assertThat(model.number).isEqualTo("#1001")
         assertThat(model.date).matches("^May [12]$")
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
@@ -3,7 +3,6 @@ package com.woocommerce.android.ui.aiassistant
 import android.content.Context
 import com.woocommerce.android.R
 import com.woocommerce.android.aiassistant.ui.cards.AssistantCard
-import com.woocommerce.android.aiassistant.ui.cards.AssistantCardAction
 import com.woocommerce.android.util.CurrencyFormatter
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
@@ -14,15 +13,16 @@ import java.math.BigDecimal
 
 class WooAssistantCardRendererTest {
     private val currencyFormatter: CurrencyFormatter = mock()
+    private val context: Context = mock()
 
     @Test
     fun `given assistant order card, when mapped, then host row model formats total with order currency`() {
         whenever(currencyFormatter.formatCurrency("12.34", "USD")).thenReturn("$12.34")
 
-        val model = orderCard().toOrderSummaryRowModel(currencyFormatter)
+        val model = orderCard().toOrderSummaryRowModel(context, currencyFormatter)
 
         assertThat(model.number).isEqualTo("#1001")
-        assertThat(model.date).isEqualTo("2026-05-01T10:00:00Z")
+        assertThat(model.date).matches("^May [12]$")
         assertThat(model.customerName).isEqualTo("Jane Doe")
         assertThat(model.status).isEqualTo("processing")
         assertThat(model.statusColor).isEqualTo(R.color.tag_bg_processing)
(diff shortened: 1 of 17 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
```diff
@@ -14,15 +13,16 @@ import java.math.BigDecimal
...
     @Test
     fun `given assistant order card, when mapped, then host row model formats total with order currency`() {
         whenever(currencyFormatter.formatCurrency("12.34", "USD")).thenReturn("$12.34")
 
-        val model = orderCard().toOrderSummaryRowModel(currencyFormatter)
+        val model = orderCard().toOrderSummaryRowModel(context, currencyFormatter)
 
         assertThat(model.number).isEqualTo("#1001")
-        assertThat(model.date).isEqualTo("2026-05-01T10:00:00Z")
+        assertThat(model.date).matches("^May [12]$")
```
staskus, line 32: Claude flagged if the system under test uses a different locale then it could fail.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/cards/AssistantCardEntry.kt
```diff
@@ -0,0 +1,11 @@
+package com.woocommerce.android.aiassistant.ui.cards
+
+data class AssistantCardEntry(
```
staskus: > 1. AssistantCardEntry and AssistantCardKey accidentally public. Other :feature types (ShowCardsUiStructured, parser classes) are correctly internal. :WooCommerce consumes only AssistantCard via the host renderer, never the Entry/Key envelope. Both should be internal — otherwise this becomes part of the cross-module API by accident.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/runtime/AgenticLoopAssistantRuntime.kt
```diff
@@ -85,21 +90,42 @@ internal class AgenticLoopAssistantRuntime @Inject constructor(
...
-                    AssistantRuntimeEvent.ToolCallFinished(
-                        toolCallId = event.result.toolCallId,
-                    )
-                )
+                is LoopEvent.ToolCallStarted -> {
+                    emit(event.toRuntimeEvent(toolNamesById))
+                }
+                is LoopEvent.ToolCallFinished -> {
+                    val toolName = toolNamesById.remove(event.result.toolCallId)
+                    event.result.toRuntimeEvents(toolName).forEach { emit(it) }
```
staskus, line 101: Could also be removed on `LoopEvent.Failed`

staskus (approved):
✅ Works well!

Found some things I should do on iOS as well.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/cards/AssistantCardEntry.kt
hichamboushaba: Good catch.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/runtime/AgenticLoopAssistantRuntime.kt
hichamboushaba: Addressed as defensive cleanup. This does not change cross-turn behavior in practice because `toolNamesById` is scoped to a single `runTurn` flow and is discarded when the turn completes; clearing it on `Failed` just drops stale in-flight tool state within that turn.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aiassistant/WooAssistantCardRendererTest.kt
hichamboushaba: Fixed in 57b2c83

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/cards/AssistantCardEntry.kt
hichamboushaba: Fixed in 57b2c83

PR #15822, opened 2026-05-05
[AI Support Chat] Add support chat repository and bookmark persistence

iamgabrielma:
Closes WOOMOB-2952. Stacked on https://github.com/woocommerce/woocommerce-android/pull/15815 , which needs to be reviewed/merged first.

### Description

Adds the non-UI support chat repository layer and lightweight Room persistence for chat bookmarks. I'm assuming the 6k diff is expected when doing a DB migration due to `schema 83.json`?

### Changes
- Add SupportChatRepository for sending, fetching, registering, touching, loading, and deleting support chats.
- Add SupportChatBookmarkEntity and DAO in libs:fluxc-plugin.
- Scope bookmarks by selected local site ID and store remote site/user metadata.
- Bump WC Room DB version from 82 to 83 with auto migration.
- Add Room schema 83.json.
- Add DAO, migration, and repository unit coverage.

### Test Steps
- Synced Gradle successfully.
- Ran the app on simulator and confirmed it compiles/launches.
- ./gradlew :libs:fluxc-plugin:testDebugUnitTest --tests "*SupportChatBookmarkDaoTest*"
- ./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*SupportChatRepositoryTest*"
- ./gradlew :WooCommerce:assembleWasabiDebug

cc: @itsmeichigo


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    suspend fun fetchChat(botSlug: String, chatId: Long): Result<SupportChatResponse> = withContext(dispatchers.io) {
+        restClient.fetchChat(botSlug = botSlug, chatId = chatId).toResult()
+    }
+
+    suspend fun registerChat(chatId: Long, botSlug: String, firstUserMessage: String): Unit = withContext(dispatchers.io) {
+        val selectedSiteModel = selectedSite.get()
+        val now = System.currentTimeMillis()
+        bookmarkDao.insertIgnore(
```
malinajirka: ❓ `insertIgnore` returns `-1L` on conflict and we throw the value away here, so a re-`registerChat` for an existing `chatId` silently drops the new title/timestamps. Whether that's intentional (idempotent registration) or a bug (should bump `updatedAt` to keep recents accurate) isn't obvious from the call site. If it's intentional, renaming to `registerChatIfAbsent` would make it self-documenting.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepositoryTest.kt
```diff
@@ -0,0 +1,239 @@
...
+            botSlug = BOT_SLUG,
+            message = MESSAGE,
+            context = CONTEXT,
+            chatId = CHAT_ID
+        )
+
+        assertThat(result.getOrNull()).isEqualTo(response)
+        verify(restClient).sendFollowUpMessage(BOT_SLUG, CHAT_ID, MESSAGE)
+        verify(restClient, never()).sendMessage(any(), any(), any())
+    }
```
malinajirka, line 88: 🔍 We have an error-path test for `fetchChat` but not for `sendMessage` / `sendFollowUpMessage` — given that `sendMessage` is the most-used method in a multi-turn session, adding `given failed send response, when sending new message, then SupportChatRepositoryException carries error type and message` (and the follow-up symmetric variant) would catch a future regression in `toResult()` for the path that matters most.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
```diff
@@ -0,0 +1,113 @@
...
+import kotlinx.coroutines.withContext
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response
+import org.wordpress.android.fluxc.persistence.dao.SupportChatBookmarkDao
+import org.wordpress.android.fluxc.persistence.entity.SupportChatBookmarkEntity
+import org.wordpress.android.fluxc.store.AccountStore
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
malinajirka: 🔍 Seems like this class is stateless, so there isn't any value in having it as a singleton, it just increases memory fingerprint. I'd consider replacing this with `@Reusable`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
```diff
@@ -0,0 +1,113 @@
...
+        response.toResult()
+    }
+
+    suspend fun fetchChat(botSlug: String, chatId: Long): Result<SupportChatResponse> = withContext(dispatchers.io) {
+        restClient.fetchChat(botSlug = botSlug, chatId = chatId).toResult()
+    }
+
+    suspend fun registerChat(chatId: Long, botSlug: String, firstUserMessage: String): Unit = withContext(dispatchers.io) {
+        val selectedSiteModel = selectedSite.get()
+        val now = System.currentTimeMillis()
```
malinajirka: 💡  Calling `System.currentTimeMillis()` directly makes the class tricky to test - I'd consider injecting `CurrentTimeProvider` and getting current time from there.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SupportChatBookmarkDao.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.persistence.entity.SupportChatBookmarkEntity
+
+@Dao
+interface SupportChatBookmarkDao {
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
+    suspend fun insertIgnore(bookmark: SupportChatBookmarkEntity): Long
+
+    @Query("UPDATE SupportChatBookmarkEntity SET updatedAt = :updatedAt WHERE chatId = :chatId")
+    suspend fun touch(chatId: Long, updatedAt: Long): Int
```
malinajirka: ❓ I'm not familiar with this terminology - is it a common name or something created here?

PRD-Android-AI-Chat.md
```diff
@@ -0,0 +1,561 @@
+# PRD — Android AI-Powered Troubleshooting (Chat with Support)
```
malinajirka: ❓ Was this file committed on purpose or by accident? Both is fine, just want to double check.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SupportChatBookmarkDao.kt
```diff
@@ -0,0 +1,26 @@
...
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.persistence.entity.SupportChatBookmarkEntity
+
+@Dao
+interface SupportChatBookmarkDao {
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
```
malinajirka: ❓ It is a bit unusual to ignore inserts for a single entity (non-collection). Unless there is a specific reason for it, I'd personally opt for REPLACE. Wdyt?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SupportChatBookmarkEntity.kt
```diff
@@ -0,0 +1,17 @@
...
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+
+@Entity(tableName = "SupportChatBookmarkEntity")
+data class SupportChatBookmarkEntity(
+    @PrimaryKey val chatId: Long,
+    val localSiteId: LocalId,
+    val remoteSiteId: Long,
+    val wpcomUserId: Long,
```
malinajirka: 🗨️ I don't have the full context, but tieing this feature to wpcom users only on the data layer feels a bit risky to me. What if we ever decide to give access to any user, including self-hosted users. I mean, it'd make a lot of sense for us as a business, since Woo isn't .com specific and AI support is likely cheaper than human support. 

Moreover, since only one .com user can be logged into the app, why do we need to store their ID with the chatbookmark?

Wdyt?

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepositoryTest.kt
```diff
@@ -0,0 +1,239 @@
...
+        val bookmarkCaptor = argumentCaptor<SupportChatBookmarkEntity>()
+        val firstMessage = "  abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ  "
+
+        repository.registerChat(
+            chatId = CHAT_ID,
+            botSlug = BOT_SLUG,
+            firstUserMessage = firstMessage
+        )
+
+        verify(bookmarkDao).insertIgnore(bookmarkCaptor.capture())
```
malinajirka: 🔍 The test verifies different things than the title suggests.

malinajirka (approved):
Thanks for working on this @iamgabrielma, looks great overall! I have left some suggestions, none of them is blocking.

P.S. Just to double check, the migration file was generated not written by AI from scratch, right?

iamgabrielma:
> P.S. Just to double check, the migration file was generated not written by AI from scratch, right?

That's a good question, I'm pretty sure that these were generated and not written, otherwise would have raised a flag when it did, but I'm not 100% sure either. 

If I re-run `./gradlew :libs:fluxc-plugin:kspDebugKotlin --rerun-tasks` locally `83.json` stay unchanged, so the committed schema matches Room’s generated output. I assume that should be enough to confirm it was generated?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SupportChatBookmarkEntity.kt
iamgabrielma: I think at the moment we should keep this, it might change later, but from what we've seen in iOS so far we cannot continue odie chats unless we have an authenticated WPCOM, as this is needed in the POST request to continue the chat session, meaning most likely unauthenticated users won’t have this feature (will be able to see previous conversations but not continue them). 

Persisting the user ID lets us scope saved chat bookmarks to the account that can actually resume them, if we just rely on site ID we could be fetching different chats from different users for the same site ID.

PRD-Android-AI-Chat.md
iamgabrielma: It started purposefully to make it easier to share work, but after a talk with my counterpart they won't do Android, so the PRD is no longer needed. Removed on bbd64a61e4d4bda0776f0a655372210c35f05c3a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
iamgabrielma: Updated! 70797b7ac312236f9d52c4978cb001b7e8beb9e2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
iamgabrielma: Injected and replaced with `currentTimeProvider.currentDate().time` on f6c8488cbd06fc762280ddff066ab66c0078c005

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepositoryTest.kt
iamgabrielma: Added additional tests on 5a96591039630d2b8e77cbd0a2834fad4e6e9e5e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepository.kt
iamgabrielma: Done! https://github.com/woocommerce/woocommerce-android/commit/f6c8488cbd06fc762280ddff066ab66c0078c005

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SupportChatBookmarkDao.kt
iamgabrielma: It sounds good to me, changed on 6f0a3687a8ea337e06d07a4e9224851edbd0b625

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/aisupportchat/SupportChatRepositoryTest.kt
iamgabrielma: 👀 77429160ad75f7159f84c5795ca30db8a17c9995

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SupportChatBookmarkEntity.kt
malinajirka: The app doesn't support having more than one account in the app - we wipe the database on logout. 

Moreover, localSiteId already guarantees it's just for the current user - if a new user logs in, they won't reuse the same local site id.

I think with this logic, all entities in the whole database should have a userId to ensure they are not used by a different user.

Having said all that, if you think it's better, I don't mind keeping it.

malinajirka:
> If I re-run ./gradlew :libs:fluxc-plugin:kspDebugKotlin --rerun-tasks locally 83.json stay unchanged, so the committed schema matches Room’s generated output. I assume that should be enough to confirm it was generated?

If you delete the file before running it -> the newly generated file has identical content -> then yes, that's enough to verify. I have checked it and looks good 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/SupportChatBookmarkDao.kt
iamgabrielma: I've seen it used when you need to update timestamps, to avoid the awkard `updateUpdateAt`, but we can use others. Renamed in 45395b78e97cb4894010759e81027ffe19c9b02d

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SupportChatBookmarkEntity.kt
iamgabrielma: That makes sense, removed on https://github.com/woocommerce/woocommerce-android/pull/15822/commits/c615b906660c0ed61368c0ec896194f14052dd88

iamgabrielma:
Holding this one for a bit, might need re-review later.
- Bumped schema from 83 to 84 for now, since 83 already exists in trunk and was conflicting.
- Commit history should be cleaned up once I merge networking, which is the parent. Will also address any further conflicts then.

PR #15826, opened 2026-05-05
[WOOMOB-2985] Polish assistant chat and rich cards

hichamboushaba:
### Description
Part of WOOMOB-2985

This PR is the first part of the Android AI Assistant UI polish work. It focuses on making streamed replies, rich cards, confirmation cards, and the composer behave more consistently with the iOS implementation. There are still additional polish pieces to add in follow-up work.

#### What changes for users

The assistant chat now feels more stable while a turn is streaming. Messages stay top-aligned instead of visually recentering as the transcript grows, assistant bubbles keep a steadier width while tokens stream in, and autoscroll only follows new output when the user is already pinned near the latest response. If the user scrolls up during a long streamed reply, the transcript no longer force-snaps back to the bottom.

Rich order and product cards now render as one grouped assistant card surface, with a localized group header and dividers between rows. The previous expansion-style footer controls are intentionally not shown, so the card group reads like a compact result set rather than a paginated list.

#### `show_cards` display ordering

The UI now owns rich-card display order instead of relying on prompt timing or raw tool-event arrival order:

- `show_cards` card groups are hidden while the active assistant message is still streaming.
- Once the assistant turn is no longer streaming, prose and non-card segments render before card groups when the message has text.
- Card-only assistant turns still show cards once the turn finishes.
- Confirmation cards are not treated as normal rich cards, because their placement carries user-action semantics.
- The `show_cards` tool activity pill is suppressed when rich cards are the visible output.

This avoids final transcripts like `cards → text` or `text → cards → text` when the model emits cards and prose in different event phases.

#### Confirmation card polish

Confirmation cards now better match the iOS confirmation-card treatment:

- Pending, confirmed, and cancelled states use the updated inline card styling.
- Resolved confirmation cards persist in the transcript after confirm/cancel.
- Diff rows show the label with a trailing colon, prior values with strikethrough, and new values with emphasized text.
- Long values, including billing emails, wrap naturally without field-specific email or length heuristics.
- Bulk confirmations omit prior values when there is no reliable per-item prior state.
- Confirmation flow behavior remains unchanged: confirm/cancel still dispatch `ConfirmationResult` and conversation cancel/reset paths resolve pending confirmations cleanly.

#### Composer polish

The assistant composer has been rebuilt as an iOS-style input bar while preserving the existing behavior contract:

- Pill-shaped input container using Material/Woo theme tokens.
- Circular icon send and stop controls with accessibility labels.
- Send remains enabled only for non-blank input while no turn is active.
- IME send still goes through the same guarded send path.
- Stop appears only while a turn can be cancelled.
- Multiline input growth is bounded.
- Awaiting-confirmation state can show a short pending-change hint.

#### Compose previews

This PR adds or expands previews for the assistant components touched by the polish work, including chat text/card states, composer states, in-flight dots, and confirmation-card variants. The previews use realistic assistant sample states so future UI changes have a more useful local visual target.

All production changes stay in `:libs:ai-assistant:feature`. There are no `:WooCommerce` host changes in this PR, no `:libs:ai-assistant:core` dependency-direction changes, no new auth path, no Jetpack AI JWT changes, and no Gson usage for AI payloads.

### Test Steps
1. Open the AI Assistant entry point in the app.
2. Ask for recent orders or products and verify the assistant reply remains top-aligned while streaming.
3. Verify rich cards render in one grouped card container below the assistant prose, without Show all / View all / +N footer controls.
4. Scroll up while a response is streaming and verify the thread does not force-scroll back to the latest token.
5. Ask for a write action that requires confirmation, then confirm and cancel separate attempts; verify confirmation cards keep their resolved state in the transcript.
6. Type in the composer and verify send, stop, multiline, and pending-confirmation states behave correctly.

### Images/gif

https://github.com/user-attachments/assets/b28fcbf3-b3c6-4180-afe6-1b0e059b2214

### DIFF
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/safety/WooCommerceConfirmationPreviewBuilder.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/safety/WooCommerceConfirmationPreviewBuilder.kt
@@ -59,24 +59,15 @@ internal class WooCommerceConfirmationPreviewBuilder @Inject constructor() {
             }
         }
 
-        if (status != null) {
-            val message = if (status in CUSTOMER_NOTIFYING_STATUSES) {
-                string(
-                    R.string.ai_assistant_confirmation_order_set_status_emails_customer,
-                    raw(id.toString()),
-                    raw(status),
-                )
-            } else {
-                string(R.string.ai_assistant_confirmation_order_set_status, raw(id.toString()), raw(status))
-            }
-            return ConfirmationPreview(message = message, fields = fields)
-        }
-
         return ConfirmationPreview(
             message = string(
                 R.string.ai_assistant_confirmation_order_update_summary,
                 raw(id.toString()),
-                fields.toChangeSummary(),
+                fields.toChangeSummary(
+                    statusEmailImpact = status
+                        ?.takeIf { it in CUSTOMER_NOTIFYING_STATUSES }
+                        ?.let { R.string.ai_assistant_confirmation_change_summary_status_emails_customer },
+                ),
             ),
             fields = fields,
         )
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
@@ -352,8 +352,16 @@ private fun LazyListState.isPinnedToRenderedEnd(
     if (lastVisibleItem.index != targetIndex) return false
 
     val distanceFromViewportEnd = layoutInfo.viewportEndOffset - (lastVisibleItem.offset + lastVisibleItem.size)
-    return distanceFromViewportEnd <= bottomPinThresholdPx
+    return isRenderedTargetPinnedToViewportEnd(
+        distanceFromViewportEnd = distanceFromViewportEnd,
+        bottomPinThresholdPx = bottomPinThresholdPx,
+    )
 }
+
+internal fun isRenderedTargetPinnedToViewportEnd(
+    distanceFromViewportEnd: Int,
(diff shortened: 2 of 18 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
✅ I tested and it works very wel! 
 
One observation to fix:
⚠️  No back button from the products view

Difference from iOS:
- You list products and variations together.  As on iOS, this interaction is not perfect. The agent doesn't always understand that it needs to update variations or vice-versa. Updates variations one by one, not bulk.

I like:
- More user friendly tool action descriptions compared to iOS
- Feel stable and solid, no weird glitches

Will keep testing more

hichamboushaba:
> You list products and variations together

Actually we don't, but in Woo Core, a product variation is still a product (in the sense that `/wc/v3/products/<id>` will return it). For now, the Android app doesn't have a specific card for variations, something that's still missing, and I'm planning to add it if possible.

> ⚠️ No back button from the products view

We show a Close button, that's the default behavior of the screen, the back button is shown only when it's shown from the product list. do you feel this is a major issue that requires changing the screen logic? if yes, I can address it in a separate PR @staskus

staskus:
> We show a Close button, that's the default behavior of the screen, the back button is shown only when it's shown from the product list. do you feel this is a major issue that requires changing the screen logic? if yes, I can address it in a separate PR @staskus

@hichamboushaba I don't see any close button, have to press a system back button.

no close

> do you feel this is a major issue that requires changing the screen logic? 

Not sure, unless it's acceptable on Android not having any in-app button to going back.

hichamboushaba:
> I don't see any close button, have to press a system back button.

Oh, you test in a tablet, this might be the cause, let me take a look.

staskus:
@hichamboushaba 

I should probably do it on a phone. A habit that comes from POS.

PR #15829, opened 2026-05-05
[RSM-1626] Fetch default low stock threshold for stock notifications

irfano:
### Description
Fixes RSM-1626

This PR fetches the store-wide default low stock threshold from WooCommerce product settings and uses it in the Stock notification settings screen.

It adds the default low stock threshold to the FluxC product settings model, maps and persists the value, and refreshes it when the Stock notification settings screen resumes so changes made in WC Admin are reflected when returning to the app.

### Test Steps
1. Open the app with a Woo-driven site.
2. Go to `Settings > Notifications > Stock`.
3. Verify the Low stock copy shows the store default threshold.
4. Tap `Edit default threshold` and update the low stock threshold in WC Admin.
5. Return to the app.
6. Verify the Low stock copy updates with the new threshold value.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsFragment.kt
```diff
@@ -36,6 +36,7 @@ class NewStockNotificationSettingsFragment : BaseFragment() {
 
     override fun onResume() {
         super.onResume()
+        viewModel.refreshDefaultLowStockThreshold()
```
AdamGrzybkowski: ❓ Can't we call this from the viewModel's `init` block?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
```diff
@@ -45,12 +58,17 @@ class NewStockNotificationSettingsViewModel @Inject constructor(
...
+            _viewState.update { it.copy(defaultLowStockThreshold = threshold) }
+        }
+    }
+
     data class ViewState(
         val notificationsEnabled: Boolean = true,
         val lowStockNotificationsEnabled: Boolean = true,
         val outOfStockNotificationsEnabled: Boolean = true,
         val backorderNotificationsEnabled: Boolean = true,
-        // Temporary value until the default threshold is fetched from the network.
```
AdamGrzybkowski, line 53: We should remove the default completely and show some kind of shimmer when we don't have the value.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
```diff
@@ -36,6 +39,16 @@ class NewStockNotificationSettingsViewModel @Inject constructor(
         }
     }
 
+    fun refreshDefaultLowStockThreshold() {
+        launch {
+            val productSettings = wooCommerceStore.fetchSiteProductSettings(selectedSite.get()).model
+
+            productSettings?.defaultLowStockThreshold?.let { threshold ->
+                updateDefaultLowStockThreshold(threshold)
+            }
```
AdamGrzybkowski: I think it would be good to handle the error here. We change the copy to one without a number and show a snackbar. WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
```diff
@@ -36,6 +39,16 @@ class NewStockNotificationSettingsViewModel @Inject constructor(
         }
     }
 
+    fun refreshDefaultLowStockThreshold() {
+        launch {
+            val productSettings = wooCommerceStore.fetchSiteProductSettings(selectedSite.get()).model
```
AdamGrzybkowski: We are adding the low stock threshold to the DB, but we are not taking advantage of that here - why? 

This value is something that will not change often. I think it would be better to show the stored value and refresh it in the background.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsFragment.kt
AdamGrzybkowski: This can lead to an excessive number of network requests for a value that is unlikely to change. Could we listen for the webview close event/result to know if we should refresh that value?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
```diff
@@ -45,12 +58,17 @@ class NewStockNotificationSettingsViewModel @Inject constructor(
         )
     }
 
+    private fun updateDefaultLowStockThreshold(threshold: Int) {
+        if (_viewState.value.defaultLowStockThreshold != threshold) {
```
AdamGrzybkowski: This if is unnecessary - MutableStateFlow should handle that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
AdamGrzybkowski: I would suggest observing the value from DB and refreshing only when needed. A shimmer when refreshing and no value visible would be nice, but not necessary.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsFragment.kt
irfano: Good point! Fixed in e5e068a.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: Good point. I removed the hardcoded fallback and now keep the value empty until we have either a cached or fetched threshold. If there’s a cached value, we show it immediately while refreshing in the background; otherwise we show a small inline loading indicator.

I tried a value-only shimmer first, but at this size it looked more like a static gray block than a shimmer, so I switched to the inline spinner. I’ll attach a screenshot as well. Fixed in 480ef7c.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: Good point. I think this is worth handling, but I’d prefer to split it into a follow-up PR to keep this one focused.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: I added the cache for that purpose but forgot to use it. Anyway, I’ve handled it: 480ef7c3dfc4c22e66ba20cb05a786257433ac4e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: Fixed in c9fdd22.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: Opened a follow-up PR for this here: #15845.

AdamGrzybkowski (approved):
Looks good, thanks for the updates!

One thing: can we add a non-breaking space before the threshold value?

Image

PR #15830, opened 2026-05-05
[RSM-1747] Add stock notification channel support

irfano:
### Description

Fixes RSM-1747

Adds Android notification channel support for the upcoming stock notification type.

This adds `STOCK` as a recognized notification kind and maps it to a new `Stock alerts` Android channel. Stock notification taps currently use the existing dashboard fallback rather than introducing a new navigation flow.

A follow-up PR will add in-app warnings for disabled notification channels, so channel availability handling is intentionally not included here.

### Test Steps

1. Open the device notification settings for the app.
2. Verify `Stock alerts` appears as a notification category.

### Images/gif
Screenshot_20260505_173103


WooCommerce/src/main/kotlin/com/woocommerce/android/model/Notification.kt
```diff
@@ -103,6 +109,7 @@ fun NotificationModel.getNoteMessage(resourceProvider: ResourceProvider): String
     return when (this.type) {
         NotificationModel.Kind.STORE_ORDER -> this.getMessageSnippet()
         NotificationModel.Kind.COMMENT -> "${this.getTitleSnippet()}: ${this.getMessageSnippet()}"
+        NotificationModel.Kind.STOCK -> getMessageSnippet() ?: getBodySnippet() ?: getTitleSnippet()
```
AdamGrzybkowski: Curious, but why do we need this fallback here, but not in other types?

irfano: You're right, this was unnecessarily defensive. I removed the fallback and kept stock notifications using the message snippet only.

AdamGrzybkowski (approved):
LGTM :shipit:

PR #15831, opened 2026-05-05
[WOOMOB-2937] Implement AgenticLoop tool replay dedupe

hichamboushaba:
### Description
Fixes WOOMOB-2937

This adds per-turn replay defenses to `AgenticLoopImpl` so repeated tool calls do not burn the loop budget by re-dispatching the same work, and so the model receives successful replay guidance instead of a hard error envelope.

#### Runtime changes
- Adds a package-level `ToolReplayTracker` used by `AgenticLoopImpl` for per-turn replay state.
- Tracks calls for the lifetime of a single user turn only; there is no cross-turn cache or conversation-wide dedupe.
- Builds canonical signatures from `toolName + sorted JSON arguments` using `kotlinx.serialization-json`.
  - JSON object keys are sorted recursively before serialization.
  - Arrays keep their order.
  - This makes `{"a":1,"b":2}` and `{"b":2,"a":1}` resolve to the same signature.
- Caches only the first successful `ToolResult.Success` for a canonical signature.
  - Validation, safety, and transport failures are not cached for replay.
- Replays duplicate cached successes as `ToolResult.Success`, not as an error response.
  - The replayed tool message uses the current model-requested tool call id so history remains aligned with the current assistant tool call.
  - The cached structured payload is preserved and receives an `_assistant_runtime_hint` field.
  - `uiStructured` is preserved through replay.
- Adds the two replay stages required by the ticket:
  - Second identical call: returns cached success with a neutral "use the result above" hint and skips `ToolRegistry` execution.
  - Third and later identical calls: returns cached success with an escalated "finish now" hint and skips `ToolRegistry` execution.
- Adds a same-tool per-turn cap of 4 valid calls regardless of arguments.
  - The fifth same-tool call returns a `ToolResult.ValidationError` nudge before safety evaluation or registry execution.
  - The cap intentionally runs before replay, so a fifth identical call is capped rather than replayed.

#### Scope and architecture
- Changes are contained to `:libs:ai-assistant:core`; `ToolReplayTracker` now lives in its own core loop file instead of as an `inner` class on `AgenticLoopImpl`.
- No host app, feature module, UI, navigation, auth, dependency, or Gradle configuration changes.
- AI payload handling stays on `kotlinx.serialization-json`; no Gson usage was introduced.
- Android keeps its existing normal finish / `show_cards` loop contract. The escalated replay hint uses Android-compatible "finish now" wording rather than introducing an iOS-style terminal `respond` tool.

#### Test coverage
- `AgenticLoopImplTest` covers the ticket-level behavior:
  - reversed-key JSON arguments collapse to the same canonical signature;
  - the first call executes through the registry;
  - the second identical call replays the cached result with the soft hint;
  - the third identical call replays the cached result with the escalated hint;
  - varied-argument same-tool fanout caps on the fifth call;
  - the same tool with different arguments still executes normally below the cap.
- `ToolReplayTrackerTest` was added for the project’s class-to-test coverage check and for two boundary cases:
  - cap wins over replay on the fifth identical call;
  - replay preserves `uiStructured` while adding the model-visible replay hint to `structured`.

#### CI follow-up
- The initial PR passed local verification but Danger required an exact class-name test signal for `ToolReplayTracker`.
- Added `ToolReplayTrackerTest` and renamed the test methods to include the exact helper token expected by the checker.
- Moved `ToolReplayTracker` into its own file after feedback that an `inner` class was unnecessary for this state holder.
- Danger, detekt, lint, manifest diff, Firebase prototype build, wrapper validation, and setup checks were green at last handoff. The longer Buildkite unit and instrumented test jobs were still pending when the watcher was stopped by request.

### Test Steps
No manual app flow is required for this core runtime change. Review the added unit tests to validate the duplicate replay and same-tool cap behavior.

### Images/gif
N/A

### DIFF
--- a/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
+++ b/libs/ai-assistant/core/src/main/kotlin/com/woocommerce/android/aiassistant/core/loop/AgenticLoopImpl.kt
@@ -22,8 +22,12 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.flow
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
 
 class AgenticLoopImpl(
     private val chatService: ChatService,
@@ -47,6 +51,7 @@ class AgenticLoopImpl(
         val newTurnMessages = mutableListOf<AssistantMessage>(initialTurn.currentUserTurn)
         var modelMessages: List<AssistantMessage> = initialTurn.modelMessages
         var visibleOutputStarted = false
+        val replayTracker = ToolReplayTracker(json)
         var iteration = 0
 
         while (iteration < MAX_ITERATIONS) {
@@ -83,7 +88,7 @@ class AgenticLoopImpl(
                 return@flow
             }
 
-            val toolExecution = executeTools(assembledResults, validCalls, toolDescriptors)
+            val toolExecution = executeTools(assembledResults, validCalls, toolDescriptors, replayTracker)
             if (toolExecution is ToolExecutionOutcome.Cancelled) {
                 emitCancelledToolExecution(toolExecution, stream.assistantText, history, newTurnMessages)
                 return@flow
@@ -198,6 +203,7 @@ class AgenticLoopImpl(
         assembledResults: List<ToolCallAssembler.AssemblyResult>,
         validCalls: List<ToolCall>,
         toolDescriptors: List<ToolDescriptor>,
+        replayTracker: ToolReplayTracker,
     ): ToolExecutionOutcome {
         val completedTools = mutableListOf<CompletedToolCall>()
         for (r in assembledResults) {
@@ -209,10 +215,31 @@ class AgenticLoopImpl(
         }
         for (call in validCalls) {
             val descriptor = toolDescriptors.find { it.name == call.name }
-            val result = executeToolIfAllowed(call, descriptor)
-                ?: return ToolExecutionOutcome.Cancelled(completedTools)
-            completedTools += CompletedToolCall(call, result, descriptor?.safetyLevel ?: ToolSafetyLevel.SAFE)
-            emit(LoopEvent.ToolCallFinished(result))
+            when (val replayDecision = replayTracker.prepare(call)) {
+                is ToolReplayDecision.CapExceeded -> {
+                    completedTools += CompletedToolCall(
+                        call,
+                        replayDecision.result,
+                        descriptor?.safetyLevel ?: ToolSafetyLevel.SAFE,
+                    )
+                    emit(LoopEvent.ToolCallFinished(replayDecision.result))
+                }
+                is ToolReplayDecision.Replay -> {
+                    completedTools += CompletedToolCall(
+                        call,
+                        replayDecision.result,
+                        descriptor?.safetyLevel ?: ToolSafetyLevel.SAFE,
+                    )
+                    emit(LoopEvent.ToolCallFinished(replayDecision.result))
+                }
+                is ToolReplayDecision.Execute -> {
+                    val result = executeToolIfAllowed(call, descriptor)
+                        ?: return ToolExecutionOutcome.Cancelled(completedTools)
+                    replayTracker.record(replayDecision.signature, result)
+                    completedTools += CompletedToolCall(call, result, descriptor?.safetyLevel ?: ToolSafetyLevel.SAFE)
+                    emit(LoopEvent.ToolCallFinished(result))
+                }
+            }
         }
         return ToolExecutionOutcome.Completed(completedTools)
     }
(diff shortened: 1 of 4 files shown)


staskus:
[Generated with the help of Claude]

**iOS parity note** — same pattern shipped on iOS in [#17002](https://github.com/woocommerce/woocommerce-ios/pull/17002). We're at parity on the core mechanism (signature, two-tier hint, threshold, intra-batch). One notable asymmetry I caught while comparing:

| | Android (this PR) | iOS today |
|---|---|---|
| Per-tool-name cap (catch "vary args to escape") | ✅ `PER_TOOL_TURN_CALL_LIMIT = 4` → `ValidationError` | ⚠️ Constant `perToolPerTurnCap = 4` is **declared and threaded through init, but never enforced in the dispatch path**. Filed as [WOOMOB-2997](https://linear.app/a8c/issue/WOOMOB-2997). |
| Structured stop markers (`must_respond_now`, `stop_reason`) on 3rd+ replay | hint string only (`_assistant_runtime_hint`) | ✅ extra JSON fields (`must_respond_now: true`, `stop_reason: "duplicate_tool_call"`) beside the hint |

**Why we dropped 3 → 2** (verbatim from inline comment): *"the previous error-shape result (`duplicate_call_blocked`) was read by mini as a retryable failure and just provoked more variations, saturating the 5-iter cap."* — same reasoning probably worth surfacing on Android if you bumped 3 → 2 too.

**Two questions:**

1. **Why is the soft hint just a string on Android?** We added structured `must_respond_now: true` + `stop_reason: "duplicate_tool_call"` on the 3rd+ replay (alongside the hint string) because `gpt-4o-mini` was bargaining on prose-only hints. Smoke regression improved measurably after. Worth porting?

2. **Per-tool cap envelope shape — `ValidationError` vs success-shape?** Android nudges through `ToolResult.ValidationError`. We're planning to wire ours as a success-shape `tool_call_cap_reached` envelope (matching the existing `cached_result_reused` pattern) because of the same "error-shape provokes retries" finding. Curious whether you've observed the cap-result-shape mattering, or if `ValidationError` works fine in your testing.

staskus:
[Generated with the help of Claude]

One more parity observation while comparing the replay paths — possibly a UX glitch worth checking.

**iOS replay path** ([`AgenticLoopOrchestrator.swift` line 320](https://github.com/woocommerce/woocommerce-ios/blob/trunk/Modules/Sources/WooAIAssistant/Loop/AgenticLoopOrchestrator.swift)):

```swift
if priorSeen >= 1, let cachedPayload = liveResults[signature] {
    continuation.yield(.toolCallStarted(...))
    let replay = Self.duplicateReplayJSON(...)
    continuation.yield(.toolCallCompleted(..., resultJSON: replay))
    continue   // bypasses handleToolResult, the only place .cardRender is emitted
}
```

iOS caches only the **model-visible structured payload** (`liveResults: [String: String]`), not the `Success` envelope. So a duplicate `show_cards` call:
- emits the `cached_result_reused` replay envelope to the model ✅
- does **not** re-emit `.cardRender` to the UI

**Android (from this PR's diff):**

```kotlin
ToolReplayDecision.Replay(cached.toReplayResult(...))   // copies the entire ToolResult.Success
```

`Success.copy(...)` preserves `uiStructured`, and your `LoopEvent.ToolCallFinished` bridge (best guess based on iOS-side handling) presumably reads `uiStructured.cards` and surfaces them. If so, a duplicate `show_cards` call **renders the same card twice** in the transcript on Android — once on the primary, once on the replay.

**Question:** is this intentional on Android, or have you observed duplicate cards stacking when `gpt-4o-mini` repeats `show_cards`? If the latter, scoping `toReplayResult` to drop `uiStructured` (or a UI-side dedupe by `card.id`) would match the iOS behavior. Easy to repro: in smoke, prompt `"show me my last order"` twice in the same turn — the model often calls `show_cards(order, id=N)` twice.

staskus (approved):
✅ Approved, left two comments based on small divergencies with iOS

hichamboushaba:
Thanks @staskus for the detailed review.

I ran manual real-model checks with ApiFaker-backed Woo API responses. The replay/cap logic seems to be working well in practice: repeated same-turn calls are handled without loops, duplicate card rendering did not reproduce, and the per-tool cap stopped excessive calls. The test sets that I ran are explained below.

### 1. Duplicate replay: empty order list

Commands:

```bash
cat > /tmp/orders-empty.json <<'EOF'
[]
EOF
adb push /tmp/orders-empty.json /data/local/tmp/orders-empty.json
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT --es api_type "wp-api" --es path "/wc/v3/orders" --es http_method "GET" --ei response_status_code 200 --es response_body_file "/data/local/tmp/orders-empty.json"
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
```

Prompt:

```text
Find processing orders from 2026-05-01 to 2026-05-31. Check the exact same query three times before answering. Do not change the filters between checks. If all three checks match, summarize the result.
```

Expected/result observed: first matching `orders_list` call executes, repeated identical calls are replayed, and the assistant stops with a no-matching-orders answer.

### 2. Duplicate `show_cards` rendering

Commands:

```bash
cat > /tmp/order-1001.json <<'EOF'
{"id":1001,"number":"1001","status":"processing","total":"42.50","currency":"USD","date_created_gmt":"2026-05-01T10:00:00","payment_method_title":"Credit card","billing":{"first_name":"Jane","last_name":"Doe","email":"jane@example.com"},"line_items":[],"shipping_lines":[],"fee_lines":[],"tax_lines":[]}
EOF
adb push /tmp/order-1001.json /data/local/tmp/order-1001.json
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT --es api_type "wp-api" --es path "/wc/v3/orders/1001" --es http_method "GET" --ei response_status_code 200 --es response_body_file "/data/local/tmp/order-1001.json"
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
```

Prompt:

```text
Show me order 1001 as a card. To verify rendering, render the exact same order card twice using the same order reference, then answer briefly.
```

Expected/result observed: only one visible card is rendered for order `1001`; duplicate replay does not duplicate the card UI.

### 3. Per-tool cap: five different order IDs

Commands:

```bash
for id in 1001 1002 1003 1004 1005; do
  cat > /tmp/order-$id.json <<EOF
{"id":$id,"number":"$id","status":"processing","total":"10.00","currency":"USD","date_created_gmt":"2026-05-01T10:00:00","billing":{"first_name":"Jane","last_name":"Doe","email":"jane@example.com"},"line_items":[],"shipping_lines":[],"fee_lines":[],"tax_lines":[]}
EOF
  adb push /tmp/order-$id.json /data/local/tmp/order-$id.json
done
for id in 1001 1002 1003 1004 1005; do
  adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT --es api_type "wp-api" --es path "/wc/v3/orders/$id" --es http_method "GET" --ei response_status_code 200 --es response_body_file "/data/local/tmp/order-$id.json"
done
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
```

Prompt:

```text
Fetch full details separately for order IDs 1001, 1002, 1003, 1004, and 1005. Do not use a bulk/list call. Inspect each order by ID, one at a time, then summarize status and total.
```

Expected/result observed: the first four `orders_get` calls can execute; the fifth same-tool call is stopped by the cap, and the assistant produces a partial useful answer instead of continuing to call tools.

### 4. Retry pressure: failing order

Commands:

```bash
cat > /tmp/order-error.json <<'EOF'
{"code":"fake_error","message":"Temporary fake order failure"}
EOF
adb push /tmp/order-error.json /data/local/tmp/order-error.json
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT --es api_type "wp-api" --es path "/wc/v3/orders/1001" --es http_method "GET" --ei response_status_code 500 --es response_body_file "/data/local/tmp/order-error.json"
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
```

Prompt:

```text
Fetch order 1001. If the fetch fails or looks incomplete, retry the exact same fetch until you are confident, then tell me what happened.
```

Expected/result observed: failures are not cached as replay successes, but the per-tool cap still prevents an endless retry loop. The assistant stops and explains that the order could not be fetched.

### 5. Analytics duplicate replay

Commands:

```bash
cat > /tmp/revenue-stats.json <<'EOF'
{"totals":{"orders_count":3,"total_sales":"123.45","net_revenue":"100.00","items_sold":5},"intervals":[{"interval":"2026-05-01","subtotals":{"orders_count":3,"total_sales":"123.45","net_revenue":"100.00","items_sold":5}}]}
EOF
adb push /tmp/revenue-stats.json /data/local/tmp/revenue-stats.json
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT --es api_type "wp-api" --es path "/wc-analytics/reports/revenue/stats" --es http_method "GET" --ei response_status_code 200 --es response_body_file "/data/local/tmp/revenue-stats.json"
adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
```

Prompt:

```text
Compare revenue from 2026-05-01 to 2026-05-07. Run the exact same revenue query once for the initial result, once as verification, and once as final confirmation. Then answer only if all three match.
```

Expected/result observed: first `analytics_revenue` call executes; repeated identical analytics calls replay from cache, and the assistant summarizes the matching result and stops.

------ 
If you want this re-checked independently, you can ask an agent to run the same manual scenarios with the `/verify-on-device` skill against a debug build and ApiFaker enabled, hopefully it'll be able to figure it out.

hichamboushaba:
I'm going to merge this now, and we can always align later when needed.

PR #15832, opened 2026-05-06
[WOOMOB-2973] Preserve deterministic AI write failures

hichamboushaba:
### Description
Fixes WOOMOB-2973

The AI assistant previously treated every unsafe write tool transport error as an unknown outcome after the merchant confirmed the action. That was too broad: for some failures, the app can tell that the write did not happen, so the user should see the regular tool-failed state instead of being asked to verify whether the outcome happened.

This PR adds a small internal distinction to `ToolResult.TransportError` through a binary `ToolFailureKind`:

- `OUTCOME_UNKNOWN` remains the default, preserving existing behavior for ambiguous failures and for tools that do not classify their errors yet.
- `DETERMINISTIC_FAILURE` marks stable failure cases where the app can tell the write did not succeed.

`AgenticLoopImpl` now uses that distinction only for confirmed unsafe tool failures. Unknown outcomes still emit `AssistantError.OutcomeUnknown`; deterministic failures emit the existing `AssistantError.ToolFailed`. Model-facing tool JSON remains the generic `{"error":"Tool execution failed"}`, so raw FluxC/Woo exception details are not exposed to the model or UI.

The PR classifies deterministic failures for two single-write tools:

- `orders_update` maps stable order update failures such as invalid IDs, invalid parameters, missing order status, empty billing email, and the local "order not found" result to `DETERMINISTIC_FAILURE`. Generic/ambiguous failures keep `OUTCOME_UNKNOWN`.
- `products_update` maps product-not-found, invalid product ID, invalid parameters, duplicate SKU, invalid min/max quantity, and matching Woo API validation/ID failures to `DETERMINISTIC_FAILURE`. Generic/ambiguous failures keep `OUTCOME_UNKNOWN`.

`AIProductsDataSource` now throws a typed `ProductNotFoundException` when fetching a product succeeds but the product remains absent locally. That gives `products_update` a stable not-found signal without relying on raw error text.

The change intentionally keeps the scope narrow. Bulk/partial-write tools continue to use the default unknown-outcome behavior because their partial success semantics need separate handling. No UI, auth, telemetry, retry, diagnostics, or broad error-model changes are included.

### Test Steps
N/A - this is an internal AI assistant error-classification change covered by focused unit tests and module test suites.

### Images/gif
N/A

### DIFF
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/orders/OrdersUpdateToolHandler.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/orders/OrdersUpdateToolHandler.kt
@@ -1,8 +1,10 @@
 package com.woocommerce.android.aiassistant.tools.orders
 
+import com.woocommerce.android.OnChangedException
 import com.woocommerce.android.aiassistant.core.chat.AssistantToolHandler
 import com.woocommerce.android.aiassistant.core.chat.ToolCall
 import com.woocommerce.android.aiassistant.core.chat.ToolDescriptor
+import com.woocommerce.android.aiassistant.core.chat.ToolFailureKind
 import com.woocommerce.android.aiassistant.core.chat.ToolResult
 import com.woocommerce.android.aiassistant.core.chat.ToolSafetyLevel
 import com.woocommerce.android.aiassistant.core.chat.inputSchema
@@ -13,6 +15,7 @@ import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.encodeToJsonElement
+import org.wordpress.android.fluxc.store.WCOrderStore
 import javax.inject.Inject
 
 internal class OrdersUpdateToolHandler @Inject constructor(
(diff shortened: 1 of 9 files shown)


libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/orders/OrdersUpdateToolHandler.kt
```diff
@@ -80,3 +89,27 @@ internal class OrdersUpdateToolHandler @Inject constructor(
...
+            ToolFailureKind.DETERMINISTIC_FAILURE
+        } else {
+            ToolFailureKind.OUTCOME_UNKNOWN
+        }
+        else -> ToolFailureKind.OUTCOME_UNKNOWN
+    }
+}
+
+private fun WCOrderStore.OrderError.isLocalMissingOrder(): Boolean =
+    message == "Order not found" ||
```
staskus, line 112: Should we match by a strong type / code and not a string?

staskus (approved):
✅ 

Left one comment.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/orders/OrdersUpdateToolHandler.kt
hichamboushaba: We already do using `WCOrderStore.OrderErrorType.INVALID_ID` which is mapped from the network error `woocommerce_rest_shop_order_invalid_id`, this is just a defensive last check. TBH I'm not sure it's worth it, but IMO it doesn't hurt.

PR #15833, opened 2026-05-06
[RSM] Phone POS TTP - Part 7: Payment-failed screen padding fix

samiuelson:
> Stacked on Part 6 (#15825). Re-target as parents merge.

### Description

Phone-layout fix for the shared `WooPosPaymentFailedScreen` (used by external readers and Tap to Pay). The outer `Column` previously had no horizontal padding, so on phone the title/subtitle hugged the screen edges. Adds breakpoint-aware horizontal padding (`Large` on phone, `XLarge` on tablet) and centers the title/subtitle via `fillMaxWidth()` + `TextAlign.Center`. Tablet spacing is preserved.

Split out from the larger Tap to Pay payment-flow PR so it can land independently.

### Stack

Part 7 of the [RSM] Phone POS TTP stack:
- Part 1 — Multi-method payment buttons row (#15838)
- Part 2 — Phone back button on POS checkout (#15839)
- Part 3 — Feature flag and TTP availability tracking (#15840)
- Part 4 — Tap to Pay promoted state in checkout (#15841)
- Part 5 — Tap to Pay in payment row and overflow dialog (#15842)
- Part 6 — Cash + Other payment methods row (#15825)
- **Part 7 — this PR**
- Part 8a — TTP reader connector foundation (#15850)
- Part 8b — Wire payment collection (#15834)
- Part 9 — Polish state transitions (#15835)

### Build setup

To test on phone, build the `wasabiRelease` variant and temporarily flip these flags in `FeatureFlag.kt`:

```kotlin
WOO_POS_PHONE("woo_pos_phone", localValue = true),
WOO_POS_TAP_TO_PAY("woo_pos_tap_to_pay", localValue = true),
```

Without these overrides the POS phone layout and Tap to Pay aren't reachable on a release-style build.

### Test Steps

1. On phone (or POS in phone breakpoint), open POS, add an item, tap **Checkout**.
2. Trigger a payment failure (decline test card or kill the network mid-collect) so the **Payment failed** screen appears.
3. Verify the title/subtitle no longer hug the screen edges and are centered horizontally.
4. Repeat on tablet to confirm the existing layout is unchanged.

- [ ] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15825, opened 2026-05-05
[RSM] Phone POS TTP - Part 6: Cash + Other payment methods row

samiuelson:
> Stacked on Part 5 (#15842). Re-target as parents merge.

### Description

Replaces the form-factor-aware payment row from Parts 1–5 with a fixed two-button layout: Cash payment + Other payment methods (both outlined). The overflow dialog from Part 5 is repurposed to host the "other" methods. On phones the dialog is rendered as a fullscreen sheet with a back button.

`WooPosPaymentMethod` is rebuilt around the dialog's options: `CARD_READER`, `TAP_TO_PAY`, `SCAN_TO_PAY`, `MARK_ORDER_AS_PAID`. Cash is wired directly from its dedicated button and is no longer an enum member. The dialog's contents are derived from reader status and TTP availability:

- TTP available, reader disconnected: **Card reader**, Scan to pay, Mark order as paid. Card reader routes to the existing connect-reader flow.
- TTP available, reader connected: **Tap to Pay**, Scan to pay, Mark order as paid.
- TTP unavailable: Scan to pay, Mark order as paid only.

`Scan to pay` and `Mark order as paid` are placeholders that close the dialog without further action.

### Stack

Part 6 of the [RSM] Phone POS TTP stack:
- Part 1 — Multi-method payment buttons row (#15838)
- Part 2 — Phone back button on POS checkout (#15839)
- Part 3 — Feature flag and TTP availability tracking (#15840)
- Part 4 — Tap to Pay promoted state in checkout (#15841)
- Part 5 — Tap to Pay in payment row and overflow dialog (#15842)
- **Part 6 — this PR**
- Part 7 — Payment-failed screen padding fix (#15833)
- Part 8a — TTP reader connector foundation (#15850)
- Part 8b — Wire payment collection (#15834)
- Part 9 — Polish state transitions (#15835)

### Build setup

To test on phone, build the `wasabiRelease` variant and temporarily flip these flags in `FeatureFlag.kt`:

```kotlin
WOO_POS_PHONE("woo_pos_phone", localValue = true),
WOO_POS_TAP_TO_PAY("woo_pos_tap_to_pay", localValue = true),
```

Without these overrides the POS phone layout and Tap to Pay aren't reachable on a release-style build.

### Test Steps

1. Open POS, add an item, tap Checkout.
2. Verify the bottom shows Cash payment + Other payment methods.
3. Tap Cash payment — existing cash flow runs.
[...]

PR #15834, opened 2026-05-06
[RSM] Phone POS TTP - Part 8b: Wire payment collection

samiuelson:
> Stacked on Part 8a (#15850) → Part 7 (#15833) → Part 6 (#15825). Re-target as parents merge.

### Description

Wires the POS checkout's `OnTapToPayClicked` to a working built-in card-reader payment flow, on top of the connector foundation introduced by Part 8a (#15850). Tapping **Pay with Tap to pay** now requests `ACCESS_FINE_LOCATION` if needed, drives `WooPosBuiltInReaderConnector.connect()` for `TapToPayDevice`, and starts a `BUILT_IN` payment via the existing `CardReaderPaymentController`. While the SDK's full-screen overlay is up, the in-app reader-status / payment-state observers are silenced; on the SDK's terminal states we route to **Payment success** (success), or a toast with the SDK reason (failure / pre-overlay errors like missing permission, no reader, location lookup failed).

Key technical decisions:
- `ACCESS_FINE_LOCATION` is checked before discovery; on missing permission the ViewModel emits `RequestFineLocationPermission`, the screen launches the system prompt via `rememberLauncherForActivityResult`, and the result is fed back through `OnFineLocationPermissionResult`.
- The BuiltIn reader is explicitly disconnected on every TTP terminal state so the next checkout starts with the TTP CTA again instead of auto-collecting from a still-connected reader.
- A `paymentButtonsState` (derived from `isTapToPayInProgress`) on `Checkout` disables Cash + Other while a TTP payment is in flight, preventing parallel payments.
- `isTapToPayPayment` is persisted via the generalized `TTPPaymentProgressDelegate` (now keyed) so the in-flight flag survives process death — observers correctly silence themselves on resume.
- The double-tap race is closed by flipping the in-flight flag synchronously before the launched coroutine. A `launchTapToPayConnect(orderId)` helper unifies the entry points used by `OnTapToPayClicked` and `OnFineLocationPermissionResult(granted = true)`.
- `listenToPaymentState` dispatches at the top level: `if (isTapToPayPayment) handleTapToPayPaymentState else handleExternalReaderPaymentState`, so per-branch silencing guards aren't needed and adding new states can't accidentally regress TTP.
- `BuiltInReaderDiscoveryFailedException` propagates a nullable message so the VM can correctly differentiate "SDK gave a reason" from "no reader available" and pick the right toast string.

### Stack

Part 8b of the [RSM] Phone POS TTP stack:
- **Part 8b — this PR** — wire payment collection

### Build setup

```kotlin
```

### Test Steps

1. With `WOO_POS_TAP_TO_PAY` enabled and on a TTP-eligible phone, open POS, add an item, tap **Checkout**.
2. With **no card reader connected**, verify the **Tap to pay** promoted state is shown.
3. Tap **Pay with Tap to pay**:
   - First time: system **Allow location** prompt should appear. Allow it. Stripe SDK overlay should follow.
   - Decline location: toast "Grant location permission to use Tap to Pay" appears, TTP CTA remains clickable.
4. With permission granted: tap **Pay with Tap to pay**, verify SDK overlay; tap a contactless card; on success verify the **Payment success** screen.
5. Trigger an SDK failure (test fraud-decline card or cancel from the overlay) → toast with the SDK reason; checkout state restored, TTP CTA clickable for retry.
6. Repeat with airplane mode on → "no internet" toast and the SDK overlay does not appear.
7. Open **Other payment methods** while a card reader is connected, choose **Tap to Pay**, verify the same flow runs.
8. After a successful TTP payment, start a new transaction. Verify the new checkout shows the TTP CTA again (no auto-collect).
9. While **Pay with Tap to pay** is loading (before the Stripe overlay appears), confirm the **Cash payment** and **Other payment methods** buttons are disabled.
10. **Process death**: with the Stripe SDK overlay active, kill the app via Android Studio's "Stop" or `adb shell am kill`, then reopen. The flag survives — no rogue auto-collect.

11. **Cancel during pre-overlay connect window**: with airplane mode off but a slow connection (e.g. throttled to 2G), tap **Pay with Tap to pay**, then immediately back out of checkout before the Stripe overlay appears. Confirm there is no rogue payment auto-collected on the next checkout entry, and the TTP CTA returns cleanly.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

toupper (approved):
Big PR but well-structured. Top-level dispatch in `listenToPaymentState` (`if (isTapToPayPayment) handleTapToPayPaymentState else handleExternalReaderPaymentState`) is the right call — keeps each branch readable and means new states can't silently regress TTP. The double-tap race fix (sync flip of `isTapToPayPayment` before launching) is correct, the `launchTapToPayConnect` helper unifies the two entry points cleanly, and the auto-collect block on built-in readers from previous sessions is a nice catch with a good comment. Test coverage is thorough across the connect/permission/in-flight/fail/success matrix.

A few questions, none blocking:

**1. `buildTotalsReaderNotConnectedError()` is misnamed in the new built-in-reader branch.** When a built-in reader is already connected from a prior transaction we deliberately mask it as `Disconnected` to gate auto-collect, but the helper name says "NotConnectedError" — there's nothing erroneous about it, it's an intentional UX gate. Reads as if something went wrong. Worth either splitting into two helpers (`Disconnected` vs an explicit `buildBuiltInReaderHiddenState`) or renaming the existing one to something neutral.

**2. Three TTP flags is a lot.** `isTapToPayPayment` (persisted), `isTapToPayInProgress` (UI Checkout state), and the existing `isTTPPaymentInProgress` (controller factory hook) all coexist with overlapping semantics. The relationships are correct as far as I can tell, but a one-line comment on each flag's responsibility — or even consolidating two of them — would save the next reader from tracing all three across the file. The TTP-specific naming on a generalized `TTPPaymentProgressDelegate` (now `key`-parameterized) also obscures that the delegate is just a SavedStateHandle-backed Boolean.

**3. `cancelPaymentAction` doesn't reset `isTapToPayInProgress`.** The current call sites (line 162, plus the retry/cleanup paths outside this diff) either run when `isTapToPayPayment` is already false (early-return at line 156) or are followed by a state transition out of `Checkout`, so the leak isn't currently observable. But the SDK-driven failure path goes through `handleTapToPayPaymentState` → `resetTapToPayProgress()`, and `cancelPaymentAction` doesn't — divergence is asking for a future bug if a new cancel path lands. Worth either calling `resetTapToPayProgress()` from `cancelPaymentAction` or commenting why it doesn't apply.

**4. Spinner state during the connect phase.** `TapToPayConnecting` shows while we're calling `builtInReaderConnector.connect()`. If the user backs out or the screen rebuilds before the SDK overlay appears, the coroutine keeps running and the UI flag stays `true`. SDK-cancel handles itself once the SDK overlay is up, but the pre-overlay window may need a thought — at minimum the test plan's "process death" step covers the persisted flag, but a cancel-during-connect manual test would be reassuring.

samiuelson:
Thanks! Addressed in b26ea27e867:

1. **`buildTotalsReaderNotConnectedError` rename** — now `buildReaderDisconnectedState`. Drops the misleading "Error" suffix; reads correctly at all three call sites (genuine disconnected state, deliberate built-in-reader masking, and the totals-build path).

2. **Flag-purpose comments** — added one-liners on each of `isTTPPaymentInProgress` (controller hook for SDK-overlay-up survival across process death), `isTapToPayPayment` (VM-level flow flag), and `isTapToPayInProgress` (UI rendering hint). Skipped the `TTPPaymentProgressDelegate` rename — it has another caller in `WooPosCardPaymentViewModel` and the cascade would touch the connector PR as well; happy to do it as a follow-up after the stack lands if you'd like.

3. **`cancelPaymentAction` divergence** — now calls `resetTapToPayProgress()` instead of manually resetting only `isTapToPayPayment`, so the UI flag is also cleared on cancel paths. Aligns with the SDK-driven failure path.

4. **Cancel-during-connect manual test** — added as test step 11 in the PR description.

I'll skip flag consolidation as you noted — too risky for this PR. Filed in my head as a follow-up if the multi-flag pattern keeps growing.

PR #15835, opened 2026-05-06
[RSM] Phone POS TTP - Part 9: Polish state transitions

samiuelson:
> Stacked on Part 8 (#15834) → Part 7 (#15833) → Part 6 (#15825). Re-target as parents merge.

### Description

UX polish on top of Part 8's coarse loading + toast-only failure handling. Three changes:

1. **Two-phase TTP loading.** `Checkout.isTapToPayInProgress: Boolean` becomes a `TapToPayProgress` sealed type with `Preparing` (location lookup + SDK init + reader connect — shows "Preparing Tap to Pay") and `SdkActive` (Stripe overlay is up or just dismissed — shows a bare spinner so the gap before `PaymentInProgress` doesn't read as "preparing again").
2. **Reuse `PaymentFailed` for SDK failures.** `BuiltInReaderFailedPayment` now renders the standard `PaymentFailed` screen via `buildBuiltInPaymentFailedState`, with retry routed through the SDK's `onRetry` if offered, otherwise a clean rebuild back to checkout. Cancels (`PaymentFlowError.Canceled`) silently return to checkout — they aren't failures.
3. **In-app `PaymentInProgress` on `PaymentCapturing`** for TTP, so the gap between Stripe's overlay closing and `PaymentSuccessful` has a consistent state instead of flashing back to the checkout.

### Stack

Part 9 of the [RSM] Phone POS TTP stack:
- Part 8 — Wire payment collection (#15834)
- **Part 9 — this PR**

### Build setup

```kotlin
```

### Test Steps

1. With `WOO_POS_TAP_TO_PAY` enabled on a TTP-eligible phone, open POS, add an item, **Checkout** → **Pay with Tap to pay**. Verify the "Preparing Tap to Pay" title shows during reader discovery, then collapses to a bare spinner once the Stripe overlay surfaces.
2. Tap a contactless card. After the green checkmark, the in-app "Processing payment" state shows briefly, then **Payment success**. No flash of "Preparing Tap to Pay".
3. Use a fraud-decline card. Verify the standard **Payment failed** screen renders with the SDK reason and a working **Try payment again**.
4. Start TTP, then tap **Cancel** in the Stripe overlay → returns silently to checkout (no Payment failed screen).
5. Confirm the totals table doesn't jump between TTP CTA / Preparing / SdkActive states.


toupper (approved):
Nice polish layer on top of Part 8. Each of the three changes pulls its weight:

- The two-phase `TapToPayProgress` (`Preparing` / `SdkActive`) cleanly distinguishes "we're getting ready" from "the SDK overlay is active," and the `paymentButtonsState` becoming a derived getter is a small but welcome cleanup.
- Routing `BuiltInReaderFailedPayment` through the standard `PaymentFailed` screen with `buildBuiltInPaymentFailedState` (mirroring `buildPaymentFailedState`) is the right consolidation, and silently returning to checkout on `PaymentFlowError.Canceled` is the correct UX — cancels aren't failures.
- `ensurePaymentInProgressStateForTapToPay()` on `PaymentCapturing` closes the "Stripe overlay just dismissed but PaymentSuccessful hasn't arrived yet" gap.

Tests cover the new transitions: LoadingData → SdkActive, PaymentCapturing → PaymentInProgress, Canceled → silent return, retry-with-SDK-callback vs retry-without. Solid.

Two small notes, neither blocking:

1. **Three TTP flags is still three flags.** Part 8 introduced `isTapToPayPayment` / `isTTPPaymentInProgress` / the now-derived `paymentButtonsState`; Part 9 adds `tapToPayProgress`. The relationships still aren't obvious from a cold read — would benefit from a comment or a lightweight unifying type.

2. **`returnToCheckoutAfterTapToPayFailed()` re-fetches the order via `totalsRepository.getOrderById(...)`.** If that hits the network, the cancel-back transition could feel laggy. If it's a cache lookup in practice, fine — just worth eyeballing on a slow connection.

samiuelson:
Thanks for the review!

1. **Three TTP flags & added `tapToPayProgress`** — addressed in two places. The flag-purpose comments landed on `isTTPPaymentInProgress`, `isTapToPayPayment`, and (the now-replaced) `isTapToPayInProgress` in Part 8b's response (which just cascaded into this branch). For Part 9 specifically, I added a one-line comment on the new `tapToPayProgress` field in `WooPosTotalsViewState.Checkout` clarifying that `Preparing` is the pre-overlay phase and `SdkActive` is set once the Stripe overlay surfaces. Holding off on a unifying type for now — the flags cover related-but-distinct concerns (process-death-survivable VM mode vs UI rendering hint) and consolidating them risks more than it saves.

2. **`returnToCheckoutAfterTapToPayFailed()` re-fetching the order** — checked: `totalsRepository.getOrderById` reads from the local FluxC `OrderStore` via `getOrderByIdAndSite(...)` on `Dispatchers.IO`. It's a Room cache lookup, not a network call. The cancel-back transition will be instant on any device that's already shown the order on the checkout screen (which is always the case at this code path).

PR #15836, opened 2026-05-06
Route POS payment-success back press to Products home

samiuelson:
## Summary

System back on the POS payment-success screen now lands the user on Products home, matching the **New Order** affordance. Previously back fell through to the parent `BackHandler` and dropped the user onto an empty cart, which was confusing post-payment.

## Implementation

- `WooPosTotalsPaymentSuccessScreen` installs a `BackHandler` that fires a new `onBackPressed` callback wired to `WooPosTotalsUIEvent.OnBackClicked`.
- `WooPosTotalsViewModel.handleBackPress` adds a `PaymentSuccess` branch that emits `ChildToParentEvent.OnNewTransactionStarted`; the home VM already resets `screenPositionState` to `Products`.
- `CreateNewOrderTapped` analytics are intentionally **not** fired on back — that event represents an explicit button tap.

## Test Steps

1. Open POS, add an item, checkout, complete a payment (cash, card, or Tap to Pay).
2. On the success screen, press system back → land on Products home with an empty cart.
3. Repeat and tap **New Order** instead → same destination, and `CreateNewOrderTapped` analytics fire.
4. Press system back from Products → exit-confirmation dialog still appears as before.


toupper (approved):
Tight, well-scoped change. The `PaymentSuccess` branch in `handleBackPress` cleanly mirrors the `OnNewTransactionClicked` handler minus the analytics, and the home VM already maps `OnNewTransactionStarted` to a Products screen-position reset, so the back path piggybacks on existing behavior without introducing a new event. Tests cover both invariants (event sent, `CreateNewOrderTapped` not tracked).

One small observation, not a blocker — the event name `OnNewTransactionStarted` reads as user-intent ("user wants a new transaction"), but a back press is more of a navigation intent. Functionally identical because the home VM only uses it to flip `screenPositionState` to `Products`, but a future reader might wonder why pressing back fires a "transaction started" event. If the event ever grows side-effects beyond the screen-position reset, the back-press path would inherit them. Keeping a mental note that today these two intents collapse onto one handler is fine — easy enough to split into a dedicated event later if needed.

Manual sanity check worth adding to the test plan: pressing back *during* the success animation (the BackHandler is registered as soon as the screen composes, before the checkmark animation completes) — should be fine but worth eyeballing once.

PR #15837, opened 2026-05-06
[WOOMOB-2994] Implement analytics stats card for AI Assistant

hichamboushaba:
### Description
Fixes WOOMOB-2994

This PR adds the Android AI Assistant analytics stats card. It extends the existing rich-card pipeline so analytics results are produced through `show_cards`, then rendered as a native assistant card with Total Sales, Net Sales, and compact interval charts.

#### What changes for users

When the assistant is asked for store analytics, it can now return a rich Analytics card in the conversation instead of only prose. The card shows the requested period, Total Sales, Net Sales, and two compact trend charts built from the revenue interval subtotals.

Tapping the card opens Analytics for the same date range. Date-only ranges are converted using the selected site's timezone so the Analytics screen lands on the same range the assistant card represents.

If chart data is missing for one or both metrics, the card shows the unavailable trend state rather than an empty graph.

#### `show_cards` analytics flow

Analytics cards now follow the same card-producing tool path as order and product cards:

- `analytics_revenue` remains the data-fetching tool and returns the shared analytics payload shape: `totals` and `interval_subtotals`.
- `show_cards` now supports an `analytics_stats` family.
- The assistant requests analytics cards with an ID-only reference. It does not copy `totals`, `interval_subtotals`, or other analytics result data into the `show_cards` call.
- Analytics IDs use a deterministic format: `analytics_revenue:after:<YYYY-MM-DD>:before:<YYYY-MM-DD>:interval:<hour|day|week|month|year>:currency:<ISO|none>`.
- `ValidatedRef` stays generic and ID-only: `index`, `family`, and `id`.
- `ShowCardsReferenceValidator` validates analytics refs by parsing the synthetic ID with `AnalyticsStatsCardId.parse(...)`.
- `ShowCardsResolver` parses the ID again at resolution time, refetches the stats through `AIAnalyticsDataSource.fetchRevenueStats(...)`, and builds typed `ShowCardDetails.AnalyticsStats` / `AnalyticsStatsSummary` payloads from the fresh result.

This keeps the model-facing `show_cards` contract defensive and avoids adding analytics-specific payload fields to the validated reference model.

#### Analytics payload and card parsing

`analytics_revenue` intentionally does not emit Android-only chart arrays. The chart data is derived on the card side from the same `interval_subtotals` payload that the tool already returns.

The stats card parser maps:

- Total Sales from `total_sales`, falling back to `gross_sales` when needed.
- Net Sales from `net_revenue`.
- Total Sales chart points from each interval's `total_sales` / `gross_sales` value.
- Net Sales chart points from each interval's `net_revenue` value.
- Chart dates from interval keys, with `date_start` used as a fallback where the interval label is not directly parseable.

For `currency:none` IDs, the resolver still refetches without a currency override, but fills the card display metadata from the selected site's currency when available. That keeps the API request semantics unchanged while avoiding raw decimal display in the card.

#### Native assistant UI

This adds a new assistant-local stats card surface:

- `AssistantCard.Stats` for the parsed render model.
- `AiAssistantStatsCard` and `AiAssistantStatsCardState` for the Compose card UI.
- `AssistantStatsTrendChart` for the compact chart primitive.
- Dedicated stats card preview states for multi-point, changed-shape, single-point, all-zero, negative/refund, no-trend, and partial-data scenarios.
- `AssistantCardAction.OpenAnalytics(after, before)` plus host navigation into Analytics for the represented range.

The chart is a lightweight Compose `Canvas` component local to `:libs:ai-assistant:feature`. It is intentionally compact: no axes, tooltips, date controls, report links, or Analytics Hub parent callbacks.

#### Why this does not reuse the Analytics Hub card view

I considered reusing the existing Analytics Hub revenue card, including embedding it with `AndroidView`, but this PR does not do that.

The reusable-looking pieces in Analytics Hub are not a clean assistant component boundary today:

- `AnalyticsHubInformationCardView` is a `MaterialCardView` + view-binding view designed for the Analytics Hub `RecyclerView` adapter, with its own card chrome, loading/no-data states, report-link behavior, and `SeeReportClickListener` callback model.
- Reusing it inside the assistant would either nest an Analytics Hub card inside the assistant card chrome, or require special casing the assistant renderer around that host-owned view.
- The Hub card is built around Analytics Hub screen semantics. The assistant card needs a smaller transcript-friendly result card that is produced by `show_cards`, keyed by assistant card IDs, and tapped as a shortcut into Analytics.

So this PR reuses the Analytics data concepts and destination, not the Analytics Hub view implementation. The assistant owns a small native Compose chart/card, while tapping the card still opens the full Analytics Hub for detailed exploration.

#### Renderer and host integration

The host renderer is split into dedicated order, product, and stats renderers so `WooAssistantCardRenderer` does not keep growing as more assistant card families are added. Host-specific work stays in `:WooCommerce`: formatting money through `CurrencyFormatter`, adapting card actions to navigation, and opening Analytics with a `StatsTimeRangeSelection`.

All assistant payload parsing, card models, runtime routing, `show_cards` validation/resolution, and card UI stay in `:libs:ai-assistant:feature`.

There are no new auth paths, no Jetpack AI JWT changes, no Gson usage for AI payloads, and no dependency-direction reversal. AI payload handling remains on `kotlinx.serialization-json`.

The direct `analytics_revenue` → stats-card side effect has been removed. Analytics cards are now emitted only from successful `show_cards` `uiStructured` output.

### Test Steps
1. Build and install the Wasabi debug app from this branch on a store with Analytics data.
2. Open the AI Assistant and ask for analytics over a recent date range, for example: `Show me analytics for the last 7 days`.
3. Verify the assistant reply includes a rich Analytics card in the transcript.
4. Verify the card shows the requested period, Total Sales, Net Sales, and two compact trend charts.
5. Tap the Analytics card and verify the app opens Analytics for the same date range.
6. Ask for a range with sparse or unavailable trend data and verify the card shows the unavailable trend state instead of a blank or broken chart.

### Images/gif
Screenshot_20260506_125638


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
✅ Nice! 

I tested it. It generally works. The biggest improvement can come for nudging system prompt to interpret the dates better.

⚠️  "This week" is treated as last 7 days, not as this week from the beginning of the week depending on the calendar. You can see how performance card on the dashboard interprets this week.

Image

⚠️ Empty space in the message

I had this issue as well, had to trim down. 

Image

⚠️ Expected analytics path don't show the card

Check iOS implementation. You may need to improve tool calls or the prompt for it to display a card for such queries as well

Image

Image

🤖  Claude feedback:

```
1. interval_subtotals leaks to the model via structured
ShowCardsToolHandler.kt:181-188 adds interval_subtotals to ANALYTICS_STATS_SUMMARY_KEYS, which means the chart bucket array flows back to the model in structured (model-visible). With MAX_ANALYTICS_INTERVALS = 100 per range, every analytics show_cards returns the full subtotals payload twice per turn (once via analytics_revenue, again via show_cards).

The architectural rule from earlier reviews: chart arrays go in uiStructured.cards[].details, not structured. The data IS in ShowCardDetails.AnalyticsStats already, so dropping it from the summary keys closes the leak without losing UI capability.

Note from the deep parity: iOS [#17079](https://github.com/woocommerce/woocommerce-android/issues/17079) has the same leak (AnalyticsStatsSummary.swift:13-19). Cross-platform contract issue, but worth fixing on the Android side first since this PR is the trigger.

2. AnalyticsStatsCardId.parse() has no dedicated unit test
The parser is the trust boundary between an LLM-controlled string and AIAnalyticsDataSource.fetchRevenueStats(after, before, interval, currency). Current coverage is indirect through 3 cases in ShowCardsToolHandlerTest + 1 in ShowCardsResolverTest.

iOS already has AnalyticsCardSpecTests.swift with parameterized cases. Worth adding AnalyticsStatsCardIdTest.kt covering:

Length cap, wrong section count, wrong labels at any position
Bad date formats (2026/01/01, ISO with time, not-a-date)
Invalid intervals (decade, Hour mixed-case)
Invalid currencies (special chars, lowercase, empty between colons, 4-letter)
after > before range invariant
Round-trip: parse(toSyntheticId(parsed)) == parsed (would also surface the dead toSyntheticId() helper)
The domain side proves the parser fails safely (returns null → validator emits InvalidId), but the safety property isn't pinned by tests.

3. analytics_orders cards aren't supported despite the tool existing
AnalyticsStatsCardId.kt:63: private const val ANALYTICS_STATS_TOOL = "analytics_revenue" — hardcoded. So:

AnalyticsOrdersToolHandler exists, model can call it
Model gets analytics_orders results in prose
But if it tries to render those as a card via show_cards(family=analytics_stats, id=analytics_orders:after:...), validator rejects with InvalidId
System prompt Pattern 6 only documents the revenue path
iOS [#17079](https://github.com/woocommerce/woocommerce-android/issues/17079) supports both via AnalyticsKind { revenue, orders } (AnalyticsCardSpec.swift:106-108).
```

hichamboushaba:
Thanks @staskus for the review, taking a look now.

For:

>⚠️ Empty space in the message
>
> I had this issue as well, had to trim down.

Can you explain what you mean?

hichamboushaba:
Thanks @staskus for the thorough review, I tried to address the issue, I ported the date anchors you use in iOS.

For the `show_cards` hints, I went with a slightly different approach, I updated the `analytics_revenue` definition to make it clear that `show_cards` has to be called after it : https://github.com/woocommerce/woocommerce-android/blob/94045acac50f018af404fd9d3ec6a423442b6512/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/tools/analytics/AnalyticsRevenueToolHandler.kt#L24-L28

I prefer this approach because it doesn't couple the prompt to the supported show_cards, let's discuss it tomorrow.

> 3. analytics_orders cards aren't supported despite the tool existing
> AnalyticsStatsCardId.kt:63: private const val ANALYTICS_STATS_TOOL = "analytics_revenue" — hardcoded. So:
> 
> AnalyticsOrdersToolHandler exists, model can call it
> Model gets analytics_orders results in prose
> But if it tries to render those as a card via show_cards(family=analytics_stats, id=analytics_orders:after:...), validator rejects with InvalidId
> System prompt Pattern 6 only documents the revenue path
> iOS [#17079](https://github.com/woocommerce/woocommerce-android/issues/17079) supports both via AnalyticsKind { revenue, orders } (AnalyticsCardSpec.swift:106-108). 

Regarding this point from Claude, as it's a bigger change, I'm deferring it to another PR, I logged an issue about. I didn't know we are having two analytics cards.

staskus:
> Can you explain what you mean?

@hichamboushaba 

image

> For the show_cards hints, I went with a slightly different approach, I updated the analytics_revenue definition to make it clear that show_cards has to be called after it :

That's a good idea 👍 I have some of it in some tool prompts but not all. I'll align.

staskus:
> I updated the analytics_revenue definition to make it clear that show_cards has to be called after it 

@hichamboushaba the possible drawback here is with future MCP tools what wouldn't have this guidance, so in the MCP scenario system prompt pointing to show_cards would make more sense. However, we don't have to think about it this very moment and can iterate.

hichamboushaba:
@staskus for the empty space in the screenshot, I'm not sure what happened there, could the model sent empty characters?
Anyway, I'll see if it happens more, but I've never seen it in my tests.

PR #15838, opened 2026-05-06
[RSM] Phone POS TTP - Part 1: Multi-method payment buttons row

samiuelson:
### Description

Refactors the bottom of the POS checkout: replaces the single Cash payment button (and the fullscreen "Reader disconnected" CTA) with `WooPosCheckoutPaymentButtons`, a form-factor-aware vertical stack driven by a list of `WooPosPaymentMethod` enum values. The first method renders as a primary button, the rest as outlined. When the reader is disconnected, both Card reader and Cash render; otherwise only Cash. The new component is structured to accept Tap to Pay and an overflow dialog in later parts without further refactoring.

### Note on disconnected-state UX

The fullscreen `ReaderDisconnected` info card (icon + title + subtitle + CTA) is removed in this PR; the upper area's `Disconnected` branch is intentionally rendered as `Unit`. Until later parts cascade, a disconnected reader shows nothing on top — only the Card reader + Cash buttons at the bottom. The full info card is restored by Part 4 (#15841) for the TTP-available path and Part 8b (#15834) for the TTP-unavailable path. During the merge window between this PR and Part 8b, the disconnected screen looks bare; that is intentional, not a regression.

### Stack

Part 1 of the [RSM] Phone POS TTP stack:
- **Part 1 — this PR**

### Test Steps

1. Open POS, add an item, tap **Checkout**.
2. With **no reader connected**: verify two stacked buttons appear — primary **Card reader** and outlined **Cash payment**. Tap **Card reader** and confirm the connect-reader flow opens.
3. Connect a reader and return to checkout: verify only **Cash payment** is shown and the cash flow runs on tap.
4. Repeat on phone and tablet to verify horizontal padding for each form factor.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCheckoutPaymentButtons.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCheckoutPaymentButtons.kt
@@ -1,0 +1,47 @@
+package com.woocommerce.android.ui.woopos.home.totals
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.navigationBarsPadding
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.res.stringResource
+import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosOutlinedButton
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosBreakpoint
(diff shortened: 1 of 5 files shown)


toupper (approved):
Clean foundation for the stack. Lifting Cash + Card reader into a method enum and a small `WooPosCheckoutPaymentButtons` composable (first method primary, rest outlined) is straightforward, sets up the rest of the parts cleanly, and the layout decision (CARD_READER + CASH when disconnected, CASH when connected) is easy to read in the call site.

One thing worth calling out: the disconnect UX takes a temporary visual hit. The fullscreen `ReaderDisconnected` info card (icon + title + subtitle + CTA) is removed and the upper area's `Disconnected` branch becomes `Unit` — so as of this PR alone, a disconnected reader shows nothing on top, just the Card reader + Cash buttons at the bottom. Part 8b reintroduces `ReaderDisconnected` for the no-TTP path, so the info card returns down the stack, but during the merge window between Part 1 and Part 8b the disconnected screen looks bare. Worth keeping in mind for the merge order — or noting in the PR description so reviewers don't read it as a regression.

Otherwise LGTM.

PR #15839, opened 2026-05-06
[RSM] Phone POS TTP - Part 2: Phone back button on POS checkout

samiuelson:
### Description

Adds a back arrow at the top-start of the POS checkout (totals) screen on phone. The arrow is opt-in via a nullable `onPhoneBack: (() -> Unit)?` parameter on `WooPosTotalsScreen`, propagated to `TotalsLoaded`. When non-null, the screen renders a `WooPosBackButton` with status-bar padding; the host (`WooPosPhoneTotalsScreen`) wires the callback to the existing `WooPosHomeUIEvent.SystemBackClicked` flow, so tapping the arrow returns to the cart with identical behavior to the system back gesture. Tablet doesn't pass `onPhoneBack`, so its layout is unchanged. No new UI event is introduced.

### Stack

Part 2 of the [RSM] Phone POS TTP stack. Independent of Parts 3-9 and can land in any order:
- **Part 2 — this PR**

### Test Steps

1. On phone (or phone breakpoint), open POS and navigate Products → Cart → Checkout.
2. Verify a back arrow appears at top-start of the checkout, with status-bar padding.
3. Tap the arrow and verify it returns to the cart (same as system back).
4. On tablet, verify the back arrow is **not** rendered.


toupper (approved):
Tight, well-scoped change. Reusing the existing `WooPosHomeUIEvent.SystemBackClicked` rather than adding a new event is the right call — the back arrow and system back gesture are the same intent, and a separate event would just duplicate the home VM dispatch.

The opt-in `onPhoneBack: (() -> Unit)?` keeps tablet behavior identical and renders the back-button row on phone with status-bar padding while the rest of `TotalsLoaded` keeps its existing inset handling.

One small note: this overlaps with the back-button code that bled into PR #15833's diff (which added a separate `PhoneBackFromCheckoutClicked` event). The cleaner version is here — once one of the two merges, the other will need a quick rebase to drop the duplicate. Worth coordinating the merge order so the rebase falls on whichever PR you'd rather touch.

Stylistic nit (same one I left on #15833): `onPhoneBack: (() -> Unit)? = null` uses the parameter's presence as a render flag (`if (onPhoneBack != null) { ... }`). A boolean + always-non-null callback would be slightly more idiomatic, but harmless either way. LGTM.

PR #15840, opened 2026-05-06
[RSM] Phone POS TTP - Part 3: Feature flag and TTP availability tracking

samiuelson:
> Stacked on Part 1 (#15838). Re-target `trunk` after the parent PR merges.

### Description

Foundation for the upcoming Phone POS Tap-to-Pay feature. Adds the `WOO_POS_TAP_TO_PAY` feature flag (debug-on by default) and, when the flag is enabled, forwards any `TapToPayAvailabilityStatus.Result.NotAvailable` reason to `PaymentsFlowTracker.trackTapToPayNotAvailableReason(...)` from `WooPosTotalsViewModel.init` with `source = "woo_pos_checkout"`. Tracking fires once per ViewModel construction (i.e. once per checkout entry, including after process death) and is skipped when the flag is off so we don't pollute analytics before the feature ships.

No user-visible change — the TTP UI lands in Part 4.

### Stack

Part 3 of the [RSM] Phone POS TTP stack:
- **Part 3 — this PR**

### Build setup

```kotlin
```

### Test Steps

1. Flag on, TTP-ineligible device (e.g. emulator without NFC): open POS → Checkout, confirm `tap_to_pay_not_available_reason` fires with `source=woo_pos_checkout`.
2. Re-enter checkout (back to cart, then checkout again): confirm the event fires again — once per checkout entry.
3. Flag off, same device: confirm the event is **not** fired.
4. Flag on, TTP-eligible device: confirm the event is **not** fired.
5. Confirm no UI change anywhere.


samiuelson:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

toupper (approved):
Tight, well-scoped foundation PR. The flag is wired correctly (debug-on default), the tracking guard checks the flag before invoking `tapToPayAvailabilityStatus()` so we don't pollute analytics pre-launch, and the test matrix covers all four combinations (flag × availability state). Source string `woo_pos_checkout` is consistent with the rest of the analytics surface.

One thing worth keeping an eye on: the PR description correctly calls out that tracking fires once per ViewModel construction, including after process death — so a user who gets killed mid-checkout and resumes will produce duplicate `tap_to_pay_not_available_reason` events for the same device/reason. Probably acceptable since process death is rare, but if the dashboards start seeing inflated counts, dedup keying on a session marker (or only tracking on a fresh `CheckoutClicked` rather than every VM init) would be the lever. Not blocking.

PR #15841, opened 2026-05-06
[RSM] Phone POS TTP - Part 4: Tap to Pay promoted state in checkout

samiuelson:
> Stacked on Part 3 (#15840) → Part 1 (#15838). Re-target as parents merge.

### Description

Adds the user-visible TTP entry point at the top of the POS checkout. A new `WooPosIsTapToPayAvailable` use case combines the `WOO_POS_TAP_TO_PAY` flag with `TapToPayAvailabilityStatus.isAvailable` and exposes `isFeatureFlagEnabled()` so Part 3's tracking shares one source of truth. `WooPosTotalsViewState.Checkout.isTapToPayAvailable` is set when the Checkout state is built; when true and the reader is disconnected, a new `TapToPayPromoted` composable renders an illustration, "Tap to pay" heading, subtitle, and a primary CTA.

`OnTapToPayClicked` records `checkout_tap_to_pay_payment_tapped` via `WooPosTotalsAnalyticsTracker` — the actual payment collection lands in Part 8. The Part 1 payment-buttons row is unchanged here; Part 5 layers TTP into it.

### Note on disconnected-state UX

This PR fills in the upper-area UI for the **reader-disconnected + TTP-available** path via the new `TapToPayPromoted` composable. The **reader-disconnected + TTP-unavailable** path still renders nothing in the upper area (the `Disconnected` branch was made `Unit` in Part 1 #15838 and stays that way here); the fullscreen `ReaderDisconnected` info card is restored by Part 8b (#15834). During the merge window between this PR and Part 8b, non-TTP users will see only the bottom buttons on a disconnected reader. That is intentional, not a regression.

### Stack

Part 4 of the [RSM] Phone POS TTP stack:
- Part 1 — Multi-method payment buttons row (#15838)
- Part 2 — Phone back button on POS checkout (#15839)
- Part 3 — Feature flag and TTP availability tracking (#15840)
- **Part 4 — this PR**
- Part 5 — Tap to Pay in payment row and overflow dialog (#15842)
- Part 6 — Cash + Other payment methods row (#15825)
- Part 7 — Payment-failed screen padding fix (#15833)
- Part 8a — TTP reader connector foundation (#15850)
- Part 8b — Wire payment collection (#15834)
- Part 9 — Polish state transitions (#15835)

### Build setup

To test on phone, build the `wasabiRelease` variant and temporarily flip these flags in `FeatureFlag.kt`:

```kotlin
WOO_POS_PHONE("woo_pos_phone", localValue = true),
[...]

PR #15842, opened 2026-05-06
[RSM] Phone POS TTP - Part 5: Tap to Pay in payment row and overflow dialog

samiuelson:
> Stacked on Part 4 (#15841) → Part 3 (#15840) → Part 1 (#15838). Re-target as parents merge.

### Description

Adds Tap to Pay to the bottom payment row introduced in Part 1. `WooPosCheckoutPaymentButtons` now derives its layout from `readerStatus` and `isTapToPayAvailable` via `derivePaymentButtonsLayout`, producing one of three shapes: a single button, a primary/secondary pair, or a primary CTA plus an "All payment methods" overflow that opens `WooPosAllPaymentMethodsDialog`. Form-factor preference selects the primary (phone primes TTP, tablet primes Card reader). A new `OnAllPaymentMethodsVisibilityChanged` UI event drives the dialog.

This layout is intentionally short-lived: Part 6 (#15825) replaces it with a fixed two-button row. Shipping it as its own PR keeps the TTP integration reviewable in isolation.

### Stack

Part 5 of the [RSM] Phone POS TTP stack:
- **Part 5 — this PR**

### Build setup

```kotlin
```

### Test Steps

Enable `WOO_POS_TAP_TO_PAY` on a TTP-eligible device, open POS, add an item, tap Checkout:

1. Reader connected + TTP available: bottom shows primary Tap to Pay (phone) / Card reader (tablet), and the other as outlined.
2. Reader disconnected + TTP available: primary CTA + outlined "All payment methods"; tapping it opens a dialog with Card reader, Tap to Pay, Cash.
3. Reader connected + TTP off: only the Cash button is shown.
4. Reader disconnected + TTP off: Card reader + Cash (Part 1 behavior).


toupper (approved):
Solid. `derivePaymentButtonsLayout` is well-factored — pure function, easy to test, and the `Single` / `Pair` / `WithOverflow` sealed type maps cleanly to the actual UI. The form-factor primary preference (phone primes TTP, tablet primes Card reader) reads as intended, and the test matrix in `DerivePaymentButtonsLayoutTest` is exhaustive across the 2×2×2 of formFactor × readerStatus × isTapToPayAvailable.

Couple of small notes, none blocking:

1. **Intentionally short-lived code, as called out in the PR description.** Worth reviewing in its own right because the structural decisions (the layout sealed class, the use case wiring, the dialog) carry forward. Just noting that Part 6 collapses this back to a fixed two-button row, so the form-factor preference logic and `derivePaymentButtonsLayout` won't survive — fine, it bought a clean intermediate step.

2. **`pickPrimary()` has a `?: methods.first()` fallback** for an empty preference list, which is unreachable since `availablePaymentMethods` always includes at least `CASH`. Defensive but dead — not worth removing if Part 6 deletes it anyway.

PR #15843, opened 2026-05-06
[RSM-1747] Add Woo notification preferences endpoints

irfano:
### Description
Adds infrastructure support for the Woo notification preferences endpoint:

`/wp-json/wc/v3/push-notifications/preferences`

This includes:
- The generated endpoint path.
- Network request handling for fetching and updating preferences.
- Local cache support for notification preferences.
- App repository access needed for the upcoming notification settings UI work.

### Test Steps
N/A - infrastructure-only change.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/utils/NullJsonAdapter.kt
```diff
@@ -3,25 +3,28 @@ package org.wordpress.android.fluxc.utils
...
+            else -> throw MalformedJsonException("Unexpected value: $value")
         }
         out.serializeNulls = defaultSerializeNullValue
     }
 
-    override fun read(input: JsonReader): String? {
+    override fun read(input: JsonReader): Any? {
         return when (val token = input.peek()) {
             STRING -> input.nextString()
+            NUMBER -> input.nextString().toBigDecimalOrNull()
```
AdamGrzybkowski: I don't think `BigDecimal` is an obvious type for a `NUMBER` here - should we at least mention it in the docs? I imagine this could lead to a runtime error if used on the wrong type. 

Or maybe having a shared class for that is not a good way forward (given the hidden return types). What about something like this:


```kotlin
abstract class NullableJsonAdapter<T : Any> : TypeAdapter<T>() {
    override fun write(out: JsonWriter, value: T?) {
        val previous = out.serializeNulls
        out.serializeNulls = true
        if (value == null) {
            out.nullValue()
        } else {
            writeValue(out, value)
        }
        out.serializeNulls = previous
    }

    override fun read(input: JsonReader): T? {
        return if (input.peek() == NULL) {
            input.nextNull()
            null
        } else {
            readValue(input)
        }
    }

    protected abstract fun writeValue(out: JsonWriter, value: T)
    protected abstract fun readValue(input: JsonReader): T?
}

class NullStringJsonAdapter : NullableJsonAdapter<String>() {
    override fun writeValue(out: JsonWriter, value: String) {
        out.value(value)
    }

    override fun readValue(input: JsonReader): String? {
        return if (input.peek() == STRING) input.nextString() else throw MalformedJsonException("Expected STRING, got ${input.peek()}")
    }
}

class NullBigDecimalJsonAdapter : NullableJsonAdapter<BigDecimal>() {
    override fun writeValue(out: JsonWriter, value: BigDecimal) {
        out.value(value)
    }

    override fun readValue(input: JsonReader): BigDecimal? {
        return input.nextString().toBigDecimalOrNull()
    }
}
```

We extract some of the common code and keep the return types.

AdamGrzybkowski (commented):
Looks good except for the changes to the `NullableJsonAdapter` - I left my worries there, let me know what you think.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/utils/NullJsonAdapter.kt
irfano: Good point, I applied your suggestion in 4231c08.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/utils/NullableJsonAdapter.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+class NullBigDecimalJsonAdapter : NullableJsonAdapter<BigDecimal>() {
+    override fun writeValue(out: JsonWriter, value: BigDecimal) {
+        out.value(value)
+    }
+
+    override fun readValue(input: JsonReader): BigDecimal? {
+        return input.nextString().toBigDecimalOrNull()
```
AdamGrzybkowski: Would it make sense to be more strict here and also throw `MalformedJsonException` instead of returning null? That would be more consistent. Null is already handled in the base class. 

I know this was in my suggestion, but it was based on the previous code and wasn't a full implementation :D Sorry

irfano: Good point, updated in da06fca9c2ea7898f02b07138fa14fb291b3a442.

AdamGrzybkowski (approved):

PR #15845, opened 2026-05-06
[RSM-1626] Handle stock threshold fetch errors

irfano:
### Description

Follow-up to #15829.

Adds error handling for loading the store default low stock threshold on the stock notification settings screen.

If the fetch fails, the screen now:
- keeps showing the cached threshold when one is available
- stops the inline loading state when no value is available
- shows an error snackbar
- shows the generic “store default” copy without “of <value>” when there is no cached threshold to display

### Test Steps

1. Install a fresh debug build and log in to a Woo store.
2. Configure ApiFaker to return an error for the product settings endpoint:
   ```bash

   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT \
     --es api_type "wp-api" \
     --es path "/wc/v3/settings/products" \
     --es http_method "GET" \
     --ei response_status_code 500

   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
   ```
3. Open the stock notification settings screen.
4. Verify the default low stock threshold does not keep loading indefinitely.
5. Verify an error snackbar is shown.
6. Disable ApiFaker:
   ```bash
   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled false
   ```
7. Reopen the stock notification settings screen and verify the real default low stock threshold is shown.
8. Enable the same ApiFaker error again:
   ```bash
   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.ADD_ENDPOINT \
     --es api_type "wp-api" \
     --es path "/wc/v3/settings/products" \
     --es http_method "GET" \
     --ei response_status_code 500

   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true
   ```
9. Reopen the stock notification settings screen.
10. Verify the cached threshold remains visible and an error snackbar is shown.

### Images/gif

| No cached threshold error| Cached threshold error|
| --- | --- |
| |Screenshot_20260507_011340|


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
```diff
@@ -62,26 +62,51 @@ class NewStockNotificationSettingsViewModel @Inject constructor(
 
             wooCommerceStore.getProductSettings(site)?.defaultLowStockThreshold?.let { threshold ->
                 updateDefaultLowStockThreshold(threshold)
+            } ?: run {
+                if (_viewState.value.defaultLowStockThreshold == null) {
```
AdamGrzybkowski: ❓ We know that the value from DB was null, so the view state value is unlikely to be non-null. Is there a reason why for this check?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2293,6 +2293,8 @@
     <string name="settings_notifs_stock_low_stock_title">Low stock</string>
     <string name="settings_notifs_stock_low_stock_description">When a product variant reaches its low stock threshold.</string>
     <string name="settings_notifs_stock_low_stock_threshold">Products can use their own threshold or the store default of &lt;b&gt;%1$s&lt;/b&gt;. &lt;a href=\'\'&gt;Edit default threshold&lt;/a&gt;</string>
+    <string name="settings_notifs_stock_low_stock_threshold_unavailable">Products can use their own threshold or the store default. &lt;a href=\'\'&gt;Edit default threshold&lt;/a&gt;</string>
```
AdamGrzybkowski: What about something like this - `Products can use their own threshold or the store default. View default threshold to see the current value.`?

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2293,6 +2293,8 @@
...
+    <string name="settings_notifs_stock_low_stock_threshold_error">Unable to load default threshold</string>
```
AdamGrzybkowski: That's not necessarily a default threshold - a user could have modified it, so it's not really a default? At least that's my understanding.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsViewModel.kt
irfano: You're right, we don't need this check. This branch only runs when getProductSettings(site)?.defaultLowStockThreshold is null, so checking the current view state here was redundant. I removed it and also made finishDefaultLowStockThresholdRefresh() just clear the loading state, since it doesn't modify the current threshold value.

WooCommerce/src/main/res/values/strings.xml
irfano: Good suggestion, this reads better for the unavailable state. I updated the copy in cd015f7.

Screenshot_20260508_200859

WooCommerce/src/main/res/values/strings.xml
irfano: You're right, "default" was a bit misleading here since the value can be edited. I switched the user-facing copy to "store-wide threshold" consistently across this section in 7f66ecd. Does that read better to you?

AdamGrzybkowski: I like it, thanks!

AdamGrzybkowski (approved):
Thanks for the changes :shipit:

PR #15846, opened 2026-05-07
[WOOMOB-3027] Reconcile login URL with canonical SiteModel URL after WP-API discovery

AdamGrzybkowski:
Fixes WOOMOB-3027

### Description

When a user logged in to a self-hosted WooCommerce site that 301/302-redirects HTTP→HTTPS, every login attempt could fail in an unrecoverable loop with `INVALID_NONCE`. The attached Linear ticket captured a session with 30+ failed attempts in a row.

Root cause is two URL holders that disagreed and were never reconciled:

- **`LoginSiteCredentialsViewModel.siteAddress`** — captured once from `savedStateHandle` (sourced from `LoginActivity.gotConnectSiteInfo` → `result.urlAfterRedirects`). For the failing user, WP.com's `connect/site-info` endpoint returned `http://…` because of a stale cache entry on its side (most likely). 
- **`SiteStore.SiteModel.url`** — written by FluxC's local WP-API discovery (`SiteWPAPIRestClient.fetchWPAPISite`), which actually probes `/wp-json/`, follows redirects, and resolves the canonical scheme correctly. Every iteration of the failing log shows this returning `https://…` and persisting to Room.

The fix adds a one-shot reconciliation in `fetchSiteForTutorial.onSuccess`: if `SiteModel.url` differs from the current `siteAddress`, adopt it, persist via `SavedStateHandle`, and re-enter `login()`. A `hasReconciledSiteUrl` flag prevents ping-pong if a misbehaving server were to keep flipping. `siteAddress` itself is now a mutable property backed by `SavedStateHandle` so the reconciled value survives process death.

### Test Steps

**Setup — replace `<YOUR_TEST_SITE_URL>` with a hostname you have wp-admin credentials for** (e.g. a `*.jurassic.ninja` sandbox). Run the build on an emulator/device first.

```bash
# 1. Enable apifaker
adb shell am broadcast \
  -a com.woocommerce.android.apifaker.SET_STATUS \
  --ez enabled true \
  com.woocommerce.android.dev

# 2. Register the mock endpoint with an inline http response
adb shell "am broadcast \
  --es api_type wp-com \
  --es path /rest/v1.1/connect/site-info \
  --es http_method GET \
  --es response_body '{\"exists\":true,\"isWordPress\":true,\"hasJetpack\":true,\"isJetpackActive\":false,\"isJetpackConnected\":false,\"isWordPressDotCom\":false,\"isCommerceGarden\":false,\"urlAfterRedirects\":\"http://<YOUR_TEST_SITE_URL>\",\"jetpackVersion\":\"15.7.1\",\"skipRemoteInstall\":false}' \
  com.woocommerce.android.dev"

# 3. Verify (look for 1 endpoint with http://… in the body)
adb shell am broadcast -a com.woocommerce.android.apifaker.LIST_ENDPOINTS com.woocommerce.android.dev
adb logcat -d -s WCApiFaker | tail -30
```

Then in the app:

1. Force-stop the app and re-launch it (so the apifaker DB is read fresh).
2. Tap **Log in**, type any domain, tap **Continue**. (apifaker intercepts `connect/site-info`; the typed value doesn't matter — `urlAfterRedirects` will always be the http URL above.)
3. On the username/password screen, enter valid wp-admin credentials for `<YOUR_TEST_SITE_URL>` and tap **Continue**.
4. Watch logcat:
   ```bash
   adb logcat -s WooCommerce-LOGIN:* WooCommerce-WP:* WordPress-MAIN:*
   ```
5. **Expected**: first attempt logs `Authenticating in to site http://<YOUR_TEST_SITE_URL>` and fails with `INVALID_NONCE`; immediately after, `DB Updating site: https://<YOUR_TEST_SITE_URL>`; then a second `Authenticating in to site https://<YOUR_TEST_SITE_URL>` that succeeds and proceeds to the eligibility/site-picker step. **Without this fix**: the http attempt loops indefinitely — every user retry starts from the http URL.

Negative-control sanity check (no fix needed for this case): use `EDIT_ENDPOINT` (or repeat step 2 with `https://` in the JSON), force-stop, log in again — should succeed first attempt with no retry visible.

**Tear down when done:**

```bash
adb shell am broadcast -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS com.woocommerce.android.dev
adb shell am broadcast -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled false com.woocommerce.android.dev
```

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 24.8
 -----
 - [Internal] Added REMOTE_TAP_TO_PAY feature flag (off in production) for the upcoming Woo POS Remote Tap-to-Pay feature [https://github.com/woocommerce/woocommerce-android/pull/15696]
+- [*] Fixed login getting stuck in a retry loop when an http URL was returned for an HTTPS-only site [https://github.com/woocommerce/woocommerce-android/pull/15846]
 
 24.7
 -----
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
@@ -69,9 +69,24 @@ class LoginSiteCredentialsViewModel @Inject constructor(
         private const val SUCCESS_PARAMETER = "success"
         private const val USERNAME_PARAMETER = "user_login"
         private const val PASSWORD_PARAMETER = "password"
+        private const val HAS_RECONCILED_SITE_URL_KEY = "has-reconciled-site-url"
     }
 
-    private val siteAddress: String = savedStateHandle[SITE_ADDRESS_KEY]!!
+    private var siteAddress: String
+        get() = savedStateHandle[SITE_ADDRESS_KEY]!!
+        set(value) {
+            savedStateHandle[SITE_ADDRESS_KEY] = value
+        }
+
+    // The URL the WP.com `connect/site-info` endpoint reports as the canonical site URL is
+    // sometimes wrong — its cache can return an http URL for a site that 301s to https. Once
+    // FluxC's WP-API discovery resolves the real scheme, we adopt it and re-attempt login
+    // (guarded so a misbehaving server can't trigger a ping-pong update).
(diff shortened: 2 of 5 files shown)


hichamboushaba (approved):
Thanks for addressing this @AdamGrzybkowski

PR #15847, opened 2026-05-07
Fix null Volley error message crashes

irfano:
### Description

Volley can return timeout errors with a null message. Since 24.6, we pass Volley's error message into FluxC network errors, and Kotlin callers that expect a non-null message can crash when a timeout returns null.

This was introduced in [55917b22c50](https://github.com/woocommerce/woocommerce-android/commit/55917b22c502d0e92ff4783c306dcac692e2dffa), which started defaulting to Volley's error message when FluxC does not have a more specific one.

This normalizes null Volley error messages to an empty string. It covers the timeout-only crashes seen while fetching What's New / announcements during app startup and while fetching notification details after receiving a push notification.

### Test Steps

1. Force the What's New / announcements request to time out during app startup.
2. Confirm the app stays open instead of crashing.
3. Receive a push notification and force the follow-up notification detail fetch to time out.
4. Confirm the notification can still be opened and the app does not show a crash dialog when the timeout completes.

### Images/gif

N/A

### DIFF
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -86,7 +86,8 @@ public abstract class BaseRequest<T> extends Request<T> {
         }
 
         public BaseNetworkError(@NonNull GenericErrorType error, @NonNull VolleyError volleyError) {
-            this.message = volleyError.getMessage();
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15850, opened 2026-05-07
[RSM] Phone POS TTP - Part 8a: TTP reader connector foundation

samiuelson:
> Stacked on Part 7 (#15833). Re-target as parents merge.

### Description

Foundational types for the Phone POS Tap-to-Pay (TTP) feature, split out from Part 8 to keep the wiring PR (#15834) focused on the state-machine work.

Adds three pieces with **zero behavior change**:

- **`WooPosBuiltInReaderConnector`** orchestrates a built-in (TTP) card reader connection: permission check → `CardReaderManager` initialization (pinned to `Dispatchers.Main.immediate` per Stripe Terminal SDK requirements) → location lookup → reader discovery (`BuiltInReaders(listOf(TapToPayDevice))`) → connect. Returns `Result<Unit>` so callers can branch on success/failure. Exposes `disconnectIfConnected()` for cleanup. Errors are surfaced via two typed exceptions: `MissingFineLocationPermissionException` and `BuiltInReaderDiscoveryFailedException(message: String?)` — both `internal`.
- **`WooPosFineLocationPermissionCheck`** is a thin testable wrapper around `WooPermissionUtils.hasFineLocationPermission(context)`. `@Reusable`-scoped Hilt binding.
- **`WooPosCardReaderPaymentControllerFactory.create(...)`** gains a `cardReaderType: CardReaderType = CardReaderType.EXTERNAL` parameter so the same controller can drive `BUILT_IN` for TTP. The default keeps every existing call site compiling and behaving identically.

The connector class is not invoked from anywhere in this PR; Part 8 (#15834) wires it from `WooPosTotalsViewModel.startTapToPayPayment(...)`.

### Stack

Part 8a of the [RSM] Phone POS TTP stack:
- **Part 8a — this PR** — TTP reader connector foundation
- Part 8b — Wire payment collection (#15834) — depends on this PR

### Test Steps

This PR has no user-visible behavior change. Verification is via the unit-test suite for `WooPosBuiltInReaderConnector`:

1. Run `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*WooPosBuiltInReaderConnectorTest*"`. All five cases should pass:
   - already-connected reader → returns success without discovering
   - missing fine-location permission → returns `MissingFineLocationPermissionException`
   - location fetch fails → returns failure
   - no built-in reader discovered → returns failure
   - reader discovered + connection succeeds → returns success
2. Verify the existing tests for `WooPosCardReaderPaymentControllerFactory`'s consumers still pass with the new optional parameter (default `EXTERNAL` preserves existing behavior):
   - `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*WooPosTotalsViewModelTest*"`
3. Build the app and confirm the POS flow is unchanged from Part 7's behavior.


toupper (approved):
Solid foundation. `WooPosBuiltInReaderConnector` reads as a clean sequential pipeline (already-connected? → permission → init → location → discover → connect → wait for terminal status), the typed exceptions are `internal` so the VM can pattern-match without leaking implementation, and pinning SDK calls to `Dispatchers.Main.immediate` is the right concession to the Terminal SDK threading model. `WooPosFineLocationPermissionCheck` as a thin `@Reusable` wrapper is a minor but high-leverage testability win. The optional `cardReaderType` parameter default-keeps every existing call site behaving identically.

Two small notes, neither blocking:

1. **`initializeCardReaderManager` is invoked twice per `connect()`** — once at the top, then again inside `discoverFirstBuiltInReader`. The `if (!cardReaderManager.initialized)` guard makes the heavyweight `initialize` a no-op on the second call, but `setupTapToPayUx(...)` runs *outside* that guard, so it gets called twice every connect. If `setupTapToPayUx` is cheap it doesn't matter; if not, hoisting the second `initializeCardReaderManager` call out of `discoverFirstBuiltInReader` (or moving `setupTapToPayUx` inside the guard) would dedupe. Worth a quick eye.

2. **Test gap: `ReadersFound(emptyList())`.** The diff distinguishes `BuiltInDiscoveryResult.NoReaders` (empty list) from `BuiltInDiscoveryResult.Failed(message)` (SDK error), and Part 8b uses that distinction to pick the right toast. The test suite covers `Failed("nope")` but not `ReadersFound(emptyList())`. Quick to add and would lock in the message-vs-no-message branching.

samiuelson:
Thanks for the review! Both points addressed in 7eba7f395f4:

1. **`setupTapToPayUx` double-invocation** — moved inside the init guard (now uses `return@withContext` early-out when already initialized) so it only runs on the first init per session. The redundant `initializeCardReaderManager()` call inside `discoverFirstBuiltInReader` becomes a no-op (skipped early), leaving it for defensive idempotency.

2. **Empty-list test** — added `given discovery emits empty reader list, when connect, then returns failure with no message`. Asserts `BuiltInReaderDiscoveryFailedException` is thrown with a null message, locking in the `NoReaders` vs `Failed(message)` distinction that Part 8b uses for toast selection.

Cascaded to Part 8b (#15834) via merge.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15854, opened 2026-05-07
[WOOMOB-2958] Implement AI Assistant tool parameter expansion

hichamboushaba:
### Description
Fixes WOOMOB-2958

This PR expands the model-facing payloads and query arguments for the existing WooCommerce Android AI Assistant tools. The goal is to give the assistant enough structured store data to answer merchant questions accurately while staying inside the existing tool catalog and runtime architecture.

#### Why this diverges from the initial consensus

The research consensus originally proposed an opt-in `extra_fields` parameter for read tools. That approach had a good token-budget rationale: keep list responses lean by default, then let the model request billing, payment, line-item, or other heavier fields only when needed.

In practice, that made the model choose whether to ask for fields it had not yet seen. The assistant often answered from the default payload instead of predicting that a hidden field existed and making a second, more specific request. That was especially fragile for natural merchant questions like asking for customer email, payment method, billing details, product dimensions, variation attributes, or order notes.

This PR therefore intentionally diverges from that specific part of the consensus: fields that were previously candidates for opt-in expansion are now first-class response fields returned by default. To keep that from becoming an unbounded payload expansion, the implementation still keeps the consensus safety constraints that matter most:

- compact nested projections instead of raw API objects or metadata dumps
- capped long text fields with explicit truncation markers
- allowlisted/validated tool arguments so unsupported model parameters fail loudly
- no new tools, no new card families, and no runtime architecture changes

#### What changes for assistant answers

Read tools now return richer structured payloads by default, so the model can answer from the first tool result instead of guessing which optional fields to request. Write tools also return richer success payloads or compact receipts where the current Android data path supports it.

The system prompt no longer teaches the model to request optional response fields. It continues to trust the dynamic tool catalog as the source of truth and keeps the existing guidance around concise prose, card rendering, confirmation handling, analytics windows, and prompt-injection resistance.

#### Query argument parity

This PR also adds iOS-parity query arguments where the existing Android/FluxC paths support them cleanly:

- `orders_update` now accepts `customer_note` and `billing_email`, and `status` is optional as long as at least one editable field is provided.
- `orders_update` and `orders_bulk_update` no longer allow `refunded`, matching the prompt policy that refunds must be handled from the order screen rather than chat writes.
- `products_list` now accepts `category`, `sku`, `include`, `stock_status`, and supported `orderby` / `order` values.
- Unsupported iOS arguments were intentionally left out rather than exposed as no-ops: `products_list.tag` and `orderby=id|price|rating` do not have clean existing FluxC support.
- `show_cards` customer/variation families and iOS-only variation bulk update remain out of scope.

#### Tool family changes

Orders now return more useful list/detail rows with customer, contact, payment, billing, shipping, capped line-item, coupon, fee, tax, totals, note, and date fields where available. Long `customer_note` values are capped with `customer_note_truncated` so the model knows when it is seeing partial note text. `orders_update` fetches and returns the widened updated-order payload after a successful write, with a partial-response fallback if the follow-up fetch fails. `orders_bulk_update` now returns a compact receipt with requested, updated, failed, partial-success, ID, and patch-key information.

Products and product variations now expose aligned detail, list, update, and bulk receipt payloads. Product responses include stock, pricing, permalink, parent/variation context, dimensions, categories, tags, attributes, images, and capped text fields. Variation responses include product IDs, attributes, image, SKU/price/stock fields, dimensions, dates, and capped descriptions with `description_truncated` markers.

Customers now expose the available customer list fields such as username, dates, role, avatar URL, billing, and shipping fields. The implementation intentionally does not fabricate `orders_count` or `total_spent` because those values are not reliably available from the existing Android customer list data path.

Analytics tools now support `compare_to=previous_period`, echo useful interval/currency metadata, and include `previous_period_totals` when available. If the primary analytics request succeeds but the previous-period comparison request fails, the tools return the primary data with a partial comparison warning instead of discarding the successful result.

`show_cards` summaries are widened for the existing order and product card families, while keeping the structured model summary separate from host-owned UI rendering details. This PR does not add new card families beyond the existing scope.

#### Review fixes included

The final branch includes the high-priority review fixes from whole-branch review:

- `products_list(include=...)` now paginates included IDs, reports `can_load_more`, and rejects oversized include lists.
- Analytics previous-period comparison failures no longer turn a successful primary fetch into a retryable transport error.
- Order update confirmation previews show billing email before/after from the resolved snapshot.
- Customer note confirmations intentionally omit the previous note because it can be long/customer-authored, and instead show a capped preview of the new note.
- Order response `customer_note` fields are capped and marked with `customer_note_truncated` when partial.

#### Scope and safety

This intentionally stays within the existing Android AI Assistant modules:

- No new tools are registered, described, or advertised.
- No optional response-field parameter is exposed in the final tool schemas.
- No `:WooCommerce` host integration, navigation, resources, manifests, or layouts changed.
- No Gradle dependencies, auth paths, Jetpack AI JWT behavior, registry bindings, or DI bindings changed.
- AI payload code continues to use `kotlinx-serialization-json`.
- Existing WooCommerce/FluxC REST payload behavior is left untouched except for assistant-specific query usage through existing supported paths.

The PR also adds catalog and prompt guards to catch accidental out-of-scope tool names, plus focused JVM coverage for the expanded response payloads, validation paths, capped text behavior, update receipts, analytics comparison metadata, card summaries, query argument mapping, confirmation previews, and the high-priority review regressions.

### Test Steps
Use an AI Assistant-enabled test store with recent orders, products, customers, and analytics data. Exact names and totals will depend on the store data, but the assistant should use the available tools/cards and should not fail because it lacks hidden optional fields.

1. Ask: “Show me products of my last order.”
2. Ask: “Who placed my most recent order, and what else have they ordered?”
3. Ask: “How much revenue did I make this week, and how does it compare to last week?”
4. Ask: “Show recent orders, then tell me which one has the biggest discount.”
5. Ask: “Find orders with free shipping and show the customers.”

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
✅ I tested and the model handles these new parameter well. Suddenly much more useful!

I am making a mirror 1:1 PR.

Claude left nit comments:

```
1. PR description mismatch on input validation (OrdersUpdateToolHandler.kt:46-58)
PR description: "capped/validated tool arguments so unsupported model parameters fail loudly"

Reality: customer_note and billing_email aren't length-capped, and email format isn't checked locally. Schema doesn't enforce maxLength either. Merchant confirms the write so safety floor holds, but the description claim doesn't match the code.

From Claude: Small note: the PR description says write-tool args are "capped/validated" but customer_note and billing_email in OrdersUpdateToolHandler.kt:46-58 aren't actually bounded for length and email format isn't checked. Merchant confirmation is the safety floor so this isn't urgent — but worth either correcting the description or adding maxLength to the schema so the model fails loudly instead of submitting a pathologically long note.

2. Truncation marker emission style is inconsistent across tools
Orders: customerNoteTruncated = compactCustomerNote.truncated.takeIf { it } — marker omitted when false (OrderToolResponses.kt:1013, 1045)
Products: descriptionTruncated = compactDescription.truncated — marker always emitted, even false (ProductDetailResponse.kt:1808-1810, ProductVariationDetailResponse.kt:1968)
Both are truthful but the model sees two different shapes for "was this field truncated?" depending on which tool called.

From Claude: Truncation markers emit differently across tools: orders use takeIf { it } so the marker is omitted when false; products always emit the boolean. Both are truthful, but the model sees two different shapes for the same concept across tools. Worth picking one convention — I'd lean toward always-emit since it makes the schema explicit, but either works.
```

PR #15860, opened 2026-05-08
[RSM-1747] Connect New orders settings to notification preferences endpoints

irfano:
### Description
Fixes RSM-1747

This PR connects the New orders notification settings screen to the new WooCommerce notification preferences endpoints.

The screen now loads the saved settings from cache, refreshes them from the API, and saves user changes back to the endpoint. Changes are debounced while the screen is open, and pending changes are saved immediately when leaving the screen so the last edit is not dropped.

Implementation note: if the user has unsaved edits, we do not partially update the screen with newly loaded values. For example, if the user changes the notification switch and the API then returns a different threshold amount, we keep the current screen values instead of updating only the threshold. This keeps the edit flow predictable and avoids mixing fresh server values with in-progress local edits.

### Test Steps
1. Open Settings > Notifications > New orders.
2. Change the New orders notification settings, leave the screen, then open it again.
3. Verify the screen shows the saved values after reopening.
4. Select Only high-value orders, change the threshold, then switch back to All new orders.
5. Leave and reopen the screen, then verify All new orders is still selected and no threshold field is shown.

### Images/gif
N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -6,51 +6,89 @@ import com.woocommerce.android.R
...
+        onBufferOverflow = BufferOverflow.DROP_OLDEST
+    )
+    private var savedOrderPreferences = _viewState.value.toStoreOrderPreferences()
+
     val viewState = _viewState.asLiveData()
 
+    init {
+        observeCachedNotificationPreferences()
+        observeOrderPreferencesChanges()
+        fetchNotificationPreferences()
```
AdamGrzybkowski: I would imagine the push preferences will be refreshed on the main (previous) screen. Do we really need that here? Shouldn't we start with the PN list types first - that screen also needs this data (at least on the designs)?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -86,6 +124,112 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
         refreshNotificationSettings()
     }
 
+    private fun observeCachedNotificationPreferences() {
+        launch {
+            pushNotificationRepository.observeWooNotificationPreferences(site).collect { preferences ->
```
AdamGrzybkowski: What about mapping it to the data we are interested in, in the stream? 

```kotlin
.mapNotNull { it?.storeOrder }
.distinctUntilChanged()
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -6,51 +6,89 @@ import com.woocommerce.android.R
...
     }
 
     fun onThresholdAmountChanged(amount: BigDecimal) {
-        _viewState.update { it.copy(thresholdAmount = amount.coerceAtLeast(MIN_THRESHOLD_AMOUNT)) }
+        updateOrderPreferences(_viewState.value.copy(thresholdAmount = amount.coerceAtLeast(MIN_THRESHOLD_AMOUNT)))
+    }
+
+    fun savePendingOrderPreferences() {
+        val orderPreferences = _viewState.value.toStoreOrderPreferences()
+        appCoroutineScope.launch(coroutineDispatchers.main) {
```
AdamGrzybkowski: ❓ `saveOrderPreferences` uses `appCoroutineScope` already - do we need it here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -86,6 +124,112 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
...
+            }
+        }
+    }
+
+    private fun observeOrderPreferencesChanges() {
+        launch {
+            saveOrderPreferencesTrigger
+                .debounce(ORDER_PREFERENCES_SAVE_DEBOUNCE_MS)
+                .conflate()
+                .collect { saveOrderPreferences(_viewState.value.toStoreOrderPreferences()) }
```
AdamGrzybkowski, line 150: There's a potential scenario where this and `savePendingOrderPreferences()` fire, leading to unnecessary API calls. What about keeping a job reference for that and cancelling it before starting a new one?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -86,6 +124,112 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
...
+        if (updatedViewState == _viewState.value) return
+
+        _viewState.value = updatedViewState
+        saveOrderPreferencesTrigger.tryEmit(Unit)
+    }
+
+    private suspend fun saveOrderPreferences(orderPreferences: StoreOrderPreferences) {
+        if (!hasUnsavedOrderPreferences(orderPreferences)) return
+
+        val result = appCoroutineScope.async(coroutineDispatchers.main) {
```
AdamGrzybkowski: Could you explain the async await usage here? 
Would something like this work here:

```
val result = withContext(NonCancellable + coroutineDispatchers.main) {
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -6,51 +6,89 @@ import com.woocommerce.android.R
...
-            currencySymbol = currencyParameters.currencySymbol.orEmpty(),
+            currencySymbol = parameterRepository.getParameters().currencySymbol.orEmpty(),
             newOrderNotificationSoundStatus = notificationChannelsHandler.checkNewOrderNotificationSound()
         )
     )
+    private val saveOrderPreferencesTrigger = MutableSharedFlow<Unit>(
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST
+    )
+    private var savedOrderPreferences = _viewState.value.toStoreOrderPreferences()
```
AdamGrzybkowski, line 65: I think we should remove any `defaults` related to user prefs from the `viewState` and keep them null until we load actual values from the DB. 

If we do that, this code would likely break, because during the construction, we don't have the cached data yet, and only the cached data should be treated as a `savedOrderPreferences`. 

Am I understanding this correctly?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
AdamGrzybkowski: Whatever we do, I think we should leave a comment for the `saveOrderPreferences` explaining what and why.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -86,6 +124,112 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
...
+                site = site,
+                preferences = WooPushNotificationPreferences(storeOrder = orderPreferences)
+            )
+        }.await()
+        val hasNewerOrderPreferences = _viewState.value.toStoreOrderPreferences() != orderPreferences
+
+        result.onSuccess { savedOrderPreferences = orderPreferences }
+            .onFailure {
+                if (!hasNewerOrderPreferences) {
+                    _viewState.update { savedOrderPreferences.toViewState(it) }
```
AdamGrzybkowski: ⛏️ Personally, I would change the API here a bit and write some code to achieve something like this: 

```kotlin
_viewState.update { it.copyWith(savedOrderPreferences) }
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
```diff
@@ -102,5 +246,6 @@ class NewOrderNotificationSettingsViewModel @Inject constructor(
     companion object {
         private const val DEFAULT_THRESHOLD_AMOUNT = 100
         private val MIN_THRESHOLD_AMOUNT = BigDecimal.ONE
+        private const val ORDER_PREFERENCES_SAVE_DEBOUNCE_MS = 500L
```
AdamGrzybkowski: 500ms is not a lot for a user action (except fast typing), so I would even lean towards a bigger value here to not make too many API requests, especially given that we will call it when the screen closes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: That’s a good point. We can fetch this once on the notification settings screen and reuse it across the sub-screens. Updated in: be6e10712f96be4a9c5df63c4544755a296bbc4a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Good advice. Updated in e03229af20d.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Good catch! Done: 439d8a172e25251787078a3c02563d5d973c1e30

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Good catch, Adam. I pushed 3df5003 for this. Instead of tracking and cancelling the existing job, I made the pending save use the same save flow as the debounced save. That should avoid the duplicate API call.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Good point. I changed this to use `withContext(NonCancellable + coroutineDispatchers.main)` in 13b904e. We don't need `async` here since we're not doing parallel work; we only need the save request to finish if the screen is closed. I also added a short comment to make that clearer.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Yes, that would be a problem if this screen had to wait for the cached prefs after it opens. In the follow-up PR, I’m planning to initialize this screen with the already-loaded order preferences from the parent screen, so the initial saved state will come from real cached data instead of these UI defaults.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: Yes, your suggestion reads more naturally. I made the change in 184f545.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
irfano: That makes sense. 500ms feels more suitable for typing, but this screen also has other preference changes, so I bumped it to 1000ms in 48f9dcb. Does that sound reasonable to you?

irfano:
Thanks for the thorough review, this was really helpful. I think I’ve addressed all the points now. Could you take another look?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewOrderNotificationSettingsViewModel.kt
AdamGrzybkowski: Yes, I think there is still plenty enough and we also do the save on exit so should be more than enough.

AdamGrzybkowski (approved):
Thanks for the changes - LGTM 👍 

One note: Now that we removed the fetch from this screen, there's no fetch anywhere, because the root screen is not yet implemented.

irfano:
> One note: Now that we removed the fetch from this screen, there's no fetch anywhere, because the root screen is not yet implemented.

I’ll handle it in my follow-up PR.

PR #15861, opened 2026-05-10
[WOOMOB-3050] Float assistant composer with gradient bottom fade

hichamboushaba:
### Description
Fixes WOOMOB-3050

Floats the assistant chat composer above the chat content and adds a ChatGPT-style bottom fade so scrolled content recedes behind the input area instead of being abruptly blocked.

This moves the composer and fallback status panel into the `Scaffold` bottom bar, keeps that container transparent, measures the bottom bar content, and draws a bottom-aligned gradient from the main screen content. The chat thread and empty state use the measured composer height plus spacing for bottom padding, so the last visible message/card can still scroll fully above the composer. Sending a message or suggestion now also clears focus to hide the keyboard.

The composer pill keeps its existing internal controls, with elevation applied to the pill itself rather than the bottom bar container.

### Test Steps
1. Open the assistant chat screen and verify the composer appears as a floating pill over the chat surface.
2. Scroll a long assistant response with cards to the bottom and verify content fades behind the composer, including around the pill corners.
3. Verify the last visible card/message can scroll fully above the composer without clipping.
4. Open the empty assistant state and verify suggestions are not obscured by the floating composer.
5. Trigger the fallback error state and verify the error panel floats with the composer above the input.
6. Send a message from the focused composer and verify the keyboard is dismissed.
7. Repeat the visual checks in dark mode.

### Images/gif

https://github.com/user-attachments/assets/55ab9e22-640b-4b9b-8320-3607699adaf8

Screenshot_20260511_180029

### DIFF
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
@@ -41,6 +41,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
@@ -173,7 +174,7 @@ fun AssistantChatScreen(
 ) {
     val focusManager = LocalFocusManager.current
     val density = LocalDensity.current
-    var bottomBarContentHeightPx by remember { mutableStateOf(0) }
+    var bottomBarContentHeightPx by remember { mutableIntStateOf(0) }
     val bottomBarContentHeight = with(density) { bottomBarContentHeightPx.toDp() }
     val bottomContentPadding = bottomBarContentHeight + FLOATING_COMPOSER_CONTENT_SPACING
     val submitMessage = {
--- a/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
+++ b/libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
@@ -6,14 +6,18 @@ import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.slideInVertically
 import androidx.compose.foundation.background
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
```diff
@@ -161,6 +171,20 @@
     assistantCardRenderer: AssistantCardRenderer? = null,
     onCardAction: (AssistantCardAction) -> Unit = {},
 ) {
+    val focusManager = LocalFocusManager.current
+    val density = LocalDensity.current
+    var bottomBarContentHeightPx by remember { mutableStateOf(0) }
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 6de6d17

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
```diff
@@ -161,6 +172,20 @@ fun AssistantChatScreen(
     assistantCardRenderer: AssistantCardRenderer? = null,
     onCardAction: (AssistantCardAction) -> Unit = {},
 ) {
+    val focusManager = LocalFocusManager.current
+    val density = LocalDensity.current
+    var bottomBarContentHeightPx by remember { mutableIntStateOf(0) }
+    val bottomBarContentHeight = with(density) { bottomBarContentHeightPx.toDp() }
+    val bottomContentPadding = bottomBarContentHeight + FLOATING_COMPOSER_CONTENT_SPACING
```
staskus, line 179: Claude:

> On first frame bottomBarContentHeightPx == 0, so a saved-state chat with prior messages can briefly render the last bubble under the composer before remeasure. Seed bottomContentPadding with COMPOSER_MIN_HEIGHT + 32.dp.

staskus (approved):
✅ I like the floating chat input view. Looks for modern, and works well. 

(Feel free to ignore) One thing that subjectively doesn't look as good for me is using light purple / pink for the background in the light mode of this input view. I would still keep using white or light grey there.

Image

hichamboushaba:
> (Feel free to ignore) One thing that subjectively doesn't look as good for me is using light purple / pink for the background in the light mode of this input view. I would still keep using white or light grey there.

Thanks @staskus, the light purple part is from the Matherial theme, a `tonalElevation` is applied there to make the surface more noticeable, and this uses the primary color (purple for us) for the tone, the goal is to make the screen more accessible, as the floating content is easily separable from the rest, unlike if we use white. I'll keep it for now.

libs/ai-assistant/feature/src/main/kotlin/com/woocommerce/android/aiassistant/ui/AssistantChatScreen.kt
hichamboushaba: I wouldn't worry about this.

PR #15865, opened 2026-05-11
[WOOMOB-2945] Make Help/Support accessible on QR login screen

malinajirka:
Fixes WOOMOB-2945

### Description

The Help/Support entry point that exists on the legacy "enter site URL" login flow was missing from the new QR login flow. This PR adds a Help icon button to the QR login screens so merchants can reach support from anywhere in the QR onboarding without having to back out to the legacy prologue first.

The QR screens are custom Compose UIs without a toolbar, so the legacy menu-item approach doesn't apply. Instead, a small `QrLoginHelpButton` is rendered as a top-right overlay:

### Test Steps

1. Confirm you're on a debug build (the `QR_LOGIN` feature flag is on by default for debug).
2. From the launch screen, tap **Log In**.
3. On **Scan to sign in** (QR prologue), verify a Help icon is visible top-right. Tap it — Help & Support should open. Press back.
4. Tap **Scan QR code** and grant camera permission. Verify the Help icon is still visible top-right over the camera preview. Tap it — Help & Support should open. Press back.
5. Without leaving the scanner, point the camera at any non-WooCommerce QR (or trigger a deep-link error with `adb shell am start -W -a android.intent.action.VIEW -d \"woocommerce://qr-login?token=tooshort\"` on a cold-started app). On the **Not a WooCommerce code** error screen, verify the Help icon is still top-right and opens Help & Support.

### Images/gif

| QR prologue | Scanner | Error screen |
| --- | --- | --- |
| dafdsa | woomob-2945-scanner-with-help | woomob-2945-error-screen-with-help |

[Internal] — QR login feature is still behind \`FeatureFlag.QR_LOGIN\` (debug only).


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginHelpButton.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.dimensionResource
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import com.woocommerce.android.R
+
+@Composable
+fun QrLoginHelpButton(
```
JorgeMucientes: This is just a generic composable Help button, so perhaps we can remove the "QrLogin" part from the component naming.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueFragment.kt
```diff
@@ -55,7 +56,12 @@ class QrLoginPrologueFragment : Fragment() {
                 analyticsTracker.track(AnalyticsEvent.LOGIN_QR_PROLOGUE_FALLBACK_TAPPED)
                 unifiedLoginTracker.trackClick(UnifiedLoginTracker.Click.LOGIN_QR_FALLBACK)
                 listener?.onQrLoginFallbackClicked()
-            }
+            },
+            onHelpClicked = {
+                analyticsTracker.track(AnalyticsEvent.LOGIN_QR_HELP_TAPPED)
+                unifiedLoginTracker.trackClick(UnifiedLoginTracker.Click.SHOW_HELP)
```
JorgeMucientes, line 62: Given we already have this `UnifiedLoginTracker.Click.SHOW_HELP` event, do we really need to add a new  `AnalyticsEvent.LOGIN_QR_HELP_TAPPED` event? 
I think we just need to make sure that all QR code login screens/steps handle the `step` tracking property value correctly so that any `UnifiedLoginTracker` event has for example "step=qr-code-scanned" values. Similar to what we do for the rest of the login steps/screens.
Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerFragment.kt
```diff
@@ -103,6 +109,11 @@ class QrLoginScannerFragment : Fragment() {
             onStartOver = ::handleStartOver,
             onRetryExchange = qrLoginViewModel::onRetryExchange,
             onFallbackClicked = { listener?.onQrLoginFallbackClicked() },
+            onHelpClicked = {
+                analyticsTracker.track(AnalyticsEvent.LOGIN_QR_HELP_TAPPED)
+                unifiedLoginTracker.trackClick(UnifiedLoginTracker.Click.SHOW_HELP)
```
JorgeMucientes, line 124: Same as above (regarding the tracking)

JorgeMucientes (approved):
Changes LGTM. I just left a couple non-blocking comments that are worth checking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginPrologueFragment.kt
malinajirka: I'll tackle all tracking in a final PR when everything else is done. If I didn't add it here, AI kept complaining that tracking isn't consistent with the rest of the feature. I think most of the tracking we added will be remove in favor of unified login events.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginHelpButton.kt
malinajirka: Fixed in https://github.com/woocommerce/woocommerce-android/pull/15865/commits/2f1b00b23513b6f3eace7ffca8d2a6519915df5c

PR #15866, opened 2026-05-11
[WOOMOB-3009] QR login: make error screen scrollable with anchored buttons

malinajirka:
### Description

Fixes WOOMOB-3009.

Make `QrLoginErrorScreen` use the same scrollable-content + bottom-anchored-buttons pattern as the prologue and confirm screens, so the primary/secondary actions stay reachable in landscape, with large fonts, or with longer translations.

Also renames the private `Hero*` composables in the prologue and confirm screens to `Content*` for consistency.

### Test Steps

1. Trigger the error screen: `adb shell am start -a android.intent.action.VIEW -d "woocommerce://qr-login?token=invalid" com.woocommerce.android.dev`.
2. Rotate to landscape and bump `font_scale` to 1.3 — both buttons stay reachable; content scrolls if needed.

### Images/gif
07-landscape-2 0x-scrolled-with-padding

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginConfirmSiteScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginConfirmSiteScreen.kt
@@ -48,7 +48,7 @@ fun QrLoginConfirmSiteScreen(
             .padding(horizontal = dimensionResource(id = R.dimen.major_150)),
         horizontalAlignment = Alignment.CenterHorizontally,
     ) {
-        // Hero scrolls so the Connect / Cancel buttons stay visible in landscape on phones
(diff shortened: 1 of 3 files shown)


JorgeMucientes (approved):
Works as expected. I'm only concern about one minor thing. In multiple cases when font size is big the error description is simply not visible and it's hard to tell that the layout is scrollable. 
Anyway this is very minor given this is really an edge case (an error screen, in landscape with increased font size). 

https://github.com/user-attachments/assets/c0459dc6-16f8-4bb5-8359-ab3a5bfd2565

malinajirka:
Thanks for the review! 

> I'm only concern about one minor thing. In multiple cases when font size is big the error description is simply not visible and it's hard to tell that the layout is scrollable.

I personally wouldn't worry about it - the CTAs are visible and that's the most important part. Chances someone will be logging on a small phone, with an increased font size, in landscape feels like a super edge case.

PR #15869, opened 2026-05-11
[Woo POS] Custom Amount, Part 2: cart rendering

samiuelson:
## Summary

Part 2 of 6. Adds the cart-row composable for custom amounts that Part 1 laid the foundation for: an initials avatar (first letter of `name` on `surfaceContainerLow`), the formatted amount, an optional "Tax included" subline, and an overflow (kebab) menu with **Edit** and **Remove** actions. Edit opens the dialog (which arrives in Part 4); Remove reuses the existing cart-item removal path.

> Note: Part 8 (#15887) later replaces the overflow menu with separate pencil + trash buttons on **tablet** while keeping the overflow menu on **phone**, for iOS parity. This PR ships the phone variant (overflow only) — the tablet split happens in Part 8.

## What's included

- `WooPosCartScreen.kt`: new `CustomAmountItem` composable + `CustomAmountOverflowMenu`; replaces the `Unit` stub from Part 1. Accessibility is consolidated on the outer `Row` (`contentDescription`) with inner text silenced via `clearAndSetSemantics { }`. The overflow menu is gated by `canRemoveItems` — this also locks editing during checkout (intentional; comment in code).
- `WooPosCustomAmountInitialsAvatar.kt`: new reusable component rendering the first character of a name on `surfaceContainerLow`. Mirrors iOS `CustomAmountAvatar`. Previews cover populated and blank-name cases.
- Strings (cart-only): `woopos_cart_item_custom_amount_content_description`, `woopos_cart_custom_amount_includes_tax`, `woopos_cart_custom_amount_menu_edit`, `woopos_cart_custom_amount_menu_remove`.

## Stacking

This PR targets **`feature/woopos-custom-amount-1-foundation`** (Part 1) as its base. It must merge after Part 1 lands on trunk.

## Verification

- `./gradlew :WooCommerce:compileWasabiDebugKotlin`
- `./gradlew detektAll`

No new unit tests in this part — the cart rendering is purely Compose; behaviour is covered by Part 1's `WooPosCartViewModelTest` (which exercises the cart's add/edit handler).

## Test plan

- [ ] Compile passes.
- [ ] When run end-to-end (after Parts 3–5 land), tapping the overflow menu on a custom-amount cart row exposes Edit (reopens the dialog prefilled) and Remove.

## Stack

This feature ships as 6 stacked PRs (≤300 prod LOC each):

- [Part 1 — foundation (data + events + cart handler)](https://github.com/woocommerce/woocommerce-android/pull/15868)
[...]

PR #15868, opened 2026-05-11
[Woo POS] Custom Amount, Part 1: foundation (data + events + cart handler)

samiuelson:
## Summary

Part 1 of 5 splitting up #15867 (the bundled draft PR) into reviewable slices. Adds the data layer and event bus plumbing for the custom-amount feature: new cart-item type, new ItemClickedData variant, new parent/child events, dialog state in `WooPosHomeState`, cart VM handling, the `WooPosCustomAmountCartHandler`, totals fee-line creation, and analytics events.

No user-visible behaviour yet — there's no UI to add or render custom amounts. The pipe is fully wired so subsequent PRs only need to plug UI in.

## What's included

- New types: `WooPosCartItemViewState.CustomAmount`, `ItemClickedData.CustomAmount`.
- Events:
  - `ChildToParentEvent.CustomAmountDialogRequested(editing)` and `ChildToParentEvent.CustomAmountSubmitted(...)`.
  - `ParentToChildrenEvent.CustomAmountSubmitted(...)`.
- `WooPosHomeState.DialogState.CustomAmountDialog(editing)` + `WooPosHomeUIEvent.DismissCustomAmountDialog`.
- `WooPosHomeViewModel` wires the new events and dialog dismissal.
- `WooPosCartUIEvent.EditCustomAmountClicked` + cart VM handling.
- `WooPosCartViewModel` handles `ParentToChildrenEvent.CustomAmountSubmitted` via the new `WooPosCustomAmountCartHandler`, updates `getCartItemsDataList`, `placeItemInList` (so custom amounts sit between coupons and products), and treats them as checkout-enabling items.
- `WooPosTotalsRepository.createOrder` now converts `ItemClickedData.CustomAmount` entries to `Order.FeeLine` with the correct `taxStatus`.
- Analytics: new `CustomAmountSubmitted(mode, isTaxable)` event; `ItemAddedToCart` / `ItemRemovedFromCart` extended with `ItemsListItemType.CUSTOM_AMOUNT`.
- Exhaustive `when` branches added everywhere new sealed-class variants land (cart updater, items search, items VM, products VM, totals VM, search VM).
- `@Suppress(\"LargeClass\")` on `WooPosCartViewModel` (the file was already near the threshold).
- A `Unit` stub branch for `CustomAmount` in `WooPosCartScreen` — replaced by the full cart row composable in Part 2.

## Tests

- `WooPosCustomAmountCartHandlerTest` (new).
[...]

PR #15870, opened 2026-05-11
[Woo POS] Custom Amount, Part 3: dialog ViewModel + state + currency helper

samiuelson:
## Summary

Part 3 of 5. Introduces the headless side of the custom-amount dialog: state model, ViewModel, and a small reusable helper that loads currency-formatting parameters (symbol, position, decimal separator, number of decimals) from the selected site's WooCommerce settings.

The Compose UI that consumes this VM lands in Part 4.

## What's included

- `WooPosCustomAmountDialogState` — `@Parcelize` state with mode (Add / Edit), amount, name, isTaxable, isSubmitting flag, and currency formatting params.
- `WooPosCustomAmountDialogViewModel` — Hilt ViewModel that:
  - eagerly loads currency params in `init {}` so user input is never wiped by an async overwrite;
  - exposes `initializeFor(editing)`, `onAmountChanged`, `onNameChanged`, `onTaxableToggled`, `onSubmit`, `onDismissed`;
  - guards `onSubmit` against double-tap via `isSubmitting` flag;
  - uses `R.string.woopos_custom_amount_dialog_name_placeholder` as the default name (no hardcoded localised string).
- `WooPosGetCurrencyFormattingParameters` — small helper that wraps `WooCommerceStore.getSiteSettings` with sensible defaults.
- New string resource: `woopos_custom_amount_dialog_name_placeholder` (used by the VM for the default fallback name).

## Tests

- `WooPosCustomAmountDialogViewModelTest` — covers add / edit / submit / validation / default-name / double-tap guard / re-initialization idempotence.

## Stacking

Targets **`feature/woopos-custom-amount-2-cart-ui`** as its base.

## Verification

- `./gradlew :WooCommerce:compileWasabiDebugKotlin`
- `./gradlew detektAll`
- `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests \"*WooPosCustomAmountDialogViewModelTest\"`

## Test plan

- [ ] Compile passes.
- [ ] Unit tests above pass.
- [ ] (End-to-end behaviour appears once Parts 4–5 land.)

## Stack

This feature ships as 6 stacked PRs (≤300 prod LOC each):

- [Part 1 — foundation (data + events + cart handler)](https://github.com/woocommerce/woocommerce-android/pull/15868)
- [Part 2 — cart rendering](https://github.com/woocommerce/woocommerce-android/pull/15869)
- **Part 3 — dialog ViewModel + state + currency helper** (this PR) — [#15870](https://github.com/woocommerce/woocommerce-android/pull/15870)
- [Part 4 — dialog Compose UI + render hookup](https://github.com/woocommerce/woocommerce-android/pull/15871)
- [Part 5 — entry row + products screen integration](https://github.com/woocommerce/woocommerce-android/pull/15872)
- [Part 6 — refund support](https://github.com/woocommerce/woocommerce-android/pull/15874)


toupper (approved):
Clean VM. The sealed `Mode` (Add / Edit(itemNumber)) reads much better than a nullable `editingItemNumber: Int?` and naturally maps to the dialog title flip in Part 4. `isSubmitEnabled` as a derived getter is the right shape and the double-tap guard via `isSubmitting` is plumbed correctly. `hasInitializedFor` keying off the item number (or `"add"`) so re-emitting the same trigger doesn't blow away user input is a nice touch — and `onDismissed` clearing it means the next open re-initializes fresh.

Eagerly loading currency params in `init {}` and `.copy()`-ing them onto the current state (rather than overwriting the whole state) is exactly right — user input that lands before the async settle is preserved. `WooPosGetCurrencyFormattingParameters` with a per-instance cache is fine for a POS session lifetime; if site settings can change mid-session, an explicit invalidation path would be needed, but I don't think that's an actual scenario here.

Tests are thorough: add/edit/submit, blank-name default, zero-amount disabled, double-tap guard, re-init idempotence. ✓

A few small things, none blocking:

1. **Redundant check in `onSubmit`.** `if (!current.isSubmitEnabled || current.isSubmitting) return` — the `isSubmitEnabled` getter already factors in `!isSubmitting`, so the second clause is a no-op. Harmless.

2. **`hasInitializedFor: Any?`** typed as `Any` with values that are either an `Int` (itemNumber) or the sentinel string `"add"`. Either a small sealed type, or just `Int? = null` with the existing `Mode.Add` carrying `null` semantics, would be slightly easier on the next reader.

3. **`@Volatile` on `hasInitializedFor`** is unusual for a VM that's effectively main-dispatcher-only. Probably defensive against unexpected dispatcher juggling, but worth a one-line comment if there's a specific reason.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15872, opened 2026-05-11
[Woo POS] Custom Amount, Part 5: entry row + products screen integration

samiuelson:
## Summary

Part 5 of 5 (final). Adds the \"Custom amount / Add a one-off charge\" entry row above the products list and wires the click through to open the dialog. Once this PR lands, the feature is end-to-end functional: merchants can add, edit, and remove custom amounts in the cart and have them included as fee lines on the resulting order.

## What's included

- `WooPosCustomAmountEntryRow.kt`: new card-style row composable with a money icon, title, and subtitle.
- `WooPosProductsUIEvent.CustomAmountEntryRowClicked` + handler in `WooPosProductsViewModel` that emits `ChildToParentEvent.CustomAmountDialogRequested()`.
- `WooPosProductsScreen.kt`: renders the entry row above the products list when content is loaded.
- Entry-row strings: `woopos_custom_amount_entry_title`, `woopos_custom_amount_entry_subtitle`.

## Stacking

Targets **`feature/woopos-custom-amount-4-dialog-ui`** as its base. After Parts 1–5 all merge, this completes the feature.

## Verification

- `./gradlew :WooCommerce:compileWasabiDebugKotlin`
- `./gradlew detektAll`

## Test plan (end-to-end now possible)

- [ ] Open POS on a tablet, see the \"Custom amount\" row at the top of the products list.
- [ ] Tap it → dialog opens.
- [ ] Enter an amount, optional name, toggle Charge taxes, tap \"Add custom amount\" → row appears in cart.
- [ ] Tap pencil on the cart row → dialog reopens prefilled → update → row updates in place.
- [ ] Tap trash on the cart row → row removed.
- [ ] Add a custom amount only (no products) → checkout button is enabled.
- [ ] Add product(s) + custom amount → Checkout → order draft contains both items and the fee line with the correct tax status.
- [ ] Verify the order on the backend has the expected `fee_lines` entry.
- [ ] Repeat in dark mode and on a phone layout.

## Stack

This feature ships as 6 stacked PRs (≤300 prod LOC each):

- [Part 1 — foundation (data + events + cart handler)](https://github.com/woocommerce/woocommerce-android/pull/15868)
- [Part 2 — cart rendering](https://github.com/woocommerce/woocommerce-android/pull/15869)
- [Part 3 — dialog ViewModel + state + currency helper](https://github.com/woocommerce/woocommerce-android/pull/15870)
[...]

PR #15871, opened 2026-05-11
[Woo POS] Custom Amount, Part 4: dialog Compose UI + render hookup

samiuelson:
## Summary

Part 4 of 5. The Compose UI for the custom-amount dialog, plus wiring into `WooPosHomeDialogs` so it shows whenever `DialogState.CustomAmountDialog` is active.

After this PR lands the dialog can be opened from the cart's edit pencil (added in Part 2). The entry point from the products list comes in Part 5.

## What's included

- `WooPosCustomAmountDialog.kt`: the Compose dialog with title (mode-dependent), amount input (`WooPosMoneyInputField`), name input (`WooPosInputField`), "Charge taxes" switch, Cancel / primary action buttons. On phone it renders as a full-screen `Surface` with a top close toolbar and a sticky bottom action bar; on tablet it uses `WooPosDialogWrapper` as a centered modal. `LaunchedEffect(isVisible, editing?.itemNumber)` drives initialise/reset.
- `WooPosHomeDialogs.kt`: render `WooPosCustomAmountDialog` when the home state's `dialogState` is `CustomAmountDialog`.
- Dialog strings: titles, labels, button captions, cancel.

## Stacking

Targets **`feature/woopos-custom-amount-3-dialog-vm`** as its base.

## Verification

- `./gradlew :WooCommerce:compileWasabiDebugKotlin`
- `./gradlew detektAll`

## Test plan

- [ ] Compile passes.
- [ ] Open the dialog by editing a custom amount in the cart (becomes possible end-to-end after Part 5 lets users add one).
- [ ] Dialog title flips between "Add custom amount" / "Edit custom amount" based on mode.
- [ ] Currency symbol / decimal separator follow site settings.
- [ ] Toggle Charge taxes — state preserved on submit.
- [ ] Cancel button dismisses without emitting; primary button disabled when amount is empty or zero.

## Stack

This feature ships as 6 stacked PRs (≤300 prod LOC each):

- [Part 1 — foundation (data + events + cart handler)](https://github.com/woocommerce/woocommerce-android/pull/15868)
- [Part 2 — cart rendering](https://github.com/woocommerce/woocommerce-android/pull/15869)
- [Part 3 — dialog ViewModel + state + currency helper](https://github.com/woocommerce/woocommerce-android/pull/15870)
- **Part 4 — dialog Compose UI + render hookup** (this PR) — [#15871](https://github.com/woocommerce/woocommerce-android/pull/15871)
[...]

PR #15874, opened 2026-05-11
[Woo POS] Custom Amount, Part 6: refund support

samiuelson:
## Summary

Part 6 of the custom-amount stack. After Parts 1–5 let merchants add custom amounts to a POS order, Part 6 makes them refundable — fee lines now appear in the POS refund items list and are correctly carried into the refund POST.

Mirrors iOS, which sends fee refunds in the existing `line_items` array with `quantity = 0`; WooCommerce resolves the id against the order's fee_lines server-side. No FluxC changes are required — `RefundRestClient.createRefundByItems` and `Refund.feeLines` already accept and parse this shape.

## What's included

- `WooPosRefundableItem`: new `isLumpSum: Boolean = false` flag; defaults preserve product behaviour. `uniqueId` switches to `fee_{feeId}` for lump-sum rows.
- `WooPosGetRefundableItems`: appends one refundable row per `order.feesLines` entry, skipping any fee whose id already appears in a prior refund's `feeLines`.
- `WooPosGroupRefundItems`: partitions rows; emits one `RefundRequestItem(quantity = 0, refundTotal = feeTotal, refundTax = …)` per fee line, sourcing the tax breakdown from the order's `FeeLine.taxes`.
- `WooPosCalculateRefundSubtotal` / `WooPosCalculateRefundTax`: lump-sum rows are summed directly (no quantity multiplication).
- `WooPosIssueRefundScreen`: renders a money-icon avatar (`ic_gridicons_money_on_surface` on `primaryContainer`) in place of the product placeholder for lump-sum rows; reuses the existing checkbox + selection wiring.

## Tests

- `WooPosGroupRefundItemsTest` — lump-sum fee emits `quantity = 0` with full total and full tax; mixed product + fee runs.
- `WooPosGetRefundableItemsTest` — fees become refundable; already-refunded fees are excluded; product + fee combination.
- `WooPosCalculateRefundSubtotalTest` — lump-sum sums without quantity multiplication.
- `WooPosCalculateRefundTaxTest` — lump-sum tax uses `unitTax` directly; combined with product math.

## Stacking

Targets **`feature/woopos-custom-amount-5-entry-row`** as its base. After Parts 1–5 land on trunk, this rebases onto trunk.

## Verification

- `./gradlew :WooCommerce:compileWasabiDebugKotlin`
- `./gradlew detektAll`
- `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*WooPosRefund*" --tests "*WooPosGetRefundableItems*" --tests "*WooPosGroupRefundItems*" --tests "*WooPosCalculateRefund*"`

## Test plan

- [ ] On a tablet, place an order with a product + a custom amount; complete payment.
- [ ] Open the order's refund flow → both the product row and the custom-amount row appear; the fee row shows a money-icon avatar instead of a product image and has no quantity stepper.
- [ ] Select only the custom amount → confirm refund → the issued refund on the backend has a fee-line refund with the correct amount and tax.
- [ ] Open the same order again → the (now-refunded) custom amount no longer appears in the refundable list.
- [ ] Place an order with only a custom amount, refund it, verify success.

## Out of scope

- Partial refunds of a custom amount (iOS only supports full refunds — matched here).
- Refund of shipping lines (separate, pre-existing gap in POS).
- FluxC / data-layer changes (not required).

## Stack

This feature ships as 6 stacked PRs (≤300 prod LOC each):

- [Part 1 — foundation (data + events + cart handler)](https://github.com/woocommerce/woocommerce-android/pull/15868)
- [Part 2 — cart rendering](https://github.com/woocommerce/woocommerce-android/pull/15869)
- [Part 3 — dialog ViewModel + state + currency helper](https://github.com/woocommerce/woocommerce-android/pull/15870)
- [Part 4 — dialog Compose UI + render hookup](https://github.com/woocommerce/woocommerce-android/pull/15871)
- [Part 5 — entry row + products screen integration](https://github.com/woocommerce/woocommerce-android/pull/15872)
- **Part 6 — refund support** (this PR) — [#15874](https://github.com/woocommerce/woocommerce-android/pull/15874)

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosCustomAmountInitialsAvatar.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosCustomAmountInitialsAvatar.kt
@@ -10,8 +10,8 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.text.font.FontWeight
-import androidx.compose.ui.unit.dp
 import com.woocommerce.android.ui.woopos.common.composeui.WooPosPreview
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosComponentSize
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosCornerRadius
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTheme
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
@@ -42,11 +42,11 @@ fun WooPosCustomAmountInitialsAvatar(
 
 @WooPosPreview
 @Composable
-private fun WooPosCustomAmountInitialsAvatarPreview() {
+fun WooPosCustomAmountInitialsAvatarPreview() {
     WooPosTheme {
         Box(
             modifier = Modifier
-                .size(96.dp)
+                .size(WooPosComponentSize.Medium.value)
                 .clip(RoundedCornerShape(WooPosCornerRadius.Medium.value)),
         ) {
             WooPosCustomAmountInitialsAvatar(name = "Service fee")
@@ -56,11 +56,11 @@ private fun WooPosCustomAmountInitialsAvatarPreview() {
 
 @WooPosPreview
 @Composable
-private fun WooPosCustomAmountInitialsAvatarBlankNamePreview() {
+fun WooPosCustomAmountInitialsAvatarBlankNamePreview() {
     WooPosTheme {
         Box(
             modifier = Modifier
-                .size(96.dp)
+                .size(WooPosComponentSize.Medium.value)
                 .clip(RoundedCornerShape(WooPosCornerRadius.Medium.value)),
         ) {
             WooPosCustomAmountInitialsAvatar(name = "")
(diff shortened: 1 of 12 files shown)


toupper (approved):
Solid refund integration. The `partition { it.isLumpSum }` pattern repeats consistently across `WooPosCalculateRefundSubtotal` / `WooPosCalculateRefundTax` / `WooPosGroupRefundItems`, which reads well and keeps the product math untouched. The `quantity = 0` request shape matches the WC REST contract for fee refunds, and `buildFeeRefundTaxes` propagating the per-rate breakdown from `feeLine.taxes` keeps the tax math server-side rather than reconstructing it client-side.

Particularly appreciate the inline comment in `buildFeeRows` calling out the "fees are refunded whole" design constraint and the explicit upgrade path if the backend ever supports partial-fee refunds — that's exactly the kind of breadcrumb that pays off later. The `refundedFeeIds` set lookup is cheap and the exclusion happens at the right place. Avatar reuse from Part 2's `WooPosCustomAmountInitialsAvatar` is clean.

Test coverage is thorough: lump-sum-only, lump-sum + product mixed, already-refunded fee excluded, quantity-0 shape, taxes propagated.

**One thing to flag — out-of-scope change.** The PR also flips `WOO_POS_PHONE` from `localValue = PackageUtils.isDebugBuild()` to `localValue = true`. That's a meaningful default change (phone POS is now enabled on every local build regardless of variant), and it's not mentioned anywhere in the description, the test plan, or the "What's included" list. Either it's intentional (and worth calling out in the description so the next reviewer doesn't miss it) or it slipped in from another branch. Worth confirming.

PR #15875, opened 2026-05-11
[WOOMOB-2975] Simplify AI Assistant error diagnostics

hichamboushaba:
### Description
Fixes WOOMOB-2975

#### Problem
The AI Assistant error pipeline was losing the information needed to understand backend failures.

For SSE failures from `/wpcom/v2/jetpack-ai-query`, the stream client only classified a few specific HTTP statuses (`401`, `403`, `408`, `429`, and `5xx`). Other client errors such as `400`, `404`, and `422` could fall through to `ChatStreamError.UNKNOWN`, so a real backend request failure could reach the app as `AssistantError.Unknown(cause = null)`.

The stream endpoint can also fail before a server-sent event stream is established. In those cases, OpenAI-compatible endpoints normally return a regular JSON error response rather than `text/event-stream`. The Jetpack AI proxy can also return `application/json` with HTTP `200` for schema validation failures, including an observed empty-body response for an invalid tool schema. The previous streaming path did not reliably preserve that response body or classify that protocol-level error as a bad request.

That meant we lost the HTTP status, request id, Retry-After value, and response context exactly when those details were needed for logs, telemetry, tests, and debugging. At the same time, those backend details must not be shown to merchants or sent back to the model.

#### Why this change
This PR makes the error model explicit enough to distinguish predictable request failures from unknown failures, while preserving safe diagnostics end-to-end. The goal is to keep useful internal debugging context available without leaking raw backend payloads, auth data, tool arguments, prompts, or model content into UI copy, model-facing tool JSON, or user-facing analytics.

It also closes the same diagnostics gap for API-backed tool calls. Tool failures previously returned mostly empty diagnostics, so the new architecture helped stream failures but not tool failures. This PR carries tool diagnostics where the existing FluxC/store error surfaces expose enough information, and deliberately avoids fabricating unavailable HTTP fields for lossy error types.

#### Changes
- Added shared diagnostics types for assistant errors and tool results, including transport diagnostics and tool diagnostics.
- Added `BadRequest` / `BAD_REQUEST` handling for `400`, other non-special `4xx` stream failures, and JSON non-SSE responses from the stream endpoint.
- Read non-SSE stream responses through OkHttp so JSON error bodies can be captured before the response is closed.
- Parsed WordPress-style JSON error bodies with `data.status` to preserve the logical HTTP status when the proxy tunnels an error inside a successful response.
- Classified successful `application/json` responses from the stream endpoint as request/protocol failures while preserving available transport diagnostics, including the observed empty-body `HTTP 200` case.
- Preserved transport diagnostics from stream failures: HTTP status, allowlisted request id, bounded Retry-After, and capped/redacted body snippets.
- Kept pre-stream `401` refresh behavior in `JetpackAiChatService`, with one retry before visible output only.
- Replaced overlapping retry booleans with a shared `RetryAffordance` at the loop/runtime/UI boundary while keeping loop-internal retry decisions separate.
- Added `TransportDiagnosticsFactory` redaction/capping support for OkHttp responses and retained raw HTTP/Volley response data.
- Added `ToolFailureDiagnosticsFactory` so API-backed tool `TransportError` results include non-empty tool diagnostics and preserve available transport diagnostics.
- Normalized tool handler exceptions at the registry boundary without leaking raw exception text to callers.
- Preserved unsafe write semantics by widening deterministic failures to `ToolFailed` and unknown write outcomes to `OutcomeUnknown` with diagnostics.
- Kept model-facing tool failure JSON generic, including for transport failures.
- Added an allowlisted `AssistantTelemetry` event shape and a no-op default implementation.
- Added and updated focused unit coverage for stream classification, JSON/non-SSE diagnostics, redaction, retry affordance, telemetry boundaries, UI/model non-leakage, and tool diagnostics.

#### Boundaries
- No `:WooCommerce` host integration changes.
- No new auth path; existing Jetpack AI JWT handling is reused.
- AI payload handling remains on `kotlinx.serialization.json`; no Gson path added.
- Raw body snippets are internal diagnostics only and are not read by UI mappers, model-facing messages, or user-facing analytics.

### Test Steps
N/A - this is internal AI Assistant error mapping and diagnostics plumbing with no manual UI flow. CI and focused unit coverage validate the behavior.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
Great addition. I will borrow some of this for iOS.

PR #15881, opened 2026-05-12
Send token_hash with every QR login session-status poll

JorgeMucientes:
### Description

Pairs with WC Core [#64503](https://github.com/woocommerce/woocommerce/pull/64503) ([commit `b54daa591e`](https://github.com/woocommerce/woocommerce/commit/b54daa591e)), which made `token_hash` a required query parameter on `/qr-login-session-status` as part of the security review hardening. The Android client at `rsm-qr-login-main` only sends `session_id`, so every poll currently returns `400 rest_missing_callback_param` and the flow stalls after scan. This PR sends the missing `token_hash` on every poll.

#### Why the server needs it

`session_id` is the one value the app **can't** avoid putting on the wire — it's printed in every 2-second poll URL, so it leaks readily through logs, packet captures, screen recordings, support-ticket attachments. Before the hardening, anyone holding a leaked `session_id` could poll the same endpoint themselves and harvest the `exchange_grant` the moment the merchant approved on wc-admin. Binding each poll to `SHA-256(token)` proves the caller actually scanned the QR — the plaintext token never leaves the device, so an attacker with only the leak can't derive the matching hash and is held at an opaque `state: expired`.

```mermaid
sequenceDiagram
  participant Phone
  participant Logs as 📋 logs / capture /<br/>support attachment
  participant Attacker as 🕵️ Attacker
  participant Server

  Phone->>Server: POST /qr-login-scan { token }
  Server-->>Phone: { session_id }
  Note right of Server: stores SHA-256(token)<br/>keyed by SHA-256(session_id)

  Phone->>Server: GET /qr-login-session-status<br/>{ session_id, token_hash }
  Phone-->>Logs: session_id leaks in URL / log lines
  Logs-->>Attacker: reads session_id

  rect rgba(214,54,56,0.10)
  Note over Attacker,Server: Old design — session_id alone was enough
  Attacker->>Server: poll { session_id }
  Server-->>Attacker: { state: approved, exchange_grant } ❌
  end

  rect rgba(26,175,93,0.10)
  Note over Attacker,Server: New design — token_hash required
  Attacker->>Server: poll { session_id, token_hash: ??? }
  Server-->>Attacker: { state: expired } (opaque) ✅
  end
```

#### What changed Android-side

- `QrLoginRestClient.checkSessionStatus(siteUrl, sessionId, token)` hashes the token and adds `&token_hash=<sha256_hex>` on every poll.
- New `sha256Hex()` helper on the companion object — lowercase-hex output matches PHP's `hash('sha256', $token)` byte-for-byte under server-side `hash_equals`.
- `QrLoginScannerViewModel` passes `ticket.token` (already in scope from `scan()`) through to each poll tick.
- Tests pin `SHA-256` for two known plaintexts so a future change to algorithm/encoding/case fails loudly.

### Test Steps

Requires the WC Core build from PR [#64591](https://github.com/woocommerce/woocommerce/pull/64591) at `7bb7a1e79d` or later (the build that contains the `token_hash` server-side check).

1. Install the debug APK from this branch on a device.
2. Generate a QR code from the test site (`wp-admin → Mobile app login`).
3. Scan the QR. The merchant browser should advance to the "match number" step within ~2 s, and the Android app should show the number-display screen with a 90 s countdown.
4. Tap the correct number on wc-admin. The app should advance through "Signing in…" and finish at the logged-in store.

Negative path:

5. Re-scan the same QR after expiry (or tap "Cancel" mid-flow). The app should land on "Sign-in timed out" rather than hanging on `state: scanned` indefinitely.

### Images/gif

N/A — wire-level change. Visible only in HTTP logs and in the fact that the flow no longer dead-ends at the polling step.

> No release-notes entry: the QR login feature has not shipped to merchants yet, so the user-visible behavior on a real install is "QR login works" rather than "a previously broken path now works." A `[Internal]` note would be appropriate if the team prefers to track every Core-pairing fix in the changelog.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/QrLoginRestClient.kt
@@ -18,6 +18,7 @@ import java.net.HttpURLConnection.HTTP_FORBIDDEN
 import java.net.HttpURLConnection.HTTP_NOT_FOUND
 import java.net.HttpURLConnection.HTTP_PRECON_FAILED
 import java.net.HttpURLConnection.HTTP_UNAUTHORIZED
+import java.security.MessageDigest
 import javax.inject.Inject
 import javax.inject.Named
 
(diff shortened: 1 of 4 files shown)


malinajirka (approved):
LGTM

PR #15880, opened 2026-05-12
[AI Support Chat] Issue picker diagnostics flow

iamgabrielma:
Closes WOOMOB-3062

### Test Steps
  1. Open the app with a Jetpack-connected store.
  2. Go to: Menu > Help & Support > AI Troubleshooting
  3.   Confirm the chat opens with:
      - greeting message,
      - issue picker,
      - disabled input bar.
  4. Tap each issue type, especially:
      - I can't see my orders
      - I can't see my products
      - My analytics aren't loading
  5. Confirm:
      - diagnostics progress appears inline,
      - checks move through pending/running/passed,
      - after success, the selected issue is sent to Odie automatically,
      - bot reply appears,
      - input becomes enabled.
  6. Failure path:
      - Disable network or use airplane mode before selecting an issue.
      - Pick an issue.
      - Confirm diagnostics fail.
      - Confirm Retry and Continue buttons appear.
      - Tap Retry with network still disabled and confirm it reruns.
      - Re-enable network, tap Retry again, confirm it can proceed.
      - Or tap Continue and confirm it starts the chat with the diagnostics failure context.

### Images/gif

https://github.com/user-attachments/assets/8c34a51d-ea6d-44ef-9476-f37b88d34a85

@itsmeichigo

PR #15883, opened 2026-05-12
[AI Support Chat] Connectivity Tool entry

iamgabrielma:
### Description

Adds an AI Support Chat entry point from the Connectivity Tool, plus active-session `session_id` plumbing.

When AI Support Chat is enabled for a Jetpack-connected site, the Connectivity Tool can launch AI Troubleshooting with the completed connectivity checks as context. The chat converts those checks into diagnostic context and starts the first Odie message automatically.

This PR also forwards the returned `session_id` for follow-up messages within the active chat session. Chat history, resume, and session persistence are intentionally left for a follow-up PR.

### Test Steps
  1. Enable the AI Support Chat feature flag if needed.
  2. Use a Jetpack-connected store.
  3. Open the Connectivity Tool and run the checks.
  4. Confirm AI Troubleshooting opens and starts a chat using the connectivity context.
  5. Send a follow-up message and confirm it succeeds.

Screenshot 2026-05-13 at 11 34 07


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -276,3 +320,35 @@ private fun greetingMessage(): AiSupportChatMessage =
...
+private fun List<ConnectivityCheckCardData>.toDiagnosticResult(): DiagnosticResult =
+    DiagnosticResult(
+        issueType = SupportIssueType.OTHER,
+        statuses = map { check ->
+            DiagnosticStatus(
+                test = check.type.toDiagnosticTest(),
+                status = check.status.toTestStatus()
+            )
+        }
+    )
```
itsmeichigo, line 422: According to Codex: The connectivity-tool launch converts every check into diagnostic context, including checks that never ran after the first failure. That means Odie receives pending statuses and is_complete = false for a completed connectivity-tool run. iOS only builds context from completed test results. I’d filter to Success/Failure checks before building the diagnostic result, or explicitly model the later checks as skipped/stopped instead of pending.

itsmeichigo (changes requested):
I tested the feature and found a few places where feature parity should be improved:
- In Troubleshoot Connection screen:
  - CTA Contact Support should be hidden if AI chat feature flag is enabled. 
  - CTA Ask AI should be renamed to Chat with support.
  - Both CTA should be hidden until all checks complete.
 - In the chat screen:
   - Let's not send the connectivity result in the chat. Start with a greeting message to ask user to describe their problem.
   - Let's not send the automatic message "Help me troubleshoot...". This together with the missing troubleshooting results in context ([commented](https://github.com/woocommerce/woocommerce-android/pull/15880#discussion_r3232365544) in a previous PR) causes the bot to mistake the problem and tries to suggest solution for login:

https://github.com/user-attachments/assets/55655dac-0d9a-4a80-8e5b-9c27ee5e0882

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/TroubleshootConnectionViewModel.kt
```diff
@@ -190,12 +199,21 @@ class TroubleshootConnectionViewModel @Inject constructor(
         }.trimEnd()
     }
 
+    private fun isAiSupportChatAvailable(): Boolean =
+        featureFlagRepository.isEnabled(FeatureFlag.AI_SUPPORT_CHAT) &&
+            selectedSite.getIfExists()?.isJetpackConnected == true
```
malinajirka, line 208: 💡 The eligibility check has two gates (flag + Jetpack) but the tests only exercise the both-enabled branch. The most likely real regression vector here is the `isJetpackConnected` check accidentally inverting or being dropped — which would expose the button to app-password / non-Jetpack merchants where the AI flow isn't supported.

A `given flag enabled, site not Jetpack-connected -> button hidden` test (and the inverse `flag disabled, jetpack-connected -> hidden`) would lock that down. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/TroubleshootConnectionViewModel.kt
```diff
@@ -94,6 +98,11 @@ class TroubleshootConnectionViewModel @Inject constructor(
         triggerEvent(OpenSupportRequest(diagnosticLog = generateDiagnosticLog()))
     }
 
+    fun onAiSupportChatClicked() {
+        if (!checksFlow.value.isFinished()) return
```
malinajirka, line 106: 🔍 np: the `!isFinished()` early-return guard has no negative test — only the positive case (`given checks finished, when clicked, then event emitted`) is covered. A future change that drops the guard would silently start firing AI chat sessions with empty/in-progress diagnostic context. Cheap to add a `given checks not finished, when clicked, then no event` test. Wdyt?

malinajirka (approved):

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
iamgabrielma: fixed https://github.com/woocommerce/woocommerce-android/pull/15883/commits/e9d55c8c648d0f865d12b4b70f64ea43ec036919

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/TroubleshootConnectionViewModel.kt
iamgabrielma: fixed https://github.com/woocommerce/woocommerce-android/pull/15883/commits/a85e12c71739e1c6b6b97756e69948ce08604445

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/TroubleshootConnectionViewModel.kt
iamgabrielma: sounds good, updsted https://github.com/woocommerce/woocommerce-android/pull/15883/commits/529ae58f4c7c615a7eb4c8533c89da2b7380853a

iamgabrielma:
> I tested the feature and found a few places where feature parity should be improved:
> 
> * In Troubleshoot Connection screen:
>   
>   * CTA Contact Support should be hidden if AI chat feature flag is enabled.
>   * CTA Ask AI should be renamed to Chat with support.
>   * Both CTA should be hidden until all checks complete.
> * In the chat screen:
>   
>   * Let's not send the connectivity result in the chat. Start with a greeting message to ask user to describe their problem.
>   * Let's not send the automatic message "Help me troubleshoot...". This together with the missing troubleshooting results in context ([commented](https://github.com/woocommerce/woocommerce-android/pull/15880#discussion_r3232365544) in a previous PR) causes the bot to mistake the problem and tries to suggest solution for login:

Thanks for the review! These have been addressed in https://github.com/woocommerce/woocommerce-android/pull/15883/commits/222852e323b4ed08cf2459321345ec18067ae662 and https://github.com/woocommerce/woocommerce-android/pull/15883/commits/4900791b41c2431caca9ab6c692daa21aec6a281 , ready for another review 👍

itsmeichigo (approved):
Thanks for the updates!

PR #15884, opened 2026-05-12
wp.com QR login via QrLoginFlow strategy pattern

malinajirka:
### Description

Adds wp.com QR login support on top of the existing site QR login baseline.

Introduces a `QrLoginFlow` strategy abstraction with two implementations:
- `SiteQrLoginFlow` — wp-admin protocol (existing behavior, extracted from the ViewModel)
- `WpComQrLoginFlow` — wp.com protocol (new)

`QrLoginFlowFactory` picks the right flow based on the scanned payload, and `QrLoginScannerViewModel` delegates to it — the VM no longer knows about REST clients or protocol details. Analytics are derived from `FlowState` transitions.

I realize this PR is huge, but considering everything, I opted for this approach. If you'd prefer me to split it into multiple smaller PRs, I can do that.

### Test Steps

**I'd consider skipping tests in this PR and testing end-to-end full test when we merge final `tracking updates` PR Wednesday/Thursday. I did run the happy path tests + mocked api error paths tests.**

1. Scan a site (wp-admin) QR code → confirm the existing approval flow still works end-to-end.
2. Scan a wp.com QR code → confirm the wp.com approval flow completes successfully.
3. Trigger error paths on both (expired code, denied approval, network error) → confirm the error screen renders correctly.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/WpComQrLoginRestClient.kt
```diff
@@ -0,0 +1,495 @@
...
+            return parsed.toStatus()
+        }
+    }
+
+    private fun buildSessionStatusRequest(sessionId: String): Request {
+        val url = WPCOMV2.auth.qr_code_app.session_status.url.toHttpUrl().newBuilder()
+            .addQueryParameter("client_id", appSecrets.appId)
+            .addQueryParameter("client_secret", appSecrets.appSecret)
+            .addQueryParameter("session_id", sessionId)
+            .build()
```
JorgeMucientes: Hmm, related to how the exchange happens during the polling the WP.com approach seems to be missing an extra security hardening that the Woo core provides. 
In `QrLoginRestClient.kt:177-184` the site flow binds session-status polling to proof of token possession (token_hash = SHA-256(token)), mirroring the [Woo Core Task 7 hardening](https://github.com/woocommerce/woocommerce/pull/64503). The WP.com flow polls with only session_id + client_id + client_secret in the GET URL, with no token-derived proof. 

Two questions before merge: 
- (1) Has the WP.com backend confirmed session_id + app credentials is sufficient for the number-matching threat model, or should we send a token proof here too? 
- (2) Can we avoid client_secret in the query string (POST body / header), since it's exposed in logs/proxies/crash traces once shipped? Or this is not critical enough given WP.com is infra we have control over?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
```diff
@@ -164,213 +141,116 @@ class QrLoginScannerViewModel @Inject constructor(
...
+            is FlowCompletion.LoggedIn -> {
+                analyticsTracker.track(AnalyticsEvent.LOGIN_QR_SUCCESS)
+                triggerEvent(Dispatch.LoggedIn(localSiteId = completion.localSiteId))
+            }
+            is FlowCompletion.OpenMagicLink -> {
+                analyticsTracker.track(AnalyticsEvent.LOGIN_QR_HANDED_OFF_WP_COM_MAGIC_LINK)
+                triggerEvent(Dispatch.OpenWpComMagicLinkUrl(url = completion.url))
+            }
         }
     }
```
JorgeMucientes, line 219: Unsure about this approach. 
We flip `loggedIn = true` before the Custom Tab opens, but opening a tab isn't completing OAuth — the user can close it, the redirect can fail, or an external browser can drop the woocommerce://magic-login app link. After that, `isIdle()` permanently rejects new scans and there's no in-flow recovery. 

Could we either (a) defer loggedIn = true until MagicLinkInterceptActivity actually completes sign-in, or (b) add a "waiting for browser sign-in" state with "Open again" / "Start over"? At minimum a test for "tab closed without redirect, then user scans again" would lock the expected behaviour.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/flow/WpComQrLoginFlow.kt
```diff
@@ -0,0 +1,220 @@
...
+        WpComQrLoginSessionStatus.Rejected -> PollOutcome.Rejected
+        WpComQrLoginSessionStatus.Expired -> PollOutcome.Expired
+        WpComQrLoginSessionStatus.Consumed -> PollOutcome.AlreadyCompleted
+    }
+
+    private fun Throwable.toPollErrorOutcome(): PollOutcome = PollOutcome.TransientError(
+        reason = toPollReason(),
+        terminal = this is WpComQrLoginSessionStatusException.RateLimited,
+        cause = this,
+    )
```
JorgeMucientes, line 171: Np.
Compared to `SiteQrLoginFlow.toPollErrorOutcome` (which marks EndpointMissing terminal and maps it to ErrorReason.EndpointMissing), the WP.com poll path only treats RateLimited as terminal and has no EndpointMissing mapping in toPollReason(), so an endpoint-missing during polling retries 4× and surfaces a generic "Unknown". Intentional, or should we mirror the site flow's terminal + mapping for parity?

JorgeMucientes (commented):
Haven't finished testing and reviewing but here's a first pass

WooCommerce/src/main/kotlin/com/woocommerce/android/network/qrlogin/WpComQrLoginRestClient.kt
malinajirka: > Can we avoid client_secret in the query string (POST body / header), since it's exposed in logs/proxies/crash traces once shipped? Or this is not critical enough given WP.com is infra we have control over?

The client secret exists for exactly this purpose right? If we didn't send it to .com backend, we wouldn't use it at all. Moreover, it's already used for the others paths and can be derived from the APK, so not sending it here doesn't help anything. It's just one extra friction for potential attackers - having it, is not enough to retrieve any unauthorized access or so.


>(1) Has the WP.com backend confirmed session_id + app credentials is sufficient for the number-matching threat model, or should we send a token proof here too?

Great question!

The number-matching threat model is closed at /exchange, which requires both the token (proof of QR-payload possession) and the exchange_grant. An attacker who only learns session_id can poll state and read the exchange_grant once approved, but can't complete login without the token — and the token is never sent in a GET URL, so it doesn't leak via access/proxy logs.

Having said that, I guess adding token_hash would be reasonable defense-in-depth, so I'll add it.

malinajirka: Fixed in 955a2583e3a12f308a45f310daf69d8f81a2d4d0. Server changes will be pushed within 10 minutes.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/flow/WpComQrLoginFlow.kt
malinajirka: The endpoint can't be missing, since there is just one server (except of load balancing). So treating it as unknown is fine imo.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/qrlogin/QrLoginScannerViewModel.kt
malinajirka: The field name is misleading, this doesn't mean we have logged in the user, but that we have dispatched the event and it's out of our control - re-trying is meaningless as magic link is a single use token.

I've fixed the UX on-back behavior and updated the field name in 2449a7bc3bc8dead7ab609011f7d2b2d209314a7

malinajirka:
I'm merging this to unblock the follow up PRs - we'll test thoroughly before the release anyway.

JorgeMucientes:
I was about to approve and merge, so you beat me to it.

PR #15888, opened 2026-05-12
[Woo POS][Phone POS] Bug: TTP buttons cut and reader-status subtitles cut on phone

samiuelson:
## Summary

Fixes four phone-only layout regressions on the Woo POS card-reader status screens (Spanish locale on a small phone):

1. **Pay with Tap to Pay** (and *Connect reader*) primary buttons were rendered as a row of dashes because the caller-side `.height(WooPosComponentSize.Small.value)` modifier collides with the inner `heightIn(min = height, max = height * 3)` in `WooPosButtons.kt::Button`, preventing the button from growing when the bold text wraps. Drop the redundant fixed `.height(...)` so the button grows vertically up to 3× when needed. Minimum height is unchanged (still enforced by the inner `heightIn`).
2. **"Tocar, deslizar o insertar tarjeta"** subtitle on the *Ready for payment* screen was clipped because it used `WooPosTypography.Heading` (36sp → 30.6sp scaled on phone). Lower to `BodyLarge` so the wrapped second line fits.
3. **"Preparando el lector para el pago"** subtitle had no horizontal padding and used the same too-large `Heading` style. Add `padding(horizontal = XLarge)` + `textAlign = Center`, and lower to `BodyLarge` for parity with the ready-for-payment subtitle.

No string changes. Tablet layout is unaffected: tablet uses the same `BodyLarge` style which renders at full 24sp; the dropped `.height(Small)` was always shadowed by the inner `Button` height clamping anyway.

## Test plan

- [ ] Build a Wasabi debug APK and install on a small phone emulator in Spanish locale.
- [ ] Open POS → start a transaction → reach the totals screen with no reader connected (TTP available): confirm the "Pay with Tap to Pay" / "Pagar con Toca para pagar" button shows full text.
- [ ] Same screen with no TTP available: confirm the "Connect reader" / "Conectar lector" button shows full text.
- [ ] Connect a reader: confirm the "Tocar, deslizar o insertar tarjeta" subtitle fully fits on screen.
- [ ] During the *Preparando el lector* transitional state: confirm the subtitle has horizontal padding and is centered.
- [ ] Sanity-check tablet emulator: no visual regression on the same screens.

## Screenshots
|Before|After|
|--|--|
|Untitled|Untitled2|

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsScreen.kt
@@ -377,8 +377,10 @@ private fun PreparingReader(title: String, subtitle: String) {
     Spacer(modifier = Modifier.height(WooPosSpacing.Medium.value))
     WooPosText(
         text = subtitle,
-        style = WooPosTypography.Heading,
-        fontWeight = FontWeight.Bold
+        style = WooPosTypography.BodyLarge,
+        fontWeight = FontWeight.Bold,
+        textAlign = TextAlign.Center,
+        modifier = Modifier.padding(horizontal = WooPosSpacing.XLarge.value)
     )
 }
 
@@ -400,7 +402,7 @@ private fun ReaderReadyForPayment(readerStatus: WooPosTotalsViewState.ReaderStat
     Spacer(modifier = Modifier.height(WooPosSpacing.Medium.value))
     WooPosText(
         text = readerStatus.subtitle,
-        style = WooPosTypography.Heading,
+        style = WooPosTypography.BodyLarge,
         fontWeight = FontWeight.Bold,
         textAlign = TextAlign.Center,
         modifier = Modifier.padding(horizontal = WooPosSpacing.XLarge.value)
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

toupper (approved):
Small, focused fix. The diagnoses all check out:

- Dropping the caller-side `.height(WooPosComponentSize.Small.value)` is the right fix — that fixed-height constraint propagates inside-out and shadows the inner `heightIn(min, max = min * 3)` in `Button`, which is exactly what prevented vertical growth on wrapping bold text. Minimum is preserved because the inner `heightIn(min = Small)` still applies, so tablet (which doesn't wrap) stays at `Small` height. ✓
- `ReaderReadyForPayment` and `PreparingReader` subtitles dropping from `Heading` (~30sp scaled) to `BodyLarge` is the obvious fix for the Spanish-locale clipping, and now both subtitles share the same Bold + Center + `XLarge` horizontal-padding treatment — nice consistency. ✓

One thing the PR description doesn't enumerate: the `weight(1f)` → `weight(if (isPhone) 2f else 1f)` change on the inner column. The description mentions "four phone-only layout regressions" then lists three, so this is presumably #4 — but it's a meaningful proportion shift (top area gets 2/3 of the vertical space on phone, totals grid gets 1/3). Worth either numbering it or adding a screenshot specifically for it so reviewers don't read past it. The change itself looks reasonable for phone where the totals grid is content-light.

Stylistic nit: `isPhone` is computed inside the layout body but used in a single `weight(...)` call — could inline or pull it earlier in `TotalsLoaded` if you ever need the same check for other phone-specific tweaks here.

LGTM.

PR #15889, opened 2026-05-12
[Woo POS][Phone POS] Bug: Email receipt button ellipsized on success screen

samiuelson:
## Summary

On the payment-success screen on a narrow phone in Spanish, the *"Enviar recibo por correo electrónico"* outlined button gets cut to *"Enviar recibo por correo electró…"*. Fix it by:

- Adding a `textStyle` parameter to `WooPosOutlinedButton` (defaults to `BodyLarge`, no behaviour change for other callers).
- Passing `WooPosTypography.BodyMedium` for the Email receipt button only — smaller text fits the localised label.
- Dropping the redundant `.height(WooPosComponentSize.Small.value)` from both success buttons so they can grow if text wraps (the inner `Button` already enforces the same minimum via `heightIn`).

No string changes. No effect on other `WooPosOutlinedButton` call sites.

## Test plan

- [ ] Complete a successful card payment; on the success screen confirm:
  - *"Enviar recibo por correo electrónico"* fits (one line, or wraps if needed) with no trailing ellipsis.
  - *"Nuevo pedido"* still fits and looks identical.
- [ ] Switch to English locale; confirm both buttons render unchanged.
- [ ] Open `WooPosPaymentSuccessScreenPreview`; confirm tablet preview is visually unchanged.

## Visual changes
Screenshot_20260513-134038
Screenshot_20260513-134054

Screenshot_20260513_141140


toupper (approved):
Tight fix. Opting into a smaller `textStyle` only at the affected call site keeps every other `WooPosOutlinedButton` consumer untouched, and dropping the hard `.height(Small)` lets the button grow vertically if the text still wraps — which combined with the new `TextAlign.Center` on the inner `Button` means multi-line content reads correctly.

One thing worth noting (not blocking): the `TextAlign.Center` on `Button` is a global change that touches every caller of `WooPosOutlinedButton` / `WooPosButton`. For single-line text it's visually identical (button has fixed shape so center-aligned and start-aligned text look the same), so this should be safe — but worth eyeballing a couple of other button-heavy POS screens just to confirm nothing reads oddly. The bottom-line — center-aligned multi-line content is strictly better than the alternative.

Conditional textStyle based on `currentWooPosBreakpoint()` at the call site reads cleanly and avoids leaking the size choice into the shared component. LGTM.

PR #15890, opened 2026-05-12
[Woo POS][Phone POS] Bug: Refund-reason placeholder cut on phone

samiuelson:
## Summary

On the *Motivo del reembolso* screen on a small phone in Spanish, the placeholder *"Motivo del pedido de reembolso"* gets hard-clipped on the right because `WooPosInputField` hardcodes `maxLines = 1, softWrap = false`.

This PR:

- Adds an opt-in `labelMaxLines: Int = 1` parameter to `WooPosInputField`. When `> 1`, the placeholder `WooPosText` renders with `maxLines = labelMaxLines`, `softWrap = true`, and `overflow = TextOverflow.Ellipsis`. Default behaviour is unchanged for all existing callers.
- Updates `WooPosRefundReasonScreen` to pass `labelMaxLines = 2` and lower `textStyle` from `Heading` (36sp → 30.6sp on phone) to `BodyLarge` (24sp → 20.4sp on phone) so the localised hint comfortably fits within two lines.

The cursor-positioning workaround inside `WooPosInputField` is untouched — only the visible label wraps; the field itself is still single-line.

## Test plan

- [ ] Open POS → orders → open an order → tap *"Emitir reembolso"* → reach the reason input screen.
- [ ] Confirm the placeholder *"Motivo del pedido de reembolso"* fits across two lines with a trailing ellipsis if the line is still longer than the available width.
- [ ] Type a few characters and confirm the input cursor and entered text render correctly (no regression to the workaround logic).
- [ ] Switch to English; the placeholder fits on one line, no visual change.
- [ ] Sanity-check tablet: refund-reason screen renders unchanged (placeholder fits on a single line as before).

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosInputFields.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.TextRange
 import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.text.input.VisualTransformation
+import androidx.compose.ui.text.style.TextOverflow
 import com.woocommerce.android.ui.compose.component.NullableCurrencyTextFieldValueMapper
 import com.woocommerce.android.ui.payments.changeduecalculator.CurrencyVisualTransformation
 import com.woocommerce.android.ui.woopos.common.composeui.WooPosPreview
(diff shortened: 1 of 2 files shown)


toupper (approved):
Clean opt-in fix. `labelMaxLines` defaults to 1 so other callers keep their single-line behaviour, and only the refund-reason screen opts into wrapping. The combination of `labelMaxLines = 2` plus dropping the font from `Heading` to `BodyLarge` is the right pair — either alone would have been borderline on the longest Spanish phrasings.

Minor — strictly speaking the diff also flips the overflow behaviour for the default `labelMaxLines = 1` path: it used to hard-clip (`softWrap = false` with no explicit overflow), and now it ellipsizes. That's strictly better visually, but the PR description's "Default behaviour is unchanged for all existing callers" isn't quite accurate. Not a problem in practice — ellipsis is always nicer than clip — but worth noting for completeness.

PR #15891, opened 2026-05-12
[Woo POS][Phone POS] Bug: Issue-refund toolbar title and selected-count overflow

samiuelson:
## Summary

Two fixes on the *Choose items to refund* step of the POS refund flow on a narrow phone in Spanish:

- **Title overflow.** *"Elige los artículos que quieres reembolsar"* was being cut to *"Elige los artículos q…"*. Lower `RefundScreenHeader` title from `Heading` to `BodyLarge` and bump `maxLines` from 1 to 2 (keep `TextOverflow.Ellipsis`) so the title fits across two lines if needed.
- **"Select all items (N seleccionado)" layout.** The count was being wrapped onto a separate line *inside* the parenthesis (`(2`, `SELECCIONA`, `DO)`), because both texts were in a `Row` with no weight or fixed width. Stack the bold label and the count vertically in a `Column` instead so each renders on its own line cleanly.

No string changes. Tablet rendering is unaffected (the title still fits on one line at the smaller `BodyLarge` size; the count layout improvement is universal).

## Test plan

- [ ] Open POS → orders → open a refundable order → tap *"Emitir reembolso"* → reach the "Choose items" step.
- [ ] Confirm the toolbar title shows the full localised text (may wrap to two lines centered).
- [ ] Confirm the *"Seleccionar todos los artículos"* row shows the count *"(N seleccionado)"* on its own line directly below the bold label.
- [ ] Repeat in English; confirm no visual regression.
- [ ] Sanity-check tablet emulator: title still fits, count still aligned next to/below the label, no broken layout.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
@@ -42,7 +42,6 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.vector.ImageVector
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.res.vectorResource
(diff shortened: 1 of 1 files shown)


toupper (approved):
Both fixes are well-scoped and breakpoint-gated. Phone gets `BodyLarge` + `maxLines = 2` for the title, and a vertical `Column` layout for the select-all + count. Tablet keeps the original `Heading` + single-line title and the inline `Row` count layout.

One thing to flag in the description: it says

> Tablet rendering is unaffected (the title still fits on one line at the smaller `BodyLarge` size; the count layout improvement is universal).

That phrasing suggests both changes apply to tablet too (just that tablet wouldn't visually regress), but the code branches both fixes on `currentWooPosBreakpoint()` so tablet keeps the existing `Heading` title and `Row` count layout. Worth tightening the description so the diff doesn't surprise the next reviewer — the actual behavior is fine, the description just slightly oversells what tablet gets.

Minor stylistic nit: `currentWooPosBreakpoint()` is invoked twice in this file (`RefundScreenHeader` and `ItemsHeaderRow` are separate composables, so the local `val isPhone` doesn't propagate). Compose-side this is cheap and idiomatic, just noting.

PR #15892, opened 2026-05-12
[Woo POS][Phone POS] Bug: Email-receipt title and email placeholder cut

samiuelson:
## Summary

> **Stacked on top of #15890** (`issue/pos-phone-refund-reason-label-wrap`) — that PR introduces the `labelMaxLines` parameter used here. Once #15890 lands and this branch is rebased onto `trunk`, GitHub will re-target the base to `trunk` automatically. (You can merge #15890 first, or merge this with the base branch updated.)

On the *Enviar recibo por correo electrónico* screen on a small phone in Spanish:

- The toolbar title was hard-clipped at *"Enviar recibo por correo"*. Lower the title style to `BodyLarge` on this call site, and let `WooPosToolbar` titles wrap to 2 lines with `TextOverflow.Ellipsis`. The toolbar's fixed `.height(XSmall)` is changed to `.heightIn(min = XSmall)` so the toolbar grows when a title wraps. Default `titleStyle` (`Heading`) and one-line layout are still the default for all existing call sites — the only screen that opts into two-line wrap is this one.
- The email placeholder *"Introducir el correo electrónico"* was hard-clipped because it used `Heading`. Lower to `BodyLarge` and pass `labelMaxLines = 2` (introduced in #15890) so it wraps and ellipsizes at the end.

No string changes. No effect on other `WooPosToolbar` consumers (default `titleStyle` is unchanged; the only behavioural change is that titles that don't fit can now wrap to a second line, which is strictly an improvement).

## Test plan

- [ ] Complete a successful card payment; on the success screen, tap *"Enviar recibo por correo electrónico"*.
- [ ] Confirm:
  - The toolbar title shows the full *"Enviar recibo por correo electrónico"*, wrapping to up to two lines, with trailing ellipsis if still longer.
  - The email-input placeholder *"Introducir el correo electrónico"* wraps to up to two lines with trailing ellipsis.
- [ ] Type a valid email; confirm the input still behaves correctly and the *Enviar* button enables/disables as before.
- [ ] Switch to English; confirm both the title and placeholder render unchanged (single line, no wrap needed).
- [ ] Sanity-check another `WooPosToolbar` consumer (e.g. another POS screen) — confirm short titles still render on a single line with no visual regression.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosToolbar.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosToolbar.kt
@@ -8,13 +8,14 @@ import androidx.compose.animation.scaleOut
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.statusBarsPadding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.constraintlayout.compose.ConstraintLayout
 import com.woocommerce.android.ui.woopos.common.composeui.WooPosPreview
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosComponentSize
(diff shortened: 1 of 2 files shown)


toupper (approved):
Reasonable fix. The `height(XSmall)` → `heightIn(min = XSmall)` change lets the toolbar grow vertically, which combined with `maxLines = 2` + `TextOverflow.Ellipsis` is strictly a visual improvement for any long-title situation. The email-receipt screen call site opts down to `BodyLarge` for both the toolbar and the input field, which is the cherry on top.

Worth tightening the description though: it claims

> Default `titleStyle` (`Heading`) and one-line layout are still the default for all existing call sites — the only screen that opts into two-line wrap is this one.

The default `titleStyle` is indeed unchanged, but the `maxLines = 2` change applies inside `WooPosToolbar` itself rather than being gated on a parameter — so every existing toolbar consumer now allows up to two lines of wrap (previously hard-clipped at one). That's a strict improvement (long titles in any locale benefit), but the description implies the wrap behavior is opt-in here only, which doesn't match the code. Worth either a one-line description update or, if you'd rather keep the change strictly opt-in, gating the new `maxLines = 2` on a parameter so only the email-receipt screen takes it.

LGTM otherwise.

PR #15893, opened 2026-05-12
[Woo POS][Phone POS] Bug: Items toolbar tabs ellipsized and crowded on phone

samiuelson:
## Summary

On a small phone, the *Productos / Cupones* toolbar tabs are ellipsized (especially in Spanish, where they currently render as *"Prod…"* and *"Cupo…"*) because they share horizontal space with the overflow menu, the *"+"* circular icon, and the search icon.

This PR fixes both visual issues from the bug report:

- **Smaller tab text on phone.** Add a `textStyle` parameter to `WooPosItemsTabsRow` (default `Heading`, no change for tablet) and pass `BodyLarge` on phone breakpoint via `WooPosItemsToolbar`.
- **Move "+" to a FAB on phone.** Add a `showAddCouponButton: Boolean = true` parameter to `WooPosItemsToolbar` (default keeps tablet behaviour). The phone screen passes `false` and instead renders a Material3 `FloatingActionButton` aligned to `BottomEnd` (with `navigationBarsPadding().imePadding()`), visible only when the active state is `WooPosItemsToolbarViewState.CouponList`. The search icon remains in the toolbar.

No string changes. Tablet UI is unaffected: the toolbar still shows the inline "+" icon and tabs at full `Heading` size.

## Test plan

- [ ] Open POS → Home (Productos tab default). Confirm:
  - "Productos" / "Cupones" tab labels are visibly smaller and not ellipsized.
  - The toolbar shows the menu (3-dot) on the left and the search icon on the right; no "+" icon.
  - No FAB on the Products tab.
- [ ] Tap the *Cupones* tab. Confirm:
  - The toolbar still has menu + search; no inline "+".
  - A "+" FAB appears at the bottom-right, sitting above the system nav bar.
  - Tapping the FAB triggers the same Add Coupon flow as before.
- [ ] Open the search overlay and confirm the FAB is still positioned correctly and the keyboard doesn't cover it (the FAB has `imePadding`).
- [ ] Switch to English. Confirm:
  - Tabs render unchanged in their smaller size and still fit.
  - FAB and toolbar layout behave identically.
- [ ] Sanity-check tablet emulator: tabs use `Heading`, *"+"* circular icon shows inline in the toolbar, no FAB. No regression.

## Visual changes
Screenshot_20260514_122516
Screenshot_20260514_122521
Screenshot_20260514_122545
Screenshot_20260514_122540

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsToolbar.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/items/WooPosItemsToolbar.kt
@@ -29,9 +29,11 @@ import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearch
 import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearchInputState
 import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosSearchUIEvent
 import com.woocommerce.android.ui.woopos.common.composeui.component.WooPosText
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosBreakpoint
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosSpacing
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTheme
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.currentWooPosBreakpoint
 import com.woocommerce.android.ui.woopos.home.items.WooPosItemsToolbarViewState.SearchState
 import com.woocommerce.android.ui.woopos.home.items.variations.WooPosVariationsNavigationData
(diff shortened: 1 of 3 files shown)


toupper (approved):
Solid two-fix bundle. Lowering the tabs to `BodyLarge` on phone via a parameter (default `Heading` so tablet is unchanged) is the right factoring, and moving the "+" from a toolbar-inline button to a Material3 FAB on phone matches the standard pattern and gets the visual breathing room the toolbar needs. `imePadding()` on the FAB so the IME doesn't occlude it is a thoughtful detail.

A couple of small things, neither blocking:

1. **The FAB appears/disappears abruptly when switching tabs.** The render is a plain `if (itemsState is CouponList)`. An `AnimatedVisibility` (scale + fade) would give a smoother transition between Products and Coupons. Pure polish.

2. **Content description string reuse.** The FAB borrows `R.string.woopos_coupons_empty_list_create_coupon_label` ("Create coupon"). Functionally fine, but a dedicated "Add coupon" string would be more semantically accurate for the FAB action — the empty-list label has a slightly different connotation.

PR #15895, opened 2026-05-12
Add Stripe smart-reader support (S700/S710/T600/T610) via internet discovery

johndcoy:
## Summary
- Adds support for Stripe internet-discovered smart readers (S700, S710, T600, T610) to the WooPos connect flow and the main Payments connect flow.
- Wires `InternetDiscoveryConfiguration` + `InternetConnectionConfiguration` through `libs/cardreader` (new `discoverInternetReaders`, `connectToInternet`, `connectToInternetReader` routing).
- Sequential discovery: internet first (short-running); bluetooth only runs if internet returned zero readers. Avoids the SDK's "operation in progress" error from concurrent calls and prevents a BT timeout from overwriting a successful internet result in the UI.

## Test plan
- [ ] End-to-end on a physical tablet on the same Wi-Fi as the reader: discovery finds the smart reader, tap to connect succeeds, payment intent runs, card tap completes, charge processes (validated locally on Pixel Tablet + S710, livemode).
- [ ] Existing Bluetooth-only allowlists (no internet readers in the list) still trigger only BT discovery — no regression for Chipper/M2/WisePad3.
- [ ] Tap to Pay (built-in) discovery path unchanged.
- [ ] All `:libs:cardreader:testDebugUnitTest` pass (329/329 locally).
- [ ] Manual smoke test: scan with no readers on the account behaves cleanly (falls through to BT scan).

## Out of scope (follow-ups)
- Reconnect-on-foreground for internet readers — TLS session drops when app backgrounds; next `createPaymentIntent` fails with "No active Reader". Mirror the existing BT auto-reconnect.
- Analytics transport label: `PaymentsFlowTracker` emits `transport=bluetooth` for connection events on internet-discovered readers; needs an `internet` value.
- Manuals/icons/strings for the new reader types in `CardReaderManualsSupportedReadersMapper`.
- iOS port (`woocommerce-ios`) — equivalent changes in `CardReaderType.swift`, `StripeCardReaderService.swift`.
- Reader-location pairing UX — currently relies on merchants registering readers via the Stripe Dashboard. Could expose a registration-code flow in WC admin.
- Custom splash/logo on smart readers — Stripe Terminal Configurations API supports per-merchant branding (1080x1920 PNG/JPG, `stripe_s710[splashscreen]=file_xxx`). Demoable feature beyond plumbing.

## Notes
- No SDK bump needed — `stripeterminal-android@5.1.1` already exposes `STRIPE_S700`, `STRIPE_S710`, `STRIPE_T600`, `STRIPE_T610`, `InternetDiscoveryConfiguration`, and `InternetReaderListener`.
- Diff: 7 files, +103/-10.

### DIFF
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/connection/ConnectionManager.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/connection/ConnectionManager.kt
@@ -5,7 +5,9 @@ import android.app.Application
 import android.content.pm.PackageManager
 import android.os.Build
 import com.stripe.stripeterminal.external.callable.Callback
+import com.stripe.stripeterminal.external.callable.InternetReaderListener
 import com.stripe.stripeterminal.external.models.ConnectionConfiguration.BluetoothConnectionConfiguration
+import com.stripe.stripeterminal.external.models.ConnectionConfiguration.InternetConnectionConfiguration
 import com.stripe.stripeterminal.external.models.ConnectionConfiguration.TapToPayConnectionConfiguration
 import com.stripe.stripeterminal.external.models.DeviceType
 import com.stripe.stripeterminal.external.models.Reader
@@ -29,6 +31,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.launch
@@ -65,17 +69,42 @@ internal class ConnectionManager(
 
                     is ExternalReaders -> {
                         checkIfNecessaryPermissionsAreGranted()
-                        discoverReadersAction.discoverExternalReaders(isSimulated)
+                        // Stripe SDK allows only one discoverReaders operation at a time. Run
+                        // internet discovery first (short-running). Skip the long-running
+                        // bluetooth scan if internet already returned a reader, so a BT timeout
+                        // doesn't overwrite a successful result in the UI.
+                        if (cardReaderTypesToDiscover.externalReaders.any { it.isInternetReader }) {
+                            flow {
+                                var foundInternetReader = false
+                                discoverReadersAction.discoverInternetReaders(isSimulated = isSimulated)
+                                    .collect { event ->
+                                        if (event is DiscoverReadersStatus.FoundReaders &&
+                                            event.readers.isNotEmpty()
+                                        ) {
+                                            foundInternetReader = true
+                                        }
+                                        emit(event)
+                                    }
+                                if (!foundInternetReader) {
+                                    emitAll(discoverReadersAction.discoverExternalReaders(isSimulated))
+                                }
+                            }
+                        } else {
+                            discoverReadersAction.discoverExternalReaders(isSimulated)
+                        }
                     }
                 }
             }
 
             UnspecifiedReaders -> {
                 checkIfNecessaryPermissionsAreGranted()
-                merge(
-                    discoverReadersAction.discoverBuildInReaders(isSimulated),
-                    discoverReadersAction.discoverExternalReaders(isSimulated)
-                )
+                // Stripe SDK serializes discoverReaders calls. Internet and Tap to Pay are
+                // short-running, so chain them before the long-running bluetooth discovery.
+                flow {
+                    emitAll(discoverReadersAction.discoverInternetReaders(isSimulated = isSimulated))
+                    emitAll(discoverReadersAction.discoverBuildInReaders(isSimulated))
+                    emitAll(discoverReadersAction.discoverExternalReaders(isSimulated))
+                }
             }
         }.map { state ->
             when (state) {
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

johndcoy:
@joshheald 👋 Are you able to help with reviewing this change? It essentially adds additional device support.

joshheald (commented):
Thanks for the PR John. 

We're keen to add support for more readers, and I think that we should offer a great experience for merchants who buy these expensive full featured readers. The [Verifone 660p](https://stripe.com/gb/terminal/v660p) is probably top of our list, because of the built in printer.

I think the real strength of the S700 line is that these readers can be used as an all-in-one device for POS. The app can run on the reader, no need for any other hardware.

Using them as a connected reader to a phone or tablet doesn't give that benefit, and it can compromise the overall implementation a little. My instinct is that we shouldn't add support for these readers until we take advantage of their hardware, because otherwise they bring no significant benefit to merchants compared to a much cheaper Bluetooth reader. That risks upsetting merchants who buy the more expensive reader thinking that we fully support it.

That said, I took a look at the code and ran it, and as mentioned in a DM, I'm looping in @malinajirka as an Android engineer for his insights too.

## Review comments

### Connection flow delays
All our merchants have Bluetooth readers or Tap to Pay at the moment, and as you mention in the code, it's not possible to search for both internet readers and bluetooth readers at the same time, we have to do the searches sequentially.

I tried out the connection flow to the bluetooth reader, and found that the delay wasn't particularly noticeable, it only lasts a second or so. That's good that it doesn't slow existing merchants down much.

However, I think that means that it's instantly failing the internet search – for people actually trying to connect an internet reader, it would be frustrating to start the connection flow before the reader's turned on, and miss the window to connect their reader, then sit on the bluetooth connection screen saying "but it is turned on! Why won't you find it?!"

If we want to support internet readers this way, I think it'd be better to have an alternative connection path, or allow people to select their type of reader at the start so that it's not prone to that failure path.

### Country support

We typically define the countries where a reader is supported in the relevant country's `CardReaderConfigForSupportedCountry`. I don't see any of the new readers added there. I'm not totally sure whether Android uses those lists for deciding what readers are available for connection, I think it might only be to show the manuals for the correct readers in settings.

S710 isn't supported in all the countries where we now have IPP, and I don't know where T610 is supported because Stripe don't have data pages for it yet. The country list changed in trunk today, but you do have the new countries in this branch already.

### Permissions

The lint job picked up a permissions issue when calling `discoverReaders` with the new configuration. I'm not an expert in Android permissions but I suspect we'll need at least location permissions (as with some of the other calls in that file) and potentially local network access permissions to check for internet connected readers on the same wifi network. If there's a new permission needed, it'll need a UI flow to introduce it and ask for it.

### Testing

I wouldn't feel comfortable shipping untested hardware support – you mentioned that you have the S710, but I think we'd want to test all the other readers added here before we actually shipped it.

The PR breaks this unit test: `CardReaderConnectViewModelTest > given external card reader type, when discovery readers, then supported readers list used FAILED`. I suspect it's because of the delay while checking for the internet readers. There's also no unit test coverage for the changes, and it seems like there should be some for this new flow.

### Release

This should be feature flagged, so we have control over the release timing and a straightforward way to remove it from a release if we find issues. One opportunity would be to use a remote feature flag, and enable it only for specific merchants who we know have the reader or want to use it in this setup. That can be done in the wpcom codebase rules files, [as we did for POS testing and launch](https://github.a8c.com/Automattic/wpcom/commit/d17e44431a7da7a6650272ad9caf78da2c279eaa).

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/connection/ConnectionManager.kt
```diff
@@ -131,6 +160,11 @@ internal class ConnectionManager(
             try {
                 val reader = when (it.cardReader.deviceType) {
                     DeviceType.TAP_TO_PAY_DEVICE -> connectToBuiltInReader(cardReader, locationId)
+                    DeviceType.WISEPOS_E,
+                    DeviceType.STRIPE_S700,
+                    DeviceType.STRIPE_S710,
+                    DeviceType.STRIPE_T600,
+                    DeviceType.STRIPE_T610 -> connectToInternetReader(cardReader)
```
joshheald, line 167: It looks like this supports more than the S710, not sure whether that's intentional based on your comments...

malinajirka:
Thanks for putting this together, @johndcoy. I'd like to second a lot of what Josh raised, but mostly I'm writing because I want to understand the goal of this work. Not trying to block anything you believe in — I'm just missing the point of introducing the support in the first place, and I'd love your help filling that in.

---------

What does this bring to merchants?

Reading through the PR, the user-visible outcome is: merchants can pair an S700/S710/T600/T610 over the network and use it like a Bluetooth reader from the phone/tablet. The S700-series costs significantly more than an M2, but in this flow it doesn't unlock anything an M2 doesn't already do — no app-on-device, no offline payments, etc.

What's the picture in your head of the merchant this helps, and what do they get that they couldn't get with an M2? A few hypotheses I came up with, but I don't know which (if any) is right:
  - Letting merchants who already own an S700-series reader re-use it rather than buying a second device.
  - Laying groundwork for a future app-on-device / offline-payments effort, with this as the first step.
  - Something else I haven't thought of.

My concern, echoing Josh, is that a merchant searching "does WooCommerce support the S700?" will see "yes" and reasonably assume they're getting the capabilities the hardware is marketed on. If we ship connect-only support, we're effectively recommending a $300+ reader to do what a $70 M2 does — and likely fielding questions from merchants asking why their expensive reader can't do the things Stripe's pages show.

--------------------

One thing I'd like to understand better is what support looks like once this is in merchants' hands. The PR adds four reader models, and the S710 is the one that's been exercised end-to-end so far. Stripe's own docs flag model-specific FAQ and known-issues sections for these readers, so each one seems to come with its own quirks.

If a merchant runs into trouble on, say, a T600, what's the path to helping them? Right now I'm not sure how I'd triage an issue on hardware I've never seen — I'd struggle to tell whether it's our integration, the Stripe SDK, or the device itself. Is there a plan for how the mobile team and HEs can get hands-on with these readers? I want to make sure we're set up to give merchants a good experience after the merge.

-------------------------

The PR description lists a substantial set of follow-ups (reconnect-on-foreground, manuals/icons, iOS parity, pairing UX). Are those things you're planning to take on as part of the RSM project, or are they genuinely out of scope? I'm asking mainly because some of these feel quite important — for example, marketing or documenting a payments feature that's Android-only tends to get confusing fast for merchants and for support, so it'd help to know what the rollout actually looks like across both apps.

Thanks!

PR #15897, opened 2026-05-13
[WOOMOB-3074] Add pre-login AI support chat entry points

itsmeichigo:
### Description
Fixes WOOMOB-3074

This adds and updates the AI support chat entry points for login-flow, Help, and Troubleshooting Connection contexts.

The launch and networking behavior supports the mixed authentication model needed on Android:

- Logged-out login-flow users open a pre-login chat mode that starts directly with the greeting and text input. These requests are sent unauthenticated, and the conversation is not persisted to site-scoped chat history.
- WP.com-authenticated users keep the normal Help chat flow with the issue picker. These requests are sent as authenticated WP.com requests and chat history persistence remains enabled.
- Application-password users can open the Help chat flow and the Troubleshooting Connection chat entry point. Their support-chat requests are sent unauthenticated because they do not have a WP.com access token.

The support chat context now tolerates missing selected-site information. If a selected site has no valid remote site id, the context omits `selectedSiteId` instead of sending an invalid blog id, while still sending the site URL, app version, platform, and troubleshooting results when available.

The chat UI now sizes message bubbles to their content with a max width cap instead of stretching every bubble across the available width.

### Test Steps
1. Enable the AI Support Chat feature flag.
2. From a login-flow Help screen, open Chat with support.
3. Verify the chat opens directly with the greeting and an enabled input field.
4. Send a message and verify the bot response appears.
5. Log in with a WP.com-authenticated Jetpack site, open Help, and verify Chat with support opens the normal issue-picker flow.
6. Log in with an application-password site, open Help, and verify Chat with support also opens the normal issue-picker flow.
7. From the application-password issue-picker flow, select an issue and verify the chat can send the initial message and receive a bot response.
8. From Troubleshooting Connection with the feature flag enabled, wait for the checks to finish and verify the CTA says Chat with support instead of Contact support.
9. Open Chat with support from Troubleshooting Connection and verify the chat starts with only the greeting, without the extra "Please describe your issue in more detail so I can help" message.
10. Send a message from the Troubleshooting Connection chat flow and verify the bot response appears.

### Images/gif

[Screen_recording_20260514_180419.webm](https://github.com/user-attachments/assets/fa28e3b0-897c-43ec-ac92-72701bd1180d)


WooCommerce/src/main/res/values/strings.xml
```diff
@@ -2410,12 +2410,12 @@
...
-    <string name="ai_support_chat_help_row_title">AI Troubleshooting</string>
+    <string name="ai_support_chat_help_row_title">Chat with support</string>
     <string name="ai_support_chat_help_row_subtitle">Chat with our AI assistant to diagnose store issues</string>
-    <string name="ai_support_chat_screen_title">AI Troubleshooting</string>
+    <string name="ai_support_chat_screen_title">Chat with support</string>
     <string name="ai_support_chat_placeholder">Coming soon</string>
-    <string name="ai_support_chat_greeting">Hi! Tell me what\'s happening with your store, and I\'ll help troubleshoot it.</string>
-    <string name="ai_support_chat_input_hint">Message AI Troubleshooting</string>
+    <string name="ai_support_chat_greeting">Hello! I\'m your Woo Mobile Support Bot. Is there anything I can help you with today?</string>
+    <string name="ai_support_chat_input_hint">Type a message...</string>
```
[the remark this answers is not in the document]
itsmeichigo: Fixed in 5a45b233fbf865eff6fb7731a0f017c5c494873a

WooCommerce/src/main/res/values/strings.xml
iamgabrielma, line 1: Do these work like in iOS, where we need to update the key as well, otherwise would serve out-of-date copy?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
iamgabrielma, line 1: From codex, my understanding is that is expected, but just in case:

> [WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt:381] canPersistChatHistory is currently set only by launch mode (false for PreLogin, defaults true elsewhere).
  For application-password users hitting the Help or ConnectivityTool flow, persistence calls go through but the WP.com endpoint will fail silently inside runCatching. Consider deriving this from auth state (e.g.
  inject AccessToken / a IsWpComAuthenticated helper) so persistence is only attempted when it can actually succeed, instead of relying on swallowed failures.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -119,12 +120,24 @@ class AiSupportChatViewModel @Inject constructor(
         }
     }
 
+    private fun startFromPreLogin() {
```
iamgabrielma, line 130: Just a consistency note here, `startFromPreLogin()` doesn't set `input = ""` like `startFromConnectivityTool()` does, no strong opinion but maybe we should stick to one or the other, at the moment is a no-op due to its default.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatActivity.kt
```diff
@@ -49,6 +49,11 @@ class AiSupportChatActivity : AppCompatActivity() {
         fun createIntent(context: Context): Intent =
             Intent(context, AiSupportChatActivity::class.java)
 
+        fun createPreLoginIntent(context: Context): Intent =
```
iamgabrielma: Codex nit:

> Two near-identical factory methods (createIntent / createPreLoginIntent) could collapse into
  createIntent(context, preLogin: Boolean = false).

iamgabrielma (approved):

WooCommerce/src/main/res/values/strings.xml
itsmeichigo: Android string resources are keyed by the name, so changing the English copy under the same key should serve the new text. I don’t think we need a new key here unless our localization tooling explicitly requires source-copy key churn.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
itsmeichigo: Updated in aa9ca4b34e145995193f4f5be414b440958e8269.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
itsmeichigo: Also updated in aa9ca4b34e145995193f4f5be414b440958e8269.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatActivity.kt
itsmeichigo: Updated in aa9ca4b34e145995193f4f5be414b440958e8269.

WooCommerce/src/main/res/values/strings.xml
malinajirka: The only time you need a new key is if you change the number of placeholders/parameters in the string (e.g. `%1$s`).

Issue #3479, opened 2021-01-26
Shipping Labels: app crashes after continuing to edit origin address when it is empty


jaclync:
**To Reproduce**

Prerequisites: the store is eligible for creating shipping labels in the app - 
1. WCS should be installed and active on the store.
2. the store is in the US. 
3. the destination address for the order is in the US.

- Create an order with a physical product in core with an empty origin address
- Launch the app
- Go to the orders tab
- Tap on the order created in core
- Tap "Create shipping label" CTA
- Tap "Continue" to verify/edit the address --> the app crashes after the dialog "Address validation in progress" is shown for a few seconds

**Stacktrace**

```
D/WordPress-API: WCShippingLabelStore: verifyAddress
D/EGL_emulation: eglMakeCurrent: 0xc9875aa0: ver 3 0 (tinfo 0xc642dab0)
I/chatty: uid=10086(com.woocommerce.android) RenderThread identical 1 line
D/EGL_emulation: eglMakeCurrent: 0xc9875aa0: ver 3 0 (tinfo 0xc642dab0)
E/Volley: [9961] BasicNetwork.performRequest: Unexpected response code 500 for https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/161199960/rest-api/?locale=en_US
E/WordPress-API: Volley error on https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/161199960/rest-api/?locale=en_US
    com.android.volley.ServerError
        at com.android.volley.toolbox.BasicNetwork.performRequest(BasicNetwork.java:205)
        at com.android.volley.NetworkDispatcher.processRequest(NetworkDispatcher.java:131)
        at com.android.volley.NetworkDispatcher.processRequest(NetworkDispatcher.java:111)
        at com.android.volley.NetworkDispatcher.run(NetworkDispatcher.java:90)
D/WooCommerce-ORDERS: com.woocommerce.android.ui.orders.shippinglabels.creation.ShippingLabelsStateMachine$Event$AddressValidationFailed@bb035b6
D/WooCommerce-ORDERS: com.woocommerce.android.ui.orders.shippinglabels.creation.ShippingLabelsStateMachine$State$OriginAddressValidationFailure@b2882b7
I/ommerce.androi: Explicit concurrent copying GC freed 14441(883KB) AllocSpace objects, 0(0B) LOS objects, 43% free, 7MB/13MB, paused 11.150ms total 314.499ms
[...]

PR #15908, opened 2026-05-14
Redesign AI Assistant confirmation cards

hichamboushaba:
### Description

Brings the Android AI Assistant confirmation cards into visual parity with iOS and adds a previously-missing affordance for bulk operations.

This branch is the tail end of the broader confirmation-preview provider refactor that landed in the earlier commits on this branch. The user-visible deltas are concentrated in the three most recent commits:

1. **`Redesign confirmation cards for improved UI and accessibility`** — replaces the stroke-outline eyebrow icons with filled colored circles (orange / green / gray for pending / confirmed / cancelled) holding a white symbol, uppercases the eyebrow text with a touch of letter-spacing, drops the per-row backgrounds and borders from diff rows so they read as plain inline `Label: ~~before~~ after`, and recolors the card surface and border to a flatter `surface` + `outlineVariant` treatment. State accent colors are sourced from the existing `woo_orange_30` / `woo_green_50` / `woo_gray_40` palette with dark-mode counterparts (`woo_orange_10` / `woo_green_20` / `woo_gray_20`).

2. **`Show bulk confirmation item IDs`** — bulk cards (e.g. "Update 3 orders") now render the affected entity IDs as a list directly under the title, capped at five visible rows with a `+N more` overflow for larger batches. iOS shows this list; Android did not.

   Each row currently shows just the ID (e.g. `#3479`), not `#3479  Jane Doe` like iOS does. That's because bulk tool-call arguments carry only IDs — names have to be fetched separately from the order / product / variation store. A follow-up Linear ticket WOOMOB-3080 will add that fetch step (orders → customer first/last, products → product title, variations → attribute summary or SKU). The same resolver will also enrich single-entity card titles (`Update order #3479` → `Update order #3479 (Jane Doe)`), which is why it's split out: it touches both the single and bulk paths, so it makes more sense as its own change than bolted onto this one.

3. **`Lift confirmation cards with subtle shadow and tonal surface`** — adds `shadowElevation = 1.dp` to the `Surface` and steps the container tone from `surface` to `surfaceContainer` so the card visibly floats above the chat thread, matching iOS and other assistant cards in the thread.

The earlier commits on this branch refactor the preview pipeline behind the scenes (provider registry, dedicated providers per tool family, builder helpers) and don't change anything user-visible.

### Test Steps

1. Trigger the assistant and ask it to make a write that requires confirmation — e.g. "Set the status of order #X to completed." Verify the **pending** card shows the orange filled-circle clock, uppercase **PENDING CHANGE** eyebrow, the diff row in plain `Status: ~~processing~~ completed` form, and an outlined Cancel + purple Confirm pair.
2. Confirm the change. Verify the card transitions to the **confirmed** state — green circle/check, **CONFIRMED** eyebrow, no actions.
3. Repeat the flow and tap **Cancel**. Verify the **cancelled** state — gray circle/X, **CANCELLED** eyebrow, de-emphasized title.
4. Ask the assistant to update several entities at once — e.g. "Mark the last 3 pending orders as completed." Verify the affected order IDs are listed between the title and the diff row.
5. Repeat step 4 with more than five entities (e.g. "Mark all out-of-stock products as in-stock" against a store with ≥6 such products). Verify only five IDs are shown and a `+N more` row appears below them.
6. Toggle device dark mode and re-check each scenario — accent colors should stay legible, the shadow becomes subtle but the tonal surface step keeps the card distinct.

### Images/gif

| Scenario | Screenshot |
|---|---|
| Pending (single update) |  Screenshot_20260514_014818|
| Confirmed | Screenshot_20260514_092432 |
| Cancelled | Screenshot_20260514_014852 |
| Bulk pending with `+N more` overflow | Screenshot_20260514_092529 |


staskus (approved):
✅ Looks and behaves like iOS now. 

Nit comments by Claude:

> P1 — AssistantConfirmationCard.kt (ConfirmationBulkEntriesSection)
> 
> The 5-cap + +N more overflow logic (entries.take(MAX_BULK_ENTRIES_VISIBLE) / overflow = entries.size - visible.size) lives entirely inside the private composable — no automated coverage, only the @Preview exercises it. Field-population and renderer pass-through are tested. Suggest extracting the cap/overflow into a pure function (visible list + overflow count) and unit-testing ≤5 / ==5 / >5. Both review engines confirmed the math is currently correct — this is purely about regression-protecting it.
> 
> P2 — strings.xml — new key ai_assistant_confirmation_bulk_entries_overflow uses the ai_assistant_* prefix while siblings use assistant_confirmation_*. Matches #15903's target rename convention, so "ahead of the rename" rather than wrong — flag for reconciliation if #15903 lands after.
> 
> P2 — AssistantConfirmationCard.kt — "#${it.id}" built inline rather than via a string resource (minor i18n; consistent with existing "${row.label}:" in the same file).
> 
> P2 (for the WOOMOB-3080 ticket) — the deferred ID→name enrichment needs two pieces: a displayName field on ConfirmationBulkEntry and a new Android resolver (no equivalent of iOS's ConfirmationSnapshotResolver bulk-fetch exists). Worth noting so the ticket isn't under-scoped.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #15911, opened 2026-05-14
Prepare WCPay terminal payments before confirmation

joshheald:
Part of: RSM-642

### Description

This is the third PR in the Android AU WooPayments card-present stack.

Stacked after #15910 and continues in #15912.

It updates the payment flow so WooPayments terminal payments can be prepared between card collection and confirmation. For AU WooPayments, this lets eftpos-capable payments be prepared before confirmation. For Canada, it keeps older WooPayments installs working by skipping preparation when the route is not available.

The PR also asks Stripe Terminal to collect the full PaymentIntent details during collection and sets `manual_preferred` for CA/AU WooPayments terminal PaymentIntents, matching the regional handling we need for Interac and eftpos-style single-step payments.

### Test Steps

Testing is possible here, but it's easier with the test tools in the final PR enabling the selection of a simulated eftpos card.

### Images/gif

N/A

### DIFF
--- a/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
+++ b/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
@@ -1,10 +1,12 @@
 package com.woocommerce.android.cardreader.internal.payments
 
+import com.stripe.stripeterminal.external.models.CardPresentDetails
 import com.stripe.stripeterminal.external.models.PaymentIntent
 import com.stripe.stripeterminal.external.models.PaymentIntentStatus
 import com.stripe.stripeterminal.external.models.PaymentIntentStatus.CANCELED
 import com.woocommerce.android.cardreader.CardReaderStore
 import com.woocommerce.android.cardreader.CardReaderStore.CapturePaymentResponse
+import com.woocommerce.android.cardreader.CardReaderStore.PreparePaymentResponse
 import com.woocommerce.android.cardreader.config.CardReaderConfigFactory
 import com.woocommerce.android.cardreader.config.CardReaderConfigForSupportedCountry
 import com.woocommerce.android.cardreader.internal.payments.actions.CancelPaymentAction
@@ -32,6 +34,8 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.flow
 
+private const val EFTPOS_AU_NETWORK = "eftpos_au"
+
 @Suppress("LongParameterList")
 internal class PaymentManager(
     private val terminalWrapper: TerminalWrapper,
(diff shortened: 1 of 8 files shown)


libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
```diff
@@ -188,16 +205,55 @@ internal class PaymentManager(
     }
 
     private suspend fun FlowCollector<CardPaymentStatus>.processPayment(
+        paymentInfo: PaymentInfo,
         paymentIntent: PaymentIntent
     ): PaymentIntent {
         emit(ProcessingPayment)
+        if (paymentInfo.terminalPaymentPreparation == PaymentInfo.TerminalPaymentPreparation.NONE) {
+            return processPaymentWithoutPreparation(paymentIntent)
```
joshheald, line 213: We keep the old flow, and use it for most payments. Only CA/AU follow this new split collect/prepare?/confirm flow.

If preferred, we could use this flow for all payments because the prepare call doesn't do anything for manual capture cards... but that might be seen as too risky for now?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/payments/PaymentManager.kt
```diff
@@ -235,14 +291,81 @@ internal class PaymentManager(
...
+    }
+
+    private suspend fun prepareTerminalPaymentIfNeeded(
+        paymentInfo: PaymentInfo,
+        paymentIntent: PaymentIntent
+    ): TerminalPaymentPreparationResult {
+        val shouldPrepare = when (paymentInfo.terminalPaymentPreparation) {
+            PaymentInfo.TerminalPaymentPreparation.NONE -> false
+            PaymentInfo.TerminalPaymentPreparation.CANADA_INTERAC -> isInteracPayment(paymentIntent)
+            PaymentInfo.TerminalPaymentPreparation.AUSTRALIA_CARD_PRESENT -> isEftposAuPayment(paymentIntent)
```
joshheald, line 317: Note that we only do this `prepare` request if we're in CA/AU, and the payment method for the presented card is (or could be) single step.

libs/cardreader/src/test/java/com/woocommerce/android/cardreader/internal/payments/PaymentManagerTest.kt
```diff
@@ -545,19 +693,40 @@ class PaymentManagerTest : CardReaderBaseUnitTest() {
...
+        network: String? = null,
+        availableNetworks: List<String>? = null,
+    ): CardPresentDetails =
+        mock<CardPresentDetails>().also {
+            brand?.let { brand -> whenever(it.brand).thenReturn(brand) }
+            network?.let { network -> whenever(it.network).thenReturn(network) }
+            availableNetworks?.let { networks ->
+                val cardNetworks = mock<CardNetworks>()
+                whenever(cardNetworks.available).thenReturn(networks)
+                whenever(it.networks).thenReturn(cardNetworks)
```
joshheald, line 726: These details are used to figure out whether a card is capable of eftpos/interac routing.

samiuelson (approved):

PR #15914, opened 2026-05-14
[RSM-3160] Persist Woo push notification type switches

irfano:
### Description

Fixes RSM-3160

This builds on the Woo push notification settings shared state foundation by persisting changes from the notification type switches on the parent screen.

The parent screen now:
- applies switch changes immediately in the shared ViewModel
- saves changed notification preferences through `PushNotificationRepository`
- sends only changed preference sections in the update request
- preserves existing detail values, such as the new-order threshold, when toggling a parent switch
- ignores stale cache emissions while a save is pending
- rolls back and shows a retry action when an update fails
- flushes pending changes when leaving the screen

### Test Steps

1. Open the smarter notification settings screen.
2. Toggle each notification type switch: New orders, New reviews, and Stock.
3. Verify the switch state updates immediately.
4. Leave the screen and return to the smarter notification settings screen.
5. Verify the changed switch state is still shown.
6. Toggle a switch and immediately leave the screen.
7. Return to the smarter notification settings screen and verify the latest switch state was saved.
8. For new orders, set a threshold in the detail screen, return to the parent screen, toggle New orders, and verify the threshold is still preserved when opening the detail screen again.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
@@ -100,6 +100,10 @@ class NotificationSettingsSharedViewModel @Inject constructor(
         saveNotificationPreferencesTrigger.tryEmit(NOTIFICATION_PREFERENCES_SAVE_DEBOUNCE_MS)
     }
 
+    fun savePendingNotificationPreferences() {
+        saveNotificationPreferencesTrigger.tryEmit(0L)
+    }
+
     fun onNotificationTypeClicked(type: NotificationType) {
         when (type) {
             NotificationType.NEW_ORDERS -> triggerEvent(OpenNewOrderNotificationSettings)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
@@ -6,13 +6,21 @@ import androidx.lifecycle.asLiveData
 import com.woocommerce.android.R
 import com.woocommerce.android.notifications.push.PushNotificationRepository
 import com.woocommerce.android.tools.SelectedSite
+import com.woocommerce.android.util.CoroutineDispatchers
 import com.woocommerce.android.viewmodel.MultiLiveEvent
 import com.woocommerce.android.viewmodel.ResourceProvider
 import com.woocommerce.android.viewmodel.ScopedViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.conflate
+import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPreferences
 import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPreferences.StoreOrderPreferences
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -6,28 +6,45 @@ import androidx.lifecycle.asLiveData
...
 @HiltViewModel
 class NotificationSettingsSharedViewModel @Inject constructor(
     savedStateHandle: SavedStateHandle,
     selectedSite: SelectedSite,
     private val pushNotificationRepository: PushNotificationRepository,
-    private val resourceProvider: ResourceProvider
+    private val resourceProvider: ResourceProvider,
+    private val coroutineDispatchers: CoroutineDispatchers
 ) : ScopedViewModel(savedStateHandle) {
+    private val site: SiteModel = selectedSite.get()
```
AdamGrzybkowski: Kinda unrelated to the changes in this PR, but I think the VM doesn't need to know about the `SiteModel` here - we could inject it in the repository or the store.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -123,7 +147,76 @@ class NotificationSettingsSharedViewModel @Inject constructor(
...
+    private suspend fun saveNotificationPreferences(preferencesToSave: WooPushNotificationPreferences) {
+        val savedPreferences = savedWooPushNotificationPreferences ?: return
+        val updateRequest = preferencesToSave.diffFrom(savedPreferences)
+        if (updateRequest.isEmpty()) {
+            return
+        }
+
+        saveInProgressWooPushNotificationPreferences = preferencesToSave
+        // Once started, let the save request finish even if the screen is closed.
+        val result = withContext(NonCancellable + coroutineDispatchers.main) {
```
AdamGrzybkowski: Small code styl nit - you can skip the `val result` and chain the `onSuccess` like this:

```kotlin
withContext(NonCancellable + coroutineDispatchers.main) {
            pushNotificationRepository.updateWooNotificationPreferences(
                site = site,
                preferences = updateRequest
            )
        }.onSuccess {
            savedWooPushNotificationPreferences = preferencesToSave
        }.onFailure {
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -123,7 +147,76 @@ class NotificationSettingsSharedViewModel @Inject constructor(
...
+            savedWooPushNotificationPreferences = preferencesToSave
+        }.onFailure {
+            if (wooPushNotificationPreferences.value != preferencesToSave) {
+                // User changed preferences after this save started, so don't rollback over the newer state.
+                return@onFailure
+            }
+            rollbackNotificationPreferences()
+            showUpdateError(preferencesToSave)
+        }
+        if (saveInProgressWooPushNotificationPreferences == preferencesToSave) {
```
AdamGrzybkowski: ❓ Can this check ever be false here?

AdamGrzybkowski (commented):
Looks good! I left some comments, but none are blocking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
irfano: Good suggestion. Done: acb262ce900fed751002227a8e1dc6153bf8b98d

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
irfano: Updated the shared VM to chain on the withContext result directly. I left the order VM as-is since that save path is moved into the shared VM in the follow-up PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
irfano: Yep, you're right. This was needed in my first implementation, but became unnecessary after the save flow was serialized. Removed it.

AdamGrzybkowski (approved):

PR #15917, opened 2026-05-14
[RSM-3275] Connect new order detail screen to shared notification state

irfano:
### Description

Part of RSM-3275

This PR connects the New Order notification detail screen to the nav-graph-scoped notification settings shared ViewModel.

The New Order detail screen now reads and updates Woo notification preference state through `NotificationSettingsSharedViewModel`, so changes made on the parent notification settings screen and the New Order detail screen stay in sync immediately. The local `NewOrderNotificationSettingsViewModel` now only owns New Order screen-specific state, like the currency symbol and cha-ching sound action.

This is intentionally limited to the New Order detail screen. Review and stock detail screens will be connected in separate follow-up PRs under the same Linear issue.

Stacked on #15914.

### Test Steps

1. Enable the `SMARTER_NOTIFICATIONS` feature flag.
2. Use a site where Woo push notifications are self-driven.
3. Go to `My store > Settings > Notifications`.
4. Toggle the `New orders` row on the parent notification settings screen.
5. Open `New orders` and verify the enable switch reflects the parent value immediately.
6. Change the New Order preference between `All orders` and `High value orders`.
7. Set a high-value threshold and leave the screen.
8. Return to the parent notification settings screen and verify the `New orders` row still reflects the current enabled state.
9. Return to `New orders` and verify the selected preference and threshold are preserved.
10. Verify the cha-ching sound action still works when the device notification channel needs fixing.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
@@ -24,6 +24,7 @@ import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPr
 import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPreferences.StoreOrderPreferences
 import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPreferences.StoreReviewPreferences
 import org.wordpress.android.fluxc.model.pushnotifications.WooPushNotificationPreferences.StoreStockPreferences
+import java.math.BigDecimal
 import javax.inject.Inject
 
 @OptIn(FlowPreview::class)
@@ -69,6 +70,8 @@ class NotificationSettingsSharedViewModel @Inject constructor(
         )
     )
     val notificationTypeItems = _notificationTypeItems.asLiveData()
+    private val _newOrderNotificationSettingsViewState = MutableStateFlow(NewOrderNotificationSettingsViewState())
+    val newOrderNotificationSettingsViewState = _newOrderNotificationSettingsViewState.asLiveData()
 
     init {
         observeWooPushNotificationPreferences()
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -73,6 +74,8 @@ class NotificationSettingsSharedViewModel @Inject constructor(
         )
     )
     val notificationTypeItems = _notificationTypeItems.asLiveData()
+    private val _newOrderNotificationSettingsViewState = MutableStateFlow(NewOrderNotificationSettingsViewState())
+    val newOrderNotificationSettingsViewState = _newOrderNotificationSettingsViewState.asLiveData()
```
AdamGrzybkowski: Do we need to keep it as a separate state and making sure this and the `wooPushNotificationPreferences` are in sync? 

What about adding only one missing piece (the threshold value) and combining it to a full state? Something like:

```kotlin 
private val displayedThresholdAmount = MutableStateFlow(BigDecimal(DEFAULT_ORDER_THRESHOLD_AMOUNT))

val newOrderNotificationSettingsViewState = combine(
    wooPushNotificationPreferences,
    displayedThresholdAmount
) { preferences, threshold ->
    preferences?.toNewOrderViewState(threshold) ?: NewOrderNotificationSettingsViewState(thresholdAmount = threshold)
}.asLiveData()

private fun WooPushNotificationPreferences.toNewOrderViewState(thresholdFallback: BigDecimal) =
    NewOrderNotificationSettingsViewState(
        notificationsEnabled = storeOrder?.enabled ?: true,
        notificationPreference = if (storeOrder?.minAmount != null)
            NewOrderNotificationPreference.HighValueOrders
        else
            NewOrderNotificationPreference.AllOrders,
        thresholdAmount = storeOrder?.minAmount ?: thresholdFallback
    )
```

irfano: Good point, updated this to avoid keeping the full state duplicated. 920559cd2c29f2d82796103d8099144f018b7496

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -92,14 +102,39 @@ class NotificationSettingsSharedViewModel @Inject constructor(
...
+    }
+
+    fun onNewOrderNotificationPreferenceChanged(preference: NewOrderNotificationPreference) {
+        val preferences = wooPushNotificationPreferences.value ?: return
+        val updatedViewState = preferences.toNewOrderNotificationSettingsViewState(
+            displayedThresholdAmount = displayedOrderThresholdAmount.value
+        ).copy(notificationPreference = preference)
+        updateDisplayedWooPushNotificationPreferences(
+            preferences.copy(storeOrder = updatedViewState.toStoreOrderPreferences())
+        )
```
AdamGrzybkowski, line 129: I think the intermediate viewState object is unnecessary here, what about something like this:

```kotlin
        val preferences = wooPushNotificationPreferences.value ?: return
        val minAmount = when (preference) {
            NewOrderNotificationPreference.AllOrders -> null
            NewOrderNotificationPreference.HighValueOrders -> displayedOrderThresholdAmount.value
        }
        updateDisplayedWooPushNotificationPreferences(
            preferences.copy(
                storeOrder = StoreOrderPreferences(
                    enabled = preferences.storeOrder?.enabled ?: true,
                    minAmount = minAmount
                )
            )
        )
```

The same could be done for the `onNewOrderThresholdAmountChanged`.

WDYT?

AdamGrzybkowski (approved):
Looks good 👍  left one suggestion.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
irfano: Agreed, that’s cleaner. I simplified it to update `StoreOrderPreferences` directly.

PR #15920, opened 2026-05-14
[RSM-3275] Connect stock detail screen to shared notification state

irfano:
### Description

Fixes RSM-3275

This PR connects the stock notification detail screen to `NotificationSettingsSharedViewModel`, so the stock parent row and stock detail switches use the same Woo push preference state and save flow.

The local `NewStockNotificationSettingsViewModel` still owns stock-screen-specific behavior that is not part of the shared preference model: loading the low stock threshold, opening the store settings web view, and showing threshold fetch errors. The shared ViewModel now owns the stock notification preferences themselves: the main stock notification switch plus low stock, out of stock, and backorder toggles.

This is stacked on the new review detail PR and keeps the change limited to the stock detail screen.

### Test Steps

1. Enable the smarter notifications flow.
2. Open notification settings and verify the Stock row reflects the current stock notification state.
3. Toggle the Stock row switch, open the Stock detail screen, and verify the main Stock switch matches the parent state.
4. Change the main Stock switch and the Low stock / Out of stock / Backorder switches from the Stock detail screen.
5. Navigate back to the parent screen and reopen Stock settings; verify the changes are kept without waiting for a delayed refresh.
6. Verify the low stock threshold still loads and the edit store settings action still opens the authenticated web view.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
@@ -18,6 +18,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.conflate
 import kotlinx.coroutines.flow.debounce
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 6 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsSharedViewModel.kt
```diff
@@ -78,6 +78,8 @@ class NotificationSettingsSharedViewModel @Inject constructor(
     val newOrderNotificationSettingsViewState = _newOrderNotificationSettingsViewState.asLiveData()
     private val _newReviewNotificationSettingsViewState = MutableStateFlow(NewReviewNotificationSettingsViewState())
     val newReviewNotificationSettingsViewState = _newReviewNotificationSettingsViewState.asLiveData()
+    private val _newStockNotificationSettingsViewState = MutableStateFlow(NewStockNotificationSettingsViewState())
+    val newStockNotificationSettingsViewState = _newStockNotificationSettingsViewState.asLiveData()
```
AdamGrzybkowski: Same suggestion here! In fact, this is even more straightforward because we don't have any other properties. The `NewStockNotificationSettingsViewState` can be derived from the current state completely, so something like this would be enough:

```kotlin
val newStockNotificationSettingsViewState = wooPushNotificationPreferences
    .map { it?.storeStock?.toNewStockViewState() ?: NewStockNotificationSettingsViewState() }
    .asLiveData()
```

irfano: Updated: 57bf9840eb68f57f11a1c9a1eccfb00009df6331

AdamGrzybkowski (approved):
LGTM :shipit:

PR #15922, opened 2026-05-15
[WOOMOB-3071] Add human support escalation to AI chat

itsmeichigo:
### Description
Fixes WOOMOB-3071

Adds the human support escalation flow to the AI support chat. The chat now reads support area metadata and escalation flags from bot responses, shows a banner when the bot asks to forward the user to human support, hides the flagged bot escalation message itself, and lets users contact support from the banner, send-error dialog, or toolbar.

The support flow pre-fills the existing Zendesk support request form with chat transcript details, support area, subject, site address, and tags. When the support area is high-confidence and Zendesk identity is already set, the app attempts direct ticket creation; otherwise it opens the pre-filled form. Ticket-created state remains in-session only. Tracking and persistence are intentionally left out for follow-up tasks.

The toolbar action is available only after the second user message receives a bot response.

### Test Steps
1. Open AI support chat and continue to the chat after diagnostics.
2. Send one message and wait for the bot response; verify the top-right Contact Support action is not shown.
3. Send a second message and wait for the bot response; verify the top-right Contact Support action appears.
4. Send a message to request human support to trigger a bot response with the human escalation flag; verify the banner appears and the flagged bot message is not shown in the chat.
5. Tap Contact Support from the banner and verify the support request form opens with the transcript and support details pre-filled. If you already set an email for contact support, the flow can send a message automatically.
6. Force a send failure (e.g. by disconnecting Internet) and  send a new message, verify that an error alert shows up with Contact Support CTA. Tap on the CTA verify the draft user message is included in the support request message.
7. Send and ticket and confirm on Zendesk that the ticket does not trigger AI response, and includes relevant tags.

### Images/gif

Entry point from toolbar and chat footer:

Screenshot_20260515_113455

New error alert: 

Screenshot_20260515_111924


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,151 @@ class AiSupportChatFragment : Fragment() {
...
+            when (event) {
+                is ContactHumanSupport -> handleContactHumanSupport(event)
+            }
+        }
+    }
+
+    private fun handleContactHumanSupport(event: ContactHumanSupport) {
+        val supportArea = event.supportArea
+        if (supportArea != null && supportArea.isHighConfidence && zendeskSettings.isIdentitySet) {
+            createTicketDirectly(event, supportArea)
```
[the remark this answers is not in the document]
itsmeichigo: Fixed in 32489cc1bb9c083042a2a15ed83315ec3580f6ca.

WooCommerce/src/main/kotlin/com/woocommerce/android/support/requests/SupportRequestFormActivity.kt
```diff
@@ -210,21 +220,50 @@ class SupportRequestFormActivity : AppCompatActivity() {
         AnalyticsTracker.track(AnalyticsEvent.SUPPORT_IDENTITY_FORM_VIEWED)
     }
 
+    private fun applyPrefill(binding: ActivitySupportRequestFormBinding) {
+        viewModel.onPrefillReceived(prefill)
```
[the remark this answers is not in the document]
itsmeichigo: Fixed in 5b8e27c04741b2fddbcfa3297d0f25c03225edde.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,151 @@ class AiSupportChatFragment : Fragment() {
...
+
+    private fun openSupportRequestForm(event: ContactHumanSupport) {
+        supportRequestLauncher.launch(
+            SupportRequestFormActivity.createIntent(
+                context = requireContext(),
+                origin = HelpOrigin.AI_TROUBLESHOOTING,
+                extraTags = ArrayList(event.extraTags),
+                preselectedTicketType = event.supportArea?.ticketType,
+                prefilledSubject = event.supportArea?.subject,
+                prefilledMessage = event.description,
```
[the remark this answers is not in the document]
itsmeichigo: Trimmed bot responses in transcript to avoid bloating: c7a92e19bd5ad5a6f63de376f1df85a94c529111

WooCommerce/src/main/kotlin/com/woocommerce/android/support/requests/SupportRequestFormActivity.kt
```diff
@@ -162,24 +174,22 @@ class SupportRequestFormActivity : AppCompatActivity() {
...
-            posBtnAction = { _, _ ->
+        setResult(Activity.RESULT_OK)
+        AlertDialog.Builder(this)
+            .setTitle(R.string.support_request_success_title)
+            .setMessage(R.string.support_request_success_message)
+            .setPositiveButton(R.string.support_request_dialog_action) { _, _ ->
                 finish()
             }
-        )
+            .show()
```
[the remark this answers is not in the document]
itsmeichigo: Reverted in 93e231a0f2970c269456425a1b060fb5c0f2dda1.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -264,6 +306,18 @@ class AiSupportChatViewModel @Inject constructor(
...
+        asReversed()
+            .firstOrNull { it.role == SupportChatRole.BOT && it.context?.supportArea != null }
+            ?.context
+            ?.supportArea
+
+    private fun List<SupportChatMessage>.shouldPromptHumanSupport(): Boolean =
+        any { it.role == SupportChatRole.BOT && it.context?.flags?.forwardToHumanSupport == true }
+
+    private fun List<SupportChatMessage>.hasBotResponse(): Boolean =
+        any { it.role == SupportChatRole.BOT }
```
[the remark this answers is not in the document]
itsmeichigo: We expect the flag comes from bot only, so ignoring this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,151 @@ class AiSupportChatFragment : Fragment() {
...
+        progressDialog?.dismiss()
+        progressDialog = null
+    }
+
+    private fun showTicketCreatedDialog() {
+        AlertDialog.Builder(requireContext())
+            .setTitle(R.string.support_request_success_title)
+            .setMessage(R.string.support_request_success_message)
+            .setPositiveButton(R.string.support_request_dialog_action) { _, _ -> requireActivity().finish() }
+            .show()
```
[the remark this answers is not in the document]
itsmeichigo: Fixed in 08118a4dc231cf4b6dc3818167367d90693d7973.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,154 @@ class AiSupportChatFragment : Fragment() {
...
+    private fun handleContactHumanSupport(event: ContactHumanSupport) {
+        val supportArea = event.supportArea
+        if (supportArea != null && supportArea.isHighConfidence && zendeskSettings.isIdentitySet) {
+            createTicketDirectly(event, supportArea)
+        } else {
+            openSupportRequestForm(event)
+        }
+    }
+
+    private fun createTicketDirectly(event: ContactHumanSupport, supportArea: SupportChatSupportArea) {
```
iamgabrielma: Should this go into the vm rather than in the fragment?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatActivity.kt
```diff
@@ -17,6 +18,9 @@ import dagger.hilt.android.AndroidEntryPoint
 @AndroidEntryPoint
 class AiSupportChatActivity : AppCompatActivity() {
     private lateinit var binding: ActivityAiSupportChatBinding
+    private var contactSupportMenuItem: MenuItem? = null
+    private var isContactSupportActionVisible = false
+    var onContactSupportClicked: (() -> Unit)? = null
```
iamgabrielma: From claude:

> AiSupportChatActivity.kt:20 — var onContactSupportClicked: (() -> Unit)? = null is a public mutable lambda on the Activity that the Fragment sets in onViewCreated and
  clears in onDestroyView. Prefer MenuProvider on the Fragment (requireActivity().addMenuProvider(..., viewLifecycleOwner)) so the Fragment owns the menu lifecycle and
  isContactSupportActionVisible / contactSupportMenuItem state on the Activity can be removed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/networking/model/SupportChatSupportArea.kt
```diff
@@ -0,0 +1,42 @@
...
+enum class SupportAreaType(val wireValue: String) {
+    MOBILE_APP("mobile-app"),
+    CARD_READER("card-reader"),
+    WOO_PAYMENTS("woopayments"),
+    WOO_COMMERCE_PLUGIN("woocommerce-plugin"),
+    OTHER_EXTENSION_PLUGIN("other-extension-plugin");
+
+    companion object {
+        fun fromWireValue(value: String?): SupportAreaType =
+            entries.firstOrNull { it.wireValue == value?.lowercase() } ?: MOBILE_APP
```
iamgabrielma, line 29: Should we add a WooLog here for when it falls back to `MOBILE_APP` on nil/unknown values?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,154 @@ class AiSupportChatFragment : Fragment() {
...
+                extraTags = ArrayList(event.extraTags),
+                preselectedTicketType = event.supportArea?.ticketType,
+                prefilledSubject = event.supportArea?.subject,
+                prefilledMessage = event.description,
+                prefilledSiteAddress = selectedSite.getIfExists()?.url.orEmpty()
+            )
+        )
+    }
+
+    private fun showProgressDialog() {
```
iamgabrielma, line 164: from claude:

>  AiSupportChatFragment.kt:141-146 — progressDialog?.isCancelable = false is set after show(). Setting it before show avoids a brief window where the dialog is
  cancelable.

iamgabrielma (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +51,154 @@ class AiSupportChatFragment : Fragment() {
...
+            activity = requireActivity(),
+            titleId = R.string.support_request_success_title,
+            messageId = R.string.support_request_success_message,
+            positiveButtonId = R.string.support_request_dialog_action,
+            posBtnAction = { _, _ -> requireActivity().finish() }
+        )
+    }
+
+    private val ContactHumanSupport.description: String
+        get() = listOf(getString(R.string.ai_support_chat_escalation_transcript_header), transcript)
```
iamgabrielma, line 191: I just noticed this after reviewing, there is no bound for the description we're sending, right? Bot paragraphs are trimmed, but the number of messages and user-message length remain unbounded. Perhaps we should add a hard transcript size/message limit.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
itsmeichigo: IMO the Fragment is appropriate for this part because it needs Context, Activity result launching, dialogs, and Zendesk SDK setup. The ViewModel already emits the intent-like event.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatActivity.kt
itsmeichigo: Fixed in ca55d4d51af5463e8fc3ddb350d7a864d00562aa.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
itsmeichigo: Fixed in 03137a6f6855ee5a6697dcc3d2ad6b0c020ecb1c.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
itsmeichigo: Added message limit in acb8924ab4aaf4de947e113241e2a4106f9c6da5 - I don't want to use character count as it can be biased by languages.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
```diff
@@ -20,6 +56,165 @@ class AiSupportChatFragment : Fragment() {
...
+    private fun observeViewEvents() {
+        viewModel.event.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is ContactHumanSupport -> handleContactHumanSupport(event)
+            }
+        }
+    }
+
+    private fun handleContactHumanSupport(event: ContactHumanSupport) {
+        val supportArea = event.supportArea
```
malinajirka: 💡 On similar note, the Fragment (view layer) is now owning a fair amount of testable business logic — the direct-vs-form routing predicate (`supportArea != null && supportArea.isHighConfidence && zendeskSettings.isIdentitySet`), the `SupportAreaType` → `TicketType` mapping, the subject string-resource mapping, and the tag list construction. With this logic in the Fragment it's only reachable via Robolectric/instrumentation, so today none of those `when` arms (`WOO_PAYMENTS`/`WOO_COMMERCE_PLUGIN`/`OTHER_EXTENSION_PLUGIN`/`MOBILE_APP`) are covered by tests — only `CARD_READER` is exercised indirectly through the VM tests.

Wdyt about moving the decision into the VM and emitting a richer `ContactHumanSupport(mode = DirectCreate | OpenForm, ticketType, subjectResId, tags, ...)`? The Fragment then just needs `getString(...)` + `startActivity`, and the routing matrix becomes unit-testable. A regression flipping `&&` to `||` or swapping two `when` arms would silently mis-route real escalations otherwise.

itsmeichigo:
@iamgabrielma I also added a fix in 4e81e084dbb7a0370d4a9fbd59e2c2946a65de40 to fix the issue with placeholder identity being used in the auto ticket creation flow.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
malinajirka: I haven't looked fully into this but from a quick check I want to second Gabriel's point - I believe this is a network request so I don't see much reason why it should be in the view layer (fragment) and not the VM. For example `SupportRequestFormViewModel` also calls the same method from a VM. It does require a context but I believe it can get away with Application Context which is available from VMs.

The fact that we orchestrate UI based on the response (e.g. hide dialog) is actually an argument for having this in the VM so it's unit testable, since in the current implementation we can't write a test to verify the dialog was hidden.

The other paths - e.g. what we do on success and what we do on failure are also not unit testable. Ideally, all this logic would be driven by the VM layer.

I don't mind if you end up using the current approach, just want to flag that it's conceptually not aligned with MVVM architecture.

malinajirka (commented):
Thanks for working on this! I don't want to block the PR or delay the release, so take all my comments as ideas for consideration not something that needs to be implemented.  Too summarize my comments, it feels like the activity/fragment are making decisions and those decisions are not unit-testable - in MVVM all those decisions would ideally be in the VM layer, not the view (activity/fragment/screen) layer.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatFragment.kt
itsmeichigo: Fixed in 7d5a4aaede3.

itsmeichigo:
Thanks for the feedback @malinajirka - I added updates to move the logic to the view model and added tests. Please let me know if you have other suggestions.

PR #15923, opened 2026-05-15
[AI Support Chat] Persist AI support chat sessionId, hasCreatedTicket, and isResolved

iamgabrielma:
Part of WOOMOB-3070

### Description

This PR introduces 86.json schema with new `sessionId`, `hasCreatedTicket`, and `isResolved ` properties for AI Support Chat feature. These will be used for bookmark persistence so we can store the session and state needed to resume chats and support later escalation/resolved-chat behavior.

###  Changes:
- Adds `sessionId`, `hasCreatedTicket`, and `isResolved` to `SupportChatBookmarkEntity`.
- Bumps the WC Room database to version 86 and adds the generated schema/migration coverage.
- Persists the Odie `sessionId` when a chat bookmark is registered or updated.
- Sends `session_id` when fetching or continuing an existing chat.
- Adds DAO/repository methods to mark a chat as ticket-created or resolved.
- Updates unit and migration tests for the new persisted fields.

This PR only adds persistence/repository plumbing. The history screen and resume UI are handled in follow-up PRs.

### Test Steps
CI should pass

@itsmeichigo

PR #15924, opened 2026-05-15
[AI Support Chat] Add AI support chat resume launch

iamgabrielma:
Part 2 of [WOOMOB-3070](https://linear.app/a8c/issue/WOOMOB-3070). Relies on https://github.com/woocommerce/woocommerce-android/pull/15923, which needs to be merged first.

### Description

This PR adds the launch path needed to resume an existing AI support chat. It introduces a new `AiSupportChatLaunchMode.Resume` mode and a matching `AiSupportChatActivity.createResumeIntent(...)` helper that carries the saved `chatId`, `botSlug`, and optional `sessionId` into the chat screen. 

When launched in resume mode, `AiSupportChatViewModel` fetches the existing transcript from Odie, rehydrates the chat state, and marks the bookmark as updated once the transcript loads successfully.

### Test Steps

Not realistically testable as stand-alone PR, since only adds the resume launch contract and vm behavior. I'd recommend testing the stack directly from the next PR: https://github.com/woocommerce/woocommerce-android/pull/15925

In this one, CI should pass.

| 1 | 2 | 3 |
|--------|--------|--------|
| Screenshot_20260518_090513 | Screenshot_20260518_091348 | Screenshot_20260518_090632  |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -133,6 +136,52 @@ class AiSupportChatViewModel @Inject constructor(
...
+                }
+                runCatching {
+                    repository.markChatAsUpdated(response.chatId, response.sessionId)
+                }
+            }.onFailure { error ->
+                WooLog.e(WooLog.T.AI, "Fetching AI support chat history failed", error)
+                _viewState.update {
+                    it.copy(
+                        isLoadingHistory = false,
+                        showSendError = true
```
itsmeichigo: This would show an error message for failure to send message, which is misleading. Please consider showing a separate alert with option to retry.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -133,6 +136,52 @@ class AiSupportChatViewModel @Inject constructor(
...
+                    )
+                }
```
malinajirka, line 288: 💡 Unless I'm missing something, on `fetchChat` failure we end up with `messages = emptyList()`, `hasProceededToChat = true`, `chatId/sessionId` still pointing at the bookmark we never actually loaded, and `showSendError = true`. That puts the user on an empty chat with a banner labelled "Unable to send your message. Check your connection and try again." — but they didn't try to send anything, and there's no retry path (Activity restart is the only recovery).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -133,6 +136,52 @@ class AiSupportChatViewModel @Inject constructor(
...
+                        chatId = response.chatId,
+                        sessionId = response.sessionId,
+                        botSlug = response.botSlug,
+                        messages = response.messages.toUiMessages(),
+                        isLoadingHistory = false,
+                        showSendError = false
+                    )
+                }
+                runCatching {
+                    repository.markChatAsUpdated(response.chatId, response.sessionId)
```
malinajirka: 🔍 `runCatching { repository.markChatAsUpdated(...) }` with no `.onFailure` swallows the throwable silently — and `runCatching` catches `Throwable`, so it'll eat `CancellationException` too (breaks structured concurrency). It's worth at least adding `.onFailure { if (it is CancellationException) throw it; WooLog.e(WooLog.T.AI, "markChatAsUpdated failed", it) }` here so future debugging has something to go on.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
iamgabrielma: Updated https://github.com/woocommerce/woocommerce-android/pull/15924/commits/268389e236c85841bd92977a6d682d1e3d990a52 , not testable till next PR in any case

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
iamgabrielma: Fixed via a retry alert: https://github.com/woocommerce/woocommerce-android/commit/268389e236c85841bd92977a6d682d1e3d990a52

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
iamgabrielma: updated https://github.com/woocommerce/woocommerce-android/pull/15924/commits/2abd81bbc5e0726810cdf67678f2077d92e4cae5

itsmeichigo (approved):

PR #15926, opened 2026-05-15
Gate QR login entry point on 10% rollout bucket

malinajirka:
### Description

Adds a per-install rollout bucket so only ~10% of installs see the in-app QR login entry point when the remote feature flag is on. The bucket is generated once (1–10) and persisted in `SharedPreferences`, so the decision is stable across restarts.

**Default-off when the remote flag is unknown.** Because the entry point lives in the login flow, the remote FF may not have loaded yet on a cold start. `QrLoginAvailability` now requires an explicit `remoteValue == true` — `null` (not yet loaded) and `false` both gate the feature off. Only a dev override bypasses this.

The bucket only restricts in-app discovery. The deep-link path (`isAvailableForDeepLink`) bypasses the bucket: if a user is already in the flow from elsewhere (scanned a QR from wp-admin with a 3rd-party camera), we honor that intent. The camera check is also skipped on this path — the user already has the token, no in-app scanning needed. The remote-flag default-off behavior still applies.

### Test Steps

1. Fresh install, remote `woo_qr_code_login` = `false` (or not yet loaded) → no QR entry point on the login prologue; scanning a wp-admin QR via 3rd-party camera should *not* enter the flow.
2. Remote `true`, no override:
   - Repeat install until your bucket is 1 → entry point appears.
   - Bucket ≠ 1 → entry point hidden.
   - From either case, scanning a QR via 3rd-party camera enters the flow (deep-link bypasses bucket).
3. Set a debug override = `true` on `QR_LOGIN` → entry point appears regardless of remote/bucket.
4. Wipe app data, open the app once → bucket is generated and persisted (verify via subsequent reads being stable).

### AI Test Results
Results of running the different combinations on emulator:

Screenshot 2026-05-15 at 12 27 02

### Images/gif

N/A — no UI change.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
@@ -210,9 +210,6 @@ class LoginActivity :
             savedInstanceState == null &&
                 intent?.action == Intent.ACTION_VIEW &&
                 intent.data?.authority == QR_LOGIN_AUTHORITY -> {
-                // Gate the deep link on the feature flag + camera, but bypass the rollout bucket:
-                // the bucket only restricts in-app discovery. A user arriving here already chose
-                // to scan a QR (e.g. from wp-admin with a 3rd-party camera), so honor that intent.
                 if (qrLoginAvailability.isAvailableForDeepLink()) {
                     intent.data?.let { uri -> handleQrLoginUri(uri) }
                 } else {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
@@ -210,9 +210,10 @@ class LoginActivity :
             savedInstanceState == null &&
                 intent?.action == Intent.ACTION_VIEW &&
                 intent.data?.authority == QR_LOGIN_AUTHORITY -> {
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
```diff
@@ -210,9 +210,10 @@ class LoginActivity :
             savedInstanceState == null &&
                 intent?.action == Intent.ACTION_VIEW &&
                 intent.data?.authority == QR_LOGIN_AUTHORITY -> {
-                // Gate the deep link on the same availability check as the in-app entry, so
-                // the system camera / 3rd-party scanner path cannot bypass the feature flag.
-                if (qrLoginAvailability.isAvailable()) {
+                // Gate the deep link on the feature flag + camera, but bypass the rollout bucket:
+                // the bucket only restricts in-app discovery. A user arriving here already chose
+                // to scan a QR (e.g. from wp-admin with a 3rd-party camera), so honor that intent.
```
JorgeMucientes: Np. The comment says the deep link is gated on “feature flag + camera”, but isAvailableForDeepLink() now intentionally skips the camera check. Minor, but worth cleaning up.

JorgeMucientes (approved):
LGTM!

PR #15927, opened 2026-05-15
[WOOMOB-3072] Add AI support chat message feedback

itsmeichigo:
### Description
Fixes WOOMOB-3072

Adds per-message feedback to AI support chat bot responses. Bot text messages now keep their Odie server message ID, show thumbs up/down controls when feedback is available, and optimistically store the selected rating in memory for the current chat session. Feedback is submitted to the Odie feedback endpoint with the chat ID, message ID, session ID, and upvote/downvote value.

This also hides the chat input while the issue picker is shown and removes the diagnostics retry button so users continue into chat after troubleshooting.

### Test Steps
1. Open AI support chat from the Help entry point.
2. Confirm the initial issue picker is shown without the message input area.
3. Select an issue and continue into chat after diagnostics.
4. Send a message and confirm bot text responses show thumbs up/down feedback controls.
5. Tap thumbs up or thumbs down and confirm the controls change to the selected Helpful or Not helpful state.
6. Confirm tapping the selected message again does not submit another rating.

### Images/gif

[Screen_recording_20260515_160426.webm](https://github.com/user-attachments/assets/384e114a-32e6-4b63-8c46-10230afcde5d)

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
@@ -213,7 +213,7 @@ private fun MessageBubble(
                 .align(if (isUser) Alignment.CenterEnd else Alignment.CenterStart)
         ) {
             Surface(
-                shape = RoundedCornerShape(16.dp),
+                shape = RoundedCornerShape(MESSAGE_BUBBLE_CORNER_RADIUS),
                 color = if (isUser) {
                     MaterialTheme.colorScheme.primary
                 } else {
@@ -262,7 +262,7 @@ private fun MessageFeedback(
             null -> {
                 IconButton(
                     onClick = { onFeedbackClicked(messageId, AiSupportChatFeedbackRating.UP) },
-                    modifier = Modifier.size(40.dp)
+                    modifier = Modifier.size(FEEDBACK_BUTTON_SIZE)
                 ) {
                     Icon(
                         painter = painterResource(R.drawable.ic_thumb_up),
@@ -272,7 +272,7 @@ private fun MessageFeedback(
                 }
                 IconButton(
                     onClick = { onFeedbackClicked(messageId, AiSupportChatFeedbackRating.DOWN) },
-                    modifier = Modifier.size(40.dp)
+                    modifier = Modifier.size(FEEDBACK_BUTTON_SIZE)
                 ) {
                     Icon(
                         painter = painterResource(R.drawable.ic_thumb_down),
@@ -302,7 +302,7 @@ private fun RatedFeedback(
         painter = painterResource(icon),
         contentDescription = null,
         tint = MaterialTheme.colorScheme.onSurfaceVariant,
-        modifier = Modifier.size(18.dp)
+        modifier = Modifier.size(RATED_FEEDBACK_ICON_SIZE)
     )
     Spacer(modifier = Modifier.width(dimensionResource(R.dimen.minor_50)))
     Text(
@@ -496,7 +496,7 @@ private fun TypingIndicator(modifier: Modifier = Modifier) {
     val typingDescription = stringResource(R.string.ai_support_chat_typing)
     BoxWithConstraints(modifier = modifier.fillMaxWidth()) {
         Surface(
-            shape = RoundedCornerShape(16.dp),
+            shape = RoundedCornerShape(MESSAGE_BUBBLE_CORNER_RADIUS),
             color = MaterialTheme.colorScheme.surface,
             border = BorderStroke(1.dp, MaterialTheme.colorScheme.outlineVariant),
             shadowElevation = 1.dp,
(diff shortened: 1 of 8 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
```diff
@@ -200,38 +207,117 @@ private fun MessageBubble(
...
+) {
+    Row(
+        modifier = modifier,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        when (rating) {
+            null -> {
+                IconButton(
+                    onClick = { onFeedbackClicked(messageId, AiSupportChatFeedbackRating.UP) },
+                    modifier = Modifier.size(40.dp)
```
iamgabrielma: Should we move all of these size numbers (40.dp, 18.dp, 16.dp) to some sort of constants?

Also worth confirming:

> AiSupportChatScreen.kt:235,247 — IconButton(modifier = Modifier.size(40.dp)) makes the thumbs up/down hit targets 40.dp. Material guidelines and accessibility require
  a minimum 48.dp touch target. Either drop the explicit size (default IconButton is 48.dp) or set it to 48.dp.

iamgabrielma (approved):
Works well 🚀

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -74,6 +74,29 @@ class AiSupportChatViewModel @Inject constructor(
...
+        launch {
+            repository.submitFeedback(
+                botSlug = state.botSlug,
+                chatId = chatId,
+                messageId = messageId,
+                sessionId = sessionId,
+                upvoted = rating == AiSupportChatFeedbackRating.UP
+            ).onFailure { error ->
+                WooLog.e(WooLog.T.AI, "Submitting AI support chat feedback failed", error)
+            }
```
malinajirka, line 99: I don't fully follow this comment from AI, but worth double-checking:

💡 The rating is written to `messageRatings` *before* the network call and the `.onFailure { }` only logs — there's no rollback. Combined with the `if (messageId in state.messageRatings) return` guard at the top and the UI swapping IconButtons for the permanent "Helpful"/"Not helpful" label once a rating exists, the failure mode is:

1. User taps thumbs up → POST fails (500, offline, auth, parse error — all collapsed into one `SupportChatRepositoryException`).
2. User sees the permanent "Helpful" label as if it was recorded.
3. They can never re-submit (the guard blocks it, and the buttons are gone).
4. Backend has no record of the rating.

Suggested fix on `onFailure`: drop the entry from `messageRatings` and surface a snackbar so the user can retry. The absence of an `ai_support_chat_feedback_submit_failed` string in this PR is itself a signal the failure path wasn't fully designed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
itsmeichigo: Good point - updated in 19a6193adbff1142d02b22b1fcf0912acc115d04.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
```diff
@@ -200,38 +207,117 @@ private fun MessageBubble(
...
+    rating: AiSupportChatFeedbackRating?,
+    onFeedbackClicked: (Long, AiSupportChatFeedbackRating) -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Row(
+        modifier = modifier,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        when (rating) {
+            null -> {
```
malinajirka: 🔍 Treating null as an expected value feels a bit off - I'd consider making rating non-nullable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
itsmeichigo: We're keeping the rating system optimistic and planning to add tracking as the main way to measure user's responses. The backend calls are additional and failures are ignored for simplicity. The iOS app is behaving similarly, so I'm keeping this for now. We can always update the error handling later if necessary.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
itsmeichigo: Separated the views to avoid nullable rating in d97bf5be184460948c7bd4ea715fd93552aa7ccb.

PR #15931, opened 2026-05-15
POS: show 'Other payment methods' as a bottom sheet

samiuelson:
### Stack

Independent PR — touches the same `WooPosTotalsScreen.kt` / `WooPosTotalsViewState.kt` / `WooPosTotalsViewModelTest.kt` as the **POS new payment methods** stack: [#15928](https://github.com/woocommerce/woocommerce-android/pull/15928) → [#15944](https://github.com/woocommerce/woocommerce-android/pull/15944) → [#15932](https://github.com/woocommerce/woocommerce-android/pull/15932) → [#15934](https://github.com/woocommerce/woocommerce-android/pull/15934). Expect small merge conflicts on the file that lands second; the conflict surface is the new `isScanToPayEnabled` / `isMarkOrderAsCompleteEnabled` fields vs. this PR's `isAllPaymentMethodsDialogVisible` → `isAllPaymentMethodsBottomSheetVisible` rename.

### Description

Replaces the existing "Other payment methods" UI in POS Checkout. Previously the picker opened as a fullscreen takeover on phones and as a centered modal dialog on tablets. It now appears as a bottom sheet that slides up from the bottom of the right (Checkout) pane on tablet and from the bottom of the screen on phone.

Why:
- Bottom sheet is the more conventional pattern for a short list of secondary actions, and is less disruptive than a fullscreen page or a centered dialog.
- On tablet, the previous centered dialog floated over the whole window. Constraining the sheet to the Checkout pane keeps the cart visible underneath and matches where the "Other payment methods" button lives.

Implementation notes:
- The sheet is rendered inline inside `WooPosTotalsScreen` rather than via Material 3 `ModalBottomSheet` (which uses a Popup window over the activity and can't be bounded by a parent pane). The inline overlay naturally inherits the totals pane width.
- Slide-up / slide-down animation via `AnimatedVisibility` + `slideInVertically` / `slideOutVertically`. The same animation plays on dismiss (back press, scrim tap) and on option selection.
- Scrim is a 40% black overlay; tapping it dismisses.
- Small horizontal margin (`WooPosSpacing.Small`) keeps the sheet slightly inset from the pane edges.
[...]

PR #15928, opened 2026-05-15
POS [1/4]: scaffold Scan to Pay and Mark Order as Complete payment methods

samiuelson:
### Stack

This PR is part of a **5-PR series** adding two iOS-parity payment methods to Woo POS:

1. 👉 **[#15928](https://github.com/woocommerce/woocommerce-android/pull/15928) — Scaffold** (this PR, base: `trunk`)
2. [#15944](https://github.com/woocommerce/woocommerce-android/pull/15944) — Mark Order as Complete domain layer (base: `feature/woopos-new-payment-methods-scaffold`)
3. [#15932](https://github.com/woocommerce/woocommerce-android/pull/15932) — Mark Order as Complete UI + integration (base: #15944's branch)
4. [#15968](https://github.com/woocommerce/woocommerce-android/pull/15968) — Scan to Pay domain layer (base: #15932's branch)
5. [#15969](https://github.com/woocommerce/woocommerce-android/pull/15969) — Scan to Pay screen and navigation wiring (base: #15968's branch)

> Related (not stacked): [#15931](https://github.com/woocommerce/woocommerce-android/pull/15931) — converts the "Other payment methods" picker to a bottom sheet. Touches the same totals files; expect minor merge conflicts depending on which lands first.

### Description

This is **PR-1 of the 5-PR series** above. It adds two iOS-parity payment methods to the Woo POS checkout flow:
- **Scan to Pay** — QR-code based payment using the order's `payment_url`. Shipped on iOS as `pointOfSaleScanToPay`.
- **Mark Order as Complete** — manual external-payment option that PUTs the order to `completed` with `payment_method=other`, `payment_method_title=Other`. Shipped on iOS as `pointOfSaleMarkOrderAsPaid`.

This PR is purely **additive scaffolding** — no new screens, no behaviour change visible to users. The per-method UI wiring (UI events, navigation, screens, repositories, polling for Scan to Pay) lands in the four follow-up PRs:

- **PR-2 (#15944)**: Mark Order as Complete domain layer — repository, ViewModel, state machine, route key, event-bus entry.
- **PR-3 (#15932)**: Mark Order as Complete UI + integration — screen, navigation route, totals/home wiring.
- **PR-4 (#15968)**: Scan to Pay domain layer — repository, ViewModel, state machine, route key, polling, shared QR composable.
- **PR-5 (#15969)**: Scan to Pay screen and navigation wiring — full-screen QR route + brightness boost + totals/home wiring.

[...]

PR #15932, opened 2026-05-15
POS [3/4]: add Mark Order as Complete screen and navigation wiring

samiuelson:
### Stack

This PR is part of the **5-PR series** adding two iOS-parity payment methods to Woo POS:

1. [#15928](https://github.com/woocommerce/woocommerce-android/pull/15928) — Scaffold (base: `trunk`)
2. [#15944](https://github.com/woocommerce/woocommerce-android/pull/15944) — Mark Order as Complete domain layer (base: `feature/woopos-new-payment-methods-scaffold`)
3. 👉 **This PR — Mark Order as Complete UI + integration** (base: `feature/woopos-mark-order-as-complete-domain`)
4. [#15968](https://github.com/woocommerce/woocommerce-android/pull/15968) — Scan to Pay domain layer (base: this PR's branch)
5. [#15969](https://github.com/woocommerce/woocommerce-android/pull/15969) — Scan to Pay screen and navigation wiring (base: #15968's branch)

> Related (not stacked): [#15931](https://github.com/woocommerce/woocommerce-android/pull/15931) — converts the picker to a bottom sheet.

### Description

**PR-3 of the 5-PR series**. Adds the user-visible piece of **Mark Order as Complete**: the full-screen confirmation route, the navigation glue, and the totals/home wiring that makes the feature reachable from the **Other payment methods** picker.

Builds on [#15944](https://github.com/woocommerce/woocommerce-android/pull/15944), which contains the domain layer (Repository, ViewModel, state, route key, event-bus entries, and analytics).

Matches iOS `pointOfSaleMarkOrderAsPaid`:
- Merchant taps **Other payment methods** → **Mark order as paid** in the totals checkout pane (visible only in debug builds for now, behind `WOO_POS_MARK_ORDER_AS_COMPLETE` from the scaffold).
- A dedicated full-screen confirmation route opens (under `ui/woopos/markorderascomplete/`), showing the order total and an optional free-form note (e.g. *"Bank transfer"*).
- On confirm, the order is PUT to `status=completed`, `payment_method=other`, `payment_method_title=Other` via the repository introduced in PR-2.
- On success the home VM transitions to `FullScreenTotals` (hides the cart pane) and the totals VM enters `PaymentSuccess`, matching the card/cash UX.
- On failure, the error renders inline with a retry-enabled button. A failing order-note request does **not** roll back the completion (the order is already done) — instead it surfaces a separate `MarkAsPaidNotePostFailed` analytic.

#### Files

- **New** — `WooPosMarkOrderAsCompleteScreen.kt`, `WooPosMarkOrderAsCompleteNavigation.kt`.
- **Strings** — 4 user-visible strings (title, body, note label, note hint). The 3 VM-used strings (total/confirm/error) live in PR-2.
- **Wiring** — `WooPosTotalsScreen.toUIEvent()` maps `MARK_ORDER_AS_PAID` to `OnMarkOrderAsCompleteClicked`; `WooPosTotalsViewModel` handles it; `ChildToParentEvent.NavigationEvent.ToMarkOrderAsComplete`; `WooPosNavigationEvent.OpenMarkOrderAsComplete`; routed through `WooPosRootHost`, `WooPosNavigationEventHandler`, and registered in `WooPosMainFlowGraph.mainGraph`.
- **2 new test tags** + small detekt-yml tweak (adds `WooPosPreview` to `UnusedPrivateMember.ignoreAnnotated` because the meta-annotation isn't followed transitively).
- **Tests** — 2 new `WooPosTotalsViewModel` tests covering the new event flow + analytic.

### Test Steps

1. Install a **wasabi debug** build.
2. Open POS, add an item to the cart, tap **Checkout**.
3. In the totals pane tap **Other payment methods** → **Mark order as paid**.
4. The full-screen confirmation route opens showing the order total and a note field.
5. Leave the note blank, tap **Mark order as complete**, confirm:
   - Order moves to **Completed** in the WordPress admin with `Payment method: Other`.
   - POS returns to totals showing the success state, **with the cart pane hidden** (full-screen totals).
6. Repeat with a note like "Bank transfer" → confirm the note shows on the order in WP admin as a non-customer order note.
7. Kill the network mid-confirm to force a failure → confirm the error message renders inline and the button re-enables for retry.
8. Toggle `WOO_POS_MARK_ORDER_AS_COMPLETE` to `false` in `FeatureFlag.kt`, rebuild, and confirm the row disappears from the **Other payment methods** dialog.
9. Run the existing Card / Tap to Pay / Cash flows once each to confirm no regression.

### Images/gif

Screenshot_20260519_174113
Screenshot_20260520_134824


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+    private fun handleConfirm() {
+        viewModelScope.launch {
+            val current = _state.value as? WooPosMarkOrderAsPaidState.Confirming ?: return@launch
+            if (!current.canConfirm || current.isProcessing) return@launch
+            analyticsTracker.track(MarkAsPaidConfirmed)
+            _state.value = current.copy(isProcessing = true, errorMessage = null)
+
+            val outcome = repository.markOrderAsPaid(orderId, current.note.takeIf { it.isNotBlank() })
+            when (outcome) {
+                MarkOrderAsPaidOutcome.SuccessWithFailedNote -> {
```
malinajirka: ❓ Should we inform the user saving the note failed?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidNavigation.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.navigation.NavGraphBuilder
+import androidx.navigation.NavType
+import androidx.navigation.compose.composable
+import androidx.navigation.navArgument
+import com.woocommerce.android.ui.woopos.home.HOME_ROUTE
+import com.woocommerce.android.ui.woopos.root.navigation.WooPosNavigationEvent
+import com.woocommerce.android.ui.woopos.root.navigation.navigateOnce
+
+const val MARK_ORDER_AS_PAID_ROUTE =
+    "$HOME_ROUTE/mark_order_as_complete/{$MARK_ORDER_AS_PAID_ROUTE_ORDER_ID_KEY}"
```
malinajirka: 🔍 Seems like a leftover still mentioning `complete`.

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidViewModel.kt
samiuelson: Checked iOS for parity: it deliberately does **not** surface the note-post failure to the merchant. In `Modules/Sources/PointOfSale/Controllers/PointOfSaleOrderController.swift`:

```swift
// Order note attaches separately from completion. We only attempt it if the merchant
// supplied content; the order completion is the critical path, and a stale note can
// always be added later via order detail.
if !trimmed.isEmpty {
    do {
        try await orderService.addOrderNote(...)
    } catch {
        DDLogWarn("⚠️ [MarkAsPaid] Order completed but failed to attach merchant note: \(error)")
    }
}
```

Warning log only — no analytic, no UI surface. The merchant's order detail screen already surfaces the note state, so a missing note can be re-added there if needed. Going with the iOS approach on Android: no user-facing message. The `MarkAsPaidNotePostFailed` analytic (which Android had added in #15944 but iOS doesn't track) is being removed in #15991 along with the rename.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidNavigation.kt
samiuelson: Good catch — same iOS-parity check led to opening #15991 which renames the entire `markorderascomplete` package on trunk back to `markorderaspaid` (iOS uses `MarkAsPaid` consistently, and the remote feature-flag key on iOS is `woo_pos_mark_order_as_paid` while Android had `woo_pos_mark_order_as_complete` — those were silently diverging cross-platform). After #15991 merges, this branch's `mark_order_as_complete` URL path becomes the odd one out — I'll fix it to `mark_order_as_paid` during the trunk-merge rebase that this branch needs anyway.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidViewModel.kt
samiuelson: Updating my earlier reply: #15991 (the trunk-side rename) was closed, we're going with the lighter approach — keep the existing `markorderascomplete` naming on trunk, just align the *UI* with iOS.

The iOS-parity finding for this question still stands: iOS deliberately doesn't surface the note-post failure to the merchant (warning log only in `PointOfSaleOrderController.swift`, with the explicit rationale that "the order completion is the critical path, and a stale note can always be added later via order detail"). So Android follows suit — no UI surface, no snackbar.

The `MarkAsPaidNotePostFailed` analytic stays on Android (we're keeping cross-platform divergence in tracking-only, since that's not user-visible and removing it would lose the existing dashboard signal). UI parity only.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidNavigation.kt
samiuelson: Updating my earlier reply: #15991 (the trunk-side rename) was closed. Resolved the leftover the simpler way — merged trunk into this branch and moved `Screen.kt` + `Navigation.kt` into the canonical `markorderascomplete/` package (commit 9d24b4a1ff4). The route path `mark_order_as_complete` is now naturally consistent with the surrounding package name and the renamed constant `MARK_ORDER_AS_COMPLETE_ROUTE_ORDER_ID_KEY`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderaspaid/WooPosMarkOrderAsPaidViewModel.kt
malinajirka: I think whether it's critical or not depends very much on the use-case, doesn't it? I can imagine, user marks the order as paid with cash and leaves a note with transaction ID from a different system. If the note isn't saved, they will never be able to pair the transaction with the order -> they might end up with some accounting troubles etc. Wdyt?

I mean it's fine with me, since chances note saving will fail but transaction succeeds feel very low - but the rationale that "stale note can always be added later via order detail" feels off to me.

samiuelson: Correcting myself: full analytics parity with iOS is the right rule. iOS doesn't have a `mark_as_paid_note_post_failed` event — it was an Android-only addition. Removed it (and its call site in `WooPosMarkOrderAsCompleteViewModel`, plus the two tests) in commit dd89bf4e9ee. The `SuccessWithFailedNote` outcome now falls through to the same success arm as plain `Success` — UI **and** analytics now match iOS exactly. All 5 remaining mark-as-paid events (`checkout_mark_as_paid_tapped`, `mark_as_paid_confirmed`, `mark_as_paid_failed`, `mark_as_paid_success`, `back_to_checkout_from_mark_as_paid`) match iOS verbatim.

PR #15937, opened 2026-05-15
[WOOMOB-3096] Use variation attributes for AI assistant card titles

hichamboushaba:
### Description
Fixes WOOMOB-3096

Variation rich cards in the AI Assistant were relying on an ambiguous variation-level `name` value for the card title. That field is not a reliable source for this card path, so cards could fall back to `Variation <id>` even when the variation had useful attributes/options such as size or color.

This PR updates the variation card data flow so the card title is built from the variation attributes first, and the parent product name is used only as supporting context.

> [!NOTE]
> Showing the parent product name intentionally diverges from the current iOS presentation. It is especially useful when the model returns variations for multiple products in the same response, as shown in the screenshot, because it keeps each variation tied to its parent product while the attributes remain the primary variation identifier.

- Changed the variation resolver to build the primary card title from compact variation attributes/options, for example `Size: M • Color: Blue`.
  Why: attribute values are the merchant-recognizable identifier for a variation, while `Variation <id>` is an implementation fallback that does not help users distinguish similar cards.

- Added parent product name support as contextual card metadata when the parent product can be resolved.

- Removed the ambiguous variation-level `name` field from the assistant variation card model and parser path.
  Why: this field was not a real source of variation display text in the current flow, and keeping it encouraged UI code to prefer a value that is usually blank or misleading.

- Preserved attribute label fields such as `CompactVariationAttribute.name` while changing the card-level title behavior.
  Why: attribute labels remain useful for formatting readable titles like `Size: M`, and the ticket only calls for removing the ambiguous card-level variation name.

- Updated the WooCommerce host renderer to apply the same attribute-first title rules when rendering assistant variation rich cards.
  Why: both the assistant payload model and the host UI need to agree on the display contract so cards do not regress to `Variation <id>` after parsing.

- Kept variation open actions wired to the same parent product and variation identifiers.
  Why: this change should only affect how variation cards are presented, not which variation detail screen opens when a user taps a card.

- Added focused unit coverage for resolver, parser, segment mapping, host rendering, telemetry-adjacent fixtures, and ViewModel card handling updates.
  Why: the behavior crosses serialization/model mapping and host rendering, so tests cover the points where a fallback title or lost parent context could be reintroduced.

### Test Steps
1. Launch the Wasabi debug app and open the AI Assistant entry point.
2. Ask for product variations for a product that has variation attributes, such as size and color.
3. Confirm variation cards use the attributes/options as the card title instead of `Variation <id>`.
4. Confirm the parent product name appears as supporting context when available.
5. Tap a variation card and confirm it opens the expected product variation detail screen.

### Images/gif
Screenshot 2026-05-15 at 15 49 25


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/289">24.8 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
:shipit: Works well. Thanks for addressing.

PR #15938, opened 2026-05-16
[RSM-3386] Support store stock notifications

irfano:
### Description

Fixes RSM-3386

Recognizes incoming `store_stock` notifications and handles them as stock notifications instead of treating them as unknown notifications.

The notification product id is used as the stock notification identity.

For now, tapping a stock notification only opens the app. Opening the related product screen will be handled in a follow-up PR.

### Test Steps

1. Trigger a stock notification for a test store.
2. Verify the notification is displayed as a stock notification.

### Images/gif


AdamGrzybkowski (approved):
Looks good 👍 The deep linking is not handled yet, right? 

I think we might have to switch the stacking logic for the `store_stock` PN. Currently, each new PN will be added to the tray, but with stock, it might make more sense to replace the PN. For example, when the user receives low stock and soon after out of stock, only the last one should be visible. WDYT? 

This doesn't have to be done in this PR - I added RSM-3452, and we can discuss there.

irfano:
> The deep linking is not handled yet, right?

Yes, I’ll handle it in a follow-up PR.

PR #15939, opened 2026-05-18
[WOOMOB-2947] Prevent QR site URL prefill crash

JorgeMucientes:
### Description
Fixes WOOMOB-2947

Scanning a tokenless QR-login payload routes the app to the site-address login screen with a prefilled `siteUrl`. The Woo override was setting that text in `onViewCreated()`, before the base `LoginSiteAddressFragment` creates `mLoginSiteAddressValidator` in `onActivityCreated()`. Setting the text triggers `afterTextChanged()`, so the validator was still null and the app crashed.

This moves the prefill and auto-submit work until after the base fragment finishes `onActivityCreated()`, so the validator exists before the text watcher runs.

### Test Steps
1. Build and install the QR login branch.
2. In the WooCommerce.com mobile-login fallback (`https://woocommerce.test/mobilelogin/?wpcom_auth=missing`) for an older WooCommerce store, scan the QR code that encodes `woocommerce://qr-login?siteUrl=...`.
3. Verify the app no longer crashes when routing to site-address login.
4. Verify the prefilled site address is submitted automatically and the login flow continues.
5. Verify the normal "Enter site address" fallback still opens without a prefilled URL.

### Images/gif
N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/overrides/WooLoginSiteAddressFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/overrides/WooLoginSiteAddressFragment.kt
@@ -1,7 +1,6 @@
 package com.woocommerce.android.ui.login.overrides
 
 import android.os.Bundle
-import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.LayoutRes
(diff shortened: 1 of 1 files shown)


malinajirka (approved):
LGTM, I noticed this crash yesterday too. Thanks!

PR #15941, opened 2026-05-18
[WOOMOB-2922] AI Assistant Headless Testing: Add smoke deterministic foundation

hichamboushaba:
### Description
Fixes WOOMOB-2922

Stack: 1/3, base: `trunk`.

This PR adds the deterministic foundation for the Android Woo AI smoke regression suite. It defines the reusable headless contracts, hard-check evaluation, baseline comparison, deterministic debug runner support, and focused unit coverage that later PRs build on without requiring live credentials, Hilt/Robolectric live wiring, or network access.

Detailed changes:

- Extends the `:libs:ai-assistant:core` headless test-fixture contracts with structured scenario metadata, hard checks, run metadata, suite results, tool traces, and approved baseline models.
- Adds strict scenario and approved-baseline parsing so smoke resources fail fast when required fields are missing.
- Adds hard-check evaluator coverage for outcomes, assistant text checks, tool call presence/absence/counts, confirmation decisions, tool result kinds, and JSON argument matching.
- Adds the baseline comparator used by later live checks, including metadata staleness, new/missing/regressed scenarios, known-failure handling, and exact known-failure shape validation.
- Adds debug-only deterministic smoke runner support in `:libs:ai-assistant:feature`: config, scenario mapping, summary rendering, artifact writing, redaction, baseline approval helper, and a stable deterministic scenario resource.
- Adds deterministic support fixtures and unit tests that run the headless loop with fake chat and fake tools, assert stable artifacts, cover direct hard-check failures, and verify changed known-failure approval shapes are rejected.

Intentionally excluded from this PR:

- Live credential parsing, selected-site bootstrap, direct JWT minting, and the live `JetpackAiChatService` factory. These are in PR #15942.
- Hilt/Robolectric `testDebug` graph and live opt-in test rule. These are in PR #15942.
- Full iOS parity `live-scenarios.json`, `live-baseline.json`, docs, sampled/focused live controls, and the `woo-ai-smoke` agent skill. These are in PR #15943.
- Assistant tool-test expectation cleanup is out of scope unless it proves required by the smoke work.

### Test Steps
CI should cover the unit tests for this foundation PR. No manual QA is required.

### Images/gif
N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #15942, opened 2026-05-18
[WOOMOB-2922] AI Assistant Headless Testing: Add AI smoke live Robolectric harness

hichamboushaba:
### Description
Fixes WOOMOB-2922

Stack: 2/3, base: PR #15941 (`issue/WOOMOB-2922-add-regression-suite-gated-model`).

This PR layers the live, no-device Robolectric smoke harness on top of the deterministic foundation from PR #15941. It keeps the live infrastructure reviewable on its own before PR #15943 adds the full iOS parity scenario and baseline set, docs, and sampled/focused workflow controls.

> [!NOTE]
> - The Android live harness uses Robolectric because the AI Assistant path is not a pure Kotlin service boundary: it depends on Android/Hilt wiring, selected-site state, preferences, resources/context, and FluxC-backed services. Robolectric keeps the run headless while still giving the test the Android runtime pieces needed to exercise that production-like integration path.
> - This differs from iOS, where the smoke harness can drive the headless assistant stack more directly from XCTest. Android pays more setup cost because it intentionally reuses FluxC and the app cache, but that also means the harness validates the same selected-site and cached store data path the feature uses in the app.

Detailed changes:

- Adds explicit live credential parsing for opt-in smoke runs using `WOO_AI_SMOKE_RUN_LIVE`, `WOO_SITE_URL`, `WOO_SITE_ID`, `WOO_USERNAME`, `WOO_APP_PASSWORD`, optional mode, output directory, and store label values.
- Adds direct Jetpack AI JWT minting with application-password basic auth, token caching/invalidation, timeout handling, and redacted error messages.
- Adds a live chat service factory that wires the direct token provider into the existing `JetpackAiChatService` transport path.
- Adds selected-site/bootstrap support that installs Robolectric application-password preferences, selects the configured site, and performs safe tool preflight checks before the live scenarios run.
- Adds the debug bridge and Hilt entry point used by the live tests to run the smoke suite, write preflight artifacts, and copy per-run artifacts into the stable `latest` directory.
- Adds the `src/testDebug` Hilt/Robolectric graph for the feature module, including app bindings and network modules required to run the real tool registry without a device.
- Adds live opt-in Robolectric tests and focused unit tests for credential parsing, JWT minting, live chat factory wiring, preflight artifact writing, and the live environment rule.
- Adds compact one-scenario `live-scenarios.json` and `live-baseline.json` resources so this PR's live harness is runnable before PR #15943 replaces them with the full parity suite.

Intentionally excluded from this PR:

- The full iOS parity live scenario set, full approved live baseline, sampled/focused controls, docs, and agent skill. Those are in PR #15943.

### Test Steps
CI should cover the unit tests and the default skipped live Robolectric entrypoints for this harness PR.

A live smoke run is optional for reviewers with `~/.woo-ai-smoke/store.env` configured; full live validation is covered by PR #15943.

### Images/gif
N/A


staskus (commented):
Stack-wide review pass on the WOOMOB-2922 stack. P1 findings below as inline comments — each tagged with the engine that surfaced it.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDebugBridge.kt
```diff
@@ -0,0 +1,122 @@
...
+import dagger.hilt.EntryPoints
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.withTimeout
+import kotlinx.serialization.encodeToString
+import java.io.File
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+object WooAiSmokeDebugBridge {
```
staskus, line 16: **From Claude — P1**

`WooAiSmokeDebugBridge` is `public` (default for top-level `object`), and `runLive(application, credentials)` has no env-gate — the opt-in lives in `WooAiSmokeCredentialSource` / `WooAiSmokeLiveEnvRule`, not here. Any code in the debug APK that constructs a `WooAiSmokeCredentialConfig` and calls `runLive` boots the full harness against the supplied site.

Release builds are safe (release variant excludes `src/debug/`), but debug APKs (wasabi/jalapeno/vanilla, Buildkite PR builds, internal distribution) materially widen the attack surface.

**Suggested fix:** any of (a) tighten visibility to `internal`, (b) move the bridge into `src/testDebug/` so it never ships in any APK, (c) gate `runLive` on `BuildConfig.DEBUG && System.getenv("WOO_AI_SMOKE_RUN_LIVE") == "true"`. (a)+(b) preferred.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDirectJwtTokenProvider.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import okhttp3.Credentials
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import java.io.IOException
+
+class WooAiSmokeDirectJwtTokenProvider(
```
staskus, line 21: **From Claude — P1**

`WooAiSmokeDirectJwtTokenProvider` is `public` (default class visibility). Same concern as `WooAiSmokeDebugBridge`: any code in the debug APK can construct one and mint a real JWT against any site URL it's given. Combined with the bridge, this widens the debug-build attack surface.

**Suggested fix:** tighten to `internal` and/or move to `src/testDebug/`.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeApplicationPasswordStore.kt
```diff
@@ -0,0 +1,20 @@
...
+internal object WooAiSmokeApplicationPasswordStore {
+    fun installRobolectricPreferences(
+        context: Context,
+        applicationPasswordsStore: ApplicationPasswordsStore,
+    ) {
+        val preferences = context.getSharedPreferences(
+            "woo-ai-smoke-application-passwords",
+            Context.MODE_PRIVATE,
+        )
+        val field = ApplicationPasswordsStore::class.java.getDeclaredField("encryptedPreferences\$delegate")
```
staskus, line 16: **From Claude — P1**

`installRobolectricPreferences` uses reflection to overwrite `ApplicationPasswordsStore.encryptedPreferences$delegate` with plain `SharedPreferences`. This downgrades the running app's encrypted credential storage to plaintext on the process-wide singleton. It's `internal`, but the bytecode (with a stable Kotlin-mangled field name) ships in every debug APK.

**Suggested fix:** move this helper (and the rest of the bridge / runner / factory) into `src/testDebug/` so it never ships in any APK, OR gate the reflection on a `BuildConfig.DEBUG && System.getenv("WOO_AI_SMOKE_RUN_LIVE") == "true"` precondition, OR implement Robolectric-only via a `@TestOnly` `ApplicationPasswordsStore` Hilt binding rather than reflection on the production class.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeLiveRobolectricApprovalTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+@Config(
+    application = HiltTestApplication::class,
+    manifest = Config.NONE,
+)
+class WooAiSmokeLiveRobolectricApprovalTest {
+    private val liveEnvRule = WooAiSmokeLiveEnvRule(
+        System.getenv() + ("WOO_AI_SMOKE_MODE" to "approve"),
```
staskus: **From Codex — P1**

Approval test unconditionally injects `WOO_AI_SMOKE_MODE=approve`. Any `:libs:ai-assistant:feature:testDebugUnitTest` invocation with `WOO_AI_SMOKE_RUN_LIVE=true` but no `--tests` filter (or one that matches both test classes) will run the approval path too, burning an extra live run and writing `approved-live-baseline.json` without explicit approval-mode opt-in.

The check test (`WooAiSmokeLiveRobolectricTest`) has the inverse concern: it passes `System.getenv()` through unfiltered, so a developer who exports `WOO_AI_SMOKE_MODE=approve` globally from a previous approval run silently re-runs as approval.

**Suggested fix:** both tests should pin their mode explicitly. The check test should override any inherited `WOO_AI_SMOKE_MODE` to `check`; the approval test should require `WOO_AI_SMOKE_MODE=approve` to already be set in the env (not inject it), so the developer must opt in per invocation.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeCredentialSource.kt
```diff
@@ -0,0 +1,73 @@
...
+    ): WooAiSmokeCredentialParseResult {
+        if (environment["WOO_AI_SMOKE_RUN_LIVE"] != "true") {
+            return WooAiSmokeCredentialParseResult.Skipped(
+                "Set WOO_AI_SMOKE_RUN_LIVE=true to run live Woo AI smoke tests."
+            )
+        }
+
+        val missingKeys = requiredKeys.filter { environment[it].isNullOrBlank() }
+        if (missingKeys.isNotEmpty()) {
+            return WooAiSmokeCredentialParseResult.Invalid(
```
staskus, line 27: **From Codex — P1 (design call)**

With `WOO_AI_SMOKE_RUN_LIVE=true` but a required credential missing, `fromEnvironment` returns `Invalid` here; `WooAiSmokeLiveEnvRule` (~line 22) then maps `Invalid` to `error(...)`. So a CI job that opts into live mode but cannot inject all four secrets (e.g. one secret rotated out of the runner) fails noisily rather than skipping.

Reasonable arguments either way:
- **Loud fail (current):** no silent fake-credentials path — opt-in means run.
- **Skip:** matches the no-opt-in path's behavior; less brittle on incomplete CI secret config.

**Suggested fix:** decide and document the contract. If `Skipped` is the intent for partial-creds-with-opt-in, change `WooAiSmokeLiveEnvRule` to map `Invalid` due to missing keys (vs malformed values) to `assumeTrue(false)`. If `Invalid` is correct, document it explicitly in the SKILL.md.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeLiveChatServiceFactoryTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        private val SAMPLE_SSE_BODY = """
+            data: {"choices":[{"delta":{"content":"Hello"}}]}
+
+            data: {"choices":[{"finish_reason":"stop"}]}
+
+            data: [DONE]
+
+        """.trimIndent()
+    }
+}
```
staskus, line 122: **From Claude — P1**

The factory test does not assert that the minted JWT (`"jwt-token"` in the test) is absent from chat-error messages. The wired `JetpackAiChatService` transport may surface `Authorization: Bearer <jwt>` on diagnostic paths.

**Suggested fix:** add a 5xx-chat-response test asserting the resulting error does NOT contain the JWT string (`"jwt-token"`) or the bearer header value. Regression here would silently mint live tokens visible in artifacts.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDirectJwtTokenProviderTest.kt
```diff
@@ -0,0 +1,119 @@
...
+            username = "merchant@example.com",
+            appPassword = "app password",
+        ),
+    )
+
+    private fun jsonResponse(body: String): MockResponse = MockResponse()
+        .setResponseCode(200)
+        .setHeader("Content-Type", "application/json")
+        .setBody(body)
+}
```
staskus, line 146: **From Claude — P1**

Redaction tests cover the 401-body and missing-token bodies, but no test exercises the IOException path (`Jetpack AI JWT network failure ... ${e.message}`). OkHttp's IOException messages often include the full URL / host.

**Suggested fix:** add a test that points `httpClient` at an unroutable host and asserts the resulting `AssistantAuthException` message does NOT contain the host substring. Without this, a network failure could leak the merchant's site URL via `e.message`.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDebugBridgeTest.kt
```diff
@@ -0,0 +1,32 @@
...
+            exit = exit,
+            preflightJson = """{"safeToolResults":[{"toolName":"orders_list","resultKind":"SUCCESS"}]}""",
+        )
+
+        assertThat(latestDirectory.resolve("preflight.json")).exists()
+        assertThat(runDirectory.resolve("preflight.json")).exists()
+        assertThat(runDirectory.resolve("preflight.json").readText())
+            .isEqualTo(latestDirectory.resolve("preflight.json").readText())
+    }
+}
```
staskus, line 62: **From Claude — P1**

32 lines of test for a 192-line file with security-sensitive orchestration. `runLive`'s catch-all `getOrElse { error -> ... redactor.redact(error.message ...) }` is the last line of defense for credential leakage and is untested.

**Suggested fix:** add a unit test that injects a failing phase and asserts the returned `failureMessage` does not contain the site URL, username, or app password from a synthetic `WooAiSmokeCredentialConfig`. The redactor itself may be tested elsewhere, but the bridge's wiring of it is not.

staskus (approved):
✅ with comments

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDebugBridge.kt
hichamboushaba: Fixed in 54fadf6.

I would treat the attack path itself as effectively a false positive: this was already constrained to debug builds, and the AI smoke harness only exercises read actions. The move still makes sense for organization and keeps smoke-only plumbing out of APK source sets.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDirectJwtTokenProvider.kt
hichamboushaba: Same as above.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeApplicationPasswordStore.kt
hichamboushaba: Same as above.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeLiveRobolectricApprovalTest.kt
hichamboushaba: Fixed in c406b2d

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeLiveChatServiceFactoryTest.kt
hichamboushaba: Fixed in c65e901

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeCredentialSource.kt
hichamboushaba: PR3 owns the skill/docs update. The intended PR2 contract is loud failure after explicit live opt-in: if the caller passes `-PwooAiSmokeRunLive=true`, missing or malformed credentials fail instead of silently skipping.

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDebugBridgeTest.kt
hichamboushaba: Fixed in c65e901

libs/ai-assistant/feature/src/testDebug/kotlin/com/woocommerce/android/aiassistant/headless/WooAiSmokeDirectJwtTokenProviderTest.kt
hichamboushaba: Fixed in c65e901

PR #15944, opened 2026-05-18
POS [2/4]: add Mark Order as Complete domain layer

samiuelson:
### Stack

This PR is part of the **5-PR series** adding two iOS-parity payment methods to Woo POS:

1. [#15928](https://github.com/woocommerce/woocommerce-android/pull/15928) — Scaffold (base: `trunk`)
2. 👉 **This PR — Mark Order as Complete domain layer** (base: `feature/woopos-new-payment-methods-scaffold`)
3. [#15932](https://github.com/woocommerce/woocommerce-android/pull/15932) — Mark Order as Complete UI + integration (base: this PR's branch)
4. [#15968](https://github.com/woocommerce/woocommerce-android/pull/15968) — Scan to Pay domain layer (base: #15932's branch)
5. [#15969](https://github.com/woocommerce/woocommerce-android/pull/15969) — Scan to Pay screen and navigation wiring (base: #15968's branch)

> Related (not stacked): [#15931](https://github.com/woocommerce/woocommerce-android/pull/15931) — converts the picker to a bottom sheet.

### Description

This PR is the **domain layer only** — no UI, no navigation route, no checkout-pane wiring. The feature is unreachable from the app surface until the follow-up PR lands.

#### Contents

- **Repository** (`WooPosMarkOrderAsCompleteRepository`) — calls `WCOrderStore.updateOrderStatusAndPaymentDetails` with `payment_method=other`, posts an optional non-customer order note as best-effort, returns `MarkOrderAsCompleteOutcome` (`SUCCESS` or `SUCCESS_WITH_FAILED_NOTE`) so the VM can surface a separate analytic when the note posting fails.
- **ViewModel** (`WooPosMarkOrderAsCompleteViewModel`) — confirm flow, in-memory error surfacing, process-death recovery (a `Confirming(LOADING)` restored from saved state is reset to `ENABLED`). On success emits `ChildToParentEvent.OrderSuccessfullyPaidExternally` so `WooPosHomeViewModel` can flip the layout to full-screen totals and broadcast `OrderSuccessfullyPaid(EXTERNAL)` to children (matching the card flow).
- **State** sealed class + route-id constant (`Initiating`, `Confirming` with `Button(ENABLED|LOADING)`).
- **UI events** sealed class (`NoteChanged`, `ConfirmClicked`).
- **Event bus** entry `ChildToParentEvent.OrderSuccessfullyPaidExternally` + matching handler in `WooPosHomeViewModel`.
- **Analytics** event `MarkAsPaidNotePostFailed` for the silent note-post failure path.
- **Strings** used by the VM (`_total`, `_confirm_button`, `_error_message`).
- **Tests**: full Repository coverage (happy/failure/note-fail/synthetic-status-fallback) + VM coverage (success, repo failure, note-fail analytic, back-clicked, process-death recovery).

### Test Steps

No user-visible changes. Manual verification is in the follow-up PR. Reviewer focus here:
* Domain tests pass: `./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "com.woocommerce.android.ui.woopos.markorderascomplete.*"`

### Images/gif

N/A — no UI.

- [x] I have considered if this change warrants release notes. The release-notes entry for the new feature flags lives in the scaffold PR #15928.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderascomplete/WooPosMarkOrderAsCompleteViewModel.kt
```diff
@@ -0,0 +1,154 @@
...
+                    status = WooPosMarkOrderAsCompleteState.Confirming.Button.Status.LOADING,
+                ),
+                errorMessage = null,
+            )
+
+            val outcome = repository.markOrderAsComplete(orderId, current.note.takeIf { it.isNotBlank() })
+            when (outcome) {
+                MarkOrderAsCompleteOutcome.SuccessWithFailedNote -> {
+                    analyticsTracker.track(MarkAsPaidNotePostFailed)
+                    onMarkAsPaidSucceeded()
```
malinajirka, line 129: ❓ When the note post fails (`SuccessWithFailedNote`) we track an analytic but otherwise handle it exactly like `Success` — emit `OrderSuccessfullyPaidExternally` and navigate away. So a merchant who deliberately typed a note sees a clean success and never learns it didn't save.

Fine for the domain layer, and I see the UI lands in the follow-up PR — flagging it now so the decision isn't lost: worth surfacing this to the merchant (e.g. a snackbar on the destination) rather than only to analytics. The dedicated `SuccessWithFailedNote` outcome already makes that easy. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
```diff
@@ -192,6 +192,10 @@ class WooPosHomeViewModel @Inject constructor(
                         PaymentMethod.CARD
                     )
 
+                    is ChildToParentEvent.OrderSuccessfullyPaidExternally -> onOrderSuccessfullyPaid(
```
malinajirka, line 195: 💡 This new branch isn't covered in `WooPosHomeViewModelTest` — the sibling `OrderSuccessfullyPaidByCard` and `OnPaymentCompletedViaCash` branches both have tests.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/markorderascomplete/WooPosMarkOrderAsCompleteRepositoryTest.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    private val selectedSite: SelectedSite = mock()
+    private val orderStore: WCOrderStore = mock()
+    private val orderMapper: OrderMapper = mock()
+
+    private lateinit var repository: WooPosMarkOrderAsCompleteRepository
+
+    @Before
+    fun setUp() {
+        repository = WooPosMarkOrderAsCompleteRepository(selectedSite, orderStore, orderMapper)
```
malinajirka: 🔍 The test methods are relatively long - I'd consider mocking the happy path in the setUp method, so each test method doesn't need to mock everything.

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeViewModel.kt
samiuelson: Good catch — covered in 746d802b40b with a test that mirrors `OnPaymentCompletedViaCash`: emits an `OrderSuccessfullyPaidExternally` child→parent event and asserts the VM (a) sends `OrderSuccessfullyPaid(EXTERNAL)` back to children, (b) plays cha-ching, and (c) sets the screen position to `FullScreenTotals`.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/markorderascomplete/WooPosMarkOrderAsCompleteRepositoryTest.kt
samiuelson: Done in 746d802b40b — hoisted the shared `site`, `statusModel`, `selectedSite.get()` stub, and the happy-path `updateOrderStatusAndPaymentDetails` stub into `setUp` (wrapped in `runBlocking` because `getOrderStatusForSiteAndKey` and `updateOrderStatusAndPaymentDetails` are suspending). Each test now only stubs what it specifically diverges from — failure paths override the update flow, the note tests stub `postOrderNote`, the synthetic-status test overrides the status lookup to null. Net: -85/+60 lines and each method reads as just its GIVEN/WHEN/THEN delta.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/markorderascomplete/WooPosMarkOrderAsCompleteViewModel.kt
samiuelson: Agree the merchant should know. Leaving the domain layer intact in this PR (the `SuccessWithFailedNote` outcome is already plumbed through so the follow-up UI PR has everything it needs). I'll address the UX in the follow-up where the screen lives — likely a snackbar on the destination so the success/navigation still happens but the merchant sees the note didn't save. Will reference this thread there.

samiuelson: Follow-up shipped as #15988 — the Payment Success screen now shows a secondary line ("The order was marked as complete, but the customer note couldn't be saved.") below the order-total message when the outcome is `SuccessWithFailedNote`. The plumbing flows through a new child→parent event variant + a `hasFailedNote` flag on the existing `OrderSuccessfullyPaid` parent→child event. Landing as a separate PR off trunk rather than in #15932 because that branch needs its own rename/state-shape reconciliation first; #15932 will plug into the new wiring once rebased.

PR #15949, opened 2026-05-19
[WOOMOB-3079] Format markdown in AI support bot responses

itsmeichigo:
### Description
Fixes WOOMOB-3079

AI support chat bot responses can include Markdown for emphasis and links, but the support chat screen rendered those responses as plain text. This reuses the existing AI assistant Markdown renderer for bot text messages in AI support chat, while keeping user messages and static support chat prompts as plain text.

### Test Steps
1. Open Help & Support and start an AI support chat.
2. Send a prompt that returns a bot response containing Markdown, such as a link or bold/italic text.
3. Confirm the bot response renders formatted Markdown and links are tappable.
4. Confirm user-sent messages still render as plain text.

### Images/gif
image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatScreen.kt
@@ -56,6 +56,7 @@ import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.woocommerce.android.R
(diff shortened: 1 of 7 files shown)


hichamboushaba (approved):
Pre-approving this, it works well, and it's an acceptable solution if we want to move fast.

But, ideally, we should not make the internal components of the `:libs:ai-assistant` modules public, for this change, I would suggest moving the `AssistantMarkdownText` to the `:libs:commons` module (including the parser component) to make it reusable by all modules (including the app host module). We can also think of renaming it to `MarkdownText` and dropping the `Assistant` prefix.

itsmeichigo:
Thanks for the suggestion @hichamboushaba! I added changes following your advice and confirmed that the markdown styling still works on AI assistant chat:

image

itsmeichigo:
I'm merging now to unblock, let me know if something else needs to be updated, I'll do it in a subsequent PR 🙇

PR #15951, opened 2026-05-19
[RSM-3534] Sync smarter notification settings with system permissions

irfano:
### Description

Fixes RSM-3534

Updates the smarter push notification settings screen to respect system notification settings.

When app notifications are disabled, the notification type rows are disabled and the existing device settings warning remains available. When an individual notification channel is disabled from system settings, the related row is shown as disabled with a device-settings subtitle, and tapping the row or switch opens that channel’s system settings. The screen refreshes these states when returning from system settings.

### Test Steps

1. Disable app notifications from system settings, open **Push notifications**, and verify the notification type rows cannot be opened or toggled.
2. Enable app notifications, disable one notification channel from system settings, return to **Push notifications**, and verify that row is disabled with the “Turn on in device settings” subtitle.
3. Tap the disabled row or switch and verify the related system notification channel settings screen opens.
4. Re-enable the channel, return to **Push notifications**, and verify the row and switch become active again.

### Images/gif

Screenshot_20260519_162343 Screenshot_20260519_143203 

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsFragment.kt
@@ -2,6 +2,7 @@ package com.woocommerce.android.ui.prefs.notifications
 
 import android.content.Intent
 import android.os.Bundle
+import android.provider.Settings
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -52,6 +53,9 @@ class NotificationSettingsFragment : BaseFragment() {
     override fun onResume() {
         super.onResume()
         viewModel.refreshNotificationSettings()
+        if (navArgs.showSmarterNotifications) {
+            sharedViewModel.refreshNotificationChannelSettings()
+        }
         AnalyticsTracker.trackViewShown(this)
     }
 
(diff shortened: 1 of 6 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsFragment.kt
```diff
@@ -52,6 +53,9 @@ class NotificationSettingsFragment : BaseFragment() {
     override fun onResume() {
         super.onResume()
         viewModel.refreshNotificationSettings()
+        if (navArgs.showSmarterNotifications) {
+            sharedViewModel.refreshNotificationChannelSettings()
+        }
```
AdamGrzybkowski, line 58: ❓ Shouldn't we run the `viewModel.refreshNotificationSettings()` in the else branch?

AdamGrzybkowski (approved):
LGTM - only one question, but that shouldn't be a blocker.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NotificationSettingsFragment.kt
irfano: We still need this in both paths because the smarter screen also observes `isAppNotificationsEnabled`. The shared VM refresh is only for channel-level state, so that one is gated by `showSmarterNotifications`.

PR #15952, opened 2026-05-19
[WOOMOB-3080] Resolve entity display names for AI Assistant confirmation cards

hichamboushaba:
### Description
Fixes WOOMOB-3080

AI Assistant confirmation cards were showing affected orders, products, and variations mostly by ID. That makes a confirmation harder to review, especially for bulk actions where several rows can look nearly identical.

This PR resolves display names before the confirmation card is shown and uses them in both places where the merchant reviews the pending action:
- the single-entity card summary, for example `Update order #3479 (Jane Doe)`
- bulk affected-item rows, for example `#3479  Jane Doe`

The lookup is cache-first and failure-tolerant. If the assistant already has the entity from a previous tool call, the confirmation preview can use that data without forcing a new network fetch. If lookup data is unavailable or an error occurs, the card still appears with the existing ID-only fallback.

Resolved display names now cover:
- Orders: billing name first, then registered customer email, then `Customer #id`; guest orders with no billing name display `Guest`.
- Products: product name.
- Variations: selected attribute options, with SKU as a fallback.

The preview is resolved before the card enters the pending confirmation state, so the UI does not flash an ID-only card and then update it with names.

### Test Steps
1. Open WooCommerce and launch AI Assistant.
2. Trigger an order confirmation for an order with a billing name; confirm the title and affected row include the billing name.
3. Trigger or inspect order confirmations with no billing name; confirm guests show `Guest`, registered customers with email show the email, and registered customers without email show `Customer #id`.
4. Trigger or inspect product and variation confirmations; confirm product names and variation attribute/SKU labels are shown when available.
5. Confirm lookup fallback behavior still leaves the confirmation usable with ID-only text when entity data is unavailable.

### Images/gif
Screenshot 2026-05-19 at 16 19 18
Screenshot 2026-05-19 at 16 17 46


staskus (approved):
✅ Works well, thanks for the improvement!

PR #15953, opened 2026-05-19
Fix POS totals screen content clipping on small phones

samiuelson:
### Description
On phone landscape (and short phone portrait), the POS Totals "Checkout" view was clipping content — the "Reader not connected" subtitle was cut off, and on small phones the totals grid's "Total" row was also clipped because the inner `Column.weight(2f)` (reader area) and `Column.weight(1f)` (totals grid) couldn't both fit alongside the bottom payment buttons in the remaining height.

Replaced the weighted split with a `BoxWithConstraints` + scrollable `Column` whose `heightIn(min = maxHeight)` lets the content centre with `Arrangement.SpaceEvenly` when there is room (current tablet / tall-phone behaviour preserved) and scroll when there isn't (small phones no longer clip).

Extracted the reader-status `when` block into a `ReaderStatusContent` composable for readability, and dropped the now-redundant inner `verticalScroll` from `ReaderDisconnected` (the parent scrolls). `CheckoutPaymentButtons` remain pinned at the bottom so Cash / Other payment methods are always one tap away.

Added a thin (1dp) scroll-edge shadow above the payment buttons and below the back-button row that fades in only while content is hidden behind that edge (`scrollState.canScrollForward` / `canScrollBackward`), so the boundary between the pinned regions and the scrollable middle reads clearly when overflow exists.

### Test Steps
- [ ] Open POS on a tablet in landscape, add an item to the cart, and verify the Totals screen looks unchanged: reader-status block in the upper area, totals grid below, payment buttons at the bottom, no scrolling artifacts.
- [ ] Open POS on a phone in landscape, add an item to the cart, and verify the "Reader not connected" title, full subtitle ("To process this payment, please connect your reader."), and "Connect to reader" button are all fully visible (scroll if the device is very short).
- [ ] On a small phone (the layout shown in the original report), verify the totals grid (Subtotal / Taxes / Total) is fully visible — scroll if needed; the "Total" row must no longer be clipped behind the Cash payment button.
- [ ] Verify the Cash payment / Other payment methods buttons stay pinned at the bottom on both phone and tablet.
- [ ] When the Totals content is scrollable, verify a thin shadow appears above the payment buttons (while more content is hidden below) and below the back-button row (while content has been scrolled up out of view). Both shadows should fade out when there is no longer content hidden on that edge.
- [ ] Verify the Reader Ready For Payment, Preparing Reader, and Tap to Pay Promoted variants of the Totals screen still render correctly.

### Images/gif

* Small phone - shadows at top and bottom of content, visible when scrolled
Screenshot_20260521_125238
Screenshot_20260521_125232

* Small tablet
Screenshot_20260521_130104
Screenshot_20260521_125934


toupper (commented):
## Review

No blockers. Two minor notes.

### Suggestions

- `WooPosTotalsScreen.kt:370` — The file-level constants `SCROLL_EDGE_SHADOW_HEIGHT` and `SCROLL_EDGE_SHADOW_ALPHA` sit mid-file between `ScrollEdgeShadow` and the unrelated `WooPosPaymentMethod.toUIEvent()` extension. Move them either to the very bottom of the file or just above `ScrollEdgeShadow` so the magic numbers live next to their only call site.
- PR description nit: it says "dropped the now-redundant inner `verticalScroll` from `ReaderDisconnected`" but `ReaderDisconnected` on trunk has no `verticalScroll`. The diff is otherwise accurate — just the description text is misleading.

### Positives

- The `BoxWithConstraints` + `Column.verticalScroll(...).heightIn(min = maxHeight)` pattern is the canonical Compose way to express "center vertically when it fits, scroll when it doesn't" — preserves the tablet/tall-phone look while making small phones non-clipping, exactly as the PR description claims.
- Correct use of `derivedStateOf { scrollState.canScrollForward }` / `canScrollBackward` to gate the edge shadows — avoids a recomposition on every scroll-offset change.
- Pinning `CheckoutPaymentButtons` outside the scrollable region (and keeping the back row in the top pinned slot) maintains the "Cash / Other always one tap away" guarantee called out in the PR description.
- Extracting `ReaderStatusContent` from the inline `when` block is a clean readability win — the outer scrollable column with `Arrangement.SpaceEvenly` and the inner status column with `Arrangement.Center` correctly mirror the original visual behaviour.
- The hairline `1.dp` shadow is consistent with the existing `HorizontalDivider(thickness = 1.dp)` used by `TotalsGrid` in the same file, and is deliberately NOT routed through `WooPosSpacing` (which is adaptive) — that's the right call for a precise hairline that must not scale with breakpoint.
- Adding `textAlign = TextAlign.Center` + horizontal padding to the `ReaderDisconnected` subtitle and title brings it in line with the treatment already used in `ReaderReadyForPayment` and `PreparingReader` — the long "To process this payment, please connect your reader." string now wraps consistently across all status variants.
- Architecturally clean: still pure Compose, `@HiltViewModel` untouched, no Fragments / nav-graph / Android-framework imports introduced.
- Existing `@WooPosPreview` annotations already cover phone + small-tablet + big-tablet variants across all status states — visual coverage for this fix is in place without needing new previews.

samiuelson:
Thanks @toupper! Addressed the code nit:

- **`SCROLL_EDGE_SHADOW_HEIGHT` / `SCROLL_EDGE_SHADOW_ALPHA` placement** — moved to sit directly above `ScrollEdgeShadow` (their only call site) in [f31ddc79d1c](https://github.com/woocommerce/woocommerce-android/commit/f31ddc79d1ceb9e1f0ff852f4472d36fc5d31b81).

For the PR description nit about `ReaderDisconnected` / `verticalScroll` — fair point, leaving the description as-is for now since the diff itself is accurate.

Re-requesting review.

toupper (approved):
Good fix for the totals screen clipping on small phones.

**What I reviewed:** `WooPosTotalsScreen.kt` layout rework, scroll-edge shadows, and `ReaderDisconnected` text alignment.

**What works well:**
- `BoxWithConstraints` + `heightIn(min = maxHeight)` + `verticalScroll` is the correct pattern for "fill-and-center when there's room, scroll when cramped" — the existing tall-phone and tablet behaviour is preserved while small phones no longer clip.
- `derivedStateOf { scrollState.canScrollForward/Backward }` for shadow visibility is efficient and avoids unnecessary recompositions.
- Removing the now-redundant inner `verticalScroll` from `ReaderDisconnected` is the right cleanup since the parent column scrolls.
- `ReaderStatusContent` extraction considerably reduces the complexity of `TotalsLoaded`.
- `textAlign = TextAlign.Center` on the `ReaderDisconnected` title and subtitle is a nice improvement.
- Payment buttons correctly stay pinned at the bottom (outside the scrollable column).

**Minor observation:** The scroll-edge shadow is 1dp tall at 12% black opacity — very subtle, but the screenshots suggest it reads clearly on device. Just noting it in case it needs a tweak after real-device testing.

No blocking concerns. ✅

PR #15954, opened 2026-05-19
[WOOMOB-3102] Migrate AI Assistant chat to Woo Mobile AI

hichamboushaba:
Fixes WOOMOB-3102

## Summary
- Switch the AI Assistant `ChatService` binding to `WooMobileAiChatService` now that the WPCOM `woo-mobile-ai` endpoint is deployed and ready to use.
- Route chat completions through `POST /wpcom/v2/woo-mobile-ai/chat/completions` with direct WPCOM `AccessToken` bearer auth.
- Keep the legacy Jetpack AI query implementation in the module for now, but move both OpenAI-style SSE clients behind a shared configurable `OpenAiSseChatService` transport.
- Make `OpenAiRequestBody` the canonical OpenAI chat-completions body used by `woo-mobile-ai`, with a legacy `JetpackAiQueryRequestBody` adapter preserved for the old endpoint until cleanup.

## Architecture
The production binding now points to `WooMobileAiChatService`, but the shared client architecture remains the important long-term shape. Endpoint-specific classes describe what differs between backends instead of reimplementing the whole streaming client.

The shared architecture separates these concerns:
- **Transport:** `OpenAiSseChatService` owns OkHttp calls, SSE framing, parser integration, cancellation, diagnostics, and generic transport error handling.
- **Auth:** each endpoint provides its own bearer-token adapter. The Woo Mobile AI path uses the WPCOM `AccessToken` directly; the legacy Jetpack path keeps JWT invalidation/retry behavior while it remains in the codebase.
- **Request shape:** `OpenAiRequestBody` is the canonical OpenAI chat-completions shape. OpenAI-compatible endpoints can send it unchanged; legacy endpoints can adapt it into their own top-level DTO, such as `JetpackAiQueryRequestBody` with `feature`.
- **Error mapping:** generic OpenAI/SSE failures stay in the shared layer, while endpoint-specific REST envelopes and retry semantics live in endpoint mappers.

This remains useful after the Jetpack AI query service is removed. At that point, the legacy adapter can go away, but the shared transport and explicit seams still make the Woo Mobile AI client easier to reason about and test. The same shape also gives us a cleaner path for future OpenAI-compatible services that differ in auth, request adaptation, or error envelopes. For example, a future bring-your-own-key flow could provide a different auth provider without forking the SSE transport, parser, diagnostics, and retry/error plumbing.

## Notes
The legacy Jetpack AI query implementation remains only as temporary cleanup debt. A follow-up PR can remove or deprecate that path and update live smoke coverage now that the WPCOM endpoint is deployed and the production binding has switched.

## Testing
Regression test the currently enabled AI Assistant experience in the app:

1. Launch the app with a store that has AI Assistant access.
2. Open AI Assistant.
3. Send a normal assistant prompt and confirm the streamed response completes successfully.
4. Send a prompt that uses store data, such as asking for recent orders or products, and confirm the assistant returns store-specific results without errors.
5. Confirm there are no visible regressions in the AI Assistant conversation UI or error handling.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

staskus (approved):
✅ Works well!

Let's keep gpt 5.1

Additional changes suggestion:
- Update OrdersGetTool and ProductsGetTool like on iOS https://github.com/woocommerce/woocommerce-ios/pull/17188/changes#diff-f43af896840b6e8c6f4b2ad87aa83e6f0195e7163df1a5b933b9bfd3a72150bf, this is a description change I found useful when switching to gpt 5.1 to get more predictable outcomes when diving deeper into orders. 

One code comment:

- `stream_options.include_usage` can be removed, it was a leftover from a previous approach where I called the proxy directly. Now it's not used.

hichamboushaba:
Thanks @staskus 

> stream_options.include_usage can be removed, it was a leftover from a previous approach where I called the proxy directly. Now it's not used.

This is handled.

> Let's keep gpt 5.1

Actually, I'll keep 4o in this PR, the update to 5.1 was not intentional, I'll update in a follow-up PR after updating the smoke tests.

> Update OrdersGetTool and ProductsGetTool like on iOS https://github.com/woocommerce/woocommerce-ios/pull/17188/changes#diff-f43af896840b6e8c6f4b2ad87aa83e6f0195e7163df1a5b933b9bfd3a72150bf, this is a description change I found useful when switching to gpt 5.1 to get more predictable outcomes when diving deeper into orders.

I'll do this also in the followup PR.

PR #15958, opened 2026-05-20
[WOOMOB-3076] Add analytics troubleshooting to AI support chat

itsmeichigo:
### Description
Part of WOOMOB-3076

Adds the analytics troubleshooting path to AI Support Chat on Android. Merchants can now choose the analytics loading issue, see the selected issue inserted into the transcript, and run diagnostics in the expected order: internet connection, WordPress.com servers, site connection, then analytics setting.

The analytics check now reads the actual WooCommerce Analytics setting (`woocommerce_analytics_enabled`) before falling back to the existing dashboard stats path. This prevents disabled analytics from being reported as a passing diagnostic when stats calls still return successfully. When analytics is disabled, the diagnostic shows iOS-aligned wording and offers an `Enable Analytics` fix action, which updates the WooCommerce setting and reruns the analytics diagnostic.

This also updates the generic failure presentation to show specific failure wording instead of a generic retry prompt.

### Test Steps
1. Open AI Support Chat and select the analytics loading issue.
2. Confirm the selected issue appears as a user message before diagnostics begin.
3. With WooCommerce Analytics enabled, confirm diagnostics complete and allow continuing to chat.
4. Disable WooCommerce Analytics for the store, select the analytics loading issue again, and confirm the analytics setting check fails with an `Enable Analytics` action.
5. Tap `Enable Analytics` and confirm the setting is enabled and the analytics diagnostic reruns.
6. Select `Other` and confirm diagnostics are skipped and the chat input appears immediately.

### Images/gif
[Screen_recording_20260520_110328.webm](https://github.com/user-attachments/assets/4e1034f4-54b8-4c54-be5f-320fb3818914)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/useCases/StoreAnalyticsCheckUseCase.kt
```diff
@@ -0,0 +1,128 @@
...
+            is LoadStatsResult.VisitorsStatsSuccess,
+            LoadStatsResult.VisitorStatUnavailable,
+            LoadStatsResult.VisitorStatsLoading -> false
+        }
+
+    private fun LoadStatsResult.toConnectivityCheckStatus(durationMs: Long): ConnectivityCheckStatus =
+        when (this) {
+            is LoadStatsResult.RevenueStatsSuccess -> Success(durationMs = durationMs)
+            LoadStatsResult.PluginNotActive -> analyticsInactiveFailure(durationMs)
+            is LoadStatsResult.RevenueStatsError -> Failure(
```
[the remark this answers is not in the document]
itsmeichigo: This can be an edge case - keeping the logic as-is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/useCases/StoreAnalyticsCheckUseCase.kt
```diff
@@ -0,0 +1,128 @@
...
+        val range = StatsTimeRangeSelection.build(
+            selectionType = StatsTimeRangeSelection.SelectionType.TODAY,
+            referenceDate = Date(),
+            calendar = Calendar.getInstance(),
+            locale = Locale.getDefault()
+        )
+        val (result, duration) = measureTimedValue {
+            getStats(refresh = true, selectedRange = range)
+                .filter { it.isRevenueCheckResult }
+                .first()
```
malinajirka, line 62: 🚫`getStats(refresh = true)` emits cached value first, and since we use `first()` operator, the remote request will be cancelled => the user will always see cached data.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
```diff
@@ -250,6 +270,22 @@ class AiSupportChatViewModel @Inject constructor(
...
+    private fun enableAnalytics() {
+        launch {
+            _viewState.update { it.copy(isExecutingFixAction = true) }
+
+            diagnosticsService.enableAnalytics()
+                .onSuccess {
+                    _viewState.update { it.copy(isExecutingFixAction = false) }
+                    rerunDiagnostics()
+                }
+                .onFailure { error ->
```
malinajirka, line 291: 💡 Should we display an error to the user?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -347,6 +347,35 @@ open class WooCommerceStore @Inject internal constructor(
         }
     }
 
+    suspend fun enableAnalytics(site: SiteModel): Boolean {
```
malinajirka: ❓ `enableAnalytics` collapses every REST error into a bare `false`, so callers can't tell a network/auth failure apart from a genuine "setting didn't stick". The sibling `fetchAnalyticsEnabled` you added right below returns `WooResult<Boolean>` and keeps `response.error` — could `enableAnalytics` do the same for consistency? It would also let `StoreAnalyticsCheckUseCase` propagate the real error type instead of a generic `error("Failed to enable analytics setting")`, which ties into the silent-failure point on the VM. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/diagnostics/SupportDiagnosticsService.kt
```diff
@@ -82,6 +85,24 @@ class SupportDiagnosticsService @Inject constructor(
         STORE_CONNECTION -> storeConnectionCheck()
         STORE_ORDERS -> storeOrdersCheck()
         STORE_PRODUCTS -> storeProductsCheck()
+        ANALYTICS_SETTING -> storeAnalyticsCheck()
+    }
+
+    suspend fun enableAnalytics(): Result<Unit> =
```
malinajirka, line 92: ⚠️ runCatching catches even cancellation exception, we should re-throw it to avoid breaking structured concurrency.

malinajirka (commented):
Thanks for working on this! I haven't tested this, since I assume @iamgabrielma will be a better tester, but I left some in-line comments. In case you'd want me to test it, just let me know.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/useCases/StoreAnalyticsCheckUseCase.kt
itsmeichigo: Updated in e37b76e49026b1789198d45f13ba5a249a288524.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
itsmeichigo: Added alert in e050ef13c6cc57a52f1f24720643b85a1ed7cbc1.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
itsmeichigo: Updated in 353837bdca28cea8ab3c3ef1af63b8a4b8c5596f.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/diagnostics/SupportDiagnosticsService.kt
itsmeichigo: Updated in ebbe52c0d3d624da118944e68aff383fe481289e.

itsmeichigo:
Thanks @malinajirka for the feedback! I addressed them in subsequent commits, pls let me know if anything should be updated. I'll wait for Gabriel to test the feature 🙇

iamgabrielma (approved):
Works well! Code-wise I have nothing additional to add to the existing review 👍 

Just noting that claude thinks the issue copilot reported might not be fully resolved yet:

> WooCommerce/src/main/kotlin/com/woocommerce/android/ui/troubleshooting/useCases/
    StoreAnalyticsCheckUseCase.kt:93
    I’d reopen the resolved Copilot thread here. After fetchAnalyticsEnabled() has already confirmed the
    setting is enabled, a later GetStats.PluginNotActive is still converted into analyticsInactiveFailure().
    SupportDiagnosticsService then sees PLUGIN_NOT_ACTIVE and offers Enable Analytics, while the UI shows
    the “Analytics is not enabled” copy. In that path the proposed fix is misleading and likely loops:
    tapping Enable Analytics writes the same enabled value, reruns diagnostics, and can land on the same
    failure again. I’d use a distinct error type/details for stats PluginNotActive, and reserve
    PLUGIN_NOT_ACTIVE/EnableAnalytics for the explicit fetchAnalyticsEnabled(site) == false branch.

itsmeichigo:
@iamgabrielma I responded to Copilot about that comment - that is likely an edge case and I want to keep the behavior. User is not blocked - if they cannot update the settings they can proceed with support chat to ask about the issue.

PR #15974, opened 2026-05-21
Seed manifest + sidecar baseline for the 16 existing locales

JorgeMucientes:
## Summary

Seeds the sidecar baseline that lets the AI translation engine preserve the 16 existing human-translated Android locales without re-translating them, and restores the source-string context comments that the engine uses for future translation runs.

This PR adds:

- `fastlane/ai_translation/baseline/values-XX/strings.xml` for `ar`, `de`, `es`, `fr`, `he`, `id`, `it`, `ja`, `ko`, `nl`, `pt-rBR`, `ru`, `sv`, `tr`, `zh-rCN`, and `zh-rTW`.
- `fastlane/ai_translation/translation-manifest.json` records for those same 16 locales with `origin=glotpress-import`.
- Schema/model metadata aligned with the current engine contract.
- `<!-- Context: ... -->` comments above translatable source keys in `WooCommerce/src/main/res/values/strings.xml`, restored from the earlier closed backfill stack so future translations get code-derived context.
- The metadata char-cap re-prompt success-path unit test that was originally split out as #15982.
- Import/backfill behavior that refreshes the sidecar baseline from the canonical localized `strings.xml` files.
- GitHub linguist markers for the generated baseline files and manifest.

The baseline files are byte-for-byte copies of the committed localized Android resources. The manifest is intentionally scoped to the 16 existing locales only; the 15 new app-string locale backfill is handled by #16004.

## Why

The engine treats the committed localized XML as the source of truth for existing translations. The sidecar baseline lets the preserved-line writer emit unchanged `glotpress-import` keys byte-for-byte while only source changes or real gaps flow through AI translation.

Without this baseline, a missing or incomplete manifest could make the engine queue existing human translations for AI translation. Without the restored source comments, the 15-locale backfill and future PR-time translations lose important per-key context.

## Stack

Base: #15962.

Open stack above this PR:

- #15983 - Play Store metadata backfill for the 15 new locales.
- #16004 - Android app-string backfill for the 15 new locales.
- #16011 - draft PR-time AI translation pipeline fixture.

Merged/folded stack PR:

- #15982 - metadata char-cap re-prompt success-path test, now included in this PR's history.

Full stack map: #15961.

## Test Plan

- [x] Manifest JSON parses and contains records only for the 16 existing locales.
- [x] Sidecar files are byte-identical to the current `WooCommerce/src/main/res/values-XX/strings.xml` files for the 16 existing locales.
- [x] `ruby fastlane/ai_translation/spec/woo_ai_translation_test.rb` - `99 runs / 368 assertions / 0 failures`.
- [x] `git diff --check`
- [x] Confirmed the re-seeded manifest uses the current `source_sha` schema.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

toupper (commented):
Looked at this as a data-only baseline PR: the 16 sidecar XMLs, the manifest, the source-side context comments, and the folded-in metadata test. Functionally it does what the description claims, no translation content is changed, and the engine consumes the artefacts correctly. A few schema/operational nits worth a follow-up, but no blockers.

## Positives

- **All 16 sidecars are byte-identical to `WooCommerce/src/main/res/values-XX/strings.xml`.** Verified locally with `diff -q` across `ar de es fr he id it ja ko nl pt-rBR ru sv tr zh-rCN zh-rTW`. No drift, no surprise re-serialization. Legacy aliases (`values-in` → `values-id`, `values-iw` → `values-he`) are also byte-identical to their modern counterparts, so collapsing to the 16 canonical codes in the baseline is correct.
- **Source `strings.xml` diff is strictly additive.** Only `<!-- Context: ... -->` comments added; zero deletions; no `<string>` value change. Confirmed by `git diff origin/issue/woomob-translations-p2-engine...HEAD -- WooCommerce/src/main/res/values/strings.xml | grep '^-'` returning only the diff header. ~3681 source keys get a `Context:` line (~87.5% of translatable keys), and section-header comments like `<!-- General Strings -->` still flow through as sticky context for the rest.
- **Manifest source_sha values are correct.** Spot-checked `app_name`, `signout`, `my_store`, `orders`, `email`, `shipping` — every recorded key-level `source_sha` matches `SHA256(unit.source_signature)` computed against the current source. Hash algo and signature shape are documented in `manifest.rb` and `android_resources.rb`.
- **Schema preamble matches the engine.** `schema_version: 2`, `prompt_version: "2026-05-19.1"`, `models: {default: claude-sonnet-4-6, escalated: claude-opus-4-7}` all line up with `Manifest::SCHEMA_VERSION` and `WooAiTranslation.model_for(...)`.
- **Manifest correctly captures the partial-coverage edge cases.** 4122 source keys → 3870 with all 16 locales, 3 with 15 locales (the missing locale really is absent from the corresponding `values-XX/strings.xml`), 249 with 0 locales (post-GlotPress source additions like `dashboard_ai_assistant_entry_point_label`, `notification_channel_stock_title`). Exactly what we want: the AI engine will queue those for translation; existing human ones are reused byte-for-byte.
- **End-to-end engine consumption works against the committed manifest.** A quick `Manifest#stale?` smoke against the live source returns `false` for current keys and `true` for a fabricated source change. Origin lookup correctly returns `glotpress-import`.
- **Folded-in test for metadata char-cap re-prompt (#15982) is well-scoped** — locks the SUCCESS path explicitly (ship the tightened retry, not the English fallback, not flagged), asserts the exact call count.

## Suggestions

- **Schema drift between the seed and current `Manifest#record`.** The committed locale entries carry `cache_key` rather than `source_sha`. The seed commit (`b60e9176bf8`, 2026-05-21 15:01) predates `8a5d22f76fc` "Source-only invalidation" (2026-05-21 16:10), which dropped `cache_key` from app-string locales and added `current_source_sha` at the key level. `stale?` still works because it falls back to `key_entry['source_sha']`, but the manifest now mixes two formats in one file (loaded with v1 layout; next AI write will use v2 layout) — confusing for future readers and for `--keys` diff tooling. Worth re-running `mode:import` once against current `manifest.rb` before merge so the entire file lands on the current schema, or at least adding a one-line note in `manifest.rb` that the legacy `cache_key` field may appear on imported entries until they're re-translated.
- **Sidecars duplicate `values-XX/strings.xml` content (~6.3 MB).** Documented as intentional ("file copy, not re-serialization"), and that's fine, but it introduces a silent drift risk: any future edit to `values-XX/strings.xml` that doesn't also update `baseline/values-XX/strings.xml` will cause the preserved-line writer to emit stale XML. Two cheap mitigations: (a) a CI check (a one-liner `diff -rq baseline/values-XX WooCommerce/src/main/res/values-XX` for the 16 baseline locales) and/or (b) extend `Importer#import` to also `FileUtils.cp` the localized file into `baseline/` so `mode:import` is self-sufficient and operators can't forget the second step. Right now there is no script that re-seeds the sidecar XMLs.
- **dangermattic spam (177 warnings).** All of them are spurious — `<item>@string/date_timeframe_today</item>` etc. already exist in `WooCommerce/src/main/res/values-XX/strings.xml` and were just file-copied. The danger gate `android_strings_checker.check_strings_do_not_refer_resource` only inspects diffs, so the new sidecar paths trip every entry. Worth scoping that check (in `Dangerfile`) to `WooCommerce/src/main/res/` only, otherwise every regenerate of the baseline will repeat this noise on its PR. Not a blocker for *this* PR but a near-certainty for the next one.
- **No `.gitattributes` markers for the manifest / baseline.** The 20 MB JSON + 16 large XMLs would benefit from `linguist-generated=true` (collapses on GitHub) and `merge=union` on the manifest — without those, two concurrent translation runs will conflict every release and the bulky JSON diffs will clutter every PR view. Optional.

## Out of scope (flagged for awareness, not for this PR)

- The 2 local test errors I see (`invalid byte sequence in US-ASCII`) reproduce against the base branch too, so they're a pre-existing local-encoding issue, not introduced here. Re-running with `LANG=en_US.UTF-8` clears them. The PR's own test plan shows 91 runs / 348 assertions passing on a clean env, which matches what I'd expect.

JorgeMucientes:
Thanks @toupper. I pushed follow-ups for the baseline and manifest concerns.

Changed:
- Re-seeded the manifest with the current schema, so imported locale records now match the `source_sha` shape used by the engine.
- `import` and `backfill` now refresh the sidecar baseline from the canonical localized `strings.xml` files.
- Marked the baseline files and manifest as generated for GitHub linguist, so the PR diff should be easier to review.

Not changed:
- I did not add `merge=union` for the manifest because union merges can create invalid JSON or duplicate object fields. The explicit `merge-manifests` tool is safer.
- I did not change the Danger resource-reference check in this pass because I could not verify the plugin API signature locally, and I did not want to push an untested Danger change into the stack.
- I also did not add a separate baseline drift CI check yet. The import path now refreshes the baseline automatically, and a dedicated CI gate can follow if we want stricter enforcement.

PR #15977, opened 2026-05-21
[WOOMOB-3076] Fix notification diagnostics analytics mapping

irfano:
### Description
Fixes WOOMOB-3076

Adds analytics values for the notification diagnostics introduced in #15960. Without these mappings, the existing exhaustive DiagnosticTest mapper fails compilation after the new notification diagnostic checks are added.

### Test Steps
Confirm the project builds successfully.

### Images/gif
N/A


itsmeichigo (approved):
Thanks for the fix!

PR #15988, opened 2026-05-22
POS: surface failed-note message after Mark Order as Complete success

samiuelson:
## Summary
Addresses [malinajirka's review thread on #15944](https://github.com/woocommerce/woocommerce-android/pull/15944#discussion_r3271326689): when `MarkOrderAsCompleteOutcome.SuccessWithFailedNote` came back, the merchant only saw a clean success and the note-post failure was tracked only via analytics. This PR surfaces it to the merchant on the destination Payment Success screen, below the order-total message.

Why a parallel PR rather than #15932: the follow-up UI PR is currently rebased on a pre-rename state (still under `markorderaspaid/` package + older `Confirming(isProcessing, canConfirm)` state shape) and needs separate reconciliation. Landing the snackbar plumbing here on trunk keeps the scope clean and lets #15932 plug into existing wiring once it's rebased.

## Changes
- New child→parent event variant `OrderSuccessfullyPaidExternallyWithFailedNote` (sibling of the existing `OrderSuccessfullyPaidExternally`).
- `ParentToChildrenEvent.OrderSuccessfullyPaid` carries a new `hasFailedNote: Boolean = false` field.
- `WooPosHomeViewModel` routes the new variant through `onOrderSuccessfullyPaid(EXTERNAL, hasFailedNote = true)`.
- `WooPosTotalsViewState.PaymentSuccess` gains an optional `secondaryMessage: String?`.
- `WooPosTotalsViewModel.showSuccessfulPaymentState` populates `secondaryMessage` from a new string when `hasFailedNote = true`.
- `WooPosTotalsPaymentSuccessScreen` renders `secondaryMessage` as a `BodyMedium` line below the order-total message (uses `onSurfaceVariant` color). New `@WooPosPreview` for the failed-note variant.
- `WooPosMarkOrderAsCompleteViewModel` emits the new variant on `SuccessWithFailedNote`; existing `Success` still emits the plain variant.
- New string `woopos_totals_success_payment_note_post_failed`.

## Test plan
- [ ] Open POS, add an item, tap Other payment methods → Mark order as complete. With network available, confirm without a note: success screen renders normally (no secondary message).
- [ ] Repeat with a customer note. Force the note-post step to fail (e.g. drop network just after status is updated, or stub repo to return `SuccessWithFailedNote`). Verify the Payment Success screen now shows "The order was marked as complete, but the customer note couldn't be saved." under the total message.
- [ ] Confirm cha-ching still plays and the "New order" / "Email receipt" buttons are unchanged.

## Unit tests
- `WooPosHomeViewModelTest` covers both event variants → correct `hasFailedNote` flag in the forwarded `OrderSuccessfullyPaid`.
- `WooPosTotalsViewModelTest` covers `EXTERNAL` with `hasFailedNote = true` (secondaryMessage set) and `false` (secondaryMessage null).
- `WooPosMarkOrderAsCompleteViewModelTest` covers `SuccessWithFailedNote` → emits the new event variant + tracks `MarkAsPaidNotePostFailed`.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
@@ -39,6 +39,7 @@ sealed class ChildToParentEvent {
     data object GoBackToCheckoutAfterFailedPayment : ChildToParentEvent()
     data object OrderSuccessfullyPaidByCard : ChildToParentEvent()
     data object OrderSuccessfullyPaidExternally : ChildToParentEvent()
+    data object OrderSuccessfullyPaidExternallyWithFailedNote : ChildToParentEvent()
     data object ExitPosClicked : ChildToParentEvent()
     data object SetupBarcodeScannerClicked : ChildToParentEvent()
     data object CouponsValidationFailed : ChildToParentEvent()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeParentToChildCommunication.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeParentToChildCommunication.kt
@@ -49,7 +49,10 @@ sealed class ParentToChildrenEvent {
         val itemClickedDataList: List<WooPosItemsViewModel.ItemClickedData>
     ) : ParentToChildrenEvent()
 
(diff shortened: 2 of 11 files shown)


samiuelson:
Closing per iOS parity check. iOS deliberately treats note-post failure as a non-event for the merchant — silent log only (`DDLogWarn` in `Modules/Sources/PointOfSale/Controllers/PointOfSaleOrderController.swift`), no analytic, no UI surface. Comment in iOS source: "the order completion is the critical path, and a stale note can always be added later via order detail." Android will match by not surfacing the failed-note state to the user. Will reply to malinajirka's question on #15932 with this rationale.

PR #15992, opened 2026-05-25
Enable Canada POS and extract refund submission processor

joshheald:
Part of: RSM-645
Part of: RSM-644

### Description
- Enable Canada/CAD POS eligibility and cover Canada in the POS tab and launchability checks.
- Keep Interac-present payment support flowing through the Canada card-reader config.
- Extract POS refund submission behind a processor while preserving the existing backend refund path.

### Stack
Base: trunk
Next branch: issue/android-pos-interac-refunds-02-reader-routing

### Test Steps
- Unit tests pass.
- On a Canada/CAD WooPayments store, verify POS can be launched and card payments can still start from POS.
- For end-to-end payment/refund validation, test from the tip of the stack: #15996.

### Images/gif
N/A. This PR is foundation/refactor work with no standalone UI changes.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
@@ -1,7 +1,6 @@
 package com.woocommerce.android.ui.woopos.orders.details.refund
 
 import com.woocommerce.android.R
-import com.woocommerce.android.model.Order
 import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.woopos.orders.WooPosLoadPaymentGateway
 import com.woocommerce.android.util.WooLog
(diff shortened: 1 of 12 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/tab/WooPosIsCountryAllowed.kt
```diff
@@ -30,7 +30,7 @@ class WooPosIsCountryAllowed @Inject constructor(
 
     private companion object {
         val SUPPORTED_COUNTRIES = setOf(
-            "US", "PR", "GB",
+            "US", "PR", "GB", "CA",
             "FR", "DE", "IE", "NL", "AT", "BE", "FI", "IT", "LU", "PT", "ES",
             "SG", "NZ",
```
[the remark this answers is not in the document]
joshheald: Doc comment updated.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -0,0 +1,105 @@
...
+    private val refundStore: WCRefundStore,
+    private val selectedSite: SelectedSite,
+    private val loadPaymentGateway: WooPosLoadPaymentGateway,
+    private val resourceProvider: ResourceProvider,
+) : WooPosRefundSubmissionProcessor {
+    @Suppress("TooGenericExceptionCaught")
+    override fun submit(request: WooPosRefundSubmissionRequest): Flow<WooPosRefundSubmissionState> = flow {
+        try {
+            WooLog.i(
+                WooLog.T.POS,
```
samiuelson, line 24: 💡 This interface is single-implementation and single-consumer across the whole stack (only `WooPosRefundSubmissionProcessorImpl`, only used by `WooPosRefundViewModel`), so it isn't buying polymorphism. Mockito can mock the concrete class directly and Hilt can bind it, so the interface is redundant — the extracted class alone gives the same testability. Fine to keep if it matches a team convention, otherwise worth dropping.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -0,0 +1,105 @@
...
+                "WooPosRefund: submission started " +
+                    "orderId=${request.orderId}, " +
+                    "amount=${request.refundAmount}, " +
+                    "itemCount=${request.refundItems.size}"
```
samiuelson, line 28: 💡 Consider moving `WooPosRefundSubmissionRequest` and `WooPosRefundSubmissionState` into their own files. Keeping the request model and the state hierarchy separate from the processor keeps this file focused and makes the types easier to locate and reuse.

samiuelson (approved):
LGTM! Approving with two minor, non-blocking suggestions below.

PR #15993, opened 2026-05-25
Route POS Interac refunds through card reader

joshheald:
Part of: RSM-645

### Description
- Add a POS refund controller factory path that builds `PaymentOrRefund.Refund`.
- Detect `interac_present` charges and route those refunds through the card reader before backend notification.
- Keep `card_present` and manual refunds on the existing backend-only refund path.

### Stack
Previous PR: #15992
Base branch: issue/android-pos-interac-refunds-01-foundation
Next branch: issue/android-pos-interac-refunds-03-reader-ui

### Test Steps
- Unit tests pass.
- From the tip of the stack (#15996), refund an Interac-present POS payment and verify the card-reader refund runs before backend refund creation.
- Refund a non-Interac card-present POS payment and a manual/backend refund path, and verify they still submit directly through the existing backend refund flow.

### Images/gif
N/A. This PR routes refund submission logic but does not add the POS-native reader UI yet.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCardReaderPaymentControllerFactory.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCardReaderPaymentControllerFactory.kt
@@ -87,9 +87,7 @@ class WooPosCardReaderPaymentControllerFactory @Inject constructor(
     fun createRefund(
         orderId: Long,
         refundAmount: BigDecimal,
-        cardReaderType: CardReaderType = CardReaderType.EXTERNAL,
         isTTPPaymentInProgress: KMutableProperty0<Boolean>,
-        allowCancelledStatus: Boolean = false,
     ): CardReaderPaymentController = CardReaderPaymentController(
         cardReaderManager = cardReaderManager,
         orderRepository = orderRepository,
@@ -114,8 +112,8 @@ class WooPosCardReaderPaymentControllerFactory @Inject constructor(
             orderId = orderId,
             refundAmount = refundAmount
         ),
-        cardReaderType = cardReaderType,
+        cardReaderType = CardReaderType.EXTERNAL,
         isTTPPaymentInProgress = isTTPPaymentInProgress,
-        allowCancelledStatus = allowCancelledStatus,
+        allowCancelledStatus = false,
     )
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCardReaderPaymentControllerFactory.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCardReaderPaymentControllerFactory.kt
@@ -24,6 +24,7 @@ import com.woocommerce.android.ui.payments.tracking.PaymentsFlowTracker
 import com.woocommerce.android.util.CoroutineDispatchers
 import com.woocommerce.android.util.CurrencyFormatter
 import org.wordpress.android.fluxc.store.WooCommerceStore
+import java.math.BigDecimal
 import javax.inject.Inject
 import kotlin.reflect.KMutableProperty0
 
@@ -77,6 +78,41 @@ class WooPosCardReaderPaymentControllerFactory @Inject constructor(
         paymentOrRefund = PaymentOrRefund.Payment(
             orderId = orderId,
             paymentType = paymentType
+        ),
+        cardReaderType = cardReaderType,
+        isTTPPaymentInProgress = isTTPPaymentInProgress,
+        allowCancelledStatus = allowCancelledStatus,
+    )
+
+    fun createRefund(
+        orderId: Long,
+        refundAmount: BigDecimal,
+        cardReaderType: CardReaderType = CardReaderType.EXTERNAL,
+        isTTPPaymentInProgress: KMutableProperty0<Boolean>,
+        allowCancelledStatus: Boolean = false,
+    ): CardReaderPaymentController = CardReaderPaymentController(
+        cardReaderManager = cardReaderManager,
+        orderRepository = orderRepository,
+        selectedSite = selectedSite,
+        appPrefs = appPrefs,
+        paymentCollectibilityChecker = paymentCollectibilityChecker,
+        interacRefundableChecker = interacRefundableChecker,
+        tracker = tracker,
+        trackCancelledFlow = trackCancelledFlow,
+        currencyFormatter = currencyFormatter,
+        errorMapper = errorMapper,
+        interacRefundErrorMapper = interacRefundErrorMapper,
+        wooStore = wooStore,
+        dispatchers = dispatchers,
+        cardReaderTrackingInfoKeeper = cardReaderTrackingInfoKeeper,
+        paymentStateProvider = paymentStateProvider,
+        cardReaderPaymentOrderHelper = cardReaderPaymentOrderHelper,
+        paymentReceiptHelper = paymentReceiptHelper,
+        cardReaderOnboardingChecker = cardReaderOnboardingChecker,
+        paymentReceiptShare = paymentReceiptShare,
+        paymentOrRefund = PaymentOrRefund.Refund(
+            orderId = orderId,
+            refundAmount = refundAmount
         ),
         cardReaderType = cardReaderType,
         isTTPPaymentInProgress = isTTPPaymentInProgress,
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -23,83 +36,436 @@ data class WooPosRefundSubmissionRequest(
...
+    data class WaitingForCard(@StringRes val cardReaderHint: Int? = null) : WooPosRefundSubmissionState()
+    data object ProcessingReaderRefund : WooPosRefundSubmissionState()
+    data object NotifyingStore : WooPosRefundSubmissionState()
     data object Success : WooPosRefundSubmissionState()
-    data class Failure(val message: String) : WooPosRefundSubmissionState()
+    data class Failure(
+        val message: String,
+        val retryBackendNotificationOnly: Boolean = false,
+        val retryCardRefund: Boolean = false,
+        val canRetry: Boolean = !retryBackendNotificationOnly,
```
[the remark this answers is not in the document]
joshheald: Fixed in the updated PR2 branch: reader-originated failures now set canRetry from state.onRetry, and the processor test covers both retryable and non-retryable Interac reader failures.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -23,83 +36,436 @@ data class WooPosRefundSubmissionRequest(
...
+                        WooLog.e(
+                            WooLog.T.POS,
+                            "WooPosRefund: failed to fetch payment charge metadata orderId=${request.orderId}"
+                        )
+                        trySendState(
+                            WooPosRefundSubmissionState.Failure(
+                                message = resourceProvider.getString(R.string.error_generic),
+                            )
+                        )
+                        return@channelFlow
```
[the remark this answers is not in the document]
joshheald: I think failing closed is intentional here. For orders with a chargeId, if charge metadata lookup fails we cannot tell whether the original payment was interac_present; falling back to the backend refund path could bypass the terminal refund requirement. Orders without a chargeId still use the backend path, and confirmed card_present/manual payments continue through the backend path once metadata is available. The failure case is covered so we do not accidentally create a backend refund when the lookup is inconclusive.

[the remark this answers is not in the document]
joshheald: Failing closed is safer there because if lookup fails for an order with a chargeId, we cannot prove it was not interac_present; falling back to backend refunding could bypass the terminal requirement.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundViewModel.kt
```diff
@@ -292,6 +306,10 @@ class WooPosRefundViewModel @AssistedInject constructor(
         }
     }
 
+    private fun handleConnectReaderClicked() {
+        resumePendingRefundIfReaderConnected()
+    }
```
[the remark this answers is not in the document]
joshheald: The button handler in WooPosIssueRefundScreen first sets showCardReaderConnectionDialog = true, which opens WooPosCardReaderConnectionDialog. So I think we can leave this as-is.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundViewModel.kt
```diff
@@ -416,50 +434,129 @@ class WooPosRefundViewModel @AssistedInject constructor(
         refundJob?.cancel()
         refundJob = viewModelScope.launch {
             refundSubmissionProcessor.submit(request).collect { submissionState ->
```
samiuelson, line 436: ⚠️ Potential crash on the normal backend refund path. `WooPosRefundSubmissionProcessor.notifyBackend` rethrows unexpected exceptions when `retryBackendNotificationOnly = false` (see processor lines 184–185, and the test "…then exception is propagated"). This `collect` isn't wrapped in try/catch, and POS VMs are plain `ViewModel()` with no `CoroutineExceptionHandler`, so a thrown flow becomes an uncaught exception in `viewModelScope` → app crash mid-refund. The foundation PR caught all non-cancellation exceptions and emitted `Failure(error_generic)`. Was fail-fast intended? At point of sale, surfacing a retryable `Failure` seems safer than crashing.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -23,83 +40,570 @@ data class WooPosRefundSubmissionRequest(
...
+            return RefundSubmissionPath.Backend(paymentMethodType = null)
+        }
 
-            if (result.isError) {
-                emit(
-                    WooPosRefundSubmissionState.Failure(
-                        message = result.error.message ?: resourceProvider.getString(R.string.error_generic),
-                    )
+        return when (val result = paymentChargeRepository.fetchCardDataUsedForOrderPayment(chargeId)) {
+            PaymentChargeRepository.CardDataUsedForOrderPaymentResult.Error -> {
```
samiuelson, line 124: ⚠️ Interac orders can silently fall back to a backend refund that can't move the money. If `order.chargeId == null` (line 126) or charge-metadata fetch returns `Error` (here), the path becomes `Backend(...)`. For an order that was actually `interac_present`, the backend refund then runs with `autoRefund = supportsRefunds` (true for WooPayments), but Interac can't be auto-refunded by the gateway — so we'd record a refund without actually refunding the card, with only a `WooLog.w`. The happy path is unaffected, but should a charge-metadata failure on a card order surface an error so the merchant retries, rather than degrade to a backend-only refund?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosCardReaderPaymentControllerFactory.kt
```diff
@@ -82,4 +83,39 @@ class WooPosCardReaderPaymentControllerFactory @Inject constructor(
         isTTPPaymentInProgress = isTTPPaymentInProgress,
         allowCancelledStatus = allowCancelledStatus,
     )
+
+    fun createRefund(
+        orderId: Long,
+        refundAmount: BigDecimal,
+        cardReaderType: CardReaderType = CardReaderType.EXTERNAL,
+        isTTPPaymentInProgress: KMutableProperty0<Boolean>,
+        allowCancelledStatus: Boolean = false,
```
samiuelson: 💡 `allowCancelledStatus` (and `cardReaderType`) don't need defaults here. `createRefund` has a single caller across the whole stack (`WooPosRefundSubmissionProcessor.createInteracRefundController`), which never overrides `allowCancelledStatus` (always `false`) and passes `cardReaderType = EXTERNAL` explicitly — which is also the default, so it's dead. The defaults look copied from the general-purpose `create()`, but this method is purpose-built for the one Interac path. Either drop the param and hardcode `false` in the body, or keep it without a default. Minor: `cardReaderType` (defaulted) is declared before `isTTPPaymentInProgress` (non-defaulted) — conventionally defaulted params trail.

samiuelson (approved):
LGTM. Added a few improvement suggestions, take a look.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundViewModel.kt
joshheald: Resolved by using WooPosRefundSubmissionState.Failure and displaying errors

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
joshheald: The refund would fail if we did this, not silently. We only need a chargeId for interac refunds, so the revert to Backend is deliberate if we don't have it. This is the same as we've done on iOS, and it is a bit of a trade off but it's a worthwhile one to keep other refunds resillient.

PR #15994, opened 2026-05-25
Add POS-native reader refund UI

joshheald:
Part of: RSM-645

### Description
- Add POS-native reader refund screens and copy instead of reusing payment text.
- Show connect-reader and reader-processing states inside the POS refund flow.
- Keep card-reader refunds on the reader processing experience through backend notification.
- Let the full-screen refund modal fill the window so reader processing has no top/bottom bars.

### Stack
Previous PR: #15993
Base branch: issue/android-pos-interac-refunds-02-reader-routing
Next branch: issue/android-pos-interac-refunds-04-hardening

### Test Steps
- Unit tests pass.
- From the tip of the stack (#15996), start an Interac-present POS refund with no reader connected and verify the POS connect-reader view appears.
- Connect a reader and verify the refund uses POS-native refund copy and the card-reader processing screen through backend notification.
- Refund a non-Interac payment and verify it keeps the API refund processing UI rather than the card-reader UI.

### Images/gif
N/A. Reader states are best validated in-device with a reader/simulator setup.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -200,7 +493,20 @@ fun WooPosIssueRefundScreen(
...
 @Composable
-private fun resolveToolbarTitle(state: WooPosRefundState): String {
+private fun RefundPaymentInProgressContent(
+    state: WooPosRefundState.Content,
+) {
+    WooPosPaymentInProgressScreen(
+        state = WooPosTotalsViewState.PaymentInProgress(
+            title = stringResource(R.string.woopos_refund_reader_processing_title),
+            subtitle = state.formattedTotal,
+        ),
```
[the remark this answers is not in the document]
joshheald: No, we want to keep showing the processing UI in this case to avoid UI jumpiness or over the top detail. This matches iOS.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -116,38 +185,189 @@ fun WooPosIssueRefundScreen(
...
+                            viewModel.onUIEvent(WooPosRefundUIEvent.ConnectReaderClicked)
+                        }
+                    )
+                }
+            }
+        }
+    }
+
+    if (showCardReaderConnectionDialog && !presentModalAsDialog) {
+        RefundCardReaderConnectionDialog(
```
samiuelson: 💡 `RefundCardReaderConnectionDialog` is invoked twice — once inside the `Dialog` block (line 318) and once here, each guarded by `presentModalAsDialog`. The trigger (`showCardReaderConnectionDialog`) and `onConnectionSuccess` effect are identical in both branches, so it could be hoisted to a single conditional render. Same for the two `onConnectReaderClicked = { showCardReaderConnectionDialog = true; viewModel.onUIEvent(...) }` lambdas (lines 270 and 309) — they could be a single `val` near the top of the composable.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -116,38 +185,189 @@ fun WooPosIssueRefundScreen(
...
+        if (dismissRequestToken > 0) {
+            handleDismiss()
+        }
     }
 
     val state by viewModel.state.collectAsStateWithLifecycle()
 
-    val toolbarTitle = resolveToolbarTitle(state)
+    LaunchedEffect(state) {
+        onPendingChangesChanged(state.hasPendingChanges())
```
samiuelson: 💡 `LaunchedEffect(state) { onPendingChangesChanged(state.hasPendingChanges()) }` re-invokes the callback on every state emission, even when the boolean is unchanged — parents will receive a stream of redundant `false`s during item-selection flicks. Consider `derivedStateOf { state.hasPendingChanges() }` + `LaunchedEffect(pendingChanges)`, or compare against a `remember`'d previous value before invoking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -116,38 +185,189 @@ fun WooPosIssueRefundScreen(
...
+            handleDismiss()
+        }
+    }
+
+    Box(
+        modifier = modifier
+            .fillMaxSize()
+            .background(MaterialTheme.colorScheme.surface)
+    ) {
+        RefundSelectionLayer(
```
samiuelson, line 251: 💡 When `presentModalAsDialog = false` and `modalState != null`, the modal is painted on top in the `Box`, but `RefundSelectionLayer` (including its `RefundScreenButtons`) stays in the composition tree underneath. Visually fine because of the surface background, but accessibility services may still announce the hidden buttons. Consider conditionally rendering the selection layer's button row, or applying `Modifier.semantics { invisibleToUser() }` / `clearAndSetSemantics {}` when a modal is active.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -116,38 +185,189 @@ fun WooPosIssueRefundScreen(
...
+                contentInsetsModifier = Modifier
+                    .statusBarsPadding()
+                    .navigationBarsPadding(),
+                disablePartialRefund = disablePartialRefund,
+            )
+        }
+    }
+
+    if (modalState != null && presentModalAsDialog) {
+        Dialog(
```
samiuelson, line 287: 💡 `Dialog(usePlatformDefaultWidth = false, decorFitsSystemWindows = false, dismissOnClickOutside = false, dismissOnBackPress = !modalIsProcessing)` is a non-standard full-bleed dialog. Should work, but it's a fragile combination across Android versions / 3-button vs gesture nav. Worth an explicit manual-test row in the #15996 test plan ("verify full-bleed modal on phone + tablet, gesture and button nav, Android 13/14/15") since end-to-end testing happens at the tip of the stack.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -1048,6 +1418,110 @@ private fun ConfirmRefundContent(
...
+        WooPosRefundState.Content.RefundStep.ProcessingRefund ->
+            stringResource(R.string.woopos_refund_reader_processing_title)
+        WooPosRefundState.Content.RefundStep.NotifyingStore ->
+            stringResource(R.string.woopos_refund_notifying_store_title)
+        WooPosRefundState.Content.RefundStep.SelectItems,
+        WooPosRefundState.Content.RefundStep.ReviewRefund,
+        WooPosRefundState.Content.RefundStep.ConfirmRefund,
+        WooPosRefundState.Content.RefundStep.ReaderDisconnected,
+        WooPosRefundState.Content.RefundStep.Processing ->
+            error("RefundReaderContent rendered for unsupported step")
```
samiuelson: 💡 `error("RefundReaderContent rendered for unsupported step")` (here and again at line 1458) is defensive but would crash mid-refund if a future state-transition refactor ever routed an unexpected step here. The dispatcher in `ContentStateHandler` already guarantees correctness, so this is meant as "unreachable" rather than a real check. A no-op (`Unit`) branch — or surfacing the canned subtitle — would be safer with the same readability.

samiuelson (approved):
LGTM! Added a few non-blocking improvements ideas.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
joshheald: We do full-bleed dialogs commonly in POS though, right? Or have I done it a strange way here?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
joshheald: Thanks, good spot... crept in because of the two modes for display

PR #15995, opened 2026-05-25
Harden POS refund cancellation flows

joshheald:
Part of: RSM-645

### Description
- Cancel in-flight checkout payment work before opening orders/refunds.
- Align refund back and cancel behavior across selection, review, confirmation, and reader states.
- Avoid crashing or retrying terminal work when cancellation races with Stripe Terminal processing.

### Stack
Previous PR: #15994
Base branch: issue/android-pos-interac-refunds-03-reader-ui
Next branch: issue/android-pos-interac-refunds-05-detail-pane

### Test Steps
- Unit tests pass.
- From the tip of the stack (https://github.com/woocommerce/woocommerce-android/pull/15996), start POS checkout with the reader disconnected, open Orders, start a card-present refund, connect the reader, and verify the old checkout payment flow does not continue or race the refund.
- During an Interac-present refund, verify Back returns to the previous refund step where allowed, Cancel exits the refund flow, and cancellation is blocked once terminal processing is non-cancelable.
- While the card is being tapped/processed, attempt to cancel/back out and verify the app does not crash.

### Images/gif
N/A. This PR hardens flow behavior and is best validated in-device.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
@@ -787,11 +787,16 @@ class CardReaderPaymentController(
         }
     }
 
+    @Suppress("TooGenericExceptionCaught")
     fun stop() {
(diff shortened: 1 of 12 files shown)


samiuelson (approved):
LGTM!

PR #15996, opened 2026-05-25
Show POS refund selection in order detail pane

joshheald:
Closes: RSM-645
Closes: RSM-644

### Description
- Move refund item selection into the order detail pane.
- Open the remaining refund review, confirmation, and processing flow in the full-screen modal.
- Preserve pending refund selection state and guard stale dismiss events when switching orders or cancelling.

### Stack
Previous PR: #15995
Base branch: issue/android-pos-interac-refunds-04-hardening
Final PR in this stack

### Test Steps
Note that to test Interac refunds, you need to use a WisePad and an Interac card, even in the simulator. If you connect to a simulated M2 or Chipper, it will not work. This does not apply in real use, as the real M2/Chipper will not connect on a Canadian store.

- [ ] Using a CA store, take several card payments using both an Interac card and a non-Interac card in POS.
- [ ] Refund the Interac payments:
  - [ ] Partial refund
  - [ ] Full refund
  - [ ] Refund with a card reader already connected
  - [ ] Refund without a card reader connected at the start
  - [ ] Observe that the refunds are processed showing the card payment-style UI
- [ ] Refund the non-Interac payments:
  - [ ] Partial refund
  - [ ] Full refund
  - [ ] Observe that regardless of the card reader connection, refunds are not processed using the card payment UI
- [ ] Start editing a refund, select another order, and verify pending changes are confirmed before switching orders.
- [ ] Continue a refund from the detail pane into the full-screen modal, including reader connection, and verify the modal covers the selection pane, respects top/bottom insets, and stays non-dismissible while processing.

I have also tested the error handling. These PRs do not offer retries for Interac refunds once the card-present refund section is complete. Just because we fail to submit it to the website, does not mean we should do the refund again. It is not perfect: if we do not submit to the site, the products do not show as refunded, but the refund itself does show because of the webhooks we have. This is the same behavior as Interac refunds in the main app. We could improve it by adding metadata in the refund request which identified the refunded line items, so that the webhook could update that bit of the order too, but that is separate.

### Images/gif

https://github.com/user-attachments/assets/effb1629-b63e-47e9-9297-fccfa07cfdee


malinajirka (commented):
Great work! I've ran the suggested testing steps + a few other and it works as expected. I have NOT reviewed the code.

P.S. In case some changes come out of the reviews, let me know and I'll be happy to retest.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
```diff
@@ -199,9 +201,11 @@ fun WooPosIssueRefundScreen(
         }
     }
 
+    var lastDismissRequestToken by rememberSaveable(viewModelKey) { mutableIntStateOf(0) }
     LaunchedEffect(dismissRequestToken) {
```
samiuelson, line 197: ⚠️ `lastDismissRequestToken` is advanced unconditionally, even when `handleDismiss()` was refused by `viewModel.onDismissRequest()` (which returns `false` for `Processing` / `ProcessingRefund` / `NotifyingStore`). The host's `pendingOrderSelectionAfterRefundDismiss` was already set *before* bumping the token (`WooPosOrdersScreen.kt:182`), and is only consumed inside `handleIssueRefundDismissed` — which fires from `onDismissed`, which never gets called on a refused dismiss. So the token is "spent" but the pending order id stays latched.

Failure mode: merchant taps Issue Refund on order A → refund reaches a non-cancelable step → merchant taps order B in the list to glance at it → host bumps the token and stores `pendingOrderSelectionAfterRefundDismiss = B` → refund screen silently consumes the token without dismissing → merchant waits for the refund to finish and taps Close on the success screen → `handleIssueRefundDismissed` fires → app navigates to **order B**, which the merchant never asked for in that moment.

Two composable fixes:
1. Have `handleDismiss` return a `Boolean` and only advance `lastDismissRequestToken` when it actually dismissed.
2. Either give the merchant feedback when a dismiss is refused (a brief snackbar / log), or clear `pendingOrderSelectionAfterRefundDismiss` on the host when the dismiss didn't propagate (e.g., reset it whenever the state changes to a non-cancelable step after a token bump).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -120,12 +129,83 @@ fun WooPosOrdersScreen(
...
+        detailPaneIssueRefundOrderId = null
+        detailPaneIssueRefundHasPendingChanges = false
+        detailViewModel.onBackFromIssueRefund(refundedOrderId)
+        pendingOrderSelectionAfterRefundDismiss?.let { orderId ->
+            listViewModel.onOrderSelected(orderId, WooPosScreenType.DualPane)
+        }
+        pendingOrderSelectionAfterRefundDismiss = null
+    }
+
+    val requestRefundDismissBeforeOrderSelection: (Long) -> Unit = { orderId ->
```
samiuelson, line 170: 💡 The ~80 lines of host-side orchestration here — token bump, pending-selection latch, the three-way `handleOrderSelected` (same order / `hasPendingChanges` → confirmation / direct dismiss), and `handleIssueRefundDismissed` — are the merchant-facing safety net against losing partial refund edits, but they ship without test backing. The new `WooPosOrderDetailsViewModelTest` test (nice!) covers the VM-side regression (refresh the *correct* order after a cross-order switch), but none of the screen-level state machine. A couple of small additions would lock this down:

- Extract `handleOrderSelected` + `handleIssueRefundDismissed` into a tiny holder class so the three branches can be unit-tested without UI.
- Add a VM test asserting that order-2 state is **untouched** after `onBackFromIssueRefund(1L)` — that's the actual regression the cross-order guards at `WooPosOrderDetailsViewModel.kt:298` (`sideLoadJob` cancel only when `current.details.id == selectedOrderId`) and `:313` (`applyOrderUpdate` skip when ids differ) prevent. The current test covers "the refunded order is refreshed" but not "the other order is left alone."

samiuelson (approved):
LGTM. Added a few improvements ideas, take a look!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosIssueRefundScreen.kt
joshheald: Done, we're clearing the selection in this case.

PR #16005, opened 2026-05-26
Fix open-in-new icon position when low stock threshold is null

irfano:
### Description
When the store-wide low stock threshold is null, the open-in-new (↗) icon in the "View store-wide threshold" helper renders at the end of the sentence instead of next to the link. The original `LowStockDetails` composable always appended the icon to the end of the text, which lined up with the link in the "set" copy (link at end) but broke in the "null" copy (link mid-sentence). This change anchors the icon to the link annotation's end so it stays adjacent to the link in both copies.

Reported by Adam Grzybkowski in a DM.

### Test Steps
1. Open Settings, Notifications, Stock notifications.
2. To produce the "null" state: tap "View store-wide threshold", clear the "Low stock threshold" field in the store admin, save, and return to the app.
3. Verify the ↗ icon sits between "View store-wide threshold" and "to see the current value."
4. To produce the "set" state: open the link again, set the threshold to a value, save, and return to the app.
5. Verify the ↗ icon sits right after "Edit store-wide threshold" at the end of the sentence.

### Images/gif
|before|after|
|-|-|
|||

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/notifications/NewStockNotificationSettingsScreen.kt
@@ -212,25 +212,27 @@ private fun LowStockDetails(
     }
     val openInNewIconId = "openInNewIcon"
     val textWithIcon = remember(text) {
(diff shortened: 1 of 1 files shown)


AdamGrzybkowski (approved):
LGTM 👍

PR #16020, opened 2026-05-28
[WOOMOB-2983] Make "Shipment details" title tappable

irfano:
### Description

Fixes WOOMOB-2983

Only the chevron toggled the shipping label bottom sheet. Adding the same click handler to the "Shipment details" title so tapping it also expands or collapses the sheet.

### Test Steps

1. Log in to a store where Woo Shipping is enabled and there is an order eligible for shipping labels.
2. Open the Orders tab and select that order.
3. Tap "Create shipping label" to start the shipping label flow.
4. Tap directly on the "Shipment details" text. Verify the bottom sheet expands.
5. Tap the chevron. Verify the sheet collapses.


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/ShipmentDetails.kt
```diff
@@ -176,7 +177,11 @@ fun ShipmentDetails(
                     Text(
                         text = stringResource(R.string.shipping_label_shipment_details_title),
                         color = MaterialTheme.colors.primary,
-                        modifier = Modifier
+                        modifier = Modifier.clickable(
+                            onClick = toggleSheet,
+                            interactionSource = remember { MutableInteractionSource() },
+                            indication = null
```
AdamGrzybkowski: ❓ Is there a reason we are removing the ripple?

AdamGrzybkowski (commented):
Would it make sense to expand the width so that the user can tap on the whole area? 

Image

irfano:
> Would it make sense to expand the width so that the user can tap on the whole area?

Yes, that makes sense. I made the change: 29f54e5dadbbad4758be52d4dacf1aa336e88ec2

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/wooshippinglabels/ShipmentDetails.kt
irfano: I kept the current behavior we have for the chevron. Also, I think it looks better to expand the bottom sheet without a ripple effect.
I checked the bottom sheet in the Maps app for reference. Even though the app has a ripple effect on buttons, there isn’t an effect on the bottom sheet itself.

AdamGrzybkowski (approved):
Thanks for the changes!

PR #16021, opened 2026-05-28
[WOOMOB-2520] Prevent verifying address with invalid input

irfano:
Fixes WOOMOB-2520

### Description

On the shipping label address screen, each input field is validated with a 500 ms debounce (`transformLatestWithDelay`) so the inline error text does not flash while the user is still typing. The problem was that this same debounced result also drove the Verify button's enabled state, so the button lagged ~500 ms behind the input: right after typing an invalid or empty email, the button stayed enabled and tappable until the debounce elapsed.

### Test Steps

1. Open an order that supports shipping labels.
2. Start the Create Shipping Label flow.
3. From the bottom sheet, open the edit destination address screen.
4. Clear the email field and verify the Verify button is disabled immediately.
5. Enter a valid email (e.g. `test@example.com`) and verify the Verify button is enabled.
6. Enter an invalid email (e.g. `abc`) and verify the Verify button is disabled immediately.

### Images/gif

N/A


AdamGrzybkowski (commented):
Hey @irfano - this works, but I have this feeling that the fix is a bit of a plaster. I'm worried it could easily be broken again by someone who's not familiar with the code, and there's a lot of reactive flows there to be aware of. 

I had no idea why the bug was occurring. Initially, I thought that maybe because of the network call. Only later on, I learned about the delay we have for each input field. I think it would be great to explain this in a bit more detail in the PR description.

The fix relies on the custom operator `transformLatestWithDelay` that folks might not be familiar with (me included) and if ever this was changed to a standard debounce, the current fix would stop working.

My suggestion is to split the concern and keep the current flows with delays for inline errors, and create a new flow that will be used for the button state. The new flow won't be using a delayed flows (because we don't want it for the button) and hence it will be harder to break by a mistake in the future. Something like this:

```kotlin
private val isFormValidFlow = combine(
        snapshotFlow { name },
        snapshotFlow { company },
        snapshotFlow { address },
        snapshotFlow { city },
        snapshotFlow { postalCode },
        snapshotFlow { email },
        snapshotFlow { phone },
        country
    ) { name, company, address, city, postalCode, email, phone, country ->
        addressValidator.validateAtLeastOneOf(name.value, company.value) == null &&
            addressValidator.validateFieldRequired(address.value) == null &&
            addressValidator.validateFieldRequired(city.value) == null &&
            addressValidator.validateFieldRequired(postalCode.value) == null &&
            (!email.isRequired || addressValidator.validateEmail(email.value) == null) &&
            (!phone.isRequired || validatePhoneByCountry(phone.value, country) == null)
    }

private suspend fun observeChanges() {
        combine(
            editableAddress,
            isCompanyExpanded,
            countriesState,
            statesState,
            addressValidationState,
            currentAddress,
            isFormValidFlow
        ) { address, isExpanded, countriesState, statesState, addressValidation, currentAddress, isFormValid ->
```

Additionally, it would be great to add unit tests for such bugs. 

WDYT?

irfano:
This is really a clean approach, thanks @AdamGrzybkowski! I've applied all your suggestions. You're right that the PR description could have been clearer, I've updated it too.

AdamGrzybkowski (approved):
Looks good, thanks!

PR #16030, opened 2026-06-01
[WOOMOB-3159] Add data layer for analytics scheduled-import setting

AdamGrzybkowski:
Fixes WOOMOB-3159

### Description

This is the **data layer only** for the Analytics "scheduled import" (12h-delay) setting — no UI. It's a building block for the parent issue [WOOMOB-3153](https://linear.app/a8c/issue/WOOMOB-3153/android-add-setting-for-analytics-updates), which will surface a "Stats may be up to 12 hours delayed" notice and a settings bottom sheet on the dashboard cards.

WooCommerce exposes this setting at `wc/v3/settings/wc_admin/woocommerce_analytics_scheduled_import` as a radio with two string values: `yes` ("Scheduled (recommended)") and `no` ("Immediately"). We map it to a `Boolean`, mirroring how `woocommerce_enable_coupons` is handled, where `value == "yes"` means the scheduled (delayed) import is enabled.

The change adds the standard three layers, each mirroring the existing analytics order-date-type plumbing:

- **`WooCommerceRestClient`** — adds the `ANALYTICS_SCHEDULED_IMPORT_SETTING_ID` constant (reusing the existing `wc_admin` group) plus `fetchAnalyticsScheduledImportEnabled` / `updateAnalyticsScheduledImportEnabled`, returning `WooPayload<Boolean>` and reusing the existing `SiteSettingOptionResponse` DTO.
- **`WooCommerceStore`** — thin `WooResult<Boolean>` wrappers via `coroutineEngine.withDefaultContext`.
- **`AnalyticsScheduledImportRepository`** (new, in `ui/dashboard/data`) — shaped like `CashOnDeliverySettingsRepository`, exposing `isEnabled()` / `setEnabled(Boolean)`. It returns `WooResult<Boolean>` (rather than collapsing errors to a bare `false`) so the upcoming UI can distinguish a network error from a real value.

The read/update round-trip was verified against a live store response (`value: "yes"` → `isEnabled() == true`).

### Test Steps

This is a non-user-facing data-layer change with no UI, so there's nothing to tap through in the app. Behaviour is covered by the unit tests added in this PR (`WooCommerceStoreTest`, `AnalyticsScheduledImportRepositoryTest`).

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/AnalyticsScheduledImportRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/AnalyticsScheduledImportRepository.kt
@@ -1,0 +1,17 @@
+package com.woocommerce.android.ui.dashboard.data
+
+import com.woocommerce.android.tools.SelectedSite
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
+import org.wordpress.android.fluxc.store.WooCommerceStore
+import javax.inject.Inject
+
+class AnalyticsScheduledImportRepository @Inject constructor(
+    private val wooCommerceStore: WooCommerceStore,
+    private val selectedSite: SelectedSite,
+) {
+    suspend fun isEnabled(): WooResult<Boolean> =
+        wooCommerceStore.fetchAnalyticsScheduledImportEnabled(selectedSite.get())
+
+    suspend fun setEnabled(enabled: Boolean): WooResult<Boolean> =
+        wooCommerceStore.updateAnalyticsScheduledImportEnabled(selectedSite.get(), enabled)
+}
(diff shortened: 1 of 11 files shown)


hichamboushaba (approved):
Code looks good @AdamGrzybkowski, thanks.

Just one question: do we need to cache the value? I mean this will be shown on the dashboard card, so without cache, we'll have to wait for the fetch and show a loading UI 🤔, WDYT?

AdamGrzybkowski:
Good call @hichamboushaba - I will add local cache and request another review.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba (approved):
Thanks @AdamGrzybkowski

PR #16035, opened 2026-06-02
[WOOMOB-3160] Show scheduled-import delay notice and settings sheet on dashboard cards

AdamGrzybkowski:
### Description
Fixes WOOMOB-3160

WooCommerce Analytics can import order/sales data on a schedule (the "Updates" option in Analytics → Settings) rather than in real time. When scheduled, the figures on the `/reports`-backed dashboard cards can lag by up to 12 hours. This PR surfaces that to merchants and lets them change the setting from inside the app.

**What changed**

- The **Performance** and **Top performers** cards now show a small line — *"Stats may be up to 12 hours delayed"* with a tappable ⓘ — when the scheduled-import setting is enabled. When it's disabled they keep the existing *"Last update: %s"* line, now also with the ⓘ so the setting is always reachable. The line sits in each card's existing last-update spot (Performance: between the metrics and the chart; Top performers: at the bottom of the list), matching trunk's placement.
- Tapping ⓘ opens a new **Analytics updates** bottom sheet — a single-choice picker (**Scheduled** vs **Immediately**) that writes the setting via the repository, shows a spinner in place of the checkmark while saving, and reverts with an inline error on failure. The setting is store-wide.
- The cards react to the change via the Room-backed `observeIsEnabled()` flow from the data layer, so flipping the option in the sheet updates the cards without any manual result plumbing.

### Test Steps
1. On a store, open the **My Store** tab.
2. On the **Performance** and **Top performers** cards, find the small centered line under the data ("Stats may be up to 12 hours delayed" or "Last update: …") with an ⓘ icon, and tap the ⓘ.
3. The **Analytics updates** sheet opens with two options (Scheduled / Immediately); the current setting is checked.
4. Tap the other option → a spinner replaces the checkmark while it saves, the sheet closes on success, and the card line updates (delay notice ↔ "Last update: …").
5. Reopen the sheet and tap **Learn more** → the WooCommerce analytics docs open in a Custom Tab.
6. Reopen the sheet, disable the network, and tap the other option → the spinner stops, an inline error appears, and the selection reverts.

### Images/gif

https://github.com/user-attachments/assets/7c4ab30c-4ff8-4e7b-ad0e-a88c231ecf4e

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/ScheduledImportInfoBottomSheetFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/ScheduledImportInfoBottomSheetFragment.kt
@@ -1,0 +1,293 @@
+package com.woocommerce.android.ui.dashboard
+
+import android.content.res.Configuration
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.material3.Icon
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.livedata.observeAsState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.ComposeView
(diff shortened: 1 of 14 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/ScheduledImportInfoBottomSheetFragment.kt
```diff
@@ -0,0 +1,293 @@
...
+        shape = RoundedCornerShape(
+            topStart = dimensionResource(id = R.dimen.minor_100),
+            topEnd = dimensionResource(id = R.dimen.minor_100)
+        )
+    ) {
+        // The horizontal padding is applied per child (not on the Column) so the tappable option
+        // rows can draw their ripple across the full width of the sheet.
+        val contentPadding = Modifier.padding(horizontal = dimensionResource(id = R.dimen.major_100))
+        Column(
+            modifier = Modifier.fillMaxWidth(),
```
hichamboushaba: np, I would make the content scrollable here, in case the screen doesn't fit, for example when using split screen:

Image

hichamboushaba (approved):
Nice work @AdamGrzybkowski, left one np comment, but other than that :shipit:

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/ScheduledImportInfoBottomSheetFragment.kt
AdamGrzybkowski: Good call! Thanks

PR #16041, opened 2026-06-03
[WOOMOB-3161] Show scheduled-import delay notice on dashboard pull-to-refresh

AdamGrzybkowski:
### Description

Fixes WOOMOB-3161

Some stores enable a "scheduled import" setting that batches analytics updates roughly every 12 hours, so the dashboard stats can be stale. PR 2 added a delayed-stats footer on the cards plus an info bottom sheet that explains the delay and lets the merchant change the setting. This PR adds the proactive notice:

- When the merchant **pulls to refresh** the My Store dashboard, if scheduled import is enabled **and** they have never opened the info bottom sheet, we show an action snackbar — "Stats may be up to 12 hours delayed" with a **Learn more** action that opens the PR 2 info sheet.
- Opening the info sheet (from the card (i) icon **or** the snackbar action) sets an `AppPrefs` flag `hasSeenAnalyticsScheduledImportInfo` (a `DeletablePrefKey`, so it is cleared on logout), so the notice is only shown until the merchant has seen the explanation once.

To make testing easier, I've added a Developer Options entry ("Reset analytics scheduled import notice") that resets the seen flag so the notice can be re-triggered.

### Test Steps

Use a store with the analytics scheduled-import setting **enabled**.

1. In **Settings → Developer options**, tap **Reset analytics scheduled import notice** (this clears the "seen" flag).
2. Go to the My Store dashboard and **pull to refresh**.
3. Verify a snackbar appears: "Stats may be up to 12 hours delayed" with a **Learn more** action.
4. Tap **Learn more** → the scheduled-import info bottom sheet opens.
5. Dismiss the sheet and **pull to refresh** again → verify the snackbar no longer appears.
6. Reset the flag again (step 1), then open the sheet via a card's (i) icon instead, dismiss it, and pull to refresh → verify the snackbar does not appear (opening the sheet from either entry point marks it as seen).
7. On a store with scheduled import **disabled**, pull to refresh → verify no snackbar appears.

### Images/gif

https://github.com/user-attachments/assets/1ef8f2d5-79dc-47f5-99ba-434370a210ce

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
@@ -153,6 +153,7 @@ object AppPrefs {
         AI_PRODUCT_DESCRIPTION_CELEBRATION_SHOWN,
         AI_ASSISTANT_EARLY_ACCESS_NOTICE_DISMISSED,
         AUTO_TAX_RATE_ID,
+        HAS_SEEN_ANALYTICS_SCHEDULED_IMPORT_INFO,
     }
 
     /**
@@ -361,6 +362,13 @@ object AppPrefs {
                 .commit()
             check(committed) { "Failed to persist AI Assistant early access notice dismissal" }
         }
+
+    var hasSeenAnalyticsScheduledImportInfo: Boolean
+        get() = getBoolean(
+            key = DeletableSitePrefKey.HAS_SEEN_ANALYTICS_SCHEDULED_IMPORT_INFO,
+            default = false,
+        )
+        set(value) = setBoolean(key = DeletableSitePrefKey.HAS_SEEN_ANALYTICS_SCHEDULED_IMPORT_INFO, value = value)
 
     private const val FEATURE_FLAG_OVERRIDE_PREFIX = "feature_flag_override"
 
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
```diff
@@ -153,6 +153,7 @@ object AppPrefs {
         AI_PRODUCT_DESCRIPTION_CELEBRATION_SHOWN,
         AI_ASSISTANT_EARLY_ACCESS_NOTICE_DISMISSED,
         AUTO_TAX_RATE_ID,
+        HAS_SEEN_ANALYTICS_SCHEDULED_IMPORT_INFO,
```
irfano: Despite the `DeletableSitePrefKey` name, this isn't per-site: it's stored by enum name with no site id and only cleared on logout, so it's effectively per-account (seen on one store hides it on all). Since the description calls for per-site, let's store it under a site-scoped key instead, the way the store onboarding prefs do, e.g. `getBoolean(PrefKeyString("HAS_SEEN_ANALYTICS_SCHEDULED_IMPORT_INFO:$siteId"), default = false)` with `siteId` from selectedSite, so each store tracks it independently.

AdamGrzybkowski: Thanks - that's not how it should be, but the description was wrong. This doesn't have to be per-site. It should match the [iOS implementation](https://github.com/woocommerce/woocommerce-ios/pull/17304) - a per-account flag that's cleared on logout.

irfano (approved):
Thanks for the change. LGTM! 👍🏻

AdamGrzybkowski:
Hey @irfano I've pushed an extra commit to only show the notice when any of the affected cards are visible on the dashboard.

irfano (approved):
The change makes sense! LGTM after fixing the failing tests. 👍🏻

PR #16044, opened 2026-06-04
[WOOMOB-3141] Hide TTP entry points and surface a clear message on unsupported devices

samiuelson:
Fixes [WOOMOB-3141](https://linear.app/a8c/issue/WOOMOB-3141).

### Description

On Android devices without a Trusted Execution Environment / hardware-backed key attestation (e.g. Samsung Galaxy A72), the Tap to Pay setup entry point was reachable from multiple places — payment method selection, Payments Hub, POS — only to fail at Stripe Terminal discovery with `TerminalException: Device does not use Trusted Execution Environment, or does not support hardware-backed key attestation.`, surfaced to the user as the generic *"No reader connected"* message. Merchants and support had no signal that the device itself was the problem.

This PR moves device-level capability into `TapToPayAvailabilityStatus` by adding a Stripe-backed check (`Terminal.supportsReadersOfType(TAP_TO_PAY_DEVICE, …)`) on top of the existing OS / NFC / Google Play Services / country / CIAB pre-flight gates. When Stripe says the device is incapable, a new `Result.NotAvailable.DeviceNotSupported` is returned and:

- **Hides** TTP from Collect Payment → method selection, POS checkout, and the More-menu "NEW" badge (all already gate on `.isAvailable`).
- **Keeps** the *"About Tap to Pay"* row visible in Payments Hub. Tapping it shows a *"Tap to Pay is not available"* dialog with a *Check Requirements* button that opens https://woocommerce.com/document/woopayments/in-person-payments/tap-to-pay-android/ — so support can point merchants there.

As defense-in-depth, the discovery path now propagates Stripe's TTP error codes (`TAP_TO_PAY_UNSUPPORTED_DEVICE`, `_TAMPERED`, `_UNSUPPORTED_ANDROID_VERSION`) through a new `CardReaderDiscoveryEvents.FailedTapToPayDeviceUnsupported`, and `CardReaderConnectViewModel` shows a new `TapToPayDeviceUnsupportedState` instead of the generic "No reader connected" — so even if a gate is bypassed the user still gets a useful message.

The Stripe check requires Terminal to be initialized (token provider tied to a logged-in WPCom site), so the existing custom checks are kept as a fast offline path and the Stripe answer is cached per-process (device capability cannot change at runtime). When Terminal isn't yet initialized the checker returns `null` and we treat the device as available, letting the discovery-error path catch any false positives.

**Out of scope:** replacing the OS / NFC / GPS / country custom checks (kept as offline fast-path); a dedicated POS error UI for the new event (POS hides TTP upstream — connector isn't reached).

**Heads-up for review:** the new string `card_reader_tap_to_pay_not_available_error_device` ("This device doesn't meet the security requirements for Tap to Pay on Android…") is a placeholder — flagged for copy review with PM / the issue reporter.

### Test Steps

**Supported device** (Pixel etc.):
- Collect Payment, POS checkout, and Payments Hub all show TTP rows as before; setting up TTP completes normally.

**Unsupported device** (Samsung Galaxy A72 / Android 14, per issue):
- Collect Payment → no *Tap to Pay* row in the payment method list.
- POS checkout → no Tap-to-Pay button.
- More menu → no Payments "NEW" badge driven by TTP.
- Payments Hub → *Tap to Pay* header + *About Tap to Pay* row visible; *Try out Tap to Pay* is hidden. Tapping *About…* shows the new dialog; tapping *Check Requirements* opens https://woocommerce.com/document/woopayments/in-person-payments/tap-to-pay-android/.
- If discovery is somehow reached (manual deep link, stale state), the error screen now reads *"Tap to Pay is not available"* with a *Check Requirements* button instead of *"No reader connected"*.

**Regression**: confirm the existing `NfcNotAvailable` / `SystemVersionNotSupported` / `GooglePlayServicesNotAvailable` / `CountryNotSupported` dialogs still appear correctly on devices that hit those cases.

**Analytics**: `CARD_PRESENT_TAP_TO_PAY_NOT_AVAILABLE` should fire with `reason=DeviceNotSupported` on the affected device.

### Images/gif

_N/A — no UI surface added beyond the existing "Tap to Pay is not available" dialog pattern; behaviour change is "row hidden" / "different dialog text and button"._


samiuelson:
@toupper Could you please verify this PR on your device (Samsung S21) which passes previous TTP compatibility checks (NFC, OS), but was not supported by Stripe Terminal SDK?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/connection/ConnectionManager.kt
```diff
@@ -84,7 +84,13 @@ internal class ConnectionManager(
                 }
 
                 is DiscoverReadersStatus.Failure -> {
-                    CardReaderDiscoveryEvents.Failed(state.exception.errorMessage)
+                    when (state.exception.errorCode) {
+                        TerminalErrorCode.TAP_TO_PAY_UNSUPPORTED_DEVICE,
+                        TerminalErrorCode.TAP_TO_PAY_DEVICE_TAMPERED,
+                        TerminalErrorCode.TAP_TO_PAY_UNSUPPORTED_ANDROID_VERSION ->
```
malinajirka: 🔍  Should we include `TAP_TO_PAY_UNSUPPORTED_PROCESSOR` and `TAP_TO_PAY_INSECURE_ENVIRONMENT` Stripe's states as well?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubTapToPayUnavailableHandler.kt
```diff
@@ -44,6 +44,14 @@ class PaymentsHubTapToPayUnavailableHandler @Inject constructor() {
                     triggerEvent
                 ) { positiveButtonClick(ActionType.TAP_TO_PAY_REQUIREMENTS) }
             }
+
+            TapToPayAvailabilityStatus.Result.NotAvailable.DeviceNotSupported -> {
+                showDialog(
+                    R.string.card_reader_tap_to_pay_not_available_error_device,
+                    R.string.card_reader_tap_to_pay_not_available_error_check_requirements_button,
+                    triggerEvent
+                ) { positiveButtonClick(ActionType.TAP_TO_PAY_LEARN_MORE) }
```
malinajirka: ❓Is the LearnMore action here intentional? The button seems to use `card_reader_tap_to_pay_not_available_error_check_requirements_button` so I'd expect it to re-run the checks. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/connect/CardReaderConnectViewState.kt
```diff
@@ -103,6 +103,17 @@ sealed class CardReaderConnectViewState(
         secondaryActionLabel = R.string.cancel,
     )
 
+    data class TapToPayDeviceUnsupportedState(
+        override val onPrimaryActionClicked: () -> Unit,
+        override val onSecondaryActionClicked: () -> Unit,
+    ) : CardReaderConnectViewState(
+        headerLabel = UiString.UiStringRes(R.string.card_reader_tap_to_pay_not_available_error_title),
```
malinajirka, line 110: 🔍 We usually don't use UIStrings in POS. It also seems inconsistent with how the other strings are defined below.

malinajirka (approved):
Great job, thanks for working on this @samiuelson! I have left a couple minor comments, but looks good to me overall.

Having said that, I don't have the Samsung device that was causing troubles - @toupper could you please smoke test the behavior and ensure that TTP is not visible on your Samsung phone?

libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/connection/ConnectionManager.kt
samiuelson: Good catch — added both. Fixed in cdac179db7e.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubTapToPayUnavailableHandler.kt
samiuelson: Good point — the LearnMore action did not match the "check requirements" button label. Switched to TAP_TO_PAY_REQUIREMENTS so it opens the Stripe device-requirements page, consistent with the CountryNotSupported branch. Fixed in 837149bf20f.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/connect/CardReaderConnectViewState.kt
samiuelson: This file is part of the main app (not POS). `headerLabel` is typed as `UiString?` in the parent `CardReaderConnectViewState`, so every existing state (e.g. `ExternalReaderScanningState`, `BuiltInReaderScanningState`, `ScanningFailedState`, `ConnectingFailedState`) wraps it with `UiString.UiStringRes(...)`. The fields below (`hintLabel`, `primaryActionLabel`, `secondaryActionLabel`) are `@StringRes Int?` in the parent, which is why they use raw `R.string.xxx`. So the new state follows the existing convention in the file. Happy to revisit if I am missing something though!

toupper:
Thanks for the ping @malinajirka! @samiuelson I tested it with Samsung S21. First time I entered into checkout I saw TTP and got the debug message alert. The second time, I didn't see TTP:


https://github.com/user-attachments/assets/1ee9a02a-d381-422f-b099-5ab9959b22be

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/291">25.0</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

samiuelson:
> Thanks for the ping @malinajirka! @samiuelson I tested it with Samsung S21. First time I entered into checkout I saw TTP and got the debug message alert. The second time, I didn't see TTP:
> 
>  Screen_Recording_20260609_152146_Woo.Dev.mp4

@toupper Thanks for testing and sharing the recording! I spent a while analysing the code to check if we can further improve TTP check UX in POS checkout, and decided to merge the PR because I believe the current behavior it's optimal.

Whether a phone can act as a TTP reader is answered by Stripe's Terminal SDK, and the answer exists only after the SDK is initialized. The SDK can't start when POS launches because it needs location permission and it would add startup work for merchants who never use TTP, so it starts when a payment flow needs it. Your first checkout rendered before the answer existed, so the app had to show or hide the option without knowing.

TTP is shown while the answer is unknown, for two reasons:
 1. On a phone-only setup, tapping the TTP option is the only thing that starts the SDK. If we hid the option until we knew, the SDK would never start, the answer would never arrive, and the option would never appear — including on phones that do support TTP. 
2. The cost lands only on devices like S21: it passes every check we can run ourselves (Android version, NFC, Play Services, country) and fails only inside Stripe's own check. After that the option stays hidden for the rest of the session.

Let me know in case you have idea for improving this flow, @malinajirka, @toupper, I'll be happy to improve the UX further!

PR #16045, opened 2026-06-05
Remove fiscalization countries from IPP and POS support

joshheald:
### Description
Removes IPP/card-reader and POS availability for Austria, Belgium, France, Italy, Germany, Portugal, and Spain.

The card-reader config factory now resolves those country codes as unsupported, the app-level IPP feature-flag gate no longer includes them, and POS supported country/currency pairs no longer expose them. Tests cover the remaining supported expansion countries and assert that the removed fiscalization countries are unsupported.

### Test Steps
1. Open the app on a store based in Austria, Belgium, France, Italy, Germany, Portugal, or Spain.
2. Confirm in-person payments/card reader setup is not offered as a supported flow for that store.
3. On a tablet using the same store, confirm the POS tab/launch eligibility is not shown as supported for that store country.
4. Repeat with a supported expansion country such as Ireland and confirm POS/IPP eligibility still uses the expected EUR configuration.

### Images/gif
N/A


samiuelson (approved):
LGTM!

PR #16056, opened 2026-06-10
[WOOMOB-2721] Replace requireNotNull with soft guard in POS orders replaceOrders

malinajirka:
### Description

Fixes WOOMOB-2721

In `WooPosOrdersListViewModel.replaceOrders()`, when the previously selected order is no longer present in the new list, the code resolved the new selection with `requireNotNull(newFirstOrderId) { "Content requires at least one order" }`. If `replaceOrders` were ever called with an empty list, that assertion throws `IllegalArgumentException` and crashes the app. While I'm a fan of fail fast principle, I believe here it's too risky, since the data of the list might change in between the calls, resulting in some race condition crashes. 

### Test Steps
No test needed

### Images/gif

N/A


samiuelson (approved):
Thank you. LGTM :shipit:

PR #16079, opened 2026-06-12
Clean up AI assistant dependency hygiene

hichamboushaba:
### Description
Why:
The AI assistant boundary was looser than the host actually needs. `:WooCommerce` declared both `:libs:ai-assistant:feature` and `:libs:ai-assistant:core`, but the host only enters the assistant through the feature module. Keeping `:core` visible to the host makes it easier for future host code to couple to assistant internals, and the dependency-analysis report also identified a few genuinely unused or mis-scoped dependencies in the assistant modules.

How:
- Narrowed AI assistant feature implementation details to `internal`, keeping `AssistantRoute`, card renderer contracts, and telemetry wiring as the host-facing API.
- Removed the host app's direct dependency on `:libs:ai-assistant:core`.
- Switched assistant state collection to `collectAsStateWithLifecycle()`, so `lifecycle-runtime-compose` remains an intentional dependency.
- Removed unused `okhttp-sse` and Apache HTTP test dependencies.
- Added or adjusted direct declarations where they document real module usage: lifecycle-viewmodel, Hilt lifecycle ViewModel Compose, serialization-core, and runtime-only Android coroutines.

### Test Steps
1. Open the AI assistant screen from the host app and verify it loads normally.
2. Send a basic prompt and verify assistant responses still render.
3. Trigger any assistant card result and verify card taps still route through the host.

### Images/gif
N/A


AdamGrzybkowski (approved):
LGTM :shipit:

PR #16080, opened 2026-06-12
[WOOMOB-2501] Android fallback to site credentials after site-info failure

hichamboushaba:
### Description
Fixes WOOMOB-2501

When Woo login enters a site address, Android currently relies on WordPress.com `/connect/site-info` to classify the site before continuing. That blocks login when `/site-info` cannot verify a site even though the device can still reach the site directly through the WordPress REST API.

This PR adds an explicit, opt-in fallback path for the Woo site-address preflight:

- eligible `/connect/site-info` failures run device-side WordPress REST API discovery before blocking login;
- successful discovery shows a `Connection Issue` dialog that lets the user continue with site credentials;
- failed discovery keeps the user on the site-address step with the existing connection-style error;
- existing explicit non-fallback paths are preserved for malformed local URLs, HTTP 200 non-WordPress/not-found payloads, suspended WordPress.com sites, remote cURL 60 certificate failures, and Android TLS certificate validity failures.

A few details are intentional:

- The discovery path is opt-in via `FetchConnectSiteInfoPayload(discoverWPAPIOnFailure = true)`, and only the Woo site-address flow opts in. No-Jetpack checks, site picker, and other default callers keep their existing no-discovery behavior.
- WordPress.com network/DNS/timeout failures from the opted-in `/connect/site-info` request are also discovery-gated. Remote site timeouts are now returned by WordPress.com as structured `follow_redirects_failed` / cURL 28 responses, so local Volley `NO_CONNECTION`, `NETWORK_ERROR`, and `TIMEOUT` mean Android failed reaching WordPress.com. Direct REST discovery is a safe recovery gate for those cases.
- The login/app layer owns the fallback decision and prompt. FluxC only carries mechanism-level discovery state (`WPAPIDiscoveryResult`) plus the original error context.
- No fallback-specific Tracks funnel is added. The existing site-info failure event remains in place, and fallback attempts add targeted `AppLog` context for support/debugging.

### Test Steps
1. Install and open a debug build on an emulator or device.
2. Configure ApiFaker to make WordPress.com `/connect/site-info` return the same fallback-eligible error used in manual verification:

```bash
cat > /tmp/woomob2501-site-info-error.json <<'JSON'
{"error":"origin_failed","message":"origin failed"}
JSON

adb push /tmp/woomob2501-site-info-error.json /data/local/tmp/woomob2501-site-info-error.json

adb shell am broadcast \
  -p com.woocommerce.android.dev \
  -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS

adb shell am broadcast \
  -p com.woocommerce.android.dev \
  --es api_type wp-com \
  --es path /rest/v1.1/connect/site-info \
  --es http_method GET \
  --ei response_status_code 400 \
  --es response_body_file /data/local/tmp/woomob2501-site-info-error.json

adb shell am broadcast \
  -p com.woocommerce.android.dev \
  -a com.woocommerce.android.apifaker.SET_STATUS \
  --ez enabled true
```

3. Launch the Woo app and start the Woo login flow.
4. On the site-address screen, enter `wordpress.org` and tap Continue. This keeps `/connect/site-info` mocked while allowing live REST discovery at `https://wordpress.org/wp-json/`.
5. Verify the `Connection Issue` dialog appears with the message `We couldn't verify your site details. You can try logging in with your site credentials instead.` and the `Log In with Site Credentials` CTA.
6. Tap Cancel and verify the dialog closes, the app remains on the site-address screen, and `wordpress.org` is still present.
7. Tap Continue again, then tap `Log In with Site Credentials` and verify the site-credentials login screen opens for `wordpress.org`.
8. Optional logcat confirmation:

```bash
adb logcat -d | rg -i 'WCApiFaker|connect/site-info|Found valid WP-API endpoint|site-info fallback offered'
```

9. Cleanup ApiFaker after testing:

```bash
adb shell am broadcast \
  -p com.woocommerce.android.dev \
  -a com.woocommerce.android.apifaker.SET_STATUS \
  --ez enabled false

adb shell am broadcast \
  -p com.woocommerce.android.dev \
  -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS
```

10. As a preserved exclusion check, use a malformed local URL or a non-WordPress/not-found `/connect/site-info` success response and verify the fallback dialog is not shown.

### Images/gif
image

### DIFF
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
@@ -55,7 +55,6 @@ import org.wordpress.android.fluxc.tools.CoroutineEngine
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.UrlUtils
 import java.net.URI
-import java.net.UnknownHostException
 import java.security.cert.CertificateException
 import java.security.cert.CertificateExpiredException
 import java.security.cert.CertificateNotYetValidException
@@ -430,18 +429,25 @@ class SiteRestClient @Inject constructor(
             apiError == "connection_disabled" -> SiteError(SiteErrorType.WPCOM_SITE_SUSPENDED, message)
             isTlsCertificateValidityIssue(this) -> SiteError(TLS_CERTIFICATE_VALIDITY_ERROR, message)
             isRemoteSiteCertificateIssue(this) -> SiteError(REMOTE_SITE_CERTIFICATE_ERROR, message)
-            isWordPressComConnectivityIssue(this) -> SiteError(WORDPRESS_COM_CONNECTIVITY_ERROR, message)
+            isWordPressComConnectivityIssue(this) -> SiteError(
+                WORDPRESS_COM_CONNECTIVITY_ERROR,
+                message,
+                wpApiDiscovery = toWPAPIDiscoveryResult(discoverWPAPIOnFailure)
+            )
             else -> SiteError(
                 INVALID_SITE,
                 message,
-                wpApiDiscovery = if (discoverWPAPIOnFailure) {
-                    WPAPIDiscoveryResult(connectSiteInfoApiError = apiError)
-                } else {
-                    null
-                }
+                wpApiDiscovery = toWPAPIDiscoveryResult(discoverWPAPIOnFailure)
             )
         }
     }
+
+    private fun WPComGsonNetworkError.toWPAPIDiscoveryResult(discoverWPAPIOnFailure: Boolean) =
+        if (discoverWPAPIOnFailure) {
+            WPAPIDiscoveryResult(connectSiteInfoApiError = apiError)
+        } else {
+            null
+        }
 
     private fun isTlsCertificateValidityIssue(error: WPComGsonNetworkError): Boolean {
         if (error.type !in TLS_CERTIFICATE_VALIDITY_ERROR_TYPES) {
@@ -490,23 +496,9 @@ class SiteRestClient @Inject constructor(
         return when (error.type) {
             GenericErrorType.NO_CONNECTION,
             GenericErrorType.NETWORK_ERROR,
-            GenericErrorType.TIMEOUT -> {
-                val volleyError = error.volleyError ?: return false
-                val cause = volleyError.cause ?: return volleyError.message.containsWordPressComApiHost()
-                when (cause) {
-                    is UnknownHostException -> cause.message.containsWordPressComApiHost()
-                    else -> {
-                        volleyError.message.containsWordPressComApiHost() ||
-                            cause.message.containsWordPressComApiHost()
-                    }
-                }
-            }
+            GenericErrorType.TIMEOUT -> true
             else -> false
         }
-    }
-
-    private fun String?.containsWordPressComApiHost(): Boolean {
-        return this?.contains("public-api.wordpress.com", ignoreCase = true) == true
     }
 
     /**
--- a/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
+++ b/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
@@ -5,6 +5,7 @@ import android.text.TextUtils
 import androidx.annotation.VisibleForTesting
 import com.android.volley.RequestQueue
 import com.google.gson.reflect.TypeToken
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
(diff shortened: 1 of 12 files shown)


libs/login/src/main/java/org/wordpress/android/login/LoginSiteAddressFragment.java
```diff
@@ -223,7 +225,8 @@ protected void discover() {
         String cleanedUrl = stripKnownPaths(mRequestedSiteAddress);
 
         mAnalyticsListener.trackConnectedSiteInfoRequested(cleanedUrl);
-        mDispatcher.dispatch(SiteActionBuilder.newFetchConnectSiteInfoAction(cleanedUrl));
+        mDispatcher.dispatch(SiteActionBuilder.newFetchConnectSiteInfoAction(
+                new FetchConnectSiteInfoPayload(cleanedUrl, true)));
```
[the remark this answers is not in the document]
hichamboushaba: The LoginMode itself is just a legacy part from before when the login library was reused by different apps, so for simplicity, let's use `true` for all paths.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -400,26 +406,49 @@ class SiteRestClient @Inject constructor(
...
-                        when {
-                            isTlsCertificateValidityIssue(response.error) -> TLS_CERTIFICATE_VALIDITY_ERROR
-                            isRemoteSiteCertificateIssue(response.error) -> REMOTE_SITE_CERTIFICATE_ERROR
-                            isWordPressComConnectivityIssue(response.error) -> WORDPRESS_COM_CONNECTIVITY_ERROR
-                            else -> INVALID_SITE
-                        }
+                var siteError = response.error.toConnectSiteInfoError(discoverWPAPIOnFailure)
+                val discovery = siteError.wpApiDiscovery
+                if (discovery != null) {
+                    val wpApiBaseUrl = withContext(kotlinx.coroutines.Dispatchers.IO) {
```
AdamGrzybkowski:
```suggestion
                    val wpApiBaseUrl = withContext(Dispatchers.IO) {
```

Also, is it too complicated to inject the dispatcher?

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -467,18 +496,7 @@ class SiteRestClient @Inject constructor(
...
-                    when (cause) {
-                        is UnknownHostException -> {
-                            cause.message?.contains("public-api.wordpress.com", ignoreCase = true) == true
-                        }
-                        else -> {
-                            error.volleyError.message?.contains("public-api.wordpress.com", ignoreCase = true) == true
-                        }
-                    }
-                } ?: false
-            }
```
AdamGrzybkowski, line 481: ❓ This will now change the logic for all callers, not only those that set `discoverWPAPIOnFailure` to true, correct? 

Any other flows in the app that would be good to test?

AdamGrzybkowski (commented):
Testing steps confirmed ✅  but... what should I expect before that PR when using wordpress.org when logging in? 

I have assumed I would be blocked in some way, but when I test it, I can enter that URL, the app asks me for my WPCOM email and there I can use site credentials 🤔

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
hichamboushaba: This code was needed before because the endpoint `/site-info` was timing out when the site times out, but now, it doesn't, we get a specific error when it times out (as explained here https://linear.app/a8c/issue/WOOMOB-1044/better-error-handling-when-site-connectinfo-request-fails#comment-ab82c01c), this means that any `GenericErrorType.TIMEOUT` that we see from our side is specifically WPCom failure, and this shouldn't cause any change to the behavior.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
hichamboushaba: Fixed in a99c699dac7. I checked FluxC first: core FluxC exposes an injected default CoroutineContext via CoroutineEngine, but there is no existing IO dispatcher binding, so this keeps the change scoped to the requested Dispatchers import.

hichamboushaba:
> what should I expect before that PR when using wordpress.org when logging in?
> 
> I have assumed I would be blocked in some way, but when I test it, I can enter that URL, the app asks me for my WPCOM email and there I can use site credentials 🤔

Thanks @AdamGrzybkowski, are you sure the ApiFaker was used when you tested? Before this change, the app should block you, if `/site-info` is not able to identiy the site, the app shouldn't show you the WPCom login flow.

AdamGrzybkowski:
>Thanks @AdamGrzybkowski, are you sure the ApiFaker was used when you tested?

🤦 I did have it to test this PR, but disabled it for trunk... Sory!

AdamGrzybkowski (approved):
LGTM 👍 Thanks for the explanation!

PR #16087, opened 2026-06-16
[WOOMOB-2720] Surface POS order-detail refresh failures

malinajirka:
### Description

Fixes WOOMOB-2720

In POS, the order-detail refresh that runs after issuing a refund / sending an email receipt silently swallowed network failures, leaving stale data on screen with no feedback. This shows a snackbar when that refresh fails (and fails fast when offline). Uses an in-Compose snackbar because system `Toast`s don't render in the POS activity.

### Test Steps

1. POS → **Orders**, select a **Completed** order so it loads in the detail pane.
2. Tap **Issue refund** and complete a refund (any item).
3. On the **"Refund complete"** screen, **before** tapping Done, enable **Airplane mode** (or turn off Wi-Fi + mobile data).
4. Tap **Done**.
5. Expect: a snackbar at the bottom — *"Couldn't refresh the order. Please check your connection and try again."* — appears (near-)immediately, and the detail pane keeps showing the existing data instead of going blank/erroring.
6. Turn connectivity back on, refund another Completed order, and tap Done → no snackbar, and the detail pane updates to reflect the refund.

### Images/gif
Screenshot_1781616221


samiuelson (approved):
LGTM! :shipit:

PR #16088, opened 2026-06-16
[WOOMOB-3182] Fix orders missing from the order list until a filter or refresh

AdamGrzybkowski:
### Description
Fixes WOOMOB-3182

The order list is backed by FluxC's `ListStore`: the unfiltered **All** list and every filter combination are each a *separately cached, server-defined sequence of order ids* (one `ListEntity` + `ListItemEntity` set per descriptor `uniqueIdentifier`), while order data lives in its own tables. Because of this split, an order can be present locally yet missing from a given list until that **specific** list is refetched — and each list is only refetched when its 30‑minute freshness window (`ShouldUpdateOrdersList` / `LastUpdateDataStore`) expires.

Visible symptom: a newly‑created order shows under a filter (e.g. *Processing* / *Last 30 days*) but not under **All** — and the reverse too (after pull‑to‑refresh on **All**, a previously‑cached filter still omits the order). Orders appear/disappear when switching filters until a manual pull‑to‑refresh.

This change makes the lists self‑heal. When an order‑list fetch reveals an order id we have **never seen before** (checked against the order summary table *before* the fetched summaries are upserted), every **other** cached order list of the same type is marked `ListState.NEEDS_REFRESH`, so each refetches the next time it is shown. The just‑fetched list is excluded (it already contains the order), and the server decides membership on refetch, so an order only appears where it actually belongs. When no new order arrives, nothing is invalidated, so existing caching is preserved.

Implementation:
- New FluxC `ListStore` action `MARK_LISTS_OF_TYPE_NEED_REFRESH` (+ `MarkListsNeedRefreshPayload`) and `ListDao.markListsOfTypeNeedRefresh(typeId, excludedUniqueId)` — a single `UPDATE` setting `NEEDS_REFRESH` on all lists of a type except the excluded one. **No change event is emitted**, so the new state is consumed lazily by the existing `ShouldUpdateOrdersList` gate — avoiding both refetch loops and UI flashes.
- `WCOrderStore.handleFetchOrderListCompleted` detects new orders (vs the summary table) and dispatches the action, excluding the just‑fetched descriptor.
- No app‑layer change was needed; `ShouldUpdateOrdersList` already treats `NEEDS_REFRESH` as "must refetch".

### Test Steps
Reproduce with a real store (orders created outside the app).
1. Open **Orders** → ensure no filter (**All**); let it load.
2. Make sure a new order exists on the store that isn't yet in **All**.
3. Apply a filter that includes it (Status = *Processing*, or Date = *Last 30 days*) → the new order appears.
4. Clear the filter back to **All** *(without pull‑to‑refresh)* → the order now also appears in **All** (previously it was missing).
5. Reverse direction: pull‑to‑refresh **All** to load a new order, then open a previously‑used filter the order matches → it appears there too (previously missing until a manual refresh).

### Images/gif

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b218c58d-2366-41d2-920e-727eebf39e4a" /> | <video src="https://github.com/user-attachments/assets/f25e0153-a67a-4edf-bf51-0cac7d27ff3b" /> |

### DIFF
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
@@ -47,6 +47,7 @@ import org.wordpress.android.fluxc.persistence.entity.OrderNoteEntity
 import org.wordpress.android.fluxc.store.ListStore.FetchedListItemsPayload
 import org.wordpress.android.fluxc.store.ListStore.ListError
 import org.wordpress.android.fluxc.store.ListStore.ListErrorType
+import org.wordpress.android.fluxc.store.ListStore.MarkListsNeedRefreshPayload
 import org.wordpress.android.fluxc.store.ListStore.OnListDataFailure
 import org.wordpress.android.fluxc.store.WCOrderStore.OrderErrorType.GENERIC_ERROR
 import org.wordpress.android.fluxc.store.WCOrderStore.OrderErrorType.PARSE_ERROR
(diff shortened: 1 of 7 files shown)


libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -935,43 +936,82 @@ class WCOrderStore @Inject internal constructor(
...
+     */
+    private suspend fun markOtherOrderListsForRefreshIfNewOrders(
+        fetchedDescriptor: WCOrderListDescriptor,
+        fetchedSummaries: List<WCOrderSummaryModel>
+    ) {
+        if (fetchedSummaries.isEmpty()) return
+        val knownOrderIds = orderSummaryDao
+            .getOrderSummaries(fetchedDescriptor.site.localId(), fetchedSummaries.map { it.orderId })
+            .mapTo(HashSet()) { it.orderId.value }
+        val hasNewOrders = fetchedSummaries.any { it.orderId.value !in knownOrderIds }
```
hichamboushaba, line 1007: Remark here: This only invalidates sibling lists when the fetched order ID is new locally. So it fixes the “new order missing from cached lists” case, but not all stale-cache cases for existing orders. 

For example, if order #123 is already cached as Processing and gets changed to Completed outside the app, refreshing Processing would remove it from that list, but it still would not make it appear in a previously cached Completed list until Completed refreshes on its own.

I'm not sharing this as a blocker, since the PR improves and fixes one part of the issue, but I think it should still be called out that we're missing some edge cases.

hichamboushaba (approved):
Really nice work @AdamGrzybkowski, I left one remark, but it's not a blocker, unless you think fixing it on this same PR is simple enough.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
AdamGrzybkowski: That's a valid point, but it's tricky. Although you can implement this by adding another check to verify whether the order has changed, that won't always work both ways. 

When you are on a filtered page and looking only at `on hold` orders, when one of those orders changes the state, and you refresh the list, the Order is gone. So the comparison "whether it changed" won't work; the order is gone, there's nothing to compare. 

The above is solvable, but there are more edge cases, especially around modifying the Order state from the app. It feels like patching all possible vectors would be hard and a pain to maintain.

If we want a truly reactive state, maybe we should consider dropping the `ListStore` and using Room directly, like we did with Bookings. That is definitely out of scope here, though...

WDYT?

hichamboushaba: Yes, I agree, I just shared this to acknowledge there are more edge cases, but `ListStore` doesn't make this easy.

PR #16093, opened 2026-06-17
[WOOMOB-3214] Fix refund notice icon appearing after rotation on Issue Refund screen

malinajirka:
### Description

Fixes WOOMOB-3214

On the Issue Refund screen, a stray refund-notice info icon (plus a duplicate divider) appeared after rotating the device. Both views were referenced by the `issueRefund_totalsGroup`, which re-forces its members visible on rotation and overrode the fragment's `hide()`. Removed them from the group and defaulted them to `gone` so the controlling fragment owns their visibility; `RefundDetailFragment` now toggles the divider explicitly to preserve its behavior.

### Test Steps

1. Open a completed, refundable single-shipping order → **Issue refund**.
2. Confirm no info `ⓘ` icon below Taxes and a single divider above "Products refund".
3. Rotate the device and confirm the icon stays hidden and there's still a single divider.

### Images/gif
WOOMOB-3214_pr_overview


irfano (approved):
Thanks for handling this, LGTM! 👍🏻

PR #16094, opened 2026-06-17
[WOOMOB-3210] Time out expired number-challenge retries instead of re-showing stale number

JorgeMucientes:
Fixes WOOMOB-3210

### Description

In the QR login number-matching flow, the screen shows a 3-digit number with a 90-second countdown that the merchant must confirm in their browser/wp-admin session. If a **retryable** poll error (network blip, server error, rate limit) interrupts polling, the user lands on an error screen whose primary button is **"Try again"**.

The bug: each flow keeps a `retainedWaitingForApproval` snapshot captured at scan time, including a fixed `expiresAtEpochMs`. When the user sits on that "Try again" screen until the 90-second window elapses and then taps it, `retry()` called `resumePolling(retainedWaitingForApproval)`, which restored that stale snapshot. Because its timestamp is now in the past, the number screen reappeared showing **"Expires in 0s"** for a session the web side has already expired — a dead end, since there is no longer any way to confirm the number.

The fix adds an expiry guard to the poll-retry branch in both flows (`WpComQrLoginFlow` and `SiteQrLoginFlow`). If the retained challenge's window has already elapsed, the flow no longer resumes polling; instead it converges to the existing terminal timeout (`ErrorReason.MatchTimedOut`) — the very same state a clean server-side expiry already produces. Its error screen ("Sign-in timed out / Generate a new code in your store and try again") has a **"Scan a new code"** primary button that routes through `onStartOver` → `Idle` → the QR scanner, so the user can re-scan and start a fresh login attempt.

This keeps the in-window recovery behavior unchanged (a transient error followed by a prompt retry still resumes polling the same live session).

### Test Steps

This path is easiest to exercise by forcing a retryable poll error so the "Try again" screen appears, then letting the challenge expire before retrying:

1. Start the QR login flow and reach the number-challenge screen (3-digit number + "Expires in Xs" countdown).
2. Put the device into airplane mode (or otherwise drop connectivity) so polling fails. After a few seconds the **"Try again"** error screen appears.
3. Leave the app on that screen for **at least 90 seconds** so the original challenge window elapses, then re-enable connectivity.
4. Tap **"Try again"**.
5. **Expected:** the app shows the "Sign-in timed out" screen (not the old number with "Expires in 0s"); tapping **"Scan a new code"** returns to the QR scanner so you can scan a fresh code.
6. Regression check — retry still works mid-window: trigger the same network error but tap **"Try again"** *immediately* (well within the 90s window) with connectivity restored. Polling should resume on the same session and the flow should continue normally.

Applies to both the self-hosted (wp-admin) and wordpress.com QR login paths.

### Images/gif

https://github.com/user-attachments/assets/f3748b79-e190-47ed-bb4a-816aec0cdeae


malinajirka (approved):
LGTM, thanks!

PR #16099, opened 2026-06-18
Skip blocking POS catalog sync wait when the host blocks the catalog file

malinajirka:
### Description

When a store uses the POS **local catalog**, the catalog is generated server-side and downloaded as a file. Some hosts block that file (HTTP 403, or an HTML error/login page served instead of JSON). When that happens we surface a "contact your hosting provider" notice (Woo 11+) and fall back to remote sync so POS keeps working.

The problem: a blocked catalog **never completes a full sync**, so `checkSyncRequirement()` keeps returning `BlockingRequired` on every POS launch. Each launch then re-runs the full generate → poll → download flow, which can take **minutes**, only to hit the same 403 and fall back to remote anyway. Merchants on a blocking host were forced to wait every single time they opened POS, for a catalog they'd never actually get.

### Test Steps

Reproducing the block requires a host that 403s / serves HTML for the catalog file. The simplest way to simulate it is to make `WooPosCatalogFileDownloader.downloadCatalogFile()` return `Result.failure(WooPosCatalogFileBlockedException())` (e.g. gated on a marker file), then on a POS-local-catalog store:

1. Clear the POS sync state so a full sync is required (fresh install, or clear the `woo_pos_data_store` datastore), with the simulated block active.
2. Open POS. It should attempt the full sync, fail on the (simulated) block, and enter POS via remote fallback — showing the "contact your hosting provider" notice on Woo 11+. Logcat: `Blocking full sync failed (catalog file blocked)`.
3. Close and re-open POS. It should load **immediately** with **no** "Syncing Catalog…" wait and no generation/download attempt. Logcat: `Catalog file is blocked by the host; skipping the blocking sync wait`.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
@@ -1,0 +1,16 @@
+package com.woocommerce.android.ui.woopos.localcatalog
+
+import com.woocommerce.android.ui.woopos.util.datastore.WooPosSyncTimestampManager
+import javax.inject.Inject
+
+/**
+ * Reads the persisted "catalog file blocked by the host" flag. The flag is set when a full sync fails
+ * because the host blocks the generated catalog file (HTTP 403 / HTML body) and cleared on the first
+ * successful full sync. POS entry uses it to skip the blocking download wait that would otherwise
+ * repeat every time, while the background worker keeps retrying.
+ */
+class WooPosIsCatalogFileBlocked @Inject constructor(
+    private val syncTimestampManager: WooPosSyncTimestampManager,
+) {
+    suspend operator fun invoke(): Boolean = syncTimestampManager.isCatalogFileBlocked()
+}
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
@@ -49,7 +49,12 @@ class WooPosLocalCatalogSyncRepository @Inject constructor(
             when (result) {
                 is PosLocalCatalogSyncResult.Success -> {
                     syncTimestampManager.storeFullSyncLastCompletedTimestamp(dateTimeProvider.now())
+                    syncTimestampManager.setCatalogFileBlocked(false)
                     trackSyncCompleted(site, SyncType.FULL, result)
+                }
+                is PosLocalCatalogSyncResult.Failure.CatalogFileBlocked -> {
+                    syncTimestampManager.setCatalogFileBlocked(true)
+                    trackSyncFailed(SyncType.FULL, result)
                 }
                 is PosLocalCatalogSyncResult.Failure -> {
                     trackSyncFailed(SyncType.FULL, result)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosPerformInstantCatalogFullSync.kt
@@ -14,9 +14,19 @@ class WooPosPerformInstantCatalogFullSync @Inject constructor(
     private val syncTimestampManager: WooPosSyncTimestampManager,
     private val syncScheduler: WooPosLocalCatalogSyncScheduler,
     private val selectedSite: SelectedSite,
+    private val isCatalogFileBlocked: WooPosIsCatalogFileBlocked,
     private val wooPosLogWrapper: WooPosLogWrapper
 ) {
     suspend operator fun invoke(): Result<Unit> {
+        if (isCatalogFileBlocked()) {
+            // The host blocked the catalog file on a previous attempt and a full sync has never
+            // completed. Fail fast instead of waiting on catalog generation/download again (which would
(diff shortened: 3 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/291">25.0 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
```diff
@@ -49,8 +49,13 @@ class WooPosLocalCatalogSyncRepository @Inject constructor(
             when (result) {
                 is PosLocalCatalogSyncResult.Success -> {
                     syncTimestampManager.storeFullSyncLastCompletedTimestamp(dateTimeProvider.now())
+                    syncTimestampManager.setCatalogFileBlocked(false)
                     trackSyncCompleted(site, SyncType.FULL, result)
                 }
+                is PosLocalCatalogSyncResult.Failure.CatalogFileBlocked -> {
+                    syncTimestampManager.setCatalogFileBlocked(true)
```
samiuelson, line 56: If I understand correctly, once we detect problem with catalog file, permanently switch to fallback product sync. Do we plan to allow any way to switching back to file-based catalog sync after?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Reads the persisted "catalog file blocked by the host" flag. The flag is set when a full sync fails
+ * because the host blocks the generated catalog file (HTTP 403 / HTML body) and cleared on the first
+ * successful full sync. POS entry uses it to skip the blocking download wait that would otherwise
+ * repeat every time, while the background worker keeps retrying.
+ */
+class WooPosIsCatalogFileBlocked @Inject constructor(
+    private val syncTimestampManager: WooPosSyncTimestampManager,
+) {
+    suspend operator fun invoke(): Boolean = syncTimestampManager.isCatalogFileBlocked()
```
samiuelson, line 15: In case `WooPosIsCatalogFileBlocked()` is false, should we also disable "Product catalog" setting? Or maybe alternatively add some error info in the Product catalog setting screen about catalog file being blocked and a new setting/CTA allowing to retry and reset this flag 🤔

samiuelson (approved):
LGTM, thanks! Shared some thoughts in inline comments, none of them are blocking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
malinajirka: Good question - yeah, the background worker actually keeps trying the file sync every 24h -> when it succeeds, it sets the flag back to `false`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
malinajirka: The way I saw it is that the `update catalog` path in the settings is one of the mechanism for the user to retry. We already show a dedicated error in case we detect the 403 issue there, so the error message should be fine. In case the sync succeeds there, it also resets the flag back to `false`. Wdyt @samiuelson ?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
malinajirka: https://github.com/woocommerce/woocommerce-android/blob/e9a09c802ca6362d2cca9a7bb23fd4e786d4b797/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt#L52

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
malinajirka: https://github.com/woocommerce/woocommerce-android/blob/e9a09c802ca6362d2cca9a7bb23fd4e786d4b797/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt#L52

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosLocalCatalogSyncRepository.kt
samiuelson: Cool, thanks for explanation 👍

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/localcatalog/WooPosIsCatalogFileBlocked.kt
samiuelson: This makes sense, thanks 👍

PR #16100, opened 2026-06-18
[WOOMOB-3293] Constrain OrderNotificationWorker to network + fix app-password order sync

AdamGrzybkowski:
Fixes WOOMOB-3293

### Description

`OrderNotificationWorker` runs when an order push notification arrives and syncs the affected order (and the order list) in the background. The main goal of this PR was to add a constraint to the worker, but I found a bug that would make the worker fail every time in a specific scenario. More about that below.

**1. Add a `NetworkType.CONNECTED` constraint to the worker (WOOMOB-3293)**

The worker was enqueued as a `OneTimeWorkRequest` with no constraints, so WorkManager could run it while the device had no connectivity. Its work is entirely network-bound (fetch single order + refresh order list), so without a network it fails immediately and we emit a spurious `PUSH_NOTIFICATION_ORDER_BACKGROUND_SYNC_ERROR`. The fix attaches a `Constraints` with `setRequiredNetworkType(NetworkType.CONNECTED)`, so WorkManager defers the work until the device is online.

**2. Fix background order sync failing for site-credentials (application-password) logins**

While testing the constraint, the worker was still reaching a FAILED state with the network ON. Root cause: the worker resolved its target site **only** by the WordPress.com blog id, via `SiteStore.getSiteBySiteId(siteId)` — once for the single-order fetch (`UpdateOrderAndOrderList`) and once for the order-list refresh (`GetWCOrderListDescriptorWithFiltersBySiteId`). For a login done with **site credentials / application password**, the stored `SiteModel.siteId` (the wpcom blog id) is `0`, so the lookup returns null, the worker short-circuits to a "Site not found" error, and returns `Result.failure()`.

The codebase already solved this exact problem for analytics in `NotificationAnalyticsTracker`: for `SiteConnectionType.ApplicationPasswords` it falls back to the selected site (multi-site is unsupported with app passwords, so the single selected site is the target). This PR extracts that logic into a reusable, app-password-aware resolver and uses it for the worker's two site lookups:

- New `ResolveSiteBySiteId` in `com.woocommerce.android.tools` — a drop-in replacement for `SiteStore.getSiteBySiteId` that falls back to the selected site for app-password connections. It lives in `tools` (next to `SelectedSite`/connectivity helpers) rather than in `notifications` because it is a general site-resolution utility, not notification-specific.
- `UpdateOrderAndOrderList` and `GetWCOrderListDescriptorWithFiltersBySiteId` now use it (the latter's only caller is the worker path, so the blast radius is contained).
- `NotificationAnalyticsTracker` now reuses the shared resolver, removing its private duplicate.

Unit tests added/updated for the resolver and both consumers.

### Test Steps

This is background behavior, so the cleanest way to observe it is Android Studio's **Background Task Inspector** (App Inspection).

Setup:

In Android Studio open **View → Tool Windows → App Inspection**, select the app process, and open the **Background Task Inspector** tab.

Part A — network constraint:
1. Log into a store (any login type) and make sure order push notifications are enabled.
2. Put the device/emulator in **airplane mode** (or `adb shell svc wifi disable && adb shell svc data disable`).
3. Trigger an order notification (place/update an order on the store so a push is delivered).
4. In the inspector, find the `OrderNotificationWorker` entry and open it — confirm its **Constraints** list shows `NetworkType: CONNECTED`, and that the work stays **ENQUEUED** rather than running while offline.
5. Re-enable the network → the worker transitions to **RUNNING** → **SUCCEEDED**.

Part B — site-credentials (application-password) login:
1. Log in using **site credentials / application password** (a site that authenticates via app password rather than a WordPress.com account).
2. With the network ON, trigger an order notification for that store.
3. In the inspector, confirm `OrderNotificationWorker` reaches **SUCCEEDED** (before this change it ended in **FAILED** with a "Site not found" error), and that the order/order list is updated in the app.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/background/OrderNotificationWorker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/background/OrderNotificationWorker.kt
@@ -2,8 +2,10 @@ package com.woocommerce.android.background
 
 import android.content.Context
 import androidx.hilt.work.HiltWorker
+import androidx.work.Constraints
 import androidx.work.CoroutineWorker
 import androidx.work.Data
+import androidx.work.NetworkType
 import androidx.work.OneTimeWorkRequest
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
@@ -29,8 +31,13 @@ class OrderNotificationWorker @AssistedInject constructor(
             dataBuilder.putLong(ORDER_ID, remoteOrderId)
             val data = dataBuilder.build()
 
+            val constraints = Constraints.Builder()
+                .setRequiredNetworkType(NetworkType.CONNECTED)
+                .build()
+
             val workRequest = OneTimeWorkRequest.Builder(OrderNotificationWorker::class.java)
                 .setInputData(data)
+                .setConstraints(constraints)
                 .build()
 
             WorkManager.getInstance(context).enqueue(workRequest)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateOrderAndOrderList.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateOrderAndOrderList.kt
@@ -1,20 +1,20 @@
 package com.woocommerce.android.background
 
+import com.woocommerce.android.tools.ResolveSiteBySiteId
 import org.wordpress.android.fluxc.network.BaseRequest
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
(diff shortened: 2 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba (approved):
Nice catch, thanks @AdamGrzybkowski

PR #16102, opened 2026-06-18
Handle malformed site responses as parse errors

hichamboushaba:
### Description
Related Slack thread: p1781780401063659-slack-C6H8C3G23

An upstream WordPress.com bug caused `/rest/v1.1/sites/<site>` to respond with HTTP 200 and an error-shaped JSON body. That response is not valid for a successful site fetch, but FluxC previously continued as if the request had succeeded because the HTTP status was 200 and Gson could deserialize the body shape into `SiteWPComRestResponse`. When required site fields such as `URL` were absent, conversion could throw and crash the app during site refresh/startup.

This PR makes the site response conversion defensive and applies the same parse-error handling used by other FluxC endpoints when a successful HTTP response cannot be mapped into the expected model. `siteResponseToSiteModel` now returns `Result<SiteModel>`, and both single-site and site-list success paths convert mapping exceptions into `BaseNetworkError(GenericErrorType.PARSE_ERROR)`. The conversion failure is also emitted as `OnUnexpectedError("API response parse error")` with the request URL, matching the existing FluxC parse-error reporting path so crash/error reporting still sees the malformed response.

Regression tests cover malformed successful responses for both `/rest/v1.1/sites/<site>` and `/rest/v1.1/me/sites/`, asserting that FluxC returns `PARSE_ERROR` and emits the parse-error report instead of throwing.

### Test Steps
1. Install a debug build and use a device/emulator that is already logged in with a selected Jetpack-connected site.

2. Create and push the mocked WordPress.com 200 response body:

   ```bash
   cat > /tmp/pr16102-site-error-response.json <<'JSON'
   {"error":"unknown_blog","message":"Unknown blog"}
   JSON

   adb push /tmp/pr16102-site-error-response.json \
     /data/local/tmp/pr16102-site-error-response.json
   ```

3. Configure Api Faker for the malformed `/rest/v1.1/sites/<site>` response:

   ```bash
   adb logcat -c

   adb shell am broadcast -p com.woocommerce.android.dev \
     -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS

   adb shell am broadcast -p com.woocommerce.android.dev \
     --es api_type "wp-com" \
     --es path "/rest/v1.1/sites/%" \
     --es http_method "GET" \
     --es response_body_file "/data/local/tmp/pr16102-site-error-response.json"

   adb shell am broadcast -p com.woocommerce.android.dev \
     -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true

   adb shell am broadcast -p com.woocommerce.android.dev \
     -a com.woocommerce.android.apifaker.LIST_ENDPOINTS

   adb logcat -s WCApiFaker -d
   ```

   Confirm the log contains `Endpoint added - WPCom GET /rest/v1.1/sites/%` and lists the endpoint with status code `200`.

4. Restart the app to trigger the selected-site refresh:

   ```bash
   adb shell am force-stop com.woocommerce.android.dev
   adb shell am start -n \
     com.woocommerce.android.dev/com.woocommerce.android.ui.main.MainActivity
   ```

5. Check the mocked request and crash/error behavior:

   ```bash
   adb logcat -d | grep -E \
     "Matched request: https://public-api.wordpress.com/rest/v1.1/sites/|AppInitializer.onUnexpectedError|FATAL EXCEPTION|AndroidRuntime"

   adb shell pidof com.woocommerce.android.dev
   adb shell dumpsys window | grep -E "mCurrentFocus|mFocusedApp"
   ```

   Expected result: Api Faker logs a matched request for `/rest/v1.1/sites/...`, the app process remains alive in `MainActivity`, and there is no `FATAL EXCEPTION`/`AndroidRuntime` crash. The malformed 200 response is handled as a FluxC parse error and reported through `AppInitializer.onUnexpectedError`.

6. Clean up Api Faker after testing:

   ```bash
   adb shell am broadcast -p com.woocommerce.android.dev \
     -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled false

   adb shell am broadcast -p com.woocommerce.android.dev \
     -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS

   adb shell rm -f /data/local/tmp/pr16102-site-error-response.json
   ```

### Images/gif
N/A


libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -566,8 +575,29 @@ class SiteRestClient @Inject constructor(
         add(request)
     }
 
+    private fun siteResponseParseError(url: String, exception: Throwable): BaseNetworkError {
+        reportSiteResponseParseError(url, exception)
```
AdamGrzybkowski, line 579: This should make sure we will be able to spot such issues in the future, right? Although it's not great, the app crashed for many users, but that allowed us to catch the problem quickly, so it would be great if it stayed like this. 

If yes, will Sentry group those similarly so that a spike can be spotted?

AdamGrzybkowski (approved):
Looks good! Left one comment to make sure we don't lose the option to spot such issues in the future.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
hichamboushaba: > This should make sure we will be able to spot such issues in the future, right?

Yes, hopefully, as this would increase the number of parse errors that we monitor daily as part of the exceptions catcher rotation.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #16108, opened 2026-06-19
Surface non-JSON order responses through the orders error banner

AdamGrzybkowski:
### Description

Closes [WOOMOB-3298](https://linear.app/a8c/issue/WOOMOB-3298/surface-html-raw-body-response-failures-to-the-user).

When a merchant's store returns a non-JSON body while loading orders — a WAF/maintenance page, a PHP fatal/warning, or markup injected by a caching plugin — the app couldn't load the order list. How that surfaces depends on the networking path:

- **Direct REST / Application Passwords:** the corrupted body fails JSON parsing, which FluxC already maps to `OrderErrorType.PARSE_ERROR` → `ListErrorType.PARSE_ERROR`. The order list already shows the **"We couldn't load your data"** troubleshooting banner for this. ✅ (no change needed)
- **Jetpack tunnel proxy:** the WordPress.com proxy (see CONNECT-38) wraps the upstream bytes as a JSON error with code `no_response_body` and the raw payload in `data.raw_body`. The app mapped this to `GENERIC_ERROR` → the generic *"Error fetching orders"* snackbar, which gives the merchant no diagnosis. ❌

This PR closes that gap and improves the banner's action.

#### Changes

1. **Error parsing** — `OrderRestClient.wpAPINetworkErrorToOrderError` now maps a proxy error of `no_response_body` **with** `data.raw_body` present to `OrderErrorType.PARSE_ERROR`, so it reaches the existing banner. The occurrence is also reported through the existing FluxC parse-error reporting path (`OnUnexpectedError`, matching #16102) so the impact is measurable in error reporting. The raw body is intentionally not attached to the report as it may contain the merchant's page content. A proxy `no_response_body` *without* `raw_body` keeps its previous `GENERIC_ERROR` behavior.
2. **Troubleshooting navigation** — the parse-error banner's *Troubleshooting* button now opens the in-app **Troubleshoot Connection** screen (the same one reachable from Settings and already used by the timeout banner) instead of launching a web view. The now-unused `AppUrls.ORDERS_TROUBLESHOOTING` constant is removed.

### Testing

A small WordPress plugin reproduces the corrupted responses: **WC REST API Parse Error Simulator** - 
[wc-rest-api-parse-error-simulator.zip](https://github.com/user-attachments/files/29138155/wc-rest-api-parse-error-simulator.zip). It keeps the HTTP status `200` and `Content-Type: application/json` but prepends non-JSON content to the body, in two modes:
- **Mode 1 — PHP warning injection** (mirrors WC Cost of Goods v2.8.3). Hooks the REST *dispatch* path, so it also corrupts requests proxied through the Jetpack tunnel.
- **Mode 2 — HTML/script injection** (mirrors a caching plugin).

Install the zip via **wp-admin → Plugins → Add New → Upload Plugin**, activate it, then enable a mode from its settings screen. Use two Jurassic Ninja (JN) sites so you can cover both networking paths:

#### Scenario A — Application Passwords (direct REST)
1. On JN site nr 1 (not Jetpack connected, Application Passwords working), log into the app with site credentials.
2. Install + activate the simulator plugin and enable **Mode 1** (or Mode 2).
3. Pull to refresh the **Orders** list.
4. **Expected:** the **"We couldn't load your data"** troubleshooting banner appears (not the generic "Error fetching orders" snackbar). Tapping **Troubleshooting** opens the in-app **Troubleshoot Connection** screen; tapping **Contact support** opens the support form.

#### Scenario B — Jetpack tunnel proxy
1. On JN site nr 2, **disable Application Passwords** so the app falls back to the Jetpack tunnel — install a "Disable Application Passwords" plugin (or otherwise turn the feature off). This is required, otherwise the app uses the direct REST path (Scenario A).
2. Log into the app with this site selected and confirm requests go through the Jetpack tunnel.
3. Install + activate the simulator plugin and enable **Mode 1** (its dispatch-path hook is what corrupts the proxied response).
4. Pull to refresh the **Orders** list.
5. **Expected:** same as Scenario A — the troubleshooting banner appears instead of the generic snackbar, and the corrupted-response event is reported via `OnUnexpectedError("Order API response parse error")`.

After testing, deactivate/remove the simulator plugin (and re-enable Application Passwords on site #2).

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 25.1
 -----
 - [*] Fixed QR login retrying an expired number challenge re-showing the old number with "Expires in 0s"; it now returns you to the QR scanner to scan a fresh code [https://github.com/woocommerce/woocommerce-android/pull/16094]
+- [*] Show the orders troubleshooting banner (instead of a generic error) when your store returns an unexpected non-JSON response, and open the in-app Troubleshoot Connection screen from it [https://github.com/woocommerce/woocommerce-android/pull/16108]
 - [*] Renamed the empty Orders "test order" prompt to "Place your first order" and clarified that processing fees aren't refundable [https://github.com/woocommerce/woocommerce-android/pull/16086]
 - [*] Improved login error messages for site security certificate issues [https://github.com/woocommerce/woocommerce-android/pull/16042]
 - [Internal] Fixed a screen flash when navigating from the login prologue to the (feature-flagged) QR login screen [https://github.com/woocommerce/woocommerce-android/pull/16095]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppUrls.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppUrls.kt
@@ -32,9 +32,6 @@ object AppUrls {
         "https://jetpack.com/support/getting-started-with-jetpack/troubleshooting-tips/"
     const val PRODUCT_IMAGE_UPLOADS_TROUBLESHOOTING =
         "https://woocommerce.com/document/troubleshooting-image-upload-issues-in-the-woo-mobile-apps/"
-    const val ORDERS_TROUBLESHOOTING =
-        "https://woocommerce.com/document/android-ios-apps-troubleshooting-error-fetching-orders/"
-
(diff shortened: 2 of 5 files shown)


hichamboushaba (approved):
Works well, thanks @AdamGrzybkowski

PR #16116, opened 2026-06-22
[WOOMOB-3322] POS: present Issue Refund flow full screen from the start

samiuelson:
### Description

Fixes WOOMOB-3322

In Woo POS on tablets (dual-pane orders layout), tapping **Issue refund** did not open the refund flow full screen. The first step (item selection) rendered inside the **order details pane**. Only after tapping **Continue** did the flow expand to full screen, because the later steps are presented in a full-screen `Dialog` while the selection step was constrained to the detail pane.

**Fix:** in the dual-pane layout (`OrdersListWithDetails`), the refund flow is now presented full screen from the very first step instead of inside the detail pane, mirroring the rest of the flow. While the refund flow is active, the orders toolbar is hidden (the refund screen has its own header), so the condition that previously only hid it in single-order mode now hides it whenever a refund is in progress.

No ViewModel or refund logic changed — this is a presentation-layer fix. The phone (single-pane) flow was already full screen and is unaffected.

### Test Steps
1. On a tablet, open Woo POS → **Orders**.
2. Select an order that can be refunded.
3. Tap **Issue refund**.
4. Confirm the refund flow (loading + item selection) now appears **full screen** immediately — the orders list on the left is no longer visible, and the orders toolbar is hidden.
5. Select items and tap **Continue**; confirm the transition to the review step stays full screen with no jump or resize.
6. Use the refund screen's back/close to exit; confirm you return to the orders list + details as before.
7. On a phone, repeat the flow and confirm it is unchanged (still full screen, as before).

### Images/gif

* After
Screenshot_20260623_130530
Screenshot_20260623_130545

* Before
Screenshot_20260623_131751
Screenshot_20260623_131758


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
```diff
@@ -97,6 +97,16 @@ import kotlinx.coroutines.flow.filter
...
 
+/**
+ * The orders toolbar is shown only when no search is active and no refund is in progress. While a
+ * refund is being issued the flow is presented full screen with its own header, so the orders
+ * toolbar must stay hidden regardless of single- vs dual-pane mode.
+ */
+internal fun shouldShowOrdersToolbar(
+    searchInputState: WooPosSearchInputState,
+    isIssuingRefund: Boolean,
+): Boolean = searchInputState is WooPosSearchInputState.Closed && !isIssuingRefund
```
malinajirka: 🔍 Ideally, this would be coming from the VM - but maybe it's over-engineering, especially considering there are more places like this within this file.

malinajirka (changes requested):
Thanks @samiuelson!

It looks good overall, however, I noticed a couple issues:

1. The second step doesn't have a toolbar - or the `up` button.

Image

2. When the items are loading, there is a barely visible grey horizontal divider below the toolbar, which disappears as soon as the items finish loading:

Image


P.S. Back button on the second step feels a bit off, considering there is OS back button + likely we'll have the `up` button there too. Having said that, I don't mind having it.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/293">25.1</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/WooPosOrdersScreen.kt
samiuelson: moved to VM's UI state: fcb1c7b

samiuelson:
Thanks for review, @malinajirka 🙇 

> 1. The second step doesn't have a toolbar - or the `up` button.

> 2. When the items are loading, there is a barely visible grey horizontal divider below the toolbar, which disappears as soon as the items finish loading

Addressed in: 456c54e

Also aligned UI of `Review refund` step in: e022e15

Screenshot_20260624_114820

malinajirka (approved):
LGTM, thanks for the changes!

PR #16117, opened 2026-06-22
[WOOMOB-3294] Detect unknown_blog response and reset the selected site

AdamGrzybkowski:
### Description

Fixes WOOMOB-3294

WPCom returns the `unknown_blog` error code when the app sends a site ID it no longer recognises — e.g. the selected site was deleted, disconnected from Jetpack, or the user lost access. Today the app has no recovery: once a site becomes "unknown", every WooCommerce request keeps failing with the same error and the merchant is stuck on a broken dashboard

This PR detects that error and recovers from it:

- **Detect (`WooNetwork`)** — `WooNetwork.handleRequest()` is the single choke point every WooCommerce REST call goes through (both the Jetpack tunnel and direct Application-Passwords), and `unknown_blog` is preserved in `WPAPINetworkError.errorCode` on both paths. When a response for the selected site carries that code, it notifies an optional `UnknownBlogListener`.
- **Recover (`AppInitializer`)** — `UnknownBlogNotifier` (the app-side listener, wired via Hilt) exposes the event as a flow. `AppInitializer` observes it and, when the failing site matches the currently selected one, tracks an event, resets the selected site, and restarts into the site picker. Recovery only runs while the app is in the **foreground** (resetting in the background would strip the site without being able to route anywhere; when the app next becomes visible it re-fetches the selected site, the error recurs, and recovery runs then).
- **Explain (site picker)** — the picker is opened as a store switcher (not the login flow) so it doesn't silently auto-select the failing store, and it shows a snackbar explaining why the store was signed out. The message is carried through a pref as a string-resource id, so it can be reused for other recovery reasons.
- **Durability fix** — `SelectedSite.reset()` and the pending-message write are persisted synchronously (`commit()` instead of `apply()`). The recovery restart can kill the process (`UncaughtErrorsHandler.exitProcess`) before an async write flushes, which previously left the relaunched process reading stale prefs and bouncing back to the broken dashboard.

Analytics: tracks `selected_site_reset_due_to_unknown_blog` when recovery fires (matches the iOS sibling).

### Test Steps

This is reproduced with the built-in API Faker (debug builds), which can return a fake `unknown_blog` for WooCommerce endpoints. See `docs/api-faker-adb.md`.

1. Install the **debug** build and log in with a **WordPress.com** account that has a connected store, then open the **Orders** tab and let it load normally.
2. Enable API Faker to return `unknown_blog` for the orders endpoint (run with a device/emulator connected; adjust the `-p` package if needed):
   ```bash
   PKG=com.woocommerce.android.dev
   APIFAKER=com.woocommerce.android.apifaker
   printf '%s' '{"error":"unknown_blog","code":"unknown_blog","message":"Unknown blog"}' > /tmp/unknown_blog.json
   adb push /tmp/unknown_blog.json /data/local/tmp/unknown_blog.json
   adb shell am broadcast -p $PKG -a $APIFAKER.CLEAR_ENDPOINTS
   adb shell am broadcast -p $PKG -a $APIFAKER.ADD_ENDPOINT --es api_type wp-api --es path /wc/v3/orders --ei response_status_code 404 --es response_body_file /data/local/tmp/unknown_blog.json
   adb shell am broadcast -p $PKG -a $APIFAKER.SET_STATUS --ez enabled true
   ```
   Confirm a red "ApiFaker Enabled" banner appears at the bottom of the app.
3. On the **Orders** tab, **pull to refresh**. The orders request now returns `unknown_blog`.
4. **Expected:** the app resets the selected store and lands on the **site picker** (as a store switcher — it does *not* auto-open a store), showing a snackbar: *"We couldn't access your store, so it's been signed out. Please select a store to continue."*
5. Disable API Faker and pick your store again to return to normal:
   ```bash
   adb shell am broadcast -p com.woocommerce.android.dev -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled false
   ```

Optional: watch the flow in logs with `adb logcat -s WooCommerce-LOGIN WCApiFaker`.

### Images/gif

https://github.com/user-attachments/assets/255d0975-5aa0-43cb-898f-8517018c169c


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
```diff
@@ -280,6 +281,19 @@ object AppPrefs {
...
+    // Persisted synchronously because it is written right before a recovery restart that may kill the
+    // process before an async apply() flushes.
+    var sitePickerErrorMessage: Int
+        get() = getInt(DeletablePrefKey.SITE_PICKER_ERROR_MESSAGE, 0)
+        set(value) {
+            getPreferences()
+                .edit()
+                .putString(DeletablePrefKey.SITE_PICKER_ERROR_MESSAGE.toString(), value.toString())
+                .commit()
+        }
```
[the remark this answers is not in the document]
AdamGrzybkowski: Good catch — fixed in 4bc2249. The message is now committed synchronously only when set to a non-zero value (right before the recovery restart); clearing it (value 0, from the site picker init on the main thread) uses `remove().apply()` to avoid blocking disk I/O.

libs/commons/src/main/java/com/woocommerce/android/tools/SelectedSite.kt
```diff
@@ -113,7 +113,9 @@ class SelectedSite @Inject constructor(
     fun reset() {
         wasReset = true
         state.value = null
-        getPreferences().edit().remove(SELECTED_SITE_LOCAL_ID).apply()
+        // Persist synchronously: recovery flows restart (and may kill) the process right after, and an
+        // async apply() can be lost before it flushes, leaving the site still "selected" on relaunch.
+        getPreferences().edit().remove(SELECTED_SITE_LOCAL_ID).commit()
```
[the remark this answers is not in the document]
AdamGrzybkowski: Done in 4bc2249. `reset()` keeps `apply()` by default; I added an opt-in `persistSynchronously` parameter and only the `unknown_blog` recovery passes `persistSynchronously = true`. Normal UI-thread resets (logout, `showLoginScreen()`, etc.) keep `apply()`, avoiding main-thread I/O / StrictMode.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooNetwork.kt
```diff
@@ -123,6 +125,14 @@ class WooNetwork @Inject constructor(
                 throw IllegalArgumentException("Unsupported site origin: ${site.origin}")
             }
         }
+        notifyIfUnknownBlog(site, response)
+        return response
+    }
+
+    private fun <T : Any> notifyIfUnknownBlog(site: SiteModel, response: WPAPIResponse<T>) {
+        if (response is WPAPIResponse.Error && response.error.errorCode == UNKNOWN_BLOG_ERROR_CODE) {
+            unknownBlogListener.takeIf { it.isPresent }?.get()?.onUnknownBlog(site.siteId)
```
irfano:
```suggestion
            unknownBlogListener.ifPresent { it.onUnknownBlog(site.siteId) }
```

np: Using `ifPresent` would be simpler here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -81,6 +81,12 @@ class SitePickerViewModel @Inject constructor(
 
     private val navArgs: SitePickerFragmentArgs by savedState.navArgs()
 
+    // Non-zero when the app routed to the picker to recover from an error; read and consumed once to
+    // show the message. Auto-login is prevented separately by opening the picker with openedFromLogin=false.
```
irfano: > Auto-login is prevented separately by opening the picker with openedFromLogin=false.

np: this is already documented where `openedFromLogin=false` is set (MainActivity / MainActivityViewModel), so repeating it here on the message field reads as redundant. Could drop it. Feel free to ignore.

irfano (approved):
Good job @AdamGrzybkowski! It works as expected. I only have one non-blocking note.

Image

After recovery, the store picker shows a back button, but there's no back stack here, so tapping it just closes the app. I'd prefer to hide it in this case.

AdamGrzybkowski:
Thanks @irfano - good feedback! The PR was updated - let me know if you want to check it again

irfano (approved):
I tested it and LGTM! Thanks for changes. :shipit:

libs/commons/src/main/java/com/woocommerce/android/di/UnknownBlogListenerModule.kt
irfano, line 1: @AdamGrzybkowski, I noticed that in the sibling PR #16132 the optional binding (`@BindsOptionalOf`) is declared once in commons, so there's no Wear module. Here we declare it per app instead (main + Wear). Could we align this PR with that approach for consistency?

AdamGrzybkowski: Good point! Let me update that.

AdamGrzybkowski: Done!

PR #16120, opened 2026-06-22
Add Store design system foundations module

hichamboushaba:
Fixes WOOMOB-3329

### Description

This PR builds on the design-system docs branch and adds the Android foundation module for the Woo Mobile Design System.

#### What this adds

- A Store-only `:libs:store-design-system` module, separate from `:libs:commons`.
- Gradle/settings wiring for the new module.
- `WooDesignSystemTheme` and `WooDesignSystemThemeWithBackground` as the opt-in design-system root.
- `WooTheme.*` accessors for foundation values:
  - colors
  - text
  - spacing
  - padding
  - radius
  - icon size
- XML-backed Store color roles for Compose and future XML/View usage, with Material 3 projections for Compose interop.
- Public palette ramps, including gray, exposed through `WooTheme.colors.palette.*`.
- Source-backed container, surface, status, overlay, and alert roles, including `alert.orange` and alert on-colors.
- Foundation primitives for spacing, padding, radius, icon size, internal stroke, and typography.
- Typography scale, line heights, letter spacing, and weights. Runtime font-family strategy remains pending.
- Foundation previews for light/dark and large-font review.
- Jacoco config so the new module participates correctly in coverage tooling.
- Compose UI tooling runtime moved to debug-only configurations for the app, Wear app, and design-system module.
- Review cleanup that colocates composition locals with their token files and keeps invariant palette XML resources defined once.

#### Why this shape

The goal is to land the foundation layer before components or product-screen migration depend on it. This keeps the first implementation PR focused on the Android API, token shape, module boundary, and preview surface.

The module boundary is intentional:

- Pure foundations and future production DS components live in `:libs:store-design-system`.
- App rollout policy, `composeView`, feature flags, legacy app resources, and migrated screen previews stay in `:WooCommerce`.
- The design-system module should not depend on app `R`, Hilt, `FeatureFlag`, POS, or Store feature packages.

#### What this does not do

- Does not add DS components yet.
- Does not migrate product screens.
- Does not touch POS.
- Does not globally remap XML/View styles or legacy app themes.
- Does not settle the runtime font-family strategy for the source `Roboto` typography tokens.
- Does not expose unresolved state-layer semantics as a public API.
- Does not add Code Connect, generated tokens, or Figma automation.

#### Helpful review focus

- Is `:libs:store-design-system` the right boundary for Store-only foundations?
- Are the public names (`WooDesignSystemTheme`, `WooTheme.*`) clear enough for migration-era use?
- Are the token groups practical for product-screen and component authors without exposing raw Figma names?
- Does the Material 3 projection stay internal enough while still making M3 components usable?
- Are the previews enough to review the foundation output before component work starts?

#### Context references

- Docs base branch: `store-design-system/docs`
- Woo Mobile Design System i1 P2: `pe5sF9-5ox-p2`
- Android adoption RFC: `pe5sF9-5sW-p2`
- Mobile Design System Integration Strategy P2: `pe5sF9-5vr-p2`

No release notes: this is an internal design-system foundation change.

### Test Steps

1. Open `WooDesignSystemFoundationPreview` in `:libs:store-design-system`.
2. Render the light/dark preview.
3. Render the large-font preview and verify the token catalog still lays out correctly.

### Images/gif

N/A - this is an internal foundation module; render the previews from the test steps for the current token catalog.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,82 @@
...
+    <color name="woo_ds_color_status_caution_container">#E6FEE995</color>
+    <color name="woo_ds_color_status_on_caution_container">#281D00</color>
+    <color name="woo_ds_color_status_success_container">#E6C6F7CD</color>
+    <color name="woo_ds_color_status_on_success_container">#002900</color>
+    <color name="woo_ds_color_status_info_container">#E6DEEBFA</color>
+    <color name="woo_ds_color_status_on_info_container">#001B4F</color>
+    <color name="woo_ds_color_status_neutral_container">#E6F4F4F4</color>
+    <color name="woo_ds_color_status_on_neutral_container">#1E1E1E</color>
+    <color name="woo_ds_color_status_neutral_outlined_container">#E6DBDBDB</color>
+    <color name="woo_ds_color_status_on_neutral_outlined_container">#DBDBDB</color>
```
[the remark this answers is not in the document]
hichamboushaba: There was an issue, the color is not meant as a `container`, but as a border, I removed the `container` suffix, and updated the token-map file.

WooCommerce/build.gradle
```diff
@@ -445,7 +445,7 @@ dependencies {
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.compose.material.main)
     implementation(libs.androidx.compose.animation.main)
-    implementation(libs.androidx.compose.ui.tooling.main)
+    debugImplementation(libs.androidx.compose.ui.tooling.main)
```
hichamboushaba, line 454: This change is unrelated to the PR, just an improvement, the ui-tooling library is needed only for debug builds.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooColors.kt
```diff
@@ -0,0 +1,283 @@
...
+    val shade60: Color,
+    val shade70: Color,
+    val shade80: Color,
+    val shade90: Color,
+    val shade100: Color,
+)
+
+internal val LocalWooColors = staticCompositionLocalOf<WooColors> {
+    error("WooTheme.colors is not available. Wrap content in WooDesignSystemTheme.")
+}
```
AdamGrzybkowski, line 152: ⛏️ Some `Local...` properties are defined in `WooDesignSystemTheme` and in separate files, like here - is that on purpose?

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/preview/WooDesignSystemFoundationPreview.kt
```diff
@@ -0,0 +1,383 @@
...
+import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import com.woocommerce.android.ui.compose.designsystem.WooTheme
+import com.woocommerce.android.ui.compose.designsystem.foundation.DefaultWooStroke
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooDesignSystemThemeWithBackground
+
+@PreviewLightDark
+@Composable
+private fun WooDesignSystemFoundationPreview() {
```
AdamGrzybkowski, line 39: ❓ Would it make sense to put this in a debug src set?

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/WooTheme.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.woocommerce.android.ui.compose.designsystem.foundation.LocalWooRadius
+import com.woocommerce.android.ui.compose.designsystem.foundation.LocalWooSpacing
+import com.woocommerce.android.ui.compose.designsystem.foundation.LocalWooText
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooColors
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooPadding
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooRadius
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooSpacing
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooTypography
+
+object WooTheme {
```
AdamGrzybkowski, line 20: ❓ Any other alternatives for this name? We already have one `WooTheme` so when typing and using autocomplete (who does that these days?!) we will have to make sure to pick the correct one.

libs/store-design-system/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,82 @@
...
+
+    <color name="woo_ds_color_overlay_overlay20">#33000000</color>
+    <color name="woo_ds_color_overlay_overlay50">#BF000000</color>
+
+    <color name="woo_ds_color_alert_red">#DC3545</color>
+    <color name="woo_ds_color_alert_yellow">#EAAB2D</color>
+    <color name="woo_ds_color_alert_green">#69B66F</color>
+    <color name="woo_ds_color_alert_blue">#1E94D0</color>
+
+    <color name="woo_ds_color_palette_sandstone_shade5">#FBF9F6</color>
```
AdamGrzybkowski: ❓ All the palette colors are duplicated in both files - would it make sense to keep them defined in Kotlin? It's not something we would change often, but having the same color in two places could cause a drift in the future.

AdamGrzybkowski: Additionally, the `WooPaletteColors` is once created with the XML colors and once with the colors defined directly in Kotlin - is there a reason for that? (`WooFoundationDefaults.palette` reads the Kotlin copy, `WooTheme.colors.palette` reads the XML copy.)

AdamGrzybkowski (commented):
>Are the previews enough to review the foundation output before component work starts?

Enough? Probably, but I think we could quite easily add either a screen within the Debug panel or a completely new target to browse all the components in the future. That would give us a real feel of them, being clickable and all - also for Wagner. 

That's more for the future - not this PR ofc.

libs/store-design-system/src/main/res/values-night/colors.xml
hichamboushaba: Good points; both are just oversights due to a last-minute decision to move the colors to XML. I will fix both.

> would it make sense to keep them defined in Kotlin?

Given we have many XML screens in the app, I decided to move the colors to XML, to make them usable for any styles that we might need to keep the app consistent.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/WooTheme.kt
hichamboushaba: My plan is to rename the other `WooTheme` to `WooLegacyTheme` in the feature branch when we are done with the first wave of migration.

But anyway, given this p1782389669493429-slack-C047J88KD1D, things might change, depending on the changes, I will see whether we need a `WooTheme` for locals anyway or not.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/preview/WooDesignSystemFoundationPreview.kt
hichamboushaba: I'm not sure there is any advantage to that, these will be stripped anyway from the release build by R8, and being close to the code is better IMO.
Do you think there are any advantages to moving it?

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooColors.kt
hichamboushaba: Will look into that.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/preview/WooDesignSystemFoundationPreview.kt
AdamGrzybkowski: No, not really - it's true they should be stripped by R8.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooColors.kt
hichamboushaba: This was addressed in last changes

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/WooTheme.kt
hichamboushaba: I kept the `WooTheme` name, with the plan to remove the existing `WooThemeWithBackground` to `LegacyWooTheme...` when we are done with the first wave of migrations. If anyone has a better suggestion, please let me know.

libs/store-design-system/src/main/res/values-night/colors.xml
hichamboushaba: These points were addressed.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooStroke.kt
```diff
@@ -0,0 +1,29 @@
+package com.woocommerce.android.ui.compose.designsystem.foundation
+
+import androidx.compose.runtime.Immutable
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@Immutable
+@Suppress("LongParameterList")
+internal data class WooStroke(
```
AdamGrzybkowski: ❓ This is not exposed via `WooTheme` - is that on purpose?

libs/store-design-system/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,54 @@
...
+<resources>
+    <color name="woo_ds_color_primary">#873EFF</color>
+    <color name="woo_ds_color_on_primary">#FFFFFF</color>
+    <color name="woo_ds_color_secondary">#383146</color>
+    <color name="woo_ds_color_on_secondary">#F1EDFE</color>
+
+    <color name="woo_ds_color_container_primary_container">#FFFFFF</color>
+    <color name="woo_ds_color_container_on_primary_container">#FFFFFF</color>
+    <color name="woo_ds_color_container_secondary_container">#FFFFFF</color>
+    <color name="woo_ds_color_container_on_secondary_container">#FFFFFF</color>
```
AdamGrzybkowski, line 11: ⚠️ This looks wrong. Both the container and the `on` tokens share the same color.

libs/store-design-system/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,54 @@
...
+    <color name="woo_ds_color_overlay_overlay50">#BF000000</color>
+
+    <color name="woo_ds_color_alert_red">#DC3545</color>
+    <color name="woo_ds_color_alert_on_red">#DC3545</color>
+    <color name="woo_ds_color_alert_orange">#EAAB2D</color>
+    <color name="woo_ds_color_alert_on_orange">#EAAB2D</color>
+    <color name="woo_ds_color_alert_green">#69B66F</color>
+    <color name="woo_ds_color_alert_on_green">#69B66F</color>
+    <color name="woo_ds_color_alert_blue">#1E94D0</color>
+    <color name="woo_ds_color_alert_on_blue">#1E94D0</color>
```
AdamGrzybkowski: ⚠️ Same issue here with the colors.

libs/store-design-system/src/main/res/values-night/colors.xml
hichamboushaba: Good eye, this is a known issue of dark colors p1782836171453349/1782299042.411359-slack-C03L1NF1EA3

I decided to keep the `figma-export` as the source of truth, once Wagner fixes the colors, we'll use another export and fix the colors.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooStroke.kt
hichamboushaba: Good point. The idea was not to expose it until it's needed, but I think it should be exposed, given it's already needed for the components' work. I'll fix it.

AdamGrzybkowski (approved):
Thanks for the answers!

hichamboushaba:
> Enough? Probably, but I think we could quite easily add either a screen within the Debug panel or a completely new target to browse all the components in the future. That would give us a real feel of them, being clickable and all - also for Wagner.

@AdamGrzybkowski I didn't answer this part, I think this would be a good idea for the components preview itself, I will do that. For the foundations, it's not interactable, so the usefulness is limited, but I'll keep it in mind.

libs/store-design-system/src/main/res/values-night/colors.xml
hichamboushaba: Fixed in 9cd8660

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooStroke.kt
hichamboushaba: Done in 64b84e7

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooSpacing.kt
```diff
@@ -0,0 +1,44 @@
...
+    val space4: Dp,
+    val space5: Dp,
+    val space6: Dp,
+    val space7: Dp,
+    val space8: Dp,
+    val space9: Dp,
+    val space10: Dp,
+    val space11: Dp,
+    val space12: Dp,
+)
```
hichamboushaba, line 24: Please let me know what you think about exposing spacings like this?

I'm on the fence about this approach, on one hand it has few advantages:

- It mirros the Figma names exactly.
- It's flexible so that we can have different spacing values for different layouts, for example we can think about bigger spacings for tablets.

But on the other, the fact that we hide the actual values of spacings makes reasoning harder, so we need to understand that `spacing5` corrsponds to `16.dp`, this is similar to what we did with XML dimensions where we had `major_100` matching 16dp.

AdamGrzybkowski: >this is similar to what we did with XML dimensions where we had major_100 matching 16dp.

I never liked this (especially when not being familiar with this system and the reasoning behind it). My issue wasn't the naming but the fact that I never saw those corresponding names in Figma. If the new designs in Figma will use the names that map to the values we use in the new DS foundations and we won't have to map those ourselves on the fly, I don't mind having such names in the codebase - in fact, it will be easier to just use the value we see in Figma.

hichamboushaba: Yes, I understand the concern, and that why I said I'm on the fence about it, the discoverability of raw values makes reasoning easier, but it has its drawbacks.

> If the new designs in Figma will use the names that map to the values we use in the new DS foundations and we won't have to map those ourselves on the fly, I don't mind having such names in the codebase - in fact, it will be easier to just use the value we see in Figma.

Yes, that's the goal, the current examples we have in Figma use them:

Image

But it's up to the designers to respect that.

-----

For me, the other point I mentioned above about being able to have different values depending on the device config (screen size, device scale...) also seems valuable, but not sure whether we'll use it.

hichamboushaba: Anyway, I think it's OK to use these now, in the worst case if this becomes an issue, we can just rename them to `spacingX` for `Xdp`, and it will be easier to reason about.

AdamGrzybkowski: Agree 👍

PR #16122, opened 2026-06-23
[POS][New refunds API] Part 1. FluxC: v4 refund preview + simplified-create APIs

samiuelson:
### Part of a 3-PR stack — review & merge in order

1. **#16122 — FluxC: v4 refund preview + simplified-create APIs ← you are here**
2. #16123 — POS: v4 refund detection, preview & simplified-create plumbing
3. #16124 — POS: wire v4 refund preview, detection & creation into the flow

Each part is stacked on the previous, so its diff only shows that part's changes. Split from #16109.

### Description

Part of WOOMOB-2687, WOOMOB-2689 and WOOMOB-2693 — these issues are closed by Part 3 (#16124), once the whole stack lands.

Part 1 of 3 of the v4 refunds integration for Woo POS. This PR adds the **FluxC data layer** only — no UI and no behaviour change in the app yet (nothing calls these methods until Part 3).

The WooCommerce v4 refund endpoints shipped server-side in WooCommerce 10.9.0 (behind the `rest-api-v4` feature flag). This adds the Android client to talk to them:

- **Endpoints**: `wc/v4/refunds` and `wc/v4/refunds/preview` (generated from `wc-wp-api-endpoints.txt`).
- **`RefundV4LineItem`** — the simplified per-line request: `line_item_id` + `quantity` for products, or `refund_total` for fee/shipping lines.
- **`RefundPreviewRestClient`** → `POST /wc/v4/refunds/preview`, returning the server-calculated breakdown. Adds the `RefundPreviewResponse` DTO, the `WCRefundPreview` domain model, and `RefundMapper.toPreviewModel`.
- **`RefundRestClient.createSimplifiedRefund`** → `POST /wc/v4/refunds` (sends only what to refund — no client-calculated `amount`).
- **`WCRefundStore.previewRefund` / `createSimplifiedItemsRefund`** — store-level wrappers returning domain models.

When the route is not registered (older WC, or the flag is off), the request fails with `404 rest_no_route`, surfaced as `WooErrorType.API_NOT_FOUND` — the signal the POS layer (Part 2) uses to fall back to v3.

### Test Steps

Data-layer only — no user-facing change. Covered by unit tests: `RefundPreviewRestClientTest`, `RefundRestClientTest`, `RefundMapperTest`, `RefundStoreTest`.

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #16109, opened 2026-06-19
[WOOMOB-2687] Integrate v4 refund preview/creation into POS with v3 fallback

samiuelson:
### Description

Fixes WOOMOB-2687
Fixes WOOMOB-2689
Fixes WOOMOB-2693

Integrates the new WooCommerce v4 refund APIs (shipped server-side in WooCommerce 10.9.0, behind the `rest-api-v4` feature flag) into the Android POS refund flow, with a transparent fallback to the existing v3 + local-calculation flow when a store does not expose them.

**Why:** today the POS refund flow replicates the backend's refund math client-side — product subtotals, per-line and per-order taxes, currency precision, and tax rounding mode — just to call the v3 endpoint. Those calculations depend on store-level settings that can change at any time, so the client and server can silently disagree by a cent, and the preview a merchant confirms may not match the refund the server actually creates. The v4 endpoints move that math to the server: the client sends only *what* to refund, and the server returns the authoritative totals.

This is the Android counterpart to the backend stack in the [POS Refunds v4 API spec](https://peacockp2.wordpress.com/2025/12/19/pos-refunds-api-enhancements-v4/) (backend PRs [#65334](https://github.com/woocommerce/woocommerce/pull/65334), [#65335](https://github.com/woocommerce/woocommerce/pull/65335), [#65439](https://github.com/woocommerce/woocommerce/pull/65439), [#65645](https://github.com/woocommerce/woocommerce/pull/65645)).

**What changed**

FluxC (shared data layer):
- New top-level endpoints `/wc/v4/refunds/` and `/wc/v4/refunds/preview/` (generated from `wc-wp-api-endpoints.txt`).
- `RefundV4LineItem` request model — `{line_item_id, quantity?, refund_total?}`. Gson omits nulls, so product lines send `quantity` and fee lines send a tax-inclusive `refund_total` (the amount-based form added in backend #65645).
- `RefundPreviewRestClient` → `POST /wc/v4/refunds/preview`, with a `RefundPreviewResponse` DTO and a `WCRefundPreview` domain model + mapper.
- `RefundRestClient.createSimplifiedRefund` → `POST /wc/v4/refunds` (no client-calculated `amount`/`refund_total`).
- `WCRefundStore.previewRefund(...)` and `WCRefundStore.createSimplifiedItemsRefund(...)`.

POS:
- `WooPosV4RefundAvailabilityCache` — caches v4 availability per site for the session, keyed by remote site id so switching stores never reuses a stale result.
[...]

PR #16123, opened 2026-06-23
[POS][New refunds API] Part 2. POS: v4 refund detection, preview & simplified-create integration

samiuelson:
### Part of a 3-PR stack — review & merge in order

1. #16122 — FluxC: v4 refund preview + simplified-create APIs
2. **#16123 — POS: v4 refund detection, preview & simplified-create plumbing ← you are here**
3. #16124 — POS: wire v4 refund preview, detection & creation into the flow

Each part is stacked on the previous, so its diff only shows that part's changes. **Stacked on #16122 — review/merge that first.** Split from #16109.

### Description

Part of WOOMOB-2693, WOOMOB-2689 and WOOMOB-2687 — these issues are closed by Part 3 (#16124), once the whole stack lands.

Part 2 of 3 of the v4 refunds integration for Woo POS. This adds the POS **domain layer** — the units that Part 3 wires into the ViewModel and UI. On their own they are inert: nothing in the refund flow calls them yet, so the app still behaves exactly as before (v3). No user-facing change in this PR.

- **`WooPosV4RefundAvailabilityCache`** — caches, per site for the app session, whether the store exposes the v4 refund endpoints (keyed by remote site id, so switching stores never reuses a stale result).
- **`WooPosRefundPreview`** — the detection + preview use case:
  - The v4 endpoints ship in WooCommerce 10.9.0. If the **cached** WooCommerce version is older, v4 cannot be available, so it skips the probe entirely (no network call) and falls back to v3.
  - On 10.9.0+ the endpoints may still be behind the `rest-api-v4` flag, so it probes `POST /wc/v4/refunds/preview`. A `404 rest_no_route` marks the site unavailable and falls back; a success caches availability and returns the server totals.
- **`WooPosBuildRefundV4LineItems`** — maps the selected refundable rows to the simplified v4 line items (products → quantity; fees → tax-inclusive `refund_total`).
- **`WooPosRefundSubmissionRequest`** — adds an optional `v4LineItems` (null = v3 path; defaulted, so existing callers are unchanged).
- **`WooPosRefundSubmissionProcessor`** — sends the simplified v4 create when `v4LineItems` is present, otherwise the v3 create.

### Test Steps

No user-facing change (the new units aren't wired into the flow until Part 3). Covered by unit tests: `WooPosRefundPreviewTest`, `WooPosBuildRefundV4LineItemsTest`, `WooPosV4RefundAvailabilityCacheTest`, `WooPosRefundSubmissionProcessorTest`.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosBuildRefundV4LineItems.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosBuildRefundV4LineItems.kt
@@ -10,11 +10,11 @@ class WooPosBuildRefundV4LineItems @Inject constructor() {
         val productLineItems = productItems
             .groupBy { it.orderItemId }
             .map { (orderItemId, rows) ->
-                RefundV4LineItem(lineItemId = orderItemId, quantity = rows.size)
+                RefundV4LineItem.quantityBased(lineItemId = orderItemId, quantity = rows.size)
             }
 
         val feeLineItems = feeItems.map { fee ->
-            RefundV4LineItem(
+            RefundV4LineItem.amountBased(
                 lineItemId = fee.orderItemId,
                 refundTotal = fee.unitPrice + fee.unitTax,
             )
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosBuildRefundV4LineItems.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosBuildRefundV4LineItems.kt
@@ -1,0 +1,25 @@
+package com.woocommerce.android.ui.woopos.orders.details.refund
+
+import org.wordpress.android.fluxc.model.refunds.RefundV4LineItem
+import javax.inject.Inject
+
+class WooPosBuildRefundV4LineItems @Inject constructor() {
+    operator fun invoke(selectedItems: List<WooPosRefundableItem>): List<RefundV4LineItem> {
+        val (feeItems, productItems) = selectedItems.partition { it.isLumpSum }
+
+        val productLineItems = productItems
+            .groupBy { it.orderItemId }
+            .map { (orderItemId, rows) ->
+                RefundV4LineItem(lineItemId = orderItemId, quantity = rows.size)
+            }
+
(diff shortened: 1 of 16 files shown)


iamgabrielma (commented):
Codex review:

Blocker:
> WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt:25: the v4 availability cache is keyed by site.siteId,
    which is the remote WP.com site ID and can be 0 for self-hosted/WPAPI sites. If one such store marks v4 unavailable at line 31 or 44, every other store with siteId
    == 0 will skip v4 preview for the rest of the process. Key this by the local site ID, e.g. site.localId().value or selectedSite.getSelectedSiteId().

Suggestions:

> - WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt:58: avoid response.model!! per project Kotlin
    conventions. A local val model = response.model branch or ?.let keeps this null-safe without force unwrap.

>  - WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreviewTest.kt:32: add coverage for two distinct local sites with
    the same remote siteId/default 0 to lock down the cache fix above.

iamgabrielma (approved):
Pre-approving this to not block the rest of the work before EOD 👀

samiuelson:
Thanks @iamgabrielma — all three addressed in 2ac1da45fc1.

**Blocker (cache key).** Confirmed and fixed. `SiteModel.getSiteId()` is the remote WP.com id and is `0` for self-hosted/WPAPI stores, so the `@Singleton` cache would let one such store's "v4 unavailable" verdict poison every other `siteId == 0` store. The cache now keys by the local site id (`site.localId().value`, unique per stored site) at all call sites — in this use case and in the wire-up ViewModel.

**Force-unwrap.** Removed — `response.model` is captured into a local `val preview` before the `when`, so the success branch is null-safe without `!!`.

**Coverage.** Added a regression test with two sites sharing remote `siteId = 0` but distinct local ids: site A is marked unavailable, then site B still probes v4 and is marked available independently. It fails against the old `siteId` keying and passes with the local-id key.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #16132, opened 2026-06-23
[WOOMOB-3296] Detect rest_invalid_signature and guide the merchant

AdamGrzybkowski:
### Description

Fixes WOOMOB-3296

When Jetpack can't verify a request's signature on the merchant's WordPress site, it returns `HTTP 400` with `{"code":"rest_invalid_signature","message":"The request is not signed correctly.","data":{"status":400}}`. This is a **server-side** problem the app cannot fix (broken Jetpack connection, a security plugin, a recent plugin update, etc.). Today the app just keeps retrying silently — the background sync loop re-runs every ~4h indefinitely — so merchants assume the app is broken and reinstall, which doesn't help.

This PR detects that error and guides the merchant instead of failing silently:

- **Typed error (FluxC).** `rest_invalid_signature` is mapped to a new `WooErrorType.REST_INVALID_SIGNATURE` (+ a shared `WooError.REST_INVALID_SIGNATURE_CODE` constant).
- **Central detection.** `WooNetwork` (the chokepoint every WooCommerce REST call flows through) reports the error via a new optional `InvalidSignatureListener`. The app-side `StoreConnectionErrorMonitor` (`@Singleton`) records the affected store (keyed by remote site id) and clears it automatically once a request to that store succeeds again. `WooNetwork` keeps the listener as an `Optional` so the fluxc-plugin graph stays valid without the app's binding.
- **Stop the silent retry loop.** When the selected store is affected, `UpdateDataOnBackgroundWorker` cancels its unique periodic work and returns `Result.failure()` instead of retrying. It is naturally re-enqueued (`ExistingPeriodicWorkPolicy.KEEP`) the next time the app goes to background, so it recovers on its own once the server-side issue is fixed.
- **Merchant-facing dialog, app-wide.** A custom Compose dialog ("Your store can't be reached") is hosted in `MainActivity` so it overlays **every** main-app screen, not just the dashboard. It stays visible across navigation until the connection is restored. It is modal (can't be dismissed by tapping outside / back) and offers two actions:
  - **Contact support** — opens the in-app support flow
  - **Dismiss** — snoozes the dialog for the current session; it reappears the next time the app returns to the foreground if the store is still unreachable.

### Test Steps

These steps reproduce the server-side error on a [Jurassic Ninja](https://jurassic.ninja/) test site by faking the REST response with a custom mu-plugin.

1. Spin up a Jurassic Ninja site (with Jetpack), and log into the app with that store so it's the selected store.
2. On the JN site, add a mu-plugin that forces the signature error for Jetpack-tunneled requests:
   ```
   cat > ~/htdocs/wp-content/mu-plugins/simulate-invalid-signature.php << 'EOF'
   <?php
   add_filter( 'rest_authentication_errors', function ( $result ) {
       if ( isset( $_GET['_for'] ) && 'jetpack' === $_GET['_for'] ) {
           return new WP_Error(
               'rest_invalid_signature',
               'The request is not signed correctly.',
               array( 'status' => 400 )
           );
       }
       return $result;
   }, 99 );
   EOF
   ```
3. In the app, trigger a store request (e.g. open the Orders tab or pull-to-refresh the dashboard). Verify the **"Your store can't be reached"** dialog appears.
5. Tap **Contact support** — confirm it opens the support request form. 
6. Re-trigger the dialog, tap **Dismiss** — confirm it closes. Send the app to the background and reopen it; with the error still active, confirm the dialog reappears.
7. Background sync: with the error active, send the app to background; confirm the periodic data-sync worker stops retrying (it no longer loops).
8. Recovery: remove the mu-plugin to clear the error:
   ```
   rm ~/htdocs/wp-content/mu-plugins/simulate-invalid-signature.php
   ```
   Pull-to-refresh / re-enter the app; confirm a successful request clears the state and the dialog no longer appears.

### Images/gif

https://github.com/user-attachments/assets/2849ea3d-a023-4072-9c5f-07cb9d847587

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateDataOnBackgroundWorker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateDataOnBackgroundWorker.kt
@@ -3,22 +3,26 @@ package com.woocommerce.android.background
 import android.content.Context
 import androidx.hilt.work.HiltWorker
 import androidx.work.CoroutineWorker
+import androidx.work.WorkManager
 import androidx.work.WorkerParameters
 import com.woocommerce.android.analytics.AnalyticsEvent
 import com.woocommerce.android.analytics.AnalyticsTracker
 import com.woocommerce.android.analytics.AnalyticsTrackerWrapper
+import com.woocommerce.android.network.StoreConnectionErrorMonitor
 import com.woocommerce.android.ui.login.AccountRepository
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 
 @HiltWorker
+@Suppress("LongParameterList")
 class UpdateDataOnBackgroundWorker @AssistedInject constructor(
     @Assisted private val appContext: Context,
     @Assisted workerParams: WorkerParameters,
     private val accountRepository: AccountRepository,
     private val updateAnalyticsDashboardRangeSelections: UpdateAnalyticsDashboardRangeSelections,
     private val updateOrderListBySelectedStore: UpdateOrderListBySelectedStore,
-    private val analyticsTrackerWrapper: AnalyticsTrackerWrapper
+    private val analyticsTrackerWrapper: AnalyticsTrackerWrapper,
+    private val storeConnectionErrorMonitor: StoreConnectionErrorMonitor
 ) : CoroutineWorker(appContext, workerParams) {
     companion object {
         const val REFRESH_TIME = 4L
@@ -39,6 +43,13 @@ class UpdateDataOnBackgroundWorker @AssistedInject constructor(
                     mapOf(AnalyticsTracker.KEY_TIME_TAKEN to (System.currentTimeMillis() - startTime))
                 )
                 Result.success()
+            }
+
+            storeConnectionErrorMonitor.isDetectedForSelectedSite() -> {
+                // The store can't be reached due to a server-side signature problem the app can't fix.
+                // Stop the silent retry loop (it's re-enqueued the next time the app goes to background).
+                WorkManager.getInstance(appContext).cancelUniqueWork(WORK_NAME)
+                Result.failure()
             }
 
             else -> {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateDataOnBackgroundWorker.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateDataOnBackgroundWorker.kt
@@ -46,7 +46,6 @@ class UpdateDataOnBackgroundWorker @AssistedInject constructor(
             }
 
             storeConnectionErrorMonitor.isDetectedForSelectedSite() -> {
-                // The store can't be reached due to a server-side signature problem the app can't fix.
                 // Stop the silent retry loop (it's re-enqueued the next time the app goes to background).
                 WorkManager.getInstance(appContext).cancelUniqueWork(WORK_NAME)
                 Result.failure()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
@@ -96,8 +96,8 @@ class MainActivityViewModel @Inject constructor(
 
     val isUserAgeRangeEligible = ageEligibilityChecker.ageEligibilityState.asLiveData()
 
-    // Snoozes the dialog when the merchant taps Dismiss. Reset when the app returns to the foreground
-    // (see [onAppForegrounded]) so a still-unreachable store reminds the merchant again next session.
+    // Snoozes the dialog when the merchant taps Dismiss. Reset when the app goes to the background
+    // (see [onAppBackgrounded]) so a still-unreachable store reminds the merchant again next session.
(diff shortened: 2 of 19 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/background/UpdateDataOnBackgroundWorker.kt
```diff
@@ -41,6 +45,13 @@ class UpdateDataOnBackgroundWorker @AssistedInject constructor(
                 Result.success()
             }
 
+            storeConnectionErrorMonitor.isDetectedForSelectedSite() -> {
+                // The store can't be reached due to a server-side signature problem the app can't fix.
```
irfano: > // The store can't be reached due to a server-side signature problem the app can't fix.

np: `StoreConnectionErrorMonitor` already contains this comment, so I think we don’t need this line.

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -459,6 +459,11 @@
     <string name="dashboard_scheduled_import_option_selected">Selected</string>
     <string name="dashboard_scheduled_import_update_error">Couldn\'t update the setting. Please try again.</string>
 
+    <!-- Store connection error -->
+    <string name="store_connection_error_dialog_title">Your store can\'t be reached</string>
+    <string name="store_connection_error_dialog_body">We\'re having trouble connecting to your store. This is usually a connection issue on your WordPress site — often caused by a security plugin, a recent plugin update, or a Jetpack connection that needs to be refreshed. The WooCommerce app can\'t fix this from your phone.</string>
```
irfano:
```suggestion
    <string name="store_connection_error_dialog_body">We\'re having trouble connecting to your store. This is usually a connection issue on your WordPress site, often caused by a security plugin, a recent plugin update, or a Jetpack connection that needs to be refreshed. The WooCommerce app can\'t fix this from your phone.</string>
```

I just removed `—`.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
```diff
@@ -88,6 +90,59 @@ class MainActivityViewModel @Inject constructor(
...
+                when {
+                    !isAffected -> {
+                        connectionErrorSnoozed.value = false
+                        _showStoreConnectionErrorDialog.value = false
+                    }
+
+                    snoozed -> _showStoreConnectionErrorDialog.value = false
+
+                    else -> _showStoreConnectionErrorDialog.value = true
+                }
```
irfano:
```
_showStoreConnectionErrorDialog.value = isAffected && !snoozed
if(!isAffected) {
    connectionErrorSnoozed.value = false
}
```

np: Minor coding-style nit. This trims it from 8 lines to 4 and reads a bit simpler to me.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivityViewModel.kt
```diff
@@ -88,6 +90,59 @@ class MainActivityViewModel @Inject constructor(
...
+    }
+
+    fun onStoreConnectionErrorDismissed() {
+        connectionErrorSnoozed.value = true
+    }
+
+    /**
+     * Called when the app returns to the foreground; clears the dialog snooze so a still-unreachable store
+     * reminds the merchant again.
+     */
```
irfano: I’m unsure if this comment is useful. I think the code and function name are already clear.

irfano (changes requested):
I added some np comments.
Also, what do you think about dismissing the dialog with the OS back button? Currently, the OS back button does nothing when the dialog is visible.

> Tap Contact support — confirm it opens the support request form. (If you can inspect the resulting Zendesk ticket, confirm it carries the rest_invalid_signature and origin:connection-error tags.)

I also tested this step but couldn’t see the `rest_invalid_signature` and `origin:connection-error` tags in the created ticket. I guess there’s an issue in this flow. I also tested #16135, but it's the same.

AdamGrzybkowski:
>Also, what do you think about dismissing the dialog with the OS back button? Currently, the OS back button does nothing when the dialog is visible.

Done!

>I also tested this step but couldn’t see the rest_invalid_signature and origin:connection-error tags in the created ticket. I guess there’s an issue in this flow. I also tested https://github.com/woocommerce/woocommerce-android/pull/16135, but it's the same.

Yeah, I thought I removed all the references to that from the PR description... This was done in a follow-up PR https://github.com/woocommerce/woocommerce-android/pull/16135. Did you disable the `ai_support_chat` feature flag? By default, it's ON, and we open the AI chat - that won't add a tag, there should already be plenty of information in the messages.

irfano (changes requested):
> By default, it’s ON, and we open the AI chat - that won’t add a tag

Yes, I tested again, and it works as expected.

I noticed one issue: after dismissing the dialog and changing the orientation, the dialog becomes visible again. We’re good to merge after fixing this.

AdamGrzybkowski:
That's a good catch @irfano - thank you!

irfano (approved):

PR #16135, opened 2026-06-24
[WOOMOB-3357] Pass store-connection error context into the AI support chat

AdamGrzybkowski:
### Description

Fixes WOOMOB-3357

Follow-up to [#16132](https://github.com/woocommerce/woocommerce-android/pull/16132) (WOOMOB-3296), and stacked on top of it.

WOOMOB-3296 added a "Your store can't be reached" dialog shown app-wide when a `rest_invalid_signature` error is detected. Its **Contact support** button opens `HelpActivity` with `HelpOrigin.CONNECTION_ERROR`. The goal is for the resulting support interaction to carry that error context so a Happiness Engineer immediately recognizes the case.

`HelpActivity` has two Contact-Support routes. The **non-AI path** forwards caller tags to the Zendesk ticket. The **AI bot path** is the default (`FeatureFlag.AI_SUPPORT_CHAT` is locally enabled), and it opened the AI chat with no error context at all — so the context was lost on the path most merchants actually hit.

This PR makes the context survive on both paths:

- **AI path (default) — pre-seed the bot, no typing required.** A new `AiSupportChatLaunchMode.StoreConnectionError` (created via `AiSupportChatActivity.createStoreConnectionErrorIntent(...)`, mirroring the existing `ConnectivityTool` pattern). `HelpActivity.showAiSupportChat()` routes `CONNECTION_ERROR` to this intent. On load, `AiSupportChatViewModel.startFromStoreConnectionError()` seeds a failed `STORE_CONNECTION` diagnostic carrying `rest_invalid_signature`, then **auto-sends a short localized first message** so the Odie bot replies immediately with relevant guidance — the merchant doesn't have to describe or re-diagnose anything. A new `STORE_CONNECTION_ERROR` analytics entry point is recorded.
- **Non-AI fallback (AI flag remotely off).** `MainActivity` now passes `extraSupportTags = listOf(WooError.REST_INVALID_SIGNATURE_CODE)` through `startHelpActivity(...)`, so the Zendesk ticket is tagged `rest_invalid_signature` (the existing tag-forwarding plumbing already routes `extraTagsFromExtras` into the support request form). 

No new user-facing surfaces; the one new string (`ai_support_chat_store_connection_error_message`) is the auto-sent message and goes through the normal translation pipeline.

### Test Steps

These steps reproduce the server-side error on a [Jurassic Ninja](https://jurassic.ninja/) test site by faking the REST response with a custom mu-plugin (same setup as #16132).

1. Spin up a Jurassic Ninja site (with Jetpack), and log into the app with that store so it's the selected store.
2. On the JN site, add a mu-plugin that forces the signature error for Jetpack-tunneled requests:
   ```
   cat > ~/htdocs/wp-content/mu-plugins/simulate-invalid-signature.php << 'PHP'
   <?php
   add_filter( 'rest_authentication_errors', function ( $result ) {
       if ( isset( $_GET['_for'] ) && 'jetpack' === $_GET['_for'] ) {
           return new WP_Error(
               'rest_invalid_signature',
               'The request is not signed correctly.',
               array( 'status' => 400 )
           );
       }
       return $result;
   }, 99 );
   PHP
   ```
3. In the app, trigger a store request (e.g. open the Orders tab or pull-to-refresh the dashboard). Verify the **"Your store can't be reached"** dialog appears.
4. **AI path (default):** tap **Contact support**. Confirm the AI support chat opens directly in chat mode and **the bot replies on its own without you typing anything** — a short "my store can't be reached" message is sent automatically and the bot responds with guidance.
5. **Non-AI fallback:** disable the `AI_SUPPORT_CHAT` feature flag (Settings → developer feature-flag override, logged-in build), re-trigger the dialog, tap **Contact support**, and confirm the Zendesk **support request form** opens instead. If you can inspect the resulting ticket, confirm it carries the `rest_invalid_signature` and `origin:connection-error` tags.
6. Recovery: remove the mu-plugin and pull-to-refresh; confirm the dialog no longer appears.
   ```
   rm ~/htdocs/wp-content/mu-plugins/simulate-invalid-signature.php
   ```

### Images/gif

https://github.com/user-attachments/assets/2ed45dc8-194f-4ce5-b42a-a5fe09ba3ff2

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
@@ -30,6 +30,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
 import javax.inject.Inject
 
 @Suppress("LargeClass")
@@ -76,6 +77,7 @@ class AiSupportChatViewModel @Inject constructor(
             is AiSupportChatLaunchMode.Help -> Unit
             is AiSupportChatLaunchMode.PreLogin -> startFromPreLogin()
             is AiSupportChatLaunchMode.ConnectivityTool -> startFromConnectivityTool(launchMode.checks)
+            is AiSupportChatLaunchMode.StoreConnectionError -> startFromStoreConnectionError()
             is AiSupportChatLaunchMode.Resume -> resumeChat(launchMode)
         }
     }
@@ -401,6 +403,29 @@ class AiSupportChatViewModel @Inject constructor(
 
     private fun startFromConnectivityTool(checks: List<ConnectivityCheckCardData>) {
         val result = checks.toDiagnosticResult()
+        _viewState.update {
+            it.copy(
+                input = "",
+                messages = listOf(greetingMessage()),
+                selectedIssueType = SupportIssueType.OTHER,
+                diagnosticResult = result,
+                hasProceededToChat = true,
+                canPersistChatHistory = accountRepository.isUserLoggedIn(),
+                showSendError = false
+            )
+        }
+    }
+
+    private fun startFromStoreConnectionError() {
+        val result = DiagnosticResult(
+            issueType = SupportIssueType.OTHER,
+            statuses = listOf(
+                DiagnosticStatus(
+                    test = DiagnosticTest.STORE_CONNECTION,
+                    status = TestStatus.Failed(technicalDetails = STORE_CONNECTION_ERROR_DETAIL)
+                )
+            )
+        )
         _viewState.update {
             it.copy(
                 input = "",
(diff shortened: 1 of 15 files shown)


irfano (commented):
Everything works well, but I just have one question: Why don’t we send the `rest_invalid_signature` tag in the AI path? It seems like a useful tag even in that scenario.

AdamGrzybkowski:
@irfano Good call, it doesn't hurt to send it. Update the PR - also I force pushed the branch, because the other PR was merge and had to fix the commits.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
irfano, line 461: I created a ticket from a chat I resumed from the chat history screen, but the `rest_invalid_signature` tag isn't sent in this flow. The reason: the tag is only written to `launchExtraTags` when the chat is first opened from the store connection error. The resume flow doesn't restore it, because the tag isn't persisted with the chat. So in a resumed chat `launchExtraTags` stays empty and the tag isn't added on escalation. Can we fix this?

Screenshot_20260626_161302

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aisupportchat/AiSupportChatViewModel.kt
AdamGrzybkowski: Done

irfano: I confirmed it’s now sending the `rest_invalid_signature` tag when resuming chats. Thanks! 👍🏻

irfano (approved):

PR #16138, opened 2026-06-25
[WOOMOB-3232] Fix different shipping address switch losing state while editing

samiuelson:
Fixes WOOMOB-3232

### Description

When editing customer details during order creation, the **"Add a different shipping address"** switch would re-enable itself after the user disabled it and then typed in any input field.

**Root cause:** `OrderCreateEditCustomerAddFragment.observeViewState()` forced the switch back to `isChecked = true` on *every* `viewState` emission whenever the shipping address was non-empty:

```kotlin
if (newShipping?.address != Address.EMPTY) {
    showShippingAddressFormSwitch?.addressSwitch?.isChecked = true
}
```

The shipping address stays populated in `AddressViewModel` until "Done" is tapped, so every keystroke (which emits a new `viewState`) re-triggered this and flipped the switch back on, discarding the user's choice.

**Fix:** The ViewModel is now the single source of truth for the switch state:

- `AddressViewModel.initialize()` sets `isDifferentShippingAddressChecked` once, based on whether the initial shipping address differs from billing (a non-empty shipping address ⇒ checked).
- The fragment no longer derives the switch state from each `viewState` emission. Instead it observes `isDifferentShippingAddressChecked` and updates the switch's checked state (with an equality guard to avoid a listener feedback loop) alongside the shipping form visibility.

The user's toggle now persists while typing and across view recreation (e.g. returning from the country/state selector), and is still correctly applied when "Done" is tapped.

### Test Steps

1. Open **Orders** → **Create order**.
2. Tap **Add customer details** and select a customer with **different billing and shipping addresses**.
3. Open the customer details editor — confirm **"Add a different shipping address"** is on and the shipping form is shown.
4. Turn **off** "Add a different shipping address" — the shipping form hides.
5. Type into any billing field (e.g. City).
6. Verify the switch **stays off** and the shipping form stays hidden.
7. Tap **Done** and reopen the editor — verify the state was applied as expected.

### Images/gif

N/A


malinajirka (approved):
Thanks @samiuelson !

I've noticed one behavior which wasn't expected, but I guess it's fine.


1. Open Orders → Create order.
2. **Select product and recalculate order** 
3. Tap Add customer details and select a customer with different billing and shipping addresses.
4. Open the customer details editor — confirm "Add a different shipping address" is on and the shipping form is shown. Turn off "Add a different shipping address" — the shipping form hides.
5. Type into any billing field (e.g. City).
6. Verify the switch stays off and the shipping form stays hidden.
7. Tap Done and reopen the editor — **Notice the toggle is switched ON** and the billing address is filled out.

samiuelson:
> I've noticed one behavior which wasn't expected, but I guess it's fine.
> Open Orders → Create order.
> Select product and recalculate order
> Tap Add customer details and select a customer with different billing and shipping addresses.
> Open the customer details editor — confirm "Add a different shipping address" is on and the shipping form is shown. Turn off "Add a different shipping address" — the shipping form hides.
> Type into any billing field (e.g. City).
> Verify the switch stays off and the shipping form stays hidden.
> Tap Done and reopen the editor — Notice the toggle is switched ON and the billing address is filled out.

Fixed in 07e3b2c060f0614b6c486713a26fe96541a87cd4

Thanks for review!

PR #16142, opened 2026-06-25
[WOOMOB-3224] Fix duplicate product slug handling

hichamboushaba:
### Description
Fixes WOOMOB-3224

Duplicating a product copied the source product slug into the draft duplicate. Because the duplicate is created through the REST create path, sending the original slug can leave the draft with the same slug as the source product, which makes storefront preview/open-product behavior resolve to the wrong product.

This clears the duplicate-only `slug` and `permalink` values before creating the copied product. The source product remains unchanged, and the server can assign an appropriate slug for the duplicate instead of receiving the source product slug.

### Test Steps
1. Open a store with at least one published product that has a storefront permalink.
2. Duplicate that product from the product details screen.
3. Confirm the duplicated product opens as a draft copy.
4. Reopen the original product and use the storefront preview/open action.
5. Confirm the original storefront product opens, not the duplicate copy.

### Images/gif
N/A


JorgeMucientes (approved):
Changes LGTM and work as expected :shipit:

PR #16143, opened 2026-06-25
Make remote feature flag value take precedence over local

irfano:
### Description

Today the local value takes precedence over remote:

```
effectiveValue = overrideValue ?: (localValue && (remoteValue ?: true))
```

Because `localValue` is an AND gate, a known remote value can only turn a feature off, never on: if `localValue` is false the feature stays off whatever the server says. An unknown remote value (fetch pending or failed) is also treated as `true`, so a flag with `localValue = true` shows up during the first-launch window before the remote value lands. There is also no way to ship a feature off in the binary and enable it from the server later.

This PR flips the precedence so a known remote value wins:

```
effectiveValue = overrideValue ?: (remoteValue ?: localValue)
```

The remote value now decides whenever it is known (it can enable or disable), and `localValue` is used only as the fallback until a remote value is available. For flags with `localValue = true` the result is mathematically unchanged.

It also changes three recently-added, remote-controlled flags (`WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1`, `SMARTER_NOTIFICATIONS`, `QR_LOGIN`) from `localValue = true` to `localValue = PackageUtils.isDebugBuild()`. Now that remote is authoritative these features no longer have to default on, so they default off in release (on in debug) and are turned on through their remote values. For `WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1` this also fixes the first-launch window where it could become wrongly visible before the remote value loaded. None of this changes the resolved value in practice: these flags are already driven by remote, so once the remote value loads the result is the same as before, and the only difference is the brief pre-fetch window, which now defaults to off.

Background and discussion: 

### Test Steps

1. On a debug build, open **Settings → Developer options → Feature Flags**.
2. The screen lists every flag with its resolved value. On a debug build this change does not alter any existing flag, so the page should look the same as on `trunk`.
3. Compare a few flags against a `trunk` build, including `WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1`, `SMARTER_NOTIFICATIONS` and `QR_LOGIN`: the resolved values should match.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/util/FeatureFlag.kt
@@ -4,9 +4,10 @@ package com.woocommerce.android.util
  * Feature flags control feature availability.
  *
  * - If overridden (debug builds) → use override value
- * - If [localValue] is false → the feature stays disabled
- * - If [localValue] is true and remote has a value for [remoteFlagKey] → use remote value
- * - Otherwise → use [localValue]
+ * - If remote has a value for [remoteFlagKey] → use remote value
+ * - Otherwise → fall back to [localValue]
+ *
+ * Remote is authoritative once known. [localValue] is the fallback whenever no remote value is available.
  *
  * Access via [FeatureFlagRepository.isEnabled].
  */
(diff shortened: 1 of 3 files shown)


AdamGrzybkowski (approved):
LGTM 👍

PR #16150, opened 2026-06-27
[WOOMOB-998] Make ProductsMapper#mapProductIdsToProduct suspendable

irfano:
Fixes WOOMOB-998

### Description

`ProductsMapper.mapProductIdsToProduct` (and its private `asProductList` helper) were non-suspend and used `runBlocking` to call the suspend `WCProductStore` DB functions. Both callers in `ProductSelectorViewModel` (`loadRecentProducts`, `loadPopularProducts`) are already suspend and run on `viewModelScope`, so that `runBlocking` was blocking the main thread while Room queried the database, with no benefit.

This PR makes the function suspend and removes the `runBlocking`, so the DB access suspends instead of blocking the main thread. No call sites changed, since the callers were already suspend.

This is part of WOOMOB-983. This one was small because the callers were already suspend; the other sub-issues of WOOMOB-983 also need changes on the caller side, so they involve larger changes.

### Test Steps

No user-visible change; this is an internal threading fix. On a store that has at least one paid order containing products:

1. Open the **Orders** tab.
2. Start creating a new order.
3. Tap **Add products** to open the product selector.
4. Confirm the **Popular** and **Last Sold** sections show the expected products.

### Images/gif

N/A, no UI change.

- [x] I have considered if this change warrants release notes. This is an internal refactor with no behavior change, so none were added.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/selector/ProductsMapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/selector/ProductsMapper.kt
@@ -3,7 +3,6 @@ package com.woocommerce.android.ui.products.selector
 import com.woocommerce.android.model.Product
 import com.woocommerce.android.model.toAppModel
 import com.woocommerce.android.tools.SelectedSite
-import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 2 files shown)


hichamboushaba (approved):
Thanks @irfano for handling this

PR #16152, opened 2026-06-28
[WOOMOB-2212] Remove deprecated log tag migration

irfano:
### Description
Fixes WOOMOB-2212

Removes the temporary `NOTIFS` to `NOTIFICATIONS` log tag migration from `LogEntry`.

The migration was only needed for old local log entries. Since app logs are retained for up to seven days, those old entries should no longer be present. With the temporary migration removed, old `NOTIFS` entries are treated like other invalid log entries and skipped by the current log parser.

### Test Steps
1. Log in to the app.
2. Go to Help.
3. Open Application log.
4. Confirm the current log file opens and valid log entries are displayed.

### Images/gif
N/A


JorgeMucientes (approved):
LGTM 👍🏼

PR #16154, opened 2026-06-28
[WOOMOB-984] Make user eligibility fetcher suspendable

irfano:
### Description
Fixes WOOMOB-984

This PR removes `runBlocking` from the user eligibility error flow. The cached user lookup is now suspendable end to end, and the error screen loads that user data from the ViewModel coroutine scope.

No user-facing behavior is expected to change.

### Test Steps
1. Set a test site's user role to an unsupported role, such as Subscriber.
2. Log in to the app with that user and select the site.
3. Confirm the role access error screen opens and shows the user name and role.
4. Tap Retry and confirm the screen remains on the role access error state while the role is still unsupported.

### Images/gif
Screenshot_20260629_013918


hichamboushaba (approved):
Changes look good.

PR #16155, opened 2026-06-28
[WOOMOB-2011] Sanitize application password device names

irfano:
### Description
Fixes WOOMOB-2011

Device names can contain path separators, but the app password application name is also used to build an encrypted SharedPreferences filename. This PR sanitizes the device name before using it in the application name, while preserving the existing space-to-hyphen format.

### Test Steps
N/A - covered by unit tests. Manual verification is difficult because it depends on the device model name.

### Images/gif
N/A, no UI change.


hichamboushaba (approved):
Thanks for fixing this @irfano

PR #16156, opened 2026-06-29
[WOOMOB-3273] Suppress back press while card payment is being captured

malinajirka:
### Description

Fixes WOOMOB-3273

Makes the IPP payment flow non-cancellable once it reaches the payment-in-progress (capturing) state, so Back no longer exits while the payment is completing; the earlier present-card phase stays cancellable.

### Test Steps

Requires a real card reader (the capturing state isn't reachable on an emulator).

1. Create an order, add a product, and collect payment with a card reader.
2. Tap a test card and let the flow reach the **"Capturing payment"** screen.
3. Press **Back** there → it's ignored; you stay until capture completes.
4. On the earlier **"present card"** screen, press **Back** → still cancels, as before.

### Images/gif

N/A


toupper (approved):
LGTM and tests well

PR #16159, opened 2026-06-29
Add success state to WooPos primary buttons

rubenscastro:
### Description
This PR introduces a reusable success state for WooPosButton.

As a first use case, the Checkout button now briefly transitions to the success state after an item is added to the cart, providing immediate visual feedback that the action was successful. This improves user feedback by acknowledging that adding a product to the cart was successful, reducing uncertainty after the tap.

### Test Steps
1. Launch Woo POS.
2. Add a product to the cart.
3. Verify the Checkout button briefly transitions to the success state.
4. Verify the button returns to its default enabled state after the animation.
5. Verify existing button states (enabled, disabled and loading) continue to behave as expected.

### Images/gif
Tablet:
https://github.com/user-attachments/assets/d8623e31-a211-47f4-80d5-755cb92f14a3

Phone:
https://github.com/user-attachments/assets/0424b294-d696-4e13-a043-06854a7dc26b

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosButtons.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosButtons.kt
@@ -4,6 +4,13 @@ package com.woocommerce.android.ui.woopos.common.composeui.component
 
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
+import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.slideInVertically
+import androidx.compose.animation.slideOutVertically
+import androidx.compose.animation.togetherWith
+import androidx.compose.runtime.getValue
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -46,6 +53,7 @@ import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosCom
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosCornerRadius
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosIconSize
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosSpacing
+import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosIcons
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTheme
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.WooPosTypography
 import com.woocommerce.android.ui.woopos.common.composeui.designsystem.currentWooPosBreakpoint
@@ -79,7 +87,7 @@ fun WooPosButton(
         textStyle = WooPosTypography.BodyLarge,
         text = text,
         state = state,
-        colors = ButtonDefaults.buttonColors(
+        baseColors = ButtonDefaults.buttonColors(
             containerColor = MaterialTheme.colorScheme.primary,
             contentColor = MaterialTheme.colorScheme.onPrimary,
             disabledContainerColor = WooPosTheme.colors.disabledContainer,
@@ -103,7 +111,7 @@ fun WooPosButtonSmall(
         textStyle = WooPosTypography.BodySmall,
         text = text,
         state = state,
-        colors = ButtonDefaults.buttonColors(
+        baseColors = ButtonDefaults.buttonColors(
             containerColor = MaterialTheme.colorScheme.primary,
             contentColor = MaterialTheme.colorScheme.onPrimary,
             disabledContainerColor = WooPosTheme.colors.disabledContainer,
@@ -122,7 +130,7 @@ fun WooPosOutlinedButton(
     textStyle: WooPosTypography = WooPosTypography.BodyLarge,
     onClick: () -> Unit,
 ) {
-    val borderColor = if (state == WooPosButtonState.ENABLED || state == WooPosButtonState.LOADING) {
+    val borderColor = if (state != WooPosButtonState.DISABLED) {
         MaterialTheme.colorScheme.inverseSurface
     } else {
         WooPosTheme.colors.disabledContainer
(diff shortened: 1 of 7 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/phone/WooPosHomePhonePersistentButton.kt
```diff
@@ -51,7 +52,8 @@ class WooPosPhonePersistentButtonStateResolver(
                     if (body.amountOfItems > 0) {
                         WooPosPhonePersistentButtonState.Primary(
                             label = buildCartLabel(body.amountOfItems),
-                            buttonState = WooPosButtonState.ENABLED,
+                            buttonState = cartState.checkoutButtonState.toWooPosButtonState()
+                                ?: WooPosButtonState.ENABLED,
```
malinajirka, line 56: 🚫 🐛  This switches the Products-screen OpenCart button from always-`ENABLED` to inheriting `checkoutButtonState`, so it now goes `DISABLED` whenever the cart holds a loading/error item (`cartContainsLoadingOrErrorItems`).

The problem is that OpenCart is pure navigation, and on phone this button is the *only* path from Products → Cart (`PhoneOpenCartClicked`) — adding an item doesn't auto-navigate. Error items can only be removed from inside the cart. So if a barcode scan adds an `Error` item while you're on Products, the button goes dead and you can't open the cart to remove it; adding a valid item afterwards doesn't help because the error item keeps the state `Disabled`. Before this PR the button was hardcoded `ENABLED`, so you could always get in.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/composeui/component/WooPosButtons.kt
```diff
@@ -263,23 +271,35 @@ private fun Button(
...
     state: WooPosButtonState = WooPosButtonState.ENABLED,
     maxLines: Int = Int.MAX_VALUE,
     onClick: () -> Unit,
 ) {
-    val onClickLocal = if (state == WooPosButtonState.ENABLED) {
-        onClick
-    } else {
-        {}
-    }
+    val isSuccess = state == WooPosButtonState.SUCCESS
```
malinajirka, line 280: 🔍 nitpick: This feels a bit unnecessary and inconsistent. Considering, we use the full `state == WooPosButtonState....` for enabled and loading, I'd suggest removing the `isSuccess` variable and using `state == WooPosButtonState.SUCCESS` directly. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartScreen.kt
```diff
@@ -170,8 +169,9 @@ fun WooPosCartScreen(
 
         when (checkoutSlot) {
             WooPosCartCheckoutButtonSlot.Inline -> {
+                val checkoutButtonState = state.checkoutButtonState.toWooPosButtonState()
                 AnimatedVisibility(
-                    visible = state.checkoutButtonState != WooPosCartState.CheckoutButtonState.Invisible,
+                    visible = checkoutButtonState != null,
```
malinajirka, line 174: ❓ While it's not an issue to treat `null` as `invisible`, in general we try to avoid treating `null` as a valid value. What was the reasoning behind removing the explicit `Invisible` state and replacing it with `null`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/cart/WooPosCartViewModel.kt
```diff
@@ -93,8 +95,10 @@ class WooPosCartViewModel @Inject constructor(
         .map { updateStateDependingOnCartStatus(it) }
 
     private val itemNumberProvider = AtomicInteger(getInitialValueOrHighestUsedItemNumberAfterProcessDeath())
+    private var itemAddedFlashJob: Job? = null
 
     init {
+        _state.value = _state.value.copy(isItemJustAdded = false)
```
malinajirka, line 101: 📓 This is a good example of a workaround that needs to be implemented, because a fire-and-forget event is designed as state. This code ensures, that after a process death, the animation doesn't play again. Not a big deal, I'm sharing it just to better convey the idea and potential risks.

malinajirka (changes requested):
Thanks @rubenscastro! Awesome job on your first Woo Android PR 👏 !

The approach you opted for is valid overall, having said that, I'd have personally opted for a slightly different one. Let me try to describe it and elaborate on the reasons why it might be a bit safer.

Instead of changing the state of the button, I'd just draw a success animation over it. The main benefit we'd get is that we wouldn't be touching the state of the button at all, ultimately reducing risks of breaking something as well as the maintenance cost. You can check the approach I have in mind [here](https://github.com/woocommerce/woocommerce-android/compare/pos-cart-add-success-flash?expand=1) (not necessarily a production ready code, I just wanted to show the approach).

The main thing that throws me off a bit in the approach used in this PR is that `isItemJustAdded/Success state` is a very temporary state - it's technically not a state, but a fire-and-forget event. Integrating it as part of the `CheckoutButtonState` feels off to me and it can introduce some tricky situations.

If you compare it to `Loading`, it feels different. Loading might be a valid state for 0.5s but also 30s => it's a state which should survive configuration change (e.g. screen rotation) => its point is to disable the button and show progress to the user (has impact on behavior). However, `isItemJustAdded/Success state` is an event that starts an animation - the event might be emitted 10 times within a second if we add 10 items to the cart, its goal is not to change the behavior (state) of the checkout button.

One example of a tricky situation we might be introducing, is the bug mentioned in the comments below - on phone POS, we might disable the `Open Cart` button (which is Checkout button in disguise) upon adding an invalid item to the cart , essentially preventing the user from opening the cart (they can't even remove the invalid item). This is a bug we can fix - but it kind of nicely shows risks the approach might introduce.

It's fair to say, the `animation on top of the button` approach has some downsides too. E.g. We can't easily slide-down and slide-up the `Checkout` text as in your animation.. We also can't easily cancel the animation mid-flight..... and likely more. Having said that, I think it might still be a more pragmatic approach overall.

Having said all that, I don't mind merging the approach used in this PR, since in the end, it's partially a matter of preference. The only must-have is fixing the `Open Cart` issue on phones which is a new bug introduced in this PR.


https://github.com/user-attachments/assets/ca474dab-ccab-4519-aa86-daac2967ff1a

P.S. I simulated addition of an invalid item by connecting an external keyboard to my phone and typing a random barcode value as quickly as I could (e.g. "asdfg").

rubenscastro:
Closing this as this solution is not the direction we decided to go.

PR #16160, opened 2026-06-29
[WOOMOB-3245] Respect site week start in stats ranges

hichamboushaba:
### Description
Fixes WOOMOB-3245

The Android stats range builders were using the device/default calendar when generating `This Week` ranges, so stores configured to start weeks on Monday could still show Sunday-based ranges on devices whose default week starts on Sunday.

This PR fetches and caches the site's WordPress `start_of_week` setting from `/wp/v2/settings`, stores it as `DayOfWeek`, and routes My Store/dashboard and Analytics Hub range generation through a site-aware `CalendarHelper`. The helper falls back to the default calendar when the selected site or cached setting is unavailable. The calendar lookup is suspend-based so stats range selection does not block the main thread while reading the cache.

The change covers the shared dashboard stats path, Analytics Hub, Monday/Sunday week starts, custom ranges, and cache/error behavior in FluxC.

### Test Steps
1. Use a store whose WordPress week starts on Monday and a device/locale whose default calendar starts weeks on Sunday.
2. Open My Store stats and select `This Week`.
3. Confirm the displayed stats range starts on Monday and ends on Sunday.
4. Open Analytics Hub and select `This Week`.
5. Confirm Analytics Hub uses the same Monday-through-Sunday range.
6. Select a custom date range and confirm the selected start/end dates are preserved.
7. Use or switch to a store whose WordPress week starts on Sunday and confirm `This Week` still starts on Sunday.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>FetchWPSiteSettingsPayload</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/settings/WPSettingsRestClient.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.generated.endpoint.WPAPI
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooNetwork
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * This REST client intentionally lives in fluxc-plugin for now so it can reuse [WooNetwork], including the
```
AdamGrzybkowski, line 12: ❓ "for now" - Where would it land eventually?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WPSettingsStore.kt
```diff
@@ -0,0 +1,56 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class WPSettingsStore @Inject constructor(
+    private val restClient: WPSettingsRestClient,
+    private val wpSiteSettingsDao: WPSiteSettingsDao
+) {
+    private val startOfWeekConverter = WPStartOfWeekConverter()
+
+    suspend fun fetchSiteSettings(site: SiteModel): FetchWPSiteSettingsPayload {
```
AdamGrzybkowski, line 23: ❓ Curious, but why do we use `Payload` instead of `WooResult`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/analytics/hub/AnalyticsHubViewModel.kt
```diff
@@ -218,10 +221,12 @@ class AnalyticsHubViewModel @Inject constructor(
     }
 
     fun onCustomRangeSelected(startDate: Date, endDate: Date) {
-        rangeSelectionState.value = SelectionType.CUSTOM.generateLocalizedSelectionData(
-            startDate = startDate,
-            endDate = endDate
-        )
+        viewModelScope.launch {
```
AdamGrzybkowski, line 224: ❓ Any reasons for using `launch` and `viewModelScope.launch` interchangeably?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/common/RefreshWPSettings.kt
```diff
@@ -0,0 +1,23 @@
...
+    suspend operator fun invoke(site: SiteModel) {
+        if (canFetchWPSettings()) {
+            wpSettingsStore.fetchSiteSettings(site)
+        }
+    }
+
+    private fun canFetchWPSettings(): Boolean {
+        val roles = userEligibilityFetcher.getUser()?.roles ?: return true
+        // /wp/v2/settings requires manage_options, which Shop Managers do not have.
+        return UserRole.Administrator in roles || UserRole.Owner in roles
```
AdamGrzybkowski, line 21: Does it mean `shop managers` will never see correct time ranges?

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
```diff
@@ -206,6 +209,7 @@ class AppInitializer @Inject constructor() : ApplicationLifecycleListener {
                     }
                     wooCommerceStore.fetchSiteGeneralSettings(site)
                     wooCommerceStore.fetchSiteProductSettings(site)
+                    refreshWPSettings(site)
```
AdamGrzybkowski, line 212: Shouldn't we call this when switching sites as well?

AdamGrzybkowski (commented):
Code looks good but I have two questions that I would like to clarify before approving (those without ❓ ).

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/settings/WPSettingsRestClient.kt
hichamboushaba: On `fluxc` module, because it's a WP Core endpoint, and uses a table on `WPDatabase`, but moving it would require extracting some logic from `WooNetwork` first.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WPSettingsStore.kt
hichamboushaba: Related to https://github.com/woocommerce/woocommerce-android/pull/16160#discussion_r3498977281

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/analytics/hub/AnalyticsHubViewModel.kt
hichamboushaba: No reason, but they are the same, our `ScopedViewModel` is a CoroutineScope itself, and delegates to the androidx `viewModelScope`.
If it's bothering you, I can align them.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/common/RefreshWPSettings.kt
hichamboushaba: Yes if they use a locale that has a different `start of week` day than what the site is configured for 🤷‍♂️ 

Fixing this would require a Core change (either in WP or in Woo as a new endpoint).

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
hichamboushaba: Yes, I thought about adding it on `SiteObserver`, but given this is `Settings`, and we are already not doing this for the other settings, I decided to put it at the same place, and if we want to fix it, we should do it for all three.

WDYT?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/common/RefreshWPSettings.kt
hichamboushaba: Just a clarification on the impact: "this week" will mean a different range than what the site shows; the stats themselves will be correct for the range the app calculates.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/analytics/hub/AnalyticsHubViewModel.kt
AdamGrzybkowski: Not bothering - just wanted to make sure there's no hidden reason behind it!

WooCommerce/src/main/kotlin/com/woocommerce/android/AppInitializer.kt
AdamGrzybkowski: Makes sense to do for all for consistency!

AdamGrzybkowski (approved):

PR #16162, opened 2026-06-29
[WOOMOB-3258] Fix account mismatch retry login route

hichamboushaba:
### Description
Fixes WOOMOB-3258

When a merchant reaches the account-mismatch error screen and taps **Log in with another account**, the app was starting the default login flow again. That sent the merchant back to the store URL entry step even though the app still had the site-login context.

This updates the account-mismatch retry path to start the existing WP.com email-login action without a known email, and teaches that action to open the WP.com email entry screen directly when no email extra is provided. The retry path also stops clearing the saved login site address so the merchant stays in the current site-login context. Focused ViewModel tests cover the retry navigation for both logged-in and already-logged-out states.

### Test Steps
1. Install the Wasabi debug app.
2. Start the site-login flow from a logged-out state.
3. Enter a valid WooCommerce site URL.
4. Authenticate with a WordPress.com account that is not associated with that site.
5. On the account-mismatch error screen, tap **Log in with another account**.
6. Verify the app opens the WordPress.com email/account entry screen directly instead of returning to the store URL entry screen.

### Images/gif
N/A


AdamGrzybkowski (commented):
>Verify the app opens the WordPress.com email/account entry screen directly instead of returning to the store URL entry screen.

That works, but when I tap the back button in the toolbar the app closes. I wonder if we could have a proper backstack and open the site URL page instead?

hichamboushaba:
> That works, but when I tap the back button in the toolbar the app closes. I wonder if we could have a proper backstack and open the site URL page instead?

Thanks @AdamGrzybkowski, yes, I'm aware of this, but having a proper backstack there would require a bigger refactoring, given the LoginActivity uses fragment transactions only, and even if we put the fragments on the backstack, when returning to them, the fields will be empty.
Do you feel this is a blocker? IMHO it's still a good experience, because if the user re-opens the app, then can enter a different site URL, but if they want to keep the same site, they can continue.

AdamGrzybkowski:
[My main worry is that this is not something that users might expect.](https://developer.android.com/guide/navigation/principles#the_up_button_never_exits_your_app)

>when returning to them, the fields will be empty.

The URL is saved, so we could restore it, no? I imagine this would not be a simple fix, but maybe now it's time to make this right.

hichamboushaba:
@AdamGrzybkowski I updated the approach here. Instead of opening the email screen directly, we now start from the site address screen with the current site prefilled. The existing site discovery flow then runs and decides the next step.

That addresses the back-stack issue you called out, and it should also be more robust: if the site’s Jetpack connection is no longer valid when the merchant taps Log in with another account, discovery can route them to the site credentials screen instead of forcing the [WordPress.com](http://wordpress.com/) email flow.


https://github.com/user-attachments/assets/c6a5eef1-7f2a-43e6-9309-5f3c747efd52

Let me know what you think.

AdamGrzybkowski:
@hichamboushaba This is slightly better, but we still end up with the "up" button that closes the app on the site URL screen. What if we modify the `ShowSiteAddress` to:

```kotlin
         is LoginActivityActionDestination.ShowSiteAddress -> {
                        // Add the prologue as the backstack root so Back returns to the initial
                        // login screen instead of exiting the app.
                        changeFragment(LoginPrologueFragment(), shouldAddToBackStack = true, LoginPrologueFragment.TAG)
                        loginViaSiteAddress(prefilledSiteUrl = destination.siteAddress)
                        // The prologue re-enables edge-to-edge in onViewCreated, so run the pending
                        // transactions first, then restore the default insets for the (non
                        // edge-to-edge) site address screen that ends up on top.
                        supportFragmentManager.executePendingTransactions()
                        disableDynamicEdgeToEdge()
}
```

The above adds the `LoginPrologueFragment` and now the "up" button correctly navigates back:


https://github.com/user-attachments/assets/7a10c417-8b07-4103-bdab-330a20d27481

hichamboushaba:
Thanks @AdamGrzybkowski, what you suggest works, but IMO it's not scalable, the issue touches all `LoginActivitiy` actions and deeplinks, and IMO it's a minor issue, so I wouldn't try to hack my way there. If we have time, we can aim for a global solution, but it's outside of the scope of the current PR.

AdamGrzybkowski (approved):
Makes sense 👍 :shipit:

PR #16166, opened 2026-06-30
Fix stale Woo push registration after token refresh

irfano:
### Description

This hotfix fixes an Android push notification blackout path after an FCM token change.

When a site is successfully registered in the Woo self-serve push notification system, WP.com notifications are disabled for that site to avoid duplicate notifications. If the FCM token later changes, the local Woo push registration can become stale. If the app then falls back to WP.com delivery, the site can remain with WP.com notifications disabled and stop receiving push notifications.

This PR:
- Clears stale Woo push registrations when the stored Woo token no longer matches the current FCM token.
- Re-enables WP.com notification settings after WP.com registration succeeds, but only for stale visible sites that are not currently Woo-registered.
- Keeps WP.com notifications disabled for sites that successfully return to the Woo push registration path, avoiding duplicate notifications.

### Test Steps

1. Use a Jetpack-connected test store that is registered through Woo self-serve push notifications.
2. Confirm order push notifications are received.
3. Disable the Woo self-serve push notification mobile feature flag.
4. Force an FCM token change and relaunch the app. To do this:

Apply this patch:
[Add_debug_broadcast_receiver_for_FCM_token_refresh.patch](https://github.com/user-attachments/files/29525899/Add_debug_broadcast_receiver_for_FCM_token_refresh.patch)

Run this command:
```
adb shell am broadcast \
  -n com.woocommerce.android.dev/com.woocommerce.android.notifications.push.DebugRefreshFcmTokenReceiver \
  -a com.woocommerce.android.dev.DEBUG_REFRESH_FCM_TOKEN
```

6. Confirm the app registers the new token with WP.com.
7. Create a new order and confirm the push notification is received.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -226,6 +226,33 @@ class PushNotificationRepository @Inject constructor(
             registration.locale != getDeviceLocale()
     }
 
+    suspend fun clearWooPushRegistrationsForStaleToken(currentToken: String) {
+        if (currentToken.isEmpty()) return
+
+        var staleSiteIds = emptyList<Long>()
+
+        pushNotificationsDataStore.edit { preferences ->
+            staleSiteIds = preferences.getSiteIdsWithStalePushRegistration(currentToken)
+            staleSiteIds.forEach { siteId ->
+                preferences.clearPushRegistration(siteId)
+            }
+        }
+        staleSiteIds.forEach { siteId ->
+            WooLog.d(WooLog.T.NOTIFICATIONS, "Cleared stale Woo Core push registration for site $siteId")
+        }
+    }
+
+    private fun Preferences.getSiteIdsWithStalePushRegistration(currentToken: String): List<Long> = asMap().keys
+        .mapNotNull { key ->
+            key.name
+                .takeIf { it.startsWith(PUSH_TOKEN_KEY_PREFIX) }
+                ?.removePrefix(PUSH_TOKEN_KEY_PREFIX)
+                ?.toLongOrNull()
+        }
+        .filter { siteId ->
+            this[getPushTokenValueKeyForSite(siteId)] != currentToken
+        }
+
     fun isWpComPushRegistered(): Boolean =
         prefsWrapper.getFluxCPreferences()
             .getString(WpComPushNotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
@@ -242,13 +242,7 @@ class PushNotificationRepository @Inject constructor(
         }
     }
 
-    private fun Preferences.getSiteIdsWithStalePushRegistration(currentToken: String): List<Long> = asMap().keys
-        .mapNotNull { key ->
-            key.name
-                .takeIf { it.startsWith(PUSH_TOKEN_KEY_PREFIX) }
-                ?.removePrefix(PUSH_TOKEN_KEY_PREFIX)
-                ?.toLongOrNull()
-        }
+    private fun Preferences.getSiteIdsWithStalePushRegistration(currentToken: String): List<Long> = registeredSiteIds()
         .filter { siteId ->
             this[getPushTokenValueKeyForSite(siteId)] != currentToken
         }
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/297">25.0.1</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
Thanks @irfano for the work on this. I asked AI to draw a diagram for the flow of the new changes, and below is what it produced:

```mermaid
flowchart TD
    A["RegisterDevice runs with current FCM token"] --> B["clearWooPushRegistrationsForStaleToken(token)"]
    B --> C["Remove local Woo registrations where stored token != current token"]
    C --> D["Return stale site IDs"]

    D --> E{"Woo self-driven push flag enabled?"}
    E -- yes --> F["Try Woo registration for visible sites"]
    F --> G["Successful Woo registrations save current token and disable WP.com"]
    E -- no --> H["Skip Woo path"]
    G --> I["Evaluate WP.com registration"]
    H --> I

    I --> J{"WP.com registration succeeds?"}
    J -- no --> K["Do not change WP.com settings"]
    J -- yes --> L["Find stale visible sites not currently Woo-registered"]
    L --> M["Enable WP.com notifications for those sites"]
```

I think this still leaves one case unsolved: the cleanup only happens when the stored Woo token differs from the current FCM token. So if we disable the feature flag without an FCM token refresh, the site can stay registered in Woo, and WP.com notifications can remain disabled.

We might also still get duplicate notifications if, for some reason, the `/me/notifications/settings` call failed when we registered the site with Woo.

What do you think about a different approach?

```mermaid
flowchart TD
    A["RegisterDevice starts"] --> B{"Woo feature flag enabled?"}

    B -- yes --> C["Use Woo path as today"]
    C --> D["Successful Woo registration disables WP.com settings"]

    B -- no --> E["Find locally Woo-registered site IDs"]
    E --> F["Ensure current FCM token is registered with WP.com"]
    F --> G{"WP.com registration succeeds?"}
    G -- no --> H["Do not unregister Woo / do not clear local state"]
    G -- yes --> I["Re-enable WP.com notifications for those sites"]
    I --> J{"WP.com settings update succeeds?"}
    J -- no --> H
    J -- yes --> K["Unregister those sites from Woo PN"]
    K --> L["Clear local Woo registration state"]
```

The idea is simple: when the feature flag is disabled, we do the cleanup path. For each Woo-registered site, we first make sure WP.com PNs are enabled again, then unregister from Woo PN and clear the local Woo registration state.

irfano:
Thanks @hichamboushaba. I closed this PR for now since we decided not to include this in the hotfix. I’ll either retarget it to the beta branch or open a new PR.

Your suggestion makes sense for beta, and I’ll update the implementation in that direction.

I intentionally didn’t include the proactive migration in the hotfix because there’s a group of users who don’t have Jetpack connected but do have a valid Woo Core push token. Even though the Woo-driven FF is disabled now, they can still receive notifications from Woo Core as long as their token is valid.

If we added the proactive migration in 25.0.1, we’d unregister those users from Woo Core and move them back to WP.com. For stores without Jetpack, that means they would stop receiving notifications.

For 25.1 this should be fine, because by then we expect the FF to be enabled again and rolled out to 100%.

hichamboushaba:
Thanks @irfano 

> If we added the proactive migration in 25.0.1, we’d unregister those users from Woo Core and move them back to WP.com. For stores without Jetpack, that means they would stop receiving notifications.

I think that's still the correct thing to do if we decide to disable the feature flag, because it might be because of something wrong (for example a security issue or something similar...).

irfano:
> I think that's still the correct thing to do if we decide to disable the feature flag, because it might be because of something wrong (for example a security issue or something similar...).

I agree. This time it happened to be safe to keep using Woo Core notifications because the issue wasn't on the Woo Core delivery side, but we shouldn't rely on that when the FF is disabled in the future.

I'll apply your suggestion in 25.1.

irfano:
@hichamboushaba, I took another look at your suggestion and realized you had already covered my concern: in your flow, if WP.com registration (or the settings update) fails, we don't unregister from Woo. So my objection about non-Jetpack users didn't actually apply to your proposal, those users would simply stay on Woo Core.

That said, as you suggested in your [follow-up](https://github.com/woocommerce/woocommerce-android/pull/16166#issuecomment-4856860943), I'll diverge from your chart on this point: when the FF is disabled, I'll unregister from Woo Core unconditionally, even if the WP.com steps fail. If we ever disable the flag, it might be for a safety reason, so we shouldn't keep devices registered with Woo Core.

hichamboushaba:
> That said, as you suggested in your https://github.com/woocommerce/woocommerce-android/pull/16166#issuecomment-4856860943, I'll diverge from your chart on this point: when the FF is disabled, I'll unregister from Woo Core unconditionally, even if the WP.com steps fail. If we ever disable the flag, it might be for a safety reason, so we shouldn't keep devices registered with Woo Core.

Thanks @irfano, I see, there is one concern with this though: given the Woo registration state is how we know we should re-enable WPCom notifications, if we remove it unconditionally, then if WPCom call fails, those uses will be stuck.
What about splitting the decision depending on the site:
- For Jetpack users: unregister from Woo only if WPCom call succeeds, even if fails on current launch, we'll retry on future launches.
- For other users (site credentials and Jetpack CP): unregister unconditionally.

WDYT?

irfano:
Thanks @hichamboushaba, I misread your security point earlier, I agree with the split. The new PR implements it as you suggested, ready for review: [#16183](https://github.com/woocommerce/woocommerce-android/pull/16183)

PR #16175, opened 2026-07-01
[WOOMOB-3391] Revert dashboard widget + More menu CIAB gating

AdamGrzybkowski:
### Description

Fixes WOOMOB-3391

This PR removes those gates so the dashboard cards and Inbox behave as they do on non-CIAB sites (available, subject to their normal non-CIAB conditions).

`CIABSiteGateKeeper` itself stays — it's still used by other features (payments, Blaze, products, POS, order-status editing, …) and gets torn down last in WOOMOB-3386. This PR only stops the four dashboard/More-menu callers from using it.

**Stacked PR:** based on `issue/WOOMOB-3385_remove-bookings` (WOOMOB-3385). Merge that one first; this branch shares the CIAB enum edit with it, so basing on trunk would conflict.

### Test Steps

1. On a non-CIAB store, open **My Store** and confirm the Onboarding, Inbox, and Stock dashboard cards appear as before (behavior unchanged).
2. Open **More menu** and confirm the Inbox entry shows when the store's Woo core version is `>= 6.4.0`.
3. On a CIAB store (if available), confirm the same cards/Inbox are now shown rather than hidden.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/ObserveInboxWidgetStatus.kt
```diff
@@ -1,27 +1,15 @@
...
     operator fun invoke() = selectedSite.observe()
         .filterNotNull()
-        .flatMapLatest {
-            if (ciabSiteGateKeeper.isFeatureUnsupported(CIABAffectedFeature.Inbox)) {
-                flowOf(DashboardWidget.Status.Hidden)
-            } else {
-                flowOf(DashboardWidget.Status.Available)
-            }
-        }
+        .map { DashboardWidget.Status.Available }
```
irfano: This class no longer does anything: it just maps to `Available`, and it didn't exist before the CIAB gating. Can we delete it and let `INBOX` fall into the `else -> Available` branch in `DashboardRepository`?

irfano (approved):
I have minor feedback, but pre-approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/data/ObserveInboxWidgetStatus.kt
AdamGrzybkowski: Good call, removed.

PR #16179, opened 2026-07-01
[WOOMOB-3392] Revert Blaze / Plugins / split-shipments CIAB gating

AdamGrzybkowski:
### Description

Fixes WOOMOB-3392

This PR removes three gates so these features behave on all sites the way they already do on **non-CIAB** sites. 

Changes — each drops the `CIABAffectedFeature` gate and keeps the non-CIAB path:

- **`IsBlazeEnabled.kt`** — dropped the `CIABAffectedFeature.Blaze` clause and the now-unused `ciabSiteGateKeeper` dependency. Blaze eligibility is still gated by `isAdmin`, a valid Jetpack connection, and `canBlaze`.
- **`MainSettingsPresenter.kt`** — `isPluginsSectionVisible` is now always `true`; removed the unused `ciabSiteGateKeeper` dependency.
- **`WooShippingLabelCreationViewModel.kt`** — removed the split-shipments gate. The `isSplitShipmentsSupported` field was introduced solely by the CIAB gating commit, so this reverts to the exact pre-CIAB shape: the field and param are removed and `shouldShowSplitShipmentButton` no longer checks it (keeping the later multi-item refactor intact).
- **`CIABAffectedFeature.kt`** — removed the three now-unused enum entries (`Blaze`, `Plugins`, `WooShippingSplitShipments`). They fell through the `else` branch in `CIABSiteGateKeeper`, so no gatekeeper logic changed.
- Tests updated to drop the CIAB-specific cases/mocks.

### Test Steps

Confirm no behavior change:
1. Blaze entry points still appear for an admin site with a valid Jetpack connection and Blaze capability.
2. Settings still shows the **Plugins** section.
3. In shipping label creation for an order with multiple items, the **split shipment** option still appears.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/MainSettingsPresenter.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/MainSettingsPresenter.kt
@@ -126,9 +126,6 @@ class MainSettingsPresenter @Inject constructor(
     override val isThemePickerOptionVisible: Boolean
         get() = selectedSite.get().isWPComAtomic
 
-    override val isPluginsSectionVisible: Boolean
-        get() = true
-
     override fun setupEnablePushNotificationsOption() {
         coroutineScope.launch {
             shouldShowEnablePushNotificationsUi().collect { shouldShowOption ->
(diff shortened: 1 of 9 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/MainSettingsPresenter.kt
```diff
@@ -130,7 +127,7 @@ class MainSettingsPresenter @Inject constructor(
         get() = selectedSite.get().isWPComAtomic
 
     override val isPluginsSectionVisible: Boolean
```
irfano: `isPluginsSectionVisible` was introduced with CIAB project. Since it's now always `true`, can we just remove `isPluginsSectionVisible`?

irfano (approved):
I have minor feedback, but I’m pre-approving. LGTM! 👍🏻

PR #16180, opened 2026-07-02
[WOOMOB-3387] Revert In-Person Payments CIAB gating

AdamGrzybkowski:
### Description

Fixes WOOMOB-3387

Part of the CIAB ("Commerce In A Box") retirement on Android. In-Person Payments call-sites were previously gated behind `CIABSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.InPersonPayments)`, which hid/blocked IPP on non-Pro CIAB sites. This PR unwinds that gating so IPP behaves as it always has on non-CIAB sites, and removes the `InPersonPayments` feature entirely.

What changed:
- **CardReaderPaymentCollectibilityChecker** — dropped the gate; card payments are collectible on the normal path.
- **TapToPayAvailabilityStatus** — removed the `Result.Hidden` branch and the `Hidden` result type; TTP availability is determined by device/country only.
- **SelectPaymentMethod (ViewModel/ViewState/Fragment)** — full revert of the "learn more IPP" footer to its pre-CIAB form (single `LearnMoreIpp` data class), removing the CIAB plan-upgrade "learn more" dialog branch and the post-upgrade `onResumed` refresh.
- **AppShortcutsHandler** — the Payments launcher shortcut is always added.
- **MoreMenuViewModel** — the "Payments" more-menu item is no longer gated; reverted to always visible.
- **CIABAffectedFeature / CIABSiteGateKeeper** — removed the `InPersonPayments` enum value and its dedicated gatekeeper branch (plus the now-orphaned `isCurrentSiteProCIAB()`). `CIAB_PRO_PLAN_SLUGS` and `buildPlanUpgradeUrl()` are kept — still used by POS eligibility.
- Tests updated to drop CIAB cases and assert the unconditional (available) behavior.

### Test Steps

On a standard (non-CIAB) store, behavior should be unchanged:

1. Open an unpaid order → **Collect Payment** is available and can be started with a card reader / Tap to Pay (subject to the usual device/country support).
2. On the payment method selection screen, confirm the "Learn more about In-Person Payments" footer link is shown and opens the learn-more web page.
3. Long-press the app launcher icon → the **Payments** shortcut is present.
4. Open **More menu** → the **Payments** entry is visible.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsViewModelTest.kt
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsViewModelTest.kt
@@ -179,7 +179,7 @@ class WooPosTotalsViewModelTest {
     private val featureFlagRepository: FeatureFlagRepository = mock()
     private val builtInReaderConnector: WooPosBuiltInReaderConnector = mock()
     private val tapToPayAvailabilityStatus: TapToPayAvailabilityStatus = mock {
-        on { invoke() } doReturn TapToPayAvailabilityStatus.Result.Hidden
+        on { invoke() } doReturn TapToPayAvailabilityStatus.Result.Available
     }
     private val remoteReaderSession: WooPosRemoteReaderSession = mock {
         on { state }.thenReturn(MutableStateFlow(WooPosRemoteReaderSession.State.Idle))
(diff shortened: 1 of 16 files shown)


AdamGrzybkowski:
@malinajirka Tagging you here as this one touches the IPP

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsViewModelTest.kt
```diff
@@ -2314,10 +2314,10 @@ class WooPosTotalsViewModelTest {
         }
 
     @Test
-    fun `given flag on and TTP Hidden, when ViewModel created, then reason not tracked`() = runTest {
+    fun `given flag on and TTP Available, when ViewModel created, then reason not tracked`() = runTest {
```
malinajirka: 🔍 np: This test is now identical to `given flag on and TTP Available, when ViewModel created, then NotAvailable reason not tracked` above — same stubs, same verification. I'd just delete this one instead of renaming it.

malinajirka (approved):
LGTM!

P.S. I'm adding `Do not merge` label as I assume we don't want to merge until the target is `trunk`.

PR #16181, opened 2026-07-02
[WOOMOB-3390] Revert order-status CIAB gating

AdamGrzybkowski:
### Description

Fixes WOOMOB-3390

Sub-task of the CIAB ("Commerce In A Box") retirement on Android. On CIAB sites, order statuses are currently gated in two ways: the full status list is collapsed into a single "Open" status, statuses are shown read-only (non-editable), and status filters are grouped/remapped. This PR reverts that gating so **order statuses behave the same on every site** — full editable status list, no "Open" collapsing — and removes the now-unused order-status CIAB machinery entirely.

**Stacked PR:** this targets `issue/WOOMOB-3387-revert-ipp-gating` (the IPP-gating revert), not `trunk`. It should merge after that base. It is otherwise independent and safe to ship on its own.

What changed:

- **Editing gate removed (always editable).** `OrderDetailFragment`, `OrderCreateEditFormFragment`, and `OrderListFragment` no longer branch on `CIABAffectedFeature.OrderStatusEditing`; the status view is always `Mode.OrderEdit` and multi-select is always enabled. The `ciabSiteGateKeeper` field was dropped from these fragments (it was used only for this gate).
- **Status-mapper routing removed.** Every `CIABOrderStatusMapper` transform call (`mapOrderStatus`, `mapOrderStatusOptionsList`, `mapFilterOptions`, `resolveFilterKeys`) was inlined to use the input unchanged, and the injected `ciabOrderStatusMapper` was removed from `OrderListViewModel`, `OrderDetailViewModel`, `OrderCreateEditViewModel`, `DashboardOrdersViewModel`, `GetOrderStatusFilterOptions`, `GetWCOrderListDescriptorWithFilters`, and `GetWCOrderListDescriptorWithFiltersBySiteId`.
- **`CIABOrderStatusMapper` deleted.** With all transform callers inlined, the class and its remaining static `OPEN_KEY` references became dead. Removed the class + its test, the three now-dead "Open"-status display branches (`OrderStatusTag`, `AiAssistantOrderCardRenderer`, and the `DashboardOrdersViewModel` status-color helper — a custom "open" status now falls through to the default color like any other custom status), and the now-unused `ciab_order_status_open` source string. (Translated copies are GlotPress-managed and cleaned up by the translation sync.)
- **Enum cleanup.** Removed the now-unused `OrderStatusEditing` value from `CIABAffectedFeature` (mirrors the base branch removing `InPersonPayments`).
- **Tests.** Dropped the `ciabOrderStatusMapper` mock wiring from the four affected ViewModel tests (they already stubbed it as identity, so behavior is unchanged) and removed the CIAB "open"-status color test from `AiAssistantOrderCardRendererTest`. Kept the `ciabSiteGateKeeper` mock in `OrderDetailViewModelTest` (still used for the unrelated "is CIAB -> fetch fulfillments" logic).

### Test Steps

1. Open the **Order list** — the status filter shows the full set of statuses (Processing, Pending, On hold, Completed, Failed, etc.), not a single "Open" status.
2. Long-press an order in the list — multi-select is available.
3. Open an **order's details** — tap the order status; the status is editable (opens the status picker) rather than read-only, and shows the real underlying status.
4. **Create/edit an order** — in edit mode the order status is editable.
5. Open the **dashboard Orders card** — order rows show their real statuses/colors and the status filter uses the full status list.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailFragment.kt
```diff
@@ -241,12 +236,7 @@ class OrderDetailFragment :
         setupResultHandlers(viewModel)
         setupOrdersCommunicationObservers(communicationViewModel)
 
-        val statusMode = if (ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.OrderStatusEditing)) {
-            Mode.OrderEdit
-        } else {
-            Mode.ReadOnly
```
irfano, line 247: `Mode.ReadOnly` was added in 41d09b30238 only for this CIAB gate, and this PR removes its last consumers. After this change nothing uses `ReadOnly`, so the value and its `when` branch in `OrderDetailOrderStatusView` are dead, and with a single remaining value the `mode` param itself becomes meaningless. Can we remove the mechanism too?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/aiassistant/AiAssistantOrderCardRenderer.kt
```diff
@@ -85,10 +84,6 @@ private fun resolveOrderStatusColor(status: String): Int =
         is Order.Status.Completed -> R.color.tag_bg_completed
         is Order.Status.Failed -> R.color.tag_bg_failed
         is Order.Status.OnHold -> R.color.tag_bg_on_hold
-        is Order.Status.Custom -> if (status == CIABOrderStatusMapper.OPEN_KEY) {
-            R.color.tag_bg_processing
-        } else {
-            R.color.tag_bg_other
-        }
+        is Order.Status.Custom -> R.color.tag_bg_other
```
irfano: This arm is now identical to the `else` below it. It was introduced CIAB-aware in f1e125e8376, and this PR already folds the equivalent arm into `else` in `DashboardOrdersViewModel`. Can we do the same here?

irfano (changes requested):

irfano (approved):
LGTM! 👍🏻 Thanks for the changes.

PR #16183, opened 2026-07-02
Migrate push registrations back to WPCom when the Woo-driven system is disabled

irfano:
### Description

When the `woo_self_driven_push_notifications_m1` feature flag is disabled remotely, devices that already registered with the Woo Core push system keep their local Woo registration. That state suppresses the WP.com fallback (the app drops incoming WP.com notifications for Woo-registered sites, and WP.com settings for those sites stay disabled), so those users can end up without notifications, permanently after an FCM token rotation. This PR adds recovery logic for both paths, following the approach [suggested by Hicham](https://github.com/woocommerce/woocommerce-android/pull/16166#issuecomment-4865272019) (supersedes #16166):

**Flag on:** before re-registering, local Woo registrations whose stored token no longer matches the current FCM token are cleared, so a rotated token can't leave stale state behind (mirrors the iOS behavior of clearing registered site IDs on token change).

**Flag off:** Woo push registrations are migrated back to WP.com:
- Sites with a working WP.com fallback (visible + full Jetpack connection) are re-enabled on WP.com first, and unregistered from Woo Core only if that succeeds. On failure they stay on Woo push and the migration retries on the next run.
- Sites without a fallback (site credentials, Jetpack CP, hidden sites) are unregistered unconditionally.

```mermaid
flowchart TD
    A["RegisterDevice runs with current FCM token"] --> B{"Woo self-driven push flag on?"}

    B -- yes --> C["Clear local Woo registrations whose stored token != current FCM token"]
    C --> D["Woo registration for sites, as today"]
    D --> E["Successful registrations save current token and disable WP.com notifications"]

    B -- no --> G{"Trigger is SITE_SWITCH?"}
    G -- yes --> H["Skip migration"]
    G -- no --> I["Find locally Woo-registered site IDs"]
    I --> J{"Any Woo-registered sites?"}
    J -- no --> K["Nothing to migrate"]
    J -- yes --> L["Determine WP.com fallback sites: visible + full Jetpack, with WP.com account"]
    L --> M["Ensure WP.com device is registered"]
    M --> N["Re-enable WP.com notifications for fallback sites"]
    N --> O{"WP.com took over successfully?"}
    O -- yes --> P["Unregister ALL Woo-registered sites from Woo Core"]
    O -- no --> Q["Unregister only sites without a WP.com fallback"]
    P --> R["Local state cleared per site only after server delete succeeds"]
    Q --> R
```

Also adds `wpcom_device_enable_push_notifications_success/error` events, mirroring the existing disable events.

### Test Steps

Use a debug build with a Jetpack-connected store on Woo Core **10.9.2**. Until 10.9.2 is released, either install a nightly/dev build of the plugin or temporarily lower `PUSH_NOTIFICATIONS_MIN_WC_VERSION` to `10.9.1`; on older versions the app treats even successfully registered sites as unregistered, which skews the results. The flag can be toggled from Developer Options → Feature Flags. To force an FCM token rotation, apply [this debug receiver patch](https://github.com/user-attachments/files/29525899/Add_debug_broadcast_receiver_for_FCM_token_refresh.patch) and use the adb broadcast from #16166.

**Migration (flag off):**
1. With the flag on, register for Woo push notifications and confirm an order notification arrives.
2. Disable the flag and foreground the app.
3. Verify logs: `Migrating Woo push registrations back to WP.com...` → `WPCom notifications enabled for sites...` → `Woo Core push token deleted for site...`.
4. Create an order and confirm a single notification arrives (via WP.com, no duplicates).
5. Foreground again and verify the migration does not re-run.

**Token rotation while the flag is off (the original blackout scenario):**
1. With the flag on, register for Woo push notifications and confirm an order notification arrives.
2. Disable the flag, then force an FCM token rotation using the debug receiver broadcast.
3. Verify the migration logs from the previous scenario run.
4. Create an order and confirm the notification arrives. Since the old token is dead after rotation, Woo Core can no longer deliver, so an arriving notification proves WP.com took over.

**Stale token cleanup (flag on):**
1. With the flag on and the store registered, force an FCM token rotation.
2. Verify logs: `Cleared stale Woo Core push registration for site...` followed by re-registration with the new token.
3. Create an order and confirm the notification still arrives.

**Site-credentials store (flag off):** repeat the migration steps with a site-credentials login; verify no WP.com calls are made and the Woo token is still deleted.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/293">25.1 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Messages
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt
```diff
@@ -246,15 +286,33 @@ class PushNotificationRepository @Inject constructor(
...
+        }
+        staleSiteIds.forEach { siteId ->
+            WooLog.d(WooLog.T.NOTIFICATIONS, "Cleared stale Woo Core push registration for site $siteId")
+        }
     }
 
+    private fun Preferences.registeredSiteIds(): Set<Long> = asMap().keys
+        .mapNotNull { key ->
+            key.name
+                .takeIf { it.startsWith(PUSH_TOKEN_KEY_PREFIX) && !it.startsWith(PUSH_TOKEN_VALUE_KEY_PREFIX) }
```
irfano, line 311: No behavior change here. The old parsing in `getWooPushRegisteredSiteIds` already excluded `push_token_value_*` keys, but only as a side effect: `removePrefix("push_token_")` left "value_123", which fails `toLongOrNull()` and gets dropped. This condition states that rule explicitly instead of relying on the leftover string not parsing as a number. Same site IDs come out either way.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -82,6 +84,8 @@ class RegisterDevice @Inject constructor(
         val shouldForce = trigger == Trigger.TOKEN_REFRESH
 
         if (featureFlagRepository.isEnabled(FeatureFlag.WOO_SELF_DRIVEN_PUSH_NOTIFICATIONS_M1)) {
+            pushNotificationRepository.clearWooPushRegistrationsForStaleToken(token)
```
hichamboushaba: ❓ isn't the existing logic in https://github.com/woocommerce/woocommerce-android/blob/0119ac66c849f9caf617f7cfc4a03e00fe5d8063/WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/PushNotificationRepository.kt#L265 enough here? The token staleness was part of the refactor I did when I introduced `RegisterDevice`, and I'm trying to understand why we need to remove the registration state instead of what `shouldRegisterWooPushForSite` does.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
```diff
@@ -123,6 +129,47 @@ class RegisterDevice @Inject constructor(
...
+
+        val wooRegisteredSiteIds = pushNotificationRepository.getWooPushRegisteredSiteIds()
+        if (wooRegisteredSiteIds.isEmpty()) return
+
+        WooLog.d(
+            WooLog.T.NOTIFICATIONS,
+            "Migrating Woo push registrations back to WP.com for sites $wooRegisteredSiteIds"
+        )
+
+        val wpComFallbackSiteIds = if (accountStore.hasAccessToken()) {
```
hichamboushaba: np, I would call this just `jetpackConnectedSiteIds`, understanding what `wpComFallback` meant took me a while.

hichamboushaba (commented):
Thanks @irfano, code looks good, there is just one question to clarify before approving.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
irfano: `shouldRegisterWooPushForSite` decides whether to retry the registration, but it doesn't change the registration state between attempts. Until a retry succeeds, the site still counts as Woo-registered. If the failure is persistent, this can last forever, and two problems remain during that time:

1. The status keeps saying "registered". For app-password and Jetpack CP sites this also keeps the "Enable push notifications" entry points hidden, so the user can't recover from within the app, the only fix is clearing app data. And Woo push is the only channel these sites have.
2. `NotificationMessageHandler` keeps dropping incoming WP.com notifications for the site. The Woo registration holds the dead token, so if WP.com is actually sending, e.g. when the disable call failed at registration time, those notifications never show up.

Clearing on token mismatch makes the local state match reality: the site isn't registered until a registration with the current token succeeds. It doesn't replace the staleness check, re-registration is still driven by `shouldRegisterWooPushForSite` for non-forced triggers and locale changes.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
irfano: Good point. I renamed it in c6b99da. I went with `visibleJetpackSiteIds` since the set only contains the visible ones, not all Jetpack-connected sites.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
hichamboushaba: Thanks, that makes sense. My remaining concern is mostly with the flow here:

```kotlin
clearWooPushRegistrationsForStaleToken(token)
shouldRegisterWooPushForSite(token, site.siteId)
```

`shouldRegisterWooPushForSite` already checks `registration.token != currentToken`, but after the up-front clear, stale-token cases become `registration == null` instead. Could we keep the stale-token cleanup closer to the per-site registration decision?

Something like:

```kotlin
val shouldRegisterSite = shouldForce ||
    pushNotificationRepository.shouldRegisterWooPushForSite(token, site.siteId)

if (shouldRegisterSite) {
    pushNotificationRepository.clearWooPushRegistrationForStaleToken(site.siteId, token)
    pushNotificationRepository.registerPushTokenInWooCoreSystem(...)
}
```

The reason I find this easier to reason about is that the same per-site flow owns both decisions: first decide whether this site needs registration, then clear stale local state only as part of that registration path. That makes the token-mismatch case explicit without having a separate pre-pass that changes what `shouldRegisterWooPushForSite` sees. Not blocking.

hichamboushaba (approved):
Thanks @irfano, pre-approving, I left one last comment on the previous thread, but it's not a blocker.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/RegisterDevice.kt
irfano: Agreed, moved the cleanup into the per-site path in eca6fdb.

PR #16186, opened 2026-07-03
[WOOMOB-3389] Revert product-type CIAB gating

AdamGrzybkowski:
### Description

Fixes WOOMOB-3389

Part of the CIAB ("Commerce In A Box") retirement on Android. Before this change, several product types were hidden on CIAB sites via per-type `CIABAffectedFeature` gates. This PR removes that gating so product types behave the same everywhere — as they already do on non-CIAB sites (all types visible), which is the pre-CIAB behavior we're restoring.

**Stacked PR:** targets `issue/woomob-3390-revert-order-status-gating` (the order-status revert), not `trunk`. Review/merge that one first.

Changes:
- `ProductTypeBottomSheetBuilder` — drop the CIAB gates on the **Variable** and **Grouped** "add product" options (now always visible). Subscription / Variable-subscription remain gated by `isEligibleForSubscriptions` (subscription eligibility, not CIAB).
- `ProductFilterListViewModel` — `ProductType.isVisible` now returns `true` for Simple/External/Grouped/Variable/Subscription/Variable-subscription/Bundle/Composite (Variation/Other stay hidden), dropping the `ciabSiteGateKeeper` checks.
- Removed the now-unused `CIABSiteGateKeeper` constructor dependency and imports from both classes.
- Tests updated to drop the CIAB cases; added always-visible assertions for Variable & Grouped in the bottom-sheet test.

`CIABSiteGateKeeper` / `CIABAffectedFeature` stay in the codebase — they're removed in the final teardown (WOOMOB-3386), which merges last.

### Test Steps

1. On a store, tap to add a new product and open the product-type bottom sheet.
2. Confirm all product types are offered — including **Variable** and **Grouped** (plus Simple/Virtual/External; Subscription types appear when the site is subscription-eligible).
3. Go to the Products list → Filters → Product type.
4. Confirm the filter lists Simple, Grouped, External, Variable, and the Bundle/Composite/Subscription explore options — none are hidden.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterListViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterListViewModel.kt
@@ -9,8 +9,6 @@ import com.woocommerce.android.R
 import com.woocommerce.android.analytics.AnalyticsEvent
 import com.woocommerce.android.analytics.AnalyticsTracker
 import com.woocommerce.android.analytics.AnalyticsTrackerWrapper
-import com.woocommerce.android.ciab.CIABAffectedFeature
-import com.woocommerce.android.ciab.CIABSiteGateKeeper
 import com.woocommerce.android.model.PluginUrls
 import com.woocommerce.android.model.ProductCategory
 import com.woocommerce.android.model.WooPlugin
@@ -55,7 +53,6 @@ class ProductFilterListViewModel @Inject constructor(
     private val pluginRepository: PluginRepository,
     private val selectedSite: SelectedSite,
     private val analyticsTracker: AnalyticsTrackerWrapper,
-    private val ciabSiteGateKeeper: CIABSiteGateKeeper,
 ) : ScopedViewModel(savedState) {
     companion object {
         private const val KEY_PRODUCT_FILTER_OPTIONS = "key_product_filter_options"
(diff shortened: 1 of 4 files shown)


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterListViewModel.kt
```diff
@@ -577,16 +574,14 @@ class ProductFilterListViewModel @Inject constructor(
...
+            ProductType.EXTERNAL,
+            ProductType.GROUPED,
+            ProductType.VARIABLE,
+            ProductType.SUBSCRIPTION,
+            ProductType.VARIABLE_SUBSCRIPTION,
+            ProductType.BUNDLE,
+            ProductType.COMPOSITE -> true
 
-            ProductType.BUNDLE -> ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.BundleProducts)
-            ProductType.COMPOSITE -> ciabSiteGateKeeper.isFeatureSupported(CIABAffectedFeature.CompositeProducts)
```
irfano, line 589: After this PR, `POS` is the only `CIABAffectedFeature` entry still in use: the rest lose their last usages here, and `GiftCardEditing` was never used anyway. Can we remove everything except `POS`?

irfano (approved):
I added minor feedback, but I’m pre-approving. 👍🏻

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterListViewModel.kt
AdamGrzybkowski: I have a final teardown PR that cleans this completely - https://github.com/woocommerce/woocommerce-android/pull/16205

PR #16187, opened 2026-07-03
[WOOMOB-3388] Revert POS eligibility CIAB gating

AdamGrzybkowski:
Fixes WOOMOB-3388

### Description

Part of the CIAB ("Commerce In A Box") retirement on Android. This sub-task reverts the POS-specific CIAB / Pro-plan eligibility gating so POS eligibility behaves as it did before CIAB on all sites — there is no longer a Pro-plan upgrade gate, and POS launchability/visibility depends only on the normal checks (WooCommerce version, screen size, country).

What changed:

- `WooPosCanBeLaunchedInTab` — removed the CIAB free-plan check; POS is always launchable subject to the version check. Dropped the now-unused `CiabPlanUpgradeRequired` non-launchability reason.
- `WooPosTabShouldBeVisible` — removed the CIAB-unsupported early return (and its `CIABSiteGateKeeper` dependency).
- `WooPosEligibilityViewModel` / `WooPosEligibilityScreen` — removed the `CiabPlanUpgradeRequired` state, the learn-more/upgrade button, and the resume-recheck round-trip that existed only for the upgrade flow.
- `WooPosSettingsStoreViewModel` — the "edit receipt" deep link always uses the standard `wc-settings` path (dropped the CIAB `next-admin` path).
- `WooPosAnalyticsEvent` — removed the `ciab_plan_upgrade_required` analytics reason mapping.
- Removed three now-dead POS eligibility CIAB strings from `values/strings.xml`.
- Updated the POS eligibility/tab/settings unit tests to drop CIAB cases and assert the standard behavior.

### Test Steps

1. Open the app and confirm the POS tab is visible (on a supported tablet/screen size and country) and POS launches normally.
2. In POS settings, tap "Edit receipt" and confirm it opens the standard `wp-admin/admin.php?page=wc-settings&tab=point-of-sale` URL.
3. Eligibility screen (only reached when POS can't launch — e.g. WooCommerce core older than 9.6.0):
   verify via the `WooPosEligibilityScreenRetryablePreview` Compose preview that the screen offers
   only "Retry"/"Exit POS" and that the former CIAB "Pro plan required" / "Learn More" variant is gone
   (its preview has been removed).

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/eligibility/WooPosEligibilityScreen.kt
@@ -9,7 +9,6 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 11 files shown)


AdamGrzybkowski:
Hi @malinajirka - adding you here again as this is fully POS related.

malinajirka (approved):
LGTM, thanks!

PR #16189, opened 2026-07-03
[WOOMOB-3228] Fix guest order search

JorgeMucientes:
### Description
Fixes WOOMOB-3228

Order list rows show the localized Guest label when an order has no billing name, but WooCommerce does not store that label on guest orders, so searching for it returns no results.

This PR keeps text search untouched and instead adds an explicit escape hatch: when a search for the displayed Guest label comes back empty (and no customer filter is active), the search empty view offers a **Show guest orders** button that re-submits the search as a guest-customer filter (`customer=0`). Compared to silently rewriting the query, this keeps a customer actually named "Guest" searchable, avoids surprising result changes while typing, and keeps the descriptor use case free of presentation strings.

It also switches the order-list descriptor builder to `copy(...)` so active status/date/product/customer/sales-channel filters are preserved when searching (they were previously dropped).

### Test Steps
1. Open Orders on a store with at least one guest order and at least one named-customer order.
2. Search for `guest` — the empty view appears with a **Show guest orders** button.
3. Tap the button and confirm guest orders are listed.
4. Search for a named customer and confirm named-customer results still work.
5. Apply an order filter (e.g. status), search again, and confirm the filter still applies.
6. With a customer filter active, search for `guest` and confirm the regular "no results" empty view is shown (no guest button).

### Images/gif

| Before  | After |
| <video width="300" controls src="https://github.com/user-attachments/assets/c9f11c6b-fd88-4dfd-9996-2eacce97c8c8"></video>  | <video width="300" controls src="https://github.com/user-attachments/assets/8a7b05b0-0a2c-4f80-a196-3964eabe4a13"></video>  |

_Note: the After video shows the previous approach (automatic query rewrite). The current implementation shows the search empty view with a "Show guest orders" button instead — updated recording to follow._


AdamGrzybkowski (approved):
Works as described! :shipit:

AdamGrzybkowski:
Actually, there's one thing: the guest search doesn't survive the config change:


https://github.com/user-attachments/assets/6a20a5eb-b8de-47da-a99a-3e1f7ede040e

Could we fix this?

JorgeMucientes:

PR #16192, opened 2026-07-03
[Design System] Add buttons and display components

hichamboushaba:
> [!NOTE]
> This is the first split PR for Store Design System components. It builds on `store-design-system/components0` and targets that branch, so review the diff against that base rather than `trunk`.

### Description

This PR introduces the revised first previewable Store Design System component slice in `:libs:store-design-system`: Material 3-aligned buttons, icon actions, display/status components, and library-local catalog preview infrastructure.

#### What this adds

- Production Compose button components under `com.woocommerce.android.ui.compose.designsystem.component`:
  - `WooFilledButton`
  - `WooFilledTonalButton`
  - `WooOutlinedButton`
  - medium and small sizing through `WooButtonSize`
- Production icon action components:
  - `WooIconButton`
  - `WooOutlinedIconButton`
  - neutral and primary emphasis through `WooIconButtonEmphasis`
- Production display/status components:
  - `WooBadge` and `WooBadgeTone`
  - `WooNoticeBanner` and `WooNoticeBannerTone`
  - `WooIconContainer` and `WooIconContainerTone`
  - `WooSectionHeader`
  - `WooDivider` and `WooVerticalDivider`
  - `WooLinearProgressIndicator` and `WooCircularProgressIndicator`
- Component-local Android Studio preview/demo coverage for the new action and display/status components.
- Library preview/catalog infrastructure under `designsystem.preview`:
  - `WooDesignSystemPreviewTheme`
  - a reduced `WooDesignSystemComponentCatalogPreview` scoped to the production components in this PR.
- Demo/navigation icons use existing Store Design System icon accessors and resources from `components0`; this PR adds no drawable resources.

#### Why this shape

The split keeps the first component PR independently previewable while giving reviewers a coherent starter set: action primitives plus passive display/status primitives. It lands the first reusable production component surface and a catalog preview shell without pulling in later input, navigation, or app-side runtime wiring.

- The catalog preview is the review surface for this slice: it renders the real PR1 action and display/status demos, rather than placeholder rows or references to components from later PRs.
- The module boundary from `components0` is preserved: this branch stays in `:libs:store-design-system` and does not depend on app `R`, Hilt, `FeatureFlag`, POS, or Store feature packages.
- The commit stack separates action components, catalog preview infrastructure, and display/status components so follow-up stacked branches can build from a clear base.

#### What this does not do

- Does not migrate or adopt the components in any product screen.
- Does not add Developer Options runtime wiring.
- Does not include cells, settings rows, switches, search, choice controls, tabs, page headers, top app bars, XML toolbar bridge, or preview-only explorations.
- Does not touch POS.
- Does not globally remap XML/View styles or legacy app themes.
- Does not add generated token, icon, Code Connect, or Figma automation.

#### Helpful review focus

- Are the button and icon-action APIs clear enough for migrated Store Management screens?
- Are the display/status component APIs scoped and named clearly enough for migrated Store Management screens?
- Do the components stay aligned with the foundations tokens and avoid app/POS dependencies?
- Does the reduced catalog preview clearly exercise only the production components available in this PR?
- Are the reused DS icons acceptable for the notice-banner demo states?

No release notes: this is an internal design-system components change.

### Test Steps

1. In Android Studio, render `WooDesignSystemComponentCatalogPreview` from `:libs:store-design-system`.
2. Verify the catalog preview contains only the PR1 production slices: buttons, icon actions, and display/status components.
3. Render the light and dark preview variants and verify the actual component demos appear.
4. Optionally render the component-local previews in the added action and display/status component files.

### Images/gif

| Light | Dark |
| --- | --- |
| components1-catalog-preview-light | components1-catalog-preview-dark |

### DIFF
--- a/libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooBadge.kt
+++ b/libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooBadge.kt
@@ -1,0 +1,138 @@
+package com.woocommerce.android.ui.compose.designsystem.component
+
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Icon
+import androidx.compose.material3.LocalContentColor
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.dp
+import com.woocommerce.android.ui.compose.designsystem.WooTheme
+import com.woocommerce.android.ui.compose.designsystem.foundation.WooDesignSystemTheme
+import com.woocommerce.android.ui.compose.designsystem.icons.Star
+import com.woocommerce.android.ui.compose.designsystem.icons.WooDsIcons
+
+@Composable
+fun WooBadge(
+    text: String,
+    modifier: Modifier = Modifier,
+    tone: WooBadgeTone = WooBadgeTone.Neutral,
+    leadingIcon: (@Composable () -> Unit)? = null,
+) {
+    val colors = tone.toBadgeColors()
+
+    Surface(
+        modifier = modifier.heightIn(min = BADGE_MIN_HEIGHT),
+        color = colors.containerColor,
+        contentColor = colors.contentColor,
+        border = colors.border,
+        shape = RoundedCornerShape(WooTheme.radius.medium),
+    ) {
+        Row(
+            modifier = Modifier.padding(
+                horizontal = WooTheme.padding.padding3,
+            ),
+            horizontalArrangement = Arrangement.spacedBy(WooTheme.spacing.space1),
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            if (leadingIcon != null) {
+                CompositionLocalProvider(LocalContentColor provides colors.contentColor) {
+                    Box(
+                        modifier = Modifier.size(WooTheme.iconSize.size14),
+                        contentAlignment = Alignment.Center,
+                    ) {
+                        leadingIcon()
+                    }
+                }
+            }
+            Text(
+                text = text,
+                style = WooTheme.text.bodySmall.regular,
+            )
+        }
+    }
+}
+
+@Suppress("UnusedPrivateMember")
+@PreviewLightDark
+@Composable
+private fun WooBadgePreview() {
+    WooDesignSystemTheme {
+        Surface(color = WooTheme.colors.background.section) {
+            WooBadgeDemo(modifier = Modifier.padding(WooTheme.padding.padding5))
+        }
+    }
+}
+
+@Composable
+internal fun WooBadgeDemo(
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier,
+        verticalArrangement = Arrangement.spacedBy(WooTheme.spacing.space2),
+    ) {
+        Row(horizontalArrangement = Arrangement.spacedBy(WooTheme.spacing.space2)) {
+            WooBadge("Error", tone = WooBadgeTone.Error, leadingIcon = { BadgeLeadingIcon() })
+            WooBadge("Caution", tone = WooBadgeTone.Caution, leadingIcon = { BadgeLeadingIcon() })
+            WooBadge("Warning", tone = WooBadgeTone.Warning, leadingIcon = { BadgeLeadingIcon() })
+        }
+        Row(horizontalArrangement = Arrangement.spacedBy(WooTheme.spacing.space2)) {
+            WooBadge("Success", tone = WooBadgeTone.Success, leadingIcon = { BadgeLeadingIcon() })
+            WooBadge("Info", tone = WooBadgeTone.Info, leadingIcon = { BadgeLeadingIcon() })
+            WooBadge("Neutral", tone = WooBadgeTone.Neutral, leadingIcon = { BadgeLeadingIcon() })
+            WooBadge(
+                "Outlined",
+                tone = WooBadgeTone.NeutralOutlined,
+                leadingIcon = { BadgeLeadingIcon() },
+            )
+        }
+    }
+}
+
+@Composable
+private fun BadgeLeadingIcon() {
+    Icon(
+        imageVector = WooDsIcons.Regular.Star,
+        contentDescription = null,
+    )
+}
+
+@Composable
+private fun WooBadgeTone.toBadgeColors(): WooBadgeColors {
+    val colors = WooTheme.colors
+    return when (this) {
+        WooBadgeTone.Neutral -> WooBadgeColors(colors.status.neutralContainer, colors.status.onNeutralContainer)
+        WooBadgeTone.NeutralOutlined -> WooBadgeColors(
+            containerColor = Color.Transparent,
+            contentColor = colors.surface.onDefault,
+            border = BorderStroke(WooTheme.stroke.regular, colors.outlineVariant),
+        )
+        WooBadgeTone.Info -> WooBadgeColors(colors.status.infoContainer, colors.status.onInfoContainer)
+        WooBadgeTone.Success -> WooBadgeColors(colors.status.successContainer, colors.status.onSuccessContainer)
+        WooBadgeTone.Warning -> WooBadgeColors(colors.status.warningContainer, colors.status.onWarningContainer)
+        WooBadgeTone.Caution -> WooBadgeColors(colors.status.cautionContainer, colors.status.onCautionContainer)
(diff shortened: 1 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/295">25.2</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooBadge.kt
```diff
@@ -0,0 +1,138 @@
...
+
+@Composable
+private fun WooBadgeTone.toBadgeColors(): WooBadgeColors {
+    val colors = WooTheme.colors
+    return when (this) {
+        WooBadgeTone.Neutral -> WooBadgeColors(colors.status.neutralContainer, colors.status.onNeutralContainer)
+        WooBadgeTone.NeutralOutlined -> WooBadgeColors(
+            containerColor = Color.Transparent,
+            contentColor = colors.surface.onDefault,
+            border = BorderStroke(WooTheme.stroke.regular, colors.outlineVariant),
```
irfano: There's an inconsistency with this border color. In Figma, the border color is `#F4F4F4`, but here we're using `outlineVariant` - `#DCDCDE`.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooIconButton.kt
```diff
@@ -0,0 +1,131 @@
...
+    onClick: () -> Unit,
+    contentDescription: String,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    emphasis: WooIconButtonEmphasis = WooIconButtonEmphasis.Neutral,
+    icon: @Composable () -> Unit,
+) {
+    require(contentDescription.isNotBlank()) {
+        "WooIconButton contentDescription must not be blank"
+    }
```
irfano: Since the parameter is already required and non-null, this check only catches blank values. For hardcoded strings it fails fast in development, which is fine, but if a description ever comes from dynamic data that can be empty, this crashes in production instead of leaving the button unlabeled. Can we rely on the required non-null parameter and drop the runtime crash? Same applies to `WooOutlinedIconButton` and `WooIconContainer`.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooProgressIndicator.kt
```diff
@@ -0,0 +1,104 @@
...
+    modifier: Modifier = Modifier,
+    progress: Float? = null,
+) {
+    val colors = WooTheme.colors
+
+    if (progress == null) {
+        LinearProgressIndicator(
+            modifier = modifier,
+            color = colors.primary,
+            trackColor = colors.secondary,
```
irfano: In light mode the track is darker than the fill, which makes the progress hard to read; the fill should be the darker one. Can we adjust the colors? Applies to all four indicators here.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooButton.kt
```diff
@@ -0,0 +1,305 @@
...
+
+@Composable
+private fun WooButtonStyle.toButtonColors(enabled: Boolean): ButtonColors {
+    val colors = WooTheme.colors
+    val disabledStateLayerColor = colors.surface.onDefault.copy(alpha = DISABLED_STATE_LAYER_ALPHA)
+    val disabledContentColor = colors.surface.onDefault.copy(alpha = DISABLED_CONTENT_ALPHA)
+
+    return when (this) {
+        WooButtonStyle.Filled -> ButtonDefaults.buttonColors(
+            containerColor = if (enabled) colors.primary else disabledStateLayerColor,
```
irfano: M3 Button already switches to the disabled colors when `enabled` is `false`, so the `if (enabled)` branches in `toButtonColors` never affect what is rendered; the disabled look comes only from `disabledContainerColor` and `disabledContentColor`. Can we pass the colors unconditionally and drop the `enabled` parameter from this function?

irfano (changes requested):
The dark mode of the “Filled tonal” button is incorrect; the text is unreadable in your screenshot.

Image

hichamboushaba:

> The dark mode of the “Filled tonal” button is incorrect; the text is unreadable in your screenshot.

Yes, I should have mentioned this in the description, the dark colors are broken from the source, we are matching Figma, and will update when Wagner provides an update.
We just need to ensure the colors of our code match the "token names".

I will review the other comment points, and address what needs addressing.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooIconButton.kt
hichamboushaba: > but if a description ever comes from dynamic data that can be empty, this crashes in production instead of leaving the button unlabeled

But why would a button's content description be empty? This is not a simple icon, it's an actionable button, so it requires a content description.
Personally I prefer the crash to catch the issue than leave it untracked.

If you are convinced, we can discuss it more 🙂

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooProgressIndicator.kt
hichamboushaba: Fixed in 7fb36c0. Since progress indicators do not have Figma-specific DS colors yet, the wrappers now defer to Material3 defaults inside `WooDesignSystemTheme` instead of overriding the active/track colors with `primary`/`secondary`.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooBadge.kt
hichamboushaba: Good find, not sure how this ends up like this, I suspect it's one of the cases where Figma was still using hardcoded colors when I worked on this.

Anyway, fixed in 583c592

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooButton.kt
hichamboushaba: Good find. Fixed in 9fd13f5.

hichamboushaba:
Ready for another round @irfano

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooIconButton.kt
irfano: Something like this looks quite likely to happen at some point:
```
val contentDescription = when (order.status) {
    Processing -> stringResource(R.string.mark_order_complete)
    Completed -> stringResource(R.string.view_order_receipt)
    else -> ""
}

WooIconButton(
    imageVector = WooIcons.Regular.Ellipsis,
    contentDescription = contentDescription,
    ...
)
```
Plugins can add custom order statuses, so this would never fire in development, but it would crash in production on every render of the screen. In some cases an empty content description can also make sense, e.g. for unknown states.

We even have a real case where we use a blank content description: [WooPosFloatingToolbar](https://github.com/woocommerce/woocommerce-android/blob/c5a3764f918962e24acb80d020b5ce733817d780/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/toolbar/WooPosHomeFloatingToolbar.kt#L364-L372). If it used `WooIconButton`, it would crash during the menu's exit animation.

I think a crash is a disproportionate defense for a content description, so I'd still prefer dropping the runtime check. Do you still think fail-fast is a better approach here?

hichamboushaba: > We even have a real case where we use a blank content description: [WooPosFloatingToolbar](https://github.com/woocommerce/woocommerce-android/blob/c5a3764f918962e24acb80d020b5ce733817d780/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/toolbar/WooPosHomeFloatingToolbar.kt#L364-L372). If it used WooIconButton, it would crash during the menu's exit animation.

Honestly I think if the item is clickable, then having an empty content description is a bug that should be fixed.

For the exampe you gave, I think we would use a default value for custom ones, because this is a clickable action, so for someone who is using a screen reader, not saying the action is a deal breaker, it's not just nice to have.

> I think a crash is a disproportionate defense for a content description

I think this would be true if we were talking about icons, not icon buttons.

If you are still not convinced, I'll delete it 😄

irfano: Fair enough. This is always an actionable button, so an empty content description would be a bug anyway. Let's keep it. 👍🏻

irfano (approved):

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooIconButton.kt
hichamboushaba: Thanks @irfano, after thinking more about this and given we'll be migrating old screens here, I decided to drop `require` and replace it `assert` instead, this way it will crash only on debug builds. Check the last commit.

Thanks again for the discussion.

PR #16195, opened 2026-07-03
Add store design system catalog runtime and toolbar bridge

hichamboushaba:
> [!NOTE]
> This is the fourth split PR for Store Design System components. It builds on `store-design-system/components3`, so review the diff against that base rather than `trunk`.

### Description

This PR lands the final Store Design System component slice: the XML toolbar bridge for XML-hosted screens, Developer Options runtime access to the catalog, and the remaining preview-only/demo catalog surfaces.

#### What this adds

- XML toolbar bridge support in `:libs:store-design-system`:
  - `WooDesignSystemToolbar`
  - toolbar XML styles, dimensions, ids, tint/background resources, and debug-only XML test resources
  - focused Robolectric coverage for toolbar behavior and XML inflation
- XML toolbar demo coverage under `designsystem.preview`:
  - `WooDesignSystemToolbarDemo`
  - `WooDesignSystemToolbarPreviewTest`
  - catalog entry for the XML toolbar bridge
- Final catalog/runtime access in `:WooCommerce`:
  - `designSystemComposeView` helper for DS-rooted Compose fragments
  - `StoreDesignSystemComponentCatalogFragment`
  - Developer Options row, event, navigation action, and catalog string wiring
  - `DeveloperOptionsViewModelTest` coverage for row presence and click event emission
- Final preview/catalog updates under `designsystem.preview`:
  - runtime-ready catalog navigation with `WooTopAppBar`
  - preview-only catalog group and deferred demo surfaces in `WooDesignSystemPreviewOnlyComponents`
  - stress/demo catalog entries for large font, long text, and RTL preview review
- Preview-only resource additions first used by the deferred demo catalog.

#### Why this shape

This split keeps the remaining non-production and integration work isolated from the earlier component PRs while still making the final review surface usable in both Android Studio and the debug app.

- The XML toolbar bridge stays scoped to the cases called out in the rollout guidance: XML-hosted screens that still need toolbar behavior such as existing menu inflation and AppCompat interoperability.
- The XML toolbar demo lives in `designsystem.preview`, not in the toolbar implementation file, so the production bridge stays focused on runtime behavior while preview/demo content stays catalog-owned.
- Developer Options becomes the debug-only runtime entry point for the catalog without pulling design-system components into unrelated product flows.
- The final catalog now shows both runtime-ready production slices and explicitly preview-only exploratory work, which keeps reviewer expectations clear.

#### What this does not do

- Does not migrate or adopt these components in any product screen.
- Does not change the PR1, PR2, or PR3 production component implementations beyond the minimal catalog/runtime integration needed for this final slice.
- Does not include unrelated Product List layout or content churn.
- Does not reintroduce removed source-text fidelity tests.
- Does not touch POS.
- Does not add generated token, icon, Code Connect, or Figma automation.

#### Helpful review focus

- Does the XML toolbar bridge preserve the intended XML-screen use case without leaking preview/demo concerns into the production implementation?
- Do the toolbar spacing, action decoration, and AppCompat theme assumptions match the existing XML fragment behavior closely enough?
- Is the Developer Options runtime wiring minimal and consistent with existing debug navigation patterns?
- Does the final catalog clearly separate production components from preview-only explorations and stress/demo surfaces?
- Are the preview-only additions and their resources limited to the deferred final-slice review surface?

No release notes: this is an internal design-system components change.

### Test Steps

1. In Android Studio, render `WooDesignSystemComponentCatalogPreview` from `:libs:store-design-system`.
2. Verify the catalog includes the XML toolbar bridge, preview-only component group, and the deferred stress/demo surfaces from this final slice.
3. In a debug app build, open `Developer Options` and launch `Store Design System Component Catalog`.
4. Verify the runtime catalog navigation opens, supports back navigation, and shows the XML toolbar catalog demo.
5. Optionally render `WooDesignSystemToolbarDemo` and the stress previews directly in Android Studio.

### Images/gif

| Light | Dark |
| --- | --- |
| components4-xml-toolbar-preview | components4-xml-toolbar-preview-dark |
| components4-catalog-preview-light | components4-catalog-preview-dark |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
@JorgeMucientes please note that there are some small differences from what we have on Figma today, as this was created last week, so the review should focus on the API of components, more than visual fidelity.

I will have one PR that handles all differences from Figma after this.

JorgeMucientes:
Hey @hichamboushaba I started reviewing the code and some of the renders but I'd appreciate if you fix the attached images in the description so I can get a better picture of the scope of the expected changes: 
Screenshot 2026-07-10 at 13 04 42

hichamboushaba:
@JorgeMucientes sorry I missed that, I added the screenshots.

For info: the only product related change of the PR is the XML toolbar, the other changes are just related to the components preview screen, which is a dev tool.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooDesignSystemToolbar.kt
```diff
@@ -0,0 +1,390 @@
...
+                }
+                if (item.icon != null) {
+                    changed = child.applyOutlinedToolbarActionStyle(item.icon, iconSize) || changed
+                }
+            }
+        return changed
+    }
+
+    private fun View.applyOutlinedToolbarActionStyle(icon: Drawable?, iconSize: Int): Boolean {
+        var changed = applyToolbarIconTouchTarget()
```
JorgeMucientes, line 118: I think this might be problematic in some wide screen scenarios. This line unconditionally calls `applyToolbarIconTouchTarget()`, which hard-sets `layoutParams.width = height = 48dp` (via `woo_ds_toolbar_icon_touch_target`). An AppCompat ActionMenuItemView is a TextView; when a `SHOW_AS_ACTION_WITH_TEXT` item shows its label (which AppCompat only does on wide screens), the label will get clipped into that 48dp square and clips and only the icon is show. One solution could be to check if the action is icon only before enforcing the `48dp` dimensions

I'm not sure we have any such case currently in the app, but I think is worth fixing this.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooDesignSystemToolbar.kt
```diff
@@ -0,0 +1,390 @@
...
+            .filterIsInstance<ActionMenuView>()
+            .flatMap { actionMenuView -> actionMenuView.children.asIterable() }
+            .forEach { child ->
+                val item = menu.findItem(child.id) ?: return@forEach
+                if (item.actionView === child) {
+                    return@forEach
+                }
+                if (item.icon != null) {
+                    changed = child.applyOutlinedToolbarActionStyle(item.icon, iconSize) || changed
+                }
```
JorgeMucientes: **Non-blocking**, but I think there's a small gap here: the decoration only goes one way. If a menu item's icon is removed at runtime (`item.setIcon(null)), the item.icon != null` guard skips the child on the next pass, but nothing clears the previous decoration — the background still holds a cloned copy of the old icon (via `constantState.newDrawable())`, so it keeps rendering inside the outlined box, and the view stays locked at `48dp`.

Icon changes are handled fine (the !== tag check catches the new instance), it's only removal that falls through.

Definitely an edge case since it needs someone mutating the icon at runtime, but since this is a shared DS component, I'd add a symmetric cleanup path, something like:
```kotlin
if (item.icon != null) {
    changed = child.applyOutlinedToolbarActionStyle(item.icon, iconSize) || changed
} else {
    changed = child.clearOutlinedToolbarActionStyle() || changed
}
```
with `clearOutlinedToolbarActionStyle()` gated on the `woo_ds_toolbar_action_view` tag (so text actions that were never decorated stay untouched), restoring the theme's actionBarItemBackground instead of just nulling the background (to keep press feedback), and resetting the layout params / padding. Returning true from it means the existing re-measure pass in onMeasure picks it up in the same frame.

A quick Robolectric regression test would lock it in: decorate → setIcon(null) → re-measure → assert the background and width are back to defaults.

JorgeMucientes (approved):
Nice job @hichamboushaba. Looks good and works well. Really neat to have a DS components catalog available in the dev options. 

I'm approving but leaving a couple suggestions I think are worth addressing, however I don't think they are blocking 

Also one thing I noticed when tapping on the toolbar actions is that the clicked background color is barely noticeable in dark mode, and it doesn't match the one used in the toolbar arrow back button. Sharing an example of what I mean: 

https://github.com/user-attachments/assets/a5a9cbb0-3c28-4219-9e17-e79ce9ce6d79

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooDesignSystemToolbar.kt
hichamboushaba: Fixed in 5f5ef4c

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooDesignSystemToolbar.kt
hichamboushaba: Fixed in a5e7196

hichamboushaba:
Thanks @JorgeMucientes for the great review, I fixed the points of the review, and fixed the dark-mode toolbar pressed-state feedback in 4166117.

PR #16204, opened 2026-07-06
Add store design system icons

hichamboushaba:
### Description
Introduces the Store Design System icon layer that later component split branches can depend on instead of duplicating app-module drawable icons.

This PR adds source-backed 24dp icon vector drawables to `:libs:store-design-system`, grouped by the Figma styles `light`, `regular`, and `solid`. It also adds the `WooDsIcons` Compose catalog so component code can use named `ImageVector` accessors such as `WooDsIcons.Regular.Store` while keeping the XML vector resources as the source of truth.

Icons source is 50XIH5MmOf4xUYEkM6fAm6-fi-920_6067

This branch is intentionally limited to icon resources, icon accessors, and a short README note. It does not add app runtime wiring, migrate app screens, or include later component slices.

### Test Steps
1. Review the new `libs/store-design-system` icon resources and `WooDsIcons` accessors.
2. Confirm component follow-up branches can reference DS-owned icons instead of adding duplicate app drawables.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_angle_right_24dp.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
```
irfano: Can we add `android:autoMirrored="true"` to icons that should be mirrored in RTL mode, like this one?

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_user_shield_24dp.xml
irfano, line 1: There’s a typo in the Figma file. This icon’s name should have been “User Shield”. Can you please fix it here? I commented in the Figma file, but it’s better to fix it manually here.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/icons/WooIcons.kt
irfano, line 1: I think we could use the `WooIcons` name. We already use `WooIconSize` and `WooFoundationDefaults` for this project, so I don’t see a need for the “Ds” addition.

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_shield_24dp.xml
```diff
@@ -0,0 +1,13 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <group
+        android:translateX="-72"
+        android:translateY="-360">
```
irfano: The path data here keeps the Figma page coordinates and relies on a group translate to land in the viewport, and 141 of the 164 icons are like this. Can we normalize the coordinates to 0-24 and drop the wrapper group?

irfano (changes requested):

irfano (commented):
I also noticed some icons from Figma were not imported in this PR: Check Small and Minus Small. Can we import these two as well and update the README counts to match?

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/icons/WooIcons.kt
hichamboushaba: For the other tokens, they are meant to be used as locals of `WooTheme`, so they are already well scoped. For the icons, I went with the `Ds` because I wanted them scoped well for the store part of the project, but now after thinking more about, I think you are right, it doesn't add any value.

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_angle_right_24dp.xml
hichamboushaba: Good point, fixed in 4e7b86b

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_user_shield_24dp.xml
hichamboushaba: Fixed in 69fa470

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/icons/WooIcons.kt
hichamboushaba: Fixed in 688243e

libs/store-design-system/src/main/res/drawable/woo_ds_ic_regular_shield_24dp.xml
hichamboushaba: Fixed in c0ca1a7

hichamboushaba:
Thanks @irfano for the review, I addressed all the comments, and for the missing icons, these were added after the PR was created, I added them in 1dee391.

irfano (approved):
LGTM! Thanks for the changes. 👍🏻

PR #16208, opened 2026-07-07
[WOOMOB-3372] Allow skipping overdue Stripe requirements in card reader onboarding

malinajirka:
### Description

Fixes WOOMOB-3372

[iOS PR](https://github.com/woocommerce/woocommerce-ios/pull/17491)

Merchants whose Stripe account has an *overdue* requirement were fully blocked from taking in-person payments — but a lot of them can actually still charge. An overdue *payout* requirement (a missing bank account, say) doesn't stop card payments on Stripe's side; the account still has `charges_enabled: true`. We were treating any `past_due` item as "restricted → no payments", which is too broad. A merchant hit this and escalated after being wrongly blocked.

The precise fix would be to read Stripe's real `charges_enabled` signal, but that means changes across both payment extensions and both apps — a lot of surface for something very few people have run into in 5+ years. So this takes the cheap route instead: let them skip the block, the same way we already let merchants skip *pending* requirements. If an account genuinely can't charge, Stripe still rejects the payment, so the worst case is a payment that fails at collection time — same outcome as today, just later and with the choice in the merchant's hands.

I also fixed the screen copy, which claimed in-person payments were "currently unavailable" — no longer true once you can skip and continue.

Kept the block in place for under-review / rejected / misconfigured accounts, which are genuinely not chargeable.

### Test Steps
1. Use a store whose account has an overdue requirement (or mock the onboarding state to `StripeAccountOverdueRequirement`).
2. Start the card reader / In-Person Payments flow.
3. The onboarding screen should now offer a **Skip** button alongside **Resolve now**.
4. Tap **Resolve now** → still opens the external CTA to fix the requirement.
5. Tap **Skip** → the flow continues to reader connection / the card reader hub instead of dead-ending.

### Images/gif
Screenshot_1783406114


samiuelson (approved):
LGTM!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #16213, opened 2026-07-07
Address security alerts and update to Ruby 3.4.9

mokagio:
Addresse the following Dependabot alerts as prompted internally, ref p1783415840767459-slack-C07DX5SGUVA

- https://github.com/woocommerce/woocommerce-android/security/dependabot/183
- https://github.com/woocommerce/woocommerce-android/security/dependabot/197
- https://github.com/woocommerce/woocommerce-android/security/dependabot/192

I'm surprised we got alerts for those only ~2 hours ago, given other repos received days ago. As such, I decided to address them myself instead of waiting for Dependabot to kick in.


.bundle/config
```diff
@@ -1,5 +1,4 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
 BUNDLE_JOBS: "16"
-BUNDLE_SPECIFIC_PLATFORM: "false"
```
iangmaia, line 4: Is this change intentional?
Usually I think we've been using this in combination with `BUNDLE_FORCE_RUBY_PLATFORM`, but I don't recall if having both is really needed.

.ruby-version
```diff
@@ -1 +1 @@
-3.2.2
+3.4.9
```
iangmaia, line 1: 👍 nice. For Android projects, this is indeed something we can do sooner.

iangmaia (approved):
I've disabled auto merge in case you want to revisit any point (mainly the `BUNDLE_SPECIFIC_PLATFORM` I think) but otherwise it looks good 👍

Gemfile
```diff
@@ -5,7 +5,6 @@ source 'https://rubygems.org'
 gem 'danger-dangermattic', '~> 1.3'
 gem 'fastlane', '~> 2.237'
 gem 'fastlane-plugin-firebase_app_distribution', '~> 1.0'
-gem 'nokogiri', '>= 1.19.3' # GHSA-c4rq-3m3g-8wgx — drop floor once toolkit is on >= 14.4.1
 gem 'rubocop', '~> 1.88'
```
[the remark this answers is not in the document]
mokagio: Fair enough, but we'll update at the `release-toolkit` level.

.bundle/config
mokagio: It's intentional.

https://github.com/woocommerce/woocommerce-android/pull/16213/changes/9be5a3bea5314fc2fff7e962fff4af9dd7517bcb

The setting was removed back in Bundler 2.2.0. More details in https://github.com/wordpress-mobile/WordPress-iOS/pull/25748/commits/281e8b031bac865dab110a63a7d392795b5f610c

> Bundler stopped reading `specific_platform` in 2.2.0 (2020-12-10): commit `0bcf7492d` deleted its `settings_flag`, making platform-specific gem resolution unconditional. The `"false"` value has been a no-op since.

Gemfile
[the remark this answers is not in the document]
mokagio: > but fastlane-plugin-wpmreleasetoolkit only constrains Nokogiri to ~> 1.11 (Gemfile.lock)

Actually, that's incorrect. https://github.com/wordpress-mobile/release-toolkit/blob/93464afa1588dda9f1aea4ab15f338136b32905f/fastlane-plugin-wpmreleasetoolkit.gemspec#L37

PR #16214, opened 2026-07-07
[WOOMOB-3424] Chunk notification deletes to avoid SQLite variable limit crash

irfano:
Fixes WOOMOB-3424

### Description

Sentry reports a `SQLiteException: too many SQL variables` crash from `NotificationDao.deleteAllByRemoteIds`: the notification hash sync deletes every stale cached notification in one `DELETE ... WHERE remoteNoteId IN (...)`, and Room binds one variable per id. With a large enough stale set this exceeds SQLite's bind variable limit (999 on older devices).

The delete now runs in chunks of 200 inside a single transaction, following the existing pattern in `OrderSummaryDao`. The public DAO API and its caller are unchanged.

### Test Steps

Reproducing the crash manually needs tens of thousands of stale cached notifications, so it isn't practical; the chunking is covered by the new `NotificationDaoTest` case instead.

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt`.


AdamGrzybkowski (approved):
:shipit:

PR #16222, opened 2026-07-08
Restore ui-tooling-preview on the release compile classpath

malinajirka:
### Description

Release builds don't compile on `trunk`. ddcbab0f18f moved `ui-tooling` to `debugImplementation`, which also dropped `ui-tooling-preview` (where `@Preview` lives) from the release classpath. This adds it back explicitly, keeping `ui-tooling` debug-only.

### Test Steps

1. `./gradlew :WooCommerce:compileVanillaReleaseKotlin` — fails on `trunk`, passes on this branch.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/build.gradle
+++ b/WooCommerce/build.gradle
@@ -451,6 +451,7 @@ dependencies {
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.compose.material.main)
     implementation(libs.androidx.compose.animation.main)
+    implementation(libs.androidx.compose.ui.tooling.preview)
     debugImplementation(libs.androidx.compose.ui.tooling.main)
     implementation(libs.androidx.compose.runtime.livedata)
     implementation(libs.androidx.compose.ui.text.google.fonts)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/295">25.2</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/build.gradle
```diff
@@ -451,6 +451,7 @@ dependencies {
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.compose.material.main)
     implementation(libs.androidx.compose.animation.main)
+    implementation(libs.androidx.compose.ui.tooling.preview)
```
JorgeMucientes, line 454: I think we should probably revert this [commit](https://github.com/woocommerce/woocommerce-android/commit/ddcbab0f18f35cad2647b0b7f83fd2a6f55985cd) completely. Or else wear app will also fail in case we need to build it.

irfano (approved):
Thanks for noticing this and for the quick fix. LGTM! 👍🏻

WooCommerce/build.gradle
irfano: This won’t fail on Wear because the Wear module already has  `implementation(libs.androidx.compose.ui.tooling.preview)`

malinajirka: Yeah, as Irfan said, Wear is fine. Also, adding just `libs.androidx.compose.ui.tooling.preview` has smaller footprint compared to `libs.androidx.compose.ui.tooling.main`.

PR #16223, opened 2026-07-08
[WOOMOB-3448] Update Stripe Terminal SDK to 5.6.0

malinajirka:
One reviewer is enough.

### Description

Fixes WOOMOB-3448

Updates Stripe Terminal 5.1.1 → 5.6.0. Two API changes: `TapToPayConnectionConfiguration` now takes `TapUseCase.Pay(locationId)`, and `Terminal.init` takes a `LocaleConfig` — we pass `CardLanguagePreferenceIfAvailable`. Note the old `init` overload hardcoded `en-US`, so reader prompt language changes.

TTP now requires an Android security patch from the last 12 months (5.2.0). While this might break TTP on devices of some merchants, considering it's a security improvement, I believe it's worth it. It should be very rare for someone to ignore security patches for 12+ months.

P.S. I verified [the changes in the manifest](https://github.com/woocommerce/woocommerce-android/pull/16223#issuecomment-4912945305) are internal only and don't change any values we use, such as identifier of TTP reader.

### Test Steps
**Requires release build variant**
1. Tap to Pay: connect a reader, take a payment.
2. Bluetooth reader (M2 / Chipper): discover, connect, collect.

### Images/gif

N/A


libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/wrappers/TerminalWrapper.kt
```diff
@@ -54,7 +55,14 @@ internal class TerminalWrapper {
...
         tokenProvider: ConnectionTokenProvider,
         listener: TerminalListener
-    ) = Terminal.init(application, logLevel, tokenProvider, listener, null)
+    ) = Terminal.init(
+        application,
+        logLevel,
+        tokenProvider,
+        listener,
+        null,
+        LocaleConfig.CardLanguagePreferenceIfAvailable
```
malinajirka, line 64: 📓 LocaleConfig is a new parameter. I opted for this `card language preference` + by default, if the card doesn't provide any language, it fallbacks to system language which at least the merchant understands. Previously, it'd always use en-us, which the customer and the merchant might not understand.

samiuelson (approved):
I tested collecting payment with TTP and WisePad. The flow works as expected 👍 LGTM!

PR #16128, opened 2026-06-23
Document Store design system rollout direction

hichamboushaba:
Fixes WOOMOB-3328

> [!NOTE]
> The foundation code is in https://github.com/woocommerce/woocommerce-android/pull/16120, if you want to check how these docs are in real code.

### Description

This PR documents the Store Management App path for adopting the Woo Mobile Design System, including the foundation source contract, Android adapter guidance, Material 3 projection rules, rollout direction, and migration playbooks future foundation/component/screen work should build on.

#### What this adds

- `docs/design-system/figma-export.json`, a full Figma variables export used as the source artifact for foundation docs.
- `token-map.md` as the source-backed Store foundation map for colors, text, spacing, padding, radius, and icon sizes.
- Parser/source rules that keep top-level `Semantic` traceability-only and use non-`Semantic` foundation sections such as `Woo theme`, `Typescale`, `Spacing`, `Shape`, `Font theme`, and `Icon` for runtime/public mapping.
- Updated color guidance based on the current Figma variables export:
  - palette ramps, including gray, are public production `WooTheme.colors.palette.*` tokens
  - `surfaceDim`, `surfaceBright`, and `surfaceContainerHighest` are public source-backed roles
- Typography guidance that treats Android default font as the runtime equivalent for Figma's Android `Roboto` source.
- Material 3 reference docs that describe how Store source roles project into Material interop without making Material aliases the Store authoring surface.
- Android adapter and implementation-plan docs for future foundation/module work, including module boundaries and migration-era wrapper naming.
- Rollout, integration checkpoint, component catalog, pilot, and screen-migration docs for future Store screen adoption work.
- Store Compose guidance that points future Store UI work toward the design-system rollout and toolbar policy.

#### What this does not do

- Does not add implementation code.
- Does not add the foundations module or design-system components.
- Does not migrate product screens.
- Does not touch POS.
- Does not globally remap XML/View styles or legacy app themes.
- Does not add Code Connect, generated tokens, or Figma automation.
- Does not introduce permanent duplicate legacy and design-system screen implementations.
- Does not add hand-maintained contrast ratios or alpha-derived contrast notes.

#### Helpful review focus

- Is `WooTheme` clearly documented as the Store authoring API, with Material roles kept as interop projections?
- Are the public color groups clear enough?
- Is the split between public foundations and internal Material projection details precise enough?
- Is `rollout-direction.md` clear enough as the canonical scope source for future agents and PRs?
[...]

PR #16230, opened 2026-07-08
[Design System] Update component documentation

hichamboushaba:
> [!NOTE]
> This is the fifth split PR for Store Design System components. It builds on `store-design-system/components4`, so review the diff against that base rather than `trunk`.

### Description

This docs-only PR recovers the Store Design System documentation updates from the broad component branch and refreshes them against the live `components4` stack.

#### What this adds

- Restores the Figma navigation guide for agents:
  - source shorthand rules
  - live Figma discovery flow
  - component search terms
  - evidence and troubleshooting rules
- Updates the component catalog to match the live `:libs:store-design-system` APIs:
  - production component inventory
  - Material/token adapter boundaries
  - preview-only catalog boundaries
  - not-public `section-header` guidance for agents
- Refreshes rollout, adapter, checkpoint, and token docs for the split stack:
  - XML toolbar bridge boundaries without claiming product-screen adoption
  - Developer Options catalog access as app-level debug wiring
  - `Semantic` as traceability-only for runtime/public mappings
  - public palette ramps and alert `orange` / `onOrange` API names

#### What this does not do

- Does not change Kotlin, XML, resources, Gradle files, or app behavior.
- Does not migrate or adopt components in any product screen.
- Does not document Section Header as a public production API.
- Does not add contrast fixes or override Figma token values.
- Does not add screenshots.
- Does not touch POS.

#### Helpful review focus

- Do the docs describe only the live public APIs and intentional adapter utilities from the split stack?
- Is the Figma navigation guidance clear enough for future source-fidelity checks?
- Are the boundaries clear for section header, `Semantic`, state layers, and search field token naming?
- Does the docs-only PR stay scoped enough to review independently from components4?

No release notes: this is internal design-system documentation.

### Test Steps

1. Review the docs diff against `store-design-system/components4`.
2. Confirm the component catalog entries match the public APIs in `:libs:store-design-system`.
3. Confirm the token docs do not promote top-level `Semantic` or `search.fieldContainer` as public runtime APIs.

### Images/gif

N/A

### DIFF
--- a/docs/design-system/android-adapter.md
+++ b/docs/design-system/android-adapter.md
@@ -31,7 +31,7 @@ New Compose APIs live under:
 com.woocommerce.android.ui.compose.designsystem
 ```
 
-Suggested subpackages:
+Subpackages:
 
 - `foundation`: theme, color, typography, spacing, shape, elevation, and token helpers.
 - `component`: production-ready Woo Mobile Design System components.
@@ -56,6 +56,9 @@ rollout wiring should stay outside the module.
 - The new `WooTheme` accessor lives under `com.woocommerce.android.ui.compose.designsystem`.
   The existing `com.woocommerce.android.ui.compose.theme.WooTheme` composable remains the legacy Store
   wrapper until deliberately removed; new design-system code should not import it.
+- Follow-up `WOOMOB-3515` tracks a detekt guardrail for files that import both the legacy
+  `com.woocommerce.android.ui.compose.theme.*` APIs and the design-system
+  `com.woocommerce.android.ui.compose.designsystem.*` APIs during migration.
 - Do not expose raw Figma variable names as public Android API.
 - Public APIs should expose only production-ready tokens and components.
 - In-progress i1 areas may be documented, tracked, or preview-only until signed off.
--- a/docs/design-system/android-adapter.md
+++ b/docs/design-system/android-adapter.md
@@ -18,6 +18,8 @@ doc defines the technical boundaries that support that rollout.
 - Figma is the design-intent source of truth. Android owns the runtime API contract.
 - Source references use public-repo shorthands: P2 `pe5sF9-5ox-p2`, Figma
   `50XIH5MmOf4xUYEkM6fAm6-fi`. Do not expand them into raw URLs in public repo docs.
+- Agents should use [figma-navigation.md](figma-navigation.md) when inspecting Figma components or
+  collecting live Figma screenshots.
 - Foundation source values come from `docs/design-system/figma-export.json`. This docs branch is
   allowed to lack runtime foundation code; future implementation consumes these docs as the contract.
 
@@ -29,7 +31,7 @@ New Compose APIs live under:
 com.woocommerce.android.ui.compose.designsystem
 ```
 
-Subpackages:
+Suggested subpackages:
 
 - `foundation`: theme, color, typography, spacing, shape, elevation, and token helpers.
 - `component`: production-ready Woo Mobile Design System components.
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

docs/design-system/android-adapter.md
```diff
@@ -29,7 +31,7 @@ New Compose APIs live under:
 com.woocommerce.android.ui.compose.designsystem
 ```
 
-Subpackages:
+Suggested subpackages:
```
irfano: I guess the AI insists on the word "suggested" here. 😄 I suggested removing it [before](https://github.com/woocommerce/woocommerce-android/pull/16128#discussion_r3523733944), but we can keep it if you like. Not a big deal.

docs/design-system/android-adapter.md
```diff
@@ -54,25 +56,22 @@ rollout wiring should stay outside the module.
 - The new `WooTheme` accessor lives under `com.woocommerce.android.ui.compose.designsystem`.
   The existing `com.woocommerce.android.ui.compose.theme.WooTheme` composable remains the legacy Store
   wrapper until deliberately removed; new design-system code should not import it.
-- The foundation implementation should add a detekt guardrail that reports files importing both legacy
-  `com.woocommerce.android.ui.compose.theme.*` and design-system
-  `com.woocommerce.android.ui.compose.designsystem.*` APIs.
```
irfano, line 59: We're dropping this guardrail completely here, and it's not implemented or kept
anywhere else. It still helps app-side screens that mix both roots though. Can we
open a Linear issue to track the detekt rule instead of just deleting it?

docs/design-system/token-map.md
```diff
@@ -255,7 +255,6 @@ the Material builder from another supplied source role.
 | `outline` | `WooTheme.colors.outline` | production | Direct source-backed projection. |
 | `outlineVariant` | `WooTheme.colors.outlineVariant` | production | Direct source-backed projection. |
 | `scrim` | `WooTheme.colors.overlay.overlay50` | production | Source-backed projection; dark alpha is 75%. |
-| `surfaceBright` | `WooTheme.colors.surface.surfaceBright` | production | Direct source-backed projection. |
```
irfano, line 258: No need to drop `surfaceBright` here, this table lists Material roles and what
they project from, not just Figma colors (`surfaceContainerHigh` below isn't in
Figma either). So let's keep the row but point it at `surface.default` instead, since that's what the
code projects now (`surfaceBright = surface.default`).

docs/design-system/integration-checkpoint.md
```diff
@@ -134,9 +146,29 @@ These remain rejected for i1:
 ## Screen Migration Candidate Criteria
 
 For first-wave screens, use [rollout-direction.md](rollout-direction.md) as the assignment source.
-For future unassigned screens, use the
-[Candidate Assessment](screen-migration-playbook.md#candidate-assessment) section in the migration
-playbook.
+The criteria below apply to future unassigned screens.
```
irfano: This block used to link to `screen-migration-playbook.md#candidate-assessment`; now it inlines a near-verbatim copy of both lists. With two copies, editing one and forgetting the other leaves them inconsistent. Can we keep the link instead of duplicating?

irfano (changes requested):

docs/design-system/integration-checkpoint.md
hichamboushaba: Fixed in 0f94fe9

docs/design-system/token-map.md
hichamboushaba: Good point, the agent went far than I asked, fixed in 0f94fe9

docs/design-system/android-adapter.md
hichamboushaba: Fixed in 0f94fe9, I also created WOOMOB-3515 to track this.

docs/design-system/android-adapter.md
hichamboushaba: True 😅, sorry for this, fixed in in 0f94fe9

irfano (approved):
All the changes look good to me. Thank you! Approving. 👍🏻

PR #16234, opened 2026-07-09
Fix time-dependent assertion in RefundStoreTest

malinajirka:
### Description

`RefundStoreTest > when createSimplifiedItemsRefund succeeds, then response is mapped to model` us flaky on CI while passing locally. The failure is a timing bug in the test, not in production code.

`RefundMapper.toModel(SimplifiedRefundResponse)` maps a missing date to the current time:

```kotlin
dateCreated = response.dateCreated?.let { fromFormattedDate(it) } ?: Date(),
```

The test passes `dateCreated = null`, so both the store's mapping and the test's own `mapper.toModel(response)` call fall into the `?: Date()` branch — capturing the wall clock at two different moments, with a coroutine context switch (`coroutineEngine.withDefaultContext`) in between. `WCRefundModel` is a `data class`, so its `equals` compares `dateCreated`; the assertion fails whenever those two reads land in different seconds.

The fix pins the fixture's `dateCreated` to a fixed value, matching the convention already used in `RefundMapperTest`, so both mappings are deterministic. 

### Test Steps

CI is green

### Images/gif

N/A


irfano (approved):
Thanks for handling this. LGTM! 👍🏻

PR #16246, opened 2026-07-11
[WOOMOB-1034] Fix the store login error showing a placeholder instead of the status code

irfano:
Fixes WOOMOB-1034

### Description

When a store is connected with site credentials and the login fails with an HTTP status code, the failure is shown as the subtitle of the Application Password Tutorial screen. That message, `login_site_credentials_http_error` ("Login failed with status code %1$s"), is built as a `UiStringRes` carrying the status code as a param.

The subtitle is produced by a view-model-local `toPresentableString()` helper. It resolved a `UiStringRes` to its string resource but ignored the `params`, so the format placeholder was never filled and the screen showed a raw `%1$s` instead of the code. The fix makes `toPresentableString()` resolve the params recursively, the same way the shared `UiHelpers.getTextOfUiString` already does.

### Test Steps

This message only appears when the login endpoint returns an HTTP error status. To force one, add this snippet to the store (Code Snippets or the theme's `functions.php`):

```php
add_action('login_init', function () {
    if (!empty($_POST['log'])) {
        status_header(403);
        exit;
    }
});
```

Then:

1. Tap **Log In** on the welcome screen.
2. Tap **Log in with site address**.
3. Enter the store address.
4. Tap **Continue**.
5. Enter the site username and password.
6. Tap **Continue**.
7. On the "We couldn't log in into your store" screen, confirm the subtitle shows the actual status code, e.g. "Login failed with status code 403", and not "Login failed with status code %1$s".

### Images/gif

| Before | After |
| --- | --- |
|Screenshot_20260711_132724|Screenshot_20260711_172707|

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,7 +4,7 @@
 
 25.3
 -----
-- [*] Fixed the store login error message showing a placeholder instead of the status code [PR URL]
+- [*] Fixed the store login error message showing a placeholder instead of the status code [https://github.com/woocommerce/woocommerce-android/pull/16246]
 
 25.2
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,7 +4,7 @@
 
 25.3
 -----
-
+- [*] Fixed the store login error message showing a placeholder instead of the status code [PR URL]
 
 25.2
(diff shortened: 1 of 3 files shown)


hichamboushaba (approved):
thanks for fixing this @irfano

Issue #9488, opened 2023-07-27
[Points & Rewards/AutomateWoo Refer A Friend] Discount from redeemed points appears in order details but the amount is not displayed correctly


rachelmcr:
When points from the Points & Rewards extension are redeemed in an order, a discount is applied to the order. The discount and corresponding coupon code appear in the Payment section in order details, but the amount is in quotes.

The same issue occurs when the discount is from a Refer A Friend referral code.

<img src="https://user-images.githubusercontent.com/8658164/217245500-4d3e2983-4e19-422a-9370-7619cd9e07b3.png" width="300px">

malinajirka:
Automated test and comment: I was able to reproduce this in WooCommerce Android dev build 24.9-rc-3 on the Medium_Phone_API_35 emulator. I created a test order (#6132) with a $9.99 product subtotal, a $2.00 discount, and a coupon line whose code is `$2.00`; in Order Details > Payment totals the app shows `Discount ($2.00)` alongside `- $2.00`, matching the reported amount-like discount label behavior.

I migrated the reproducible bug to Linear for triage: https://linear.app/a8c/issue/WOOMOB-3256/order-discount-labels-show-amount-like-coupon-codes-in-payment-totals

PR #16258, opened 2026-07-14
[WOOMOB-3256] Keep order discount codes on one line

irfano:
### Description

Fixes WOOMOB-3256

Long coupon codes wrap onto a second line in the Order Details payment totals.

This change keeps coupon codes on one line and truncates overflowing text with an ellipsis.

### Test Steps

**Prerequisite:** Have an order with a non-zero discount and a long coupon code, such as `wc_points_redemption_27375291_2023_02_07_12_14`.

1. Launch the app.
2. Open the store containing the test order.
3. Tap **Orders** in the bottom navigation.
4. Locate and open the test order.
5. Locate the **Discount** row in the **Payment totals** section.
6. Confirm the coupon code remains on a single line.
7. Confirm overflowing text is truncated with an ellipsis instead of wrapping onto a second line.

### Images/gif
|Before|After|
|-|-|
|image|image|

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.3
 -----
+- [*] Fixed long coupon codes wrapping onto a second line in order details [https://github.com/woocommerce/woocommerce-android/pull/16258]
 - [*] Fixed the store login error message showing a placeholder instead of the status code [https://github.com/woocommerce/woocommerce-android/pull/16246]
 - [*] Fixed the order list not scrolling to the top after creating a new order [https://github.com/woocommerce/woocommerce-android/pull/16240]
 
--- a/WooCommerce/src/main/res/layout/order_detail_payment_info.xml
+++ b/WooCommerce/src/main/res/layout/order_detail_payment_info.xml
@@ -80,6 +80,8 @@
                     android:layout_marginStart="@dimen/minor_00"
                     android:layout_marginEnd="@dimen/minor_00"
                     android:layout_weight="1"
+                    android:ellipsize="end"
+                    android:maxLines="1"
                     tools:text="(sale4theday, firstTime, anyothertime, blah)"/>
 
                 <com.google.android.material.textview.MaterialTextView


hichamboushaba:
@irfano I don't think the bug was about the wrapping; it was specifically about discounts using the "Points & Rewards" and "Refer A Friend" extensions, and that they show the amount like a label. Check the original Github issue https://github.com/woocommerce/woocommerce-android/issues/9488

I think even the AI reproduction that led to the Linear ticket wasn't done correctly, as it didn't use the extension too, it just forced the amount in the discount code, and took this as a confirmation, where it's not really a confirmation.

So, we need to first confirm whether the issue is still reproducible, and if so, we need to fix it.

Side note: I think we should keep the previous wrapping as it provided useful information, and there’s no need to cut the text.

hichamboushaba:
After requesting an agent to investigate this further, here is what it found (TL;DR: the issue was already fixed)

> After checking the original report, the extension behavior, the WooCommerce API shape, and the historical Android code, I don’t think this PR fixes the issue described in WOOMOB-3256/#9488.
> 
> The original screenshot shows two separate values:
> 
> - The Points & Rewards coupon code, `wc_points_redemption_...`, appears in parentheses on the left and wraps onto another line.
> - The actual `$2.00` discount is already in the correct amount column on the right, but is incorrectly surrounded by quotation marks.
> 
> Points & Rewards creates a virtual coupon with a generated code and a separate discount amount. WooCommerce likewise returns these separately as `coupon_lines[].code` and `discount_total`. The original problem was therefore not an amount being mistaken for a coupon code.
> 
> The historical Android code explains the quotation marks exactly: `discountTotal` was formatted using `R.string.orderdetail_customer_note`, whose template was `"%1$s"`. This was changed to the correct negative-currency resource in #8684, merged in April 2023.
> 
> The later reproduction using `$2.00` as the coupon code does not reproduce the original issue. If the coupon code is intentionally `$2.00`, displaying it as `Discount ($2.00)` is expected; the app cannot infer that an intentionally supplied code should be hidden.
> 
> This PR addresses a separate visual issue that happens to be visible in the original screenshot: long coupon codes wrapping. That is a valid improvement, but I suggest tracking or explicitly rescoping it as such rather than presenting it as the fix for WOOMOB-3256/#9488, whose quoted-amount bug appears to have already been fixed by #8684.

But I didn't check this manually.

irfano:
I missed the “Points & Rewards” mention in the original issue and couldn’t reproduce “$2.00.” I fixed two lines of code because, in the web admin, long discount codes are always displayed on a single line. However, I can’t defend that cutting off the long code is better, so there’s no need to add it.

Screenshot_20260717_000331

Also, I checked the “Points & Rewards” extension and confirmed that this issue no longer exists (see the screenshot above). So I’m closing this PR without merging.

PR #16260, opened 2026-07-14
[WOOMOB-3442] Fix crash when toggling the unread reviews filter

AdamGrzybkowski:
### Description

Fixes WOOMOB-3442

**Root cause:** The Reviews list is a Compose `LazyColumn` keyed by `review.remoteId`, which requires globally unique keys. The unread-filter path resolves review ids from the selected site's notifications and reads them back via `ProductReviewsDao.getProductReviews(ids)` — the **only** query in that DAO that filtered solely by `remoteProductReviewId IN (:ids)`, without scoping to `localSiteId` (every sibling query is site-scoped). Because WordPress review/comment ids are per-site, a merchant with more than one store can have the same review id cached under two `localSiteId`s. The un-scoped query then returned both rows, producing two reviews with the same `remoteId` → duplicate `LazyColumn` key → crash.

**Fix:** Scope `getProductReviews(ids)` to `localSiteId` (matching the pattern of every other query in the DAO) and thread the selected site through the single call chain:

### Test Steps

Requires an account with **two connected stores** whose product reviews have overlapping ids (natural for fresh test stores, which number reviews `1, 2, 3…`), and at least one **unread review** on the selected store (so the "Unread reviews only" switch and the mark-all-read ✓ appear).

1. Open the app on the store that has unread reviews (Jetpack-connected, push not taken over by Woo).
2. Go to Menu → Reviews.
3. Toggle the "Unread reviews only" switch several times.
4. Verify the list keeps rendering unread reviews and the app no longer crashes.

### Images/gif

N/A


irfano (approved):
LGTM! 👍🏻

PR #16262, opened 2026-07-14
[WOOMOB-3586] Use WooCommerce's product duplication endpoint

hichamboushaba:
### Description
Fixes WOOMOB-3586

Android now calls WooCommerce Core's server-side product duplicate endpoint so duplicated products preserve the data managed by WooCommerce. The existing client-side implementation remains as a compatibility fallback only when the endpoint is conclusively absent: the error must be `API_NOT_FOUND` with the `rest_no_route` API code.

Saved products with unsaved local edits keep the Duplicate action available, but now ask for confirmation before duplication. Confirming duplicates the captured last-saved `ProductAggregate` through the server endpoint, and the same saved aggregate is used if the exact route-absence response requires the legacy fallback. Cancel keeps the unsaved edit and performs no duplication. Never-saved products hide Duplicate because they do not have a server product to copy.

Other or ambiguous errors remain terminal. Product duplication is non-idempotent, so retrying with the local implementation after an uncertain server response could create two products.

The network response intentionally maps only the duplicated product ID. WooCommerce Core returns raw product data from this endpoint rather than the canonical REST product representation, and the destination product screen performs the normal canonical fetch and cache flow.

`ProductDetailRepository` exposes the operation as Kotlin `Result<Long>`. Once duplication begins, the existing progress dialog, destination navigation, and analytics behavior remain unchanged.

### Test Steps
1. On a WooCommerce 8.9+ test store, choose a saved product with images, a description, a price, inventory, shipping details, and a category.
2. Without editing the product, select the overflow menu > Duplicate. Verify duplication starts immediately without confirmation, the progress dialog appears, and the app opens one distinct Draft with the server-style `(Copy)` suffix and the saved fields populated.
3. Return to the saved source product, make a visible local edit such as changing its title, and do not save it. Select the overflow menu > Duplicate.
4. Verify the confirmation title is “Discard changes and duplicate?”, the message is “Your unsaved changes will be lost. The duplicate will use the last saved version.”, and the actions are “Cancel” and “Discard & duplicate”.
5. Select Cancel. Verify the local edit remains visible and no duplicate is created.
6. Select Duplicate again, then select Discard & duplicate. Verify exactly one Draft is created from the source's last saved version rather than the unsaved edit.
7. Start the New Product flow and open its overflow menu. Verify Duplicate is not shown.
8. To verify the compatibility fallback, configure ApiFaker with exactly one WP API endpoint matching `POST /wc/v3/products/<source-id>/duplicate`, status 404, and this response body:
   ```json
   {"code":"rest_no_route","message":"No route was found matching the URL and request method.","data":{"status":404}}
   ```
9. Enable ApiFaker and duplicate an unchanged saved source once. Verify logcat shows the exact `/duplicate` endpoint match followed by one legacy `ADD_PRODUCT` request, the app opens one new Draft, and no second duplicate is created.
10. Clear the ApiFaker endpoint, then configure the same request with status 404 and this regular invalid-product response body:
    ```json
    {"code":"woocommerce_rest_product_invalid_id","message":"Invalid product ID.","data":{"status":404}}
    ```
11. Duplicate the saved source once. Verify the app shows “Cannot duplicate product,” remains on the source product, creates no Draft, and logcat contains no `ADD_PRODUCT` request.

### Images/gif
N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.3
 -----
+- [*] Improved product duplication to preserve product details managed by WooCommerce [https://github.com/woocommerce/woocommerce-android/pull/16262]
 - [*] Fixed the store login error message showing a placeholder instead of the status code [https://github.com/woocommerce/woocommerce-android/pull/16246]
 - [*] Fixed the order list not scrolling to the top after creating a new order [https://github.com/woocommerce/woocommerce-android/pull/16240]
 
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/DuplicateProduct.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/DuplicateProduct.kt
@@ -10,6 +10,7 @@ import com.woocommerce.android.util.WooLog
 import com.woocommerce.android.viewmodel.ResourceProvider
 import org.wordpress.android.fluxc.network.BaseRequest.GenericErrorType.NETWORK_ERROR
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType.API_NOT_FOUND
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType.GENERIC_ERROR
 import javax.inject.Inject
 
@@ -20,6 +21,21 @@ class DuplicateProduct @Inject constructor(
 ) {
 
     suspend operator fun invoke(productAggregate: ProductAggregate): Result<Long> {
+        val coreResult = productDetailRepository.duplicateProduct(productAggregate.remoteId)
+        if (coreResult.isSuccess) return coreResult
+
+        val error = (coreResult.exceptionOrNull() as? WooException)?.error
+        val shouldUseLegacyFlow = error?.run {
+            type == API_NOT_FOUND && apiErrorCode == REST_NO_ROUTE_ERROR_CODE
+        } == true
+        return if (shouldUseLegacyFlow) {
+            duplicateProductWithLegacyFlow(productAggregate)
+        } else {
+            coreResult
+        }
+    }
+
+    private suspend fun duplicateProductWithLegacyFlow(productAggregate: ProductAggregate): Result<Long> {
         val newProduct = productAggregate.copy(
             product = productAggregate.product.copy(
                 remoteId = 0,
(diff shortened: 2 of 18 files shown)


JorgeMucientes (commented):
Thanks for working on this @hichamboushaba. Code changes look good and work as expected overall. I also tested products with variations and work well (did not test with a 100 or more variations though). 

One thing I noticed is that the Woo core duplication endpoint fails to duplicate all the stock details. The `GTIN | UPC | EAN | ISBN` is lost. 

| Original  | Duplicated |
|    |   |

Although, now that I think about it, it makes sense as those are unique IDs and the endpoint can't make them up like the SKU, so I guess it makes sense to leave them empty. 

Another thing I'd like to surface is the behavioral changes this new use of duplicate endpoint introduces over the previous implementation. In particular around products with unsaved edits:

| Scenario | Previous implementation | New endpoint path |
|---|---|---|
| Existing product with unsaved edits | Passed the current `productAggregateDraft`, reset identity fields, then created a product from that aggregate | Passes only `remoteId`; Core duplicates the persisted server product |
| Never-saved product (`remoteId = 0`) | Created a new draft from the current local aggregate | Calls `/products/0/duplicate`, which fails with invalid product ID |

I wonder if we should: 
- Disable product duplication if there are unsaved edits or the product was never saved. 
- Show a confirmation dialog when attempting to duplicate product with local changes that are going to be lost. I checked `wc-admin` and product editor has an option `copy to a new draft` always available. I presume this option is the same as product duplication. When tapping on that option with unsaved changes there's a confirmation dialog: 

Image

I lean towards adding something similar on mobile with a confirmation dialog before loosing local changes. Wdyt?

hichamboushaba:
Good find, @JorgeMucientes, about the unsaved changes.

I think the simplest change would be to hide the button in those scenarios, the dialog that web displays is like the "discard changes" dialog we have on mobile, and I'm not sure it's very clear there for the user.

So I will proceed to hiding the button unless you have a strong opinion towards the dialog.

Edit: I decided to go with a dialog after all 😅, we'll hide the button when the product is not saved yet, and we'll show a dialog when there are unsaved changes.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
Thanks @JorgeMucientes — addressed in 913635f9b0e. Saved products with unsaved edits now show a confirmation and duplicate the captured last-saved aggregate (including the legacy fallback); never-saved products hide Duplicate.

image

JorgeMucientes (approved):
Thank you for addressing the changes @hichamboushaba. Works perfect!

PR #16263, opened 2026-07-15
[WOOMOB-1896] Attach full application logs as a file to Zendesk support tickets

AdamGrzybkowski:
### Description

Fixes WOOMOB-1896

This PR keeps the trimmed logs in the sidebar field (unchanged, for convenience) **and additionally attaches the full, untruncated logs as an `application_log.txt` file** to every support request. This mirrors the iOS counterpart [WOOMOB-1800](https://linear.app/a8c/issue/WOOMOB-1800), which shipped the same approach in [woocommerce-ios#16445](https://github.com/woocommerce/woocommerce-ios/pull/16445) and has been running in production.

### Test Steps

1. Log in to a store.
2. Go to **Settings → Help & Support → Contact Support**.
3. Fill in the form and submit a support request.
4. In Zendesk, open the created ticket and verify **both**:
   - the trimmed logs still appear in the left sidebar `logs` field, and
   - an `application_log.txt` file is attached, containing the full (untruncated) logs.
5. Trigger the connectivity-test support flow and confirm it still attaches `connectivitytest_log.txt` (both attachments can coexist).

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
@@ -26,6 +26,7 @@ import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
 import kotlinx.parcelize.Parcelize
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.store.SiteStore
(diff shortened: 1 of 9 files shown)


JorgeMucientes:
Quick question @AdamGrzybkowski 
I created a Zendesk ticket 11448350-zd-a8c after running the connectivity troubleshooting and didn't see the  `connectivitytest_log.txt` attached in it:



Is there something specific I need to do to get them attached after running the troubleshoot and tapping on "Contact Support" directly?

WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
```diff
@@ -130,37 +134,42 @@ class ZendeskTicketRepository @Inject constructor(
...
-        diagnosticLog: String
+    private suspend fun uploadTextAttachment(
+        fileName: String,
+        content: String
     ): String? {
         val tempFile = withContext(dispatchers.io) {
-            File.createTempFile(DIAGNOSTIC_LOG_FILENAME, null).also {
-                it.writeText(diagnosticLog)
+            File.createTempFile(fileName, null).also {
+                it.writeText(content)
```
JorgeMucientes: Could we also make the local file preparation best-effort? `File.createTempFile()` and `writeText()` run before the guarded upload, so an I/O failure—for example, low disk space—would exit the flow before the Zendesk ticket is created. Since the application-log upload now runs for every request, an optional attachment could block contacting support entirely. 
Wdyt if we catch file-preparation failures, log them, and continue without the attachment token?

AdamGrzybkowski:
>Is there something specific I need to do to get them attached after running the troubleshoot and tapping on "Contact Support" directly?

Looks like a bug. With the AI chat feature enabled with lose the diagnostic there - with it disabled, the `contact support` goes straight to the submit ticket flow.

AdamGrzybkowski:
@JorgeMucientes I've applied your suggestion and also fixed the lost diagnostic logs (it's out of scope, but I hope you don't mind).

JorgeMucientes (approved):

PR #16265, opened 2026-07-15
[Build] Skip Remote Build Cache When Secrets Are Missing

ParaskP7:
Closes: [AINFRA-2560](https://linear.app/a8c/issue/AINFRA-2560/build-wcandroid-fix-fresh-clone-debug-build-to-not-require-mobile)

---

### Description

A fresh `git clone` of this repo cannot build the debug variant. `./gradlew assembleJalapenoDebug` aborts during settings evaluation with:

> The build requested remote build cache, but secrets file is not found. Please run `bundle exec fastlane run configure_apply` to apply secrets.

The cause is the combination of two deliberate behaviors: `developer.properties-example` ships with `use_remote_build_cache_locally=true` (a deliberate default from [#15189](https://github.com/woocommerce/woocommerce-android/pull/15189), which this PR keeps), and `settings.gradle` falls back to that example when the developer has no `developer.properties` of their own — so every fresh clone implicitly "requests" the remote build cache and then hard-fails on the missing secrets it can never have. External contributors and secret-less machines are blocked before any task runs, even though the rest of the build degrades gracefully (`secrets.properties` → `defaults.properties`, `google-services.json` → example file).

This PR contains two commits:

1. Register the remote build cache only when secrets are applied. When they are missing, a developer who explicitly opted in via their own `developer.properties` file still gets the hard stop (they are expected to have secrets), while a fresh clone — where the flag is `true` only through the example fallback — gets a warning and continues with the local build cache. Developers with secrets applied keep the default-on remote build cache unchanged, and `checkForRemoteBuildCacheOptimizedExperience()` stays aligned with the shared implementation in [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/settings.gradle#L141-L145). cd3de6de45d
2. Drop `install_gems` + `configure_apply` from the CI jobs that do not need secrets (`unit tests`, `gradle cache`, `manifest diff`), so CI stops exposing secrets to jobs that never consume them and permanently exercises the same no-secrets path a fresh clone goes through. The jobs that genuinely need secrets keep them: `release builds` (upload keystore, Play Store credentials, `sentry.properties`), `instrumented tests` (the Firebase Test Lab service account at `.configure-files/firebase.secrets.json` plus real e2e API credentials), and `prototype builds` (the distributed APK must carry real OAuth and `google-services` values and be signed with the shared debug keystore). 88d0a0dd407

The resulting behavior matrix when `use_remote_build_cache_locally=true`:

| Scenario | Before | After |
| --- | --- | --- |
| CI | Remote cache via env var | Unchanged |
| Secrets applied | Optimized-experience checks + remote cache | Unchanged |
| No secrets, own `developer.properties` | Hard stop | Unchanged (hard stop) |
| No secrets, fresh clone (example fallback) | Hard stop | ⚠️ Warning, local cache only |

---

### Considered alternatives

- Flipping `use_remote_build_cache_locally=false` in `developer.properties-example` fixes the fresh clone too, but reverts the deliberate default-on decision from [#15189](https://github.com/woocommerce/woocommerce-android/pull/15189), so it was discarded.
- Warning for everyone (no hard stop at all): simplest, but an internal developer who forgot `configure_apply` would silently lose the remote cache behind an easily-ignored warning — the hard stop exists precisely for them.
- Distinguishing internal developers via `git config user.email` ending in `@automattic.com`/`@a8c.com`: identity-based and covers new machines, but adds process execution to settings evaluation; kept as a fallback idea.
- Probing for internal-infra directories (e.g. the configure tool's output directory) was rejected outright — we do not want a public build script to reference internal secrets-infrastructure layout.

The chosen design (hard stop only behind an explicit, personally-created `developer.properties`) matches the semantics JP/WPAndroid already has: an explicit request for a secrets-dependent feature may fail hard; an implicit default never should.

---

### Test Steps

All on a machine without applied secrets (`~/.configure/woocommerce-android/secrets/secrets.properties` absent):

1. Fresh clone (no `developer.properties`): `./gradlew assembleJalapenoDebug` succeeds and logs the ⚠️ remote-cache warning.
2. Create `developer.properties` with `use_remote_build_cache_locally=true`: `./gradlew help` fails with the original `configure_apply` error.
3. Flip it to `false`: `./gradlew help` succeeds with the plain ℹ️ disabled message.
4. With secrets applied: behavior is unchanged (remote cache + optimized-experience checks).
5. CI on this PR: the `unit tests`, `gradle cache`, and `manifest diff` jobs run without `configure_apply` and pass.
    - Manual `Dependency Cache` Build: [40672](https://buildkite.com/automattic/woocommerce-android/builds/40672)

---

### Images/gif

N/A — build and CI configuration only.

---


settings.gradle
```diff
@@ -81,16 +81,25 @@ buildCache {
...
+                credentials {
+                    username = "developer"
+                    password = gradle.ext.secretProperties.get('gradleCacheNodePassword')
+                }
             }
+        } else if (file("${rootDir}/developer.properties").exists()) {
+            // The developer explicitly opted in to the remote build cache via their own
+            // developer.properties file, so they are expected to have secrets applied.
+            assertSecretsApplied()
+        } else {
```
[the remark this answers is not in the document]
ParaskP7: The merge semantics are described accurately, but the existence check is intentional — the disagreement is about the intended contract, not the mechanics. A `developer.properties` file is created by copying `developer.properties-example` — which contains `use_remote_build_cache_locally=true` — so having the file means having (and accepting) that line: there is no other documented way to create it, and `checkDeveloperPropertiesMatchDefaults()` already warns whenever the file diverges from the example. A hand-crafted partial file without the key is an anomaly, and in that case the holder is almost certainly an internal developer who forgot `configure_apply` — for whom the hard stop is the desired behavior, not a regression.

settings.gradle
```diff
@@ -81,16 +81,25 @@ buildCache {
...
+
+            remote(HttpBuildCache) {
+                url = gradle.ext.secretProperties.get('gradleCacheNodeUrl')
+                push = false
+                credentials {
+                    username = "developer"
+                    password = gradle.ext.secretProperties.get('gradleCacheNodePassword')
+                }
             }
+        } else if (file("${rootDir}/developer.properties").exists()) {
```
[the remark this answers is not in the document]
ParaskP7: [SAME](https://github.com/woocommerce/woocommerce-android/pull/16265#discussion_r3593452799)

settings.gradle
```diff
@@ -81,16 +81,25 @@ buildCache {
...
+            // The developer explicitly opted in to the remote build cache via their own
+            // developer.properties file, so they are expected to have secrets applied.
+            assertSecretsApplied()
```
mokagio: I'm unsure about the code flow here. The way I read it is:

```
if secretsApplied {
  ...
} else if ... {
 assertSecretsApplied // throws if secretsApplied == false
...
```

So, the `else if` branch runs only if `secretsApplied == false` but then it checks for the secrets to be applied, which we already know to be false, resulting in a `throw`.

What am I missing?

ParaskP7: 👋 @mokagio and thanks for this comment, your reading is exactly right and the always-throws is intentional! The `else if` isn't there to re-check the secrets, it is there to distinguish **who** is missing them:

- **A developer with their own `developer.properties`** (created by copying `developer.properties-example`, i.e. explicitly accepting the `use_remote_build_cache_locally=true` opt-in) but without secrets applied. That's almost certainly an internal developer who forgot `configure_apply`, and for them the branch deliberately keeps the pre-existing hard stop, with its actionable message.
- **A fresh clone** (no personal `developer.properties`, the flag is `true` only via the `developer.properties-example` fallback) never reaches that branch. It falls through to the `else` below and gets the ⚠️ warn-and-continue instead, which is the whole point of this PR.

So in that branch `assertSecretsApplied()` is effectively "fail with the canonical message": we know it will throw, and calling it (instead of inlining a `throw`) keeps the error message defined in one place and `checkForRemoteBuildCacheOptimizedExperience()` byte-identical to [JP/WPAndroid's](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/settings.gradle#L141-L145). There's more on the why (and the rejected alternatives) in the PR description's "Considered alternatives" section, plus the [bot thread above](https://github.com/woocommerce/woocommerce-android/pull/16265#discussion_r3588936359) where the "own `developer.properties` = explicit opt-in" contract is spelled out.

Let me know what you think, happy to revisit, but indeed, I thoughts about this, on various angled, before deciding to go with this approach. 💭

---

FYI: It is traveling day for me today, I might be slow to reply.

wzieba: I think @mokagio 's right on the flow and I'm not sure if the answer above addressed the concern: I don't think "asserting secrets applied after we check that secrets **don't** exist" is helpful here in any way.

Maybe instead, you just want a `throw` on line 98?

ParaskP7: 👋 @wzieba , that's fair, and you're both right on the mechanics: in that branch the assert can only ever throw. The reason it called `assertSecretsApplied()` instead of throwing inline was to keep the error message defined in one place (and reused by `checkForRemoteBuildCacheOptimizedExperience()`), not the re-check itself. But two of you tripping over the same line is good evidence the intent doesn't read from the code. 🫣

I've split the throw into its own `throwSecretsNotApplied()`, which the branch now calls directly and `assertSecretsApplied()` delegates to. The call site says what it does, the message stays single-sourced, and `checkForRemoteBuildCacheOptimizedExperience()` stays untouched. Pushed in ffd49699b71. Let me know what you think! 🙏 Cc @mokagio

wzieba: looks good, thanks!

mokagio: @ParaskP7 @wzieba not blocking! But I've been thinking about this and I think where the confusion came for me was that `use_remote_build_cache_locally` is carrying two meanings.

`gradle.ext.developerProperties` is the *merge* of `developer.properties-example` (fallback) under
`developer.properties` (primary), and the example ships `use_remote_build_cache_locally=true`. So by the
time `buildCache {}` reads it, `true` means either:

- this developer asked for the remote cache, or
- nobody said anything and the example default is on.

Those two want different behaviour when secrets are missing (hard fail vs. quiet skip) but the merge has
already erased the difference.

I asked Opus 5 for suggestion to disentangle and this is the proposal I liked the most.

<blockquote>

### Declare the second meaning instead of inferring it

Add to `developer.properties-example`:

```properties
# Fail the build if the remote cache was requested but secrets are not applied.
require_remote_build_cache=false
```

Then the block flattens to a single chain — no nesting, nothing deduced from the filesystem:

```groovy
def remoteCacheRequested = gradle.ext.developerProperties.get(USE_REMOTE_BUILD_CACHE_LOCALLY).toBoolean()
def remoteCacheRequired = gradle.ext.developerProperties.get(REQUIRE_REMOTE_BUILD_CACHE).toBoolean()

if (gradle.ext.isCi) {
    remote(HttpBuildCache) { /* CI, unchanged */ }
} else if (remoteCacheRequested && secretsApplied()) {
    checkForRemoteBuildCacheOptimizedExperience()
    remote(HttpBuildCache) { /* developer, unchanged */ }
} else if (remoteCacheRequested && remoteCacheRequired) {
    throwSecretsNotApplied()
} else if (remoteCacheRequested) {
    logger.warn("\n⚠️ …secrets are not applied, continuing without the remote cache. Run `bundle exec fastlane run configure_apply` to enable it.")
} else {
    logger.warn("\nℹ️ Remote build cache is disabled…")
}
```

Every branch states its condition completely. A fresh clone can't accidentally satisfy the strict path,
anyone who wants the hard stop opts into it with one line, and the "hand-crafted partial `developer.properties`"
edge case we've been arguing about stops existing.

On the objection to warning everyone: the warning isn't silent — it names the exact `configure_apply` command
to run. If the worry is that warnings get scrolled past, `require_remote_build_cache=true` is the honest
version of that: opt into the failure rather than have the build infer you wanted one.

**If adding a property is more than you want to take on here**, the same flat chain works with the current policy —
just compute both bits once *before* the block, `remoteCacheIsOwnChoice = file("${rootDir}/developer.properties").exists()`
in place of `remoteCacheRequired`. The win is deciding once rather than nesting the decision.

One leftover in either version: `checkForRemoteBuildCacheOptimizedExperience()` still calls
`assertSecretsApplied()`, now only reachable when secrets *are* applied. Harmless, and it keeps parity with
WPAndroid's `settings.gradle` — worth a line in the description so the next reader doesn't re-file it.

</blockquote>

ParaskP7: 👋 @mokagio , liked it, so the property-free variant of your suggestion is now in, see cd4447b1383 ! ❤️ 👍 

Both inputs are computed once (`remoteCacheRequested`, `remoteCacheIsOwnChoice`) and the block is a single flat chain where every branch states its full condition, no nesting, no behavior change.

FYI: Two deliberate deviations from the full proposal:

1. I skipped the new `require_remote_build_cache` property for now. It would grow the developer-facing surface for a policy nobody has asked for yet, and with the flat chain in place, adding it later is a one-line branch.
2. I kept the two branch comments. `remoteCacheIsOwnChoice` is an interpretation of "the file exists", and the hand-crafted partial `developer.properties` edge case is exactly where the name alone would over-promise.

ParaskP7:
👋 @mokagio and @wzieba , just a friendly reminder of this! 🙏

settings.gradle
mokagio: Sounds good! Thanks for following up @ParaskP7

mokagio (approved):
As far as my Gradle understanding goes, this looks good. Let's see how it goes in prod.

settings.gradle
ParaskP7: Awesome, thanks for taking a final look at it @mokagio! 🙇❤️🚀

PR #16261, opened 2026-07-14
[WOOMOB-3477] Migrate Dashboard header and Customize action

hichamboushaba:
### Description

Part of WOOMOB-3477

> [!NOTE]
> This is the second part of WOOMOB-3477 and does not close the ticket. A follow-up PR will migrate the full Dashboard screen to Compose and close the ticket.
>
> This is a stacked PR targeting `issue/woomob-3477-dashboard-design-system`. It depends on #16248 and should remain a draft until the parent PR lands.

> [!NOTE]
> When JITM is hidden, its existing XML host still reserves 8dp between the page header and Dashboard content. This pre-existing View/Fragment spacing is intentionally deferred to the full-Compose follow-up, where spacing will be applied only when JITM displays visible banner content.

This follow-up completes the Dashboard shell migration by replacing the Activity-owned toolbar and menu with a Compose `WooPageHeader`. The header uses the current store name as its title, preserves the conditional Share Store action, collapses with Dashboard scrolling, and expands after reselecting **My Store**.

It also moves Customize below the complete dashboard grid, preserving the existing edit-widgets flow and analytics. When new sections are available, the button shows a design-system indicator and exposes a localized state description to TalkBack.

The legacy Dashboard menu, toolbar badge, and expanded-toolbar wiring are removed. The parent PR already includes the Dashboard design-system release note, so this stack does not add a duplicate entry.

### Test Steps

1. Open **My Store** and verify the current store name appears in the expanded Compose page header without a duplicate Activity toolbar.
2. Scroll the Dashboard and verify the page header collapses. Return to the top or reselect **My Store** and verify it expands again.
3. For an eligible store, verify Share Store appears in both header states and opens the existing sharing flow.
4. Scroll past every dashboard section and verify Customize appears after the complete grid.
5. Tap Customize, verify the widget editor opens, then navigate back without changing the configuration.
6. If new dashboard sections are available, verify the indicator appears and TalkBack announces “New dashboard sections available.”
7. Verify pull-to-refresh activates after the Dashboard content and header have returned to the top.
8. Repeat the expanded/collapsed header and Customize checks in light and dark themes. On a wide device, verify Customize remains below both dashboard columns.

### Images/gif

| Header state | Light | Dark |
| --- | --- | --- |
| Expanded | Expanded Dashboard page header in light theme | Expanded Dashboard page header in dark theme |
| Collapsed | Collapsed Dashboard page header in light theme | Collapsed Dashboard page header in dark theme |

[...]

PR #16266, opened 2026-07-16
[WOOMOB-3477] Complete Dashboard migration to Compose

hichamboushaba:
### Description

Fixes WOOMOB-3477

Completes the Dashboard migration by replacing the remaining XML screen shell and Jetpack benefits banner with one Store Design System Compose root.

The Fragment remains responsible for XML Navigation, SafeArgs, scoped ViewModels, events, analytics, and lifecycle behavior. Existing Dashboard card content is unchanged. The JITM fragment is hosted inside Compose at its natural content height, its inset belongs to the scroll body so content meets the collapsed header seamlessly, and the Dashboard E2E components now target the retained root ID.

This PR is stacked on #16261, which provides the collapsible Dashboard header and Customize action.

### Test Steps

1. Open My Store and confirm the Dashboard header, share action, cards, and bottom Customize action render correctly.
2. Scroll down until the header collapses and confirm content meets the header divider without a persistent padding band; then reselect My Store and confirm the body returns to the top while the header expands.
3. Pull to refresh and confirm the Dashboard remains usable and its last-updated value advances.
4. Open and return from Customize, then use Share and confirm the Android share sheet opens.
5. Rotate the device and switch between light and dark themes; confirm the shell remains responsive without clipping or duplicate content.
6. If a JITM or Jetpack benefits banner is available, confirm it uses its natural height and that banner activation and dismissal remain separate actions.

### Images/gif

Pixel 8a emulator · Android 15 · 1080×2400

| Theme | Expanded page header | Collapsed page header |
|---|---|---|
| Light | Light theme with expanded Dashboard page header | Light theme with collapsed Dashboard page header |
| Dark | Dark theme with expanded Dashboard page header | Dark theme with collapsed Dashboard page header |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please ensure the PR has the <code>category: tracks</code> label.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/DashboardScreen.kt
```diff
@@ -0,0 +1,289 @@
...
+                    contentDescription = stringResource(R.string.share_store_button),
+                    onClick = onShareStoreClicked,
+                )
+            }
+        },
+    )
+}
+
+@Composable
+fun JetpackBenefitsBanner(
```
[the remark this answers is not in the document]
hichamboushaba: Fixed

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/DashboardScreen.kt
```diff
@@ -0,0 +1,283 @@
...
+                Box(modifier = Modifier.padding(vertical = WooTheme.padding.padding2)) {
+                    jitmContent(Modifier.fillMaxWidth())
+                }
+            }
+        }
+        AnimatedVisibility(visible = showJetpackBenefitsBanner) {
+            JetpackBenefitsBanner(
+                onClick = onJetpackBenefitsBannerClicked,
+                onDismiss = onJetpackBenefitsBannerDismissed,
+            )
```
JorgeMucientes, line 134: Perhaps we can avoid adding the JetpackBenefitsBanner completely. Not saying we need to tackle the whole Jetpack benefits banner cleanup in this PR, that is out of scope, but I don't think it makes much sense to add a new composable view to this screen that we know it's never going to be displayed (now that new PN system has been enabled to all users). It feels like adding unnecessary noise. 
Wdyt?

WooCommerce/src/main/res/values/ids.xml
```diff
@@ -5,4 +5,5 @@
     <item type="id" name="product_selector_compose_view" />
     <item type="id" name="product_configuration_view" />
     <item type="id" name="analytics_hub_settings_view" />
+    <item type="id" name="dashboard_container" />
```
JorgeMucientes, line 8: Should we name this `dashboard_widget` instead? I mean both usages of this id refer to stats and top performers components in e2e tests.

JorgeMucientes (commented):
Great job @hichamboushaba! I'm pre-approving as I haven't seen any blocking issues, but I'm leaving a couple suggestions and points for discussion before merging. 

- The outlined button in the Blaze card to create campaign in dark mode is barely visible. Is that the correct color for outlined button? 

Image

- Same goes for the info icon in dark mode. It is very dark. Shouldn't it have the brighter purple used for drawing the line chart for example? 

Image

- I'm aware if part of the foundation icons, but perhaps the shared icon should be one fo those cases were Android needs to defer. I mean it is a very iOS-ish icon and from an Android user perspective its hard to understand that is a Share CTA

- The customise button at the end of the dashboard feed, feels very hidden. Was there any discussion with @wagveloso on why it is a good idea to remove the action from the toolbar?

JorgeMucientes (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/DashboardScreen.kt
hichamboushaba: I think staying close to what we have would make keeping the branch up to date, as we know all the changes are related just to migration, nothing else, especially since we need to remove the business logic too, and this has to wait for us to remove the feature flag.

WooCommerce/src/main/res/values/ids.xml
hichamboushaba: We are [setting](https://github.com/woocommerce/woocommerce-android/blob/7b8b3af9521bb71692de1ffd0c8a03646b77d726/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/DashboardFragment.kt#L125) this to the container itself, this replaces the old XML id `my_store_stats_container`.

> both usages of this id refer to stats and top performers components in e2e tests

In e2e tests, the `id` is just a way to find the components and confirm they exist, and there for stats and top performers we used the parent id, and we keep doing so.

hichamboushaba:
Thanks @JorgeMucientes for the review.

> The outlined button in the Blaze card to create campaign in dark mode is barely visible. Is that the correct color for outlined button?

For dark color, the current branch is out of sync with the last Figma export, so the color is not correct. But Figma is actually darker than that 50XIH5MmOf4xUYEkM6fAm6-fi-3157_1138.
But what matters here: for the screen migrations, what matters is the screen itself, any component issues can be addressed separately, and fixed at the root. We are continuously iterating on the components whenever we notice any issue.

> Same goes for the info icon in dark mode. It is very dark. Shouldn't it have the brighter purple used for drawing the line chart for example?

I didn't touch this, it's part of the XML view, for cards, I tried to keep things close to what they are now, and we'll update the stats anyway in the analytics project.

> I'm aware if part of the foundation icons, but perhaps the shared icon should be one fo those cases were Android needs to defer. I mean it is a very iOS-ish icon and from an Android user perspective its hard to understand that is a Share CTA

Good point, but similar to the components issues, it should be addressed in a separate PR after getting @wagveloso approval.

> The customise button at the end of the dashboard feed, feels very hidden. Was there any discussion with @wagveloso on why it is a good idea to remove the action from the toolbar?

Yes, it was discussed with @wagveloso, we can discuss it async and adjust if needed in separate PRs.

PR #16267, opened 2026-07-16
[WOOMOB-3609] Fix system back discarding edits on editor screens

AdamGrzybkowski:
Fixes WOOMOB-3609

### Description

In 25.1 / 25.2-rc-1, editing a value and leaving a screen via the **system/gesture back button** silently discarded the edit. Reported symptom: product **stock** wouldn't save (no Save button appeared). It affects every screen that saves-or-confirms on back.

**Root cause.** These screens implement `MainActivity.BackPressListener` and do their save/discard in `onRequestAllowBackPress()`, which was only invoked from `MainActivity`'s single `OnBackPressedDispatcher` callback. `androidx.activity` 1.12.4 (first shipped in 25.1) changed predictive-back dispatch so a `NavHostFragment`'s own back callback now consumes the gesture **before** MainActivity's callback runs — popping the screen without ever calling `onRequestAllowBackPress()`. The toolbar arrow calls save-and-exit directly, bypassing the dispatcher, so it stayed working.

**Fix.** `BaseFragment` registers its own `OnBackPressedCallback` in `onResume()` when the fragment is a `BackPressListener`, so it out-prioritizes the nav host and routes the gesture back through `onRequestAllowBackPress()` — same semantics MainActivity used (`true` → allow normal back; `false` → fragment handled it). One central change covers all affected screens; no per-screen or ViewModel edits. Mirrors the earlier `OrderCreateEditFormFragment` fix (#15016).

### What changed

- **`BaseFragment`** — the central per-fragment back callback (the fix above).
- **Order filter dialogs** (`OrderFilterCategoriesFragment`, `OrderFilterOptionsFragment`) — these are `DialogFragment`s whose back is dispatched by the **dialog's own** dispatcher, not the activity's, so the `BaseFragment` fix can't reach them; each now registers a callback on its dialog's dispatcher. (Pre-existing gap: system back lost the filter selection / skipped the discard prompt.)
- **`MainActivity`** — removed the now-redundant `BackPressListener` dispatch (every such screen self-handles now); kept the analytics / app-bar / default-back behavior.

> [!WARNING]
> **Back-press analytics is only partially restored by this PR.** The `back_pressed` Tracks event is recorded for `BackPressListener` screens (the consumed path) but remains under-reported on other screens: since androidx.activity 1.12.4 (25.1) a `NavHostFragment`'s back callback consumes the gesture before MainActivity's tracker runs, so the event only fires at nav roots. Fully restoring it needs a nav-observer approach rather than the back dispatcher, and is tracked separately in WOOMOB-3614 to keep this PR focused on the data-loss fix.

### Test Steps

Android 16+ (predictive back active). For each: change a value, leave via **system/gesture back** (not the toolbar arrow), confirm the edit is kept.

1. Products → open a product → **Inventory** → change stock → back → product detail shows **Update/Save** + the new value.
2. Same product → **Price**, then **Description** (Aztec editor) → back → change kept.
4. Orders → **Filters** → change a order stats → back → change saved.
5. Regression: repeat the steps via the **toolbar back arrow** — still works.

Repeat the above steps on the tablet.

### Images/gif

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/ea2ffb0f-27f0-4ee4-a78f-e70d36072062" /> | <video src="https://github.com/user-attachments/assets/24bc6cca-6a86-4492-b916-6af56ac684a5" /> |
| <video src="https://github.com/user-attachments/assets/417a7099-dba1-4f35-8c30-83dc86ec0f5d" /> | <video src="https://github.com/user-attachments/assets/529c72b5-dace-4cc3-b136-d35dde59373f" /> |

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,11 @@
 -----
 
 
+25.1.1
+-----
+- [**] Fixed product and order edits (stock, price, description, notes, etc.) being discarded when leaving a screen with the system/gesture back button [https://github.com/woocommerce/woocommerce-android/pull/16267]
+
+
 25.1
 -----
 - [*] Fixed a crash when the Play Store age signals service disconnected during the age eligibility check [https://github.com/woocommerce/woocommerce-android/pull/16206]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/WooCommerce.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/WooCommerce.kt
@@ -1,6 +1,8 @@
 package com.woocommerce.android
 
 import android.app.Application
+import androidx.activity.ActivityFlags
+import androidx.activity.ExperimentalActivityApi
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
 import com.android.volley.VolleyLog
@@ -28,8 +30,15 @@ open class WooCommerce : Application(), HasAndroidInjector, Configuration.Provid
             .setWorkerFactory(workerFactory)
             .build()
 
+    @OptIn(ExperimentalActivityApi::class)
     override fun onCreate() {
         super.onCreate()
+
+        // Hotfix for WOOMOB-3609: androidx.activity 1.12 changed OnBackPressedDispatcher so a hosting
+        // NavHostFragment's back callback out-prioritizes MainActivity's, which stopped BackPressListener
+        // screens from saving/confirming on system back. Opt back into the previous dispatch order.
+        // Proper follow-up (drop this flag, adopt per-screen callbacks) is tracked in WOOMOB-3614.
+        ActivityFlags.isOnBackPressedLifecycleOrderMaintained = false
 
         // Stripe Tap to Pay library starts it's own process. That causes the crash:
         //  > Caused by: java.lang.IllegalStateException: Default FirebaseApp is not initialized in this process
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesFragment.kt
@@ -4,6 +4,8 @@ import android.os.Bundle
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
+import androidx.activity.OnBackPressedCallback
+import androidx.activity.OnBackPressedDispatcherOwner
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.core.view.WindowCompat
 import androidx.fragment.app.DialogFragment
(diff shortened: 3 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/300">25.1.1</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Thanks for working on this @AdamGrzybkowski 🙏🏼 

Changes look good as per de AndroidX `OnBackPressedCallback`. Tested it on Phone and tablets and everything works as expected. 

:shipit:

hichamboushaba (approved):
Thanks @AdamGrzybkowski for the fix, it works well, so I'm approving it.

But I want to suggest a smaller fix, unless it has any downsides that I didn't think of: The cause of the bug is that Activity 1.12 [changed](https://android.googlesource.com/platform/frameworks/support/+/81852a92e5047549ca6f94cc093477bd38db525b) lifecycle-aware back callbacks to be inserted immediately and enabled/disabled in place, preserving registration order. MainActivity currently registers before layout inflation, so the NavHost’s FragmentManager callback is added later and wins LIFO dispatch. We can revert to the previous behavior with one of two solutions:
1. Either by registering MainActivity’s callback after the NavHost is inflated, this restores the intended priority; MainActivity handles BackPressListener first and then redispatches to the NavHost for the normal pop, I mean something like this patch: 

```patch
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
@@ -340,7 +340,6 @@
         val bundle = if (SelectedSite.hasSelectedSiteId(this)) savedInstanceState else null
         super.onCreate(bundle)
-        setOnBackNavigationCallback()
         ChromeCustomTabUtils.registerForPartialTabUsage(this)

@@ -364,6 +363,7 @@
         animatorHelper.toolbarHeight = binding.collapsingToolbar.layoutParams.height

         val navHostFragment = supportFragmentManager.findFragmentById(R.id.nav_host_fragment_main) as NavHostFragment
+        setOnBackNavigationCallback()
         val graphInflater = navHostFragment.navController.navInflater
```

2. Or by simply calling `ActivityFlags.isOnBackPressedLifecycleOrderMaintained = false` to disable the new behavior, anywhere (maybe in the app's `onCreate`).

I tested both approaches, and both fix the issue for me.

hichamboushaba (approved):
Thanks @AdamGrzybkowski, good catch with the dialogs issue too.

PR #16272, opened 2026-07-16
[WOOMOB-3229] Cache edit and view context variation images separately

irfano:
### Description

This PR restores the variation images that #16245 unintentionally changed in the order creation picker and POS back to the trunk behavior, fixes the issues #16245 introduced, and fixes a variation image bug that exists on trunk.

WooCommerce renders a variation image in two contexts: edit context returns only the variation's own image, while view context returns the variation's own image, falling back to the parent product's image when the variation has none. #16245 switched the product screens to edit context but kept a single cached image, so the cache carried whichever context last wrote it. This PR stores the two values separately: `image` keeps the view value for the display surfaces, and the new `editContextImage` column keeps the variation's own image for the editing screens and the update request.

#16245 fixed the image removal, gave the editing screens their placeholder, and stopped the details screen from copying the image on save. This PR fixes what remained on that branch. In the examples below, a T-shirt product has a product image showing the white T-shirt, and its Blue variation has no image of its own, so the store shows the white T-shirt image for Blue.

1. This issue also exists on trunk. Updating Blue's stock by scanning its barcode, or updating it through the AI assistant, silently set the white T-shirt image as Blue's own image on the server. From that point on Blue stopped following the product image: when the merchant later changed the product image, Blue kept the old white one. Updates now send the image only when it actually changed.
2. Right after the merchant visited the variations screen, the order creation picker and POS listed Blue without any image until their next refresh. They now always show the white T-shirt image: these are view context screens, so a variation without its own image shows the parent product's image.
3. Saving Blue from the details screen sent the right data to the server, but the screen then showed the white T-shirt image again as if it were Blue's own, until the next refresh. The editing screens now stay correct after a save.

Outside the product editing flows everything keeps fetching with view context and showing the same images as trunk. No POS code is touched in this PR.

#### Saving no longer copies the parent image onto the variation
1. Create a variable product with a product image and a variation without its own image.
2. In the app, go to Products.
3. Open the variable product.
4. Tap Variations.
5. Open the variation.
6. Change the regular price.
7. Tap the update icon.
8. Open the variation in web admin.
9. Verify it still has no image of its own.

#### Placeholder on the editing screens
1. Go to Products.
2. Open the variable product.
3. Tap Variations.
4. Verify variations without their own image show a placeholder.
5. Open one of them.
6. Verify the details screen shows a placeholder.

#### Order creation picker keeps the product image after visiting the variations screen
Opening the variations screen first is what used to break the picker, so the first steps set that condition up.
1. Go to Products.
2. Open the variable product.
3. Tap Variations.
4. Go back.
5. Go to Orders.
6. Tap + to create an order.
7. Tap Add products.
8. Tap the variable product.
9. Verify variations without their own image show the product image.

#### POS shows the product image as before
POS reads the view context image, so it looks the same as trunk even after the variations screen writes the edit data. The first steps set that condition up.
1. Go to Products.
2. Open the variable product.
3. Tap Variations.
4. Go back.
5. Tap Point of Sale in the bottom bar.
6. Tap the variable product.
7. Verify variations without their own image show the product image.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/variations/VariationDetailViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/variations/VariationDetailViewModel.kt
@@ -195,7 +195,7 @@ class VariationDetailViewModel @Inject constructor(
 
     fun onAddImageButtonClicked() {
         AnalyticsTracker.track(PRODUCT_VARIATION_IMAGE_TAPPED)
-        val images = viewState.variation?.image?.let { listOf(it) } ?: emptyList()
+        val images = viewState.variation?.editContextImage?.let { listOf(it) } ?: emptyList()
         triggerEvent(ViewImageGallery(navArgs.remoteVariationId, images, showChooser = true))
     }
 
@@ -245,7 +245,7 @@ class VariationDetailViewModel @Inject constructor(
                     remoteVariationId = remoteVariationId ?: variation.remoteVariationId,
                     sku = sku ?: variation.sku,
                     globalUniqueId = globalUniqueId ?: variation.globalUniqueId,
-                    image = if (image != null) image.value else variation.image,
+                    editContextImage = if (image != null) image.value else variation.editContextImage,
                     regularPrice = regularPrice,
                     salePrice = salePrice,
                     saleEndDateGmt = saleEndDate,
@@ -324,7 +324,10 @@ class VariationDetailViewModel @Inject constructor(
                 showVariation(variation)
                 loadVariation(variation.remoteProductId, variation.remoteVariationId)
                 triggerEvent(Event.ShowSnackbar(string.variation_detail_update_product_success))
-            } else if (variation.image == null && result.error.type == ProductErrorType.INVALID_VARIATION_IMAGE_ID) {
+            } else if (
+                variation.editContextImage == null &&
+                result.error.type == ProductErrorType.INVALID_VARIATION_IMAGE_ID
+            ) {
                 triggerEvent(Event.ShowSnackbar(string.variation_detail_update_variation_image_error))
             } else if (result.error.canDisplayMessage) {
                 triggerEvent(ShowUpdateVariationError(result.error.message))
@@ -450,7 +453,7 @@ class VariationDetailViewModel @Inject constructor(
         mediaFileUploadHandler.observeSuccessfulUploads(navArgs.remoteVariationId)
             .onEach { media ->
                 viewState.variation?.let {
-                    val variation = it.copy(image = media.toAppModel())
+                    val variation = it.copy(editContextImage = media.toAppModel())
                     showVariation(variation)
                 }
             }
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
@@ -104,16 +104,19 @@ object ProductVariationMapper {
         }
         // The variation's own image lives in editContextImage; stored rows never fetched with
         // edit context only have the display image. An updated model without edit-context
-        // knowledge (null) never touches the image.
-        val storedImage = storedVariationModel.editContextImage ?: storedVariationModel.image
-        val updatedImage = updatedVariationModel.editContextImage
-        if (updatedImage != null && storedImage != updatedImage) {
-            body["image"] = if (updatedImage.isNotBlank()) {
-                updatedVariationModel.getEditContextImageModel()?.toJson() ?: ""
-            } else {
-                // A blank image means it was removed. WooCommerce deletes the variation image
+        // knowledge (null) never touches the image. Ids are compared because the stored json
+        // comes from the server while the draft rebuilds a smaller one for the same image.
+        val storedImageId = if (storedVariationModel.editContextImage != null) {
+            storedVariationModel.getEditContextImageModel()?.id
+        } else {
+            storedVariationModel.getImageModel()?.id
+        }
+        if (updatedVariationModel.editContextImage != null) {
+            val updatedImageModel = updatedVariationModel.getEditContextImageModel()
+            if (storedImageId != updatedImageModel?.id) {
+                // A missing image means it was removed. WooCommerce deletes the variation image
                 // when the image id is 0 (supported since WooCommerce 4.7).
-                WCProductImageModel(id = 0L).toJson()
+                body["image"] = (updatedImageModel ?: WCProductImageModel(id = 0L)).toJson()
             }
         }
         if (storedVariationModel.menuOrder != updatedVariationModel.menuOrder) {
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
@@ -102,9 +102,14 @@ object ProductVariationMapper {
         if (storedVariationModel.shippingClass != updatedVariationModel.shippingClass) {
             body["shipping_class"] = updatedVariationModel.shippingClass
         }
-        if (storedVariationModel.image != updatedVariationModel.image) {
-            body["image"] = if (updatedVariationModel.image.isNotBlank()) {
-                updatedVariationModel.getImageModel()?.toJson() ?: ""
+        // The variation's own image lives in editContextImage; stored rows never fetched with
+        // edit context only have the display image. An updated model without edit-context
+        // knowledge (null) never touches the image.
+        val storedImage = storedVariationModel.editContextImage ?: storedVariationModel.image
+        val updatedImage = updatedVariationModel.editContextImage
+        if (updatedImage != null && storedImage != updatedImage) {
+            body["image"] = if (updatedImage.isNotBlank()) {
+                updatedVariationModel.getEditContextImageModel()?.toJson() ?: ""
             } else {
                 // A blank image means it was removed. WooCommerce deletes the variation image
                 // when the image id is 0 (supported since WooCommerce 4.7).
--- a/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -1259,7 +1259,7 @@ class WCProductStore @Inject internal constructor(
                 val variation = if (context == "edit") {
                     listOf(result.variation).mergeEditContextResponse(site, remoteProductId).first()
                 } else {
-                    result.variation
+                    listOf(result.variation).mergeViewContextResponse(site, remoteProductId).first()
                 }
                 productVariationsDao.upsertProductVariation(variation)
                 OnVariationChanged().also {
@@ -1288,6 +1288,20 @@ class WCProductStore @Inject internal constructor(
                 image = storedImages[variation.remoteVariationId].orEmpty()
             )
         }
+    }
+
+    /**
+     * View-context responses return the display image. Keep the stored
+     * [WCProductVariationModel.editContextImage], which edit-context fetches fill.
+     */
+    private suspend fun List<WCProductVariationModel>.mergeViewContextResponse(
+        site: SiteModel,
+        remoteProductId: Long
+    ): List<WCProductVariationModel> {
+        if (isEmpty()) return this
+        val storedEditContextImages = productVariationsDao.getVariations(site.localId(), RemoteId(remoteProductId))
+            .associate { it.remoteVariationId to it.editContextImage }
+        return map { it.copy(editContextImage = storedEditContextImages[it.remoteVariationId]) }
     }
 
     private fun fetchProductSkuAvailability(payload: FetchProductSkuAvailabilityPayload) {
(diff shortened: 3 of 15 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/variations/VariationDetailViewModel.kt
```diff
@@ -450,7 +453,7 @@ class VariationDetailViewModel @Inject constructor(
         mediaFileUploadHandler.observeSuccessfulUploads(navArgs.remoteVariationId)
             .onEach { media ->
                 viewState.variation?.let {
-                    val variation = it.copy(image = media.toAppModel())
+                    val variation = it.copy(editContextImage = media.toAppModel())
```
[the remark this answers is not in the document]
irfano: 👍🏻 Fixed in a8ca73ca299e430f0d5f37f8b25201b2006400a1

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1256,15 +1256,54 @@ class WCProductStore @Inject internal constructor(
...
+     * Edit-context responses return the variation's own image. Store it as
+     * [WCProductVariationModel.editContextImage] and keep the stored display
+     * [WCProductVariationModel.image], which view-context responses fill.
+     */
+    private suspend fun List<WCProductVariationModel>.mergeEditContextResponse(
+        site: SiteModel,
+        remoteProductId: Long
+    ): List<WCProductVariationModel> {
+        if (isEmpty()) return this
+        val storedImages = productVariationsDao.getVariations(site.localId(), RemoteId(remoteProductId))
```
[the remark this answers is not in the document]
irfano: Fixed in c2b448fd32fa3b218e14d431ebd4c30c8c35cbab

irfano:
@samiuelson, can you please test the POS case in the test steps and confirm that the code changes don’t affect POS screens?

samiuelson:
> @samiuelson, can you please test the POS case in the test steps and confirm that the code changes don’t affect POS screens?

Hi, @irfano, I verified POS displays variation images as expected 👍

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1256,15 +1256,74 @@ class WCProductStore @Inject internal constructor(
...
             }
         }
     }
 
+    /**
+     * Edit-context responses return the variation's own image. Store it as
+     * [WCProductVariationModel.editContextImage] and keep the stored display
+     * [WCProductVariationModel.image], which view-context responses fill.
+     */
+    private suspend fun List<WCProductVariationModel>.mergeEditContextResponse(
```
hichamboushaba, line 1284: This only preserves the other image for the page being fetched. The first-page write deletes all cached variations, so when page 2 is fetched, its previous image is already gone and these rows are stored with image = "". The order picker can then show blank images from cache or while offline.
The same problem happens in reverse: a view-context fetch can remove editContextImage from later pages.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1256,15 +1256,74 @@ class WCProductStore @Inject internal constructor(
...
+    }
+
+    /**
+     * Single-variation variant of [mergeEditContextResponse] that reads only the stored row.
+     */
+    private suspend fun WCProductVariationModel.mergeEditContextResponse(
+        site: SiteModel
+    ): WCProductVariationModel {
+        val stored = productVariationsDao.getVariation(site.localId(), remoteProductId, remoteVariationId)
+        return copy(editContextImage = image, image = stored?.image.orEmpty())
```
hichamboushaba: When the cache is empty, this throws away a valid image from the response.
In edit context, `image` is the variation’s own image, so a non-empty value is also a valid display image.
Replacing it with an empty string means a variation with its own image has no display image until a view-context fetch happens.
Could we at least keep the incoming image when there is no cached view value? The list version above has the same problem.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1256,15 +1256,74 @@ class WCProductStore @Inject internal constructor(
...
+    /**
+     * View-context responses return the display image. Keep the stored
+     * [WCProductVariationModel.editContextImage], which edit-context fetches fill.
+     */
+    private suspend fun List<WCProductVariationModel>.mergeViewContextResponse(
+        site: SiteModel,
+        remoteProductId: Long
+    ): List<WCProductVariationModel> {
+        if (isEmpty()) return this
+        val storedEditContextImages = productVariationsDao.getVariations(site.localId(), RemoteId(remoteProductId))
```
hichamboushaba, line 1318: This is a read-modify-write across separate Room calls. If a view-context fetch and an edit-context fetch run at the same time, both can read the old rows before either writes. The last upsert then removes the fresh value written by the other fetch.
withDefaultContext does not serialize these operations, so the race is still possible.
Could we make the merge and upsert atomic, or avoid storing two values that need to be merged independently?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
```diff
@@ -102,9 +102,14 @@ object ProductVariationMapper {
...
         }
-        if (storedVariationModel.image != updatedVariationModel.image) {
-            body["image"] = if (updatedVariationModel.image.isNotBlank()) {
-                updatedVariationModel.getImageModel()?.toJson() ?: ""
+        // The variation's own image lives in editContextImage; stored rows never fetched with
+        // edit context only have the display image. An updated model without edit-context
+        // knowledge (null) never touches the image.
+        val storedImage = storedVariationModel.editContextImage ?: storedVariationModel.image
+        val updatedImage = updatedVariationModel.editContextImage
+        if (updatedImage != null && storedImage != updatedImage) {
```
hichamboushaba: Minor: this issue is not caused by the new changes, but just mentioning it: The cached image is the full JSON returned by the server, while the details screen converts it to `Product.Image` and rebuilds smaller JSON in `toDataModel()`. Because those strings have different fields, the same image ID can compare as changed and be included in an unrelated save.

hichamboushaba (commented):
Thanks @irfano for the work on this, I left few comments to check.

And I want to discuss the solution as a whole, IMO storing two columns that get updated on different paths comes with many issues. WDYT about a simpler approach: we save only the edit context IMAGE, and we fetch with `context = edit` in all paths.

Then when displaying, depending on the flow, we either show the own image, or we fallback to the parent image for view flows:

```
displayImage = variation.ownImage ?: parentProduct.coverImage
```

Where `ownImage` is the image we get with `context = edit`.

This removes the context merge completely, which also removes the pagination and cold-cache behavior issues. It also solves another issue, when the parent image changes, variations without their own image will also follow it automatically, while with the current approach, we'll show stale images.

irfano:
> WDYT about a simpler approach: we save only the edit context IMAGE, and we fetch with context = edit in all paths.

Thanks for the review! I considered this approach while working on the PR, but I don't think it works. My reasons:

- The screens that need the view image: the order creation variation picker (also used by coupons, Blaze and order list filters) and POS (barcode lookup and the non-catalog list path). Web admin renders all of them with view context, so the server does the fallback there.
- The context difference is already bigger than the image. In the variations endpoint `tax_class` also resolves to the parent's tax class in view while edit returns the raw value, and `downloads` is always included in edit but only for downloadable products in view. The app doesn't consume those two fields for variations today, so nothing breaks there, but it shows the pattern: with edit everywhere, every context dependent field becomes potential wrong data on view screens.
- The fallback rule lives in core and that area is changing right now: 10.9 shipped native [variation galleries](https://developer.woocommerce.com/2026/05/19/bringing-variation-galleries-into-core/). If the resolution logic changes on the server, we'd be maintaining a separate copy of it on the client, even though the endpoints already give us the resolved value.
- The stale image point flips for the picker: it force fetches variations on open, so the view response brings the fresh parent fallback exactly when we display it. The fallback approach reads the parent product from cache, and the picker doesn't refetch the product, so it keeps the old image until some product screen refetches it. Both are stale until a fetch, ours refetches at display time.

hichamboushaba:
> The context difference is already bigger than the image

This is an additional reason against the approach of this PR, then; it's either we are fine with editing context and then the fallback I suggested above is better, or we need two separate tables. Because otherwise we keep overwriting incompatible data on each fetch, and it becomes worse than the images. WDYT?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
irfano: Made them atomic with a Room transaction in 9427afe4e6da823b30a5312f6e265c6db5e3127d. Two overlapping fetches now read, merge and write one after the other.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
irfano: Fixed in 397f11c425f52c7f7fe311c9e3554ba63e0be35c: the diff now compares the parsed image ids instead of the json strings, so the format difference can't produce a phantom image send anymore.

irfano:
Stepping back, I think there are three options here:

1. **Your suggestion: use edit context everywhere and add the fallback logic to the client.** One image field, and the app decides what to show: the variation's own image if there is one, otherwise the parent product image. Simplest option: no extra column, no merging, and pagination stops being a problem because every fetch carries everything it needs. The cost: that rule belongs to core today, and once we copy it, a core-side change means the app shows the old rule until an update ships.

2. **What the PR does: two fields, each context writes its own.** The view context image comes only from view responses, the edit context image only from edit responses. We copy nothing from core: whatever the server decides to show is what we show. The cost is complexity: every write has to preserve the other field. Also, when the edit screen paginates, the first page clears the cache and later pages can't bring the view context images back, so those variations show placeholders on view screens until the next view fetch fills them. Screens fetch on open, so it heals itself. We could prevent this too, but that needs another layer of code on top.

3. **Two tables, one per context.** Each context replaces only its own data, and both problems above disappear. The cost is a schema migration, and it feels like more than this problem needs right now. If this area keeps getting more complex, it could become worth it.

So IMO the choice is mainly between 1 and 2: maintain core's rule on the client and drop the extra column, or keep the PR as is and accept the temporary empty view context images after an edit pagination. I'm torn between the two, wdyt @hichamboushaba?

hichamboushaba:
Thanks @irfano, great recap above.

Personally I would still lean towards option 1, I think the chances of Core changing the behavior of the fallback are very small (if not impossible even), because there is a single cover image for the parent product, and that what we can use anyway.

But this is just a personal opinion, and it's not a strong one, if you feel 2 is safer, let's improve the logic a bit and merge this.

irfano:
Before settling I tried option 1 on a branch and tested it with a product that has no product image but has gallery images. It surfaced a concrete blocker: the client can't reproduce core's rule even today. The payload doesn't say which image is the cover, `images` is the cover plus the gallery merged with no marker. For that product view context returns no image and web shows placeholders for its variations, while the client fallback shows the first gallery image. The rule needs data the API doesn't expose, so I'm staying with 2. I also applied your other suggestion in the meantime, so the cache is now correct sooner for variations that have their own image.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
irfano: The remaining gap here is a tradeoff we accept: after a paginated edit fetch replaces the first page, the later pages can't restore the view context images, so those variations show placeholders on view screens until the next view fetch. These screens fetch with view context when they open, so the images come back at that point.

I tried closing it by caching those images, but the logic got complex and still couldn't solve it 100%, so I didn't want to add that mechanism for a rare, self-healing issue: a temporary cache loss that only shows up on products with more than a page of variations, and after af999e1b46a2b35e9a2e24e462a76fc5337e562e only for the ones without their own image.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
irfano: You're right for the non-empty case: if the variation has its own image, that's exactly what view returns, so we can keep it. Done in af999e1b46a2b35e9a2e24e462a76fc5337e562e, and update and batch responses now go through the same merge.

hichamboushaba (approved):
Thanks @irfano for taking the time to fix the issues I raised, nice work.

> It surfaced a concrete blocker: the client can't reproduce core's rule even today. The payload doesn't say which image is the cover

This is a wider issue than this change actually, we already put "cover" badge on the first image of the gallery, which means we are already not matching Core. So this option 1 and its fallback Okayish for the time being, while we should aim for the root cause fix: expose the featured image of the product in the API. But going with option 2 is a good choice too.

irfano:
Thanks a lot for the detailed review, @hichamboushaba, it helped us catch important points. And you're right about the cover image, we already don't match core there. I opened WOOMOB-3670 for it, and noted there that we can revisit option 1 for this logic once that's fixed.

PR #16277, opened 2026-07-17
[WOOMOB-3618] Update Store design system color tokens

hichamboushaba:
Fixes WOOMOB-3618

### Description

Updates the Store Design System color foundations to the newer Figma token export and applies the semantic roles across existing Store components.

- Refreshes primitive and semantic light/dark values while preserving their alias relationships.
- Keeps Surface and Surface Bright distinct and adds the missing state/tint roles.
- Aligns Cells, Filter Chips, dividers, Page Header, Search, Segment, Sheet, tabs, tables, and toolbar surfaces/states.
- Updates the token documentation, component-catalog previews, and focused resource/component tests.
- Leaves the separate POS design system unchanged.

The supplied export is the baseline for token values. For Dark Primary Container tints, this PR applies the WOOMOB-3552-approved `#6D469C` base across 08/10/16/24; the checked-in export remains stale at 08/10/16. The published Figma library still resolves Secondary Container at 24%; this PR intentionally implements the newer export's 16% alias. The current live Badge examples were also checked at 50XIH5MmOf4xUYEkM6fAm6-fi-3163_4479.

Validation completed successfully: Store Design System unit tests, lint, token-schema checks, diff hygiene, and deterministic screenshot generation.

### Test Steps

1. Open the Store Design System foundation and component-catalog previews in Android Studio.
2. Render the Light and Dark variants and verify Surface and Surface Bright remain visibly distinct, especially for Cell, Search, Segment, Sheet, tabs, Table, and toolbar surfaces.
3. Verify Filled, Tonal, and Outlined buttons; resting and selected Filter Chips; disabled choice controls; and Search placeholder/action states use the expected semantic state/tint layers.
4. Compare the status Badge variants with Figma node 3163:4479, including distinct Warning and Caution treatments.
5. Render the Large Font and RTL catalog variants and confirm the token changes introduce no clipping or layout regressions.

### Images/gif

WOOMOB-3618-UI-Review


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooColorResourcesTest.kt
```diff
@@ -0,0 +1,56 @@
...
+class WooColorResourcesTest {
+    private val context = ApplicationProvider.getApplicationContext<android.content.Context>()
+
+    @Test
+    fun `given light mode, when colors load, then secondary container uses the 16 percent primary container tint`() {
+        val colors = loadWooColors(context = context, useDarkTheme = false)
+
+        assertThat(colors.container.secondaryContainer)
+            .isEqualTo(colors.tintLayers.primaryContainer.opacity16)
+        assertThat(colors.container.secondaryContainer)
```
irfano: This negative comparison is unnecessary since `isEqualTo` already ensures that `isNotEqualTo` will pass. The same applies to the next test.

docs/design-system/android-adapter.md
```diff
@@ -221,20 +223,29 @@ constants:
...
-`primaryContainer.opacity08`, `opacity10`, `opacity16`, and `opacity24`. The canonical Segmented
-Control binds its track to `primaryContainer.opacity10`; the other published components do not bind
-to this tint-layer family. Keep these as complete mode-aware colors rather than public alpha floats.
+The runtime `tintLayers` group exposes complete mode-aware colors under `primaryContainer`,
+`onSurface`, and `primary`, each with `opacity08`, `opacity10`, `opacity16`, and `opacity24`.
+
+- Segmented Control binds its track to `primaryContainer.opacity16`.
+- Dividers and subtle component boundaries bind to `onSurface.opacity16`.
+- Light `onSurface.opacity24` intentionally uses `#1E1E1E` as its RGB base, unlike the State Layer.
+- WOOMOB-3552 approves `#6D469C` as the dark Primary Container tint base for the complete family:
```
irfano: Could we remove this from the adapter and keep the exact WOOMOB-3552 override values only in `token-map.md`? Keeping them in both documents creates two sources of truth that can drift.

irfano (changes requested):
Image

This is the only issue I noticed. In dark mode, the search field container and its surrounding search shell use the same background color, so the field boundary is not visible in the catalog preview.

irfano (approved):
My findings are minor, so I don’t need to re-review. Pre-approving! 👍🏻

hichamboushaba:
Thanks @irfano for the review, nice find about the search field background, I missed it. The cause is that `Surface Bright` and `Surface Dim` seem to share the same color, the issue is from the source, so I shared it with Wagner 50XIH5MmOf4xUYEkM6fAm6-fi-3175_2352#1857883730, I won't change the colors for Android, trying to stay consistent with Figma to make adapting always easier.

I will fix the two other issues.

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/foundation/WooColorResourcesTest.kt
hichamboushaba: Fixed in 5898df9

docs/design-system/android-adapter.md
hichamboushaba: Fixed in c46a11c

PR #16278, opened 2026-07-17
[WOOMOB-3225] Update Products list stock automatically after order changes

AdamGrzybkowski:
### Description

Fixes WOOMOB-3225

The Products list showed **stale stock** after an order operation changed a product's stock server-side — taking payment, fulfilling/completing, changing status, refunding, or editing an order. None of those flows updated the local product cache, so the merchant had to pull-to-refresh (or open the product) to see the correct quantity.

This adds a small app-wide signal, `ProductStockChangedSignal`, that order flows use to tell the Products list *which* product IDs just had their stock changed. The list reacts by re-fetching **only those products** (a targeted `fetchProductListSynced` upsert — no cache wipe) and re-reading the cache, so the affected rows update **in place, on any page, without resetting pagination or scroll**, and without a manual refresh.

Design notes:
- The signal accumulates IDs in a `StateFlow<Set<Long>>` (deduped), so nothing is lost while the list isn't collecting or is busy processing a previous batch. The list clears IDs only after it has **successfully** refreshed them — a failed/empty fetch stays pending for retry.
- All emit points fire **after the server confirms** the change, so the follow-up product fetch sees the already-reduced stock:
  - **Cash payment** (`SelectPaymentMethodViewModel`) — emitted on the server-confirmed remote result. Because cash navigates back to the order list on the optimistic result, the update result is collected on the application scope (`@AppCoroutineScope`) so it survives the ViewModel being cleared.
  - **Card payment / order status change** (`OrderDetailViewModel`) — on payment-completed and on the remote-success of a status change.
  - **Refund** (`RefundSummaryViewModel`) — only the items actually refunded (`quantity > 0`).
  - **Order edit** (`OrderCreateEditViewModel`) — on each successful edit sync (status/item changes made via the toolbar edit "pencil").

### Test Steps

Set up a product with **Manage stock** enabled and a visible quantity, then verify the Products list reflects each change **without pull-to-refresh**:

1. Create an order containing that product → **Collect Payment → Cash** → confirm → open the **Products** tab. The quantity is reduced.
2. Same with **Collect Payment → Card** (card reader flow).
3. On an existing (not-yet-stock-reduced) order, tap the **status chip** and set status to **Completed** → Products list updates.
4. On an existing order, use the **toolbar edit (pencil)** to set status to **Completed** or change an item's quantity → Products list updates.
5. **Refund** one item (quantity > 0) of an order with **restock** enabled → that product's stock increases; other items in the order are not touched.
6. Pagination check: scroll the Products list to page 2+, keep it visible (e.g. tablet two-pane), complete an order for a product on page 2 → that row updates in place and your scroll position / loaded pages are preserved.

### Images/gif

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/d69e3301-648f-449e-ab4f-b7acf30a9b25"/> | <video src="https://github.com/user-attachments/assets/e5c8b3df-1514-41b9-b05d-fe87097efd3d"/> |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodViewModel.kt
```diff
@@ -437,10 +443,14 @@ class SelectPaymentMethodViewModel @Inject constructor(
...
             when (result) {
-                is WCOrderStore.UpdateOrderResult.OptimisticUpdateResult -> exitFlow()
+                is WCOrderStore.UpdateOrderResult.OptimisticUpdateResult ->
+                    withContext(dispatchers.main) { exitFlow() }
                 is WCOrderStore.UpdateOrderResult.RemoteUpdateResult -> {
                     if (result.event.isError) {
-                        handleUpdateOrderStatusError()
+                        withContext(dispatchers.main) { handleUpdateOrderStatusError() }
+                    } else {
+                        // Emit only after the server confirms so the products' fetch sees the reduced stock.
```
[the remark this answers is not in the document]
AdamGrzybkowski: Fixed in https://github.com/woocommerce/woocommerce-android/pull/16278/changes/fd6e3ca8d9eac5b074f10a5ae235002630d7734e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
```diff
@@ -113,6 +118,30 @@ class ProductListViewModel @Inject constructor(
...
+     * they are refreshed and cleared, so nothing is lost while the list isn't collecting or is busy.
+     */
+    private fun onProductStockChanged(productIds: Set<Long>) {
+        // Keep the ids pending while offline so they're retried once the set changes or the list is reopened.
+        if (!networkStatus.isConnected()) return
+        launch {
+            val refreshed = getProductsByIds(productIds.toList())
+            if (refreshed.isNotEmpty()) {
+                // Don't clobber active search results; the browse list re-reads the fresh cache when search closes.
+                if (!isSearching()) reloadProductsFromDb()
```
irfano, line 140: Is there a reason for search mode to behave differently here? We already have the fresh products in hand, so I think we could update the rows in both cases and drop the `isSearching()` check.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductStockChangedSignal.kt
```diff
@@ -0,0 +1,38 @@
...
+ * isn't collecting (e.g. it isn't open, or it's busy processing a previous batch) and repeated ids are deduped.
+ * The collector drains the set via [clearProcessed] once it has refreshed those products.
+ *
+ * Note: because pending state is presence-only, two stock changes for the *same* product within a single refresh
+ * round-trip are coalesced; the rare "second change lands and is cleared before it's fetched" case self-heals on the
+ * next stock change for that product or a manual refresh. Ids left pending while offline are retried the next time
+ * the set changes or a new [ProductListViewModel][ui.products.list.ProductListViewModel] instance starts collecting.
+ */
+@Singleton
+class ProductStockChangedSignal @Inject constructor() {
```
irfano: This makes us re-fetch the whole product, not just its stock, and some of the emit points fire when stock didn't change at all. Can we name it after what it does, something like `RefreshProductsSignal` with `notifyProductsChanged()`?

irfano (changes requested):
The `notifyStockChanged` calls live in the ViewModels, so every entry point needs its own call and it's easy to miss one. Three slip through today: swipe an order to complete it on the order list, create an order with the status set to Processing, or take a card reader payment. In all three, open the Products tab after and the stock is still stale. Can we emit from the layer that actually makes the server call, the repositories and the card reader controller, so the shared paths are covered in one place?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
AdamGrzybkowski: Two reasons: 
1. Search gets the data directly from the API response, not from DB - so we can reload in the same way.
2. You can't open orders or other screens from the search easily. 

So it's unlikely to hit the same issue there. That said, we can replace the individual row if you think it's worth it.

irfano: You're right, this is hard to hit in practice. Fine as is.

irfano (approved):
Thanks for the changes, Adam. LGTM! 👍🏻

PR #16281, opened 2026-07-20
[WOOMOB-3627] Show order currency in the product selector

malinajirka:
~Do not merge - there is an issue in WooPayments that stores the currency into session when it comes from Jetpack tunnel, breaking idempotency of the API.~ (The added gating for stores on Woo 11.1+ makes this safe to merge)

### Description

Fixes WOOMOB-3627

`onAddProductClicked()` built the `SelectItems` event without passing `orderCurrency`, so it took the null default and the product selector fell back to the store currency. Everything else was already wired up. This also fixes the duplicate `/products` fetch (the second one went out without `currency=`, overwriting the cached prices).

Note the *amount* on the added line item is still wrong for non-default-currency orders — separate root cause, tracked in WOOMOB-3630.

### Test Steps

1. On a USD store, open a EUR order → Edit → tap **+** in Products.
2. Prices should show `€` and EUR amounts.
3. New order (creation flow) should still show `$` and USD amounts.

### Images/gif

N/A

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #16282, opened 2026-07-20
[WOOMOB-3630] Send the order currency when creating or updating an order

malinajirka:
### Description

Fixes WOOMOB-3630

Stacked on #16281 — review that one first.

Adding a product to a non-default-currency order priced the line item in the store's base currency (a EUR order got the USD number verbatim). The app deliberately posts new line items without a price so the server stays the pricing authority, but WooPayments Multi-Currency forces the store currency and disables conversion for any REST request without a `currency` query param. So we now send it on order create/update.

Went with appending to the URL rather than adding query-param support to `executePutGsonRequest`/`executePostGsonRequest` — verified on device that the param survives the transport.

### Test Steps

Needs a multi-currency store and a product with a manually set price in both the store's base currency and a second one.

1. Open an existing order in the non-default currency → **Edit** → **+** → add that product.
2. The added line item should be priced from the product's price in the *order's* currency. Before this change it used the base-currency price.
3. Regression: start a new order on the same store and add the same product — it should still use the base-currency price.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka:
Superseded by #16284 (query-param support) and #16285 (the order-currency fix), which replace the URL-append approach.

PR #16283, opened 2026-07-20
Resolve unused dependency warnings in commons

irfano:
### Description

This fixes the three `:libs:commons` entries from the [dependency analysis alert](https://a8c.slack.com/archives/C8ZFBDQC9/p1784439678537379).

The report listed these:

```text
Advice for :libs:commons
Unused dependencies which should be removed:
  implementation libs.google.dagger.hilt.android.main
  testFixturesImplementation project(':libs:fluxc-plugin')
  testImplementation libs.androidx.arch.core.testing
```

Each one needed something different.

`testFixturesImplementation project(':libs:fluxc-plugin')` is really unused. Nothing in the test fixtures uses it, and the two modules that consume the fixtures already declare `fluxc-plugin` themselves. So I removed it.

`testImplementation libs.androidx.arch.core.testing` is not really unused. `BaseUnitTest` in the test fixtures uses `InstantTaskExecutorRule` from it, so the tests need it at runtime. If I remove it, the commons tests fail with `NoClassDefFoundError`. I changed it to `testRuntimeOnly` instead, which is what it actually is, and that is not reported as unused.

`implementation libs.google.dagger.hilt.android.main` was only needed by the Hilt Gradle plugin. Commons only uses the `hilt-core` annotations, so nothing points at `hilt-android` in the code, but the plugin does not let the module configure without it:

```text
The Hilt Android Gradle plugin is applied but no com.google.dagger:hilt-android dependency was found.
```

Since commons has no `@AndroidEntryPoint` classes, it does not need the plugin at all. So I removed the plugin and replaced the dependency with `hilt-core`.

There is one more unused dependency in `:libs:store-design-system`, but it appeared on trunk after this alert was sent, so it is out of scope here.

### Test Steps

1. Open the [dependency analysis alert](https://a8c.slack.com/archives/C8ZFBDQC9/p1784439678537379) and follow its instructions to find the latest scheduled build.
2. Open the `dependency analysis` job.
3. Download the `build-health-report.txt` artifact.
4. Confirm that under `Advice for :libs:commons` the report lists:
   ```text
   Unused dependencies which should be removed:
     implementation libs.google.dagger.hilt.android.main
     testFixturesImplementation project(':libs:fluxc-plugin')
     testImplementation libs.androidx.arch.core.testing
   ```
5. Check out this branch.
6. Run `./gradlew buildHealth --console=plain`.
7. Open `build/reports/dependency-analysis/build-health-report.txt`.
8. Confirm that `Advice for :libs:commons` no longer has an `Unused dependencies which should be removed` section.

### Images/gif

N/A

### DIFF
--- a/libs/commons/build.gradle
+++ b/libs/commons/build.gradle
@@ -67,5 +67,4 @@ dependencies {
     testFixturesCompileOnly(libs.junit)
     testFixturesCompileOnly(libs.mockito.kotlin)
     testFixturesCompileOnly(libs.androidx.arch.core.testing)
-    testFixturesImplementation(project(":libs:fluxc-plugin"))
 }
--- a/libs/commons/build.gradle
+++ b/libs/commons/build.gradle
@@ -68,3 +68,12 @@ dependencies {
     testFixturesCompileOnly(libs.mockito.kotlin)
     testFixturesCompileOnly(libs.androidx.arch.core.testing)
 }
+
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            // This dependency is required by the Hilt Gradle plugin, otherwise the build fails to configure.
+            exclude(libs.google.dagger.hilt.android.main.get().module.toString())
+        }
+    }
+}
--- a/libs/commons/build.gradle
+++ b/libs/commons/build.gradle
@@ -58,7 +58,7 @@ dependencies {
     testImplementation(libs.assertj.core)
     testImplementation(libs.junit)
     testImplementation(libs.mockito.kotlin)
-    testImplementation(libs.androidx.arch.core.testing)
+    testRuntimeOnly(libs.androidx.arch.core.testing)
     testImplementation(libs.androidx.lifecycle.runtime.testing)
 
     testFixturesImplementation(platform(libs.androidx.compose.bom))
--- a/libs/commons/build.gradle
+++ b/libs/commons/build.gradle
@@ -3,7 +3,6 @@ plugins {
     alias(libs.plugins.kotlin.compose)
     alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.ksp)
-    alias(libs.plugins.google.dagger.hilt)
     alias(libs.plugins.dependency.analysis)
 }
 
@@ -44,7 +43,7 @@ dependencies {
     implementation(libs.androidx.lifecycle.process)
     implementation(libs.androidx.preference.main)
     implementation(libs.eventbus.android)
-    implementation(libs.google.dagger.hilt.android.main)
+    implementation(libs.google.dagger.hilt.core)
     implementation(libs.androidx.compose.material3)
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.wordpress.utils)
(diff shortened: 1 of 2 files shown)


libs/commons/build.gradle
```diff
@@ -67,5 +67,13 @@ dependencies {
...
     testFixturesCompileOnly(libs.androidx.arch.core.testing)
-    testFixturesImplementation(project(":libs:fluxc-plugin"))
+}
+
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            // This dependency is required by the Hilt Gradle plugin, otherwise the build fails to configure.
+            exclude(libs.google.dagger.hilt.android.main.get().module.toString())
+        }
```
hichamboushaba: I think we can avoid adding an exclusion here. `:libs:commons` only uses annotations from `hilt-core` and does not appear to need the Hilt Android Gradle plugin or the full `hilt-android` dependency.

Could you test the following alternative, which removes the plugin and declares `hilt-core` directly?

```patch
diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
index 3f0926c53e0..d5c32a05c47 100644
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -224,6 +224,7 @@ google-dagger-android-processor = { group = "com.google.dagger", name = "dagger-
 google-dagger-android-support = { group = "com.google.dagger", name = "dagger-android-support", version.ref = "google-dagger" }
 google-dagger-compiler = { module = "com.google.dagger:dagger-compiler", version.ref = "google-dagger" }
 google-dagger-hilt-android-main = { group = "com.google.dagger", name = "hilt-android", version.ref = "google-dagger" }
+google-dagger-hilt-core = { group = "com.google.dagger", name = "hilt-core", version.ref = "google-dagger" }
 google-dagger-hilt-android-testing = { group = "com.google.dagger", name = "hilt-android-testing", version.ref = "google-dagger" }
 google-dagger-hilt-compiler = { group = "com.google.dagger", name = "hilt-compiler", version.ref = "google-dagger" }
 google-firebase-analytics = { group = "com.google.firebase", name = "firebase-analytics" }
diff --git a/libs/commons/build.gradle b/libs/commons/build.gradle
index 3d25054d433..e109c649e85 100644
--- a/libs/commons/build.gradle
+++ b/libs/commons/build.gradle
@@ -3,7 +3,6 @@ plugins {
     alias(libs.plugins.kotlin.compose)
     alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.ksp)
-    alias(libs.plugins.google.dagger.hilt)
     alias(libs.plugins.dependency.analysis)
 }
 
@@ -44,7 +43,7 @@ dependencies {
     implementation(libs.androidx.lifecycle.process)
     implementation(libs.androidx.preference.main)
     implementation(libs.eventbus.android)
-    implementation(libs.google.dagger.hilt.android.main)
+    implementation(libs.google.dagger.hilt.core)
     implementation(libs.androidx.compose.material3)
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.wordpress.utils)
@@ -68,12 +67,3 @@ dependencies {
     testFixturesCompileOnly(libs.mockito.kotlin)
     testFixturesCompileOnly(libs.androidx.arch.core.testing)
 }
-
-dependencyAnalysis {
-    issues {
-        onUnusedDependencies {
-            // This dependency is required by the Hilt Gradle plugin, otherwise the build fails to configure.
-            exclude(libs.google.dagger.hilt.android.main.get().module.toString())
-        }
-    }
-}
```

irfano: Nice, this works. Tests pass and the app builds fine. Fixed in f384172

hichamboushaba (approved):
Looks good, sorry for the delay, forgot to re-review this.

PR #16285, opened 2026-07-21
[WOOMOB-3630] Price added products in the order's currency

malinajirka:
### Description

Fixes WOOMOB-3630

Stacked on #16284 (query-param support) — review that first.

Adding a product to a non-default-currency order priced the line item in the store's base currency (a EUR order got the USD number verbatim). The app posts new line items without a price so the server stays the pricing authority, but WooPayments Multi-Currency forces the store currency and disables conversion for any REST request without a `currency` query param. So we now send it on order updates.

Verified on device on both networking transports (Jetpack tunnel and Application Passwords). Stores using Jetpack Tunnel without [this fix](https://github.com/woocommerce/woocommerce/pull/66816) load currency from `session` so requests without an explicit currency use the last provided currency which is wrong - fixed in https://github.com/woocommerce/woocommerce/pull/66816.

### Test Steps

**Note: Consider running tests on [the following pr branch](https://github.com/woocommerce/woocommerce-android/pull/16290).**

Prerequisites:
1. Multi-Currency enabled in WooPayments.
2. WooCommerce with [this fix](https://github.com/woocommerce/woocommerce/pull/66816) - let me know, and I can share zip.
----
1. Create an order on the web in non-default currency (I enabled currency switcher setting under `/wp-admin/admin.php?page=wc-settings&tab=wcpay_multi_currency`)
1. Open the non-default currency order in the app → tap **Edit** → tap **+** → add a product.
3. The line item should be priced from the product's price in the *order's* currency. Before this change it used the base-currency price. (This PR is not about the currency symbol but about the amount - e.g. a product might cost 9euros or 10 usd, before this PR, the UI said correctly EUR but used value 10 instead of 9)

Make sure to test this flow with Application password enabled and disabled - you can switch the toggle in Settings -> Experimental -> Application Password, given you are logged in to your .com account.

### Images/gif

https://github.com/user-attachments/assets/2c273b91-57d6-403c-b53f-4ddf6690a4f1

- [x] I have considered if this change warrants release notes and have added them to \`RELEASE-NOTES.txt\` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #16290, opened 2026-07-22
[WOOMOB-3630] Block editing orders in another currency below WooCommerce 11.1

malinajirka:
Part of WOOMOB-3630. Stacked on #16285 — review that first.

### Description

WooPayments Multi-Currency only honours a `currency` query parameter on requests WooCommerce recognises as REST, and before 11.1.0 `is_rest_api_request()` missed the `?rest_route=` form the Jetpack tunnel uses. #16285 added the parameter and gated it on WooCommerce 11.1+ for that reason.

That gate leaves a gap. On a store below 11.1, opening an order whose currency differs from the store's still lets the merchant edit it — but the app can't send the currency, so the store prices every line item it adds in the store's base currency. The edit looks like it worked and quietly writes wrong totals.

This makes those orders read-only instead. When the order currency doesn't match the store's and the store can't take the currency parameter, the edit form opens with `isEditable = false` — the same mechanism already used for orders whose status makes them non-editable — and the existing lock banner explains why:

> **This order can't be edited in the app**
> Sorry, you can only edit this order on the web, as it uses EUR, and your site's currency is USD.

Reusing `isEditable` rather than intercepting the Edit button means the whole screen is already wired for the read-only state — locked product list, disabled add buttons, disabled coupon/shipping/gift-card actions — with no new UI.

### Test Steps

Needs a multi-currency store (WooPayments) running WooCommerce **below** 11.1, and an order in a non-default currency.

1. Open the non-default-currency order → tap **Edit**.
2. The form should open locked: the banner reads "This order can't be edited in the app" with the two currency codes, the product list is locked and the add/coupon/shipping buttons are disabled.
3. Open an order in the store's own currency → tap **Edit**. It should be fully editable, with no banner.
4. Regression, existing behaviour: open a **completed** order in the store's currency → tap **Edit**. The banner should still read "Parts of this order are not currently editable / To edit Products or Payment Details, change the status to Pending Payment."
5. On a store running WooCommerce **11.1 or later**, the non-default-currency order from step 1 should be fully editable.

### Images/gif

Screenshot_1784716593

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -13,7 +13,7 @@
 - [*] Fixed the product selector showing store-currency prices when adding products to an order in a different currency [https://github.com/woocommerce/woocommerce-android/pull/16281]
 - [*] Removed the outdated store setup feedback survey [https://github.com/woocommerce/woocommerce-android/pull/16273]
 - [*] Fixed products added to an order in a non-default currency being priced in the store's base currency [https://github.com/woocommerce/woocommerce-android/pull/16285]
-- [*] Orders in a currency other than the store's are now read-only in the app on WooCommerce below 11.1, where their totals can't be calculated correctly
+- [*] Orders in a currency other than the store's are now read-only in the app on WooCommerce below 11.1, where their totals can't be calculated correctly [https://github.com/woocommerce/woocommerce-android/pull/16290]
 - [*] Fixed the store name incorrectly appearing in the toolbar when opening a review from a notification [https://github.com/woocommerce/woocommerce-android/pull/16264]
 - [*] In-Person Payments onboarding no longer hard-blocks when your Stripe account has overdue requirements; you can skip and continue taking payments [https://github.com/woocommerce/woocommerce-android/pull/16208]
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/296">25.3</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

samiuelson:
Hi @malinajirka, I have question regarding the following scenario:

> WooCommerce below 11.1, and an order in a non-default currency.

I was able to update an order in a non-default currency with new status, apply coupon and add shipping. Is this expected?

malinajirka:
Thanks @samiuelson! 

> I was able to update an order in a non-default currency with new status, apply coupon and add shipping. Is this expected?

Yes, that's expected. Since shipping uses absolute numbers entered by the user, they are not impacted by currently selected currency. Similarly status and coupons are not affected.

samiuelson (approved):
LGTM!

PR #16296, opened 2026-07-23
[WOOMOB-2979] Clarify SMS option in two-step login

JorgeMucientes:
### Description

Fixes WOOMOB-2979

The WordPress.com token response does not identify whether an account expects an authenticator or SMS code, so the previous authenticator-specific instructions were wrong for SMS-only accounts.

This updates both two-step login implementations to:

- use method-neutral verification instructions initially;
- make **Send code via text message** a prominent outlined action;
- show request progress, prevent duplicate requests, and switch to sent/resend copy after success;
- treat the legacy `needs_2fa` response from an explicit SMS request as confirmation that the code was requested, including when no phone-number suffix is returned.

SMS remains opt-in to avoid duplicate sends and unnecessary throttling.

### Test Steps

Use a WordPress.com account with SMS two-step authentication enabled, ideally with SMS as its only method.

1. Reach the two-step screen through the main WordPress.com sign-in flow.
2. Confirm the initial instructions do not mention an authenticator app.
3. Confirm **Send code via text message** is shown as a full-width outlined button with a message icon.
4. Tap it and confirm the request shows progress, an SMS arrives, the instructions confirm a text was sent, and the action changes to **Send another code via text message**.
5. Enter the SMS code and confirm login completes.
6. Repeat through the Jetpack connection/setup sign-in flow to cover the Compose implementation.
7. Regression-check an account using an authenticator code and confirm it can still log in without requesting SMS.

### Images/gif

| Login Flow  | Jetpack Connection Flow |
| <video src="https://github.com/user-attachments/assets/a344f625-c64b-4da8-8bb0-a1ae8cd7600f"/>   | <video src="https://github.com/user-attachments/assets/222e67bc-3397-4b45-b604-19af35757bcb"/> |


libs/login/src/main/res/drawable/ic_login_sms_24dp.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
JorgeMucientes, line 1: It's not very common in the app tu use outlined buttons with and icon, but I feel it provides better clarity. Let me know if you think is better to remove this for consistency.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLogin2FAViewModel.kt
```diff
@@ -56,20 +57,32 @@ class WPComLogin2FAViewModel @Inject constructor(
     private val errorMessage =
         savedStateHandle.getStateFlow(scope = viewModelScope, initialValue = 0, key = "error-message")
 
+    private val hasRequestedSms =
+        savedStateHandle.getStateFlow(scope = viewModelScope, initialValue = false, key = "has-requested-sms")
+
+    private val isRequestingSms = MutableStateFlow(false)
```
malinajirka, line 63: ❓ It seems there is a pre-existing bug in the compose version. The Java screen gates its SMS button on supportedAuthTypes (Login2FaFragment:212 — hidden unless the list is empty or contains PUSH), but the compose screen doesn't. Wdyt?


P.S. I guess this filtering is the reason I don't see the button on my account in the login flow?

libs/login/src/main/java/org/wordpress/android/login/Login2FaFragment.java
```diff
@@ -441,12 +440,18 @@ protected void endProgress() {
...
     private void handleAuthError(AuthenticationErrorType error, String errorMessage) {
+        handleAuthError(error, errorMessage, false);
+    }
+
+    private void handleAuthError(AuthenticationErrorType error, String errorMessage, boolean wasRequestingSms) {
         switch (error) {
             case INVALID_OTP:
                 show2FaError(getString(R.string.invalid_verification_code));
                 break;
             case NEEDS_2FA:
```
malinajirka, line 451: AI note - I don't think any action is needed.
📓 Worth being aware of: `NEEDS_2FA` isn't a dedicated "resend succeeded" signal — it's the same generic error type the token endpoint returns whenever an OTP is still needed. If a resend genuinely fails server-side but the account still needs 2FA, it comes back as the same `NEEDS_2FA` and is now treated as a successful send (previously this branch was a no-op, so the ambiguity was harmless). The Kotlin repo has the identical assumption in `WPComLoginRepository.kt`, so this isn't new to this PR — just flagging that the Java side now surfaces it as an affirmative UI state instead of ignoring it.

malinajirka (changes requested):
Thanks @JorgeMucientes.

1. One thing I'd suggest changing before merge is the button color on the compose screen - it's grey and looks like a disabled button.

2. For my account with SMS 2FA only, I don't see the button in the login flow - however, the SMS arrives. Is that expected behavior?

Image

3. Similarly to 2, in compose, the SMS arrives as soon as I open the screen. However, I do see the button - it says `Send` not `Resend` though.

Any ideas, how come my account behaves differently?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLogin2FAViewModel.kt
hichamboushaba: Actually, I think the bug could be in the XML screen, because the current OAauth `/token` endpoint we use has some limitations around 2FA, we can't identify which 2FA methods are supported for the account, unless it's `webauthn`.
Can you check what the API returns for your account?

Check p91TBi-bWA-p2 for more information on the issue, and a comparison with the newer endpoint that Calypso uses, and why we can't switch yet. I think we need to start considering migrating to the new endpoint.

The limitations are also related to the other comment below.

malinajirka: Thanks @hichamboushaba!

I must admit I'm a bit lost here. I just noticed that for Google SSO, we still use "/wp-login.php?locale=en_CZ" endpoint -> for my account:
```
{"success":true,"data":{"user_id":18....69,"two_step_supported_auth_types":["backup","sms"],"two_step_nonce_backup":"...42c6","two_step_nonce_sms":"...740","phone_number":"17","two_step_notification_sent":"sms"}}
```

For username + pwd login, the `/token` endpoint is used and returns:
```
{"error":"needs_2fa","error_description":"Please enter the verification code generated by your Authenticator mobile application."}
```
--------
From my very limited understanding, I agree that it seems the compose implementation is actually correct. However, I don't understand why we use different endpoints between username+pwd vs Google SSO, seems like a bug as well 🤔 😕 ❓

malinajirka: Seems like one of the bugs is that the app doesn't treat `SMS` in `two_step_supported_auth_types` as valid enum value. Wdyt? 
2. [fromString — no "sms" case, so it returns UNKNOWN](https://github.com/woocommerce/woocommerce-android/blob/8665e1cb8c21dd1770412f2ff6c98d83df7baa67/libs/login/src/main/java/org/wordpress/android/login/Login2FaFragment.java#L671-L684)
3. [UNKNOWN is discarded](https://github.com/woocommerce/woocommerce-android/blob/8665e1cb8c21dd1770412f2ff6c98d83df7baa67/libs/login/src/main/java/org/wordpress/android/login/Login2FaFragment.java#L650-L662)
4. [The gate that then hides the send SMS button](https://github.com/woocommerce/woocommerce-android/blob/8665e1cb8c21dd1770412f2ff6c98d83df7baa67/libs/login/src/main/java/org/wordpress/android/login/Login2FaFragment.java#L210-L214)

malinajirka: I should have mentioned one vital piece - the `Text me a code instead` is missing only on the Google SSO path, it's present on the `username + pwd` path.

JorgeMucientes:
Hey @malinajirka thanks for the feedback! 

I've addressed partially some of your feedback: 

> 1. One thing I'd suggest changing before merge is the button color on the compose screen - it's grey and looks like a disabled button.

Done! [82e61eb](https://github.com/woocommerce/woocommerce-android/pull/16296/commits/82e61ebb1cdc6383c86485b0ee94160d960d59e1)

> 2. For my account with SMS 2FA only, I don't see the button in the login flow - however, the SMS arrives. Is that expected behavior?

Hmm, the XML login flow shows the SMS button only when `supportedAuthTypes` is empty or contains `PUSH`. I wonder what the `/token` API is returning for you account as Hicham asked above. Would you mind sharing? 
In any case I've added the following commit [76b043d](https://github.com/woocommerce/woocommerce-android/pull/16296/commits/76b043d6862f2476ccef79d33dc25e8c41784d17) to ensure we display the button always. 

> 3. Similarly to 2, in compose, the SMS arrives as soon as I open the screen. However, I do see the button - it says Send not Resend though.

Yeah, the resend is only shown after you tap on the button. Perhaps we can update the copy and use "Resend" wording by default without waiting to tap on the button. Added that change: [b315869](https://github.com/woocommerce/woocommerce-android/pull/16296/commits/b3158697bf1b3a888907ed7de856daae07b67203)

malinajirka (commented):
I just noticed, that for some reason my replies to Hicham were set as `pending` 🤦

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/wpcom/WPComLogin2FAViewModel.kt
hichamboushaba: > However, I don't understand why we use different endpoints between username+pwd vs Google SSO, seems like a bug as well 🤔 😕 ❓

I'm not familiar with that part too, social login was [added](https://github.com/woocommerce/woocommerce-android/commit/1e4e2603447a111500f61a0d63593f087c38520c) in the login library since 2017, and it used `/wp-login.php` since then, maybe the OAuth flow (the `/token` endpoint) doesn't support social login.
Anyway, we know that the OAuth flow is kinda legacy, all the new changes regarding webauthn went to the `/wp-login.php`, so the goal should be to migrate the email/password flow to use that too, but it's outside of the scope of the current PR.

> Seems like one of the bugs is that the app doesn't treat SMS in two_step_supported_auth_types as valid enum value. Wdyt?

Nice find, the new `sms` value could have been added as part of the webauthn changes, and that's why the app didn't expect it before, I think the safest change now is to support it, and adjust 76b043d6862f2476ccef79d33dc25e8c41784d17 to make sure we handle both cases correctly:
- When `sms` is explicitly mentioned or the list is empty we show it.
- We hide it in the other cases.

WDYT?

malinajirka: Agreed! cc @JorgeMucientes

JorgeMucientes: Thanks both — that makes total sense. I’ve added explicit support for the sms auth type and updated the visibility logic so the SMS action is shown when sms is supported or when the list is empty, and hidden otherwise. This preserves the /token fallback while fixing the Google SSO case. Addressed in [cfc5ffc](https://github.com/woocommerce/woocommerce-android/pull/16296/commits/cfc5ffc6b93bb2bfffd99241dcfce336a032dcc6).

JorgeMucientes:

malinajirka (approved):

JorgeMucientes:

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #16297, opened 2026-07-23
[WOOMOB-3665] Add Products Compose UI foundation

hichamboushaba:
### Description
Part of WOOMOB-3665

Adds the Compose UI foundation for the Products list without changing the production route. [PR #16298](https://github.com/woocommerce/woocommerce-android/pull/16298) is stacked on this PR and performs the runtime cutover.

Figma reference: 50XIH5MmOf4xUYEkM6fAm6-fi-452_5318

#### Changes

- Adds a stateless `ProductListScreen` covering browsing, active search, sorting and filtering controls, loading and empty states, paging, selection visuals, accessibility semantics, and the Add Product FAB.
- Adds the shared UI capabilities needed by Products:
  - `WooActionChip` with button semantics.
  - Optional focus-request support for `WooSearchField`.
  - Optional pull-to-refresh enablement.
  - Configurable product-thumbnail corners.
- Simplifies `WooTabRow` to use Material 3 `PrimaryTabRow` for equal full-width tab allocation, RTL placement, and indicator movement while retaining Woo styling.
- Adds previews and stable semantics/test tags used by the existing Products E2E page object.

#### Scope

- `ProductListFragment`, `ProductListViewModel`, navigation, dialogs, analytics, and runtime behavior are unchanged in this PR.
- The new screen is not reachable in the app until PR #16298 is applied.
- The existing contextual selection mode remains unchanged; its Compose redesign will follow separately under WOOMOB-3665.

#### Verification

- `WooChoiceControlsTest` passes.
- App and Android-test Kotlin compilation pass.
- Wasabi debug assembly passes.
- Screenshot verification covers light/dark mode, RTL, 200% font scale, landscape, a narrow pane, active-search tab endpoints, and a three-tab design-system fixture.

### Test Steps

1. Render the `ProductListScreen` previews for browsing, active search, loading, empty, selection, large-font, dark-mode, and RTL states.
2. In active search, confirm the All Products and SKU tabs divide the available width equally and the indicator spans the selected half.
3. Confirm the header actions, sort/filter controls, product metadata, selection treatment, paging indicators, empty states, and Add Product actions render without overlap or clipping.
4. Use the interactive `WooTabs` and choice-control previews to confirm tab selection and action-chip behavior.

### Images/gif
Screenshot 2026-07-27 at 15 04 17

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #16298, opened 2026-07-23
[WOOMOB-3665] Cut over Products list to Compose

hichamboushaba:
### Description
Part of WOOMOB-3665

This PR builds on the merged [PR #16297 — Products Compose UI foundation](https://github.com/woocommerce/woocommerce-android/pull/16297) and contains the production cutover relative to that stateless UI boundary.

Figma reference: 50XIH5MmOf4xUYEkM6fAm6-fi-452_5318

#### Changes

- Makes the Compose screen the production Products list route.
- Retains `ProductListFragment` as a thin host for XML navigation, dialogs, snackbars, activity-shell behavior, and adaptive phone/tablet coordination.
- Adds a ViewModel-aware `ProductListScreen` overload that collects existing state and maps products into the stateless UI models from PR #16297.
- Moves selected product IDs into SavedState-backed ViewModel state while preserving loaded-only Select All and the existing bulk-action flows.
- Preserves existing behavior for:
  - Search threshold, debounce, and SKU scope.
  - Cached-first loading, refresh, and manual append paging.
  - Sorting, filtering, Add Product entry points, and upload state.
  - Bottom-navigation visibility, phone navigation, and tablet detail highlighting.
- Retains existing analytics event names and properties. Search tracking now records each accepted debounced query once, rather than every post-threshold keystroke.
- Migrates the Products E2E page-object selectors to Compose semantics.
- Removes obsolete main-list toolbar, search-tab, and sort/filter helpers, and simplifies the selection-only RecyclerView layout without changing its behavior.

#### Intentionally unchanged

- **AppCompat `ActionMode` remains the selection-mode presentation. Its Compose replacement will follow separately under WOOMOB-3665.**
- XML Navigation/SafeArgs, dialogs, snackbars, the adaptive two-pane shell, and the detail `FragmentContainerView` remain Fragment-owned.
- Phone product navigation uses the normal Fragment transition without a row shared element.
- `ProductSelectionListFragment` remains RecyclerView-based and continues using its dedicated selection layout and shared row assets.

#### Verification

- Focused `ProductListViewModelTest` coverage passes.
- App and Android-test Kotlin compilation pass.
- Wasabi debug assembly passes.
- The dependent cutover patch remains unchanged after each PR #16297 restack.

### Test Steps

1. On a phone with a populated store, open Products and verify the list, metadata, header controls, and Add Product action.
2. Exercise All Products and SKU search, sorting, filters, pull to refresh, and append paging; confirm results update without duplicate loads.
3. Open a product and return, then select multiple products. Confirm the existing contextual action bar and bulk actions work, selection survives a configuration change, and bottom navigation hides and restores correctly.
4. Verify empty catalog, empty search/filter results, offline refresh failure, and trash/undo feedback.
5. On a tablet or expanded window, verify list/detail panes, first-product behavior, selected-row highlighting, and back navigation.

### Images/gif
| Browsing | Search | Selection mode |
| --- | --- | --- |
| products-list-browsing-api36 | products-list-search-api36 | products-list-selection-api36 |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
@claude review

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListScreen.kt
```diff
@@ -54,9 +61,127 @@ import com.woocommerce.android.ui.compose.designsystem.icons.BarsFilter
...
+) {
+    val viewState by viewModel.viewStateLiveData.liveData.observeAsState(ProductListViewState())
+    val products by viewModel.productList.observeAsState(emptyList())
+    val selectedProductIds by viewModel.selectedProductIds.collectAsStateWithLifecycle()
+    val highlightedProductId by viewModel.selectedProductIdOnBigScreen.collectAsStateWithLifecycle()
+    val uploadProductIds by activeUploadProductIds.collectAsStateWithLifecycle(initialValue = emptySet())
+    val pullToRefreshEnabled by isPullToRefreshEnabled.collectAsStateWithLifecycle()
+    val isBarcodeScanningAvailable = remember(viewModel) { !viewModel.isSquarePluginActive() }
+    val context = LocalContext.current
+    val itemModels = remember(
```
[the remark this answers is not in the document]
hichamboushaba: Fixed and confirmed in the second review.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
```diff
@@ -260,19 +266,14 @@ class ProductListViewModel @Inject constructor(
     }
 
     fun onSearchRequested() {
-        val searchFilter = if (viewState.isSkuSearch) {
-            AnalyticsTracker.VALUE_SEARCH_SKU
-        } else {
-            AnalyticsTracker.VALUE_SEARCH_ALL
+        if (viewState.query.orEmpty().length > 2) {
```
irfano: This changes the submit behavior: searching a 1-2 char query with the keyboard search button used to work, now it does nothing. The typing path already has the threshold in `onSearchQueryChanged`, can we relax this check to `isNotEmpty()` so the keyboard search button still works for short queries?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListFragment.kt
```diff
@@ -160,89 +141,43 @@ class ProductListFragment :
...
-
-        productListToolbar.onViewCreated(this, productListViewModel, binding)
-    }
-
-    private fun addSelectionTracker() {
-        tracker = SelectionTracker.Builder(
-            "productSelection", // a string to identity our selection in the context of this fragment
-            binding.productsRecycler, // the RecyclerView where we will apply the tracker
-            ProductSelectionItemKeyProvider(binding.productsRecycler), // the source of selection keys
-            DefaultProductListItemLookup(binding.productsRecycler), // the source of information about recycler items
```
irfano, line 211: This was the only usage of `DefaultProductListItemLookup`. Can we delete the class in `SelectableProductListItemLookup.kt`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
```diff
@@ -558,9 +572,9 @@ class ProductListViewModel @Inject constructor(
                     skuSearchOptions == productRepository.lastIsSkuSearch
                 ) {
                     if (loadMore) {
-                        _productList.value = _productList.value.orEmpty() + products
+                        updateProductList(_productList.value.orEmpty() + products)
```
irfano: This append is unchanged from trunk, but the consequence changes here: the RecyclerView's DiffUtil tolerated a repeated product and just drew the row twice, while `items(key = remoteId)` throws on a duplicate key. Offset paging can return the same product on two pages when the catalog changes mid-search. Can we add a `distinctBy { it.remoteId }` here?

WooCommerce/src/main/res/layout/fragment_product_list.xml
```diff
@@ -7,115 +7,25 @@
...
         app:layout_constraintEnd_toStartOf="@+id/two_pane_layout_guideline"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@id/toolbar">
-
-        <androidx.constraintlayout.widget.ConstraintLayout
-            android:layout_width="match_parent"
-            android:layout_height="match_parent">
-
-            <!-- Product search type view -->
-            <com.woocommerce.android.ui.products.WCProductSearchTabView
```
irfano, line 35: Nothing shows this view anymore. Can we drop it here and delete `WCProductSearchTabView`?

WooCommerce/src/main/res/layout/fragment_product_list.xml
```diff
@@ -7,115 +7,25 @@
...
-                android:layout_width="match_parent"
-                android:layout_height="wrap_content"
-                android:visibility="gone"
-                app:layout_constraintEnd_toEndOf="parent"
-                app:layout_constraintStart_toStartOf="parent"
-                app:layout_constraintTop_toTopOf="parent"
-                tools:visibility="visible" />
-
-            <!-- Products filter and sort view -->
-            <com.woocommerce.android.ui.products.ProductSortAndFiltersCard
```
irfano, line 46: Same for this one. Can we remove it and delete `ProductSortAndFiltersCard`?

WooCommerce/src/main/res/layout/fragment_product_selection_list.xml
```diff
@@ -0,0 +1,139 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginBottom="@dimen/minor_00"
+                android:visibility="gone"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@+id/productsSearchTabView"
+                tools:visibility="visible" />
+
+            <androidx.compose.ui.platform.ComposeView
```
irfano: This screen is single-pane and has no FAB, so `blaze_banner_view`, `addProductButton`, and the two-pane guideline + `detail_nav_container` are dead here too. Can we drop them?

irfano (changes requested):
|Before|After| 
|-|-|
|Image|Image|
 
- `Note, no change needed` Small behavior change here: on trunk the sort and filter buttons stayed on screen during selection but were disabled, now the whole rail is hidden. Hiding them reads better than showing dead buttons, just noting it.
- The description says analytics are preserved, but `PRODUCT_LIST_SEARCHED` changed: it used to fire on every keystroke past the threshold with the partial query, now it fires once per executed search with the final query. The new behavior looks right, but the event volume will drop noticeably. Can we call that out in the description?

WooCommerce/src/main/res/layout/fragment_product_list.xml
hichamboushaba: Fixed in 6699f43

WooCommerce/src/main/res/layout/fragment_product_selection_list.xml
hichamboushaba: Fixed in 9a8078a

WooCommerce/src/main/res/layout/fragment_product_list.xml
hichamboushaba: Fixed in 0a42032

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
hichamboushaba: Fixed in 8a0db8a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListFragment.kt
hichamboushaba: Fixed in 84693bd

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListViewModel.kt
hichamboushaba: Fixed in 3d0421d

hichamboushaba:
Thanks for the review. I kept the sort/filter rail hidden during selection, as suggested. I also updated the PR description to clarify that search analytics now tracks each accepted debounced query once rather than every post-threshold keystroke.

hichamboushaba:
@irfano can you please check my last changes?

irfano (approved):
I forgot to re-review this. Thanks for the reminder. Everything looks good! 🟢

PR #16299, opened 2026-07-23
[WOOMOB-3303] Show warning when coupon does not apply to custom amounts

samiuelson:
WOOMOB-3303

### Description
Fixes WOOMOB-3303

In POS, coupons apply to the products in the cart but never to custom amounts (they are fee lines, which WooCommerce coupons ignore). This is easy to miss at checkout — the coupon row shows a discount, and nothing indicates the custom amount was excluded from it.

This PR adds an amber "Discount not applied" note to each custom amount row in the cart at checkout. Per the discussion in the Linear issue, it reuses the existing cart-item note pattern (same placement and typography as the green coupon discount / red invalid-coupon notes) with the design system's **alert** color, and the row's avatar is left unchanged.

**The note is driven by the order creation response.** The REST API has no per-fee discount field (fee lines only carry `total`/`total_tax`), so `WooPosTotalsViewModel` infers the verdict from the response: when the sum of per-item discounts (`subtotal − total`) accounts for the entire `coupon_lines` discount, the coupon provably discounted products only and did not touch fee lines. That verdict travels to the cart via the existing `OrderCreated` event (`WooPosOrderCreatedData.couponDiscountAppliedToItemsOnly`), so the note appears at the same moment the coupon row shows its computed discount. Consequences:
- No note before the order response arrives, and none if the coupon produced no discount (e.g. invalid coupon — the red coupon error already covers that case).
- If an extension ever does discount fee lines, the sums won't match and the note is suppressed rather than showing a false claim.

The cart additionally gates the stored verdict on "still at checkout + coupons still in cart", so the note disappears immediately on back-navigation or coupon removal without waiting for another order update.

### Test Steps
1. On a store with at least one active "off everything" coupon, open POS mode.
2. Add a product, a custom amount (e.g. £5.00), and the coupon to the cart. Verify the custom amount row shows no warning while the cart is editable.
3. Tap **Check out**. Once totals load, verify the custom amount row shows an amber "Discount not applied" note, the coupon row shows its discount, and the discount total reflects the products only (the custom amount is not discounted).
4. Tap back to return to the cart. Verify the note disappears.
5. Remove the coupon and check out again. Verify no note is shown.

### Images/gif
| Editable cart (no note) | Checkout (note shown) |
|---|---|
|  |  |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/296">25.3</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/totals/WooPosTotalsViewModel.kt
```diff
@@ -1032,17 +1032,32 @@ class WooPosTotalsViewModel @Inject constructor(
...
     }
 
+    // The API has no per-fee discount field, so this is inferred: when the sum of per-item
+    // discounts (subtotal - total) accounts for the entire coupon discount, nothing was
+    // applied to fee lines (custom amounts).
+    private fun isCouponDiscountAppliedToItemsOnly(
+        order: Order,
+        updatedCoupons: List<CouponInfo>,
+    ): Boolean {
+        val couponsDiscount = updatedCoupons.sumOf { it.discountAmount }
```
malinajirka: While this would likely work, I'm always worried about doing any client side calculations - including these, which don't impact the order. It's quite easy to make a mistake and end up with some rounding-error edge-cases.

Moreover, it seems we display the note even when the applied coupons is tied to a specific product. I thought we'd be displaying it only for `x% off everything` or `$x off` coupons.

What do you think about something like:
```
private fun hasWholeCartCoupon(cartCoupons: List<Coupon>): Boolean =
    cartCoupons.any { it.appliesToWholeCart() }

private fun Coupon.appliesToWholeCart() =
    type != Coupon.Type.FixedProduct &&
        productIds.isEmpty() &&
        categoryIds.isEmpty()
```

samiuelson: Good points, done in d6caa2c2e14 — the note is now driven by coupon metadata and only shows for whole-cart coupons.

Two small deviations from the snippet:
- Allowlisted `Percent`/`FixedCart` instead of `type != FixedProduct`, so plugin-defined custom types fail safe (no note).
- Kept the response gating: the note only appears once the whole-cart coupon actually produced a discount, so $0 / rejected coupons don't trigger it. No arithmetic left beyond "discount > 0".

Since the cart only carries `(id, code)`, the coupon metadata is resolved from the local DB via a new `WooPosTotalsRepository.getCouponsByIds()` — an uncached coupon just means no note.

malinajirka (approved):
Thanks @samiuelson, works as expected!

PR #16300, opened 2026-07-23
[WOOMOB-3667] Send a persistent device id in remote feature flag requests

irfano:
### Description

Fixes WOOMOB-3667

The app sends an empty string as `device_id` when it fetches remote feature flags. The endpoint expects a non-empty device id for the percentage rollout to work, so logins without a WordPress.com account are not spread across the rollout percentages. Logins with a WordPress.com account are unaffected.

This sends a persistent anonymous UUID instead, generated once and stored in undeletable preferences, matching what the [WordPress iOS app](https://github.com/wordpress-mobile/WordPress-iOS/blob/57bed95278c682ed2857fd76c4f28f66aec4ed23/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift#L17-L28) does.

Before, on trunk:

```
D FeatureFlagsRequest: https://public-api.wordpress.com/wpcom/v2/mobile/feature-flags/?build_number=&device_id=&identifier=&marketing_version=25.2-rc-2&platform=android&os_version=17
```

After, on this branch:

```
D FeatureFlagsRequest: https://public-api.wordpress.com/wpcom/v2/mobile/feature-flags/?build_number=&device_id=42bb1086-****-****-****-********c021&identifier=&marketing_version=25.2-rc-2&platform=android&os_version=17
```

### Test Steps
[log-feature-flags-request.patch](https://github.com/user-attachments/files/30312627/log-feature-flags-request.patch)
1. Log in.
2. Apply the logging patch.
3. Build and install the app.
4. Launch the app.
5. Filter logcat by the `FeatureFlagsRequest` tag and confirm `device_id` carries a UUID.
6. Relaunch the app and confirm the same UUID is sent.

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.3
 -----
+- [Internal] Send a persistent device id in remote feature flag requests so logins without a WordPress.com account are included in progressive rollouts [https://github.com/woocommerce/woocommerce-android/pull/16300]
 - [*] Improved product duplication to preserve product details managed by WooCommerce [https://github.com/woocommerce/woocommerce-android/pull/16262]
 - [*] Fixed the store login error message showing a placeholder instead of the status code [https://github.com/woocommerce/woocommerce-android/pull/16246]
 - [*] Fixed the order list not scrolling to the top after creating a new order [https://github.com/woocommerce/woocommerce-android/pull/16240]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefs.kt
@@ -234,7 +234,10 @@ object AppPrefs {
 
         IS_USER_AGE_ELIGIBLE_FOR_APP_USE,
 
-        QR_LOGIN_ROLLOUT_BUCKET
+        QR_LOGIN_ROLLOUT_BUCKET,
+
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Nice job @irfano. 
Changes look good to me and work as expected. :shipit: 

One thing that AI surfaced but I think it is not concerning as the current behaviour is the desired one and we probably want to keep the UUID across app reinstalls, but sharing it in case you think this might not be the desired behavior: 

> The Linear issue specifies a “per-install” identifier, but the UUID is stored in default preferences, which the [cloud-backup/device-transfer rules](https://github.com/woocommerce/woocommerce-android/blob/e5ceec707480fe1c4ca11fbcbd23ee4d32983b38/WooCommerce/src/main/res/xml/backup_scheme.xml#L3-L23) intentionally include. A restored installation—or two devices after transfer—can therefore share a rollout ID. This may be desirable for cohort continuity, but it is not strictly per-install

irfano:
Thanks, @JorgeMucientes! Yes, keeping the id across backup and transfer is what we want, the bucket stays stable for the same user. I've fixed the "per-install" wording in the Linear issue, that part was inaccurate.

PR #16308, opened 2026-07-24
Enable R8 optimizations for release builds

malinajirka:
Fixes #WOOMOB-3699

### Description
Release builds have been running R8 in shrink-only mode: `-dontoptimize`, R8 compat mode (`android.enableR8.fullMode=false`), no resource shrinking, and several broad keep-alls (`com.woocommerce.**`, fluxc, OkHttp, Gson, Zendesk, Guava) that exempted most of the app from shrinking.

This PR (one logical change per commit):
- Enables R8 optimization, full mode, and resource shrinking
- Narrows the `com.woocommerce.**` and fluxc keep-alls to `keepclassmembers` for fields/enums — all Gson needs, since we keep `-dontobfuscate` (names never change; only removal is a risk)
- Drops keep-alls that duplicate the libraries' own consumer rules (OkHttp, EventBus, Zendesk section, Guava)

Result on `vanillaRelease`: APK **139.4 → 127.0 MB**, dex **97.5 → 60.6 MB (−38%)**, and R8 can now optimize (inline/merge) across the app.

Notes for reviewers:
- Stripe Terminal SDK consumer rules are intentionally untouched (issue submitted [here](https://github.com/stripe/stripe-terminal-android/issues/739)).
- Remaining rules were sanity-checked with [Google's R8 configuration analyzer](https://developer.android.com/topic/performance/app-optimization/r8-configuration-analyzer).
- Verified on release builds with a logged-in store: dashboard stats, orders list/detail, products list/detail, app settings, Payments hub, and full POS flows including completed cash and simulated Tap to Pay card payments. Sentry mapping upload works (R8 emits a mapping even without obfuscation, for retracing inlined frames).
- I considered whitelisting packages instead of applying `com.woocommerce.** {<fields>; }`, but AFAICT the gain would be really small and we'd risk runtime failures across many features.

### Test Steps
Use a release build (e.g. `wasabiRelease`) with a logged-in store:
1. My Store shows stats and charts
2. Open an order from the order list; open a product from the products tab
3. POS: add a product to cart, check out, complete a cash payment
4. Open Help Center (Zendesk) and browse an article
5. Receive an order push notification

### Images/gif
N/A


WooCommerce/build.gradle
```diff
@@ -170,9 +170,10 @@ android {
 
     buildTypes {
         release {
-            // Proguard is used to shrink our apk, and reduce the number of methods in our final apk,
-            // but we don't obfuscate the bytecode.
+            // R8 shrinks and optimizes the app (dead code removal, inlining, class merging),
+            // but we don't obfuscate the bytecode (-dontobfuscate in proguard-rules.pro).
             minifyEnabled = true
+            shrinkResources = true
```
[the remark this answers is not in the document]
malinajirka: @claude not-applicable — the `country_mapping_*` strings don't exist in the unshrunk baseline APK either: all 241 were deleted by 5141ca4d674 ("Remove unused strings", Sept 2025), which static analysis drove precisely because the only reference is the dynamic `getIdentifier` lookup. `getCountryByCountryCode()` has returned null in production ever since, so resource shrinking changes nothing here. Pre-existing issue tracked in [WOOMOB-3701](https://linear.app/a8c/issue/WOOMOB-3701).

WooCommerce/proguard-rules.pro
```diff
@@ -38,12 +30,16 @@
 }
 ###### Event Bus 2 - end
 
-##### WooCommerce (this is needed for Json deserializers, but generally, we should keep our own classes) - begin
--keep class com.woocommerce.** { *; }
+##### WooCommerce - begin
+# Gson populates fields reflectively, so keep fields and enum constants of our classes
+# (we don't obfuscate, so names are stable). Methods stay eligible for R8 optimization.
+-keepclassmembers class com.woocommerce.** { <fields>; }
```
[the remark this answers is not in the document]
malinajirka: @claude rejected: verified empirically on the shrunk `wasabiRelease` mapping — the Gson models from all app-module call sites survive, including the worst-case pattern described here: `SitePlanRestClient$SitePlanDto` is reached only via `TypeToken<Map<Int, SitePlanDto>>`, yet is retained because the consuming code reads its properties (`it.currentPlan`). That's the general mechanism: any deserialized model whose data is actually consumed is statically referenced by the consuming code (property reads, checkcasts, signatures), so R8 keeps the class; a model with zero static references couldn't be used by the app at all. The release-notes entry carries the [*****] flag, so full smoke tests on the final APK cover the residual tail.

wzieba:
@malinajirka ack - but can this wait for a few days? I wanted to take a deep look but I just have some other things to finish.

malinajirka:
@wzieba Sure, no rush at all - feel free to look into it whenever. It's not blocking anything, it's just an improvement.

gradle.properties
```diff
@@ -9,7 +9,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 android.nonTransitiveRClass=true
-android.enableR8.fullMode=false
+android.enableR8.fullMode=true
```
wzieba, line 12: Nice, this had to be done before ~end of the year anyway 👏 

```
WARNING: The option setting 'android.enableR8.fullMode=false' is deprecated.
[task 2026-08-17T11:17:30.901+00:00] The current default is 'true'.
[task 2026-08-17T11:17:30.901+00:00] It will be removed in version 10.0 of the Android Gradle plugin.
```

WooCommerce/proguard-rules.pro
```diff
@@ -38,12 +30,16 @@
...
+-keepclassmembers enum com.woocommerce.** { *; }
 ##### WooCommerce - end
 
-###### FluxC (was needed for Json deserializers) - begin
--keep class org.wordpress.android.fluxc** { *; }
+###### FluxC (Gson deserialization; model/network field keeps come from fluxc's consumer-rules.pro) - begin
+-keepclassmembers class org.wordpress.android.fluxc.** { <fields>; }
+-keepclassmembers enum org.wordpress.android.fluxc.** { *; }
```
wzieba, line 42: > FluxC (Gson deserialization; model/network field keeps come from fluxc's consumer-rules.pro) - begin

True - can we move this to FluxC's `consumer-rules.pro` then? 

And generally, these rules are still pretty broad, but maybe it's a good iteration.

WooCommerce/proguard-rules.pro
```diff
@@ -1,24 +1,16 @@
...
 -dontwarn okhttp3.**
--keep class okhttp3.** { *; }
--keep interface okhttp3.** { *; }
 -dontwarn com.squareup.okhttp.**
--keep class com.squareup.okhttp.** { *; }
--keep interface com.squareup.okhttp.** { *; }
 
 -keepattributes Signature
 -keepattributes *Annotation*
 ###### OkHttp - end
```
wzieba, line 10: > OkHttp (the library ships its own consumer rules) - begin

That's [100% true](https://github.com/lysine-dev/okhttp/blob/main/docs/features/r8_proguard.md): can we remove all rules then?


```suggestion
```

WooCommerce/proguard-rules.pro
```diff
@@ -38,12 +30,16 @@
 }
```
wzieba, line 30: Image

I haven't dived into it, but have you maybe checked if we can remove this?


```suggestion
```

WooCommerce/proguard-rules.pro
wzieba, line 25: AFAIU we don't use Event Bus 2 - if these lines are required, can we move them to Event Bus 3 section?

wzieba, line 46: Normally I'd suggest it too to move to FluxC `consumer-rules.pro`, but WellSql should be relatively soon removed completely, so it's not worth it.

WooCommerce/proguard-rules.pro
```diff
@@ -54,20 +50,9 @@
 -dontwarn com.google.errorprone.annotations.*
 ###### Dagger - end
```
wzieba, line 51: I think we don't really need this, do we?

```suggestion
```

WooCommerce/proguard-rules.pro
wzieba, line 78: I think we no longer use Crypto Tink, do we? At least as a direct dependency. Have you experimented with removing these `-dontwarn`?

wzieba, line 84: As comments says, this was auto-generated by old AGP update - can we check if they're actually still required?

wzieba (commented):
Thanks @malinajirka for working on this! I've left a few comments. None of them are blocking, this PR is a good iteration towards fixing our R8 config.

wzieba:
@malinajirka I just realised: if the goal of the PR was to only enable R8 optimizations, I think it's good to go.

I reviewed it from "fix the R8 config" angle, which I see could be too broad and not actually the intention of this PR.

malinajirka:
Thanks for the review @wzieba!

> I just realised: if the goal of the PR was to only enable R8 optimizations, I think it's good to go.
> I reviewed it from "fix the R8 config" angle, which I see could be too broad and not actually the intention of this PR.

Yeah, I essentially tried to keep the changes to minimum, except of obviously wrong setup. However, I like all your suggestions and I think all of them will work - I recorded them under https://linear.app/a8c/issue/WOOMOB-3903/follow-up-proguard-rule-cleanups-suggested-in-r8-pr-review.

wzieba (approved):
Got it, thanks!

PR #16312, opened 2026-07-27
Add agent auto-login for debug builds

hichamboushaba:
### Description

This PR turns the existing debug auto-login entry into a reusable bootstrap for verification agents. The goal is to let a fresh debug install reach a configured test store without entering credentials through the UI, while keeping the feature out of production builds and avoiding credential-bearing deep links or ADB arguments.

#### Android login flow

- Moves `AutoLoginActivity` and its orchestration from `src/main` to `src/debug`, so release variants neither compile nor declare the feature.
- Replaces the old browsable auto-login URI with an explicit Activity protected by `android.permission.DUMP`. The Activity has no intent filter and rejects Intents containing data, extras, or clip data.
- Supports two explicit connection routes:
  - `WP_API` discovers the direct site with the existing `WPApiSiteRepository`, stores the supplied WordPress Application Password, performs an authenticated user request, and selects the site.
  - `WPCOM` authenticates through the production `WPComLoginRepository`, fetches the account through `AccountRepository`, then reuses `JetpackActivationRepository` to refresh and select the WordPress.com REST site.
- Treats an existing session as `ALREADY_ACTIVE` only when its connection type, selected site, normalized HTTPS host/path, and required token or Application Password state match the request. Other selected or mixed session state returns `CONFLICT`; the tool never signs out or silently replaces a session.
- Navigates to the existing `MainActivity` only after `SUCCESS` or `ALREADY_ACTIVE`, using the existing task when one is already running.

#### Host-to-app handoff

`tools/agent-login/agent-login.sh` reads a default or named JSON profile from the user's external config directory. It requires a regular mode-`0600` file, validates the selected device, Android user, package, and `run-as` support, then streams the profile bytes through ADB stdin into the debug app's private `noBackupFilesDir`.

The app consumes that request once and publishes a fixed, sanitized outcome such as `SUCCESS`, `ALREADY_ACTIVE`, `CONFLICT`, or `AUTH_FAILED`. The helper polls for that outcome, maps it to a deterministic exit code, and removes the fixed request/status files. The profile payload and password are never passed through command arguments, an Intent, a URI, `BuildConfig`, or the helper's machine-readable output.

The transport is intentionally simple and single-use: it has one request slot, one status slot, and no queue or concurrent-login support. It is intended for local debug builds and test accounts.

#### Agent guidance and coverage

The on-device verification skill now documents profile setup, flavor/package mapping, sanitized outcomes, session-preservation behavior, and manual-login fallback when no profile is configured. Focused JVM tests cover strict request parsing and redaction, request consumption/status publication, session routing, both login strategies, and site matching. The manual smoke steps below exercise the explicit Activity, private handoff, status polling, and MainActivity navigation together.

This remains happy-path development tooling. It assumes the configured target is connected to the account and the account has the expected role and eligibility. OTP-based 2FA is not automated; the helper reports `AUTH_REQUIRES_2FA` and directs the operator to use a dedicated revocable WordPress.com Application Password.

### Test Steps

1. Install a current `dev` debug build and start from a logged-out app.
2. Create the external default profile described in `.agents/skills/verify-on-device/references/agent-auto-login.md`, using test credentials, and set its mode to `0600`.
3. With one authorized device connected, run `tools/agent-login/agent-login.sh --flavor dev`. Add `--serial <serial> --user 0` when needed.
4. Confirm the script prints `SUCCESS` and the configured store dashboard opens.
5. Run the same command again and confirm it prints `ALREADY_ACTIVE` without adding another Main Activity to the back stack.
6. Repeat with `connection` set to `WP_API` and `WPCOM` to exercise both supported routes.

### Images/gif

N/A — this adds debug tooling and no new product UI.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
Hey @JorgeMucientes, can you check this PR and tell me what you think about the approach? It reuses the existing `AutoLoginActivity` that was added a while back, but makes it more secure and adds support for both [WordPress.com](http://wordpress.com/) and site credentials.

I’m asking you specifically because the smoke testing you’re working on also needs login, and I’m wondering whether it would be better to reuse the Maestro login scripts you added there instead of this approach. I mean specifically for agent automation through the `/verify-on-device` skill.

I still think the Maestro scripts are better for smoke testing, since they test login as a user would experience it, while this approach is more about getting agents logged in quickly. What do you think?

hichamboushaba:
@JorgeMucientes I'm marking this as ready for review, and assigning it to you, if you get a chance, please take a quick look, I think the important part would be just to validate the idea and generate architecture, for the code itself, it can be AI reviewed given it's for debug builds only.

WooCommerce/src/debug/kotlin/com/woocommerce/android/ui/login/auto/ApplicationPasswordAutoLoginStrategy.kt
```diff
@@ -0,0 +1,63 @@
...
+        } catch (cancellationException: CancellationException) {
+            throw cancellationException
+        } catch (_: Exception) {
+            null
+        }
+    }
+
+    private suspend fun authenticateAndSelect(persistedSite: SiteModel): Result {
+        val authenticatedUser = wpApiSiteRepository.checkIfUserIsEligible(persistedSite)
+        return if (authenticatedUser.isFailure) {
```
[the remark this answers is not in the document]
hichamboushaba: Yes, this is intented, and for two reasons:
- This is a dev tool, so we expect the accounts to be properly configured.
- Even in the case when an account is not configured well, the app will handle this gracefully and will show the correct eligibility error screen when MainActivity is launched.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/auto/AutoLoginActivity.kt
```diff
@@ -1,53 +0,0 @@
...
-        if (uri == null || !autoLoginHandler.isAutoLoginUri(uri)) {
-            finish()
-            return
-        }
-
-        autoLoginHandler.handleAutoLogin(
-            uri = uri,
-            scope = lifecycleScope,
-            onSuccess = {
-                registerDevice.kickoff(RegisterDevice.Trigger.LOGIN_SUCCESS)
```
JorgeMucientes, line 40: This line registering PN was dropped in the new AutoLoginActivity.kt. Is that intentional? Looks like registration would happen anyway upon restarting the app, however this could lead to devs thinking PN don't work unless the app is restarted.

JorgeMucientes (approved):
Hey @hichamboushaba thanks for working on this convenient way to log in! Everything worked as expected. 
As per your comment [here](https://github.com/woocommerce/woocommerce-android/pull/16312#issuecomment-5094772313). 

I agree with the separation you described. I wouldn’t reuse the Maestro login scripts for /verify-on-device, and I also wouldn’t make AutoLoginActivity the default authentication path for Maestro. 
The Maestro tests avoids repeated authentication already: login flows run first, with successful login last, and non-login flows reuse that session through ensure_logged_in script. They only fall back to UI login when the device has no valid session.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/auto/AutoLoginActivity.kt
hichamboushaba: Good catch — this was just an oversight in the AutoLoginActivity rewrite. I restored the original LOGIN_SUCCESS device registration in 1a638c3cdd0.

PR #16313, opened 2026-07-27
[WOOMOB-3635] Refresh order lists after order status changes

hichamboushaba:
### Description
Fixes WOOMOB-3635

After a full refund, the order detail flow performs a single-order refresh and stores the new Refunded status in the shared order row. The filtered order list, however, keeps its server-derived order-ID snapshot in the Paging list descriptor, so updating the shared row alone does not remove the order from the Completed list.

This change captures the cached order before inserting the fetched result. When the fetch succeeds and the status differs, it dispatches the existing site-scoped LIST_REQUIRES_REFRESH action so the server recomputes filtered list membership. It avoids an additional list refresh when the status is unchanged, the fetch fails, or the order was not previously cached.

Focused regression tests cover the successful Completed-to-Refunded transition and each of those negative cases.

### Test Steps
1. Launch the Wasabi debug app on a dedicated test or staging store. Do not use production or customer orders.
2. Open Orders and apply the Status: Completed filter.
3. Open a completed test order with refundable items.
4. Issue a full refund for all refundable items and confirm it.
5. Wait until the order detail shows Refunded and the refund success confirmation appears.
6. Return to the Completed list and verify the refunded order is removed without a manual refresh.
7. Pull to refresh and verify the refunded order remains absent.
8. Navigate away and back, or relaunch the app, and verify the refunded order remains absent from Completed.

### Images/gif
N/A


JorgeMucientes (approved):
Changes look good and work as expected.

JorgeMucientes:

PR #16319, opened 2026-07-28
[WOOMOB-3615] Handle back presses explicitly

JorgeMucientes:
### Description

Fixes WOOMOB-3615

AndroidX Activity 1.12 changed the ordering of lifecycle-aware back callbacks. This allowed the NavHostFragment to consume system and predictive back before BackPressListener screens that need to save changes or show a confirmation.

The WOOMOB-3609 hotfix in 25.1.1 (#16267) restored the old behavior globally with ActivityFlags.isOnBackPressedLifecycleOrderMaintained = false. This PR removes that temporary compatibility override and gives each owner a narrowly scoped role:

- A visible BackPressListener BaseFragment handles back first. Allowed presses continue through normal Navigation; consumed presses are tracked directly because they do not change a back stack.
- MainActivity no longer intercepts or redispatches back. BackPressTracker observes committed pops in the activity NavHost child FragmentManager, so ordinary Navigation retains predictive-back behavior and forward navigation is not tracked.
- On Android API 36+, a system-navigation observer also tracks back-to-home/activity-finish. On API 35 and below, that root-exit event is intentionally not tracked; fragment pops and consumed handlers remain covered.
- ProductListToolbarHelper is enabled only while it owns back: collapsing expanded search, popping nested tablet detail history, or restoring the list pane after a compact resize. It tracks those consumed actions directly; otherwise the activity NavHost owns Products back navigation.
- Order-filter dialogs keep using their dialog-owned dispatchers, and MainActivity continues updating its shared toolbar when a destination reaches RESUMED after navigation commits.

There is no intended user-facing behavior change from the 25.1.1 fix. This replaces its temporary global mechanism with explicit, tested ownership. No additional release note is needed because that fix is already documented.

### Test Steps

Use a phone or emulator with gesture navigation enabled (Android 16/API 36 recommended). With the current manifest configuration, predictive back animations are expected only on API 36+.

1. Open Products, select a product, open Inventory, change the stock quantity, then leave with system/gesture back. Verify the product detail shows the changed quantity and the Save/Update action.
2. Repeat for Price and Description. Verify each edit is preserved after system/gesture back, then repeat with the toolbar back action.
3. From an ordinary Navigation destination such as Order details, start a slow predictive-back gesture. Verify the previous screen preview is shown, cancelling stays on the detail, and committing returns to the list.
4. Open Orders > Filters > Order status, change the selection, and use system/gesture back. Verify the selection returns to the filter categories screen. Use back again and verify Keep changes stays in the filters while Discard exits without applying the changed selection.
5. In Products, expand search and press system/gesture back. Verify search collapses without leaving Products. With search collapsed, navigate into and back out of a product and verify ordinary predictive back is owned by Navigation.
6. On a tablet multi-pane layout, open a product and navigate into a nested detail screen. Verify system/gesture back pops the detail pane once and returns to the expected detail/list state without duplicate navigation or confirmation.
7. On a resizable emulator, open a product in a wide/two-pane layout, resize to compact while the detail pane is visible, make an unsaved edit, and press system/gesture back. Choose Discard and verify the product list pane is restored.
8. Clear Logcat and filter the WooCommerce-UTILS tag for Tracked: woocommerceandroid_back_pressed. Exercise one ordinary Navigation pop, one consumed editor/filter action, one Products search collapse, and one tablet detail-pane pop. Verify each action emits exactly one event with context: MainActivity.
9. From the activity navigation root, press back. On API 36+, verify the back-to-home predictive animation completes and emits one event. On API 35 and below, verify the app exits to the previous activity or Home normally; no predictive back-to-home animation or analytics event is expected for this root-exit case.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba (commented):
**Disclosure:** I used AI to assist with the investigation and to help draft this comment. I reviewed the conclusions and manually validated the two behaviors described below on API 36. I realize the complete solution may be non-trivial; I’m raising these points so we can discuss the right long-term direction.

---

Thanks for working on this. I tested the current branch on an API 36 phone and a Tablet. I think the screen-owned handling is going in the right direction, but the overall solution still needs another pass.

What I think we should achieve here is:

1. `BackPressListener` screens get the first chance to save changes or show confirmation.
2. Normal navigation is handled by the NavHost, so predictive back works.
3. `back_pressed` is tracked exactly once.
4. We can remove the temporary `ActivityFlags` workaround.

## BaseFragment handling

I think the callback in `BaseFragment` is needed. The activity extension cannot replace it because only the active screen knows whether it needs to save, confirm, or consume the action.

Registering it from `onResume()` also seems reasonable here. It makes the resumed `BackPressListener` the latest callback, so it handles back before the NavHost. I don’t think we need to introduce state-driven enabling across all existing listeners as part of this PR.

So I would keep this part of the approach:

```text
BackPressListener screen
        ↓
BaseFragment handles back first
        ↓
Save, confirm, or continue navigation
```

The `ProductNavigator` continuation also makes sense for the tablet/detail-root case.

## Predictive back

To clarify, this PR does not introduce the predictive-back issue. It already does not work correctly on `trunk`, because the temporary flag causes the existing always-enabled `MainActivity` callback to run before the NavHost.

The concern is that this PR makes the same behavior part of the permanent solution:

```text
Predictive gesture starts
        ↓
MainActivity callback owns the gesture
        ↓
NavHost receives no start/progress/cancel
        ↓
MainActivity redispatches after commit
        ↓
NavHost performs a normal pop without the preview
```

I confirmed this on API 36. When slowly swiping back from an ordinary destination, the previous screen is not shown during the gesture. Navigation only happens after releasing it.

`addBackNavigationCallbackAfterNavHost()` makes the ordering explicit, but it cannot transparently forward the predictive gesture. When it calls `onBackPressedDispatcher.onBackPressed()`, the NavHost has already missed the gesture progress.

This also matches Android’s official guidance. The [Predictive Back codelab](https://developer.android.com/codelabs/predictive-back) specifically recommends not intercepting back at the root activity, especially just for event logging, because this prevents the system’s predictive animations. More generally, the [callback best practices](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture#callback-best-practices) explain that an enabled `OnBackPressedCallback` takes ownership of back and prevents predictive animations unless the app handles the gesture itself. The guide also recommends not creating these callbacks only for business logic or logging.

Since this PR is meant to replace the temporary workaround with the correct fix, I think we should avoid keeping an activity-wide callback that intercepts every back action.

Could we keep the `BaseFragment` callback for the screens covered by the ticket, remove the always-enabled `MainActivity` callback, and let the NavHost own back everywhere else?

That would give us:

```text
BackPressListener screen
    → BaseFragment handles first
    → ticket behavior is preserved

Ordinary screen
    → no screen callback
    → NavHost owns the predictive gesture
```

`BackPressListener` screens would still not get the predictive preview while their custom callback is enabled, but that is expected when the screen needs to intercept the action. Normal screens would no longer lose predictive back because of a global callback.

## ProductListToolbarHelper

I think `ProductListToolbarHelper` should be checked with the same rule. Its callback is always enabled and handles both special UI states and ordinary navigation.

It makes sense for it to consume back when it needs to:

- Collapse expanded search.
- Coordinate a tablet pane.
- Restore the list pane after resizing.

But when it is only calling `navigateUp()` or `popBackStack()` for normal navigation, could that be left to the NavHost? Otherwise this callback can become another global interception point inside the Products flow and prevent predictive navigation there as well.

## Tracks event

I also confirmed that `woocommerceandroid_back_pressed` can be missing with the new callback order.

The intended behavior appears to be:

- `BaseFragment` tracks when the listener consumes back.
- If the listener allows it, it redispatches and `MainActivity` tracks it.

The problem is that the redispatched action does not necessarily reach `MainActivity`:

```text
System back
    ↓
BaseFragment listener allows it
    ↓
Redispatch
    ↓
ProductListToolbarHelper consumes it
    ↓
MainActivity is never reached
    ↓
No back_pressed event
```

This happened in the product callback path we tested. Navigation completed, but there was no Tracks event in Logcat.

On `trunk`, `MainActivity` tracks before delegating, so this missing-event behavior is new. With the PR, tracking depends on which callback happens to consume the redispatch.

## Suggested approach

My concrete suggestion would be:

1. Keep the screen-owned callback in `BaseFragment`, including its `onResume()` priority.

2. Remove the always-enabled `MainActivity` callback and `addBackNavigationCallbackAfterNavHost()`.

3. Let the NavHost handle ordinary navigation and predictive gestures.

4. Keep `ProductListToolbarHelper` responsible for actual search/pane behavior, but avoid using it for normal NavHost pops where possible.

5. Move `back_pressed` tracking out of the callback-order chain. The [official callback guidance](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture#callback-best-practices) suggests observing navigation through lifecycle or back-stack changes instead of adding a consuming callback only for logging:

   - Screen-owned handlers track when they consume the action without navigation.
   - Search or pane handlers track when they consume it themselves.
   - For fragment-to-fragment navigation, the guide suggests using fragment lifecycle state or `onBackStackChangeStarted()` / `onBackStackChangeCommitted()` from [`FragmentManager.OnBackStackChangedListener`](https://developer.android.com/reference/kotlin/androidx/fragment/app/FragmentManager.OnBackStackChangedListener).
   - On API 36, [`PRIORITY_SYSTEM_NAVIGATION_OBSERVER`](https://developer.android.com/reference/kotlin/android/window/OnBackInvokedDispatcher#PRIORITY_SYSTEM_NAVIGATION_OBSERVER) can observe system-handled navigation, such as finishing the activity or back-to-home, without consuming it. This would complement rather than replace fragment back-stack observation.
   - Use a dedicated tracker backed by `AnalyticsTrackerWrapper`.

The important part is that Tracks should observe the committed outcome. It should not require `MainActivity` to own every gesture.

JorgeMucientes:
Thanks for the detailed review @hichamboushaba. 

I've applied most of the suggested changes in the following commits:

- **Keep the screen-owned BaseFragment callback and its onResume() priority**
Unchanged from the original approach, as you suggested — screens that need to save, confirm, or consume back still own it first.

- **Remove the always-enabled activity callback and let the NavHost handle ordinary back navigation**  
  Implemented in [`8aee7e3d0c6`](https://github.com/woocommerce/woocommerce-android/commit/8aee7e3d0c6). `BackNavigationCallback`, its test, and the activity callback registration were removed, restoring NavHost-owned predictive back.

- **Keep `ProductListToolbarHelper` only for search and pane coordination**  
  Implemented in [`09c00b26ac5`](https://github.com/woocommerce/woocommerce-android/commit/09c00b26ac5). Its callback is enabled only for expanded search, tablet detail history, or pane restoration. These consumed actions track directly; normal NavHost navigation bypasses the helper.

- **Move analytics outside the callback-order chain**  
  Implemented in [`babb9a9aad4`](https://github.com/woocommerce/woocommerce-android/commit/babb9a9aad4):
  - Added a dedicated `BackPressTracker` backed by `AnalyticsTrackerWrapper`.
  - Fragment navigation is tracked from `onBackStackChangeCommitted(..., pop = true)`.
  - API 36+ uses `PRIORITY_SYSTEM_NAVIGATION_OBSERVER` for system-owned root exits.
  - Screen, search, and pane handlers still report actions they consume themselves.

- **Guarantee exactly-once tracking**  
  Added in [`70b63917fbf`](https://github.com/woocommerce/woocommerce-android/commit/70b63917fbf). Direct tracking and a resulting committed pop are coalesced when they belong to the same back operation.

One consequence of observing committed pops worth calling out: programmatic pops now also emit back_pressed — e.g. save-and-exit flows that call navigateUp() (coupon editing), and forward navigations that use popUpTo (feedback survey completion, store onboarding tasks). I think this is acceptable, the event effectively measures backward stack transitions now, but happy to suppress mixed pop+push batches in BackPressTracker if we'd rather keep the event closer to literal back presses.

Two intentional limitations remain: on API 35 and below, back-to-home from the navigation root isn't tracked (there's no non-consuming observer before PRIORITY_SYSTEM_NAVIGATION_OBSERVER in API 36), and screens with an enabled screen-owned callback don't get the predictive-back preview — which, as you noted, is expected when the screen needs to intercept the action.

I've manually re-tested the changes on API 36 and everything is looking good. Let me know if there's anything else you'd change 🙂

hichamboushaba (approved):
Nice work @JorgeMucientes, all scenarios work well, except one that I'll explain below, so I'm pre-approving.

The scenario that fails is: we send duplicate `back_pressed` events when we have discard changes dialogs, the first is sent for the user action, and the second is sent when tapping on discard.

I'll leave it to you to decide whether this is a blocker and requires fixing, or we are fine with it.

JorgeMucientes:
Thanks for the review @hichamboushaba 

I looked into fixing it without introducing too many changes and unexpectedly introduce any back navigation bugs, but it looks like it's trickier than expected. And I don't want to add extra complexity for just a minor tracking issue on an event that I think we've never checked and that tbh I'm not even sure if its a useful at all. 

So, I'm going to revert this attempt [5d74445](https://github.com/woocommerce/woocommerce-android/pull/16319/commits/5d74445398cdc0dc7cf16efea899eecaf5512cc3) of fixing that particular case of back pressed tracking duplication and accept we are fine with it.

PR #16322, opened 2026-07-28
[Design System] Feature Branch

hichamboushaba:
### Description

Aggregation branch for the store design system migration. Almost nothing here is new work for this PR: the migration was built and reviewed as 27 sub-PRs that targeted this branch instead of `trunk`, each merged after its own review. The combined diff is large (700 files, +26.1k/-12.8k) because it spans the whole migration, but the per-PR reviews are the record — see [all PRs that targeted this branch](https://github.com/woocommerce/woocommerce-android/pulls?q=is%3Apr+base%3Afeature%2Fstore-design-system-migration).

The first-parent history is those 27 sub-PR merges plus five `trunk` merges, and eight commits that landed directly on the branch. **Those eight, plus the conflict resolutions in the latest `trunk` merge, are what still needs review here:**

| Commit | What it does |
|---|---|
| b80e93bdd7e | Back-press handling in `ProductListFragment`. Merging trunk brought in WOOMOB-3615's two-layer back-press model, which duplicated a permission check this branch had independently added — so both layers ran. On the approved-discard path the second gate re-showed the discard dialog after the user had already confirmed it. Drops the duplicated gate and routes the consumed-back event through the activity-scoped `BackPressTracker` (same event and payload, but deduplicated within a main-loop turn). Restores the shape trunk used in `ProductListToolbarHelper`, the file this branch deleted when the product list toolbar moved to Compose. |
| 4e01945c94f | Updates the `verify-on-device` agent skill references to the Compose orders and products lists. Documentation only — no app code, no effect on the build. |
| 5e34e56f291 | Clears the 14 Android Lint findings code scanning attributed to this branch: `modifier` reordered to be the first defaulted parameter in four composables, a redundant nested `xmlns:app` removed from five layouts, and `button_outlined_bg_selector.xml` deleted now that the two layouts referencing it are gone. Mechanical; every call site of the reordered composables already used named arguments. |
| 2c6aae6c7a2 | Back navigation icon, which differed between the two toolbars: `WooDesignSystemToolbar` used Material's `ic_back_24dp`, `WooTopAppBar` used the design system's `AngleLeft` chevron. [Figma](https://www.figma.com/design/50XIH5MmOf4xUYEkM6fAm6/Mobile-Design-System?node-id=5060-9960&m=dev) specifies neither — it calls for FontAwesome Regular `arrow-left`, which the icon set was missing. Adds it in all three weights and repoints both toolbars, 45 layouts, and the component catalog's XML demo. It also covers the 25 Kotlin sites that set the navigation icon in code, starting with `AppBarStatus`: 27 of the 75 design system toolbar layouts declare no `navigationIcon` and fall back to that shared default, so changing only the layouts would have replaced one inconsistency with another. Those Kotlin sites need `designsystem.R` under an alias because `nonTransitiveRClass` keeps library resources out of the app module's `R` class, and in three fragments that alias sits beside a legacy theme import — so `MixedStoreThemeImportsRule` now ignores the design system `R` class, which carries no theme, while still flagging everything else under that package. |
| ca0a55b1e67 | Toolbar icon-button ripple, which was masked by a plain rectangle and so lit up a square on press. `applyToolbarIconTouchTarget` pins those buttons to a 48dp square, so a 24dp corner radius on the mask yields a circle — matching the rounded state layer in Figma and the Material3 `IconButton` the Compose toolbar already uses. |
| bd773dd5630 | Blaze icon container vanishing in the More menu in dark mode. Blaze is the one row backed by a brand asset, so it renders through `BrandIconContainer` rather than `WooIconContainer`, and that container used the `sectionVariant` background token — which has a night override, so the chip collapsed into the card and left the flame floating. The [More menu design](https://www.figma.com/design/0tv5aayHznMAVOoX7lfaeh/Mobile-More-Menu?node-id=1-3&p=f&m=dev) specifies the same container for every row in both themes, so it now uses the Sandstone shade the other rows get. |
| 49590013e35 | Adds the release note. One entry for the branch as a whole rather than per sub-PR, since none of the 27 shipped on its own. |
| 39ae768252d | Merges current `trunk` (155 commits). Git flagged one conflict, in `CardReaderModeScreen`'s import block, where trunk's rewrite of that screen met this branch's `WooThemeWithBackground` → `LegacyWooThemeWithBackground` rename; trunk's rewrite and its new `WCOutlinedButton` import are kept. Two further conflicts merged cleanly as text but not as code, because trunk added call sites for APIs this branch had changed: trunk's new intro-state preview in the same file, and in `WooPushNotificationsIntroductionScreen` both a new offline-mode preview using the removed theme and a new `OfflineModeError` branch calling `ErrorContent` without the `scrollState` argument this branch added to drive the toolbar divider. |
| c9dbb6172a1 | Moves this branch's two release notes into 25.8. Both sat under versions that had already been cut — the migration note under 25.7, and the product search fix under 25.6, where it landed while that version was open. Neither change is in `release/25.6` or `release/25.7`, so both ship with this PR. |

Two notes on what code scanning reports for this PR, so they don't get re-litigated in review:

- The "new alerts" headline overcounts. This branch has **fewer** open Android Lint alerts than `trunk` (3011 vs 3116), including `UsingMaterialAndMaterial3Libraries` dropping 830 → 715. The heuristic attributes pre-existing alerts that merely sit in touched files.
- The two remaining inline comments about `@string/selected` being a private androidx resource are pre-existing on `trunk` (3 alerts on both refs, same files) and only surfaced because line numbers shifted. Worth fixing, but not from this branch.

### Test Steps

A smoke pass over the migrated surfaces rather than a per-change check:

1. Install the branch build and sign in to a store with orders and products.
2. **My Store** — confirm the dashboard renders, and that the date-range header and card filter dropdowns open and mark the current selection.
3. **Orders** — list loads and paginates; search, filters, and barcode entry work; the create-order FAB works. Long-press a row to enter selection mode, select several, and bulk-update status.
4. **Products** — list loads; sort, filters, search (both All and SKU tabs), barcode scan, and add-product work. Open a product and check the gallery and the detail sections.
5. **Chrome** — top app bars and bottom navigation across all tabs, and back navigation out of each detail screen. In particular, on **Products**, edit a product, press back, and confirm the discard dialog appears once and dismisses the screen when confirmed.
6. **Back arrow and ripple** — open a product (Compose top app bar), then tap into Price from it (XML toolbar) and confirm both back arrows are the same glyph. Press and hold either one and confirm the ripple is a circle, not a square.
7. Repeat the main screens in dark mode, and check the **More** menu: every row's icon chip, Blaze included, should sit on the same light container.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooModalBottomSheetState</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooTooltipPositionProvider</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski (approved):
All the changes look good :shipit: 🤞 🤞 🤞

PR #16323, opened 2026-07-29
[WOOMOB-3704] Send a Mobile Status Report to Zendesk - Part 2

malinajirka:
### Description

Fixes WOOMOB-3704

Stacked on #16325, which adds the MSR itself with the App / Device / Connectivity sections. This PR adds more sections. 

Every heading states whether its values are app-wide or belong to the selected store, and store-scoped headings name the store, since a merchant with several stores gets one report covering all of them.

- **Notifications** (app-wide) — Play Services, permission, disabled channels, new-order sound status, FCM token (redacted), and background/power-save/data-saver restrictions. `GetBackgroundRestrictions` is extracted out of `BackgroundUpdatesDisabled`, which already computed those values and threw them away; analytics behaviour is unchanged.
- **Account & Stores** (app-wide) — WPCom user, the store count, and every connected site.
- **Store Details** (selected store) — Blog ID, Store ID, auth method, Jetpack state, plan, Woo version and push registration. The first two fail in opposite conditions: `blog_id` is `0` for every application-password site, which is exactly the population filing login tickets, while `store_id` is absent until the first successful `system_status` fetch.
- **Payments** (selected store) — WooPayments/Stripe install state from the plugin cache, plus the plugin actually driving in-person payments (`getCardReaderPreferredPlugin`), which is not derivable from install state when both are present.
- **Point of Sale** (selected store) — tab visible / launchable, and local-catalog sync timestamps. When a flag is false the report points at `application_log.txt` for the reason rather than recomputing it, because the use cases that determine it write those prefs as a side effect.
- **Feature flags and beta toggles** (app-wide) — every flag with its effective value and where that value came from: remote, compiled-in default, or debug override. An "enabled" list would be ambiguous between disabled, renamed and deleted, and the value alone doesn't say whether a rollout reached the device.

### Test Steps

Run tests on the last PR in this chain.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefsWrapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/AppPrefsWrapper.kt
@@ -53,6 +53,10 @@ open class AppPrefsWrapper @Inject constructor() {
 
     var jetpackAppPasswordsEnabled by AppPrefs::jetpackAppPasswordsEnabled
 
+    var isProductAddonsEnabled by AppPrefs::isProductAddonsEnabled
+
+    var wooPosLocalCatalogEnabled by AppPrefs::wooPosLocalCatalogEnabled
+
     var wooCorePushDeviceUUID by AppPrefs::wooCorePushDeviceUUID
 
     var remoteFeatureFlagsDeviceId by AppPrefs::remoteFeatureFlagsDeviceId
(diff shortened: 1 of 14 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson (approved):
LGTM!

PR #16325, opened 2026-07-29
[WOOMOB-3704] Send a Mobile Status Report to Zendesk - Part 1

malinajirka:
**Do not merge label - I'll merge all the PRs in this chain together.**

### Description

Fixes WOOMOB-3704

The server SSR we attach to support tickets carries no device or app information, and it is empty on login tickets. This adds a Mobile Status Report — the app-level counterpart — sent as Zendesk custom field `51884914117268` and as `mobile_status_report.txt`, the way application logs already work. Its own field rather than prepended to the SSR, so it survives the cases where the SSR fetch fails or is not available (before login).

This PR is the plumbing plus the App / Device / Connectivity sections. #16323 stacks on top and adds Notifications, Account & Stores, Payments and POS.

### Test Steps

Test the last PR in this chain.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
```diff
@@ -72,11 +74,14 @@ class ZendeskTicketRepository @Inject constructor(
...
 
+        val msr = mobileStatusProvider(selectedSite)
+
         val deviceLogs = envDataSource.getFullDeviceLogs()
 
         val attachmentTokens = listOfNotNull(
             diagnosticLog?.let { uploadTextAttachment(DIAGNOSTIC_LOG_FILENAME, it) },
-            uploadTextAttachment(APPLICATION_LOG_FILENAME, deviceLogs)
+            uploadTextAttachment(APPLICATION_LOG_FILENAME, deviceLogs),
+            uploadTextAttachment(MSR_FILENAME, msr)
```
samiuelson, line 84: These uploads are sequential. It would be better to run uploads concurrently with async/awaitAll.

samiuelson (approved):
LGTM!

samiuelson (approved):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
malinajirka: Good idea, logged under WOOMOB-3735

PR #16326, opened 2026-07-29
[WOOMOB-3636] Don't show the new order notification for orders paid on this device

irfano:
### Description

Fixes WOOMOB-3636

Collecting a payment or marking an order as paid fires a new order push notification on the same device that did it, so the merchant gets "You have a new order!" and the cha-ching for the sale they just rang up.

WooCommerce core queues the notification when an order moves into a paid status, listed in [NewOrderNotificationTrigger.php](https://github.com/woocommerce/woocommerce/blob/bc2fe4035c6066d44aa18ca49bcbbf1ad6999a45/plugins/woocommerce/src/Internal/PushNotifications/Triggers/NewOrderNotificationTrigger.php). Nothing in the app separates a transition it caused from one that happened elsewhere.

The app now records the site and order id when it moves an order into one of those statuses, and skips the tray notification when the matching push arrives. The order still syncs and shows up in the order list, only the tray notification is skipped. The record never leaves the device, so other devices on the store still get notified.

This PR covers the store flows. Card reader payments and POS follow in a stacked PR so the Peacock team reviews the part they can test.

### Test Steps

Use a new order for every scenario, WooCommerce sends at most one new order notification per order.

**Store cash payment**
1. Open the Orders tab.
2. Open an unpaid order.
3. Tap Collect Payment.
4. Tap Cash.
5. Confirm the payment.
6. Wait 30 seconds and verify no notification arrives.

**Order status changed from order detail**
1. Open the Orders tab.
2. Open an order whose status is Pending.
3. Tap the order status.
4. Pick Completed.
5. Wait 30 seconds and verify no notification arrives.

**Order status changed while editing an order**
1. Open the Orders tab.
2. Open an order whose status is Pending.
3. Tap Edit.
4. Tap the order status.
5. Pick Processing.
6. Tap Done.
7. Wait 30 seconds and verify no notification arrives.

**Regression: order paid somewhere else**
1. Open the store admin in a browser.
2. Open an order that is not paid yet.
3. Change its status to Processing and save.
4. Verify the new order notification arrives on the device as before.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/methodselection/SelectPaymentMethodViewModel.kt
```diff
@@ -435,12 +447,16 @@ class SelectPaymentMethodViewModel @Inject constructor(
...
     private suspend fun Flow<WCOrderStore.UpdateOrderResult>.handleOrderUpdateResultBeforeExit() {
-        collect { result ->
-            when (result) {
-                is WCOrderStore.UpdateOrderResult.OptimisticUpdateResult -> exitFlow()
-                is WCOrderStore.UpdateOrderResult.RemoteUpdateResult -> {
-                    if (result.event.isError) {
-                        handleUpdateOrderStatusError()
+        // exitFlow() on the optimistic result cancels the ViewModel scope, keep collecting so the
+        // remote result is still observed. FluxC finishes the request itself in NonCancellable.
+        withContext(NonCancellable) {
```
irfano, line 454: This screen exits on the optimistic result, which used to cancel the collection before the remote one arrived, so the recording never ran. Collecting under `NonCancellable` keeps the observer alive, FluxC already finishes the request itself.

Side effect worth knowing: the remote error branch right below now actually runs, so `payments_flow_failed` gets tracked in cases where it was silently skipped before. The snackbar goes nowhere since the screen is gone, but the event is real.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
```diff
@@ -0,0 +1,37 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * In-memory record of orders this device moved into a paid status, used to suppress the
+ * new order push notification triggered by that same action.
+ */
+@Singleton
+class MarkedAsPaidOrdersCache @Inject constructor() {
+    private val entries = Collections.synchronizedSet(mutableSetOf<Entry>())
```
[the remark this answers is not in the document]
irfano: Correct that entries stick around when a push never arrives, but the cost is small: about 80 bytes per entry including the set node, so even a heavy day of on-device payments is a few tens of KB, and it goes away with the process. The process outliving delivery actually works in our favour, if the app dies the cache dies with it.

There is no guaranteed upper bound on when the push lands, so a short TTL would start letting notifications through for orders paid on this device, which is the bug we are fixing. A longer window would not move the memory number enough to justify the extra state.

Leaving it unbounded on purpose.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/OrderDetailRepository.kt
```diff
@@ -132,7 +135,11 @@ class OrderDetailRepository @Inject constructor(
             orderId,
             selectedSite.get(),
             status
-        )
+        ).onEach { result ->
```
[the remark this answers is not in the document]
irfano: Good catch, both were untested. Added in eed0e0a: a success and an error case for each of the two producers, nothing beyond that.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditRepository.kt
```diff
@@ -81,7 +83,15 @@ class OrderCreateEditRepository @Inject constructor(
...
         return when {
             result.isError -> Result.failure(WooException(result.error))
-            else -> Result.success(orderMapper.toAppModel(result.model!!))
+            else -> {
+                val updatedOrder = orderMapper.toAppModel(result.model!!)
+                newOrderNotificationSuppressionCache.onOrderMovedToPaidStatus(
+                    siteId = selectedSite.get().siteId,
+                    orderId = updatedOrder.id,
+                    newStatusKey = updatedOrder.status.value,
+                )
```
hichamboushaba: We record the passage here regardless of the status of existing order, which means we will keep a record even if the order was already paid. This should be OK in most cases, but it might result in some notifications losses if for some reason the order was changed elsewhere.

WDYT about checking the current status before recording?

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * In-memory record of orders this device moved into a paid status, used to suppress the
+ * new order push notification triggered by that same action.
+ */
+@Singleton
+class NewOrderNotificationSuppressionCache @Inject constructor() {
+    private val entries = Collections.synchronizedSet(mutableSetOf<Entry>())
+
+    fun onOrderMovedToPaidStatus(siteId: Long, orderId: Long, newStatusKey: String) {
```
hichamboushaba: Could we rename this to `recordOrderMovedToNotifiableStatus`? “Paid” is slightly misleading because `NOTIFIABLE_STATUSES` also includes statuses such as `on-hold` and `pre-order`. “Notifiable” would align the method name with the rule it actually applies.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Collections
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * In-memory record of orders this device moved into a paid status, used to suppress the
+ * new order push notification triggered by that same action.
+ */
+@Singleton
+class NewOrderNotificationSuppressionCache @Inject constructor() {
```
hichamboushaba, line 13: Just a remark: with in-memory storage, this will be best-effort, not fixing the bug 100%, because if a notification was delayed, and arrived after the process was killed, it won't be skipped.

We can think of using DataStore to keep the records on disk, did you consider this? I'm not considering this a blocker, just discussing it.

hichamboushaba (approved):
Nice work @irfano, I'm pre-approving, I left some comments for discussion.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
irfano: Good suggestion, renamed in 9ad3d07.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/OrderCreateEditRepository.kt
irfano: Good point, done in ddfb92c.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
irfano: I did consider it, I wanted the lightest solution here. DataStore would mean a new storage layer, an expiration mechanism for the records, and a blocking read in NotificationMessageHandler, which felt like overkill for this bug. Since the push almost always arrives seconds after the status change, the notification getting delayed and the process dying in the meantime felt like an edge case I could accept.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/push/NewOrderNotificationSuppressionCache.kt
irfano: Small follow-up: it ended up as `onOrderStatusChanged` in e928783. The method gets every status change, not only the ones landing on a notifiable status.

PR #16328, opened 2026-07-29
[WOOMOB-3636] Don't show the new order notification for card reader and POS payments

irfano:
### Description

Fixes WOOMOB-3636

Stacked on #16326, which adds the suppression and covers cash payments and status changes in store management.

This PR adds the same recording for card reader payments and for the POS cash and mark as paid flows. `CardReaderPaymentController` is the single path for card payments, in POS and in store management, so both are covered here and both are in the test steps.

Release notes are in #16326, the user facing behaviour is the same one.

### Test Steps

Use a new order for every scenario, WooCommerce sends at most one new order notification per order.

**POS cash payment**
1. Open POS.
2. Add a product to the cart.
3. Tap Checkout.
4. Tap Cash.
5. Confirm the payment.
6. Wait 30 seconds and verify no notification arrives.

**POS mark as paid**
1. Open POS.
2. Add a product to the cart.
3. Tap Checkout.
4. Tap the mark as paid option.
5. Confirm.
6. Wait 30 seconds and verify no notification arrives.

**POS card reader payment**
1. Open POS.
2. Add a product to the cart.
3. Tap Checkout.
4. Collect the payment with a card reader.
5. Wait 30 seconds and verify no notification arrives.

**Store card reader payment**
1. Open the Orders tab.
2. Open an unpaid order.
3. Tap Collect Payment.
4. Pick the card reader.
5. Complete the payment.
6. Wait 30 seconds and verify no notification arrives.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
@@ -85,7 +85,6 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.CoreOrderStatus
 import org.wordpress.android.fluxc.store.WooCommerceStore
 import kotlin.reflect.KMutableProperty0
 
@@ -469,10 +468,9 @@ class CardReaderPaymentController(
     ) {
         paymentReceiptHelper.storeReceiptUrl(orderId, paymentStatus.receiptUrl)
         appPrefs.setCardReaderSuccessfulPaymentTime()
-        newOrderNotificationSuppressionCache.onOrderMovedToPaidStatus(
+        newOrderNotificationSuppressionCache.onOrderPaidRemotely(
             siteId = selectedSite.get().siteId,
             orderId = orderId,
-            newStatusKey = CoreOrderStatus.PROCESSING.value,
         )
 
         triggerEvent(CardReaderPaymentEvent.PlaySuccessfulPaymentSound)
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
@@ -42,7 +42,7 @@ import com.woocommerce.android.cardreader.payments.RefundConfig
 import com.woocommerce.android.cardreader.payments.RefundParams
 import com.woocommerce.android.cardreader.payments.StatementDescriptor
 import com.woocommerce.android.model.Order
-import com.woocommerce.android.notifications.push.MarkedAsPaidOrdersCache
+import com.woocommerce.android.notifications.push.NewOrderNotificationSuppressionCache
 import com.woocommerce.android.tools.SelectedSite
 import com.woocommerce.android.ui.orders.details.OrderDetailRepository
 import com.woocommerce.android.ui.payments.cardreader.onboarding.CardReaderFlowParam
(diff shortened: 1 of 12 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/cashpayment/WooPosCashPaymentRepositoryTest.kt
```diff
@@ -157,6 +161,43 @@
...
+    }
+
+    @Test
+    fun `given order completion succeeds, when completeOrder, then order is recorded as paid`() = runTest {
+        // GIVEN
+        val orderId = 123L
+        val siteId = 999L
+        val site: SiteModel = mock { on { this.siteId }.thenReturn(siteId) }
+        val gatewayTitle = "Pay in Person"
+        val codGateway: WCGatewayModel = mock { on { title }.thenReturn(gatewayTitle) }
```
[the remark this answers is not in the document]
irfano: Fixed in a03c3da. Dropped the gateway mock entirely, returning null from getGateway makes the repository fall back to the default title, which is what the test wanted anyway.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/cashpayment/WooPosCashPaymentRepositoryTest.kt
```diff
@@ -157,6 +161,43 @@
...
+        val statusModel = WCOrderStatusModel(statusKey = Order.Status.Completed.value)
+        val updateResult = UpdateOrderResult.RemoteUpdateResult(mock { on { isError }.thenReturn(false) })
```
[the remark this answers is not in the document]
irfano: Fixed in a03c3da, using a real OnOrderChanged() instead.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
```diff
@@ -110,6 +112,7 @@ class CardReaderPaymentController(
     private val paymentReceiptHelper: PaymentReceiptHelper,
     private val cardReaderOnboardingChecker: CardReaderOnboardingChecker,
     private val paymentReceiptShare: PaymentReceiptShare,
+    private val markedAsPaidOrdersCache: MarkedAsPaidOrdersCache,
```
malinajirka: 📓 I personally find the name `MarkedAsPaidOrdersCache` quite ambiguous/unclear. We use the term "Mark as paid" throughout the app as switching the order into `Completed` after the customer accepts cash (or other form of payment) and the app has no way to verify. 

However, here, it's used with a completely different meaning in a completely different context. I'd personally opt for a more explicit name, perhaps something like `NewOrderNotificationSuppressionCache`. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
```diff
@@ -466,6 +469,11 @@ class CardReaderPaymentController(
     ) {
         paymentReceiptHelper.storeReceiptUrl(orderId, paymentStatus.receiptUrl)
         appPrefs.setCardReaderSuccessfulPaymentTime()
+        markedAsPaidOrdersCache.onOrderMovedToPaidStatus(
+            siteId = selectedSite.get().siteId,
+            orderId = orderId,
+            newStatusKey = CoreOrderStatus.PROCESSING.value,
```
malinajirka: ❓ Could you please elaborate on why do we switch to Processing here and not completed? 

As far as I can tell the app doesn't set the status in this flow at all — the backend does it while capturing the payment, and where it lands depends on the products (e.g. virtual/downloadable orders usually go straight to Completed) and on the store's configuration. So we're hardcoding an assumption about servers  behaviour - not mentioning it's configuration dependant.

Do we need to pass the key at all here? I get why it's needed in OrderCreateEditRepository (it fires on every order update), but in this flow we unconditionally want to remember the order. Perhaps a dedicated onOrderPaidRemotely(siteId, orderId) without the status gate be clearer, wdyt?

malinajirka (approved):
Thanks! Behaves as expected, I've shared some notes about the code. Let me know what you think.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
irfano: Agreed. Renamed to `NewOrderNotificationSuppressionCache` in 7b57c58f.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
irfano: Good point! Setting PROCESSING is currently harmless, but no need to keep that assumption here. Applied your suggestion: ae3c827

PR #16331, opened 2026-07-30
[WOOMOB-3704] Send a Mobile Status Report to Zendesk - Part 3

malinajirka:
### Description

Fixes WOOMOB-3704

Stacked on #16323. Makes the report readable to someone who did not write it, without paying for that in length on every ticket.

`docs/mobile-status-report.md` documents every field — what it means, the values it can hold, and what each value implies for triage. The report links to it once from its header rather than carrying a per-field gloss inline, because it is attached to every ticket and is also shown to merchants in Help & Support, where a doubled length costs more than it explains.

The values that actively misled now explain themselves instead: `Blog ID` and `Store ID` each say why they are unset, `Remote values loaded: false` says no remote fetch has ever succeeded on the install, and a sideloaded build says it did not come from Play. The `IPP *` keys become `In-person payments *` — internal jargon that merchants were being shown.

The field-reference link points at `trunk`, so it 404s until this merges.

### Test Steps

Run tests on the last PR in this chain.

### Images/gif

N/A


samiuelson (approved):
LGMT!

PR #16333, opened 2026-07-30
Keep PR descriptions short in the pr skill

malinajirka:
### Description

The `pr` skill told agents it was "fine to walk through the author's reasoning", which reliably produced multi-section PR descriptions retelling the whole investigation. This swaps that for the opposite instruction: a couple of sentences of plain prose, and the conversation that produced the PR stays out of the body.


.agents/skills/pr/SKILL.md
```diff
@@ -34,8 +34,11 @@ Create a pull request following the project's PR conventions.
 7. **Push the branch.** Run `git push -u origin <branch-name>`.
 
 8. **Create the PR.** Read `.github/PULL_REQUEST_TEMPLATE.md` and use it as the body. Strip the HTML comments and fill in each section, following the conventions from the guidelines doc above:
-   - **Description**: start with `Fixes WOOMOB-XYZ` on its own line, then a self-contained summary of what changed and why.
-     - Write so a reviewer who never saw this branch can understand it. It's fine to walk through the author's reasoning — just make sure the description carries its own context rather than assuming the reviewer was part of the conversation that produced the PR.
+   - **Description**: start with `Fixes WOOMOB-XYZ` on its own line, then a short summary of what changed and why.
+     - Keep it to what a reviewer needs: what was wrong, what this changes, and anything they can't see in the diff. A couple of sentences of plain prose is usually right — no section headers beyond the template's.
```
irfano: `A couple of sentences` fits a PR that does one thing. If a PR fixes three separate bugs, I think we'd end up saying too little.

```suggestion
     - Keep it to what a reviewer needs: what was wrong, what this changes, and anything they can't see in the diff. A couple of sentences of plain prose is usually right — no section headers beyond the template's. If the PR fixes several distinct things, one line each is fine.
```

.agents/skills/pr/SKILL.md
```diff
@@ -34,8 +34,11 @@ Create a pull request following the project's PR conventions.
...
+     - The conversation that produced the PR is not the PR. Don't retell the investigation — no measurements, no debugging path, no narrative of how you got there.
```
irfano: `no measurements` feels too strict to me. On a performance fix the number is the whole reason we made the change, so I don't think we want to drop it.

```suggestion
     - The conversation that produced the PR is not the PR. Don't retell the investigation — no debugging path, no narrative of how you got there, no measurements unless the measurement is the reason for the change.
```

irfano (approved):
Nice, this was needed. The old line was exactly what let those long descriptions through.

Left two suggestions, I think both are worth taking before merge.

PR #16334, opened 2026-07-30
[WOOMOB-3704] Send a Mobile Status Report to Zendesk - Part 4

malinajirka:
### Description

Fixes WOOMOB-3704

Stacked on #16331. Support asked whether the report could show a store's notification settings and which of the two push systems serves it, so this adds a `Store Notifications` section for the selected store: the push registration status, and the store's new-order, review and stock alert settings read from the local cache with no network call on the ticket path. `Push registration` moves here from `Store Details`, since it belongs with the rest of that store's notification state.

The fields are printed as stored rather than interpreted. The report doesn't work out *why* a store is unregistered — the inputs are already separate fields (`FCM token`, `woo_self_driven_push_notifications_m1`, `Woo core version`, `Auth method`), and deriving a cause here would be a second copy of the decisions `RegisterDevice` makes. `docs/mobile-status-report.md` gains the field rows plus a table for reading those fields together, including what the alert thresholds mean in the settings screen's own wording.

### Test Steps

Test on the last PR in the chain

### Images/gif

N/A


samiuelson (approved):
LGTM!

PR #16335, opened 2026-07-30
[POS] Switch server-calculated refunds from wc/v4 to wc/v3

samiuelson:
WOOMOB-3716, WOOMOB-3749

### Description
The wc/v4 refunds endpoints are behind the server-side `rest-api-v4` feature flag and unreachable on production stores. WooCommerce core has ported the refund enhancements to wc/v3 in this PR: https://github.com/woocommerce/woocommerce/pull/67043, (expected in WC 11.1.0), so this PR switches the POS server-calculated refund flow to the v3 endpoints and removes the v4 flow.

**💡 Do not merge until this WC PR lands on trunk: https://github.com/woocommerce/woocommerce/pull/67042 to confirm target WC version (expected 11.1.0)**

- The refund preview posts to `POST /wc/v3/orders/{id}/refunds/preview`. Request and response payloads are unchanged from v4; the order id moved from the body to the URL. `RefundV4LineItem` is renamed to `RefundPreviewLineItem`.
- Refund creation uses the v3 `compute_totals: true` mode: line items keyed by `id` (new `ComputedRefundLineItem`), explicit `api_refund`/`api_restock` (v3 defaults both to true), optional `amount` override, response parsed by the same mapping as the classic v3 create. The v4-only `createSimplifiedRefund`, `SimplifiedRefundResponse`, and the flat `/wc/v4/refunds*` endpoint nodes are removed.
- A new `WooPosResolveRefundFlow` use case consolidates the eligibility decision (feature flag `woo_pos_server_refunds`, WC version at least 11.1.0, per-site availability cache) and returns `ServerComputed` or `LocalComputed`. The local calculators stay unchanged as the fallback for older stores.
- Safety invariant preserved: a `compute_totals` create is only sent when the availability cache is true, and the cache is set true only by a successful v3 preview on that site. Older stores silently drop the unknown `compute_totals` parameter, so probing with a create would produce a ghost zero-amount refund; the preview probe (404 `rest_no_route`) is the only safe detection.
- The submission intentionally sends no `amount` override: refund amount calculation is fully delegated to the backend at create time, for Interac card-present refunds as well. The server rejects a create that would exceed the remaining refundable amount, which covers the concurrent-refund case.

### Test Steps
Verify that: 
1. On a store running WooCommerce version built from that PR: https://github.com/woocommerce/woocommerce/pull/67043 refund flow uses `preview` endpoint  and refund `create` with "compute_totals: true"

2. On a store below 11.1.0, the flow falls back to local calculation and the classic v3 create with no user-visible difference.

3. Attempt to refund more than the remaining quantity: the preview surfaces an error instead of proceeding.
4. Check the Application log (Settings → Help & Support) filtered to `POS`: the resolved flow and its reason are logged on every refund.

Refund flow should use `preview` endpoint and refunds create with "compute_totals: true", on stores running WC version from this PR: https://github.com/woocommerce/woocommerce/pull/67043

image

You can test that WC build on this test store: https://cool-shop.mystagingwebsite.com/


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

iamgabrielma:
👋 @samiuelson  do you mind to re-assign this one to an Android engineer? I'm currently packed with PR reviews from core and ios, and won't be able to tackle this one till next week.

samiuelson:
Hi @malinajirka, would you be able to review this PR? 🙏

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -228,23 +228,23 @@ class WooPosRefundSubmissionProcessor @Inject constructor(
...
-                "WooPosRefund: creating simplified v4 backend refund " +
+                "WooPosRefund: creating server-computed backend refund " +
                     "orderId=${request.orderId}, " +
-                    "itemCount=${v4LineItems.size}, " +
+                    "itemCount=${serverLineItems.size}, " +
                     "autoRefund=$shouldAutoRefund, " +
                     "backendOnlyRetry=$retryBackendNotificationOnly"
             )
-            return refundStore.createSimplifiedItemsRefund(
+            return refundStore.createComputedItemsRefund(
```
malinajirka, line 241: AI comment, but I believe there might be something to it, wdyt?

💡 I believe we should pass `amount = request.refundAmount` here. `createComputedItemsRefund` supports the override (currently no production caller uses that parameter at all), and without it the server recomputes totals at create time — if the order changed between preview and create (concurrent refund from another register, price/tax edit), the refunded amount silently differs from what the cashier confirmed, and the success screen still shows the preview total. On the Interac path it's worse: the card is reversed for the previewed amount while the store records the recomputed one. With the override, a divergence becomes a surfaced server error instead of a silent mismatch. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosResolveRefundFlow.kt
```diff
@@ -0,0 +1,97 @@
...
+class WooPosResolveRefundFlow @Inject constructor(
+    private val selectedSite: SelectedSite,
+    private val availabilityCache: WooPosServerRefundAvailabilityCache,
+    private val getWooCoreVersion: GetWooCorePluginCachedVersion,
+    private val featureFlagRepository: FeatureFlagRepository,
+) {
+    operator fun invoke(): WooPosRefundFlow {
+        val wooVersion = getWooCoreVersion()
+        val eligibleVersion = when {
+            !featureFlagRepository.isEnabled(FeatureFlag.WOO_POS_REFUND_V4) ->
```
malinajirka: 🔍 np: `WOO_POS_REFUND_V4` now gates the v3 server-computed flow — everything else in this PR shed the V4 name. The remote key has to stay, but the enum constant could be renamed (`WOO_POS_SERVER_REFUNDS` or similar) so it doesn't read as gating the removed flow.

P.S. I'd personally vote for ditching the original value and replacing it  with a new one for clarity..

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 
 25.4
 -----
+- [Internal] POS refunds: server-refund availability is now tracked per WooCommerce version, so a store that changes version mid-session is re-probed instead of reusing a stale verdict [https://github.com/woocommerce/woocommerce-android/pull/16365]
```
malinajirka: 🔍 np: This links PR 16365, which is closed without merging — the cache change ships in this PR, so the link should be 16335. Also the v4→v3 switch itself might deserve its own `[Internal]` line.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosResolveRefundFlow.kt
```diff
@@ -0,0 +1,97 @@
...
+ * WooCommerce core changes, so a preview succeeding only proves the preview route exists. A store
+ * with the preview but without `compute_totals` (partial backport, or the changes splitting across
+ * releases) would silently drop the parameter and create a zero-amount refund while restocking
+ * items. Only a version known to contain both changes unlocks the server flow.
+ *
+ * [WooPosRefundFlow.ServerComputed] is eligibility, not proof: the availability cache is only set
+ * `true` by a successful preview, and only that value permits sending the computed create (see
+ * [WooPosRefundViewModel.buildSubmissionRequest]).
+ */
+class WooPosResolveRefundFlow @Inject constructor(
```
malinajirka, line 51: ❓ ⚠️ I just realized that the new flow is used only in the POS, not in the management section of the app. Is there a plan to migrate that part as well?

Feels like a no-go to me, to have two different refund flows with different behaviors. Wdyt?

malinajirka (approved):
Great job @samiuelson!

I've tested various scenarios and everything worked as expected.

Two items on my mind:
1. Unless it's already on your plan, I think we should consider using the same strategy in POS as well as the Store Management part of the app. Having two refund flows with different results feels off.
2. I noticed that we don't log the strategy (local/serverComputed) to any of the tracking events, something worth considering unless it's also on your radar already.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosResolveRefundFlow.kt
samiuelson: Fixed in 59b536ad7ca — renamed the constant to `WOO_POS_SERVER_REFUNDS` and took your suggestion on the remote key too: it now reads `woo_pos_server_refunds` (the old key never had a production rollout attached, so nothing to migrate).

RELEASE-NOTES.txt
samiuelson: Fixed in a1d156cc8fc — link now points at 16335 and the v4→v3 switch got its own `[Internal]` entry.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
samiuelson: Valid concern, especially the Interac card-vs-store mismatch. I held off on passing `amount` here because the endpoint documents it as an override of the computed total, not a validation guard — if the server accepts it without checking it against the computed line items, we would be forcing a stale preview total instead of surfacing an error, which is arguably worse. Filed https://linear.app/a8c/issue/WOOMOB-3796 to verify the actual server semantics and then add a divergence guard (server-side rejection if supported, otherwise comparing the create response total against the preview). It also covers the interplay with WOOMOB-3752: a guard that fails the create after a successful card refund needs the backend-only retry wiring fixed first.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosResolveRefundFlow.kt
samiuelson: Agreed — two flows with potentially different results is not a state to stay in. Migrating the store management refund screen to the same preview + `compute_totals` strategy is now tracked in https://linear.app/a8c/issue/WOOMOB-3797 (includes moving the flow resolution and availability cache out of `ui/woopos`). Keeping it out of this PR since the management screen has its own ViewModel and rollout considerations.

samiuelson:
Thank you for the thorough review and testing @malinajirka! 🙇 Both points are now tracked:

1. Migrating the store management refund flow to the same server-calculated strategy: https://linear.app/a8c/issue/WOOMOB-3797
2. Adding the resolved flow (`local` / `server_computed`) to the refund tracking events: https://linear.app/a8c/issue/WOOMOB-3798

Also filed https://linear.app/a8c/issue/WOOMOB-3796 for the `amount`-override question from the inline thread. The flag rename and release-notes fixes are pushed.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
samiuelson: Follow-up: I verified the actual server semantics in core (`DataUtils::prepare_refund_creation_or_error`, shared by the wc/v3 `compute_totals` path). The override is rejected with a 400 when it is below the computed line-item total, and accepted as a goodwill over-refund above it, capped at the order's remaining refundable amount (422 beyond). So passing the confirmed total surfaces upward drift as an error before the gateway moves any money, and in the downward case keeps the books consistent with what the cashier confirmed instead of silently recording a different amount.

Applied in af54d612293 for the backend path. The Interac path deliberately still omits the override: there the card is reversed before the create, and a rejected create would loop the backend-only retry on the same stale request — extending the guard there needs the retry rework first (WOOMOB-3752). WOOMOB-3796 now tracks just that remaining piece.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
```diff
@@ -54,41 +53,41 @@ class RefundRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
...
-            "line_items" to items,
-        )
-        val response = wooNetwork.executePostGsonRequest(
-            site = site,
-            path = WOOCOMMERCE.refunds.pathV4,
-            body = body,
-            clazz = SimplifiedRefundResponse::class.java,
-        )
-        return response.toWooPayload()
+            "amount" to amount,
```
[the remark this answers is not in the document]
samiuelson: @claude addressed — the parameter now has a production caller (the POS backend path pins the cashier-confirmed total in af54d612293), and the `amount`/`compute_totals` contract is documented on `createComputedRefund`.

PR #16337, opened 2026-07-30
[WOOMOB-3711] Share all application logs at once as a zip

irfano:
### Description
Fixes WOOMOB-3711

The application log screen can only share one day at a time, and that share is capped at the last 1000 entries because the text travels inside the intent. Support usually ends up with the most recent log, which shows little activity.

The log files list now has a share action that bundles every stored log file into a single zip, along with a `device_info.txt` entry. The zip goes out as a file rather than intent text, so the entry cap does not apply on this path.

### Test Steps
1. Open the app.
2. Tap Menu.
3. Tap Settings.
4. Tap Help & support.
5. Tap Application log.
6. Verify the toolbar shows a share icon.
7. Tap the share icon.
8. Verify the share sheet offers `woocommerce-logs.zip`.
9. Share it to an app you can open files from, for example Drive.
10. Open the zip and verify it holds one `log_<date>.txt` per row in the list, plus `device_info.txt`.

### Images/gif
image image image


WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
```diff
@@ -25,7 +27,8 @@ import javax.inject.Inject
 @HiltViewModel
 class WooLogViewerViewModel @Inject constructor(
     savedState: SavedStateHandle,
-    private val wooFileLogger: WooFileLogger
+    private val wooFileLogger: WooFileLogger,
+    private val deviceInfo: DeviceInfoWrapper
```
irfano, line 31: Swapped the `DeviceInfo` object for the injectable `DeviceInfoWrapper` here. Same values, it makes the view model unit testable.

libs/commons/src/main/java/com/woocommerce/android/util/logs/WooFileLogger.kt
```diff
@@ -97,6 +102,14 @@ class WooFileLogger(
 
     suspend fun getLogFiles(): List<File> = logFileWriter.getLogFiles()
 
+    /**
+     * Bundles every stored log file into a single zip, so all of them can be shared at once.
+     */
+    suspend fun archiveLogFiles(deviceInfo: String): File? {
+        forceFlush()
```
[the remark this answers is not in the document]
irfano: The flush race is pre-existing behavior shared by all the log read paths. The single file share and the Zendesk attachment use the same flush-then-read pattern, and with the 5 second foreground flush the worst case is the last few seconds of entries missing from the archive, or a torn final line in the current day's file. Making the flush wait for the write isn't a small change to the logger internals, so I kept this path consistent with the existing ones.

WooCommerce/src/main/res/xml/provider_paths.xml
```diff
@@ -6,4 +6,7 @@
     <files-path
         name="internal_files"
         path="." />
+    <cache-path
+        name="internal_cache"
+        path="." />
```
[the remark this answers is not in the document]
irfano: Good point, scoped it to the logs directory in ac44783.

hichamboushaba (commented):
Please check the comments below.

WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
```diff
@@ -73,10 +78,26 @@ class WooLogViewerViewModel @Inject constructor(
...
+            val archive = wooFileLogger.archiveLogFiles(deviceInfo = getDeviceInfo().toString())
+            isPreparingArchive.value = false
+
+            if (archive != null) {
+                triggerEvent(ShareLogsArchive(archive))
+            } else {
+                triggerEvent(ShareLogsArchiveFailed)
+            }
+        }
+    }
```
hichamboushaba, line 103: `isPreparingArchive` only reaches the UI after the state/IO/recomposition round trip, while `shareAllLogs()` itself has no in-flight guard. A repeat tap during that window can start two IO jobs that both truncate and write woocommerce-logs.zip, potentially producing a corrupt archive and multiple chooser events. Could we atomically guard re-entry (for example, `compareAndSet(false, true))` and reset the loading flag in finally?

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 
 25.4
 -----
+- [*] The application log screen can now share all stored logs at once as a single zip, instead of one day at a time
```
hichamboushaba:
```suggestion
- [*] The application log screen can now share all stored logs at once as a single zip, instead of one day at a time [https://github.com/woocommerce/woocommerce-android/pull/16337]
```

WooCommerce/src/main/kotlin/com/woocommerce/android/support/WooLogViewerViewModel.kt
irfano: Good catch, added the guard with compareAndSet and the flag now resets in a finally, done in a512eee.

hichamboushaba (approved):
Nice work @irfano.

PR #16338, opened 2026-07-30
[POS] Map refund API error codes to cashier-facing messages

samiuelson:
WOOMOB-3718

### Description
The wc/v3 refund endpoints return specific error codes when a refund request is rejected, most commonly because the order changed since the screen was loaded (for example another register refunded part of it). The app showed a generic error for all of them. This PR maps the actionable codes to cashier-facing messages. Stacked on #16335.

- A new app-layer `WooPosRefundApiError` enum matches the raw API error code carried by `WooError.apiErrorCode` and bridges it to string resources. The codes are matched with their `woocommerce_rest_` prefix as returned by the server: `woocommerce_rest_quantity_exceeds_refundable`, `woocommerce_rest_line_item_already_refunded`, `woocommerce_rest_order_not_refundable`, `woocommerce_rest_preview_exceeds_max_refundable` and `woocommerce_rest_refund_exceeds_remaining` (shared entry, same user meaning), `woocommerce_rest_refund_total_exceeds_line`, and `woocommerce_rest_invalid_refund_amount`. FluxC is intentionally untouched: these codes are refund-specific, so typing them in FluxC's shared `WooErrorType` enum would give them app-wide meaning and force every exhaustive `when` over the enum to enumerate refund cases.
- Programming-error codes (`invalid_line_item`, `duplicate_line_item`, and similar) stay unmapped on purpose: they indicate a client bug, not something a cashier can act on, and keep the generic message.
- POS: preview failures carry the mapped error into the state (`previewErrorMessage`, cleared on selection change, retry, and success) and the selection screen shows it inline; the submission processor prefers the mapped message over the existing `error.message` passthrough and the generic fallback. The mapping applies to the server-computed refund path only; the legacy local-calculation path is unchanged.
- The availability fallback is unchanged: only `rest_no_route` (`API_NOT_FOUND`) triggers the local-calculation fallback, and mapped errors do not touch the availability cache.

String copy (sentence case): "The selected quantity is more than what's left to refund", "This item has already been fully refunded", "This order can't be refunded", "The refund amount is more than what's left to refund on this order", "The refund amount is more than what's left to refund for this item", "The refund amount isn't valid".

### Test Steps
Verified on a store running WooCommerce trunk (11.1.0-dev) with the `woo_pos_server_refunds` flag enabled (default in debug builds). "Another session" below can be wp-admin or the store management app signed into the same store.

1. Create an order with quantity 2 of a product (for example a POS cash sale) and open the POS refund flow for it, stopping on the item selection step.
2. From another session, refund 1 of the 2 units. Back in POS, select both units and continue: the selection step shows "The selected quantity is more than what's left to refund" inline (server rejects the preview with `woocommerce_rest_quantity_exceeds_refundable`), Retry stays available, and changing the selection clears the message.
3. Select only the remaining unit and continue: the server-calculated preview succeeds and the review screen opens. From the other session, refund the remaining unit, then submit in POS: the error screen shows "This order can't be refunded" (the fully refunded order flips to the Refunded status, so the create is rejected with `woocommerce_rest_order_not_refundable`, HTTP 422). The mapped copy is shown instead of the raw server message.
4. Unmapped errors (network failure, server error) still show the previous generic messaging.

For testing purposes you can use cool-shop.mystagingwebsite.com test site build .zip plugin file from the latest dev version of WC.

Unit test coverage: `WooPosRefundApiErrorTest` (dedicated mappings, shared entry for the two over-remaining codes, unknown and programming codes stay unmapped), preview propagation and cache non-interference, processor message precedence, and ViewModel surfacing/clearing, all green. `detektAll` passes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundApiError.kt
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Matched on the raw API error code carried by `WooError.apiErrorCode` rather than on a
+ * `WooErrorType`: the codes are refund-specific, so typing them in FluxC's shared error enum
+ * would give them app-wide meaning and force every exhaustive `when` over `WooErrorType` to
+ * enumerate refund cases.
+ *
+ * Programming-error codes (invalid line item ids, malformed payloads, etc.) intentionally have
+ * no entry here: they indicate a client bug and keep the generic error message.
+ */
+enum class WooPosRefundApiError(@StringRes val messageRes: Int) {
```
malinajirka, line 19: 💡 We might want to consider having `isRetryable/isRecoverable` flag on each, to ensure we don't display retry on all of them. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundState.kt
```diff
@@ -32,6 +32,9 @@ sealed class WooPosRefundState {
         // back to the local calculation and these stay at their defaults (no loading/no failure).
         val isPreviewLoading: Boolean = false,
         val previewFailed: Boolean = false,
+        // Refund-specific message for a failed preview (see [WooPosRefundApiError]); null when
```
malinajirka: Feels like a worthless comment?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt
```diff
@@ -61,13 +62,19 @@ class WooPosRefundPreview @Inject constructor(
...
+            else -> Result.Error()
         }
     }
 
     sealed interface Result {
         data class ServerCalculated(val preview: WCRefundPreview) : Result
         data object FallbackToLocal : Result
-        data object Error : Result
+
+        /**
```
malinajirka: Feels like a worthless comment, wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -295,9 +295,16 @@ class WooPosRefundSubmissionProcessor @Inject constructor(
                     "message=${error.message}, " +
                     "errorData=${error.errorData}"
             )
+            // Refund-specific errors (usually a stale order state, for example another register
```
malinajirka: Feels like a worthless comment, wdyt?

malinajirka (approved):
LGTM, thanks!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundApiError.kt
samiuelson: Good catch — done, though I landed on a slightly different shape and want to flag why.

Checking each case, **all six mapped codes are deterministic 4xx validation rejections**: quantity-exceeds, already-refunded, order-not-refundable, exceeds-remaining (order and line), and invalid-amount all produce the identical rejection when the same request is resubmitted. So a per-case `isRetryable` flag would have carried `false` six times — a constant wearing a flag's clothing, and six places to keep in sync.

Instead the rule lives in one place: `Failure` now carries the store's raw `apiErrorCode`, and `canRetry` defaults to false when `WooPosRefundApiError.fromCode` recognises it. Kept raw rather than pre-mapped so unrecognised codes stay available for the refund analytics work (#16419).

Worth noting the preview path was already fine — `continueToReviewButtonState` disables Continue on `previewFailed`, so only the create error screen was offering a button that could not work.

This only removes the misleading affordance. Actual recovery — refreshing the refundable list for the stale-quantity codes, routing `order_not_refundable` to the existing `NoRefundableItems` terminal state — is WOOMOB-3751, which I have promoted off the backlog on the strength of this comment. iOS turns out to have the same gap on *both* paths, so it is tracked there too.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundState.kt
samiuelson: Agreed — removed. It narrated the code beneath it rather than saying anything the code could not.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt
samiuelson: Agreed — removed. It narrated the code beneath it rather than saying anything the code could not.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
samiuelson: Agreed — removed. It narrated the code beneath it rather than saying anything the code could not.

PR #16339, opened 2026-07-30
[WOOMOB-3704] Send a Mobile Status Report to Zendesk - Part 6

malinajirka:
### Description

Fixes WOOMOB-3704

Stacked on #16336. Corrections to the report from a review pass over the whole stack, plus two changes for cross-platform consistency with iOS — no new fields.

`Device locale` was read from `Resources.getSystem()`, which from API 33 also carries the per-app language override, so it printed the same value as `App language` for exactly the merchants who had set one. It now reads through `LocaleManagerCompat.getSystemLocales()`. `Connected stores` counted every site on the WordPress.com account, not just the Woo ones, so a merchant with a pile of personal blogs read as having dozens of stores; both the count and the list are now filtered. `Remote values loaded: false` claimed no remote fetch had ever succeeded, which is one of three indistinguishable causes — it now states only what is observable, and the doc explains the rest.

`FCM token` becomes `Push token`, so the label means the same thing on both platforms; the internal FCM plumbing keeps its name, since that is accurate on Android. `Auth method` keeps its existing values, which iOS will map onto, and gains `unknown`: `connectionType` has to return something, so in production it answers `Jetpack` for a store it could not classify, and the report was stating that guess as fact. A new `connectionTypeOrNull` in commons lets the report see the ambiguity, leaving the 58 other callers untouched.

The remaining change is structural: `runCatching` sat around whole sections, so one throwing field cost every other field beside it. A new `safeEntry` confines a failure to its own field.

### Test Steps

1. Set a per-app language for the app (System settings → Apps → WooCommerce → Language) that differs from the device language, then open Settings → Help & Support → Mobile status report. Confirm `Device locale` shows the *device* language and `App language` shows the override.
2. On an account with both Woo stores and non-Woo WordPress.com sites, confirm `Connected stores` counts only the Woo stores and `All connected stores:` lists only those.
3. Confirm the Notifications section now reads `Push token` rather than `FCM token`.
4. On a Jetpack store and on an application-password store, confirm `Auth method` still reads `Jetpack` / `ApplicationPasswords` as before.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka:
Closing — the fixes from this PR are now applied at origin in Parts 1-5 rather than bolted on top, along with the simplifications discussed (golden-master test, leaner code comments, and a single `# Selected store:` band replacing the per-heading scope suffixes). Parts 1-5 have been force-pushed with the redistributed content.

PR #16340, opened 2026-07-30
Report the device free space Zendesk sees as the real figure

malinajirka:
### Description

The `Device free space` field on every support ticket has been reporting roughly twice the space actually free, capped at megabytes.

It came from `DeviceUtils.getTotalAvailableMemorySize()` in `org.wordpress:utils`, which adds the free space of internal and external storage together. On a device with emulated external storage — every current device — both are views of the same partition, so the same bytes are counted twice. Its formatter also stops at megabytes, so a half-full 128 GB phone reported `104,857MB`. We now measure the data partition once with `StatFs` and format up to terabytes, giving `12.4 GB`.

Found while working on the Mobile Status Report (#16325 and the PRs stacked on it), which surfaces the same value. Kept separate because the bug predates that stack and affects tickets today. The property keeps its `totalAvailableMemorySize` name so this does not conflict with those PRs, though the name is a misnomer for free disk space and is worth renaming once they land.

### Test Steps

1. Note the real free space on your device (Settings → Storage).
2. In the app, Settings → Help & Support → Contact Support, and submit a ticket.
3. Open the ticket in Zendesk and confirm `Device free space` matches step 1 and is formatted like `12.4 GB` rather than `104,857MB`.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/298">25.4</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

kidinov (approved):
LGTM, but some improvements ideas from the AI:


- **`StatFs.availableBytes`** exists since API 18, so `availableBlocksLong * blockSizeLong` can be just `StatFs(...).availableBytes`.
- **Unit is picked before rounding**, so `1 GiB - 1 byte` prints `1024.0 MB` instead of `1.0 GB`. Same on every boundary.
- Divisor is 1024 but labels are `KB/MB/GB/TB`. Android `Formatter` uses SI since API 26, so the ticket will read ~7% lower than what user sees in Settings. Either divide by 1000, or use `KiB/MiB/GiB`.
- **Test step 1** says the value should match Settings -> Storage. It will not match exactly: `StatFs` excludes reclaimable cache, Settings uses `StorageStatsManager.getFreeBytes()` which includes it. Maybe say "in the same ballpark", otherwise someone will report it as still broken.
- **Release note is missing the PR link**, all other entries in 25.4 have one.


Seems to make sense

malinajirka:
Addressed the first three and the last one: switched to `StatFs.availableBytes`, the unit is now picked after rounding (so 1 GB − 1 byte prints `1.0 GB`), the formatter divides by 1000 to line up with what Settings shows, and the release note now has the PR link.

PR #16344, opened 2026-07-30
[WOOMOB-2576] Add alt text and title editing for product images

irfano:
### Description

Closes [WOOMOB-2576](https://linear.app/a8c/issue/WOOMOB-2576/request-to-add-image-alt-editing)

Merchants can't edit a product image's alt text or title from the app and have to switch to wp-admin for SEO and accessibility work. This adds an Image details screen behind a pencil icon in the fullscreen image viewer. Like the other product subscreens, edits go into the product draft and are saved with the product's Save button, which also updates the attachment in the site's media library.

Two behaviors worth calling out:

- The products API [ignores empty image alt and name values](https://github.com/woocommerce/woocommerce/blob/54a287bab33682e0ff14dd74f7867a574da08663/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-products-controller.php#L710-L716), so clearing a field can't delete the value on the server. The screen reflects that: an emptied field shows the current value as a hint, a note under the field points to wp-admin, and the stored value is kept.
- An alt text the app doesn't know, for example on images picked from the media library, is left out of the update request entirely so the server value isn't overwritten.

The edit action is only available in the product images flow; variation images are out of scope.

### Test Steps

1. Open a product that has an image.
2. Tap the image to open the Photos screen.
3. Tap the image to open the fullscreen viewer.
4. Tap the pencil icon.
5. Type an alt text and a title.
6. Tap the back arrow.
7. Tap the back arrow again to return to the product screen: the Save button appears.
8. Tap Save.
9. In wp-admin, open the image in the Media Library: the alt text and title show the new values.
10. In the app, open the same image details again.
11. Clear the Title field: the stored title shows as a hint and a note says the app can't remove it.
12. Tap the back arrow twice to return to the product screen: the Save button does not appear.
13. Open a variation image in the fullscreen viewer: the pencil icon is not shown.

### Images/gif

| Edit action in the viewer | Image details | Cleared field keeps the value as a hint | Edited alt text | Removed alt text |
| --- | --- | --- | --- | --- |
| Pencil icon in the image viewer toolbar | Image details screen | Cleared alt text showing the stored value as a hint | Edited alt text with Done enabled | image |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/model/Product.kt
```diff
@@ -354,7 +356,8 @@ fun Product.toDataModel(storedProductModel: WCProductModel? = null): WCProductMo
                 JsonObject().also { json ->
                     json.addProperty("id", image.id)
                     json.addProperty("name", image.name)
-                    json.addProperty("source", image.source)
+                    json.addProperty("src", image.source)
```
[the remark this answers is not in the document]
irfano: src is only used when an image has no id, that's the sideload path, so wc/v3 ignores it for existing images and nothing gets re-downloaded. Also ran it on a test store: adding, reordering and deleting images all came back correct after a fresh fetch.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsScreen.kt
```diff
@@ -0,0 +1,132 @@
...
+import androidx.activity.compose.BackHandler
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.MaterialTheme
```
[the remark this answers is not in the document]
irfano: 👍🏻  Fixed in https://github.com/woocommerce/woocommerce-android/pull/16344/changes/cbfa720ea05fa7ba9bae36960e284588a9a6ee0a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsScreen.kt
```diff
@@ -0,0 +1,132 @@
...
+import androidx.compose.material.Scaffold
```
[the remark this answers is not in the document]
irfano: 👍🏻  Fixed in https://github.com/woocommerce/woocommerce-android/pull/16344/changes/cbfa720ea05fa7ba9bae36960e284588a9a6ee0a

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsFragment.kt
```diff
@@ -0,0 +1,37 @@
...
+        return composeView {
+            ProductImageDetailsScreen(viewModel)
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        viewModel.event.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is MultiLiveEvent.Event.ExitWithResult<*> ->
+                    navigateBackWithResult(ProductImageDetailsViewModel.KEY_IMAGE_DETAILS_RESULT, event.data)
```
JorgeMucientes: Np. I wouldn't use `ProductImageDetailsViewModel.KEY_IMAGE_DETAILS_RESULT` directly here, I'd use `event.key!!` to ensure we use the key supplied by the VM (which today is only `KEY_IMAGE_DETAILS_RESULT`).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsScreen.kt
```diff
@@ -0,0 +1,132 @@
...
+
+            Spacer(modifier = Modifier.height(dimensionResource(R.dimen.major_100)))
+
+            WCOutlinedTextField(
+                value = state.altText,
+                onValueChange = onAltTextChanged,
+                label = stringResource(R.string.product_image_details_alt_text_label),
+                placeholderText = state.altTextPlaceholder,
+                modifier = Modifier.fillMaxWidth()
+            )
```
JorgeMucientes, line 87: This approach has an issue when clearing the texts. A merchant clearing the alt text to remove it gets a disabled Done, no dialog on back, and the old value still live on the server, with no feedback explaining any of it. 

https://github.com/user-attachments/assets/da06239d-b4c7-442b-9ff4-e39f11a3a3ea

JorgeMucientes (approved):
Nice work @irfano! It works as described for the most part but I left a couple nits that are worth checking before merging. 

Also I'd like to suggest a different approach to this feature. But not to be addressed in this PR. I feel it's confusing that  `ProductImageDetailsScreen.kt` has a dialog to confirm or discard changes (like product detail has) and a "Done" button that is enabled/disabled based on chaging any of the values, but that never actually commits those changes. The use might leave the screen thinking title or alt text were applied were they are not until the "Save" button is tapped. I feel this particular info should be synced right away as soon as the user taps `Done` form `ProductImageDetailsScreen.kt`. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsFragment.kt
irfano: Good point, now taking the key from the event. Went with requireNotNull(event.key) instead of !! so a missing key fails with a clear message. 25f375c6624

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/images/ProductImageDetailsScreen.kt
irfano: Good point. Now clearing the text shows a note under the field: "The app doesn't support removing alt text. Please use wp-admin to remove it." We already use a similar one in the custom fields editor ("The app currently does not support creating duplicate keys. Please use wp-admin to duplicate a key if needed.").

0ad60b75de7

image

I'm also removing the DONE button in the next commit.

irfano:
Thank you, @JorgeMucientes! Great review!

> Also I'd like to suggest a different approach to this feature. 

I noticed that none of the other subscreens have a DONE button, so I took a different approach from the one you suggested: I removed it. Changes made on the Image Details screen are now saved with the SAVE button on the product screen, which also removes the discard dialog since nothing gets lost anymore. This makes Image Details consistent with the other subscreens.

Here are recordings of the Price subscreen as it is today and the updated Image Details screen. How does this look?

[price.webm](https://github.com/user-attachments/assets/30d1a5a7-bd96-4d6f-a72d-001f832a93c3)

[image-details.webm](https://github.com/user-attachments/assets/b0cab902-f7ba-4d32-931b-4c7e303b767a)

JorgeMucientes:
Nice! Thanks for addressing the feedback @irfano. 

Just one more question. While testing I noticed that we now enforce the user to add an `alt text` event though this is an option field isn't it? I mean when I open the editor the first time the `alt text` is empty. But If I add a value, then it becomes mandatory. See the screen recording: 

https://github.com/user-attachments/assets/858ce661-fffb-418b-ab9b-e1a884518603

irfano:
Great catch, Jorge! The message was shown even when the value on the server was empty, which shouldn't happen. Fixed it in the last commit. fb016f332b6

PR #16346, opened 2026-07-31
[WOOMOB-3262] Follow the locale's date order in displayed dates

irfano:
Fixes WOOMOB-3262

Dates were formatted with hardcoded `SimpleDateFormat` patterns like `MMM d` and `MMM d, yyyy`. The pattern dictates the order, so every language got the American month-day order. The orders list showed `Jul 29` in British English and `Tem 29` in Turkish, where both call for the day first.

### Description

Display formats now derive their pattern from the current locale through `DateFormat.getBestDateTimePattern`, so the app's displayed language drives the order.

Machine-readable formats are untouched. `formatToYYYYmmDD`, `formatToYYYYmmDDhhmmssGmt` and the other ISO helpers keep their fixed patterns, so API payloads and stored values are unchanged.

The legacy WCShip label screens moved too. They sit behind the WooShipping revamp flag, so they are not in the screenshots.

`formatAsRangeWith` is not part of this change, so analytics intervals still read `Jul 20 – 26, 2026`. Fixing that needs interval formatting rather than string concatenation.

### Test Steps

1. Set the app language to English (United Kingdom) from the system settings.
2. Open the Orders tab.
3. Confirm order dates read `29 Jul`, not `Jul 29`.
4. Tap Filters.
5. Tap Date Range.
6. Tap Custom Range.
7. Pick a start and an end date, then tap Save.
8. Confirm the range reads `15 Jun 2026 - 20 Jul 2026`.
9. Open the Products tab.
10. Open a product with a scheduled sale.
11. Confirm the sale date under Price reads day first.
12. Tap Price.
13. Tap the sale end date field.
14. Confirm the picker opens on the date shown in the field.
15. Pick a different date and confirm the field updates.
16. Open Menu, then Customers.
17. Open a customer and confirm Last active reads day first.
18. Open Menu, then Inbox.
19. Scroll to a note older than 30 days and confirm its date reads day first.
20. Open My store, then View all store analytics.
21. Confirm the Today range and the comparison line read day first.
22. Switch the app language to English (United States).
23. Repeat steps 2 and 3 and confirm dates read `Jul 29`.

### Images/gif

Emulator with the app language set to English (United Kingdom).

| | Before | After |
|---|---|---|
| Orders list |  |  |
| Recent orders card |  |  |
| Analytics range header |  |  |
| Order filters, custom range |  |  |
| Product detail, sale dates |  |  |
| Product pricing, scheduled sale |  |  |
| Customer detail, last active |  |  |
| Inbox, older notes |  |  |
| Blaze campaign budget |  |  |

Analytics intervals still read month-first.


WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
```diff
@@ -174,15 +169,15 @@ fun Date.isInSameMonthAs(other: Date, baseCalendar: Calendar): Boolean {
 
 fun Date.formatAsRangeWith(other: Date, locale: Locale, calendar: Calendar): String {
     val formattedStartDate = if (this.isInSameYearAs(other, calendar)) {
-        this.formatToMMMdd(locale)
+        SimpleDateFormat("MMM d", locale).format(this)
```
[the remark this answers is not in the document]
irfano: This is called out in the description, leaving `formatAsRangeWith` as is was intentional here. The same-month shortening is why a pattern swap is not enough on its own, the range assembly needs restructuring. Tracked in [WOOMOB-3725](https://linear.app/a8c/issue/WOOMOB-3725/fix-analytics-date-ranges-showing-the-wrong-date-order).

WooCommerce/src/main/kotlin/com/woocommerce/android/util/PriceUtils.kt
```diff
@@ -88,23 +88,23 @@ object PriceUtils {
...
+                    dateOnSaleFrom.formatToLocalizedMonthDay()
                 } else {
-                    dateOnSaleFrom.formatToMMMddYYYY()
+                    dateOnSaleFrom.formatToLocalizedMedium()
                 }
                 resources.getString(
                     R.string.product_sale_date_from_to,
                     formattedFromDate,
-                    dateOnSaleTo.formatToMMMddYYYY()
+                    dateOnSaleTo.formatToLocalizedMedium()
```
JorgeMucientes, line 98: **Mixed month-name and numeric formats in the same string for numeric-medium locales**

For 6 of the 16 shipped locales the localized MEDIUM style is fully numeric (the
abbreviated month name disappears) while `getBestDateTimePattern(locale, "MMMd")`
keeps it:

| locale | `formatToLocalizedMedium` | `formatToLocalizedMonthDay` |
|---|---|---|
| de | `03.08.2026` | `3. Aug.` |
| ja | `2026/08/03` | `8月3日` |
| ko | `2026. 8. 3.` | `8월 3일` |
| zh-CN / zh-TW | `2026年8月3日` | `8月3日` |
| ar | `٠٣‏/٠٨‏/٢٠٢٦` | `3 أغسطس` |

Because the two helpers are paired at sites that render them side by side, German
(and ja/ko/zh/ar) now get two different date systems at once:

- In the above piece of code — the same-year branch renders `3. Aug. - 01.09.2026` in a single
  `%1$s - %2$s` string.
- `OrderListItemDataSource.getFormattedDateWithSiteTimeZone` — the orders list
  mixes `3. Aug.` (this year) with `28.12.2025` (last year) in adjacent rows.
- `BlazeCampaignBudgetViewModel.getFormattedStartDate` — endless vs. fixed
  campaigns switch styles.

Before this PR both halves were `MMM d`-family, so they matched; this is the one
spot where the change makes things worse rather than better.

Suggestion: give the year variant a skeleton too rather than the MEDIUM style —
a `formatToLocalizedMonthDayYear()` on `getBestDateTimePattern(locale, "yMMMd")`.
CLDR's `yMMMd` is `d. MMM y` for de, `y年M月d日` for zh, `d MMM y` for ar, so it
stays consistent with `MMMd` in every locale, which MEDIUM does not. Worth
confirming on device since Android's ICU tables aren't identical to the JDK's
(the table above is JDK 21 CLDR).

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
```diff
@@ -59,27 +62,19 @@ fun Date.formatToDDMMMYYYY(locale: Locale = Locale.getDefault()): String = Simpl
...
-    "MMM d, yyyy",
-    locale
-).format(this)
-
 /**
  * Formats the date to a string in the format "MMM d, yyyy" considering the current locale.
  */
 fun Date.formatToLocalizedMedium(locale: Locale = Locale.getDefault()): String = SimpleDateFormat
     .getDateInstance(SimpleDateFormat.MEDIUM, locale)
     .format(this)
```
JorgeMucientes, line 71: Np. Outdated comment. `formatToLocalizedMedium` still claims the format is "MMM d, yyyy". That's precisely the misconception this PR exists to correct, and it's now false for de/ja/ar.
Perhaps something along the lines of 
```kotlin
/**
 * Formats the date with the locale's medium date style, e.g. "Aug 3, 2026" in English, "3 Aug 2026" in British
 * English and "03.08.2026" in German — some locales drop the month name entirely.
 */
```
would be clearer.

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
```diff
@@ -59,27 +62,19 @@ fun Date.formatToDDMMMYYYY(locale: Locale = Locale.getDefault()): String = Simpl
...
-fun Date.formatToDDyyyy(locale: Locale): String = SimpleDateFormat(
-    "d, yyyy",
-    locale
-).format(this)
+/**
+ * Formats the date and time to a string in the format "MMM d, yyyy, h:mm AM" considering the current locale.
+ */
+fun Date.formatToLocalizedMediumWithTime(locale: Locale = Locale.getDefault()): String = SimpleDateFormat
+    .getDateTimeInstance(SimpleDateFormat.MEDIUM, SimpleDateFormat.SHORT, locale)
+    .format(this)
```
JorgeMucientes, line 77: Same as above:
```kotlin
/**
 * Formats the date with the locale's medium date style and short time style, e.g. "Aug 3, 2026, 9:30 AM" in
 * English and "3 Aug 2026, 09:30" in British English. The time style follows the locale, not the system
 * 12h/24h setting.
 */
```

Wdyt?

JorgeMucientes (approved):
Nice work @irfano. The changes work as described, however I'm leaving one suggestion that should probably be addressed to ensure date formatting works well in all language. Let me know your thoughts about it.

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
irfano: Good catch, updated in 589fd86.

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/DateExt.kt
irfano: Taken as is. 2e949ca

WooCommerce/src/main/kotlin/com/woocommerce/android/util/PriceUtils.kt
irfano: Confirmed on a German device. Added `formatToLocalizedMonthDayYear` on a `yMMMd` skeleton and switched the dates with a year over to it, in both the `Date` extension and `DateUtils`. d753c86

Blaze was already mixed before this PR, the start date used `formatToMMMdd` and the end used `formatToLocalizedMedium`, so German showed `Aug. 12 - 19.08.2026`. It reads `12. Aug. - 19. Aug. 2026` now.

In German:

| | trunk | after |
|---|---|---|
| Orders list |  |  |
| Product sale dates |  |  |
| Blaze campaign budget |  |  |

PR #16347, opened 2026-07-31
[WOOMOB-3710] Fix StackOverflowError when leaving the QR session-replace warning

malinajirka:
### Description

Fixes WOOMOB-3710

A merchant who is already signed in and opens a `woocommerce://qr-login?...` deep link sees the "You're already signed in" session-replace warning. Pressing system Back on that screen crashes the app with `java.lang.StackOverflowError`.

`QrLoginSessionReplaceWarningScreen` registers `BackHandler(onBack = onCancel)` so that system Back goes through the same dismiss path as the Cancel button (keeping the `DISMISS` tracks event and the state reset symmetric). `onCancel` is `QrLoginScannerFragment.handleCancelSessionReplace()`, which reset the ViewModel state and then called `onBackPressedDispatcher.onBackPressed()` to leave the screen.

That last call is the problem: when it runs from the `BackHandler`, the handler is still registered and enabled — the ViewModel's transition to `Idle` only reaches the composition on the next recomposition — so the dispatcher picks the same handler again, which calls `onBackPressed()` again, and so on until the stack overflows. `QrLoginScannerViewModel.onCancelSessionReplace()` does guard on the current state, but that only stops the tracks event from firing repeatedly; the fragment re-dispatched Back unconditionally, so the loop was unaffected.

The fix is to stop re-entering the back dispatcher. `QrLoginScannerFragment.Listener` gains `onQrLoginDismissed()`, which `LoginActivity` implements with its existing `handleBackPress()` — pop the fragment if there is a back stack, otherwise finish, which is exactly what a system Back press would have done. All four exit points in the fragment (`handleCancelSessionReplace`, `handleStartOver`, `handleCancelNumberMatch`, and the scanner's `Exit` event) now route through a single `dismiss()` helper instead of calling `onBackPressed()` themselves.

Only `handleCancelSessionReplace` is reachable from a `BackHandler` today, so it is the only one that could crash. The other three were changed for consistency: they express the same "leave the QR surface" intent, and routing them the same way means adding a `BackHandler` to the number-match or error screen later can't reintroduce this crash.

There is no automated regression test. Reproducing the recursion needs a real `OnBackPressedDispatcher` driving a live composition, and the unit test suite has no Robolectric or Compose test infrastructure — that only exists under `androidTest`. Verified manually on an emulator instead; steps below.

### Test Steps

The session-replace warning only appears when you are already signed in, so stay signed in for all of these.

1. Sign in to a store and stay on the app's My Store screen.
2. From a terminal, open the QR login deep link:
   ```
   adb shell am start -a android.intent.action.VIEW \
     -d "woocommerce://qr-login?siteUrl=https%3A%2F%2Fexample.com" com.woocommerce.android.dev
   ```
3. The "You're already signed in" warning appears.
4. Press system Back. Confirm the app does **not** crash or freeze, the login screen closes, and you are returned to wherever you were before the deep link.
5. Reopen the app and confirm you are still signed in and the store dashboard loads — cancelling must leave the existing session untouched.
6. Repeat steps 2–5, this time tapping the **Cancel** button instead of pressing Back. The result should be identical.
7. On `trunk`, step 4 crashes the app with `java.lang.StackOverflowError: stack size 8188KB` (visible in logcat) and wedges input dispatch for several seconds first.

### Images/gif

N/A


JorgeMucientes (approved):
Thanks for fixing this @malinajirka. It works well and code changes look good

PR #16349, opened 2026-08-03
[WOOMOB-3728] Preserve COD gateway title when disabling Pay in Person

malinajirka:
### Description

Fixes WOOMOB-3728

Toggling **Pay in Person** off in the Payments menu sent the whole gateway payload — `enabled = false` plus `title`, `description` and `instructions` — so a merchant who never opted into the rename ended up with a disabled Cash on Delivery gateway now called "Pay in Person".

`toggleCashOnDeliveryOption` now only sends `enabled = false` when disabling. Null params are omitted from the request body by `GatewayRestClient`, so the merchant's own title/description survive. Enabling still sets them, matching iOS.

The second commit is a small UX fix on the same row: while the Cash on Delivery status is being fetched, the row shows a progress indicator in place of the toggle instead of rendering as "off" and then snapping to the real value. Only the initial load shows it — returning to the screen keeps the last known state and refreshes silently.

### Test Steps

1. In the store's WP admin, rename the Cash on Delivery gateway title to something custom, e.g. "Cash".
2. In the app, open **Payments** and note the Pay in Person row shows a spinner instead of the toggle until the status loads.
3. Toggle **Pay in Person** off.
4. Back in WP admin, confirm the gateway is now disabled and its title is still "Cash".
5. Toggle **Pay in Person** back on in the app and confirm the title becomes "Pay in Person" — enabling still applies it, as before.

### Images/gif

https://github.com/user-attachments/assets/c060a09e-68f4-425e-be60-0a1bb4cf8eea


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/298">25.4</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

kidinov (commented):
LGTM but I wouldn't add another loading indicator as we have the generic one already. We could just fix the bug when it was clickable even when was "disabled" (OFF and disabled visually 2 different states while look a bit similar)

Patch based on the trunk to do so:

[Fix_eligibility_check_for_phone_as_card_reader_in_PaymentsHubViewModel.patch](https://github.com/user-attachments/files/30666241/Fix_eligibility_check_for_phone_as_card_reader_in_PaymentsHubViewModel.patch)

wdyt?

malinajirka:
Thanks for the review @kidinov!

The patch says `not found 🤔` . However, I'm not sure if fixing the clickability fixes the UX problem - I believe users think the app enables the toggle upon entering the screen -> so if the app disaplys `toggle off`, no matter whether it's enabled or disabled -> then it switches it to `toggle on` (because it fetched the state from remote) => they'll still think, the app enabled the feature just upon entering the screen => that's why I leaned towards not showing the toggle in any state, until we know the state. Wdyt?

Btw the top/generic progress indicator is independent - it indicates loading of IPP status. I realize, it's better to have just one progress indicator, but AFAIK showing small sub-indicators is quite common on Android.

Having said all that, if you still prefer to not display the secondary progress indicator, I guess we could simply hide the toggle and show it when we know the state.

kidinov:
@malinajirka 
Weird bug with the patch - I can load it. Here it's like a text:

```
Subject: [PATCH] Fix eligibility check for phone as card reader in PaymentsHubViewModel
---
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewHolder.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewHolder.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewHolder.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewHolder.kt	(revision 6cd092c5073e4d223807ae070d495ab9e10a7f69)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewHolder.kt	(date 1785768130744)
@@ -77,7 +77,9 @@
                 uiState.onLearnMoreClicked()
             }
             binding.root.setOnClickListener {
-                binding.paymentsHubSwitch.isChecked = !uiState.isChecked
+                if (uiState.isEnabled) {
+                    binding.paymentsHubSwitch.isChecked = !uiState.isChecked
+                }
             }
         }
     }
Index: WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt	(revision 6cd092c5073e4d223807ae070d495ab9e10a7f69)
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt	(date 1785768125464)
@@ -90,6 +90,7 @@
             ),
             index = 2,
             isChecked = false,
+            isEnabled = false,
             onToggled = { (::onCashOnDeliveryToggled)(it) },
             onLearnMoreClicked = ::onLearnMoreCodClicked
         )
@@ -150,7 +151,8 @@
         val isCashOnDeliveryEnabled = cashOnDeliverySettingsRepository.isCashOnDeliveryEnabled()
         updateCashOnDeliveryOptionState(
             cashOnDeliveryState.value?.copy(
-                isChecked = isCashOnDeliveryEnabled
+                isChecked = isCashOnDeliveryEnabled,
+                isEnabled = true
             )!!
         )
     }
Index: WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModelTest.kt
IDEA additional info:
<+>UTF-8
diff --git a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModelTest.kt b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModelTest.kt
--- a/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModelTest.kt	(revision 6cd092c5073e4d223807ae070d495ab9e10a7f69)
+++ b/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModelTest.kt	(date 1785768172382)
@@ -47,6 +47,7 @@
 import com.woocommerce.android.viewmodel.BaseUnitTest
 import com.woocommerce.android.viewmodel.MultiLiveEvent
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.awaitCancellation
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Before
@@ -55,6 +56,7 @@
 import org.mockito.ArgumentMatchers.anyLong
 import org.mockito.kotlin.any
 import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.doSuspendableAnswer
 import org.mockito.kotlin.eq
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.never
@@ -827,6 +829,28 @@
         ).isTrue
     }
 
+    @Test
+    fun `given cod status is being fetched, when screen shown, then cash on delivery is not allowed to toggle`() =
+        testBlocking {
+            // GIVEN
+            whenever(cashOnDeliverySettingsRepository.isCashOnDeliveryEnabled()).doSuspendableAnswer {
+                awaitCancellation()
+            }
+
+            // WHEN
+            initViewModel()
+
+            // THEN
+            assertThat(
+                (
+                    viewModel.viewStateData.getOrAwaitValue().rows.find {
+                        it.label == UiStringRes(R.string.card_reader_enable_pay_in_person)
+                    }
+                        as ToggleableListItem
+                    ).isEnabled
+            ).isFalse
+        }
+
     @Test
     fun `given cash on delivery enabled, when screen shown, then cash on delivery state is enabled`() =
         testBlocking {
```


About toggle looking "off": in my patch it's not just off, it's **disabled**. So when state arrives, two things change at same time - it becomes colored and it moves to on. I think this reads differently than a live toggle that jumps by itself.

But main point for me is we don't need a second spinner, because top progress bar is already showing during this time. Disabled row + top loading together say "we are still loading something and this one is not ready yet". For me that's enough signal better UX than 2 loading things on one screen

One thing though, and here I agree with you that current behaviour is not good: we should tie the top indicator to **both loads** - COD state and IPP status - and hide it only when both are done. Right now `isLoading` comes only from the onboarding check, so top bar can disappear while COD is still fetching, and then merchant sees a disabled toggle with no loading indication anywhere. That's a real hole, but it's small fix.

If after that you still think it's not enough, then I'd prefer we just hide the toggle until we know the state, like you suggested, instead of adding second progress indicator.

malinajirka:
I went with the 'hide toggle and re-use loading indicator` option in the end: dropped the second spinner — the toggle is simply hidden until its state is known + implemented your suggestion to tie the top indicator to both loads, so it stays until the COD initial state loading and the onboarding check have both finished. Could you please review the last two commits?

kidinov (approved):
I tested and to me disabled still feels more logical and commonly used pattern as this is kind of breaking the UI element (people think the whole row as one UI element not just the toggle itself)

Having said that it's better than have it clickable and "off" as it's currently is

One thing I'd change: on `ToggleableListItem` we now have `isChecked` + `isLoading` + `isEnabled`, so 8 possible combinations for a row that really has only 3 states. Can we make it **one tri-state value** instead of two booleans? Something like `Loading` / `Checked` / `Unchecked`. Then the view holder has to handle all cases explicitly, and right now it works only because an `isInvisible` switch does not receive touches, which is a bit implicit.

Also small: `isLoading: Boolean = false` has a default but `isChecked` does not, better to keep them the same and make it required.

Not blocking, but worth a follow up: `cashOnDeliveryState` keeps a copy of the row in the VM while `viewState.rows` keeps another one, so every update writes both and we get `cashOnDeliveryState.value!!` in 7 places plus the `getNonTogggleableItems()` filter and append. If the rows were derived from the COD state instead of stored, this `isLoading` flag would not need to be carried around at all. Pre-existing, not from this PR.

Two more things I noticed while reading, up to you if you want them here or separately:

- `isCashOnDeliveryEnabled()` returns `false` when the gateways fetch fails (`?: false`), so a failed request still shows the toggle as off. The spinner only covers the time before the response.
- `initialState` is rebuilt on every `onViewVisible()` from `cashOnDeliveryState.value!!`, which keeps `isLoading = false` after first load. So the spinner shows on first entry but not on later resumes. Maybe intentional to avoid flicker, but it does not look deliberate.

malinajirka:
Thanks for the review!

> One thing I'd change: on ToggleableListItem we now have isChecked + isLoading + isEnabled, so 8 possible combinations for a row that really has only 3 states. Can we make it one tri-state value instead of two booleans? Something like Loading / Checked / Unchecked. Then the view holder has to handle all cases explicitly, and right now it works only because an isInvisible switch does not receive touches, which is a bit implicit.

Good point, I've merged `isChecked + Loading` to ensure we can't represent invalid state. However, since `So the spinner shows on first entry but not on later resumes. Maybe intentional to avoid flicker, but it does not look deliberate.` is intentional, we need independent isEnabled state (moreover it comes from parent).

> Not blocking, but worth a follow up: cashOnDeliveryState keeps a copy of the row in the VM while viewState.rows keeps another one, so every update writes both and we get cashOnDeliveryState.value!! in 7 places plus the getNonTogggleableItems() filter and append. If the rows were derived from the COD state instead of stored, this isLoading flag would not need to be carried around at all. Pre-existing, not from this PR.
> Two more things I noticed while reading, up to you if you want them here or separately:
> isCashOnDeliveryEnabled() returns false when the gateways fetch fails (?: false), so a failed request still shows the toggle as off. The spinner only covers the time before the response.

https://linear.app/a8c/issue/WOOMOB-3782/payments-hub-cod-toggle-derive-rows-from-state-and-handle-gateway

PR #16350, opened 2026-08-03
Ask for confirmation before toggling Pay in Person

malinajirka:
Fixes [WOOMOB-3729](https://linear.app/a8c/issue/WOOMOB-3729/android-add-a-confirmation-dialog-when-toggling-pay-in-person-in-the).

The Pay in Person toggle wrote to the live store the moment it was tapped — and the whole row is a tap target, so a stray tap silently changed a payment method at checkout. It now asks for confirmation first. When enabling, the dialog names the gateway's current customer-facing title and explains it will be renamed to "Pay in Person"; that sentence is skipped when the title is already "Pay in Person", so merchants only see the rename warning when something actually gets renamed.

The switch itself is no longer directly togglable — the row click drives it — so it can't flip before the change is confirmed.

Built on top of #16349, review/merge that one first.

### Testing

Verified on device against a real store and with mocked `/wc/v3/payment_gateways` responses for each state:

1. Open **Payments**.
2. With the gateway disabled and its title still custom (e.g. "Cash on delivery"), tap the **Pay In Person** row — the dialog quotes that exact title and warns about the rename.
3. With the title already "Pay in Person", tap the row — the dialog shows the short message with no rename sentence.
4. With the gateway enabled, tap the row — the dialog warns that customers won't be able to select the method.
5. Tap **Cancel** in each case — the switch stays where it was and no request is made.
6. Tap the confirm button — the toggle flips and the gateway is updated.

| Enable with rename | Enable no rename | Disable |
|--|--|--|
| 1-enable-with-rename | 2-enable-no-rename | 3-disable |


kidinov (approved):
LGTM!

malinajirka:
@copilot Merge in trunk and resolve the conflicts.

PR #16324, opened 2026-07-29
[WOOMOB-3708] Add Orders Compose UI foundation

hichamboushaba:
### Description
Part of WOOMOB-3708

Adds the Compose UI foundation for the Orders list without changing the production route. This follows the same stateless-UI-first boundary as [PR #16297 — the Products Compose UI foundation](https://github.com/woocommerce/woocommerce-android/pull/16297). Two follow-up PRs will add the Paging 2 presentation seam and perform the runtime cutover.

> [!NOTE]
> This PR is intentionally larger than a typical foundation change. Its scope is almost entirely stateless Compose UI, previews, and the UI models that define their contract; none of it is wired into the production Orders route yet. Keeping these tightly related pieces together makes the complete UI foundation easier to review and avoids splitting screen, header, row, and state contracts across artificial PR boundaries.

 

#### Changes

- Adds overloaded stateless `OrderListScreen` and `OrderListContent` APIs covering browsing, active search, filtering controls, troubleshooting, selection chrome, loading and empty states, paging placeholders, and the Create Order FAB.
- Treats search as a dedicated mode: the Create Order FAB is hidden while search is active and restored when search closes. This intentionally aligns with WooCommerce iOS and Gmail rather than retaining the legacy Android create action during search, reducing unrelated actions and keyboard-constrained viewport pressure.
- Adds Orders row rendering with the legacy information hierarchy and visual parity:
  - Date on the first line.
  - Order number and customer name as the title, with the total trailing.
  - Existing status and POS colors through a controlled custom-color `WooBadge` API.
  - White row and skeleton surfaces, section headers, dividers, and responsive large-font behavior.
- Adds Compose swipe-to-complete presentation with the physical-left gesture, full-height reveal, RTL placement, and accessibility action.
  - Slow drags complete after crossing 50% of the row.
  - A directional fling uses Material 3's 125 dp/s settling threshold; a fast reversal cancels the action.
  - The legacy zero-velocity override was introduced only to protect the since-removed synthetic glance animation ([#7362](https://github.com/woocommerce/woocommerce-android/pull/7362), [#8026](https://github.com/woocommerce/woocommerce-android/pull/8026)).
- Adds immutable UI models and a mapper from the existing Orders list item types, with focused unit coverage for status, sales-channel, and metadata mapping.
- Documents the lazy-row accessor contract: mapped items are memoized per index and callers advance `contentRevision` when content changes without replacing the accessor.
[...]

PR #16351, opened 2026-08-03
[WOOMOB-3708] Add Orders Paging 2 presentation seam

hichamboushaba:
### Description
Part of WOOMOB-3708

This stacked PR builds on [PR #16324 — Orders Compose UI foundation](https://github.com/woocommerce/woocommerce-android/pull/16324) and introduces the isolated Paging 2 presentation seam needed by the later production cutover.

The existing Orders list still uses Paging 2. Compose needs an immutable, indexed snapshot for rendering while preserving `PagedList.loadAround()` hints, stable row identity, and callback-driven updates. `OrderListPaging2Presenter` provides that bridge without changing the current production route; PR3 will connect it to the screen.

#### Changes

- Adds `OrderListPaging2Presenter`, exposing immutable `StateFlow` snapshots of the active `PagedList`.
- Keeps item count, indexed reads, stable keys, and loaded order IDs tied to one captured presenter state.
- Triggers `loadAround(index)` only when reading through the currently published state, so stale snapshots remain side-effect free.
- Separates list `generation` from `contentRevision`:
  - Replacing the `PagedList` advances both.
  - Paging callbacks and same-instance item mutations advance only content revision.
- Uses stable, saveable keys for orders, loading rows, date sections, and unresolved placeholders.
- Owns `PagedList.Callback` attachment and removal across replacement and close.
- Publishes each matching snapshot before attaching its callback, keeping list and state authority coherent even if callback behavior changes later.
- Adds JVM coverage for same-instance mutation and resubmission, list shrink and clearing, placeholder identity, loaded-order helpers, stale callbacks, prefetch gating, StateFlow revisions, and terminal cleanup.

#### Scope

- The presenter is intentionally not called by production code at this boundary.
- `OrderListFragment`, `OrderListViewModel`, the RecyclerView renderer, navigation, analytics, and visible behavior remain unchanged.
- Migrating Orders to Paging 3 is out of scope; this seam preserves the current Paging 2 behavior for the Compose cutover.
- No release-notes entry is needed because this PR is internal infrastructure with no user-facing change.

#### Verification

- All nine `OrderListPaging2PresenterTest` cases pass.
- Wasabi debug production and unit-test Kotlin compilation pass as part of the focused test run.
- `detektAll --no-auto-correct` passes.
- `git diff --check` passes at the PR1 → PR2 boundary.

### Test Steps

1. Check out this PR on top of PR #16324, launch the app, and open Orders.
2. Confirm the existing Orders list still loads, scrolls, and paginates as before. No visual or behavioral change is expected because the presenter is not wired until PR3.

### Images/gif
N/A — this PR adds an internal, intentionally unwired presentation seam.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListPaging2Presenter.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListPaging2Presenter.kt
@@ -1,0 +1,165 @@
+package com.woocommerce.android.ui.orders.list
+
+import androidx.annotation.MainThread
+import androidx.paging.PagedList
+import com.woocommerce.android.ui.orders.list.OrderListItemUIType.LoadingItem
+import com.woocommerce.android.ui.orders.list.OrderListItemUIType.OrderListItemUI
+import com.woocommerce.android.ui.orders.list.OrderListItemUIType.SectionHeader
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.util.Collections
+
+@MainThread
+internal class OrderListPaging2Presenter : AutoCloseable {
+    private val mutableState = MutableStateFlow(State())
+    val state: StateFlow<State> = mutableState.asStateFlow()
+
+    private var currentList: PagedList<OrderListItemUIType>? = null
+    private var currentCallback: PagedList.Callback? = null
+    private var isClosed = false
+
+    fun submit(pagedList: PagedList<OrderListItemUIType>?) {
+        if (isClosed || pagedList === currentList) return
+
+        removeCurrentCallback()
+        currentList = pagedList
+        val previousState = mutableState.value
+        val generation = previousState.generation + 1
+        val contentRevision = previousState.contentRevision + 1
+
+        if (pagedList != null) {
+            val callback = callbackFor(pagedList)
+            currentCallback = callback
+            pagedList.addWeakCallback(null, callback)
+        }
+
+        mutableState.value = pagedList.toState(
+            generation = generation,
+            contentRevision = contentRevision,
+        )
+    }
+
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>State</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListPaging2Presenter.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    fun loadedOrderIds(capturedState: State): List<Long> {
+        val orderIds = ArrayList<Long>()
+        capturedState.items.forEach { item ->
+            if (item is OrderListItemUI) {
+                orderIds += item.orderId
+            }
+        }
+        return Collections.unmodifiableList(orderIds)
```
AdamGrzybkowski: ⛏️  Wouldn't that be enough? The return type is already a Kotlin read-only List.
```suggestion
return capturedState.items
            .filterIsInstance<OrderListItemUI>()
            .map { it.orderId }
```

AdamGrzybkowski (approved):
Damn, I don't remember when I had to use Paging2 :D Looks good, only left one suggestion! :shipit:

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/list/OrderListPaging2Presenter.kt
hichamboushaba: Applied in 8cf299807ed. The two forms are not runtime-equivalent because Kotlin List is only read-only at the type level, but this is safe here: the result is detached navigation context, no caller mutates it, and it is copied to a LongArray before navigation.

PR #16352, opened 2026-08-03
[WOOMOB-3735] Upload Zendesk ticket attachments concurrently

malinajirka:
### Description

Fixes WOOMOB-3735

`ZendeskTicketRepository.createRequest()` uploaded the ticket attachments one after another, so submitting a support request waited for the sum of all three round trips instead of the slowest one — and in the worst case burned 3× `attachmentUploadTimeout`. This runs them in a `coroutineScope` with `async`/`awaitAll` instead.

`uploadTextAttachment()` already returns `null` on failure or timeout rather than throwing, so the best-effort semantics survive unchanged: one failing upload can't cancel its siblings or block ticket creation. Attachment order is preserved too, since `awaitAll` keeps the list order.

One existing test needed its `verify` narrowed from `any()` to `eq("application_log.txt")` — with concurrent uploads two calls are recorded by the time it runs, so the unfiltered matcher hit `TooManyActualInvocations`.

### Test Steps

1. Open **Settings → Help & Support → Contact Support**, fill in a subject and description, and submit.
2. Confirm the ticket is created successfully and that submission feels faster than before.
3. Turn off networking mid-submission (or use a flaky connection) and submit again — the ticket should still be created, just without the attachments that failed.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/support/zendesk/ZendeskTicketRepository.kt
@@ -20,7 +20,10 @@ import com.woocommerce.android.util.WCSSRModelCachingFetcher
 import com.woocommerce.android.util.WooLog
 import com.zendesk.service.ErrorResponse
 import com.zendesk.service.ZendeskCallback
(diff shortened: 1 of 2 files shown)


samiuelson (approved):
LGTM

PR #16354, opened 2026-08-03
[WOOMOB-2734] Fix duplicated device registration API calls on login

samiuelson:
### Description
Fixes WOOMOB-2734

**Problem:**
`RegisterDevice` serializes runs with an internal mutex and skips WPCom registration when a device id is already stored, but the class was unscoped in Hilt, so every injection point  got its own instance with its own mutex, and concurrent triggers raced past the "already registered" check (the device id is only persisted after a successful response) and each fired its own call. This PR scopes `RegisterDevice` as a `@Singleton` so that concurrent calls now queue, and follow-up runs no-op against the persisted state.

The same check also races within a single run: per-site Woo push registration falls back to WPCom registration when it fails, so several stores failing at once each fired an identical call before the first response persisted the device id. The fallback is now dropped from `RegisterDevice` runs, which already register WPCom once at the end of the run.

### Test Steps
1. Log out, then log in with a WP.com account that has multiple Woo stores available and select a store.
2. Watch network traffic during the whole login flow: exactly one `POST .../devices/new/` should fire, and subsequent registration runs should log "Skipping WP.com push registration".
3. Background and foreground the app a few times - no further `devices/new` calls.
4. Sanity-check re-registration still works: clear app data (or reinstall), log in again - a single registration call fires.

### Images/gif
N/A


irfano:
@samiuelson, do you plan to open this PR this week? I’m asking because this is the last remaining task from the PN project.

samiuelson:
> @samiuelson, do you plan to open this PR this week? I’m asking because this is the last remaining task from the PN project.

Hi @irfano 👋 I wasn't able to repro this issue manually on device, only in unit test. If you're able to reproduce the bug and validate this PR fixes it it would be great. Otherwise, feel free to take over this issue because I don't want to block you.

irfano (approved):
I was able to reproduce this on trunk: logging in with a multi-store account while watching traffic gives 5 identical `devices/new` calls. Your branch drops it to 4, so the singleton fix works, and the remaining duplicates all come from the per-site WPCom fallback you split out in 7177ce26ae2. Since that turned out to be the main source of the duplicates, I restored it here by reverting that commit and updated the description. With both changes in, login fires exactly one call.

PR #16358, opened 2026-08-04
[WOOMOB-3661] Use the standard decimal input for custom amounts

malinajirka:
### Description

Fixes WOOMOB-3661

The custom amount price field was the only money field in the app using the right-to-left "cents" formatting, so typing `250` produced `2.50` while the shipping field right next to it read the same keystrokes as `250`. Easy to bill someone $500 when you meant $5. Dropping `usesFullFormatting` from the field lines it up with shipping, fees and product price, and with iOS.

The state observer in `CustomAmountsFragment` compared the raw field text against the price as a string. That never matches mid-entry with the regular input — typing `2.` parses to `2`, and locales using `,` as the separator never match at all — so it would overwrite the decimal separator as soon as it was typed. It now compares `BigDecimal`s, the same way the shipping field's `setValueIfDifferent` does.

Trade-off: the custom amount input loses thousand separators while typing (`1000000`, not `1,000,000`). No other money input in the app has them either.

Second commit removes unused code.

### Test Steps

1. Open an order (new or existing) → Add Custom Amount → Fixed amount.
2. The field starts at `0`, not `0.00`.
3. Type `250` — the amount is `250`, not `2.50`.
4. Clear it and type `2.50` — the decimal point survives, including the intermediate `2.` state, and the cursor stays at the end.
5. Add the custom amount, then add a Shipping line and type `2.50` there. Both fields should behave identically.
6. Reopen the saved custom amount for editing — it pre-fills as `2.50`.
7. Add a custom amount of `1000000` and confirm the amount shown on the order line and in the totals is still grouped (`1,000,000.00` on a US store).
8. Percentage custom amounts are unchanged.

### Images/gif

https://github.com/user-attachments/assets/d0a14684-a7bd-4219-a7f0-8d1ee2e626be


samiuelson (approved):
LGTM!

PR #16359, opened 2026-08-04
[WOOMOB-3747] Fix card reader payment crash when the payment screen is recreated

kidinov:
### Description

Fixes WOOMOB-3747

`start()` runs again when the payment dialog's view is recreated, for example on rotation, while the controller survives. It always called `scope.cancel()`, which caused two bugs during a payment:

- Cancelling an in-flight Stripe operation makes Stripe throw from its cancellation handler. Coroutines pass that to the uncaught exception handler, so the app crashes. 23 users in Sentry.
- The cancelled payment job was never restarted, so the screen froze while the charge could still go through.

`start()` now returns early while a payment or refund is still running, which fixes both. `stop()` still has to cancel, so the scope now has a `CoroutineExceptionHandler`. The old `try/catch` in `stop()` never worked and is removed, because coroutines never throw that exception to the caller.

The added unit tests reproduce both bugs. All three fail on trunk and pass with this change.

Notes for review:

- The handler catches any `CompletionHandlerException`, not only Stripe's. Narrowing it would need Stripe types in the app module or matching on the message.
- The exception no longer crashes the app, so it is logged to `WooLog` and sent to Sentry as a non-fatal.
- POS is not affected. It creates a new controller before every `start()`.

### Test Steps

1. Open an order and start a card payment.
2. Rotate the device while the payment is being processed.
3. The app does not crash, the screen keeps updating, and the payment completes.
4. Start another payment and press back while it is being processed. The app does not crash.
5. Check a normal payment still works, and cancelling one before tapping a card.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
@@ -136,6 +136,11 @@ class CardReaderPaymentController(
     }
 
     fun start() {
+        // The view can be recreated while a payment is in flight (rotation, dialog recreation). Cancelling the
+        // scope here would kill the running flow without restarting it, and cancelling an in-flight Stripe
+        // operation throws from its cancellation handler.
+        if (paymentFlowJob?.isActive == true || refundFlowJob?.isActive == true) return
+
         scope.cancel()
         scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
         if (cardReaderManager.readerStatus.value is CardReaderStatus.Connected) {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
@@ -71,6 +71,7 @@ import com.woocommerce.android.util.PrintHtmlHelper.PrintJobResult.FAILED
 import com.woocommerce.android.util.PrintHtmlHelper.PrintJobResult.STARTED
 import com.woocommerce.android.util.WooLog
 import kotlinx.coroutines.CompletionHandlerException
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.InternalCoroutinesApi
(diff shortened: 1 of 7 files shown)


WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentControllerTest.kt
```diff
@@ -2511,19 +2511,73 @@ class CardReaderPaymentControllerTest : BaseUnitTest() {
...
+            // WHEN
+            controller.start()
+            paymentStatus.value = ProcessingPayment
+
+            // THEN
+            assertThat(controller.paymentState.value)
+                .isInstanceOf(CardReaderPaymentState.ProcessingPayment::class.java)
+        }
+
+    private suspend fun stubPaymentThatThrowsOnCancellation() {
```
malinajirka: 🔍 np: We might consider moving these two methods to the top/bottom of the file.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
```diff
@@ -115,7 +116,17 @@ class CardReaderPaymentController(
...
     private val allowCancelledStatus: Boolean = false,
 ) {
-    private var scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
+    @OptIn(InternalCoroutinesApi::class)
+    private val scopeExceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        // Stripe Terminal throws from its coroutine cancellation handler when the operation is still waiting for
+        // a network response. Coroutines report that as CompletionHandlerException through this handler instead
+        // of throwing it to the caller, so it cannot be caught around scope.cancel(). Anything else reaching
+        // here is a real failure and has to keep crashing.
+        if (throwable !is CompletionHandlerException) throw throwable
```
malinajirka, line 126: ⚠️ This feels a bit risky to me since we're consuming all CompletionHandlerException no matter where they come from — it's not a Stripe type, it's what coroutines wrap any misbehaving cancellation/completion handler in. 

Might be worth reporting it to Sentry as a non-fatal, wdyt?  `CrashLogging.sendReport(exception, tags, message)`

malinajirka (approved):
LGTM overall, I've left one comment to consider to make sure we have reports in Sentry in case we are missing something.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentControllerTest.kt
kidinov: Fixed in d7d3632

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/payment/controller/CardReaderPaymentController.kt
kidinov: Good point, fixed in 2750b71. It still swallows the exception, but now sends it as a non-fatal so we keep visibility if a non-Stripe handler ends up here.

PR #16361, opened 2026-08-04
[WOOMOB-3754] Block subscription products from reaching an order via bundles and scanning

malinajirka:
### Description

Fixes WOOMOB-3754

Order creation doesn't support subscription products, but the rule lived only in `ProductSelectorViewModel.getProductSelection`, so it applied to the product list and nothing else. Two paths went around it: a bundle holding a subscription child could be added with no warning, and scanning a subscription product's barcode or SKU added it straight to the order.

The rule now lives in `ProductRestriction.SubscriptionProducts` alongside the other restrictions, so the product list, the scanner and the bundle check all read the same definition. Restrictions gained an optional `nonSelectableReason`: those that carry one keep the product visible and explain why it can't be picked (what subscriptions already did), those that don't keep removing it from the list, which is what every existing restriction does.

For bundles, whether a child is a subscription can't be answered while building the list — the API's `bundled_items` has no product type and a bundle's children usually aren't cached — so it's resolved when the merchant picks the bundle rather than per row. That costs one resolve per bundle actually tapped and avoids a window where a row renders enabled before flipping to disabled. Since that resolve can take a request, the row shows a spinner and ignores further taps until it finishes, otherwise a second tap during the wait would fire the navigation twice. Both entry points are covered: the row tap and the **Configure** button, which go through different ViewModel methods.

One incidental fix: `ProductSelectorFragment` never handled `ShowSnackbar`, so the refusal had nowhere to appear.

### Test Steps

Needs a store with the Product Bundles and Subscriptions extensions.

1. Create a new order and tap **Add products**.
2. Search for a subscription product. It's greyed out with *"Subscription products are not supported for order creation"* — unchanged behaviour, worth confirming it didn't regress.
3. Find a bundle that contains a subscription product. Tap the row: nothing is selected and a snackbar reads *"This bundle contains a subscription product, which is not supported for order creation"*.
4. Tap that bundle's **Configure** button. Same snackbar, and the configuration screen does not open.
6. Find a bundle with no subscription products. It still selects and configures normally.
7. Tap **Scan products** and scan a subscription product's barcode (or its SKU). It isn't added, and the failure message explains why.
8. Scan a normal product to confirm scanning still works.

P.S. Can be tested on https://site-for-woocommerce12a3fasdf45dfs6789.mystagingwebsite.com/ with AAA Product Bundle 2.

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 25.4
 -----
 - [*] Fixed bundled variable products not asking for a variation when configuring a bundle during order creation
+- [*] Bundles containing a subscription product can no longer be added to an order, and scanning a subscription product now explains why it can't be added
 - [Internal] Attach a Mobile Status Report (device, OS, connectivity, notifications, feature flags) to support tickets
 - [*] Fixed fully refunded orders remaining visible in the Completed orders filter [https://github.com/woocommerce/woocommerce-android/pull/16313]
 - [*] Fixed coupon expiry dates drifting when unrelated coupon changes were saved for stores outside UTC [https://github.com/woocommerce/woocommerce-android/pull/16279]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/model/BundledProduct.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/model/BundledProduct.kt
@@ -2,6 +2,7 @@ package com.woocommerce.android.model
 
 import android.os.Parcelable
 import com.woocommerce.android.ui.products.ProductStockStatus
+import com.woocommerce.android.ui.products.ProductType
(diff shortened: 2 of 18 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

malinajirka:
Replaced with https://github.com/woocommerce/woocommerce-android/pull/16380

PR #16363, opened 2026-08-04
[WOOMOB-3712] Fix POS Scan to Pay false success and non-fullscreen success screen

kidinov:
### Description

@samiuelson I could not validate that the success path works as on the web it doesn't collect the money and shows an error "canot read properties of null". Both here and on trunk. Did it work before?

Fixes WOOMOB-3712

Two fixes for POS Scan to Pay:

1. Picking "Pay in Person" in the QR webview showed `Order Successfully Paid`. Polling treated `processing`/`completed` as paid, but that order looks like this:

   ```
   status=processing datePaid=null isOrderPaid=false paymentMethod='cod'
   paymentMethodTitle='Pay in Person' isCashPayment=true
   ```

   Online gateways stamp `datePaid` when the money arrives; offline ones only advance the status. So success now requires `datePaid`, and a paid status *without* it is treated as "customer picked an offline method" — keyed on status rather than the gateway id, so it holds for any offline gateway. That state stops polling and explains what happened, offering *Collect on the register* or *Show the code again*. Without it the merchant would sit on the QR for the full 5.5 min polling window and then get a generic error.

   *Collect on the register* sends `BackFromCheckoutToCartClicked`, returning to the cart so the next checkout builds a fresh order — the customer already placed the offline one, so collecting against it would conflate two sales.

2. The success screen kept the cart visible and played no cha-ching. Scan to Pay sent `OrderSuccessfullyPaid` straight down to children, skipping `WooPosHomeViewModel.onOrderSuccessfullyPaid()` which sets fullscreen totals and plays the sound. It now reports upward like the card and external flows.

While wiring that up, the three child-to-parent success events (`OrderSuccessfullyPaidByCard`, `...Externally`, plus the new Scan to Pay one) collapsed into a single `ChildToParentEvent.OrderSuccessfullyPaid(paymentMethod)`.

Offline gateways that land on `on-hold` (bacs, cheque) aren't detected and still fall through to the polling timeout — only `processing`/`completed` are treated as an offline checkout.

### Test Steps

Requires a debug build (`WOO_POS_SCAN_TO_PAY` is off in production).

1. Enable Cash on delivery / Pay in Person in the store's payment settings.
2. In POS, add an item, go to checkout, tap Scan to Pay.
3. Scan the QR, choose "Pay in Person" and place the order.
4. POS shows "Customer chose to pay in person" — no success screen

Note: the store's new-order push notification uses the same cha-ching audio file, so the sound alone doesn't tell you which fired — watch the screen.

### Images/gif

The UI of this error is just for testing - should be redone when ready

image

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
@@ -39,6 +39,7 @@ sealed class ChildToParentEvent {
     data object GoBackToCheckoutAfterFailedPayment : ChildToParentEvent()
     data object OrderSuccessfullyPaidByCard : ChildToParentEvent()
     data object OrderSuccessfullyPaidExternally : ChildToParentEvent()
+    data object OrderSuccessfullyPaidViaScanToPay : ChildToParentEvent()
     data object ExitPosClicked : ChildToParentEvent()
     data object SetupBarcodeScannerClicked : ChildToParentEvent()
     data object CouponsValidationFailed : ChildToParentEvent()
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/home/WooPosHomeChildToParentCommunication.kt
@@ -1,5 +1,6 @@
 package com.woocommerce.android.ui.woopos.home
 
+import com.woocommerce.android.ui.woopos.home.ParentToChildrenEvent.OrderSuccessfullyPaid.PaymentMethod
 import com.woocommerce.android.ui.woopos.home.cart.WooPosCartItemViewState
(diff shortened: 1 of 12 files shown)


samiuelson (commented):
Thanks for working on this @kidinov. This PR works as expected. I have, however, question about the following checkout scenario:

1. Scan the QR code, choose Pay in Person, and place the order.
2. Wait for POS to show Customer chose to pay in person.
3. Press Android Back or the top-left back button instead of Collect on the register.

 At that point order is in the "processing" state. I think in case user discards the order at that point, for example, going by back to Home, we should delete the order (or revert its status to "auto-draft" so that it's auto removed and not visible in wp-admin). Otherwise, the order will be visible with processing state while the user attempted to discard it in POS. Wdyt?

kidinov:
@samiuelson thanks for testing it. I thought about it more and I would not revert or delete anything here.

Once the customer placed the order it is a real order - stock is reduced, emails are sent, it has a number. So I think it is fine that it stays visible in `processing`, it is a record of what actually happened. If the merchant does not want it, they can cancel or delete it manually in wp-admin, explicitly. Doing it automatically from POS on a back press feels wrong to me, we would silently throw away a record the customer already got a confirmation email about.

So I would leave this as is and not add anything to this PR. Wdyt?

samiuelson (approved):
LGTM

PR #16364, opened 2026-08-04
[WOOMOB-3714] Restore back navigation in location pickers

hichamboushaba:
### Description

Fixes WOOMOB-3714

Adds a dedicated toolbar to the shared Country and State picker so back navigation remains visible when the picker opens in the Orders detail pane on tablets.

### Test Steps

1. Open an order that can create a shipping label on a tablet in landscape.
2. Tap **Create shipping label**, expand **Shipment details**, and edit the shipping address.
3. Open **Country** and confirm the picker shows a back arrow in the detail pane.
4. Tap the back arrow and confirm it returns to the address editor.
5. Repeat steps 3–4 for **State**.

### Images/gif
woomob-3714-country-picker


AdamGrzybkowski (approved):
LGTM :shipit:

PR #16365, opened 2026-08-04
[POS] Harden the server-calculated refund flow

samiuelson:
WOOMOB-3749

Follow-ups from the review of #16335. None of these block the WC 11.1.0 rollout; they were deliberately kept out of that PR to keep it focused. Stacked on #16338, which this branch's base points at.

### Description

**Correctness — the availability cache outlived the store it described.** A verdict was recorded per store with no record of the WooCommerce version it was probed against. A `false` from a pre-11.1 probe survived the store upgrading and kept a now-capable store on local math until the process restarted. Worse, a `true` survived a *downgrade*, which would allow a `compute_totals` create against a store that silently drops the parameter — the ghost zero-amount refund the whole preview-before-create design exists to prevent.

Each verdict now carries the version it was probed against and only answers for that version, so a store that changes version mid-session is re-probed. To make that impossible to get wrong at the call sites, `WooPosRefundFlow.ServerComputed` became a data class carrying `wooVersion`, and every cache read and write goes through it.

**Diagnosability — the resolver was silent.** A store on local calculation left no trace of why, and since an unknown version fails closed, that path was invisible. It now logs the resolved flow and the reason (flag / version / cache).

**Simplification.** Eligibility was evaluated in the ViewModel and again inside `WooPosRefundPreview`, which routed both outcomes to the same place. The redundant ViewModel pre-check is gone, leaving one decision point per phase and shrinking the window in which the decision can change mid-flow.

**Smaller items.** Dropped the write-only `maxRefundable` state — the server enforces the cap and reports `preview_exceeds_max_refundable`, so a client-side copy was dead weight and was never cleared on local fallback. Used `toPlainString()` for the money string. Corrected KDoc that described both flows as "v3" and that claimed `LocalComputed` was only for old stores.

**Tests added** for the gaps the review found: prerelease version resolution (`11.1.0-dev`, which trunk builds report), verdicts across a version change in both directions, the deny side of the create gate, Interac followed by a computed create, and error propagation from a failed computed create.

One test assertion changed meaning: it previously verified the ViewModel short-circuits *before* consulting the preview. That pre-check is the redundancy removed here, so it now verifies the preview is consulted; that no network call happens for an unavailable store is asserted in `WooPosRefundPreviewTest` rather than duplicated.

### Test Steps

1. On a store running WooCommerce 11.1.0+ with the `woo_pos_refund_v4` flag on, issue a POS refund by item selection. Totals come from the server preview and the created refund shows server-computed amounts in wp-admin.
2. Check the Application log (Settings → Help & Support) filtered to `POS`: the resolved flow and its reason are now logged on every refund.
3. On a store below 11.1.0, or with the flag off, the flow falls back to local calculation with no user-visible difference.
4. Version-change behaviour: refund on a pre-11.1 store (verdict recorded as unavailable), upgrade the store, then refund again without restarting the app — it should re-probe and use the server flow rather than staying local.

Unit coverage: 186 tests green across the `woopos.orders.details.refund` package, `detektAll` clean.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.4
 -----
+- [Internal] POS refunds: server-refund availability is now tracked per WooCommerce version, so a store that changes version mid-session is re-probed instead of reusing a stale verdict [https://github.com/woocommerce/woocommerce-android/pull/16365]
 - [*] Fixed fully refunded orders remaining visible in the Completed orders filter [https://github.com/woocommerce/woocommerce-android/pull/16313]
 - [*] Fixed coupon expiry dates drifting when unrelated coupon changes were saved for stores outside UTC [https://github.com/woocommerce/woocommerce-android/pull/16279]
 - [*] Card payments no longer show an error when the payment was actually captured but the backend reported a network/server issue — the app now verifies the payment status and shows success, preventing double charges [https://github.com/woocommerce/woocommerce-android/pull/16295]
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundPreview.kt
@@ -29,7 +29,8 @@ class WooPosRefundPreview @Inject constructor(
         val site = selectedSite.get()
         val localSiteId = site.localId().value
 
-        if (resolveRefundFlow() is WooPosRefundFlow.LocalComputed) {
+        val flow = resolveRefundFlow()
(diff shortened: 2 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson:
Consolidated into #16335 as 6e0abe9c9c3: the hardening changes now ride in the root PR and #16338 stays stacked on it. Closing without merge.

PR #16367, opened 2026-08-05
[WOOMOB-3757] Fall back to an RSA certificate when the platform rejects the ECDSA one

kidinov:
### Description

Fixes WOOMOB-3757

The Remote TTP phone session builds its own TLS identity with `HeldCertificate.Builder().ecdsa256().build()`. okhttp-tls encodes the certificate DER itself and round-trips it back through the platform `CertificateFactory`, so on a few devices Conscrypt refuses to parse the EC certificate okhttp just generated and `build()` throws `IllegalArgumentException: failed to decode certificate`. That happens before the session opens a socket, so the phone never advertises and the tablet has nothing to discover — with no recovery path, one throw killed the session permanently. This adds a one-shot retry with RSA-2048, keeping ECDSA first since it works everywhere else and is faster.

**The root cause is inferred from a device correlation, not reproduced.** All reported errors come from one model (`SM-S938B`, Android 16 / SDK 36) and, within it, two specific handsets that fail 100% of the time while other units of the same model work fine. That fits a Play-Services-updated Conscrypt disagreeing with the EC key encoding, but I don't have an affected handset. The fallback is cheap and low-risk either way.

Worth a reviewer's attention:

- Contrary to how it reads, okhttp's `HeldCertificate.Builder` **defaults to `ecdsa256()`** in its `init` block, so the fallback has to call `.rsa2048()` explicitly rather than just dropping the `.ecdsa256()` call.
- RSA changes nothing downstream: the tablet pins the certificate by SHA-256 of its encoding, which is algorithm-agnostic.
- The key type used is reported as `certificate_key_type` on the existing `remote_ttp_phone_session_started` event rather than as a new event, so one query gives both numerator and denominator.
- When RSA also fails, the message flattens the cause chain so `error_description` carries Conscrypt's actual reason. Deliberately scoped to the certificate path — the general `CardReaderRemoteSession` cause-chain fix belongs to WOOMOB-3756, and keeping it out of that file avoids duplicating work.

### Testing

Verified on real hardware — a phone as reader, a Redmi tablet as POS, over wifi_lan.

**ECDSA path (unchanged behaviour):** `remote_ttp_phone_session_started` carries `certificate_key_type: ecdsa_256`.

**RSA fallback path:** forced by temporarily making the certificate factory throw `IllegalArgumentException("failed to decode certificate")` for ECDSA — the same exception okhttp raises on the affected devices — then pairing and taking a payment end to end:

```
15:24:48  ECDSA certificate rejected, retrying with RSA: java.lang.IllegalArgumentException: failed to decode certificate
15:24:49  remote_ttp_phone_session_started  certificate_key_type: rsa_2048
15:24:51  TLS accepted, starting protocol loop
15:24:56  Connected to reader
15:25:24  pos_card_present_collect_payment_success  transport: wifi_lan
```

So the fallback fires, the tablet's SHA-256 fingerprint pinning accepts an RSA certificate, and a payment completes. The patched build ran on the phone only, against an unmodified tablet, so this also confirms an RSA phone interoperates with a tablet that knows nothing about the change.

**Still not verified:** that RSA actually succeeds on a device where the platform genuinely rejects ECDSA. That is the underlying hypothesis and it needs an affected handset — forcing the branch on healthy hardware cannot prove it. Once shipped, `certificate_key_type` on `remote_ttp_phone_session_started` answers it directly from Tracks.

### Test Steps

1. On a phone, go to Settings → Payments → Card Reader Mode and start a session. Confirm it reaches "ready to pair" with a pairing code.
2. On a POS tablet on the same Wi-Fi and the same store, pair with the phone and take a Tap to Pay payment through to completion.
3. Confirm `remote_ttp_phone_session_started` carries `certificate_key_type: ecdsa_256`.
4. To exercise the fallback, temporarily swap `ecdsa256()` for a `throw IllegalArgumentException("failed to decode certificate")` in `CardReaderRemoteTlsServer`'s default certificate factory, then repeat steps 1–2. Pairing and payment should still work, and the started event should report `certificate_key_type: rsa_2048`.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/cardreader/src/test/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+            logWrapper = logWrapper,
+            heldCertificateFactory = { keyType -> buildCertificate(keyType) },
+        )
+
+        // WHEN
+        server.start()
+
+        // THEN
+        assertThat(server.certificateKeyType).isEqualTo(CardReaderRemoteCertificateKeyType.ECDSA_256)
+        server.close()
```
malinajirka: 🔍 np: `server.close()` as the last statement means a failing assertion leaks the real `SSLServerSocket` for the rest of the test JVM. `server.use { ... }` would cover it for free (same in the fallback test below).

malinajirka (approved):
LGTM, one small np comment.

Note: I tested only the happy path - ECDSA.

libs/cardreader/src/test/java/com/woocommerce/android/cardreader/remote/CardReaderRemoteTlsServerTest.kt
kidinov: Good catch, fixed in 939f53e - both tests now wrap the server in `use`.

kidinov:

PR #16368, opened 2026-08-05
[WOOMOB-3756] Close the Remote Tap to Pay analytics gaps

kidinov:
### Description

Fixes WOOMOB-3756

Remote TTP (phone acting as a card reader for a POS tablet over `wifi_lan`) is instrumented badly enough that we can't diagnose the phone-to-tablet drop-off. This adds `transport` to `pos_reader_ready_for_card_payment`, keeps the cause chain in error descriptions instead of collapsing every failure to one generic string, clears the stale `card_reader_model` on disconnect, and derives `remote_ttp_phone_session_ended.reason` instead of hardcoding `user_exit`.

Worth a reviewer's attention:

- **`pos_reader_ready_for_card_payment` still never fires in the remote flow.** It is emitted on `CardReaderPaymentState.ProcessingPayment`, and `WooPosRemoteReaderPaymentFlow` never emits that state. The `transport` property added here fixes the Bluetooth case, but the wifi_lan connection→payment blind spot this issue was opened for is still open. Confirmed on device — see testing below. Tracked separately rather than pulled into this PR, since fixing it means changing what states the remote flow emits.
- `ReaderReadyForCardPayment` went from `data object` to `data class`. The old code mutated the singleton via `apply { addProperties(...) }`, so properties accumulated across payments for the process lifetime.
- `session_ended.reason` only emits `error`, `user_exit` and `dismissed`. There is no timeout or tablet-disconnect teardown path in the session, so inventing those values would be fabrication. The new `last_state` and `tablet_connected` properties carry that factually instead — a session ending at `ready_to_pair` with `tablet_connected: false` is the 94% case the issue is about.
- Out of scope: missing `amount` on POS payment events, `transport` absent on 28% of POS payment successes, and registering the four Remote TTP events in the Tracks directory.

### Testing

Verified end to end on real hardware — a phone as reader, a Redmi tablet as POS, over wifi_lan, through to a **successful payment**:

- `pos_card_reader_connection_tapped` / `_connection_success` / `_auto_connection_started` → `transport: wifi_lan`
- `pos_card_present_collect_payment_success` → `transport: wifi_lan`, and **no `card_reader_model`**, where trunk reports a stale `STRIPE_M2` left over from a prior Bluetooth connection
- `pos_reader_ready_for_card_payment` → did not fire at all, per the first bullet above

Not covered: the `session_ended` reason variants, and the Bluetooth path for `pos_reader_ready_for_card_payment`.

### Test Steps

1. Connect a tablet to a phone reader over wifi_lan and take a payment. Confirm `pos_card_present_collect_payment_success` carries `transport: wifi_lan`.
2. Connect a Bluetooth reader first, disconnect it, then connect the phone reader and take a payment. Confirm the payment events report no `card_reader_model` rather than the stale Bluetooth model.
3. Make the remote payment fail (kill the phone app mid-payment). Confirm `pos_card_present_collect_payment_failed` carries a specific `error_description` (`remote_failed: ...`, `remote_rejected: ...` or `remote_timed_out: ...`) instead of "Something went wrong. Please try again."
4. On the phone, exit reader mode via the on-screen button, then repeat with the system back button. Confirm `remote_ttp_phone_session_ended` reports `user_exit` and `dismissed` respectively, both with `last_state` and `tablet_connected`.
5. Take a payment with a Bluetooth reader and confirm `pos_reader_ready_for_card_payment` carries `transport: bluetooth`.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -477,6 +477,18 @@ class WooPosCardReaderConnectionController(
...
     }
 
+    private fun clearReaderTrackingInfo() {
+        cardReaderTrackingInfoKeeper.setTransport(null)
+        cardReaderTrackingInfoKeeper.setCardReaderModel(null)
+    }
+
+    // Must run before any phone event is tracked, or a model left over from a previous Bluetooth
+    // connection gets reported against this wifi_lan session.
+    private fun markPhoneReaderSelected() {
```
samiuelson, line 496: I guess we should also reset battery level here, otherwise the last known battery level will be tracked.

samiuelson: Probably worth resetting it in ::clearReaderTrackingInfo too.

kidinov: Both good catches, fixed in 81cda40. `setCardReaderBatteryLevel` is now reset in `clearReaderTrackingInfo()` and in `markPhoneReaderSelected()`.

The Bluetooth side had the same leak, so I added `markBluetoothReaderSelected()` to mirror it - `onConnectToReaderClicked()` and the auto-connect path set the model and transport but left the previous reader's battery in place, and `trackOnConnectTapped` fires before `CardReaderStatus.Connected` overwrites it.

samiuelson (approved):
LGTM!

PR #16369, opened 2026-08-05
[WOOMOB-3639] Hide empty Add more details section on complete products

AdamGrzybkowski:
### Description
Fixes WOOMOB-3639

The "Add more details" section on the product detail screen stayed visible even when there was nothing left to add.

Its visibility was already derived from the bottom-sheet item list (`showBottomSheetButton = list.isNotEmpty()`), and that list already depended on two inputs: the product draft and the product's displayable custom fields — which are stored separately from the product (in the metadata store) and typically load *after* it. The bug was in *when* the list got recomputed: it was rebuilt only on product-draft changes, so when the custom-fields metadata arrived without a corresponding draft change, the stale "Add custom fields" entry kept the section visible — until an unrelated rebuild happened to run (e.g. tapping into a field and navigating back).

Two changes, doing two different things:

- **ViewModel — the actual fix:** the cards and bottom-sheet list are now rebuilt whenever *either* input changes, by `combine`-ing the product-draft stream with the displayable-custom-fields stream. The section updates as soon as the metadata loads, regardless of the order the two arrive in.
- **Fragment — hardening:** `showProductDetails()` no longer force-shows the section on every render; visibility is driven solely by the computed list, so a later draft edit can't re-show an empty section. On its own this would **not** fix the reported bug (a stale `showBottomSheetButton` would still keep the section shown) — it removes a second, conflicting writer of the container's visibility.

### Test Steps
1. On the connected store, have a product that has **all** optional details populated, **including custom fields**.
2. Open **Products** and tap that product.
3. Scroll to the bottom: the **Add more details** section should **not** be shown on first render (no need to navigate away and back), and the **Custom fields** card should be visible.
4. Open a product that is still missing some optional details: the **Add more details** section should still be shown.
5. Edit a field on the fully-populated product and confirm the section does not reappear.

### Images/gif

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/1f2b408c-a0d2-45e6-a8ff-a509744f5666"></video> | <video src="https://github.com/user-attachments/assets/b0f75951-6fb9-4124-80e8-2fdef5359a9c"></video> |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailViewModel.kt
```diff
@@ -424,6 +424,27 @@ class ProductDetailViewModel @Inject constructor(
...
+    }
+
+    /**
+     * The detail cards and the "add more details" section depend on two independent sources: the product draft
+     * and the product's displayable custom fields (stored separately from the product). Combining them here
+     * rebuilds the cards whenever either changes, so the section reflects the latest state regardless of the
+     * order in which the product and its custom fields load.
+     */
+    private fun observeCardData() {
+        val hasDisplayableCustomFields = (navArgs.mode as? ProductDetailFragment.Mode.ShowProduct)
```
hichamboushaba: This observer is selected from the immutable `navArgs.mode`. In the add-new flow, saving the product replaces the draft sentinel ID with a real remote ID, but the mode remains `AddNewProduct`. This branch therefore remains `flowOf(false)`, so later custom-field additions or deletions do not rebuild the cards and bottom-sheet list until another draft change.

Could we derive the observed metadata stream from the current draft instead?

```kotlin
private fun observeCardData() {
    draftChanges
        .filterNotNull()
        .flatMapLatest { draft ->
            if (draft.remoteId == DEFAULT_ADD_NEW_PRODUCT_ID) {
                flowOf(draft)
            } else {
                customFieldsRepository.observeDisplayableCustomFields(draft.remoteId)
                    .map { it.isNotEmpty() }
                    .distinctUntilChanged()
                    .map { draft }
            }
        }
        .onEach(::updateCards)
        .launchIn(viewModelScope)
}
```

This keeps the subscription aligned when an add-new draft acquires its remote ID.

hichamboushaba: You can reproduce this with:

1. Start creating a new product.
2. Save it and stay on the product detail screen. The draft now has the new remote ID, but `navArgs.mode` is still `AddNewProduct`.
3. Open **Add more details → Custom fields**.
4. Add a custom field and go back.
5. The **Custom fields** card is not shown and **Add custom fields** is still in the bottom sheet.
6. Edit any regular product field. The cards rebuild and the UI becomes correct.

Expected: step 4 should update the cards and bottom-sheet list immediately.

AdamGrzybkowski: Thanks, fixed in https://github.com/woocommerce/woocommerce-android/pull/16369/commits/c8978d46bd5cef0768ae2b53dd1384b58dbe8eb5

hichamboushaba (approved):
Works well, thanks @AdamGrzybkowski

AdamGrzybkowski:

PR #16370, opened 2026-08-05
[WOOMOB-3759] Initialise Stripe Terminal on the payments screen and fix the simulated reader check

kidinov:
### Description

Fixes WOOMOB-3759

Three fixes around the Tap to Pay device support check, found while testing Remote TTP on real hardware.

**1. The payments screen decided Tap to Pay availability without initialising Stripe Terminal.**

`isTapToPaySupportedOnDevice()` returns `TerminalNotInitialized` while `Terminal.isInitialized()` is false, `TapToPayAvailabilityStatus` mapped that to `null`, and `null == false` quietly read as "available". Nothing here is asynchronous — Terminal simply had not been initialised. It is only initialised from four places (`CardReaderModeViewModel`, `CardReaderConnectViewModel`, `WooPosBuiltInReaderConnector`, `WooPosCardReaderConnectionController`), none of them the payments hub, so on a fresh process the hub built its rows from a guess and dropped them once a card reader flow made the answer real.

`PaymentsHubViewModel` now initialises Terminal in its `init` block, before anything on the screen reads Tap to Pay availability — the first list build and `handleOpenInHubParameter()`, which handles the deep link into the Tap to Pay summary. `Terminal.init` needs no network, connection token or permissions; it only has to run on the main thread, which ViewModel construction already does.

An earlier revision modelled the missing answer as an explicit `Unknown` state instead. That is dropped — it modelled the symptom, and initialising Terminal removes the ambiguity at the source.

**2. `isSimulated` was hardcoded to `false` in `TapToPayDeviceSupportChecker`.**

Stripe rejects any debuggable app for the production reader, so debug builds always got `NotSupported` and the entry points were hidden even with the simulated reader enabled. It now reads `DeveloperOptionsRepository` like every other call site, and the cache is keyed per reader kind since the answer is no longer constant for the process.

**3. POS lost its Tap to Pay UX config once anything else initialised Terminal.**

In `WooPosBuiltInReaderConnector` the `if (cardReaderManager.initialized) return@withContext` guard sat before both `initialize()` and `setupTapToPayUx()`, so POS fell back to Stripe's default colours whenever Terminal was already up. Reachable on trunk today via the connect flow before POS — and fix 1 would have made it the normal case rather than a rare one, which is why these ship together. The guard now covers only `initialize()`, matching `CardReaderConnectViewModel`.

Worth a reviewer's attention: every merchant who opens Payments now starts Stripe Terminal, which registers a lifecycle observer and a terminal listener process-wide.

### Verification

Trunk vs this branch on the same handset and store — OPPO CPH2671, Android 16, NFC, security patch 2026-07-01, simulated reader enabled.

**1.** On trunk, a fresh process shows "Try out Tap To Pay" and "Card Reader Mode"; entering Card reader mode calls `initialize()`, and on return **both rows are gone**. On this branch the screen is identical before and after that same round trip, and the hub asks Stripe itself on the first visit — `supportsReadersOfType deviceType=TAP_TO_PAY_DEVICE result=Supported`.

**2.** The simulated-reader option was on for the whole trunk run and changed nothing, since trunk asks about the production reader regardless. On this branch the toggle flips Stripe's answer, which is the proof the flag is read:

```
on  → result=Supported
off → result="NotSupported(error=… Debuggable applications are not supported when using the
      production version of the Tap to Pay reader. Please use a simulated version of the
      reader by setting TapToPayDiscoveryConfiguration.isSimulated to true.)"
```

with the hub correctly collapsing to just the "About Tap to Pay" row.

**3.** Kill the app → Payments (Terminal now initialised there) → POS → Tap to Pay payment. The payment completed and `setupTapToPayUx` still ran despite Terminal already being up: `class=Terminal message=setTapToPayUxConfiguration`, then `Card reader manager initialized for TTP (initialized=true)`. On trunk that path returns early and neither line can appear.

Not verified on device: the trunk-side repro of defect 3, which needs the connect flow before POS — that sequence was reasoned from the guard, not run. Note the sequence above would *not* reproduce it on trunk, where the hub leaves Terminal uninitialised and POS configures the UX normally.

### Test Steps

1. With **Settings → Developer options → Simulated card reader** enabled, open **Payments**. The Tap to Pay section is shown — before this change it was hidden on every debug build.
2. Turn the simulated reader off and reopen **Payments**. The section collapses to just "About Tap to Pay", and toggling back and forth keeps switching rather than sticking to the first answer.
3. Cold start the app and open **Payments** without touching a card reader flow first. Whatever the Tap to Pay section shows once loading finishes is what it keeps showing — go into **Tap to Pay** or **Card reader mode**, come back, and it must not change.
4. On a device Stripe rejects, the same holds: after loading it settles on the "About Tap to Pay" row and stays there across visits.
5. POS keeps Woo's colours: open **Payments** (or the card reader connect flow) so Terminal is already initialised, then run a POS Tap to Pay payment and confirm the tap screen is Woo-branded rather than Stripe's default.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/taptopay/TapToPayDeviceSupportChecker.kt
```diff
@@ -2,29 +2,33 @@ package com.woocommerce.android.ui.payments.taptopay
...
+ * answers differently for the simulated and the production reader, and the simulated reader can
+ * be toggled from developer options while the process is alive.
  *
  * Returns `null` when Terminal is not yet initialized; callers should treat that as "unknown"
  * and rely on their pre-init heuristics until Stripe can answer.
  */
 @Singleton
 class TapToPayDeviceSupportChecker @Inject constructor(
     private val cardReaderManager: CardReaderManager,
+    private val developerOptionsRepository: DeveloperOptionsRepository,
```
malinajirka: 🔍 np: I'd consider using `AppPrefs` and reading `isSimulatedReaderEnabled` directly, to avoid adding relatively complex repository as dependency here.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt
```diff
@@ -161,14 +164,26 @@ class PaymentsHubViewModel @Inject constructor(
             checkAndUpdateCashOnDeliveryOptionState()
         }
         launch {
-            viewState.value = when (val state = cardReaderChecker.getOnboardingState()) {
+            val state = cardReaderChecker.getOnboardingState()
+            initializeCardReaderManagerIfNeeded()
```
malinajirka: ❓ AFAIU the initialization is needed in order to determine whether to show the rows or not - should we call initialize in `init` instead of in `onViewVisible`?

malinajirka (approved):
Thanks, looks good.

Regarding testing, I'm not sure when the TTP row should be visible - e.g. on my tablet, it's not visible even with simulated reader enabled - is that expected? I mean the tablet doesn't even have NFC, so I guess it's fine, I just wasn't sure if that's how it's supposed to work.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/taptopay/TapToPayDeviceSupportChecker.kt
kidinov: Fixed in e4b6e0a - it reads `AppPrefs.isSimulatedReaderEnabled` directly now.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/hub/PaymentsHubViewModel.kt
kidinov: Good catch, fixed in de2d580 - it moved to `init`. It also covers `handleOpenInHubParameter()`, which runs there and reads Tap to Pay availability, and it keeps the rows shown during the onboarding check from being a guess that changes once it finishes.

kidinov:
@malinajirka 
On the tablet - yes, that is expected. `TapToPayAvailabilityStatus` returns `NfcNotAvailable` before it ever asks Stripe, and for that reason the hub hides the whole Tap to Pay section (not even the "About Tap to Pay" row, which only shows for `DeviceNotSupported`). The simulated reader flag only changes what Stripe is asked about, it does not bypass the NFC check.

PR #16371, opened 2026-08-05
[WOOMOB-3760] Fix the tablet retrying a stale port when connecting to a phone reader

kidinov:
### Description

Fixes WOOMOB-3760 (the stale ephemeral port half — the site-mismatch half of that ticket is out of scope and not handled here)

The phone binds `createServerSocket(0)` and re-advertises with a new port, service name and TLS fingerprint every session. The tablet cancelled NSD/Bluetooth discovery the moment the user tapped connect, so after a failed connect nothing refreshed the cached address and Retry kept hitting the dead port until the reader dialog was closed and reopened — the only workaround a merchant had, and not a discoverable one. Discovery now keeps running through the connection attempt and is cancelled once connected, and Retry re-resolves the phone by `deviceId` — the only identifier stable across a session restart — falling back to the known entry if it hasn't been re-advertised yet.

I picked "keep discovering" over "re-resolve on failure" because the Bluetooth path already leaves discovery running during a connect, so the phone path was the odd one out, and because re-resolution takes seconds of NSD round trips that would be spent after the user taps Retry rather than before it. The cost is that discovery events could now land while the dialog shows a connection outcome, so `handleDiscoveryEvent` ignores them in the `Connecting` / `ConnectingFailed` / `Connected` states — this also stops a late `ReadersFound` from clobbering the Bluetooth success screen, which was already possible.

The stale-entry dedupe in `WooPosUnifiedDiscoveryStream` was keyed on host, which misses a phone that comes back on a new IP; it is keyed on `deviceId` now.

Also here: a refused connect reported `"Connection to phone reader was lost"` when no session ever existed. Connect-time failures now say `"Could not connect to phone reader"`; the lost-connection message stays on `collectPayment`, where a session really did drop. The merchant-facing string was already the generic "Connection failed" in both cases, so this only changes logs and the analytics trail.

One known gap I left alone, pre-existing and not made worse here: a late NSD resolution of an already-lost service can briefly re-cache a dead address. The obvious guard — blacklisting lost service names — breaks recovery from a spurious `Lost`, and a second Retry now recovers where before it never could.

### Verification

Verified on two physical devices — Xiaomi tablet (24018RPACG, Android 16) and OPPO CPH2671 phone (Android 16), same store, same Wi-Fi.

To get a genuinely stale entry I had to drop the phone's Wi-Fi rather than close Card Reader Mode: on this network the system NSD daemon reliably sends the goodbye, even on `am force-stop`, so the tablet's entry is cleaned up before a connect can be attempted. Killing Wi-Fi kills the port with no goodbye, which is the case the original report hit.

```
16:14:24  Connecting TLS to 192.168.31.235:39589      ← phone's Wi-Fi dropped, port dead
16:14:41  Remote reader connect failed: IllegalStateException
          java.lang.IllegalStateException: Could not connect to phone reader
16:17:45  Discovered phone woopos-remote-667e fp=9X-SuVns   ← discovery still running
16:18:24  Retrying phone connection with a re-advertised address
16:18:26  Connecting TLS to 192.168.31.235:37867      ← new port
16:18:32  Remote reader connected: CPH2671
```

Three things to note in that trace. The phone came back with a different fingerprint each session (`ro-DMWB8` → `j3qLrXVc` → `PC940sXA` → `9X-SuVns`), confirming the per-session certificate and why `deviceId` is the only usable key. The 16:17:45 discovery line lands while the dialog is showing "Connection failed" and is *not* followed by a `Found N phones` line — the state guard suppressed the UI update, so the error dialog survived while the cached address was refreshed underneath it. And on the earlier successful connect the NSD listener was unregistered immediately afterwards, which is the new cancel-on-success.

Retry connected on the first tap. Before this change it would have gone back to 39589.

### Test Steps

1. Sign a tablet and an Android phone into the same store. Put the phone into Card Reader Mode (Settings → Payments → Card Reader Mode) on the same Wi-Fi.
2. On the tablet, open POS → connect a reader, and wait until the phone shows in the list.
3. Turn the phone's Wi-Fi off (not Card Reader Mode — a clean exit sends an NSD goodbye and the tablet correctly drops the entry).
4. Tap connect on the tablet's now-stale entry. It fails with "Connection failed".
5. Turn the phone's Wi-Fi back on and start Card Reader Mode again — it binds a new port.
6. Tap **Retry** — it should connect without closing and reopening the reader dialog.
7. Check logcat: the failure reads "Could not connect to phone reader", not "Connection to phone reader was lost".

### Images/gif

N/A


samiuelson (approved):
LGTM!

PR #16373, opened 2026-08-05
[WOOMOB-3761] Add Store design system segment control and modal sheet

hichamboushaba:
### Description

Fixes WOOMOB-3761

Promotes Segment Control and Modal Bottom Sheet from preview-only Store design-system examples into production `Woo*` components, then migrates both Dashboard-owned sheet flows to consume them.

`WooSegmentControl` keeps caller-owned selection and localization, exposes radio/selectable semantics, preserves the 36dp track and 32dp pill, and provides a 48dp interaction target without clipping large text. `WooModalBottomSheet` keeps experimental Material types internal while delegating gestures, scrim, back handling, focus, semantics, system insets, IME behavior, and width to Material. Its Store styling supplies the Surface Bright container, rounded top corners, and drag handle defined by the canonical Figma component.

The Performance card now uses these components for revenue segments and order-date selection. The Analytics updates sheet also moves from its legacy `WCBottomSheetDialogFragment` and navigation destination into the Dashboard's design-system composition, keeping the Dashboard visible behind the scrim while Material owns Back, scrim, and swipe dismissal.

The scheduled-import feature retains its existing copy, repository calls, Learn more behavior, success/failure handling, and Dashboard analytics. Sheet visibility and the last server-confirmed option survive recreation, while optimistic, loading, and error state remain transient. Manual dismissal cancels in-flight UI work; if the server has already applied a change, the Dashboard's existing repository observer converges to that stored value and reseeds the sheet the next time it opens.

Focused component and ViewModel tests, catalog previews, and design-system documentation are included.

> [!NOTE]
> Known dark-theme limitation: The modal sheet currently has limited contrast against the surrounding surface. We will address this in a separate PR once Design provides the intended treatment. p1786013954706019/1785758642.339779-slack-C03L1NF1EA3

### Test Steps

1. Install this branch's Wasabi debug build and open **My Store** with the **Performance** card visible.
2. Select **Total**, **Gross**, and **Net**; confirm the selected pill moves correctly, labels remain readable, and the displayed statistics update.
3. Tap the **Paid orders** selector; confirm the order-date sheet opens at the bottom with a bright surface, rounded top corners, no outlined boundary, and a drag handle.
4. Drag the sheet slightly and let it settle, then dismiss it with Back or the scrim; confirm only the sheet closes.
5. Reopen the order-date sheet and select **Placed orders** or **Completed orders**; confirm it closes after the update succeeds and the Performance card reflects the selected type.
6. Tap the **Last update** row or its Analytics update settings info icon; confirm the **Analytics updates** sheet opens over the still-visible Dashboard and shows **Scheduled**, **Immediately**, the current selection, and **Learn more**.
7. Dismiss the Analytics updates sheet with Back, the scrim, and a downward swipe, reopening it between methods; confirm each action closes only the sheet.
8. Reopen it and tap the already-selected update option; confirm the sheet closes without changing the store-wide setting.

#### Segment Control

| Light | Dark |
| --- | --- |
| Dashboard Segment Control in the light theme | Dashboard Segment Control in the dark theme |

#### Modal Bottom Sheet

| Light | Dark |
| --- | --- |
| Order date modal bottom sheet in the light theme | Order date modal bottom sheet in the dark theme |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooModalBottomSheetState</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/ScheduledImportInfoBottomSheet.kt
```diff
@@ -0,0 +1,261 @@
...
+}
+
+@Composable
+private fun ScheduledImportInfoBottomSheet(
+    state: ScheduledImportInfoViewModel.ViewState,
+    onDismissRequest: () -> Unit,
+    onOptionSelected: (Boolean) -> Unit,
+    onLearnMoreClick: () -> Unit,
+) {
+    if (!state.isVisible) return
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in `01c8737c59d`. Both Dashboard sheets now animate through `WooModalBottomSheetState.hide()` and leave composition only after reaching Hidden; I also verified the programmatic dismissal path on a Pixel 8a running API 36.

AdamGrzybkowski:
Hey @hichamboushaba, the diff here is massive - is there a problem with the branch not being updated, or is that expected?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -330,64 +352,35 @@ private fun StatsChart(
...
 ) {
     val options = DashboardStatsViewModel.RevenueStatsType.OPTIONS
-    val segmentedButtonColors = SegmentedButtonDefaults.colors(
-        activeContainerColor = WooTheme.colors.primary,
-        activeContentColor = WooTheme.colors.onPrimary,
+    WooSegmentControl(
+        options = options.map { stringResource(id = it.labelRes) },
+        selectedIndex = options.indexOf(selectedType),
+        onSelectedIndexChange = { onTypeSelected(options[it]) },
+        modifier = modifier.padding(start = 16.dp, top = 8.dp, end = 16.dp, bottom = 4.dp),
```
AdamGrzybkowski: ❓ Is this something where we could use `WooTheme.padding`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
```diff
@@ -83,14 +81,32 @@ fun DashboardStatsCard(
...
+                    orderDateTypeBottomSheetState.hide()
+                    if (!orderDateTypeBottomSheetState.isVisible) {
+                        showOrderDateTypeBottomSheet = false
+                    }
+                } finally {
+                    isOrderDateTypeBottomSheetDismissing = false
+                }
+            }
+        }
+    }
```
AdamGrzybkowski: ❓ We will need this logic every time we use the bottom sheet (two places already) - would it make sense to extract that to a helper funciton?

AdamGrzybkowski (commented):
Looks good in general, left two comments/suggestions. 

One weird thing that needs to be looked at (or at least confirmed it's real) is this weird slow animation that expand the bottom sheet to full screen height in landscape mode. 

I've tried on two emulators and got the same result, but a physical devices works fine.

https://github.com/user-attachments/assets/aabfbab6-c694-4cde-8130-ab00df4fd606

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
hichamboushaba: I had the same question, and thought about adding a helper, but then later I decided to defer that. This logic is also needed for the M3 bottom sheet (https://developer.android.com/develop/ui/compose/components/bottom-sheets#control-sheet-state), our implementation is just a thin wrapper.

I will still think more about it, and reply later.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
hichamboushaba: Fixed in e27f9735efe

hichamboushaba:
Scrim update: WooTheme.colors.overlay.overlay50 now resolves to Gray 80 (#2C3338) at 50% in both light and dark themes, replacing the previous black 50% light / 75% dark values. This matches the current Figma variable export and the guidance we received. `WooModalBottomSheet` passes that semantic token to Material as `scrimColor`, while Material continues to own scrim rendering and behavior. We kept the existing overlay50 naming and limited the implementation to the token value and component usage. The light/dark treatment is still subject to final confirmation from Design, and we will adjust the token if the designer confirms a different mode-specific value.

| Light | Dark |
| --- | --- |
| Screenshot_1786385118 | Screenshot_1786385175 |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/dashboard/stats/DashboardStatsCard.kt
hichamboushaba: I added a helper in 7fdf0dbd61673d67e4c5b628ce6c0db0d866c11e, please let me know what you think.

hichamboushaba:
Ready for another round @AdamGrzybkowski

AdamGrzybkowski (approved):

PR #16379, opened 2026-08-06
[WOOMOB-3216] Make order filter selectors follow the product filter transition

AdamGrzybkowski:
Fixes WOOMOB-3216

The Orders filter opened its nested selectors as dialogs, so they popped in without the screen transition the Products filter uses and didn't share the app's toolbar — making the two filter menus feel inconsistent (the exact comparison in the ticket).

This converts the Orders filter category and option screens from `DialogFragment`s to full-screen fragments hosted in the main nav graph, so they now use the shared activity app bar and the same push behaviour as the Products filter. The reusable product and customer selectors reached from the filter stay `DialogFragment`s (they're embedded into this graph from a couple of flows and carry their own sub-navigation), but they now slide in and out horizontally instead of popping as modal dialogs.

### Test Steps
1. Orders tab → tap **Filters**. It opens full-screen under the app toolbar (title "Filters", ✕ to close), not as a floating dialog.
2. Tap a filter such as **Order Status** → it pushes as a full screen with a back arrow under the same toolbar. Select options, tap back → the selection is reflected on the parent and **Clear** appears.
3. Tap **Clear** → filters reset; **Clear** should not be shown on the option sub-screen.
4. Tap **Show orders** → returns to the order list with the filter applied.
5. Tap the **Product** filter → the product selector slides in from the right; pick a product → it applies. Repeat for **Customer**.
6. On the filter screen with unsaved changes, press system **back** → the discard-changes dialog appears; "keep changes" cancels, discard returns to the order list.
7. Confirm the bottom **Show orders** button clears the system navigation bar (edge-to-edge, no overlap).

### Images/gif

| Before | After | Products for comparison
|---|---|---|
| <video src="https://github.com/user-attachments/assets/0685adeb-3e18-436d-9394-be8f9e697296" /> | <video src="https://github.com/user-attachments/assets/e3ec62e6-54e4-4e62-b8e4-95bf9c321c42"/> | <video src="https://github.com/user-attachments/assets/c404826d-e913-44ba-8ac0-2184ca08ed66"/> |


WooCommerce/src/main/res/anim/slide_in_from_right.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@integer/activity_transition_duration"
+    android:fromXDelta="100%"
+    android:toXDelta="0" />
```
JorgeMucientes: I think this new animations are duplicates to the ones we already have `default_enter_anim.xml / default_pop_exit_anim.xml`. We can probably remove these 2 files and reference the existing animations.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/customerlist/CustomerListDialogFragment.kt
JorgeMucientes: #### Follow-up idea: drop the two dialog wrappers for `CustomerListDialogFragment` and `ProductSelectorDialogFragment` instead of theming them.

The slide on the product and customer selectors is one-sided:

https://github.com/user-attachments/assets/43a15380-68f9-4b93-9f73-acd94e339120

 The entering screen slides in from the right, but the filter screen underneath doesn't slide out. That's not something the theme can fix, and I think this PR has actually made the proper fix cheap.

**Why it's now possible.** `ProductSelectorDialogFragment` and `CustomerListDialogFragment` are near-verbatim copies of the real `ProductSelectorFragment` and `OrderCustomerListFragment` — same Compose screens, same ViewModels, same result keys. They exist purely as dialog-shaped variants for this graph, which is why both destinations carry a *"Make sure this and X have the same list of arguments"* comment. Since this PR turns the filter screens into regular fragment destinations, the graph can host the real fragments directly.

[order-filter-selectors-to-fragments.patch](https://github.com/user-attachments/files/30898465/order-filter-selectors-to-fragments.patch)

JorgeMucientes (approved):
Nicely done @AdamGrzybkowski 

I'm approving as I didn't see any blocking issues. However, I'm sharing a couple suggestions worth looking into or applying before merging. Lmk your thoughts.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/creation/customerlist/CustomerListDialogFragment.kt
AdamGrzybkowski: Good call, added in https://github.com/woocommerce/woocommerce-android/pull/16379/commits/ed4b77e29e6413785900df8502e4f146bdc443f7

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Thanks for addressing the suggestion @AdamGrzybkowski 🙏🏼 
The code changes look good and work as expected :shipit:

PR #16380, opened 2026-08-06
[WOOMOB-3754] Move the subscription rule into product restrictions and cover the scanner

malinajirka:
### Description

Part of WOOMOB-3754

Order creation doesn't support subscription products, but the rule lived only in `ProductSelectorViewModel.getProductSelection`, so it applied to the product list and nothing else — scanning a subscription product's barcode or SKU added it straight to the order.

The rule now lives in `ProductRestriction.SubscriptionProducts` alongside the other restrictions, and the hierarchy is split into two explicit branches: `Hidden` restrictions remove the product from the list (what every pre-existing restriction does), `Unsupported` ones keep it visible and carry the reason it can't be picked (what subscriptions already did). The scanner reads the same definition and shows each restriction's own message: previously it re-derived the reasons by hand, which gave a private product with no price the "not published" message and silently ignored blocked variable products with no variations. The analytics failure-reason strings changed slightly as part of this.

The other gap — bundles holding a subscription child — is covered in a follow-up PR based on this one: #16381.

### Test Steps
Run tests on the follow up PR.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
```diff
@@ -31,33 +43,81 @@ class ProductFilterProductRestrictions @Inject constructor() : ProductRestrictio
...
+    abstract val scanningMessage: Int
+
+    /**
+     * Failure reason reported to analytics (never shown to the user) when the scanning flow rejects a product this
+     * restriction blocks.
+     */
+    abstract val scanningTrackingReason: String
+
+    /**
+     * A restriction which removes the product from the product list altogether.
```
kidinov, line 52: Do the comments like that give much? Imo, especially with agents, if they drift from the implementation we can make more mistakes as this kinda double source of truth for the AIs

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
```diff
@@ -31,33 +43,81 @@ class ProductFilterProductRestrictions @Inject constructor() : ProductRestrictio
...
+     */
+    sealed class Unsupported(@StringRes val reason: Int) : ProductRestriction() {
+        override val scanningMessage: Int get() = reason
+    }
+
     @Parcelize
-    object NonPublishedProducts : ProductRestriction() {
+    object NonPublishedProducts : Hidden() {
+        override val scanningMessage get() = R.string.order_creation_barcode_scanning_unable_to_add_draft_product
+        override val scanningTrackingReason get() = "Failed to add a product which is not published"
```
kidinov: This string is sent to Tracks as `scanning_failure_reason` in `PRODUCT_SEARCH_VIA_SKU_FAILURE`. Before it was **"Failed to add a product that is not published"**, now it is **"...which is not published"**.

Meaning is the same for us, but for analytics it is just a different value, so old and new data will not group together after the release. The two new reasons (no variations, subscription) are fine, they are new anyway. Can we keep `that` here so the series stays continuous?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
```diff
@@ -31,33 +43,81 @@ class ProductFilterProductRestrictions @Inject constructor() : ProductRestrictio
...
+     * A restriction on products the app can't sell at all. The product stays visible in the product list and
+     * [reason] is shown as why it can't be selected.
+     */
+    sealed class Unsupported(@StringRes val reason: Int) : ProductRestriction() {
+        override val scanningMessage: Int get() = reason
+    }
+
     @Parcelize
-    object NonPublishedProducts : ProductRestriction() {
+    object NonPublishedProducts : Hidden() {
```
kidinov, line 65: `NonPublishedProducts` is not really used anywhere:

- `SelectItems` (`OrderCreateEditViewModel:1203`) and `EditIncludedProducts` (`EditCouponViewModel:214`) pass it, but both navigators drop the `restrictions` list and never send it to the selector.
- `ProductFilterListViewModel:341` checks `restrictions.contains(NonPublishedProducts)`, and `ProductFilterProductRestrictions` never contains it, so that branch is always taken.

Now we also give it `scanningMessage` and `scanningTrackingReason` which can never be shown, and they are an exact copy of the `NonPurchasableProducts` ones. So if it ever did fire, analytics could not tell the two apart.

Since this PR is about unifying restrictions, maybe good moment to just delete `NonPublishedProducts` and the two dead `restrictions` params with it?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
malinajirka: I personally didn't understand the difference between Unsupported and Hidden -> I think explicitly mentioning the difference here makes sense. 

Moreover, in general for AI comments are usually useful - if it matches its understanding of the code it proceeds, if it doesn't it verifies the code again -> updates its understanding or raises inconsistency with the user.

I mean, I haven't verified it myself, but that's what the community says.

kidinov (approved):
LGTM! I left a couple of minor comments  though

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
kidinov: > Moreover, in general for AI comments are usually useful - if it matches its understanding of the code it proceeds, if it doesn't it verifies the code again -> updates its understanding or raises inconsistency with the user.

That sounds like double work for nothing then, no? The same as for people I guess

Funny enough Claude review mentioned to drop those comment

> All four blocks restate what the names and types already say. The only line carrying non-recoverable information is "never shown to the user" on the tracking reason, and that's better solved by naming it analyticsReason.

malinajirka: > That sounds like double work for nothing then, no? The same as for people I guess

I think there might be a misunderstanding, what I mean is:

1. Claude reads code and interprets it
2. Then it reads the comment: if the comment is aligned with understanding of the code, it proceeds (and yes, reading the comment was worth less), however, if the comment is not aligned, claude checks the code again to ensure it understood it correctly -> sometimes it realizes the interpretation in step `1` was wrong (the comment was helpful for the AI, btw same for humans).

I'm not suggesting we should comment all code just because sometimes it could help. However, here it helped me personally when I read it, so I believe this comment is useful.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
malinajirka: Good catch!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
malinajirka: Yep, agree it's worth cleaning up.

malinajirka:
Thanks @kidinov!

I've done the suggested fixes + I ended up creating a separate PR for the dead code cleanup, it turned into a cascade delete with 150 lines of removed code - https://github.com/woocommerce/woocommerce-android/pull/16389.

PR #16383, opened 2026-08-06
[WOOMOB-3777] Treat any bundle with bundled items as configurable

malinajirka:
### Description

Fixes WOOMOB-3777

A bundle whose only merchant choice is a variable child's variation bypassed the configuration screen: `WCProductModel.isConfigurable` guessed from the `bundled_items` metadata (quantity range, optional item, single allowed variation), but that payload carries no product type, so a variable child with fixed quantity looked like "nothing to configure" and the bundle was added to the order with no variation selected.

The heuristic is gone: any bundle with bundled items is now configurable, which is exactly iOS's rule, so this also closes a parity gap. The trade-off is that a bundle with truly fixed contents now shows a configuration step with nothing to choose — same as iOS today — in exchange for never adding a broken order line. The alternative of resolving child types at tap time is written up in the Linear issue; matching iOS was chosen as the simpler and parity-consistent fix.

### Test Steps

1. Create a bundle whose children have fixed quantities and aren't optional, one of them a variable product (e.g. "Bundle product 3" on [the test store](https://site-for-woocommerce12a3fasdf45dfs6789.mystagingwebsite.com/)).
2. In order creation, tap the bundle in the product selector — the configuration screen opens, marks the variable child with "please select a variation", and refuses to save until one is chosen.
3. The bundle's row in the selector now shows the Configure cogwheel.

### Images/gif

N/A


RELEASE-NOTES.txt
```diff
@@ -4,18 +4,16 @@
 
 25.4
 -----
+- [*] Bundles with a choice the app can't detect up front (like a variation of a bundled variable product) no longer skip the configuration screen when added to an order
```
samiuelson: Pls, append the entry with PR URL.

samiuelson (approved):
LGTM :shipit:

PR #16384, opened 2026-08-06
[WOOMOB-2745] Map remote reader failures to user-facing copy

kidinov:
Closes WOOMOB-2745.

Remote Tap to Pay failures currently reach the merchant as raw technical text or as nothing useful at all: the connect dialog renders `"${code}: ${description}"` verbatim, the phone shows `t.toString()`, and the payment path throws the error code away and always says "Something went wrong". This adds a typed `RemoteReaderError` that the phone classifies at the point of failure, and a single mapper that turns it into copy for both the connect dialog and the payment-failed state.

**Based on `issue/WOOMOB-3756-remote-ttp-analytics-gaps`** — review that one first; this will retarget to trunk once it lands.

- Wire codes are stable strings and unknown ones map to `Unknown(raw)`, because phone and tablet are separate installs that can run different app versions.
- `errorDescription` still carries `describeWithCauses()` for Tracks — copy and analytics stay separate.
- Replaced the `"Connection to phone reader was lost"` substring match with a typed `RemoteReaderConnectionLostException`; identifying an error by matching English was going to break on the first translation.
- Last commit fixes raw text I missed on the first pass: the connect dialog could show `missing_merchant_address` and `invalid_postal_code` — literal snake_case identifiers — plus raw exception text from token fetching. Reuses existing copy.

### Known gaps

- **NFC disabled at discovery time is not covered.** `CardReaderDiscoveryEvents.Failed` carries only a message, not the Stripe code, so a `TAP_TO_PAY_NFC_DISABLED` raised during discovery still shows the generic copy. It *is* covered on the collect path, where the raw `TerminalException` survives. Confirmed on device: a discovery-time failure reaches the tablet as `connect_failed` and renders as *"Connection failed"*. Closing the discovery case needs a new event on the shared sealed class.
- **A failed connect still sends `ConnectAck`.** `ConnectionManager.startConnectionToReader` catches `TerminalException` and only updates `readerStatus`, so `handleConnectRequest`'s `runCatching { … }.onSuccess { … }` runs anyway and the tablet reports a connected reader that isn't. Pre-existing and out of scope here, but it means connect-time token errors never reach this mapping. Worth its own issue.

### Test steps

Needs two devices on the same Wi-Fi: a Stripe-Tap-to-Pay-eligible phone and a POS tablet. On a debug build the phone needs **Enable Simulated Card Reader** on — Stripe refuses production Tap to Pay discovery in a debuggable app.

1. Phone: **Payments → Card Reader Mode**, wait for *Ready to pair*. Tablet: **POS → Connect your reader** → pick the phone. Confirm it still connects normally.
2. Clear the store address in WooCommerce settings, then connect again — the dialog should read *"Please provide your store address in order to proceed"*, **not** `missing_merchant_address`.
3. Take a payment and force a failure on the phone — the tablet should show mapped copy, and the phone should show its own message rather than an exception string.
4. With `woo_pos_tap_to_pay` off on both surfaces, confirm the Bluetooth reader connect and payment flows are unchanged.

Verified on a Xiaomi Pad 6S Pro + Oppo Find N5 pair: connect and pairing work end to end. The NFC-off-at-discovery case **cannot be exercised on a debug build**: with the simulated reader off, Stripe fails discovery with *"Debuggable applications are not supported when using the production version of the Tap to Pay reader"* before NFC state is read; with it on, NFC is not used at all. That path needs a non-debuggable build.


samiuelson (approved):
LGTM!

RELEASE-NOTES.txt
```diff
@@ -2,6 +2,10 @@
 *** Use [*****] to indicate smoke tests of all critical flows should be run on the final APK before release (e.g. major library or targetSdk updates).
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 
+25.5
+-----
+- [*] Woo POS: Remote Tap to Pay failures now explain what went wrong - phone not eligible for Tap to Pay, NFC turned off, or a payment service error - instead of a generic message or raw error text
```
samiuelson: Pls, add URL to the PR at the end of the entry.

kidinov: Added in 6850f6a

PR #16389, opened 2026-08-07
[WOOMOB-3754] Delete the unenforced NonPublishedProducts restriction and dead restriction params

malinajirka:
Part of WOOMOB-3754.

### Description
Follow-up to #16380, based on [this review discussion](https://github.com/woocommerce/woocommerce-android/pull/16380#discussion_r3734174095).

The `restrictions` params on `SelectItems`, `EditIncludedProducts`, and `NavigateToProductFilter` have been populated by callers but dropped by every navigator since the 2023 move to DI-provided restrictions, so `NonPublishedProducts` was never actually enforced anywhere. Removing its always-true check in `ProductFilterListViewModel` left `ProductFilterProductRestrictions` with no production usages, so that class is deleted as well. No behavior change.

### Test Steps
1. Order creation → Add products: draft products are still hidden in the product selector.
2. In the product selector, open the filter screen: the Product status filter is still offered.
3. Coupons → edit a coupon → Select products: the selector opens and works as before.

### Images/gif
N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
```diff
@@ -33,14 +33,6 @@ class OrderCreationProductRestrictions @Inject constructor() : ProductRestrictio
         )
 }
```
kidinov: Now only one implementation is left, and both callers inject `OrderCreationProductRestrictions` directly, not the interface. Can we move these 3 methods into the class and drop the interface?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
```diff
@@ -33,14 +33,6 @@ class OrderCreationProductRestrictions @Inject constructor() : ProductRestrictio
...
-class ProductFilterProductRestrictions @Inject constructor() : ProductRestrictions {
-    override val restrictions: List<ProductRestriction>
-        get() = listOf(
-            ProductRestriction.VariableProductsWithNoVariations,
-            ProductRestriction.ProductWithPriceNotSpecified,
-        )
-}
-
 @Parcelize
 sealed class ProductRestriction : (Product) -> Boolean, Parcelable {
```
kidinov: Is `Parcelable` still needed here? After this PR nothing passes restrictions in nav args, so `@Parcelize` and the imports look dead.

kidinov (approved):
LGTM! I left a couple of comments - more stuff can be removed

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
malinajirka: Good point — done in 9f17f84bb81. Also made the `restrictions` list private since nothing read it externally.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductRestrictions.kt
malinajirka: Right, nothing parcels restrictions anymore — removed in 16dcd7e8ade.

PR #16396, opened 2026-08-09
[WOOMOB-3799] Verify cookie authentication endpoints before sending credentials (1/3)

hichamboushaba:
### Description
Fixes WOOMOB-3799

Part 1 of 3. Followed by #16397 and #16398.

> [!NOTE]
> The PR is a bit big, but the bulk is just tests. 

This is one direct-site-credentials fix split into three PRs for review; the parts are not intended to ship independently. Merge order is 1 → 2 → 3, once the whole stack is approved.

Cookie/nonce authentication assumed that every store exposed `/wp-login.php` and `/wp-admin/`. It also posted credentials without first establishing that the destination was a WordPress login form. That prevents login on stores where a security plugin moves either endpoint and makes redirect handling too permissive for a credential-bearing transaction.

This PR adds a validated endpoint model containing the canonical store origin plus optional login-entry and admin-base overrides, then rebuilds nonce acquisition around it.

```mermaid
flowchart LR
    A["GET login entry<br/>up to 3 safe redirects"] --> B{"Verified WordPress<br/>login form?"}
    B -->|No| X["Fail before sending credentials"]
    B -->|Yes| C["POST credentials to the<br/>validated form action"]
    B -->|Already signed in at<br/>the expected admin base| D
    C -->|Redirect must exactly match<br/>the derived nonce URL| D["GET REST nonce<br/>without following redirects"]
```

The important constraints are:

- Every supplied or redirected URL must stay on the canonical host, contain no user information, preserve the approved port, and never downgrade HTTPS. A default-port HTTP origin may upgrade to HTTPS on port 443.
- The login preflight accepts exactly one core-compatible `loginform`, with one enabled username field and one enabled password field. Its form action is resolved against the final preflight URL and validated before use.
- The credential POST remains deliberately small: `log`, `pwd`, and `redirect_to` only.
- The successful credential redirect must equal the nonce URL derived from the validated admin base. A same-origin lookalike is classified as a custom-admin failure but is never followed or persisted.
- WPS Hide Login's already-authenticated `/login` → `/wp-admin/` path is handled without posting credentials to the dashboard; a valid nonce still has to prove the session.
- Nonce cache identity now normalizes terminal slashes while continuing to isolate usernames, subdirectories, schemes, hosts, and ports.
- Nonce failures record whether the login entry had already been verified. PR2 and PR3 use that provenance to distinguish “wrong login address” from a later credential, network, or nonce failure.

The form check intentionally fails closed. A store whose theme or plugin substantially changes WordPress's form identity or field semantics may be rejected rather than receiving credentials. Rejection is not limited to the login form itself: malformed form markup elsewhere on the page, or a comment/script marker that survives stripping, rejects the page with the same non-diagnostic invalid-response result. Supporting interactive 2FA, CAPTCHA, SSO, or arbitrary custom form markup is outside this stack.

This PR also introduces optional authenticated-dashboard verification for a supplied admin base; the user-facing recovery flow activates it in PR3. QR login, XML-RPC, and Application Password architecture are unchanged.

Review map:

- `CookieNonceAuthenticationEndpoints.kt`: endpoint validation and same-origin policy.
- `NonceRestClient.kt`: preflight, form-action selection, redirect rules, dashboard/nonce checks, and cache identity.
- `Nonce.kt`: verified-login-entry provenance.
- `NonceRestClientTest.kt`: the safety and compatibility matrix.

### Test Steps

This PR has no custom-endpoint UI. Use the top-of-stack (PR3) build for step 4.

1. Sign in to a self-hosted store that uses the default `/wp-login.php` and `/wp-admin/` paths. Confirm the store opens normally.
2. Sign out and try again with an incorrect password. Confirm an invalid-credentials error appears and the store does not open.
3. On a store whose HTTP URL redirects to HTTPS on the same host, enter its `http://` address and valid credentials. Confirm login succeeds normally.
4. Using the PR3 build, sign in to a store with WPS Hide Login enabled. Enter its custom sign-in URL when prompted and confirm login succeeds.

### Images/gif

N/A — this PR changes the networking layer; the transaction diagram above is the relevant visual.

The stack's single user-facing release note is owned by PR3.

### DIFF
--- a/libs/fluxc-tests/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClientTest.kt
+++ b/libs/fluxc-tests/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClientTest.kt
@@ -381,6 +381,20 @@ class NonceRestClientTest {
     }
 
     @Test
+    fun `given WPS reaches admin base without trailing slash, when requesting a nonce, then reuse it`() = test {
+        val dashboardUrl = "$SITE_ORIGIN/wp-admin"
+        givenGet(DEFAULT_LOGIN_URL, redirect(dashboardUrl))
+        givenGet(dashboardUrl, WPAPIResponse.Success(ADMIN_DASHBOARD, emptyList()))
+        givenNonce(DEFAULT_NONCE_URL)
+
+        val actual = subject.requestNonce(SITE_ORIGIN, USERNAME, PASSWORD)
+
+        assertIs<Nonce.Available>(actual)
+        verifyNoCredentialPost()
+        verify(requestBuilder).syncGetRequest(subject, DEFAULT_NONCE_URL)
+    }
+
+    @Test
     fun `given preflight reaches an unexpected dashboard URL, when requesting a nonce, then reject it`() = test {
         givenGet(DEFAULT_LOGIN_URL, redirect(CUSTOM_ADMIN_URL))
         givenGet(CUSTOM_ADMIN_URL, WPAPIResponse.Success(ADMIN_DASHBOARD, emptyList()))
@@ -720,7 +734,7 @@ class NonceRestClientTest {
             "<html><body class=\"wp-admin wp-core-ui index-php\">" +
                 "<div id=\"dashboard-widgets-wrap\"></div></body></html>"
         val MANUAL_ADMIN_ENDPOINTS = CookieNonceAuthenticationEndpoints(
-            canonicalSiteOrigin = SITE_ORIGIN,
+            siteUrl = SITE_ORIGIN,
             adminBaseUrl = CUSTOM_ADMIN_URL,
             adminBaseVerification = AdminBaseVerification.AUTHENTICATED_DASHBOARD
         )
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
@claude review

hichamboushaba:
I'm not sure what's wrong with Claude, the issue it complains was in previous commit (https://github.com/woocommerce/woocommerce-android/blob/1c7c9ec30579f6a8dcc3e0dc1f87213a1e28e226/libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/CookieNonceAuthenticationEndpoints.kt#L73-L75), but it's not present anymore.

AdamGrzybkowski (approved):
Great work! All manuall tests pass ✅ 

:shipit:

PR #16398, opened 2026-08-09
[WOOMOB-3799] Recover custom WordPress login and admin URLs (3/3)

hichamboushaba:
### Description

Fixes WOOMOB-3799

Part 3 of 3, stacked on #16397, which is stacked on #16396.

This is the user-facing end of one direct-site-credentials fix split into three PRs for review. Review this diff against PR2's head. Merge order is 1 → 2 → 3, once the whole stack is approved.

Stores whose security plugins move the WordPress sign-in page or dashboard could not complete native credential login because the app had no way to collect and prove the replacement addresses. This PR adds focused recovery forms for both URLs while keeping the canonical store URL as the site's identity.

```mermaid
flowchart TD
    A["Submit store credentials"] --> B["Native cookie/nonce authentication"]
    B -->|Login entry not found| C["Ask for the sign-in URL"]
    C --> B
    B -->|Admin base not found| D["Ask for the dashboard URL"]
    D --> B
    B -->|Wrong password or later<br/>network/nonce failure| E["Show the actual error<br/>and retain the verified URL"]
    E --> B
    C -->|Use WP Admin instead| F["Existing browser authorization<br/>without adopting draft endpoints"]
    D -->|Use WP Admin instead| F
    B -->|Native authentication succeeds| G["Persist and reload proven endpoints"]
    G -->|Save fails| H["Show a save error<br/>and stop login completion"]
    G -->|Saved| I["Eligibility → site selection → logged in"]
```

The recovery behavior is intentionally narrow:

- The first missing login/admin response opens the matching form with the current default. Inputs are normalized and must pass PR1's same-host, scheme, user-information, query, and port rules.
- Before prompting, the app makes one canonical-address reconciliation attempt. A discovered address is adopted only when it is a safe same-origin result; off-host moves, port changes, and HTTPS downgrades are rejected.
- Once a custom sign-in page is structurally verified, that URL is retained through wrong-password retries, later network/response/nonce failures, and process recreation. The merchant sees the real error and can correct credentials without repeatedly entering the URL.
- A manual dashboard URL requires both an authenticated WordPress dashboard response and a valid nonce from its derived admin-ajax endpoint before it can be promoted.
- Cancel discards the current draft. The explicit WP Admin fallback stays available from recovery, but browser authorization is isolated from pending native values and cannot persist them.
- After native authentication succeeds, normalized endpoints are saved and reloaded before eligibility, site selection, or the logged-in event. A persistence failure stops completion instead of leaving the session dependent on transient values.

The existing direct-credentials behavior remains unchanged for stores using the default WordPress paths. QR login, XML-RPC, interactive 2FA/CAPTCHA/SSO handling, and Application Password architecture are outside this PR.

Review map:

- `LoginSiteCredentialsViewModel.kt`: recovery state, error routing, safe reconciliation, promotion, and persistence ordering.
- `LoginSiteCredentialsScreen.kt`: login/admin URL forms and the explicit WP Admin fallback.
- `strings.xml`: merchant-facing recovery and validation copy.
- `LoginSiteCredentialsViewModelTest.kt`: sequential recovery, wrong-password retries, later failures, process recreation, browser isolation, and persistence ordering.
- `RELEASE-NOTES.txt`: the stack's single user-facing note.

### Test Steps

Use direct store credentials; QR login is out of scope.

#### WPS Hide Login — custom sign-in URL

1. Provision a Jurassic Ninja store, install and activate WPS Hide Login, and configure a custom sign-in slug.
2. Start direct store-credentials login with the canonical store URL, the correct username, and an incorrect password.
3. When asked for the sign-in address, confirm that a malformed URL and an off-site URL are rejected inline.
4. Enter the custom sign-in URL. Confirm the invalid-credentials error appears, dismiss it, correct only the password, and continue without entering the sign-in URL again.
5. Confirm the store opens natively without an Application Password tutorial, WebView, Custom Tab, external browser, or retry loop. Force-stop and relaunch the app and confirm the store remains connected.

#### WP Ghost — custom sign-in and dashboard URLs

6. With Docker Desktop running, the reviewed Wasabi debug build installed, and an emulator or device connected, open 3c30a-pb and save the standalone script as `woomob-3799-wp-ghost.sh`.
7. Prepare and start the fixture:

   ```bash
   chmod +x woomob-3799-wp-ghost.sh
   ./woomob-3799-wp-ghost.sh up
   ```

   If more than one device is connected, pass `--serial DEVICE_SERIAL`. Keep the terminal open; it prints the disposable store address, credentials, sign-in URL, and dashboard URL.
8. Open the direct store-credentials screen and use the printed store address and credentials. Enter the printed sign-in URL when prompted.
9. Confirm the dashboard-address form appears after the sign-in URL succeeds. Verify that an off-site value is rejected, then enter the printed dashboard URL.
10. Confirm the store opens natively with no Application Password tutorial or browser. Force-stop and relaunch the app, refresh an authenticated screen, and confirm neither endpoint is requested again.
11. Tear down the fixture:

    ```bash
    ./woomob-3799-wp-ghost.sh down
    ```

### Images/gif
| Login URL | Admin URL |
| --- | --- |
| Screenshot_1786384987 | Screenshot_1786385072  |

This PR carries the stack's single release note under 25.5:
`- [*] Store login now supports custom WordPress sign-in and dashboard addresses [https://github.com/woocommerce/woocommerce-android/pull/16398]`

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsScreen.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsScreen.kt
@@ -22,8 +22,8 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.dimensionResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.input.ImeAction
+import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.text.style.TextAlign
-import androidx.compose.ui.tooling.preview.Preview
 import com.woocommerce.android.R
 import com.woocommerce.android.model.UiString
 import com.woocommerce.android.ui.compose.component.ProgressDialog
@@ -33,6 +33,7 @@ import com.woocommerce.android.ui.compose.component.WCOutlinedTextField
 import com.woocommerce.android.ui.compose.component.WCPasswordField
 import com.woocommerce.android.ui.compose.component.WCTextButton
 import com.woocommerce.android.ui.compose.component.getText
+import com.woocommerce.android.ui.compose.preview.LightDarkThemePreviews
 import com.woocommerce.android.ui.compose.theme.WooThemeWithBackground
 
 @Composable
@@ -42,7 +43,9 @@ fun LoginSiteCredentialsScreen(viewModel: LoginSiteCredentialsViewModel) {
             viewState = it,
             onUsernameChanged = viewModel::onUsernameChanged,
             onPasswordChanged = viewModel::onPasswordChanged,
+            onEndpointUrlChanged = viewModel::onEndpointUrlChanged,
             onContinueClick = viewModel::onContinueClick,
+            onEndpointRecoveryCancelClick = viewModel::onEndpointRecoveryCancelClick,
             onResetPasswordClick = viewModel::onResetPasswordClick,
             onBackClick = viewModel::onBackClick,
             onHelpButtonClick = viewModel::onHelpButtonClick,
@@ -57,7 +60,9 @@ fun LoginSiteCredentialsScreen(
     viewState: LoginSiteCredentialsViewModel.ViewState,
     onUsernameChanged: (String) -> Unit,
     onPasswordChanged: (String) -> Unit,
+    onEndpointUrlChanged: (String) -> Unit,
     onContinueClick: () -> Unit,
+    onEndpointRecoveryCancelClick: () -> Unit,
     onResetPasswordClick: () -> Unit,
     onBackClick: () -> Unit,
     onHelpButtonClick: () -> Unit,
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
```diff
@@ -109,14 +126,15 @@ class LoginSiteCredentialsViewModel @Inject constructor(
         savedStateHandle.getStateFlow(USERNAME_KEY, ""),
         savedStateHandle.getStateFlow(PASSWORD_KEY, ""),
         loadingMessage.map { message -> message.takeIf { it != 0 } },
-        authError
-    ) { siteAddress, username, password, loadingMessage, authError ->
+        combine(authError, endpointRecovery) { error, recovery -> error to recovery }
+    ) { siteAddress, username, password, loadingMessage, errors ->
```
AdamGrzybkowski: ⛏️ I believe Kotlin allows destructuring inline, like:

```suggestion
    ) { siteAddress, username, password, loadingMessage, (error, recovery) ->
```

To be honest, I'm not sure what's better - we could also use the woo combine extension that allows 6 inputs.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
```diff
@@ -144,7 +162,53 @@ class LoginSiteCredentialsViewModel @Inject constructor(
...
+
+    private suspend fun continueEndpointRecovery(recovery: EndpointRecovery) {
+        val endpoints = currentAuthenticationEndpoints(recovery).withSiteSchemeFrom(recovery.url)
+        when (val validation = endpoints.validate()) {
+            is ValidationResult.Valid -> {
+                val validatedEndpoints = endpoints.withValidatedUrls(validation)
+                endpointRecovery.value = recovery.copy(
+                    url = when (recovery.type) {
+                        EndpointType.LOGIN -> requireNotNull(validatedEndpoints.loginEntryUrl)
+                        EndpointType.ADMIN -> requireNotNull(validatedEndpoints.adminBaseUrl)
```
AdamGrzybkowski: ⛏️ Can't we use the URLs from `validation` here to avoid `requireNotNull`? Not a big deal, but having compile-time null safety would be nice.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsScreen.kt
```diff
@@ -173,41 +198,110 @@ fun LoginSiteCredentialsScreen(
...
+    Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.minor_100)))
+    Text(
+        text = stringResource(strings.second),
+        style = MaterialTheme.typography.body2
+    )
+    Spacer(modifier = Modifier.height(dimensionResource(id = R.dimen.major_100)))
+    WCOutlinedTextField(
+        value = recovery.url,
+        onValueChange = onUrlChanged,
+        label = stringResource(strings.third),
```
AdamGrzybkowski, line 233: ⛏️ This field is not usable with a real keyboard (probably a rare scenario) - trying to position the cursor with arrows doesn't work. The same issue happens for username/passowrd inputs, so this is not new and likely doesn't have to be fixed in this PR.

AdamGrzybkowski (approved):
Looks good and worked as expected 👍 

I couldn't use the provided script as it was missing a few extra ones that were required , but the AI was quite helpfull here, recreating the needed setup.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
hichamboushaba: Fixed in 28acc1e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsScreen.kt
hichamboushaba: Good catch. This is pre-existing behavior in the string-backed username and password fields too, so I’d keep the broader text-field selection fix outside this PR. As part of the design-system migration, we might also move these fields to a different component, so it may make more sense to revisit hardware-keyboard handling then.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsViewModel.kt
hichamboushaba: Fixed in 479eadb

hichamboushaba:
> I couldn't use the provided script as it was missing a few extra ones that were required , but the AI was quite helpfull here, recreating the needed setup.

Thanks @AdamGrzybkowski for the review, and sorry about this, I had the wrong script there 🤦, I updated the script now, in case someone wants to re-test the change.

PR #16399, opened 2026-08-10
[WOOMOB-3219] Fix horizontal padding of bulk edit stock status dropdown

AdamGrzybkowski:
### Description
Fixes WOOMOB-3219

Updates the bulk product stock status dropdown to full width.

### Test Steps
1. Open **Products**, long-press a product to enter multi-select, and select 2+ products.
2. Open the overflow menu and tap **Update stock status**.
3. Verify the stock status dropdown spans the screen width with even side margins (aligned with the divider/helper text below it), and that the open popup (In stock / Out of stock / On backorder) matches that width.
4. Regression: edit a Subscription product and open the **Expiration** and **Free trial** pickers — verify their dropdowns still render correctly.

### Images/gif

| Device | Before | After |
| --- | --- | --- |
| Phone | Screenshot_20260810_133945 | Screenshot_20260810_140647 |
| Tablet | Screenshot_20260810_142801 | Screenshot_20260810_142154 |


irfano (approved):
np: This is user facing, can we add a release notes entry? 

Tested, and everything looks good! 👍🏻

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/UpdateProductStockStatusScreen.kt
```diff
@@ -77,7 +77,8 @@ fun UpdateProductStockStatusScreen(
                     onStockStatusChanged(newStatus)
                 },
                 modifier = Modifier
-                    .align(Alignment.CenterHorizontally)
+                    .fillMaxWidth()
```
irfano: I think this line is useless here. It works the same without it.
```suggestion
```

AdamGrzybkowski: Good catch — removed. With `fillWidth = true` the inner text field already fills the width, so the wrap-content `Box` sizes to it and this line was a no-op. Also added a release notes entry for 25.5. Thanks for the review! 🙏

AdamGrzybkowski:

PR #16400, opened 2026-08-10
[WOOMOB-3806] Require manual store selection for multi-site login

JorgeMucientes:
### Description

Fixes WOOMOB-3806

QR login without a supplied site address preselected the first returned store, which could make Continue verify a store the merchant had not chosen. The picker also rendered cached sites before the network refresh, so a cached singleton could trigger auto-login before multiple authoritative sites arrived. Login now leaves cached sites unselected and waits for a successful API refresh before auto-selecting a sole site; supplied-address login is unchanged.

The shared login-epilogue primary button now uses a state-aware tint, so Continue is visibly disabled when no site is selected while preserving its existing enabled color.

### Test Steps

1. Sign in through QR login with a WordPress.com account that has multiple sites and do not enter a site address.
2. Confirm the site picker shows no selected store and `Continue` is disabled. (I've updated the disabled button colors to match the rest of the app).
3. Select a store and confirm Continue becomes enabled and completes the normal verification flow.
4. Repeat with an account that has one site and confirm login still continues automatically.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -270,12 +271,20 @@ class SitePickerViewModel @Inject constructor(
             processLoginSiteAddress(it)
             return
         }
-        if (navArgs.openedFromLogin && wooSites.size == 1) {
-            onSiteSelected(wooSites.first())
-            onContinueButtonClick(isAutoLogin = true)
+        if (navArgs.openedFromLogin && isApiResponse && sites.size == 1) {
```
malinajirka: ❓ Should we check `wooSites.size == 1` instead of `sites.size == 1`?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
```diff
@@ -224,7 +224,7 @@ class SitePickerViewModel @Inject constructor(
         )
     }
 
-    private suspend fun onSitesLoaded(sites: List<SiteModel>) {
+    private suspend fun onSitesLoaded(sites: List<SiteModel>, isApiResponse: Boolean = false) {
```
malinajirka: 🔍 np: AFAIU defaulting to `false` pushes preexisting call sites to cached values, is that intentional? I'd consider an explicit argument without a default value.

malinajirka (approved):
LGTM overall, I've left a couple questions.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
JorgeMucientes: Definitely! Great catch @malinajirka.
I’ve changed both the initial selection and auto-login conditions to use wooSites.size, so a single WooCommerce site is selected automatically even when non-Woo sites are also present.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/sitepicker/SitePickerViewModel.kt
JorgeMucientes: Good point. The default value was intentional. I added it to avoid changing the existing call sites. After reviewing them, they should all pass false because they don’t represent a fresh site list (automatic login should only happen after the complete site list is successfully fetched from the API). And I’ve removed the default value to make it explicit in each function call. Only the full API refresh passes true.

JorgeMucientes:

PR #16405, opened 2026-08-11
[WOOMOB-3811] Fix foreground ANR from the POS promo banner blocking the main thread

kidinov:
### Description

Fixes WOOMOB-3811

`ClientSidePosBanner.shouldShow()` read the store country code with `WooCommerceStore.getStoreCountryCode()`, which wraps a suspend Room DAO call in `runBlocking`. Nothing in the chain from `JitmViewModel.init` switches dispatchers, so that read parked the main thread and Android raised a foreground ANR (Sentry `WOOCOMMERCE-ANDROID-V8N`). Only non-Jetpack sites hit this path — Jetpack sites go through `JitmBannerAdapter`.

`shouldShow()` is now `suspend` and uses the existing `getSiteSettingsAsync()`, which already does the same DAO call on the DB dispatcher. Its only caller was already a suspend function, so nothing else changed signature. I preferred this over wrapping the call in `withContext(Dispatchers.IO)` because that would leave the `runBlocking` accessor on a UI path for the next caller.

The dismissal check now runs before the country lookup, so merchants who dismissed the banner stop paying for a settings read on every My Store open. To keep that reorder from adding a lookup of its own, `isBannerHidden()` takes the already-resolved `SiteModel` instead of resolving it again.

Not addressed here: `SelectedSite.getIfExists()` is still a synchronous WellSql read on the main thread, and `WooCommerceStore` still has two `runBlocking` accessors with no suspend equivalent (`getSitePlugin`, `getActiveSitePlugin`). Neither is this ANR — no `runBlocking`, no Room writer lock — and both are pre-existing across the app. Worth a separate ticket.

Verified on an emulator against a non-Jetpack US store, with an artificial 8s delay injected into the settings read. Before: the read runs on `main` and the main thread stalls for 10.5s. After: the same 8s read, longest main-thread stall 2.9s, all of it unrelated cold-start work.

### Test Steps

Note: a debug build cannot show this banner as-is — `WOO_POS_PHONE` defaults to on in debug, so `wooPosIsScreenSizeAllowed()` returns true and `shouldShow()` bails early. Turn that flag off in Settings → Developer options → Feature flags first.

1. Log in with a **non-Jetpack-connected** store (store credentials / application password) based in the US or GB, on a phone.
2. Open My Store. The POS promotion banner appears as before, and the screen does not freeze.
3. Dismiss the banner. Leave My Store and come back — it stays gone.
4. Repeat with a store in an ineligible country (e.g. CA): no banner, no freeze.
5. Repeat with a Jetpack-connected store to confirm the regular JITM banners are unaffected.

### Images/gif

N/A


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jitm/clientside/ClientSideBannerDismissalStorage.kt
```diff
@@ -3,15 +3,15 @@ package com.woocommerce.android.ui.jitm.clientside
...
 
 @Reusable
 class ClientSideBannerDismissalStorage @Inject constructor(
     private val appPrefsWrapper: AppPrefsWrapper,
     private val selectedSite: SelectedSite,
 ) {
-    fun isBannerHidden(bannerId: String): Boolean {
-        val site = selectedSite.getIfExists() ?: return false
+    fun isBannerHidden(bannerId: String, site: SiteModel): Boolean {
         return appPrefsWrapper.isClientSideBannerHidden(
```
malinajirka, line 13: 🔍 np: `hideBanner` below gets the site internally, we could consider doing the same refactor for consistency.

malinajirka (approved):
Thanks, looks good.

I've noticed one small visual issue - I'll fix it in WOOMOB-3814.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/jitm/clientside/ClientSideBannerDismissalStorage.kt
kidinov: Good point, done in 3f0ad1d. `hideBanner` takes the site now too, so the storage does not need `SelectedSite` at all and site resolution stays in one place.

PR #16407, opened 2026-08-11
[WOOMOB-3814] Restore default padding on JITM banner CTA button

malinajirka:
Fixes WOOMOB-3814

The JITM banner's CTA button used a custom `contentPadding` that only set `start`, but Compose's `PaddingValues` constructor defaults every unspecified side to `0.dp` — so it silently zeroed out top/end/bottom too. That collapsed the button's pressed/ripple `Surface` down to the exact size of the text, so the highlight had no padding on any side. Restoring `ButtonDefaults.TextButtonContentPadding` gives the ripple normal insets again.

### Test Steps
1. Enable JITM testing mode: in `developer.properties`, set `wc.jitm_testing_json_file_name=jitm_testing.json`, then rebuild/install.
2. Open the "My store" dashboard — a JITM banner ("In-person card payments") should appear with a "Buy a card reader" CTA.
3. Press and hold the CTA button.
4. Confirm the pressed/ripple background forms a pill shape with padding around the text on all sides (not flush with the text).

### Images/gif
N/A


kidinov (approved):
I think the change is good, but it requires some layout changes as well now

Image

malinajirka:
🤦 Not sure how come my brain didn't notice this - fixed:
image.

PR #16408, opened 2026-08-12
[WOOMOB-54] Handle empty REST API namespaces during site login

JorgeMucientes:
### Description
Fixes WOOMOB-54

When the WordPress REST API root response contains no namespaces, site-credentials login currently treats it as a valid non-WooCommerce site and can continue into misleading fallback UI. This change treats an absent or empty `namespaces` field as an invalid response so the existing generic site-fetch error is shown instead. Responses with populated namespaces keep their existing behavior.

### Test Steps
1. On a disposable WooCommerce test site, install and activate the [Code Snippets plugin](https://wordpress.org/plugins/code-snippets/).
2. Go to **Snippets > Add New**, name the snippet **Hide REST API namespaces**, and add this code (without a `<?php` opening tag):
   ```php
   add_filter( 'rest_index', function ( WP_REST_Response $response ) {
       $data = $response->get_data();
       $data['namespaces'] = array();
       $response->set_data( $data );
       return $response;
   } );
   ```
3. Select **Run snippet everywhere**, then save and activate the snippet.
4. Open `https://YOUR-SITE/wp-json?_fields=namespaces` and confirm the response contains `"namespaces":[]`.
5. Start the app logged out and log in using site credentials flow (NOT WP.com)
7. Verify the app shows **An error occurred while fetching your website**. It must not open the Application Password authorization/tutorial flow or show the non-WooCommerce error.
8. Deactivate the snippet, repeat the login, and verify the normal site-credentials login flow resumes.

### Images/gif
 


AdamGrzybkowski (approved):
LGTM :shipit:

PR #16411, opened 2026-08-13
Require fastlane ~> 2.238.0 and remove faraday requirement

mokagio:
fastlane 2.238.0 allows using faraday 2.x. The constrain on faraday was there to address a security vulnerability, but now that it resolves to latest 2.x, we no longer need it.

See https://github.com/fastlane/fastlane/pull/30089

The diff is tiny because all the bumps already occurred in https://github.com/woocommerce/woocommerce-android/pull/16409

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,13 +3,9 @@
 source 'https://rubygems.org'
 
 gem 'danger-dangermattic', '~> 1.4'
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
🎉

PR #16414, opened 2026-08-13
[WOOMOB-3845] Add device_type and entry_point to POS analytics events

kidinov:
### Description

Fixes WOOMOB-3845

Adds `device_type` (`phone`/`tablet`) and `entry_point` (`pos_tab`) to every WooPos event, through the `WooPosAnalyticsCommonPropertiesProvider` hook that was already wired into the tracker but returned an empty map. The V2 POS Looker board (WOOMOB-2673) needs to split POS usage by form factor, and to attribute sessions to entry points we add later.

- `entry_point`, not `source` — `source` is already a per-event POS property (`item_added_to_cart`) and a global one would collide.
- `device_type` reads `createDisplayContext(DEFAULT_DISPLAY)`, not the app context, whose `Configuration` follows the current window: a tablet in a small multi-window pane reported `phone`.
- Tablet detection reuses `WooPosScreenSizeUtils.isTabletSize` (674x800dp), the same predicate `WooPosIsScreenSizeAllowed` uses to decide whether POS is reachable, so analytics cannot label a device `tablet` that POS itself serves the phone layout to. iOS uses the hardware idiom, so the two disagree for 600-673dp tablets and for folding devices.
- Cleared in `onPause` guarded by `isFinishing`. `onDestroy` alone was too late — `MainActivity.onResume` runs first, and the tab-visibility event fired there still carried the stale value.
- Known gap: process death while POS is open loses the entry point. Under-counting seemed better than persisting session state for one analytics property.

### Test Steps

1. On a tablet, open POS from the POS tab. Filter logcat for `Tracked: woocommerceandroid_pos`.
2. Confirm events inside POS carry `"device_type":"tablet"` and `"entry_point":"pos_tab"`.
3. Exit POS. Confirm the next `pos_tab_visibility_checked` carries `device_type` but no `entry_point`.
4. Put the app into a small freeform/split-screen window on the same tablet and reopen POS. Confirm `device_type` is still `tablet`.
5. On a phone with `WOO_POS_PHONE` enabled, repeat step 1 and confirm `"device_type":"phone"`.

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.5
 -----
+- [Internal] Woo POS: every POS analytics event now carries device_type (phone/tablet) and entry_point, so POS usage can be split by form factor and attributed to where POS was opened from [https://github.com/woocommerce/woocommerce-android/pull/16414]
 - [*] Woo POS: Remote Tap to Pay failures now explain what went wrong - phone not eligible for Tap to Pay, NFC turned off, or a payment service error - instead of a generic message or raw error text [https://github.com/woocommerce/woocommerce-android/pull/16384]
 - [*] QR login now asks you to choose a store when your WordPress.com account has multiple sites instead of preselecting one [https://github.com/woocommerce/woocommerce-android/pull/16400]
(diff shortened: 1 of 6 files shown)


malinajirka (approved):
LGTM

PR #16418, opened 2026-08-14
[WOOMOB-3851] Request the Android 17 local network permission for POS phone-as-reader

kidinov:
### Description

Fixes WOOMOB-3851

On Android 17, `ACCESS_LOCAL_NETWORK` becomes a runtime permission for apps targeting SDK 37. Nothing requested it, so Remote Tap to Pay broke on the targetSdk 37 bump: the phone's `NsdManager.registerService` threw `SecurityException: Missing local network permission`, and the tablet's TLS socket to the phone timed out. This declares the permission and asks for it on both sides.

The phone treats it as a hard requirement, gated before the session starts, next to the existing location gate. The tablet only asks when the merchant actually picks a phone reader — mDNS browsing is not gated by the permission, so putting it in the pre-discovery requirement chain would have made Bluetooth-only merchants answer a prompt they never need.

Below API 37 `hasLocalNetworkPermission()` returns true, so nothing changes on Android 16 and earlier.

### Test Steps

On an **Android 17 (API 37)** device or emulator:

1. Payments → Card Reader Mode. Confirm the "Local network permission required" screen appears, and granting takes you to **Ready to pair**.
2. Deny twice, and confirm the screen switches to the blocked variant with **Open Settings**. Grant it in Settings, come back, and confirm the session starts without reopening the screen.
3. On a tablet, open POS → Connect your reader → pick a phone reader. Confirm the "Local network permission required" dialog appears and connecting succeeds after granting.

On **Android 16 or earlier**, run the same flows and confirm no new prompt appears and pairing still works.

### Images/gif

**Phone, Android 17 — before vs after**

| Before (no permission requested) | After (permission requested) | After granting |
| --- | --- | --- |
|  |  |  |

**Phone, Android 17 — system prompt and the blocked variant**

`ACCESS_LOCAL_NETWORK` sits in the **Nearby devices** runtime group, so Android renders it with the same wording as the Bluetooth prompt. Whether a system prompt appears depends on whether the app already holds that group: with Bluetooth granted the request is auto-granted silently, and without it the prompt is shown. Reader mode never asks for Bluetooth, so the phone always prompts.

| System prompt | Denied twice |
| --- | --- |
|  |  |

**POS tablet, Android 17 — asked only when a phone reader is picked**

Since #16406 a returning phone-reader merchant skips the Bluetooth gates, so the tablet holds no Nearby devices permission and gets the system prompt too.

| In-app gate | System prompt |
| --- | --- |
|  |  |

**Android 16 regression check — real phone + tablet, no new prompt**

| Tablet finds the phone | Tablet connected | Phone paired |
| --- | --- | --- |
|  |  |  |


WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/cardreader/readermode/CardReaderModeActivity.kt
```diff
@@ -46,19 +59,26 @@ class CardReaderModeActivity : AppCompatActivity() {
...
 
-        if (savedInstanceState == null) {
-            if (WooPermissionUtils.hasFineLocationPermission(this)) {
-                viewModel.onLocationPermissionResult(granted = true, shouldShowRationale = false)
-            } else {
-                viewModel.onLocationPermissionMissing()
-            }
+    override fun onResume() {
+        super.onResume()
+        checkPermissionsAndStartSession()
```
malinajirka: 💡 Coming back from Settings without granting drops the merchant off the blocked screen and back onto the `Continue` explainer — and `Continue` re-requests a permanently denied permission, which Android auto-denies without showing anything, so the screen just flips back to blocked.

`checkPermissionsAndStartSession()` maps every not-granted case to `…PermissionMissing()`, and the denied states are only reachable from the launcher callbacks. Making them sticky (early-return from `onLocationPermissionMissing()` / `onLocalNetworkPermissionMissing()` when the state is already the matching `…Denied`) would keep the Open Settings route. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -559,6 +575,21 @@ class WooPosCardReaderConnectionController(
 
     private fun onPhoneConnectClicked(phone: WooPosDiscoveredReader.Phone) {
         if (_state.value is WooPosCardReaderConnectionState.Connecting) return
+        if (!WooPermissionUtils.hasLocalNetworkPermission(context)) {
```
malinajirka: ❓ A merchant whose last reader was a phone, and who declines this permission, can't get back to the reader list to pick a Bluetooth reader instead. `handleReadersFound` auto-connects to the remembered phone, and the gate returns before any connect attempt, so `removeLastConnectedPhoneDeviceId()` never runs — Cancel closes the dialog and reopening walks straight back into this screen.

Should the gate clear the remembered device id (or skip auto-connect once) so a decline falls through to the list? Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionController.kt
```diff
@@ -188,6 +189,14 @@ class WooPosCardReaderConnectionController(
         checkRequirementsAndStartDiscovery()
     }
 
+    fun onLocalNetworkPermissionResult(granted: Boolean, shouldShowRationale: Boolean) {
+        isLocalNetworkPermissionPermanentlyDenied = !granted && !shouldShowRationale
+        val currentState = _state.value
+        if (granted && currentState is WooPosCardReaderConnectionState.MissingLocalNetworkPermission) {
+            onPhoneConnectClicked(currentState.phone)
```
malinajirka: 💡 This reuses the phone address captured before the permission wall went up, so after an Open Settings round trip the connect can hit a dead port/fingerprint and the merchant has to tap Retry. `onPhoneRetryConnectClicked` refreshes via `latestDiscoveredPhones.refreshAddressOf(phone)` for exactly that reason, and `latestDiscoveredPhones` keeps updating while the wall is shown. Same at line 215.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/cardreader/connection/WooPosCardReaderConnectionViewModel.kt
```diff
@@ -119,6 +123,10 @@ class WooPosCardReaderConnectionViewModel @Inject constructor(
         controller.onLocationPermissionResult(granted, shouldShowRationale)
     }
 
+    fun onLocalNetworkPermissionResult(granted: Boolean, shouldShowRationale: Boolean) {
```
malinajirka: 🔍 np: The Bluetooth and Location equivalents of this delegate and of the new `RequestLocalNetworkPermission` mapping are both pinned in `WooPosCardReaderConnectionViewModelTest` — worth adding the local network pair? A copy-paste slip in the event `when` would compile fine.

samiuelson (approved):
LGTM!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/299">25.5</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

PR #16419, opened 2026-08-14
[Woo POS] Report the refund calculation flow in analytics

samiuelson:
WOOMOB-3750

## Description
The server-computed create refund and the classic create refund flows emit identical analytics, and the route-availability fallback emits nothing at all.

This PR mirrors the iOS change in woocommerce-ios#17716; both platforms report the same properties on the same events, so the funnels stay comparable.

Four changes:

- **`refund_flow` on the refund processing events.** `RefundProcessingStarted`, `RefundProcessingSuccess` and `RefundProcessingFailed` become data classes carrying `refund_flow` (`local` / `server_computed`), with the values in a new `WooPosAnalyticsEventConstant.RefundFlow` enum. The value comes from `isServerComputedRefundConfirmed()`, extracted from `buildSubmissionRequest`. `processRefund` reads it once and passes it to `buildSubmissionRequest`, so the flow reported at start is the flow the request is built on — `analyticsTracker.track()` suspends, and two separate reads around that suspension could disagree. Success and failure derive it from `request.serverLineItems`, the documented discriminator on the request, so they stay correct on the reader-reconnect path where a stored request is submitted later.
- **`refund_server_flow_unavailable`.** Emitted from `WooPosRefundPreview` where the availability cache is marked unavailable. Because the cache short-circuits the resolver on later refunds, this counts stores that fell back rather than refunds. It carries no properties of its own: `AnalyticsTracker` already attaches `cached_woo_core_version` to every event, from the same cached source `flow.wooVersion` reads. Marking the cache and reporting the event are wrapped in `withContext(NonCancellable)`: the caller cancels the preview job whenever the selection changes or the refund is confirmed, and marking without reporting would drop the event for the rest of the process, since the resolver never probes again.
- **`api_error_code` is now omitted when the failure carries no code.** It shipped in #16440 while this branch was open, sending the literal string `"unknown"` in that case. iOS omits the property instead, so a query for "failures with no code" needed different logic per platform — which works against the comparable funnel both PRs exist to build. Android now omits it too, using the same `buildMap` shape `ReaderReadyForCardPayment` already uses in this file for its optional properties.
- **`refund_processing_precondition_failed`.** `refund_processing_started` fires before two local checks in `processRefund` — the order is missing, or the store's currency settings cannot be read. Both then showed a processing error and emitted nothing, so `started` exceeded `success + failed` with no way to see why. The new event closes the funnel, carrying `refund_flow` and an enum-backed `reason` (`order_unavailable` / `currency_settings_unavailable`), following the existing `LocalCatalogSyncSkipped` / `SyncSkipReason` pattern. It reconciles per flow: `started(f) == success(f) + failed(f) + precondition_failed(f)`.

**`RefundFlowStarted` is deliberately left alone.** It fires on the item-selection step, before the preview probe runs, so only *eligibility* is known there — a store that later 404s would report `server_computed` at start and `local` afterwards. Reporting a value that can be contradicted later is worse than reporting none, and the three processing events already cover the comparison we need. Worth revisiting if a start-to-outcome funnel split turns out to be needed.

**`refund_processing_precondition_failed` has no iOS counterpart yet.** The three shared events stay schema-identical, so the cross-platform comparison is unaffected; mirroring the new event on iOS is a follow-up.

## Test Steps
`./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "com.woocommerce.android.ui.woopos.orders.details.refund.*" --tests "com.woocommerce.android.ui.woopos.util.analytics.*"` — 233 tests.

Beyond the flow reported on each processing event for both paths, the added cases cover: a server-computed refund that is *rejected* reporting `server_computed` alongside the error code; a server-computed refund that waits for the reader still reporting `server_computed` after the availability verdict flips, proving the value comes from the stored request and not from re-read state; `api_error_code` absent from the properties when the failure carries no code and present when it does; the precondition event firing when the server verdict flips before submission and the currency settings cannot be read; and the fallback event firing once per store and version rather than on every later refund, and not at all on a store that never probes the route.

To check on device with `woo_pos_server_refunds` enabled: refund from POS on a WooCommerce 11.1.0+ store and confirm `refund_processing_started` carries `refund_flow: server_computed`; repeat on a pre-11.1.0 store and confirm `refund_flow: local`.

There is no manual check for `refund_server_flow_unavailable` or `refund_processing_precondition_failed`, and none is needed. `WooPosResolveRefundFlow` fails closed on version, so a store below 11.1.0 never uses the route; the fallback event needs the app to believe the store is 11.1.0+ while the route is missing. That is not expected in production. One way to produce it would be to downgrade WooCommerce from 11.1.0 to an older version on the backend while the app still has 11.1.0 cached. A plugin or WAF blocking the REST route would do the same. The precondition reasons are narrower still: `order_unavailable` is defensive — the order is loaded before any refund UI exists — and `currency_settings_unavailable` needs a server-computed preview to succeed and the verdict to flip before submission, so the local request cannot be built. Unit tests cover all of them.

## Images/gif
N/A — no UI changes.

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

iamgabrielma:
👋  I won't be able to review the Android PRs until next week, can these be redirected to an Android dev?

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 
 25.5
 -----
+- [Internal] Woo POS refunds: refund events now report whether totals were calculated locally or by the server, the API error code on failures, and when a store falls back for lack of the server route [https://github.com/woocommerce/woocommerce-android/pull/16419]
```
iamgabrielma: This will need update to 25.6!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundViewModel.kt
iamgabrielma, line 1: Codex shares a blocker here. wdyt?

> WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/
    WooPosRefundViewModel.kt:524 RefundProcessingStarted determines the flow before
    analyticsTracker.track(), which suspends, while the actual request independently evaluates
    isServerComputedRefundConfirmed() again at line 576. If the feature flag, cached version,
    selected site, or availability verdict changes during that suspension, the started event can
    report a different flow from the submitted request and outcome events. Build the request first
    and derive all three events through refundFlowFor(request).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
```diff
@@ -1027,16 +1029,52 @@ sealed class WooPosAnalyticsEvent : IAnalyticsEvent {
...
+            val refundFlow: RefundFlow,
+            val apiErrorCode: String? = null,
+        ) : Event() {
             override val name: String = "refund_processing_failed"
+
+            init {
+                addProperties(
+                    buildMap {
+                        put(RefundFlow.REFUND_FLOW, refundFlow.value)
+                        apiErrorCode?.let { put(AnalyticsTracker.KEY_API_ERROR_CODE, it) }
```
iamgabrielma: Suggestion:

> WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/
    WooPosAnalyticsEvent.kt:1070 Importing AnalyticsTracker.KEY_API_ERROR_CODE couples POS analytics
    to the store analytics constants, contrary to the POS typed-constant convention. Define the
    property key within WooPosAnalyticsEventConstant instead.

RELEASE-NOTES.txt
samiuelson: Moved to 25.6. Also trimmed the entry: `api_error_code` shipped in #16440 while this branch was open, so it is no longer part of this PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundViewModel.kt
samiuelson: Good catch, the drift is real — `WooPosAnalyticsTracker.track()` is `suspend` and wraps `withContext(Dispatchers.IO)`, so there is a real suspension point between the two reads.

Fixed, but not by moving the started event. `processRefund` now reads the predicate once and passes it down:

```kotlin
val serverComputed = isServerComputedRefundConfirmed()
analyticsTracker.track(
    WooPosAnalyticsEvent.Event.RefundProcessingStarted(
        refundFlow = if (serverComputed) RefundFlow.SERVER_COMPUTED else RefundFlow.LOCAL
    )
)
val order = currentOrder ?: run { ... }
val request = buildSubmissionRequest(order, contentState, selectedItems, serverComputed) ?: run { ... }
```

Building the request first would also remove the drift, but `refund_processing_started` would stop firing when `currentOrder` is null or the currency decimals cannot be read. Both show the cashier a processing error, and both would then produce no analytics at all. One read, threaded down, keeps the event where it is and makes the two paths impossible to disagree.

Success and failure still use `refundFlowFor(request)`. That reads `request.serverLineItems`, so it reports the request actually submitted — which also stays correct on the reader-reconnect path, where a stored request is submitted later and re-reading ViewModel state would be wrong.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/util/analytics/WooPosAnalyticsEvent.kt
samiuelson: Agreed, and the import is gone. `api_error_code` shipped in #16440 while this branch was open, using an inline `"api_error_code"` key like the other 72 property keys in this file. I kept trunk's version and only added `refund_flow` next to it, so nothing here reaches into the store constants anymore.

iamgabrielma (approved):

PR #16421, opened 2026-08-14
Filter history: shared history screen and ViewModel

AdamGrzybkowski:
Part of [WOOMOB-3326](https://linear.app/a8c/issue/WOOMOB-3326) — WOOMOB-3825

### Description
Second PR in the **filter history** stack. Adds the shared UI, on top of the foundation from #16420. Still behind the `FILTER_HISTORY` flag and not reachable from any menu yet (wired up in the order/product PRs).

- New reusable **Filter History** screen in Jetpack Compose (hosted in a Fragment, added to both filter nav graphs), mirroring iOS `FilterHistoryView`: a "Recent" list, single-select + Apply/Cancel, swipe-to-delete, "Clear History" with a confirmation dialog, and an empty state.
- A generic `FilterHistoryViewModel` driven by `FilterHistoryRepository`, taking a `source` (orders/products) argument and returning the chosen `SavedFilter` as a nav result.
- The 4 filter-history Tracks events (`filter_history_button_tapped`, `_past_filter_applied`, `_past_filter_removed`, `_cleared`) with a `source` property, matching iOS.

### Test Steps

Not reachable in the app yet. You can verify via Compose preview in AS

### Images/gif

Screenshot 2026-08-17 at 10 19 21

- [x] I have considered if this change warrants release notes — none needed (behind feature flag, not user-facing).

PR #16422, opened 2026-08-14
Filter history: order list integration

AdamGrzybkowski:
Part of [WOOMOB-3326](https://linear.app/a8c/issue/WOOMOB-3326) — WOOMOB-3826

### Description
Third PR in the **filter history** stack: wires the shared history screen (#16421) into the **order list** filters. Behind the `FILTER_HISTORY` flag.

- Adds a flag-gated clock entry point to the order filter screen → opens the history screen with `source=orders` and tracks `filter_history_button_tapped`.
- Saves the current selection to history when the user taps **Show Orders** — from either the categories screen or an options sub-screen — via a new `SaveOrderFilterToHistory` use case that reads the full selection from `OrderFiltersRepository` (the source of truth) and runs fire-and-forget on the app scope, so navigation isn't blocked and a rare failure can't crash the app.
- Applies a picked past filter by writing the decoded selection (per category + custom date range) back to the repository and rebuilding the categories, reusing the existing persistence→UI round-trip (so filters whose values no longer exist are silently dropped).
- `OrderFilterHistoryMapper` produces a canonical JSON payload (stable dedup) and a readable label.

### Test Steps
Enable **Filter History** in the developer feature-flag screen, then on the Orders tab:
- Open Filters → tap the clock (top right) → confirm the history is empty.
- Select some filters → Show Orders → reopen the history → the filter appears at the top.
- Apply more filters and confirm newest-first + dedup; swipe to delete one; tap a past filter and confirm the filter screen repopulates and Show Orders lists correctly; Clear History → confirm dialog → empty state.
- Verify Tracks events fire with `source=orders`.

### Images/GIFs

https://github.com/user-attachments/assets/87953b25-8436-4626-8abe-76aae222981f

- [x] I have considered if this change warrants release notes — none needed (behind feature flag, not user-facing).

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
@@ -70,12 +70,18 @@ class OrderFilterCategoriesViewModel @Inject constructor(
 ) : ScopedViewModel(savedState) {
     companion object {
         const val OLD_FILTER_SELECTION_KEY = "old_filter_selection_key"
+        const val OLD_CUSTOM_DATE_RANGE_KEY = "old_custom_date_range_key"
     }
 
     val isFilterHistoryEnabled: Boolean = featureFlagRepository.isEnabled(FeatureFlag.FILTER_HISTORY)
 
     private var oldFilterSelection: List<OrderFilterCategoryUiModel> =
         savedState[OLD_FILTER_SELECTION_KEY] ?: emptyList()
+
+    // The custom date range days aren't captured in [oldFilterSelection] (which only holds option keys),
+    // so snapshot them separately to restore the range when the user discards their changes.
+    private var oldCustomDateRange: Pair<Long, Long> =
+        savedState.get<LongArray>(OLD_CUSTOM_DATE_RANGE_KEY)?.let { it[0] to it[1] } ?: (0L to 0L)
 
     /**
      * Saving more data than necessary into the SavedState has associated risks which were not known at the time this
@@ -99,6 +105,9 @@ class OrderFilterCategoriesViewModel @Inject constructor(
                 _categories = OrderFilterCategories(buildFilterListUiModel())
                 oldFilterSelection = _categories.list
                 savedState[OLD_FILTER_SELECTION_KEY] = oldFilterSelection
+                oldCustomDateRange = orderFilterRepository.getCustomDateRangeDays()
+                savedState[OLD_CUSTOM_DATE_RANGE_KEY] =
+                    longArrayOf(oldCustomDateRange.first, oldCustomDateRange.second)
             }
         }
     }
@@ -202,6 +211,7 @@ class OrderFilterCategoriesViewModel @Inject constructor(
                 ShowDialog.buildDiscardDialogEvent(
                     positiveBtnAction = { _, _ ->
                         saveFiltersSelection(oldFilterSelection)
+                        orderFilterRepository.setCustomDateRange(oldCustomDateRange.first, oldCustomDateRange.second)
                         triggerEvent(Exit)
                     },
                     negativeButtonId = R.string.keep_changes
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
@@ -125,8 +125,25 @@ class OrderFilterCategoriesViewModel @Inject constructor(
     fun onPastFilterSelected(savedFilter: SavedFilter) {
         val historyData = orderFilterHistoryMapper.fromPayload(savedFilter.payload) ?: return
         launch {
+            // Reconcile the stored selection against what this store can currently show, so values that
+            // are no longer available (e.g. the whole Sales Channel category on WooCommerce < 9.9, or a
+            // deleted order status) aren't applied and left stuck active with no visible way to remove them.
+            val availableCategories = buildFilterListUiModel()
             OrderListFilterCategory.entries.forEach { category ->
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/domain/SaveOrderFilterToHistory.kt
```diff
@@ -0,0 +1,164 @@
...
+        if (!hasSelection) return
+        // Store the range in epoch days, matching the unit setCustomDateRange expects on restore
+        // (getCustomDateRangeFilter returns millis, which would corrupt the days-based pref).
+        val customDateRange = orderFiltersRepository.getCustomDateRangeDays()
+        filterHistoryRepository.save(
+            type = FilterHistoryType.ORDERS,
+            payload = orderFilterHistoryMapper.toPayload(
+                categories = selectedCategories,
+                customDateRangeStart = customDateRange.first,
+                customDateRangeEnd = customDateRange.second
```
hichamboushaba, line 87: We should make sure to add the custom date range to the payload only when `DATE_RANGE` is set to `CustomRange`.

`getCustomDateRangeDays()` returns the last range picked by the user even after they switch the date filter back to **All**, and we serialize it unconditionally. This means the payload isn't canonical for the active selection and dedup treats identical filters as different entries.

I reproduced this:

1. Select **Processing** and custom range Jul 20–21.
2. Switch Date Range back to **All**, then tap **Show Orders**.
3. Repeat with Jul 22–23, switch back to **All**, then tap **Show Orders** again.

History shows two **Processing** rows even though both entries apply the same active filter.

Image

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
```diff
@@ -100,9 +110,29 @@ class OrderFilterCategoriesViewModel @Inject constructor(
...
+            mapOf(AnalyticsTracker.KEY_SOURCE to AnalyticsTracker.VALUE_FILTER_HISTORY_SOURCE_ORDERS)
+        )
+        triggerEvent(OpenFilterHistory)
+    }
+
+    fun onPastFilterSelected(savedFilter: SavedFilter) {
+        val historyData = orderFilterHistoryMapper.fromPayload(savedFilter.payload) ?: return
+        launch {
+            OrderListFilterCategory.entries.forEach { category ->
+                orderFilterRepository.setSelectedFilters(category, historyData.selections[category.name] ?: emptyList())
```
hichamboushaba: Can we remove saved filter values that are no longer available before applying a history entry?

The app currently applies every value stored in the history entry, then hides values it cannot show on the filter screen. If a whole category is unavailable—Sales Channel on WooCommerce versions below 9.9, for example—the filter can remain active in the background with no visible way to remove it.

I reproduced it these steps:

1. On a store running WooCommerce 9.9 or newer, select **Sales Channel → Point of Sale** and tap **Show Orders**. This adds the POS filter to history.
2. Make Sales Channel unavailable by using WooCommerce 9.8. In the emulator, I did this by temporarily changing the cached WooCommerce version to 9.8.
3. Open **Orders → Filters → Filter History**.
4. Select the saved **Point of Sale** entry and tap **Apply**.
5. The filter screen now shows only Order Status, Date Range, Customer, and Product. They all say **All**, and there is no Sales Channel row.
6. Tap **Show Orders**.

The order list still shows **Filters • 1** and returns no orders. Reopening the filter screen still shows every visible filter as **All**, so the user has no way to see or remove the active POS filter.

Image

Image

hichamboushaba (commented):
Thanks @AdamGrzybkowski, nice work. I left some comments about some issues that are worth addressing.

And below a case that's not caused by the PR changes, but if you want to address it:

When discarding changes in the filters screen, we don't restore the previous custom date range that was selected before.

You can reproduce this with the following steps:

1. Select **Processing** with the custom range **Jul 22–23**, then tap **Show Orders**.
2. Reopen Filters, change the filter to **Completed** with the custom range **Jul 20–21**, then tap **Show Orders**.
3. Reopen **Filters → Filter History**.
4. Select **Processing, Jul 22–23** and tap **Apply**.
5. On the filter screen, tap the back button without tapping **Show Orders**.
6. When asked whether to discard the changes, tap **Discard**.
7. The app returns to the order list and reloads it.
8. Reopen Filters. The order status is back to **Completed**.
9. Open Date Range and check the custom dates.

The app reloads **Completed orders from Jul 22–23**. The status was restored, but the dates were not. Reopening Date Range confirms that **Jul 22–23** is still active.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/domain/SaveOrderFilterToHistory.kt
AdamGrzybkowski: Fixed in https://github.com/woocommerce/woocommerce-android/pull/16422/changes/bfc7ce41cab687f974de0a1c2180f2fe84a7d59e

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/filters/OrderFilterCategoriesViewModel.kt
AdamGrzybkowski: Fixed in https://github.com/woocommerce/woocommerce-android/pull/16422/changes/03cf60c3b7070b0b48913352536008f425f058c3

AdamGrzybkowski:
>When discarding changes in the filters screen, we don't restore the previous custom date range that was selected before.

Good find! Fixed in https://github.com/woocommerce/woocommerce-android/pull/16422/changes/b909192678cee9bc5c8e359db1576b702987325d

hichamboushaba (approved):
Nice work, thanks @AdamGrzybkowski

PR #16423, opened 2026-08-14
Filter history: product list integration

AdamGrzybkowski:
Part of [WOOMOB-3326](https://linear.app/a8c/issue/WOOMOB-3326) — WOOMOB-3827

### Description
Final PR in the **filter history** stack: wires the shared history screen into the **product list** filters (reusing the clock entry point from #16422). Behind the `FILTER_HISTORY` flag.

- Flag-gated clock entry point on the product filter screen → history with `source=products` and `filter_history_button_tapped`.
- Saves the current selection on **Show Products** via `SaveProductFilterToHistory` (app-scope, fire-and-forget, `runCatching`). Products use a single shared ViewModel, so the save is covered whether Show Products is tapped from the list or an options sub-screen.
- Applies a picked past filter by repopulating the in-memory `productFilterOptions` map + category name, then reloading the filters.
- `ProductFilterHistoryMapper` is a pure JSON codec (matching the order side); label resolution lives in the use case. The category name is only persisted when a category id is actually selected, so clearing to "Any" can't leak a stale name or break dedup.

### Test Steps
Enable **Filter History** in the developer feature-flag screen, then on the Products tab:
- Open Filters → tap the clock → confirm empty history.
- Select filters → Show Products → reopen history → filter appears at top; confirm newest-first + dedup; swipe to delete; apply a past filter and confirm it repopulates and Show Products lists correctly; Clear History → confirm → empty state.
- Verify Tracks events fire with `source=products`.

### Images/GIFs

https://github.com/user-attachments/assets/a9064825-f566-4d20-8567-1bab2bd8c555

- [x] I have considered if this change warrants release notes — none needed (behind feature flag, not user-facing).

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterHistoryMapper.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/filter/ProductFilterHistoryMapper.kt
@@ -1,0 +1,48 @@
+package com.woocommerce.android.ui.products.filter
+
+import com.google.gson.Gson
+import com.google.gson.annotations.SerializedName
+import javax.inject.Inject
+
+/**
+ * Encodes/decodes a [ProductFilterResult] to and from the opaque `payload` string persisted in the
+ * filter history table.
+ *
+ * The payload is a JSON serialization of the selected slugs; Gson omits null fields, so logically
+ * identical selections produce identical JSON and dedup reliably. The category name is only persisted
+ * when a category id is actually selected, so clearing the category to "Any" can't leak a stale name
+ * into the payload (which would break dedup). Decoding tolerates missing/unknown fields (all data-class
+ * fields default to null).
+ */
+class ProductFilterHistoryMapper @Inject constructor(
+    private val gson: Gson
+) {
+    fun toPayload(filter: ProductFilterResult): String = gson.toJson(filter.toData())
+
+    fun fromPayload(payload: String): ProductFilterResult? =
+        runCatching { gson.fromJson(payload, ProductFilterHistoryData::class.java) }.getOrNull()?.toResult()
+
+    private fun ProductFilterResult.toData() = ProductFilterHistoryData(
+        stockStatus = stockStatus,
+        productStatus = productStatus,
+        productType = productType,
+        productCategory = productCategory,
+        productCategoryName = productCategoryName?.takeIf { productCategory != null }
+    )
+
+    private fun ProductFilterHistoryData.toResult() = ProductFilterResult(
+        stockStatus = stockStatus,
+        productType = productType,
+        productStatus = productStatus,
+        productCategory = productCategory,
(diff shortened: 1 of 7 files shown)


hichamboushaba (commented):
Thanks @AdamGrzybkowski for the PR.

Can we make sure a category restored from history is still available and displayed correctly on the filter screen?

I found two issues:

**1. Applying a category from history can leave the filter screen out of sync: the Category row shows **Any**, while the restored category remains active and filters the product list.**

To reproduce:

    1. Save a filter with Category = Decor.
    2. Change the selected category to Clothing.
    3. Apply the saved Decor filter from history.
    4. The Category row shows Any, but Filters (1) is displayed and Show products returns only Decor products.

The Category row should show **Decor** after applying the history entry.

Image

Image

</br></br>

**2. When a saved category is no longer available, it remains active and can return no products instead of being dropped.**

Image

Image

AdamGrzybkowski:
@hichamboushaba thanks for catching that! My mistake 🤦 

Issue 1 was fixed. 

Issue 2 is also fixed, but only by looking at the persisted category list. Adding a network request to fetch the list would add a noticeable delay to the filter selection, and it wouldn't work well. 

An alternative I considered: Apply based on the persisted category list and run a background fetch to get the fresh data. This would work, but in those rare cases where a category was removed, it would lead to a silent filter change that could be missed by the user. I didn't like that, so I decided not to go this route.

hichamboushaba:
Thanks @AdamGrzybkowski. Issue 2 looks fixed.

Issue 1 still isn’t working as expected. I reproduced it with these steps:

1. Open product filters and select Category → Decor.
2. Tap Show products. This also saves the Decor filter to history.
3. Open product filters again and change the category to Clothing.
4. Tap Show products, so Clothing becomes the active filter.
5. Open product filters again, tap Filter History, select Decor, then tap Apply.

Expected: the filter screen shows Category: Decor and Filters (1).

Actual: the filter screen shows Category: Any and no active filters. Decor is dropped instead of being restored.

It looks like `productCategories` contains only Clothing when the filter screen is reopened, so `applyRestoredCategory()` doesn’t load the persisted category list because the in-memory list isn’t empty. Could we also load it when the list is only partially populated? This should remain a local lookup without requiring a network request.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/301">25.6</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski:
Thanks @hichamboushaba  - this should now be fixed.

hichamboushaba (approved):
Looks good, thanks @AdamGrzybkowski

PR #11347, opened 2024-04-18
[Watch App] Introduce Wear module with build and theme configurations.

ThomazFB:
Summary
==========
Fix issue #11333 and #11334 by introducing the `WooCommerce-Wear` module with the following:

### Build configs
- Package and APK signing matching the Phone (required to allow Phone-to-Watch connection)
- Build Flavors / Variants (Vanilla and Wasabi with the same package pattern)
- ProGuard and Lint configurations
- Android Manifest configuration

###  Libraries and Dependencies
- Hilt and Dagger
- Compose
- WearOS libraries
- DataStore libraries

### Architecture support
- Dependency Injection modules configured
- Data Store declaration and injection configured
- Coroutines scope and injection configured
- Navigation through Compose

### Wear support
- Phone-to-watch Service declaration and configuration
- Watch-to-phone Service declaration and configuration
- Default complication setup
- Default tile setup

### Theme, Icons and Strings
- Round and Square resources separation (including app icon)
- Vanilla and Wasabi resources separation (including app icon)
- Colors setup
- Theme setup
- Typography setup

Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16427, opened 2026-08-17
Configure Sentry from public values and an env-injected token

mokagio:
In preparation for moving to [`a8c-secrets`](https://github.com/Automattic/a8c-secrets), reduces the surface area of secrets we'll need to migrate. The Sentry project does not need to be hidden from the repo, the only value that should not be tracked is the API token, which is not injected in the environment and provisioned in CI. Additionally, the DSN for the mobile and wear project remain secret, to avoid open source contributors accidentally sending events to our Sentry instances.

 See [AINFRA-2717](https://linear.app/a8c/issue/AINFRA-2717).

Verified via #16428 with a [throwaway step](https://buildkite.com/automattic/woocommerce-android/builds/41603/canvas?sid=01a00eca-806f-407d-802b-d2a65cdef0a7&tab=output) that exercises the Sentry code:

image

<details>
<summary>AI-generated details</summary>

### Description

`configure_apply` decrypted `sentry.properties` into the repo root, but only the auth token in it is secret — `defaults.org` and `defaults.project` are public values. Per the paaHJt-ahO-p2 RFC, the public halves move into source and the token comes from `SENTRY_AUTH_TOKEN`.

Uploads are gated on CI plus a `-PskipSentryUpload` escape hatch, so the `sentry-cli` tasks exist only when an upload is actually wanted. The missing-token guard then fires exactly then — a release we can't deobfuscate fails loudly instead of shipping. Local release builds don't need the token at all, since they no longer upload.

The split also fixes a live bug. The Wear app reports events to the `woocommerce-wear` Sentry project through its own DSN, but both modules read the single root `sentry.properties` — so its **source bundle** was uploading to `woocommerce-android`, which is why Wear stack traces carried no source context. (Wear sets `minifyEnabled false`, so there are no ProGuard mappings involved; an earlier commit message in this branch says otherwise and is corrected by a later one.) Each module now names its own project. That part is isolated in its own commit if you'd rather it went separately.

Note that a blank `SENTRY_AUTH_TOKEN` is treated as missing: a secret-injection step that resolves to nothing exports an empty string rather than leaving the variable unset.

### Test Steps

No PR or trunk build exercises the upload, so a green CI run does **not** prove the token wiring.

1. Confirm the gate, with no `sentry.properties` on disk:
   `./gradlew :WooCommerce:assembleVanillaRelease --dry-run` → no `sentry*` upload tasks
   `CI=true ./gradlew :WooCommerce:assembleVanillaRelease --dry-run` → they appear
2. Confirm the guard fires when an upload is due but the token is absent:
   `CI=true ./gradlew :WooCommerce:uploadSentryProguardMappingsVanillaRelease` → fails with `SENTRY_AUTH_TOKEN is not set`
3. Exercise the real upload and confirm each lands in its own project:
   `CI=true SENTRY_AUTH_TOKEN=… ./gradlew :WooCommerce:uploadSentryProguardMappingsVanillaRelease` → `woocommerce-android`
   `CI=true SENTRY_AUTH_TOKEN=… ./gradlew :WooCommerce-Wear:sentryUploadSourceBundleWasabiRelease` → `woocommerce-wear`

### Images/gif

N/A.

</details>


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce-Wear/build.gradle
```diff
@@ -10,7 +11,16 @@ plugins {
     alias(libs.plugins.dependency.analysis)
 }
 
+// A secret-injection step that resolves to nothing exports an empty SENTRY_AUTH_TOKEN rather than
+// leaving it unset, so treat blank as missing — otherwise the guard below passes and sentry-cli
+// fails mid-upload with an opaque auth error.
```
mokagio:
```suggestion
```

WooCommerce/build.gradle
```diff
@@ -63,7 +64,16 @@ fladle {
     ]
 }
 
+// A secret-injection step that resolves to nothing exports an empty SENTRY_AUTH_TOKEN rather than
+// leaving it unset, so treat blank as missing — otherwise the guard below passes and sentry-cli
+// fails mid-upload with an opaque auth error.
```
mokagio:
```suggestion
```

WooCommerce-Wear/build.gradle
```diff
@@ -31,6 +41,17 @@ sentry {
     ignoredBuildTypes = ["debug"]
 }
 
+// These tasks only exist when the upload flags are on, so this fires exactly when an upload is due.
```
mokagio:
```suggestion
```

WooCommerce/build.gradle
```diff
@@ -91,6 +101,17 @@ sentry {
     ]
 }
 
+// These tasks only exist when the upload flags are on, so this fires exactly when an upload is due.
```
mokagio:
```suggestion
```

RELEASE-NOTES.txt
```diff
@@ -5,6 +5,8 @@
 25.5
 -----
 - [WEAR][*] Order and stats dates on the watch now follow the language's date order [https://github.com/woocommerce/woocommerce-android/pull/16429]
+- [Internal] [WEAR] Wear crash reports are now deobfuscated: ProGuard mappings upload to the `woocommerce-wear` Sentry project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```
mokagio: @woocommerce/android-developers this is something the agent I worked with on this PR noticed and should be a nice side effect. Let us know how it goes.

RELEASE-NOTES.txt
```diff
@@ -5,6 +5,7 @@
 25.5
 -----
 - [WEAR][*] Order and stats dates on the watch now follow the language's date order [https://github.com/woocommerce/woocommerce-android/pull/16429]
+- [Internal] [WEAR] Wear stack traces in Sentry now carry source context: the source bundle uploads to the `woocommerce-wear` project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```
AliSoftware, line 8: Nice side effect 👍

WooCommerce-Wear/build.gradle
```diff
@@ -10,8 +11,17 @@ plugins {
     alias(libs.plugins.dependency.analysis)
 }
 
+def sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull?.trim() ?: null
+def shouldUploadToSentry = gradle.ext.isCi
+        && !providers.gradleProperty("skipSentryUpload").orNull?.toBoolean()
+
```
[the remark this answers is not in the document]
mokagio: Good catch — the code is deliberate here and the docs were stale. The property was renamed to `skipSentryUpload` because it also gates `includeSourceContext`, and this module has no ProGuard mappings at all, so the old ProGuard-specific name was misleading. I've updated the PR description to match.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

WooCommerce/build.gradle
```diff
@@ -63,8 +64,17 @@ fladle {
     ]
 }
 
+def sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull?.trim() ?: null
+def shouldUploadToSentry = gradle.ext.isCi
+        && !providers.gradleProperty("skipSentryUpload").orNull?.toBoolean()
+
```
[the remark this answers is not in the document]
mokagio: Same as the Wear module: `skipSentryUpload` is the intended name (the flag gates `includeSourceContext` too, not just mappings). PR description updated to match the code.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

WooCommerce-Wear/build.gradle
```diff
@@ -10,8 +11,17 @@ plugins {
...
 sentry {
-    includeSourceContext = true
+    org = "a8c"
+    projectName = "woocommerce-wear"
+    authToken = sentryAuthToken
+
+    includeProguardMapping = shouldUploadToSentry
```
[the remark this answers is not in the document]
AliSoftware: That makes me wonder if having `minifyEnabled = false` on the wear module on release is intentional here?

Is it because this is something not supported by the wearOS platform? Is it an unintended misconfiguration?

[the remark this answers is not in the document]
mokagio: <details>
<summary>Original AI comment</summary>
The premise is right — `release` does set `minifyEnabled false` — but the consequence doesn't follow: the Sentry Gradle plugin never registers mapping tasks for a module without minification, so there are no unnecessary or failing tasks to avoid.

Verified on this branch:

```
$ ./gradlew :WooCommerce-Wear:tasks --all | grep -i sentry
sentryBundleSources{Jalapeno,Vanilla,Wasabi}Release
sentryCollectSources{...}Release
sentryUploadSourceBundle{...}Release
generateSentryBundleId{...}Release
injectSentryDebugMetaPropertiesIntoAssets{...}Release

$ ./gradlew :WooCommerce-Wear:uploadSentryProguardMappingsWasabiRelease --dry-run
Cannot locate tasks that match ':WooCommerce-Wear:uploadSentryProguardMappingsWasabiRelease'
as task 'uploadSentryProguardMappingsWasabiRelease' not found in project ':WooCommerce-Wear'.
```

Only source-bundle tasks exist. `includeProguardMapping` is inert here, and leaving it symmetric with the phone module means it does the right thing automatically if Wear ever enables minification. Skipping.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

</details>

@AliSoftware apologies for the AI comment. I asked AI to address the comment and clearly the global rules I have to avoid this behavior—replying directly to people—are not good enough.

Before seeing this comment, I tasked an agent to look through the Git history for more info.

The setting came in via https://github.com/woocommerce/woocommerce-android/pull/11347, the original Wear setup PR. cc @ThomazFB 

For this PR, that means `includeProguardMapping = shouldUploadToSentry` in the Wear module is currently a no-op: with minification off there is no mapping file for Sentry to upload.

I've tasked an agent to open a stacked PR to change the configuration in isolation, so that @woocommerce/android-developers can chip in on their preferred configuration.

RELEASE-NOTES.txt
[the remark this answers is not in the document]
mokagio: Correct, and fixed in 08d535aa7ea — that note was left over from an earlier commit message in this branch that misdescribed the fix. There are no Wear mappings, so nothing about this PR changes Wear deobfuscation.

The remaining note covers what actually changed: the source bundle now uploads to `woocommerce-wear` instead of the phone app's project.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

AliSoftware (commented):
Looking good at first glance, but I think it's worth digging about the proguard not running on wear release?

mokagio:
Thanks @AliSoftware 

> Looking good at first glance, but I think it's worth digging about the proguard not running on wear release?

Addressed via https://github.com/woocommerce/woocommerce-android/pull/16443

mokagio:
@ParaskP7 @hichamboushaba can I ask you look over this PR as well? In https://github.com/woocommerce/woocommerce-android/pull/16443, you guided me against modifying the Wear setup. How does the rationale from that PR apply here? The end goal of this PR—using public Sentry values + env-injected token—is required, but the implementation can be changed if there are concerns similar to the other PR. 

Thanks!

ParaskP7:
> @ParaskP7 @hichamboushaba can I ask you look over this PR as well? In https://github.com/woocommerce/woocommerce-android/pull/16443, you guided me against modifying the Wear setup. How does the rationale from that PR apply here? The end goal of this PR—using public Sentry values + env-injected token—is required, but the implementation can be changed if there are concerns similar to the other PR.

Thanks for the ping @mokagio , and, looking at the changes at a glance, as per Hicham's comment [here](https://github.com/woocommerce/woocommerce-android/pull/16443#discussion_r3955947212), and unless totally necessary, for the project overall, and then for this `wear` app module, I wouldn't change anything on the `WooCommerce-Wear`'s `build.gradle` configuration, Sentry or not. Having said that, I am not sure why Sentry was/is configured in that `wear` app module the way it is, so leaving that for @hichamboushaba to answer and have a final say on whether this change should survive this PR. 🙏

hichamboushaba:
Sorry for the late reply @mokagio, I took a quick glance at the changes, and I think these are good to go, this is mostly infra code, it shouldn't affect runtime (right?), and with this in mind, keeping the setup similar between the main app and the wear app would be beneficial.

mokagio:
Between @hichamboushaba 's [comment](https://github.com/woocommerce/woocommerce-android/pull/16427#issuecomment-5665145793) and @ParaskP7 's 👍 reaction, I consider this ready to merge even without an explicit approval.

image

I triggered CI and will admin-merge once green.

hichamboushaba:
@mokagio before merging, check this point that was raised by AI, and I think it might be valid:

> Removing `sentry.properties` from `.configure` stops creating new copies, but it does not remove the plaintext file from existing configured checkouts. Those files contain `auth.token`; removing the `.gitignore` entry makes them immediately eligible for an accidental `git add .`, while removing `.aiexclude` can expose them to AI tooling. Please retain both ignore entries even though the build no longer consumes this file.

mokagio:
Thanks @hichamboushaba . I'm of two minds about the approach. One the one end, I appreciate the added safety. On the other, I like the clean cut move.

I implemented your suggestion because you raising it adds extra weight to the "additional safety" angle. I also tracked the need to follow up and remove those at some point in https://linear.app/a8c/issue/AINFRA-3221 .

RELEASE-NOTES.txt
```diff
@@ -80,6 +80,7 @@
 - [*] You can now apply a gift card when editing an order [https://github.com/woocommerce/woocommerce-android/pull/16445]
 - [Internal] Woo POS analytics: replaced the catch-all "other" reason on the POS ineligible screen with specific reasons, and added the missing orders_list_loaded event [https://github.com/woocommerce/woocommerce-android/pull/16439]
 - [WEAR][*] Order and stats dates on the watch now follow the language's date order [https://github.com/woocommerce/woocommerce-android/pull/16429]
+- [Internal] [WEAR] Wear stack traces in Sentry now carry source context: the source bundle uploads to the `woocommerce-wear` project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```
hichamboushaba, line 83: Let's move this to the correct version before merging.

hichamboushaba (approved):
Thanks @mokagio, looks good.
I left one comment about the release note being in the wrong version.

RELEASE-NOTES.txt
mokagio:
```suggestion
- [Internal] [WEAR] Wear stack traces in Sentry now carry source context: the source bundle uploads to the `woocommerce-wear` project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```

RELEASE-NOTES.txt
mokagio: Ah, good point!

mokagio:
```suggestion
```

RELEASE-NOTES.txt
```diff
@@ -8,7 +8,7 @@
 - [Internal] Push notification analytics now attribute store and plan properties to the notification's origin site instead of the currently selected store [https://github.com/woocommerce/woocommerce-android/pull/16575]
 - [*****] Refreshed My Store, Orders, Products, and the More menu with the new design system — new top app bars, bottom navigation, list styling, and icons [https://github.com/woocommerce/woocommerce-android/pull/16322]
 - [*] Fixed the add product button staying on screen after clearing the product search term [https://github.com/woocommerce/woocommerce-android/pull/16451]
-- [Internal] Card Reader Mode on the phone now tracks when it rejects a Remote Tap to Pay payment from the tablet [https://github.com/woocommerce/woocommerce-android/pull/16584]
+- [Internal] [WEAR] Wear stack traces in Sentry now carry source context: the source bundle uploads to the `woocommerce-wear` project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```
mokagio:
```suggestion
- [Internal] Card Reader Mode on the phone now tracks when it rejects a Remote Tap to Pay payment from the tablet [https://github.com/woocommerce/woocommerce-android/pull/16584]
- [Internal] [WEAR] Wear stack traces in Sentry now carry source context: the source bundle uploads to the `woocommerce-wear` project rather than the phone app's [https://github.com/woocommerce/woocommerce-android/pull/16427]
```

PR #16435, opened 2026-08-17
[WOOMOB-990] Load refunds asynchronously in IssueRefundViewModel

irfano:
Fixes WOOMOB-990

### Description

`IssueRefundViewModel.init` loaded the order refunds with `runBlocking`, blocking the main thread on a DB read every time the refund screen opened.

The refunds now load asynchronously next to the order, and the screen sections are built once both are in. Part of WOOMOB-983.

### Test Steps

**An order with a custom amount, not refunded yet**
1. Open the **Orders** tab.
2. Open the order.
3. Tap **Issue refund**.
4. Confirm the **Refund custom amount** toggle is offered.

**An order with products and a single shipping line, nothing refunded yet**
1. Open the order.
2. Tap **Issue refund**.
3. Confirm every product is listed with its full quantity available.
4. Confirm the shipping toggle is offered.

**An order with previously refunded products**
1. Open the order.
2. Tap **Issue refund**.
3. Confirm only the quantity that is left is offered on the partly refunded product.
4. Confirm a fully refunded product is not listed.

**An order whose custom amount was refunded before**
1. Open the order.
2. Tap **Issue refund**.
3. Confirm the custom amount toggle is not offered.

**An order whose shipping line was refunded before**
1. Open the order.
2. Tap **Issue refund**.
3. Confirm the shipping toggle is not offered.

**An order with more than one shipping line**
1. Open the order.
2. Tap **Issue refund**.
3. Confirm the multiple shipping lines notice is shown.

### Images/gif

N/A

- [x] I have considered if this change warrants release notes. This is an internal threading fix with no user-visible change, so none were added.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/IssueRefundViewModel.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/payments/refunds/IssueRefundViewModel.kt
@@ -147,7 +147,7 @@ class IssueRefundViewModel @Inject constructor(
 
     private val order: Order
         get() = requireNotNull(orderFlow.replayCache.firstOrNull()) {
-            "Please ensure that this property is not accessed before the order is loaded."
+            "Order is not loaded yet"
         }
 
     private val Order.allFeeLineIds: List<Long>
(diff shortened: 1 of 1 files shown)


AdamGrzybkowski (approved):
Looks good! :shipit:

PR #16438, opened 2026-08-18
[Woo POS] FluxC Create Refund code clean up 

samiuelson:
WOOMOB-3787

### Description
The computed create exists so the backend owns the refund total: the client sends which items to refund and the server computes every monetary value.

**1. Implicit defaults on `createComputedItemsRefund`.** `reason`, `autoRefund`, `restockItems` and `amount` each change what the store does with the merchant's money or stock, and every one of them had a default. A caller could omit any of them and still get a live gateway refund with a restock. The defaults are gone, so every call site states what it wants.

**2. The `compute_totals` availability docs were wrong.** The KDoc on `WCRefundStore.createComputedItemsRefund`, `RefundRestClient.createComputedRefund` and `WooPosResolveRefundFlow` said the `/wc/v3` refund preview route and the `compute_totals` create shipped as separate WooCommerce changes, so a successful preview proved only that the preview route exists. They shipped in one merge to WooCommerce trunk (woocommerce/woocommerce#67042, which carried #67043) and one backport to `release/11.1` (woocommerce/woocommerce#67613, both changelogs, one controller file). woocommerce/woocommerce#67043 also names the preview probe as the feature detection for `compute_totals`. The docs now say that.

The `MIN_WC_VERSION_FOR_SERVER_REFUNDS = "11.1.0"` gate stays. It is cheap, it does not depend on core keeping the two together in future backports, and it is the only gate that can answer before a preview has ever run. Its KDoc now says that instead of claiming the preview is insufficient.

**What this PR no longer does.** An earlier revision also removed the `amount` parameter from the computed create. That is reverted. `amount` pins the total the cashier confirmed, so the server rejects the create when its recomputed total drifts above it. Nothing else checks that — `handleBackendRefundResult` only reads `result.isError`, never `result.model.amount` — and a post-create comparison cannot replace the pin, because `api_refund` is `true` when the gateway supports refunds, so the money has already moved by the time the response arrives. Removing the parameter is deferred until the minimum WooCommerce version reaches 11.1.0.

### Test Steps
```
./gradlew :libs:fluxc-plugin:testDebugUnitTest --tests "*RefundStoreTest*" --tests "*RefundRestClientTest*"
./gradlew :WooCommerce:testWasabiDebugUnitTest --tests "*WooPosRefundSubmissionProcessorTest*" --tests "*WooPosResolveRefundFlowTest*"
```

No behavior changes. The diff is the four removed defaults, KDoc, and the test call sites that had relied on the defaults.

### Images/gif
N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/299">25.5</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
```diff
@@ -230,17 +230,14 @@ class WooPosRefundSubmissionProcessor @Inject constructor(
...
-            // Pinning the confirmed total makes the server reject the create when its recomputed
-            // total drifted above it (order edited between preview and create) instead of
-            // refunding more than the cashier confirmed. Omitted on backend-only retries: the
-            // card has already been reversed, and a rejection would strand that money behind a
-            // permanently failing retry (WOOMOB-3752 covers reworking that path).
-            val amountOverride = request.refundAmount.takeUnless { retryBackendNotificationOnly }
+            // No total is sent: the server computes it from the line items. That is what the
+            // computed create is for, and it keeps the amount identical on a backend-only retry,
+            // where the card has already been reversed and a rejected create would strand that
+            // money behind a permanently failing retry (WOOMOB-3752 covers reworking that path).
```
kidinov: I think this is not a no-op. The pin was added in af54d61 so the server rejects the create when its total drifts above what the cashier confirmed (order changed between preview and create). Now nothing checks it, `handleBackendRefundResult` does not look at `result.model.amount`. Maybe compare it with `request.refundAmount` instead of removing the check?

kidinov (commented):
@samiuelson overall looks good but a dew things to take a look!

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
```diff
@@ -77,17 +77,15 @@ class RefundRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
...
     ): WooPayload<RefundResponse> {
         val body = mapOf(
             "compute_totals" to true,
             "reason" to reason,
-            "amount" to amount,
             "api_refund" to apiRefund.toString(),
             "api_restock" to apiRestock.toString(),
             "line_items" to lineItems,
-        ).filterNotNull()
+        )
```
kidinov: Now the body has no money value at all. If a store drops `compute_totals`, core takes `amount` from `line_items`, and quantity-only items have no `refund_total`, so it saves a 0.00 refund, restocks and returns 201. Before, the pinned amount saved us here.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessorTest.kt
```diff
@@ -206,7 +203,6 @@ class WooPosRefundSubmissionProcessorTest {
                 reason = eq("Customer request"),
```
kidinov, line 206: The `// THEN` comment above (197-199) still talks about the amount override.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessor.kt
samiuelson: I was thinking about it a bit more and reverted in f28a873.

You are right that nothing else checks it. `handleBackendRefundResult` only reads `result.isError`. And comparing `result.model.amount` with `request.refundAmount` would not replace the pin on this path: `api_refund` is `true` whenever the gateway supports refunds, so by the time the response arrives the money has moved.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
samiuelson: I reverted the "amount" param, but removed default values, so the .filterNotNull is not needed anymore.

WooCommerce/src/test/kotlin/com/woocommerce/android/ui/woopos/orders/details/refund/WooPosRefundSubmissionProcessorTest.kt
samiuelson: Fixed by the revert — the comment matches the code again.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCRefundStore.kt
```diff
@@ -110,24 +110,24 @@ class WCRefundStore @Inject internal constructor(
...
+     * is an optional override: below the computed total the store returns 400, above it the store
+     * accepts an over-refund up to the order's remaining refundable amount.
      *
-     * On a store whose refund endpoint does not support `compute_totals`, the unknown param is
-     * silently dropped and a quantity-only body would create a ghost zero-amount refund with
-     * restock. Only call this after a successful [previewRefund] against the same store confirmed
-     * server-side computation is available.
+     * A store without `compute_totals` drops the param and books a zero-amount refund with a
+     * restock, and still answers 201. Call this only after [previewRefund] succeeded on a store
+     * running WooCommerce 11.1.0 or newer. POS checks both in `WooPosResolveRefundFlow`.
```
kidinov: `WooPosResolveRefundFlow` does not check the preview here, it only rejects a cache entry which is already `false`. The `== true` check sits in `WooPosRefundViewModel.buildSubmissionRequest`. Maybe point to that one instead?

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
```diff
@@ -56,19 +56,14 @@ class RefundRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
...
      *
-     * [apiRefund] and [apiRestock] are always sent explicitly: the v3 endpoint defaults both to
-     * `true` when omitted, and the caller must stay in control of gateway refunds and restocking.
+     * [apiRefund] and [apiRestock] are sent explicitly because the endpoint defaults both to true.
      *
-     * On stores whose refund endpoint does not support `compute_totals`, the unknown param is
-     * silently dropped and a quantity-only body would create a ghost zero-amount refund with
-     * restock. Callers must therefore only use this method after a successful preview against the
-     * same store (see the availability cache in the POS refund flow).
+     * A store without `compute_totals` drops the param and books a zero-amount refund with a
```
kidinov: `api_restock` is always sent explicitly, so with `restockItems = false` there is no restock, only the zero amount refund. Same sentence is in `WCRefundStore` too. wdyt?

kidinov (approved):
LGTM! Just a few docs changes maybe needed

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/store/WCRefundStore.kt
samiuelson: You are right. The version gate is here, the preview gate is in `WooPosRefundViewModel.buildSubmissionRequest`. Fixed in acde5e2 — the KDoc now names both.

libs/fluxc-plugin/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
samiuelson: Correct, the restock only happens when `api_restock` is true. POS always sends true, but this is the FluxC layer, so I made the sentence conditional in both files (acde5e2). I also pointed this one at both POS gates.

PR #16440, opened 2026-08-18
[WOOMOB-3751] Recover from deterministic refund rejections instead of retrying

samiuelson:
### Description

Fixes WOOMOB-3751

The mapped refund errors always fail the same way, so the Retry we show cannot work. Each error now offers what can help:

- `order_not_refundable`: the "nothing to refund" screen.
- quantity / amount / already refunded: "Review remaining items", which reloads the order and keeps the selection that is still refundable.
- `invalid_refund_amount`: no action.
- unmapped errors: Retry, as before.

Unmapped create errors now show the generic message instead of the store's raw one. Copy fixes: the missing full stops, and the already-refunded message no longer claims to name one item.

iOS: woocommerce-ios#17732

### Test Steps

Needs a second client (web admin) to change the order while the flow is open.

1. Use store that supports server-calculated refunds - this requires WooCommerce 11.1.0 or newer, you can ask claude to generate .zip, or use nightly build from woocommerce monorepo. \*
2. In POS, start a refund on a paid order with an item of quantity 2 or more.
3. In the web admin, refund one unit of that item.
4. In POS, tap Continue: the store's reason shows, Continue is disabled, "Review remaining items" appears.
5. Tap it: the list reloads with the units still refundable and keeps your selection.
6. Refund the whole order in the web admin and repeat steps 2-4: the flow ends on the "nothing to refund" screen.
7. Turn off the network mid-preview: the error is generic and still offers Retry.
8. To hit the create path instead of the preview, let the web-admin refund land after you tap Continue but before you confirm: the error screen shows the store's reason with "Review remaining items" as the primary action and "Back to order" below it.

\* You can also use the test store cool-shop.mystagingwebsite.com.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

iamgabrielma:
👋  I won't be able to review the Android PRs until next week, can these be redirected to an Android dev?

samiuelson:
> 👋 I won't be able to review the Android PRs until next week, can these be redirected to an Android dev?

Hi @kidinov 👋 Would you be able to review this PR? 

Here's the recording on the error reproduction scenario:

<video src="https://github.com/user-attachments/assets/b7454044-0854-4b79-b1b8-7f35fc007e13"/>

kidinov (commented):
Tested on a JN store with WooCommerce trunk (11.2.0-dev), the flow works as you describe. One thing I would fix before merge.

#### Double refund on a card order

The mapped error code now wins over `canRetry`, so it also kicks in when the card was already refunded on the reader and only the store call failed. The screen then offers **Review remaining items** and the cashier can go through the flow and refund the same card a second time. Before this PR that case had "Back to order" only.

Repro:

1. Card paid order, one item with quantity 2 or more.
2. Start the refund in POS and let the terminal finish the card refund.
3. In web admin refund one unit of that item right after the terminal succeeds, while POS still shows "Notifying store".
4. POS shows the error with "Review remaining items". Tap it, confirm the refund again, and the card is refunded twice.

`order_not_refundable` is the same hole: POS jumps to the "nothing to refund" screen while the money already left and the store has no record of it.

The window is narrow, but it is the same race this PR is about, so I think it is worth guarding. Looking at `retryBackendNotificationOnly` before the mapped code should be enough.

#### Smaller things

When nothing from the old selection survives the reload, everything left gets selected. So the cashier picks 1 unit out of 5, that unit is refunded elsewhere, and after the reload the other 4 are selected with Continue enabled. Empty selection looks safer to me. wdyt?

Unmapped gateway rejections now show "Something went wrong" with a Retry that keeps failing. I get the locale argument, but that message was the only hint the cashier had

samiuelson:
Thanks for testing. Pushed fixes for the first two.

**Fixed**

- **Double refund** (fa82c4ba0fb, 5d18a15c3f3). The recovery now comes from `retryBackendNotificationOnly` directly instead of the mapped code, so a store rejection after the card was reversed ends at "Back to order". I did not route it through `canRetry` — that field is overridable and two call sites already set it. `order_not_refundable` falls under the same check and no longer jumps to the terminal screen.
- **The message** (5d18a15c3f3). "This order can't be refunded." read as if nothing had happened. Backend-only failures now say: "The card was refunded, but your store wasn't updated. Record this refund in your store admin."
- **Empty selection after a reload** (a6e38be0ce9). Continue stays disabled until the cashier picks again. First load still selects everything.
- **`refund_processing_failed` now carries `api_error_code`** (a23c6b7fe7a). It was documented as kept for analytics and never sent.

**On the repro**

I don't think it can run today, so the guard is defensive rather than a live fix. `retryBackendNotificationOnly` is only set in `handleSuccessfulInteracRefund`, so the split state is Interac-only — a normal card refund is one server call with `autoRefund`, where the reversal and the record are the same operation. POS doesn't collect Interac for now: it's behind the `isInteracEnabled` developer option, and POS only lists orders it created (`createdVia = "pos-rest-api"`), so there's no Interac order for it to refund either. Interac refunds are planned along with CA support, and the guard is there for when that lands.

The main app is a different story. `RefundSummaryViewModel` routes on the charge's `cardType` and its order list isn't filtered, so it can refund an `interac_present` charge from any terminal. On a failed store call it shows a snackbar and leaves `isFormEnabled = true` — the button stays live and the order still shows the amount as refundable, so tapping Refund again goes back to the reader. Same bug you found, reachable there. Worth its own issue.

**Not fixed**

Unmapped rejections keep the generic message and a Retry that may fail again. On the message I'd keep the generic string — the store's text is in the store's locale, not the cashier's. On the retry: I can't separate a deterministic rejection from a transient one without knowing the code. With `autoRefund` on, gateway timeouts come back coded too, and those are exactly the ones where retrying works. `WooPosRefundApiError` is the set we've confirmed deterministic, and `api_error_code` in analytics is how it grows. Say if you'd rather have the message back.

One structural note: the mapping moved to `WooPosMapRefundFailure` (5d18a15c3f3) — the ViewModel was over detekt's `LargeClass` limit.

kidinov (approved):
LGTM! Thanks!

PR #16441, opened 2026-08-18
[WOOMOB-3868] Add Store Design System tooltip

hichamboushaba:
### Description
Fixes WOOMOB-3868

Adds an anchored `WooTooltipBox` component with automatic or preferred logical-side placement, viewport clamping, adaptive arrow geometry, and caller-controlled persistent presentation. Material 3 owns standard input, popup lifecycle, accessibility, and single-tooltip coordination while Woo owns the Figma-backed appearance and positioning behavior.

This PR keeps tooltip content text-only. Optional actionable content is intentionally deferred to a separate stacked change.

### Test Steps
1. Open **Settings → Developer Options → Store Design System Component Catalog**, then select **Production → Tooltips**.
2. Show the automatic tooltip and confirm it is anchored to the trigger, remains inside the viewport, and dismisses when the bubble or outside content is tapped.
3. Show the preferred-End tooltip and confirm its side and arrow target follow the anchor; scroll the anchor offscreen and confirm the popup disappears.
4. Repeat in dark mode and an RTL locale, confirming inverse colors and logical Start/End mirroring.
5. Increase the system font scale and confirm the title and supporting text wrap without clipping.

### Images/gif
| Light | Dark |
| --- | --- |
| light-all-anchor-positions | dark-all-anchor-positions |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>WooTooltipPositionProvider</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipTest.kt
```diff
@@ -0,0 +1,660 @@
...
+        lateinit var scope: CoroutineScope
+        lateinit var showJob: Job
+        composeTestRule.setContent {
+            WooDesignSystemTheme {
+                state = rememberWooTooltipState()
+                scope = rememberCoroutineScope()
+                WooTooltipBox(
+                    state = state,
+                    title = TITLE,
+                    modifier = Modifier.offset(x = if (anchorIsOffscreen) (-1000).dp else 0.dp),
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 0838b4e

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipTest.kt
```diff
@@ -0,0 +1,660 @@
...
+        composeTestRule.setContent {
+            WooDesignSystemTheme {
+                firstState = rememberWooTooltipState()
+                secondState = rememberWooTooltipState()
+                scope = rememberCoroutineScope()
+                Column {
+                    WooTooltipBox(
+                        state = firstState,
+                        title = TITLE,
+                        modifier = Modifier.offset(x = if (firstAnchorIsOffscreen) (-1000).dp else 0.dp),
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 0838b4e

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipTest.kt
```diff
@@ -0,0 +1,660 @@
...
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 0838b4e

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipTest.kt
```diff
@@ -0,0 +1,660 @@
...
+        lateinit var scope: CoroutineScope
+        lateinit var showJob: Job
+        composeTestRule.setContent {
+            WooDesignSystemTheme {
+                state = rememberWooTooltipState()
+                scope = rememberCoroutineScope()
+                WooTooltipBox(
+                    state = state,
+                    title = TITLE,
+                    modifier = Modifier.offset(x = if (anchorIsOffscreen) (-1000).dp else 0.dp),
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 0838b4e

libs/store-design-system/src/test/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipTest.kt
```diff
@@ -0,0 +1,660 @@
...
+        lateinit var rehostedShowJob: Job
+        composeTestRule.setContent {
+            WooDesignSystemTheme {
+                state = rememberWooTooltipState()
+                scope = rememberCoroutineScope()
+                if (hostIsComposed) {
+                    WooTooltipBox(
+                        state = state,
+                        title = TITLE,
+                        modifier = Modifier.offset(x = if (hostIsOffscreen) (-1000).dp else 0.dp),
```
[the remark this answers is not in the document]
hichamboushaba: Fixed in 0838b4e

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltip.kt
```diff
@@ -0,0 +1,562 @@
...
+            anchorGap = with(density) { ANCHOR_GAP.roundToPx() },
+            maxBubbleWidth = with(density) { MAX_TOOLTIP_WIDTH.roundToPx() },
+            minSideWidth = with(density) { MIN_SIDE_WIDTH.roundToPx() },
+            geometry = geometryTokens,
+        )
+    }
+    var layoutResult by remember {
+        mutableStateOf(
+            WooTooltipLayoutResult(
+                offset = androidx.compose.ui.unit.IntOffset.Zero,
```
AdamGrzybkowski: ⛏️ We can import `androidx.compose.ui.unit.IntOffset`

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltip.kt
```diff
@@ -0,0 +1,562 @@
...
+ */
+@Composable
+fun WooTooltipBox(
+    state: WooTooltipState,
+    title: String,
+    modifier: Modifier = Modifier,
+    supportingText: String? = null,
+    preferredPlacement: WooTooltipPlacement? = null,
+    content: @Composable () -> Unit,
+) = WooTooltipBoxImpl(
```
AdamGrzybkowski, line 76: ❓ Question about the API design and the planned use. It seems like the current approach might be a bit tricky to work with if we have to drive the tooltip state from the ViewModel. 

1. Showing/Hiding the tooltip will require an extra `LaunchEffect` to call `show/hide()` functions, rather than driving the visibility from the state.
2. There's no easy-to-use `onDismiss` callback. `state.isVisible` can't be used as it's not snapshot-based.

Maybe I'm overthinking and we won't need that! If we decide to keep it as is, we still might want to fix the `state.isVisible`, the one from Material 3 is snapshot-backed, and only the override here broke it, if I understand correctly.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltipState.kt
```diff
@@ -0,0 +1,190 @@
...
+    private fun pauseOrDismissOffscreenRequest() {
+        val request = currentRequest ?: return
+        if (request.priority == MutatePriority.Default) {
+            request.isPaused = true
+            attemptCounter++
+            materialState.dismiss()
+        } else {
+            finish(request)
+        }
+    }
```
AdamGrzybkowski: ❓ Do we really need the logic to bring back the tooltip when the anchor is visible again? 

Scroll triggered by the user dragging their finger across the screen is not the scenario we are covering here, because the tap will dismiss the tooltip first. So this would be for some sort of scrollable content that happens without the user tapping the screen. Seems to me a bit overkill unless you have something specific in mind?

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltip.kt
hichamboushaba: Thanks @AdamGrzybkowski. The state-driven API follows M3’s `TooltipState` design, but you are right on `isVisible` issue. The additional logic on the state, and the override was added to match iOS: where an offscreen tooltip remains logically presented and reappears with its anchor.
But after reconsidering, and also related to your comment below, permanent dismissal is simpler and preferable on Android, so I’ll remove that layer and align the state/lifecycle API fully with M3, including snapshot-backed `isVisible` and `onDismissRequest`.

hichamboushaba: The commit 66034fa simplified the implementation, our State now is just a thin wrapper.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooTooltip.kt
hichamboushaba: Fixed in 66034fa

AdamGrzybkowski (approved):
Thanks for the updates!

PR #16446, opened 2026-08-19
Fix browser fallback for custom login challenges

hichamboushaba:
### Description

Follow-up to WOOMOB-3799 and #16398, which added recovery and retention for custom WordPress login and admin URLs.

That work correctly retains a custom login URL after it has been verified, but its new error routing also hid the existing browser/Application Password fallback when the subsequent native credential submission failed. A verified login form only proves that the endpoint is correct; CAPTCHA, 2FA, and other interactive security layers can still prevent native authentication. Invalid-credentials and invalid-response failures now return to the credentials form with the verified endpoint and credentials retained, expose the existing browser fallback, and never launch it until the merchant taps it.

Sites using WPS Hide Login need two additional safeguards. While logged out, opening the site-advertised Application Password URL directly can return 404, so the fallback starts at the verified custom login URL with a redirect to the advertised authorization target. If the WebView is already authenticated, WPS can ignore that redirect and attempt to navigate to the admin dashboard instead; the app now intercepts that same-origin main-frame navigation before the dashboard loads and performs a one-shot navigation to the advertised target in the same WebView, preserving its authenticated cookies. The existing page-finished check remains as a fallback. The admin directory is derived from the advertised URL, so rewritten/custom authorization paths remain supported.

The change reuses the existing failure classifications and authorization flow instead of relying on CAPTCHA- or plugin-specific strings. Pending, unverified recovery endpoints are neither used for browser routing nor persisted.

### Test Steps

Run the fixture script below from the WordPress root of a disposable test site. It installs the same WPS Hide Login and Simple Login Captcha versions used by the test fixture, configures `/private-login/`, and leaves existing users and credentials unchanged.

<details>
<summary>WP-CLI fixture setup script</summary>

```bash
#!/usr/bin/env bash

set -euo pipefail

LOGIN_SLUG="${1:-private-login}"

if ! command -v wp >/dev/null 2>&1; then
    echo "Error: wp-cli is not available." >&2
    exit 1
fi

if [[ ! "$LOGIN_SLUG" =~ ^[a-z0-9][a-z0-9_-]*$ ]]; then
    echo "Usage: $0 [login-slug]" >&2
    exit 1
fi

wp core is-installed

install_plugin() {
    local plugin="$1"
    local version="${2:-}"

    if ! wp plugin is-installed "$plugin"; then
        if [[ -n "$version" ]]; then
            wp plugin install "$plugin" --version="$version"
        else
            wp plugin install "$plugin"
        fi
    elif [[ -n "$version" && "$(wp plugin get "$plugin" --field=version)" != "$version" ]]; then
        wp plugin install "$plugin" --version="$version" --force
    fi
}

install_plugin woocommerce
wp plugin activate woocommerce

install_plugin simple-login-captcha 1.3.6
install_plugin wps-hide-login 1.9.19

wp option update whl_page "$LOGIN_SLUG"
wp option update whl_redirect_admin 404
wp plugin activate simple-login-captcha wps-hide-login
wp rewrite flush

site_url="$(wp option get home)"
authorization_url="$(wp eval 'echo admin_url("authorize-application.php");')"

printf '\nFixture ready.\n'
printf 'Custom login URL: %s/%s/\n' "${site_url%/}" "$LOGIN_SLUG"
printf 'Application Password URL: %s\n' "$authorization_url"
printf 'Use an existing administrator account; no credentials were changed.\n'
printf 'Recovery: wp plugin deactivate wps-hide-login simple-login-captcha && wp rewrite flush\n'

if ! wp eval 'exit(function_exists("wp_is_application_passwords_available") && wp_is_application_passwords_available() ? 0 : 1);'; then
    printf 'Warning: Application Passwords are not available on this site. Check that it uses HTTPS.\n' >&2
fi
```

</details>

An AI agent with approved access can provision a disposable Jurassic Ninja site and run the same WP-CLI setup automatically.

Use a real Android device for these steps. Emulator WebViews may be classified as automated traffic by Atomic/WP Cloud edge protections, causing the Application Password creation POST to return an HTML 429 response even when the app flow is correct.

1. Start site-credentials login, submit valid credentials, and provide the generated custom login URL when prompted.
2. Verify the native CAPTCHA failure returns to the credentials form with the endpoint and credentials retained.
3. Verify the browser/Application Password action is visible and that nothing opens automatically.
4. With the WebView logged out, tap the browser action, complete the browser login challenge, and verify the site-advertised Application Password authorization page opens.
5. Approve the connection and verify authentication completes in the app.
6. Repeat with an already-authenticated WebView session. Verify that the attempted WPS redirect to the dashboard is intercepted and the Application Password authorization page opens without rendering the dashboard first.
7. Rotate the device after reaching the authorization page and verify it remains there instead of restarting from the hidden login URL.
8. If the site advertises a rewritten/custom admin authorization path, verify that exact path is used.

### Images/gif

N/A

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/LoginActivity.kt
@@ -1000,9 +1000,17 @@ class LoginActivity :
         tag = LoginSiteCredentialsFragment.TAG
     )
 
-    override fun onApplicationPasswordHelpRequired(url: String, errorMessage: String) {
+    override fun onApplicationPasswordHelpRequired(
+        verifiedLoginUrl: String?,
+        applicationPasswordAuthorizationUrl: String,
+        errorMessage: String
+    ) {
         changeFragment(
-            fragment = ApplicationPasswordTutorialFragment.newInstance(url, errorMessage),
+            fragment = ApplicationPasswordTutorialFragment.newInstance(
+                verifiedLoginUrl = verifiedLoginUrl,
+                applicationPasswordAuthorizationUrl = applicationPasswordAuthorizationUrl,
+                errorMessage = errorMessage
+            ),
             shouldAddToBackStack = true,
             tag = ApplicationPasswordTutorialFragment.TAG
         )
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsFragment.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/sitecredentials/LoginSiteCredentialsFragment.kt
@@ -83,7 +83,11 @@ class LoginSiteCredentialsFragment : Fragment() {
                 is ShowHelpScreen -> loginListener.helpUsernamePassword(it.siteAddress, it.username, false)
                 is ShowSnackbar -> uiMessageResolver.showSnack(it.message)
                 is ShowApplicationPasswordTutorialScreen ->
-                    passwordTutorialListener?.onApplicationPasswordHelpRequired(it.url, it.errorMessage)
+                    passwordTutorialListener?.onApplicationPasswordHelpRequired(
+                        verifiedLoginUrl = it.verifiedLoginUrl,
+                        applicationPasswordAuthorizationUrl = it.applicationPasswordAuthorizationUrl,
+                        errorMessage = it.errorMessage
+                    )
                 is ShowUiStringSnackbar -> uiMessageResolver.showSnack(it.message)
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/299">25.5</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski (approved):
Works as described :shipit:

PR #16448, opened 2026-08-19
[WOOMOB-3868] Add WooTooltip action support

hichamboushaba:
### Description
Fixes WOOMOB-3868

Stacked on #16441. This PR adds narrowly scoped optional `WooTooltipAction` support to `WooTooltipBox` while preserving the existing inverse appearance and text-only behavior. The component owns a transparent, leading-aligned inverse text action and integrates Material's actionable-tooltip focus behavior; activating the action dismisses first and invokes its callback exactly once, while every other dismissal path remains silent.

This follows the legacy iOS `WooCommerce/Classes/ViewRelated/Feature Highlight/Tooltip.swift` implementation used by Product Details as the inverse action-bearing precedent. The newer StoreDesignSystem `storeTooltip` at `a4385756` remains title/message-only and has the same action gap. Android Product Details remains on its current app-owned implementation and is intentionally not migrated here.

### Test Steps
1. Open **Settings → Developer Options → Store Design System Component Catalog**, then select **Production → Tooltips**.
2. Tap **With action** and confirm the inverse tooltip shows its title, supporting text, and a transparent leading-aligned **Got it** action.
3. Tap **Got it** and confirm the tooltip dismisses. Reopen it, then tap the bubble background or outside content and confirm those paths only dismiss the tooltip.
4. With a hardware keyboard or emulator keyboard navigation, focus **With action**, press Tab to enter the tooltip action, then press Enter and confirm the tooltip dismisses.
5. Repeat in dark mode and an RTL locale, confirming inverse colors and logical leading alignment.
6. Increase the system font scale and confirm the title, supporting text, and a long action label wrap without clipping or overflowing the tooltip.

### Images/gif
| light | dark |
| ---- | ---- |
| image | image |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski (approved):
Looks good :shipit:

hichamboushaba:
Thanks @AdamGrzybkowski.

If you can please just take a look at the last commit to review it too, it replaced one tooltip on the product details screen with the new component:

| Light | Dark |
| --- | --- |
| image | image |

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRows.kt
```diff
@@ -370,9 +370,18 @@ private fun ProductDetailButtonRow(
     row: ProductDetailRowUiModel.Button,
     modifier: Modifier,
 ) {
-    var showTooltip by rememberSaveable(row.key) { mutableStateOf(row.tooltip != null) }
+    var isTooltipConsumed by rememberSaveable(row.key) { mutableStateOf(false) }
+    var hasAcknowledgedTooltip by rememberSaveable(row.key) { mutableStateOf(false) }
```
AdamGrzybkowski: ❓ Do we need both of these? What is the scenario where the Tooltip would be shown twice? Or is this to prevent double taps?

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRows.kt
```diff
@@ -383,49 +392,32 @@ private fun ProductDetailButtonRow(
...
-                    shadowElevation = 4.dp,
-                    shape = MaterialTheme.shapes.medium,
-                    modifier = Modifier
-                        .width(TOOLTIP_WIDTH)
-                        .background(WooTheme.colors.surface.surfaceContainerHighest),
+            val tooltip = row.tooltip
+            if (tooltip != null && !isTooltipConsumed) {
+                val tooltipState = rememberWooTooltipState()
+                LaunchedEffect(tooltipState) {
+                    tooltipState.show()
```
AdamGrzybkowski, line 432: ❓ I wonder if this is not one of those scenarios where it would be nicer to have Tootlip's visibility controlled directly via VM state instead of the `show/hide` functions. I know this follows Material 3, but some libraries [do it differently](https://composables.com/ui/docs/tooltip).

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRows.kt
hichamboushaba: Good point. The extra flag was only a defensive repeat-tap guard, while the acknowledgement write is idempotent and `isTooltipConsumed` already prevents the host from reopening after dismissal. Removed `hasAcknowledgedTooltip` and now invoke the action callback directly in 16befae3ed4.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/details/ProductDetailRows.kt
hichamboushaba: > but some libraries [do it differently](https://composables.com/ui/docs/tooltip).

I took a look at the linked API, and it isn’t actually controlling visibility externally: `enabled` only gates hover, focus, and long-press, while the tooltip owns its visibility internally. Am I missing something?

A controlled Woo API would need at least `visible`, `onVisibleChange`, and a separate `enabled`. The tricky part is that M3’s hover, focus, long-press, outside-dismissal, and global coordination paths call `TooltipState.show()` and `dismiss()` themselves. We would need to synchronize both directions: external `visible` changes would drive M3, while every M3 visibility change would need to update the caller. We’d also need to avoid feedback loops around the suspending `show()` call and distinguish temporarily hidden from permanently consumed, otherwise hover could reopen this onboarding tooltip. It’s doable, but it would make `WooTooltip` more than a thin M3 wrapper, so I think local state is the better fit here.

AdamGrzybkowski: >Am I missing something?

No. I only skimmed the examples and had the wrong assumptions! Thanks for the clarification and sorry for the noise. 

>It’s doable, but it would make WooTooltip more than a thin M3 wrapper, so I think local state is the better fit here.

Yeah, doing this on top of Material 3 would be cumbersome. I was worried we would lose the dismiss animation with the `if (tooltip != null && !isTooltipConsumed)`, but it's not the case here. 

Interestingly, this led me to discover that `isTooltipConsumed` is never set to true. `show()` suspends until the Tooltip is dismissed, so effectively `isTooltipConsumed = true` never runs. This means the tooltip will be reshown after a config change, or even navigation to another screen, until the user taps the action button. Maybe something like this `onDismissRequest = { isTooltipConsumed = true }` would work better here?


https://github.com/user-attachments/assets/3dd631e9-ddc4-48df-acd1-3eab55ea20ba

hichamboushaba: Good catch, and sorry I missed it. AI said that `show` would resume after dismissing, but it doesn't, it cancels, so the previous logic didn't work.
Fixed 0c2eb4b

PR #16451, opened 2026-08-20
[WOOMOB-3877] Hide the add product button when the search term is cleared

kidinov:
WOOMOB-3877

### Description

On the Products screen the add product button is hidden while search is open, except on a zero result search, where it is deliberately kept because that empty state has no add button of its own. Clearing the search term left it on screen, so you end up with a floating plus over a blank search screen.

Two things caused it. `onSearchQueryChanged` never recomputed the button visibility on the short query path, and `shouldShowAddProductButton` tested `query != null`, which stays true for the empty string the field holds after clearing.

I kept the button visible on a zero result search instead of hiding it for the whole search session as the issue asks. Hiding it there takes away the only way to add a product without leaving search, which the issue itself flags as worth a design call first.

This targets `feature/store-design-system-migration` rather than trunk, as requested on the issue.

### Test Steps

1. Open the Products tab.
2. Tap the search icon.
3. Type a term of at least 3 characters that matches no products, for example `zzzqqq`.
4. Wait for the no results message. The add product button should still be visible.
5. Clear the search term using the X in the search field, leaving search open.
6. The add product button should now be gone, leaving an empty search screen.
7. Type a term that does match products. The add product button stays hidden while results are shown.

### Images/gif
Before
woomob3877-before-cleared

After
woomob3877-after-cleared
woomob3877-after-no-results

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,9 +2,12 @@
 *** Use [*****] to indicate smoke tests of all critical flows should be run on the final APK before release (e.g. major library or targetSdk updates).
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 
-25.5
+25.6
 -----
 - [*] Fixed the add product button staying on screen after clearing the product search term [https://github.com/woocommerce/woocommerce-android/pull/16451]
+
+25.5
+-----
 - [Internal] Woo POS Scan to Pay: picking an offline method such as Pay in Person on the QR page no longer reports the order as paid, and a detected payment now shows the fullscreen success screen [https://github.com/woocommerce/woocommerce-android/pull/16363]
 - [*] Woo POS: Retrying a failed connection to a phone card reader now works without having to close and reopen the reader dialog [https://github.com/woocommerce/woocommerce-android/pull/16371]
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.5
 -----
+- [*] Fixed the add product button staying on screen after clearing the product search term
 - [Internal] Woo POS Scan to Pay: picking an offline method such as Pay in Person on the QR page no longer reports the order as paid, and a detected payment now shows the fullscreen success screen [https://github.com/woocommerce/woocommerce-android/pull/16363]
 - [*] Woo POS: Retrying a failed connection to a phone card reader now works without having to close and reopen the reader dialog [https://github.com/woocommerce/woocommerce-android/pull/16371]
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,7 +4,7 @@
 
 25.5
 -----
-- [*] Fixed the add product button staying on screen after clearing the product search term
+- [*] Fixed the add product button staying on screen after clearing the product search term [https://github.com/woocommerce/woocommerce-android/pull/16451]
 - [Internal] Woo POS Scan to Pay: picking an offline method such as Pay in Person on the QR page no longer reports the order as paid, and a detected payment now shows the fullscreen success screen [https://github.com/woocommerce/woocommerce-android/pull/16363]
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

samiuelson (commented):
Nice catch on the root cause, and I agree with keeping the button on a zero result search rather than hiding it for the whole search session.

I verified the change on a Pixel 9 Pro Fold emulator (folded, 1080x2424), building both this branch (`112c6c24df6`) and its base (`252a3b8de82`) so I could compare directly. The fix works: with no filter set, clearing the term hides the button.

Two cases still show the button over a blank search screen. Both come from the same place: `shouldShowAddProductButton()` cannot tell "search open, no search run yet" from "search ran and returned nothing".

### 1. A 1-2 character query shows the button (new in this PR)

Repro:
1. Products tab, tap search.
2. Type `ho` (any 1 or 2 characters).

The button appears over an empty screen with no empty state message.

On the base branch the same input keeps the button hidden, so this state is new. The short query path now recomputes visibility, and `!viewState.query.isNullOrEmpty()` is true for `"ho"`. Before this PR the short path never recomputed, so the button stayed hidden after `onSearchOpened()`.

This test passes on the PR head:

```kotlin
viewModel.onSearchOpened()
viewModel.onSearchQueryChanged("no")
advanceUntilIdle()
assertThat(visible.last()).isTrue()   // button visible over a blank screen
```

### 2. With a filter set, clearing the term still leaves the button

Repro:
1. Filters, Stock status, In stock, Show products. The chip reads `Filters (1)`.
2. Tap search, type `zzzqqq`, wait for the no results message. Button visible, as intended.
3. Tap the X to clear the term.

The button is still there over a blank screen. The same steps without a filter hide it correctly.

The empty query falls through to `productFilterOptions.isNotEmpty()`. Filters set before opening search stay in `productFilterOptions`, and search applies them too, so the original bug is unfixed whenever a filter is on.

This test also passes on the PR head:

```kotlin
viewModel.onFiltersChanged("instock", null, null, null, null)
viewModel.onSearchOpened()
viewModel.onSearchQueryChanged("nomatch"); advanceUntilIdle()
viewModel.onSearchQueryChanged(""); advanceUntilIdle()
assertThat(visible.last()).isTrue()   // button still visible
```

### Suggested fix

Branch on `isSearching()` and gate on the search threshold instead of on emptiness:

```kotlin
private fun shouldShowAddProductButton(): Boolean =
    if (_productList.value.isNullOrEmpty()) {
        if (isSearching()) {
            viewState.query.orEmpty().length >= MIN_SEARCH_QUERY_LENGTH
        } else {
            productFilterOptions.isNotEmpty()
        }
    } else {
        !isSearching()
    }
```

with `private const val MIN_SEARCH_QUERY_LENGTH = 3` reused by the two existing `query.length > 2` checks (lines 143 and 264), which currently repeat the same threshold in three places.

I applied this in a worktree and ran the suite. All 64 tests pass, including both tests this PR adds and both cases above inverted to expect `false`.

### Minor

- Trunk keeps the bug. There the logic sits inline in `resetViewState()` as `viewState.query != null -> true`, and `onSearchQueryChanged` has no recompute at all. Users get the fix only when this feature branch merges. Worth confirming that is the intent.
- The release note has no PR link. The file header asks for `[<PR URL>]` and the neighbouring entries have one.
- The two new tests assert only `.last()`. They pass, but they would also pass if the button flickered on and off in between. Asserting the emitted sequence would pin the behaviour down.
- The new comment on `shouldShowAddProductButton()` describes only the search cases, while the function also covers the non search filter path.

kidinov:
@samiuelson thanks for the review!

Both cases reproduce, thanks. Fixed both.

Minors done too: release note link, comment dropped, and all four button tests assert the sequence now instead of .last(). Typing "no" emits nothing at all, which was your flicker point.

On trunk, yes that is the intent, base is #16322, so fix ships when that lands.

wdyt?

samiuelson (approved):
LGTM!

PR #16452, opened 2026-08-21
[WOOMOB-3858] Stop translating the app_name launcher label

kidinov:
Fixes WOOMOB-3858

### Description

`app_name` had no `translatable="false"`, so it went to GlotPress and a translator read "Woo" as the English word for excitement, not as the brand name. Brazilian Portuguese users see **Eba!** on the launcher icon, Italian has **Woo!**. This marks the string non-translatable and removes the `app_name` override from all 16 localized `strings.xml`, so every locale falls back to English source.

I did not touch `fastlane/resources/values/strings.xml` on purpose. It is the frozen snapshot, and it gets regenerated from the main source as is on every code freeze, so it picks the attribute up by itself.

Still open: GlotPress keeps the existing `app_name` translations, and `download_translations` writes back whatever it returns. If a translations import runs before the next freeze, these deletions will come back. Someone with GlotPress access needs to remove the string there.

### Test Steps

1. Build and install the app.
2. Set the device language to **Português (Brasil)**.
3. Check the launcher icon label. It should read **Woo**, not **Eba!**.
4. Repeat with **Italiano**. It should read **Woo**, not **Woo!**.
5. Switch back to English and confirm the label is still **Woo**.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kidinov:

WooCommerce/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <string name="app_name">Woo</string>
```
malinajirka, line 3: 📓 If I recall correctly, changing string attributes + removing strings from non-main strings.xml causes some issues in the release build pipeline. I think it'd be safer to simply create a new entry `app_name_v2` and removing this one only from the main strings.xml.

malinajirka (approved):
Good catch!

WooCommerce/src/main/res/values/strings.xml
kidinov: I checked this one with help of AI. MissingTranslation and ExtraTranslation are both set to ignore in WooCommerce/lint.xml, so lint can't fail on it.

The immutable-key rule is about changing the value of a string, not the attributes. The Danger check for it (Automattic/dangermattic#126) skips lines with translatable="false" and allows removals, and we don't have that check enabled here anyway.

About removing the strings from values-*: those files are rewritten from scratch by the GlotPress export on every download_translations, so my deletions are only a cleanup for now. The attribute is the part that actually fixes it, after the next code freeze sends the frozen file to GlotPress and the original gets obsoleted there.

New key would work too, but it is more churn here - app_name is used as tools:title in ~70 layouts and overridden in wasabi/jalapeno. I'd rather keep it as is, as I didn't find any places where it brings any issues, wdyt?

The only rule is that the keys should be immutable and we do not break it here - paaHJt-9N9-p2#comment-11358

PR #16453, opened 2026-08-21
[WOOMOB-3794] Format the product discount screen values

kidinov:
Fixes WOOMOB-3794

### Description

The discount screen built its text by hand, `"$productQuantity x ${state.currency}$productPrice"`, so it printed a raw `Float` and a raw `BigDecimal` and you got `2.0 x €18.00000000`. The product line now comes from `getQuantityWithTotalText()`, the same one the order card uses, and totals go through `CurrencyFormatter`. Also `getPriceAfterDiscount()` returned `ZERO` when no discount was entered, so the row showed `€0` instead of the full price. That branch is gone.

Not visible in the diff: switching Amount to Percentage puts a rounded value in the input while the internal one stays precise, so Calculated Amount can read `US$200.00` instead of `US$199.94`. Old behaviour, I did not touch it.

### Test Steps

1. Orders, create order, add a product, set quantity to 2
2. Expand the card and tap **Add discount**
3. Line reads `2 × US$913.90` and the total `US$1,827.80`
4. With the amount field empty, **Price after discount** shows the full price, not `US$0`

### Images/gif

| No discount entered | Discount of 200 |
| -- | -- |
|  |  |

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,9 +2,12 @@
 *** Use [*****] to indicate smoke tests of all critical flows should be run on the final APK before release (e.g. major library or targetSdk updates).
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 
-25.5
+25.6
 -----
 - [*] Order creation: the discount screen now formats the product line and totals like the rest of the app, and "Price after discount" shows the full price until a discount is entered
+
+25.5
+-----
 - [Internal] Woo POS analytics: replaced the catch-all "other" reason on the POS ineligible screen with specific reasons, and added the missing orders_list_loaded event [https://github.com/woocommerce/woocommerce-android/pull/16439]
 - [WEAR][*] Order and stats dates on the watch now follow the language's date order [https://github.com/woocommerce/woocommerce-android/pull/16429]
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

RELEASE-NOTES.txt
```diff
@@ -2,9 +2,12 @@
 *** Use [*****] to indicate smoke tests of all critical flows should be run on the final APK before release (e.g. major library or targetSdk updates).
 *** For entries which are touching the Android Wear app's, start entry with `[WEAR]` too.
 
+25.6
+-----
+- [*] Order creation: the discount screen now formats the product line and totals like the rest of the app, and "Price after discount" shows the full price until a discount is entered
```
irfano: The PR link is missing here.

irfano (approved):
LGTM! 👍🏻 We just need to add the PR link to RELEASE-NOTES before merging.

RELEASE-NOTES.txt
kidinov: Fixed

PR #16461, opened 2026-08-24
[WOOMOB-3785] Bump androidx.fragment to 1.9.0 and re-land explicit back handling

AdamGrzybkowski:
### Description

Fixes WOOMOB-3785

The crash is an upstream androidx.fragment bug ([b/389589678](https://issuetracker.google.com/issues/389589678)), introduced in 1.7.0 and fixed in 1.9.0: the app crashes when the activity is sent to the background during a predictive back gesture. Bumping `androidx.fragment` to 1.9.0 fixes it.

#16454 reverted the explicit back-press handling (WOOMOB-3615) as a mitigation. That only stopped us from reaching the buggy library code rather than fixing it, so this re-lands it now that the library itself is fixed — the second commit here is a straight revert of that revert, unchanged from the originally reviewed #16319.

Verified on a physical Pixel 9 Pro with the back handling re-landed: 12/12 crashes on 1.8.9, 0/12 on 1.9.0.

### Test Steps

Requires Android 16+ with **gesture navigation enabled** (the back *button* does not trigger it).

1. Menu tab → **Coupons**
2. Back gesture from the left edge: drag out, drag back, drag out again — **never lifting your finger**
3. Still holding, drag slightly up
4. **Press POWER** to lock without lifiting your finger
5. Crash

Expected on this branch: no crash. On `trunk` before this change (with WOOMOB-3615 landed) the same steps crash with `IllegalArgumentException: The fragment ... is unknown to the FragmentNavigator`.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Thanks for taking care of this @AdamGrzybkowski 🙏🏼 
Managed to reproduce the crash on trunk `f1ddd678b953e0196437119981da176a69fc86b6` before release branch was merged and then checked the crash doesn't happen anymore on this branch. 

Didn't review the code given that was already reviewed in previous PR.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/list/ProductListToolbarHelper.kt
```diff
@@ -40,47 +42,36 @@ class ProductListToolbarHelper @Inject constructor(
...
-                            // In this case we need to switch panes – show the list pane instead of details pane.
-                            listFragment?.displayListPaneOnly()
-                        } else {
-                            listFragment?.findNavController()?.navigateUp()
-                        }
-                    }
-                }
-            }
-        )
+        (activity as FragmentActivity).onBackPressedDispatcher.addCallback(owner, backPressedCallback)
```
malinajirka: 💡 Small leak — every time the merchant leaves the Products tab and comes back we add another registration to this same callback and never remove the old one. `BaseFragment.kt:48` calls `remove()` before re-adding; this one doesn't.

```suggestion
        backPressedCallback.remove()
        (activity as FragmentActivity).onBackPressedDispatcher.addCallback(owner, backPressedCallback)
```

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/BackPressTracker.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val analyticsTrackerWrapper: AnalyticsTrackerWrapper,
+) {
+    private val mainHandler = Handler(Looper.getMainLooper())
+
+    // A consumed callback can enqueue a pop, so both tracking reports arrive before the main queue advances.
+    private var trackedInCurrentMainLoop = false
+
+    fun register(activity: Activity, fragmentManager: FragmentManager) {
+        fragmentManager.addOnBackStackChangedListener(backStackChangedListener(activity))
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
```
malinajirka: ❓ Below Android 16 we stopped tracking `BACK_PRESSED` when the user backs out of the app — nothing consumes that press, and the observer we rely on here is API 36 only.

Pre-existing from #16319, so follow-up rather than blocking this? Wdyt?

PR #16464, opened 2026-08-25
Remove obsolete `android_build_preflight`

mokagio:
### Description

The action runs checks that are mostly unnecessary:

- Decrypt secrets, which should already be decrypted
- Check gems are up to date, unrelated to app builds
- Ensure `bundletool` is available, but builds don't use it — its only
  client is `android_send_app_size_metrics`, which runs its own check

See https://github.com/wordpress-mobile/release-toolkit/blob/9322fff453fb4622d4d1add8040dc8d13cee2a1d/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_build_preflight.rb#L6-L23

`skip_prechecks` goes with it. Unlike the other repos, where it guards a block of checks, here it guarded only this one call — so it no longer skips anything, and `build_and_upload_google_play` stops passing it.

Closes https://linear.app/a8c/issue/AINFRA-2964

### Test Steps

Validated on CI by throwaway PR #16465, which ran `build_and_upload_beta` and stopped before the Play Store upload. The build [passed](https://buildkite.com/automattic/woocommerce-android/builds/41806). Lint took 10m55s and the bundle 5m34s before that marker, so the 107 MB AAB is a real build.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

iangmaia (approved):
👍

PR #16465, opened 2026-08-25
[DO NOT MERGE] Validate build lane for #16464

mokagio:
## Validation only — do not merge

Stacked on #16464 ([AINFRA-2964](https://linear.app/a8c/issue/AINFRA-2964)).
Close this PR once the job has answered; never merge it.

## Why the ordinary pipeline misses this

#16464 removes `android_build_preflight` from `build_and_upload_release` and `build_and_upload_beta`.
Neither lane runs on a PR: `prototype-build.sh` calls `build_and_upload_prototype_build`, and the two edited lanes are only reached from `release-build.sh` → `build_and_upload_google_play`, which the release pipeline triggers.

## Which job to read

**Validate build_and_upload_beta (no Play Store upload)** is expected to **pass**.
Red means #16464 actually broke the lane — do not "fix" this PR.

Assertions:

- The lane prints `VALIDATION: bundle built` with a non-zero size — a signed Release AAB came out of `build_bundle`, so nothing in the lane depended on the `configure_apply` that `android_build_preflight` used to run before it.

The AAB is attached as a build artifact.

## What CI still cannot answer

Whether the Play Store would accept the bundle. `upload_to_play_store` and `create_gh_release` are deleted on this branch rather than exercised.

### DIFF
--- a/.buildkite/commands/validate-build-lane.sh
+++ b/.buildkite/commands/validate-build-lane.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+"$(dirname "${BASH_SOURCE[0]}")/restore-cache.sh"
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+bundle exec fastlane run configure_apply
+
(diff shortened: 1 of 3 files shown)


mokagio:
Answered: the lane builds a 107 MB Release AAB without the preflight. Evidence moved to #16464; closing as planned.

PR #16467, opened 2026-08-25
[WOOMOB-3856] Fix order editing crash when order is missing from local DB

AdamGrzybkowski:
### Description
Fixes WOOMOB-3856

Opening a Customer Note / Shipping / Billing address editor crashed the app when the order was not present in the local database. `OrderEditingViewModel.start()` loaded the order with `requireNotNull(orderDetailRepository.getOrderById(orderId))` inside a `launch { }`; when the lookup returned `null`, the thrown `IllegalArgumentException` was an unhandled coroutine exception and closed the app.

This handles the missing order gracefully instead of crashing: `start()` now emits a new `OrderLoadFailed` event, and `BaseOrderEditingFragment` shows an "Error fetching order" snackbar and navigates back.

I did not add a network re-fetch fallback here: the editor is always entered from the Order Detail screen (which already loads/fetches the order), so a missing order at editor start is an edge case (order purged from the local cache by the 4h background sync, or process-death restoration while offline), and a graceful close keeps the fix minimal and safe to backport. I couldn't reproduce the crash without manually removing the Order from the DB.

### Test Steps
1. Open an order's detail screen, then open the **Edit Customer Note** (or Shipping/Billing address) editor.
2. While staying on the editor, remove that order's row from the local DB (`wc-android-database` → `OrderEntity`, matching `localSiteId` + `orderId`).
3. Rotate the device (or trigger a config change / process-death restore) so the editor re-runs its load.
4. Verify: an "Error fetching order" snackbar appears and the editor closes — the app does **not** crash.
5. Sanity check the normal path: open the same editor for an order that is present and confirm it loads and saves as before.

### Images/gif

https://github.com/user-attachments/assets/4f460b18-dff5-4c77-a328-1a453c29c409


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/editing/OrderEditingViewModel.kt
```diff
@@ -60,9 +61,13 @@ class OrderEditingViewModel @Inject constructor(
             } else {
                 orderId
             }
-            order = requireNotNull(orderDetailRepository.getOrderById(orderId)) {
-                "Order $orderId not found in the database."
+            val loadedOrder = orderDetailRepository.getOrderById(orderId)
+            if (loadedOrder == null) {
+                WooLog.w(WooLog.T.ORDERS, "Order $orderId not found in the database.")
```
irfano: Now that this no longer crashes, Sentry loses the signal, so we won't know if orders keep going missing in the field. Can we also send a non-fatal report here? `CrashLogging` is injectable, something like:
```kotlin
crashLogging.sendReport(message = "Order $orderId not found in the database.")
```

irfano (approved):
LGTM! Left one small note about keeping Sentry visibility for this case, non-blocking.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/details/editing/OrderEditingViewModel.kt
AdamGrzybkowski: Makes sense! I did pass a stable exception message, hoping this will be properly grouped rather than being treated as a separate error each time. 

Let's see if it's not too noisy; if it is, we can move it to an analytic event.

AdamGrzybkowski:

PR #16468, opened 2026-08-25
[WOOMOB-3764] Treat an undecodable Blaze ad image as unsupported instead of crashing

kidinov:
### Description

Fixes WOOMOB-3764

`BitmapFactory` leaves `Options.outMimeType` null whenever it cannot decode an image, but it is a Java field, so Kotlin sees `String!` and `ImageDetails.mimeType` was declared non-null. Any corrupt, truncated or device-unsupported image killed the app in the middle of Blaze campaign creation. The mime type is nullable now, and an image without one falls into the existing "unsupported image type" path.

- `AdImageValidationResult.Valid` carries the mime type it just validated, so the two places that build a `RemoteImage` do not need `orEmpty()` or `!!` on a value they already know is good.
- Behaviour change: `RemoteImage` now gets the normalized mime type from `Valid`, so the API is sent the trimmed lowercased value instead of the raw `outMimeType`.
- Not fixed here: the local image path in `prepareCampaignImage` still sends an empty string when the uploaded media has no mime type. Pre-existing.
- My test store is connected with an application password, so Blaze is not eligible on it and the entry point never shows. To walk the flow on a device I stubbed `IsBlazeEnabled` to `true` locally. That stub is not in the diff, so you need a Blaze eligible store for the steps below.

### Test Steps

The image you need is [broken.jpg](https://github.com/user-attachments/assets/aa28a86b-4250-4ee6-8c72-4ac7dd5eaa01), attached to this PR. It is 20 bytes, a JPEG SOI plus a JFIF header and nothing after it. There is no SOF marker, so bounds decoding fails and `outMimeType` comes back null. MediaStore still lists it as `image/jpeg` from the extension.

1. Get it onto the device:

```bash
curl -L -o broken.jpg https://github.com/user-attachments/assets/aa28a86b-4250-4ee6-8c72-4ac7dd5eaa01
adb push broken.jpg /sdcard/Pictures/broken.jpg
adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Pictures/broken.jpg
```

2. Open any product, tap Promote with Blaze, then Start your campaign, wait for the preview and tap Edit ad.
3. Tap Change image, then Choose from device, and pick `broken.jpg`. It is the newest item in the picker and it has a blank grey thumbnail.
4. The "Invalid image" dialog is shown and the ad image stays as it was. On trunk the app dies here with `NullPointerException: outMimeType must not be null` at `MediaFilesRepository.kt:81`.
5. Pick a normal jpg or png instead. It is accepted, the preview shows it, and campaign creation still works.

### Images/gif

Step 4 with the fix, on a Pixel 9 emulator:

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -35,7 +35,6 @@ import org.wordpress.android.fluxc.store.MediaStore.OnMediaUploaded
 import org.wordpress.android.mediapicker.MediaPickerUtils
 import org.wordpress.android.util.MediaUtils
 import java.io.File
-import java.io.FileDescriptor
 import java.io.IOException
 import java.net.URL
 import javax.inject.Inject
@@ -74,14 +73,18 @@ class MediaFilesRepository @Inject constructor(
                     BitmapFactory.decodeStream(URL(uri).openConnection().getInputStream(), null, options)
                 } else {
                     val parcelFileDescriptor = context.contentResolver.openFileDescriptor(Uri.parse(uri), "r")
-                    val fileDescriptor: FileDescriptor = parcelFileDescriptor!!.fileDescriptor
-                    BitmapFactory.decodeFileDescriptor(fileDescriptor, null, options)
-                    parcelFileDescriptor.close()
+                    if (parcelFileDescriptor == null) {
+                        WooLog.w(T.MEDIA, "MediaFilesRepository > Couldn't open a descriptor for uri: $uri")
+                        return@withContext UNDECODABLE_IMAGE_DETAILS
+                    }
+                    parcelFileDescriptor.use {
+                        BitmapFactory.decodeFileDescriptor(it.fileDescriptor, null, options)
+                    }
                 }
                 return@withContext ImageDetails(options.outWidth, options.outHeight, options.outMimeType)
             } catch (e: IOException) {
                 WooLog.e(T.MEDIA, "MediaFilesRepository > Error getting image details from uri: $uri", e)
-                return@withContext ImageDetails(width = 0, height = 0, mimeType = "")
+                return@withContext UNDECODABLE_IMAGE_DETAILS
             }
         }
     }
@@ -231,6 +234,10 @@ class MediaFilesRepository @Inject constructor(
     data class ImageDetails(
         val width: Int,
         val height: Int,
-        val mimeType: String
+        val mimeType: String?
     )
+
+    companion object {
+        private val UNDECODABLE_IMAGE_DETAILS = ImageDetails(width = 0, height = 0, mimeType = null)
+    }
 }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -85,6 +85,9 @@ class MediaFilesRepository @Inject constructor(
             } catch (e: IOException) {
                 WooLog.e(T.MEDIA, "MediaFilesRepository > Error getting image details from uri: $uri", e)
                 return@withContext UNDECODABLE_IMAGE_DETAILS
+            } catch (e: SecurityException) {
+                WooLog.e(T.MEDIA, "MediaFilesRepository > No access to uri: $uri", e)
+                return@withContext UNDECODABLE_IMAGE_DETAILS
             }
         }
     }
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
@@ -70,7 +70,9 @@ class MediaFilesRepository @Inject constructor(
             try {
                 val options = Options().apply { inJustDecodeBounds = true }
                 if (Patterns.WEB_URL.matcher(uri).matches()) {
-                    BitmapFactory.decodeStream(URL(uri).openConnection().getInputStream(), null, options)
+                    URL(uri).openConnection().getInputStream().use {
+                        BitmapFactory.decodeStream(it, null, options)
+                    }
                 } else {
                     val parcelFileDescriptor = context.contentResolver.openFileDescriptor(Uri.parse(uri), "r")
                     if (parcelFileDescriptor == null) {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
@@ -193,14 +193,15 @@ class BlazeRepository @Inject constructor(
             productId = productId,
             tagLine = product.name,
             description = description.fastStripHtml(),
-            campaignImage = product.images.firstOrNull().let {
-                val imageDetails = it?.source?.let { uri -> getImageDetails(uri) }
-                if (imageDetails?.isValidAdImage() == true) {
-                    BlazeCampaignImage.RemoteImage(it.source, imageDetails.mimeType)
-                } else {
-                    BlazeCampaignImage.None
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
```diff
@@ -74,14 +73,21 @@ class MediaFilesRepository @Inject constructor(
                     BitmapFactory.decodeStream(URL(uri).openConnection().getInputStream(), null, options)
```
samiuelson: The `use` block on the descriptor also closes it when decoding throws. Good.

The URL branch has the same problem. The stream from `openConnection().getInputStream()` is never closed. This path runs for every product with a remote image, through `generateDefaultCampaignDetails`.

```kotlin
URL(uri).openConnection().getInputStream().use {
    BitmapFactory.decodeStream(it, null, options)
}
```

This is pre-existing, so it is optional here.

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
```diff
@@ -74,14 +73,21 @@ class MediaFilesRepository @Inject constructor(
...
                 } else {
                     val parcelFileDescriptor = context.contentResolver.openFileDescriptor(Uri.parse(uri), "r")
```
samiuelson, line 77: Some providers throw `IllegalArgumentException` ("Unknown URI") instead of `FileNotFoundException`. Neither catch covers that. The URIs come from the media picker, so this is unlikely. Only relevant if you want the method to handle every input.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
```diff
@@ -193,14 +193,15 @@ class BlazeRepository @Inject constructor(
             productId = productId,
             tagLine = product.name,
             description = description.fastStripHtml(),
-            campaignImage = product.images.firstOrNull().let {
-                val imageDetails = it?.source?.let { uri -> getImageDetails(uri) }
-                if (imageDetails?.isValidAdImage() == true) {
-                    BlazeCampaignImage.RemoteImage(it.source, imageDetails.mimeType)
-                } else {
-                    BlazeCampaignImage.None
+            campaignImage = product.images.firstOrNull()?.let { image ->
```
samiuelson, line 196: Both `generateDefaultCampaignDetails` tests stub `ImageDetails(0, 0, null)`. Only the `None` branch runs. The `Valid` branch has no test.

A case with `ImageDetails(VALID_IMAGE_SIZE, VALID_IMAGE_SIZE, "IMAGE/JPEG")` asserting `RemoteImage(imageUrl, "image/jpeg")` would cover the branch and the normalization.

samiuelson (commented):
LGTM!

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
```diff
@@ -399,15 +400,16 @@ class BlazeRepository @Inject constructor(
...
         return when {
-            normalizedMimeType !in SUPPORTED_BLAZE_IMAGE_MIME_TYPES -> AdImageValidationResult.UnsupportedMimeType
+            normalizedMimeType == null ||
+                normalizedMimeType !in SUPPORTED_BLAZE_IMAGE_MIME_TYPES -> AdImageValidationResult.UnsupportedMimeType
+
             width < BLAZE_IMAGE_MINIMUM_SIZE_IN_PIXELS ||
                 height < BLAZE_IMAGE_MINIMUM_SIZE_IN_PIXELS -> AdImageValidationResult.InvalidSize
-            else -> AdImageValidationResult.Valid
+
+            else -> AdImageValidationResult.Valid(normalizedMimeType)
```
samiuelson, line 412: `Valid` now carries the normalized mime type. The API gets the trimmed, lowercased value instead of the raw `outMimeType`. `SUPPORTED_BLAZE_IMAGE_MIME_TYPES` is all lowercase, so the value is always canonical. This is a behavior change that is not in the description. Please add it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/creation/ad/BlazeCampaignCreationEditAdViewModel.kt
```diff
@@ -142,24 +142,24 @@ class BlazeCampaignCreationEditAdViewModel @Inject constructor(
...
 
     private suspend fun handleSelectedImage(
         uri: String,
-        createCampaignImage: (MediaFilesRepository.ImageDetails) -> BlazeRepository.BlazeCampaignImage
+        createCampaignImage: (mimeType: String) -> BlazeRepository.BlazeCampaignImage
     ) {
         val imageDetails = blazeRepository.getImageDetails(uri)
-        when (imageDetails.validateAdImage()) {
-            Valid -> {
+        when (val validationResult = imageDetails.validateAdImage()) {
```
samiuelson, line 159: An undecodable JPEG takes the `UnsupportedMimeType` branch and reports "unsupported image type". The type is supported. The file is broken. Reusing the string avoids a new translation, so this is fine to ship. A separate "this image could not be read" string could be added later.

samiuelson (approved):

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
kidinov: Good catch, closed it in b4abf95

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
kidinov: Skipping, picker URIs only

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
kidinov: Added a test in b4abf95

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/BlazeRepository.kt
kidinov: Added it to the description

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/blaze/creation/ad/BlazeCampaignCreationEditAdViewModel.kt
kidinov: Agree, leaving the string

WooCommerce/src/main/kotlin/com/woocommerce/android/media/MediaFilesRepository.kt
```diff
@@ -74,14 +73,21 @@ class MediaFilesRepository @Inject constructor(
...
+                    parcelFileDescriptor.use {
+                        BitmapFactory.decodeFileDescriptor(it.fileDescriptor, null, options)
+                    }
                 }
                 return@withContext ImageDetails(options.outWidth, options.outHeight, options.outMimeType)
             } catch (e: IOException) {
                 WooLog.e(T.MEDIA, "MediaFilesRepository > Error getting image details from uri: $uri", e)
-                return@withContext ImageDetails(width = 0, height = 0, mimeType = "")
+                return@withContext UNDECODABLE_IMAGE_DETAILS
+            } catch (e: SecurityException) {
```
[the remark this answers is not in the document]
kidinov: @claude rejected: guards are three-line early returns, and testing them needs a new Robolectric class for little value.

kidinov:

PR #16469, opened 2026-08-25
[WOOMOB-3865] Enforce HTTPS during site login

JorgeMucientes:
### Description

Fixes WOOMOB-3865

Normalizes store addresses to HTTPS throughout new-site discovery and site-credentials authentication. Explicit HTTP and scheme-less inputs are upgraded before the first request, and server-advertised site, admin, login, REST root, and application-password authorization URLs are normalized before use or persistence.

Certificate failures keep the existing actionable security guidance; other HTTPS connection failures retain the generic connection error. The login flow never retries the original HTTP URL and never creates or selects a site after a failed HTTPS connection.

This is the first PR in a two-part stack. #16470 adds existing logged-in store migration and the configuration warning on top of this branch.

### Test Steps

1. Sign out, then choose the site-address login flow.
2. Enter an HTTPS-capable store using an `http://` address and complete login.
3. Confirm login succeeds and the stored site, admin, login, REST root, and application-password authorization URLs use `https://`. Use network instecptor or proxyman for this. 
4. Log in with a store already configured with `https://` and confirm the flow is unchanged.
5. Sign out, enter the following site address `http://expired.badssl.com`, and tap **Continue**. Confirm you remain on the site-address screen and see the inline certificate error.
6. Exercise the same certificate failure through the direct site-credentials flow. Wrap the complete remote command so the device shell does not interpret `&username` as a shell operator:
   ```bash
   adb shell 'am start -a android.intent.action.VIEW \
     -d "woocommerce://app-login?siteUrl=http%3A%2F%2Fexpired.badssl.com&username=reviewer"'
   ```
   Enter any dummy password and continue. Confirm the certificate dialog is shown.
7. Simulate a server accepting HTTP (setup explained below). The app correctly upgrades the site address to HTTPS, but the server responds as HTTP and cannot perform TLS. App will show an error (see screenshots below)

Use the following three-terminal setup to test a plaintext-only endpoint. All commands run on the computer connected to the emulator or device.

   **Terminal 1 — start the plaintext server on an automatically selected free port:**
   ```bash
   python3 -u - <<'PY' 2>&1 | tee /tmp/woo-http-server.log
   from http.server import ThreadingHTTPServer, SimpleHTTPRequestHandler

   server = ThreadingHTTPServer(("127.0.0.1", 0), SimpleHTTPRequestHandler)
   print(f"Serving HTTP on 127.0.0.1 port {server.server_port}", flush=True)
   server.serve_forever()
   PY
   ```
   Port `0` asks the operating system to allocate a free local port, avoiding conflicts with services already using `8080`. The server output appears live in Terminal 1, and `tee` copies it to `/tmp/woo-http-server.log`.

   **Terminal 2 — read that port, connect the emulator or physical device, and run a browser preflight:**
   ```bash
   woo_http_port="$(sed -nE 's/^Serving HTTP on 127\.0\.0\.1 port ([0-9]+)$/\1/p' /tmp/woo-http-server.log | tail -1)"
   printf 'Using local port %s\n' "$woo_http_port"

   adb reverse "tcp:$woo_http_port" "tcp:$woo_http_port"
   adb shell am start -a android.intent.action.VIEW -d "http://127.0.0.1:$woo_http_port/"
   ```
   Confirm the device browser shows the Python directory listing and Terminal 1 prints a valid request such as `"GET / HTTP/1.1" 200 -`. This proves that device `127.0.0.1:$woo_http_port` reaches the Python server through `adb reverse`.

   Once the preflight requests have appeared, record the current end of the log. This excludes those browser requests from the later app-only assertion without restarting the server:
   ```bash
   sleep 1
   wc -l < /tmp/woo-http-server.log > /tmp/woo-http-server.start-line
   ```

   **Terminal 2 — open the direct site-credentials flow:**
   ```bash
   adb shell "am start -a android.intent.action.VIEW \
     -d 'woocommerce://app-login?siteUrl=http%3A%2F%2F127.0.0.1%3A${woo_http_port}&username=reviewer'"
   ```
   If Android shows an app chooser because multiple WooCommerce builds are installed, select the PR build. Enter any dummy password and continue. Confirm login fails without creating or selecting a site. Depending on the Android TLS provider and Python version, the app may show the generic website-fetch error rather than certificate guidance; step 7 covers the certificate-specific UI. The assertion here is that the app tries HTTPS and never retries plaintext HTTP.

   **Terminal 3 — inspect only the server output produced after the app attempt:**
   ```bash
   woo_http_start_line="$(cat /tmp/woo-http-server.start-line)"
   tail -n "+$((woo_http_start_line + 1))" /tmp/woo-http-server.log \
     | tee /tmp/woo-http-server-app.log

   if grep -Eq '"(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS) ' /tmp/woo-http-server-app.log; then
     echo 'FAIL: a valid plaintext HTTP request reached the server'
   else
     echo 'PASS: no valid plaintext HTTP request reached the server'
   fi
   ```
   Run this command in any separate computer terminal, not inside the emulator/device. It first prints the app-only server output, so the captured logs are visible there as well as in Terminal 1.

   Either of these outcomes is a pass:
   - No app-only server output.
   - A `400 Bad request` containing escaped TLS bytes such as `\x16\x03`, followed by the `PASS` line. The malformed request is the HTTPS handshake reaching a plaintext server.

   A valid HTTP method such as `"GET /…"`, `"POST /…"`, or `"HEAD /…"` in `/tmp/woo-http-server-app.log` is a failure because it means the app retried plaintext HTTP.

   **Cleanup:** stop Terminal 1 with `Ctrl+C`, then remove the exact dynamic port mapping from Terminal 2 (or recompute `woo_http_port` from the log first if using another terminal):
   ```bash
   woo_http_port="$(sed -nE 's/^Serving HTTP on 127\.0\.0\.1 port ([0-9]+)$/\1/p' /tmp/woo-http-server.log | tail -1)"
   adb reverse --remove "tcp:$woo_http_port"
   ```

### Images/gif

| Step 5  | Step 6 | Step 7  |
| ------------- | ------------- | ------------- |
|    |   |  |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>

<ul>
</ul></td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/main/java/org/wordpress/android/fluxc/utils/HttpsUrlNormalizer.kt
```diff
@@ -0,0 +1,63 @@
...
+class HttpsUrlNormalizer @Inject constructor() {
+    data class Result(
+        val normalizedUrl: String,
+        val wasUpgraded: Boolean,
+    )
+
+    @JvmOverloads
+    fun normalize(rawUrl: String, addHttpsSchemeIfMissing: Boolean = false): Result {
+        val trimmedUrl = rawUrl.trim()
+        val scheme = SCHEME_REGEX.find(trimmedUrl)?.groupValues?.get(1)?.lowercase(Locale.ROOT)
```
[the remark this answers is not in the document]
JorgeMucientes: Fixed in d81590c9c52

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/site/SiteWPAPIRestClient.kt
```diff
@@ -72,10 +83,11 @@ class SiteWPAPIRestClient @Inject constructor(
 
                     applicationPasswordsAuthorizeUrl = response.authentication?.applicationPasswords
                         ?.endpoints?.authorization
+                        ?.normalizeOptionalUrl()
                     adminUrl = inferAdminBaseUrl(applicationPasswordsAuthorizeUrl)
 
                     wpApiRestUrl = discoveredWpApiUrl
-                    this.url = cleanedUrl.replaceBefore("://", urlScheme)
+                    this.url = serverUrl?.normalizedUrl ?: cleanedUrl
```
irfano: The REST index `url` field is the WordPress Address (`siteurl`). It can be different from the address the user typed, for example WordPress installed in a sub folder, `www` vs no `www`, or plugins that change the URL. I tested this with a store where the index returns `https://site/wp`. The login itself works. After that, `WPApiSiteRepository.fetchSite` looks for the stored site using the typed address, does not find it, and shows "An error occurred while fetching your website". On trunk the same login works. https://github.com/woocommerce/woocommerce-android/commit/d9dae31504156ed2a3b766847d05ff35757964ea removed this behaviour before for the same reason. Keeping `url` from `cleanedUrl` and using the server URL only for the http/https state avoids this. Keeping the server URL like iOS does would also work, but then the lookup by typed address after the fetch has to go as well.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -238,16 +238,24 @@ class SiteRestClient @Inject constructor(
...
+                    result.data.url?.let(httpsUrlNormalizer::normalize)
+                } catch (_: IllegalArgumentException) {
+                    return SiteModel().apply { error = BaseNetworkError(GenericErrorType.INVALID_RESPONSE) }
+                }
                 // Keep existing fields, and update only fields fetched from the root endpoint
                 site.apply {
                     name = result.data.name
                     timezone = result.data.gmtOffset
+                    serverUrl?.let {
+                        url = it.normalizedUrl
```
irfano: This does the same thing as the change in `SiteWPAPIRestClient`. It replaces the WP.com URL with the server's `siteurl`. I tested it with a Jetpack CP store where the REST index returns `https://site/wp`. The store list shows `site/wp`, and login with the site address ends on the "account not connected to this site's Jetpack" screen. On trunk the same login opens the store. Leaving `url` as it is here and using `serverUrl` only for the http/https state avoids this.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/site/SiteWPAPIRestClient.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/aa8bc796a47f45362f995a7e6ed8feda9841d1fe, the stored site URL is the entered address again.

libs/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/071fa75f245d67ceb677459ede9d647f0804a802, the root endpoint no longer replaces the WP.com URL.

irfano (approved):
Approving. Jorge is AFK, so I pushed the two small fixes for my own comments directly to the branch and tested them on an emulator:
- https://github.com/woocommerce/woocommerce-android/commit/aa8bc796a47 keeps the entered address as the site URL for application password stores
- https://github.com/woocommerce/woocommerce-android/commit/071fa75f245 stops replacing the WP.com site URL with the root endpoint URL

Both threads are resolved.

PR #16470, opened 2026-08-25
[WOOMOB-3865] Warn stores still configured for HTTP

JorgeMucientes:
### Description

Fixes WOOMOB-3865

Stacked on #16469.

Extends HTTPS enforcement to existing logged-in stores. Shared WPAPI request builders and refreshed server-provided URLs are normalized before use, so existing sessions migrate naturally on their next request or synchronization without probing or falling back to HTTP.

The PR also persists whether WordPress still advertises an HTTP address and shows a dismissible warning beneath the shared toolbar on top-level store-management destinations. Dismissal is scoped per local site for 24 hours, a fresh HTTPS configuration clears the warning, and failed syncs preserve the last known state. POS remains out of scope.

Application-password site refreshes retain the existing local ID. A uniqueness collision fails safely without deleting or merging stored sites. Malformed or scheme-less legacy admin/login URL values retain their previous fallback behavior instead of throwing during normalization.

### Test Steps

Use a disposable self-hosted test store with SSH and WP-CLI access. The temporary MU plugin below keeps the site reachable over HTTPS but makes the WordPress REST index advertise `http://` URLs. This reproduces the server state consumed by this PR without requiring control of a TLS listener or CDN.

Do not substitute #16469 step 8's plaintext Python server for this setup. Its `app-login` deep link exercises the login flow owned by #16469, and a plaintext server cannot return the authenticated WordPress REST-index state needed to exercise this PR's existing-store migration and warning. A genuinely HTTP-only server should fail after the app upgrades its request to HTTPS; it cannot produce the configuration warning because no REST response is available.

1. Before enabling the simulation, install this build and log in to the test store using site credentials and its normal HTTPS address. Confirm the store loads successfully.
2. SSH into the site and install the temporary simulation:
   ```bash
   wp eval 'wp_mkdir_p( WPMU_PLUGIN_DIR ); file_put_contents( WPMU_PLUGIN_DIR . "/woo-http-only-test.php", base64_decode( "PD9waHAKYWRkX2ZpbHRlciggJ3Jlc3RfaW5kZXgnLCBmdW5jdGlvbiAoICRyZXNwb25zZSApIHsKICAgICRkYXRhID0gJHJlc3BvbnNlLT5nZXRfZGF0YSgpOwoKICAgIGZvcmVhY2ggKCBhcnJheSggJ3VybCcsICdob21lJyApIGFzICRrZXkgKSB7CiAgICAgICAgaWYgKCBpc3NldCggJGRhdGFbICRrZXkgXSApICkgewogICAgICAgICAgICAkZGF0YVsgJGtleSBdID0gcHJlZ19yZXBsYWNlKCAnI15odHRwczovLyNpJywgJ2h0dHA6Ly8nLCAkZGF0YVsgJGtleSBdICk7CiAgICAgICAgfQogICAgfQoKICAgICRyZXNwb25zZS0+c2V0X2RhdGEoICRkYXRhICk7CiAgICByZXR1cm4gJHJlc3BvbnNlOwp9ICk7Cg==" ) );'
   ```
3. Verify WordPress now advertises HTTP while the site itself remains available over HTTPS:
   ```bash
   wp eval '$data = rest_do_request( "/" )->get_data(); printf( "url=%s\nhome=%s\n", $data["url"], $data["home"] );'
   ```
   Both values should start with `http://`.
4. Return to the app and pull to refresh My Store, Orders, or Products. Confirm requests continue using HTTPS, the existing store remains selected, and no duplicate store appears in the store switcher.
5. Confirm the “Secure your store with HTTPS” warning appears below the toolbar on the top-level My Store, Orders, Products, and More screens.
6. Open a nested destination and confirm the warning is hidden. Launch POS and confirm it is unaffected.
7. Tap “Learn more” and confirm the WordPress HTTPS documentation opens.
8. Dismiss the warning and confirm it hides immediately and remains hidden after relaunching or switching away from and back to the store. Move the device clock beyond 24 hours, relaunch, and confirm it becomes eligible to appear again.
9. While the warning state is present, force an offline/failed site refresh and confirm the last known warning state is preserved.
10. Remove the simulation:
    ```bash
    wp eval '$path = WPMU_PLUGIN_DIR . "/woo-http-only-test.php"; if ( file_exists( $path ) ) { unlink( $path ); }'
    ```
11. Refresh the store again. Confirm the warning clears and does not appear after relaunching.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/extensions/SiteModelExt.kt
```diff
@@ -44,10 +45,14 @@ val SiteModel.isSimpleWPComSite
...
-    get() = adminUrl ?: url.slashJoin("wp-admin")
+    get() = (adminUrl ?: url.slashJoin("wp-admin")).normalizedToHttps()
 
 val SiteModel.loginUrlOrDefault
-    get() = loginUrl ?: url.slashJoin("wp-login.php")
+    get() = (loginUrl ?: url.slashJoin("wp-login.php")).normalizedToHttps()
+
+private val httpsUrlNormalizer = HttpsUrlNormalizer()
+
+private fun String.normalizedToHttps() = httpsUrlNormalizer.normalize(this).normalizedUrl
```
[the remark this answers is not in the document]
JorgeMucientes: @claude addressed - Fixed in 04ec97b7d43.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -156,6 +160,8 @@ class MainActivity :
         private const val KEY_UNFILLED_ORDER_COUNT = "unfilled-order-count"
 
         private const val DIALOG_NAVIGATOR_NAME = "dialog"
+        private const val HTTPS_CONFIGURATION_LEARN_MORE_URL =
+            "https://developer.wordpress.org/advanced-administration/security/https/"
```
irfano: iOS links to this section of the Woo doc: https://woocommerce.com/document/ssl-and-https/#for-new-websites-stores. Using the same link here keeps both apps on the same page.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
```diff
@@ -307,6 +315,8 @@ class MainActivity :
 
             when (val appBarStatus = (f as? BaseFragment)?.activityAppBarStatus ?: AppBarStatus.Visible()) {
                 is AppBarStatus.Visible -> {
+                    httpsConfigurationWarningAllowedForDestination =
+                        f is TopLevelFragment && shouldShowBottomNavigation
```
irfano: Orders, Products and Menu return `AppBarStatus.Hidden` because they have their own toolbar. This check only allows the banner when the status is `Visible`, so the banner only shows on My Store. On my device it never showed on the other three tabs. Allowing it for every `TopLevelFragment` with bottom navigation, no matter what the app bar status is, fixes this.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/81444b3f8bc

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/main/MainActivity.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/f53dce4657d

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooNoticeBanner.kt
```diff
@@ -69,6 +91,29 @@ fun WooNoticeBanner(
...
+                    ) {
+                        Text(
+                            text = actionLabel,
+                            style = WooTheme.text.labelLarge.emphasized,
+                        )
+                    }
+                }
+            }
+            if (dismissContentDescription != null && onDismissClick != null) {
+                IconButton(
```
irfano, line 114: `TextButton` adds 12dp padding at the start, and `IconButton` uses a 48dp box. Because of this, "Learn more" starts 12dp to the right of the text, and the X is 12dp lower than the title. I saw this on a device.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooNoticeBanner.kt
```diff
@@ -33,7 +38,21 @@ fun WooNoticeBanner(
     description: String? = null,
     tone: WooNoticeBannerTone = WooNoticeBannerTone.Neutral,
     leadingIcon: (@Composable () -> Unit)? = null,
+    actionLabel: String? = null,
+    onActionClick: (() -> Unit)? = null,
+    dismissContentDescription: String? = null,
+    onDismissClick: (() -> Unit)? = null,
```
irfano, line 47: Action and dismiss support is new for `WooNoticeBanner`. The catalog row in `docs/design-system/component-catalog.md` still calls it a static banner, and the catalog demo `WooNoticeBannerDemo` does not show the new variant.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooNoticeBanner.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/d7a8f516859

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooNoticeBanner.kt
irfano: Fixed in https://github.com/woocommerce/woocommerce-android/commit/58f5b9e0864

irfano (approved):
Approving. Jorge is AFK, so I pushed the fixes for my own comments directly to the branch and tested them on an emulator:
- https://github.com/woocommerce/woocommerce-android/commit/89aa7a44371 moves the release note under 25.7
- https://github.com/woocommerce/woocommerce-android/commit/81444b3f8bc points Learn more to the WooCommerce SSL guide, same link as iOS
- https://github.com/woocommerce/woocommerce-android/commit/f53dce4657d shows the warning on Orders, Products and Menu too, not only on My Store
- https://github.com/woocommerce/woocommerce-android/commit/d7a8f516859 aligns Learn more and the X button with the banner text
- https://github.com/woocommerce/woocommerce-android/commit/58f5b9e0864 shows the action and dismiss variant in the design system catalog

All threads are resolved.

PR #16472, opened 2026-08-25
[WOOMOB-3902] Restore the POS refund selection by unit count, not by row id

samiuelson:
### Description

Fixes WOOMOB-3902

When the store rejects a refund preview or create because the order changed, the refund screen offers to reload the remaining items and tries to keep what the cashier had selected. It restored that selection by matching `WooPosRefundableItem.uniqueId`, which is `"${orderItemId}_$rowIndex"`. `rowIndex` is the unit's position within what is still refundable, so it is renumbered on every reload — and a reload only happens because the order changed, which is exactly when those positions move.

Item 42 with quantity 3 gives `42_0`, `42_1`, `42_2`. The cashier deselects `42_0` and keeps the two trailing units. Another register refunds two units. The reload yields only `42_0`, nothing matches, and the selection is emptied. The cashier has to pick again with no explanation.

The failing case is always a trailing one. The selection has to exceed what is left for the store to reject it at all, so a leading selection survives the old id intersection unchanged.

This replaces the id set with `WooPosRefundSelectionSnapshot`: a unit count per `orderItemId` plus the selected lump-sum ids. `WooPosBuildRefundContent` now selects the first *n* units of each line, capped at what is left, and selects a fee row if its id was selected. Products and fees are kept in separate fields because they restore by different rules: a count for units, whole-line presence for a lump sum.

Restoring by count rather than giving units a stable identity is deliberate: nothing downstream needs per-unit identity. `WooPosBuildRefundLineItems` (server path), `WooPosGroupRefundItems` (local path) and `WooPosCalculateRefundSubtotal` all group by `orderItemId` and use the row count, so the count is the only meaningful state.

Two notes for review:

- WOOMOB-3902 describes the no-survivor fallback as "select everything". That is no longer the code — it was changed to leave the selection empty in #16440. The live symptom is therefore a lost selection, not a widened one, and this PR keeps the empty fallback.
- Of the three `WooPosRefundViewModelTest` cases that cover the reload, only the trailing-unit one failed on `trunk` as originally written. The other two deselected the last unit, so the survivors matched the old intersection and they passed. Both have been changed to deselect the *first* unit. All three now fail on `trunk`, along with the new post-rejection case: pasting the four into `trunk`'s `WooPosRefundViewModelTest` gives 80 tests and 4 failures, the trailing-unit ones with `expected:<[["1_0"]]> but was:<[[]]>`.

Pre-existing and unchanged by this PR: if the reload itself fails, the error screen's **Retry** loads with no preserved selection, which selects every remaining unit. Worth knowing if you hit a load error while testing. Tracked as WOOMOB-3927.

### Test Steps

Needs two clients signed in to the same store. The second can be a phone running the main app, as long as it is already signed in — the store needs a WordPress.com login, so a fresh device will not do. Refunding from the main app's order detail screen is enough.

These steps assume the server-calculated refund path, so the store rejects at the preview when Continue is tapped.

Only scenario 1 fails on `trunk`. Scenarios 2 and 3 pass there too — they guard behaviour this PR must not break, so check them against the stated counts rather than reading a pass as proof of the fix.

**Scenario 1 — a shrinking line. Fails on `trunk`.**

1. On the register, open an order with a line item of quantity 3 and start a refund.
2. Deselect the **first** unit of that line. The counter reads `(2 SELECTED)`.
3. On the second client, refund **two** units of that line item. One unit is left.
4. On the register, tap **Continue**. Selected (2) exceeds what is left (1), so the store rejects with "The selected quantity is more than what's left to refund."
5. Tap **Review remaining items**.
6. Expected: one unit is listed and it is selected — `(1 SELECTED)`, Continue enabled.
   On `trunk`: `(0 SELECTED)`, Continue disabled.

**Scenario 2 — a fee alongside the units. Regression guard.**

Use a fresh order with a line item of quantity 3 plus a fee. In POS, "Custom amount" adds one.

7. Start a refund. Deselect the **first** unit, leaving the two trailing units and the fee. The counter reads `(3 SELECTED)`.
8. On the second client, refund **two** units of that line item. Leave the fee alone.
9. On the register, tap **Continue**, then **Review remaining items**.
10. Expected: two rows are listed and **both** are selected — `(2 SELECTED)`: the remaining unit and the fee.
    On `trunk`: `(1 SELECTED)`. The fee survives there because a lump sum's `uniqueId` is `fee_<id>` and carries no `rowIndex`, so only the unit is lost. Assert the count, not just the fee.

**Scenario 3 — nothing survives. Regression guard for #16440.**

Use another fresh order with a line item of quantity 3 plus a fee. Scenario 2's order will not do: it has one unit left, so step 12 cannot refund three.

11. Start a refund. Deselect the **fee**, leaving all three units. The counter reads `(3 SELECTED)`.
12. On the second client, refund **all three** units of that line item.
13. On the register, tap **Continue**, then **Review remaining items**.
14. Expected: only the fee row is listed, and it is **unselected** — `(0 SELECTED)`, Continue disabled. The cashier picks again rather than confirming an item they never chose.
    `trunk` behaves the same: this is #16440's empty fallback, which this PR keeps.

### Images/gif

N/A

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 - [*] The app name on the launcher is no longer translated - Brazilian Portuguese showed "Eba!" instead of "Woo"[https://github.com/woocommerce/woocommerce-android/pull/16452]
 - [*] Order creation: the discount screen now formats the product line and totals like the rest of the app, and "Price after discount" shows the full price until a discount is entered [https://github.com/woocommerce/woocommerce-android/pull/16453]
 - [*] Woo POS: removed the banner warning that Point of Sale would require WooCommerce 10.5.0, as that requirement is not enforced [https://github.com/woocommerce/woocommerce-android/pull/16466]
+- [*] Woo POS: reloading the remaining items after the store rejects a refund now keeps the number of units you had selected, instead of clearing the selection when the line got shorter
 
 25.5
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,7 +7,7 @@
 - [*] The app name on the launcher is no longer translated - Brazilian Portuguese showed "Eba!" instead of "Woo"[https://github.com/woocommerce/woocommerce-android/pull/16452]
 - [*] Order creation: the discount screen now formats the product line and totals like the rest of the app, and "Price after discount" shows the full price until a discount is entered [https://github.com/woocommerce/woocommerce-android/pull/16453]
 - [*] Woo POS: removed the banner warning that Point of Sale would require WooCommerce 10.5.0, as that requirement is not enforced [https://github.com/woocommerce/woocommerce-android/pull/16466]
-- [*] Woo POS: reloading the remaining items after the store rejects a refund now keeps the number of units you had selected, instead of clearing the selection when the line got shorter
+- [*] Woo POS: reloading the remaining items after the store rejects a refund now keeps the number of units you had selected, instead of clearing the selection when the line got shorter [https://github.com/woocommerce/woocommerce-android/pull/16472]
 
 25.5
 -----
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #16473, opened 2026-08-26
Hide the unread reviews badge when the app cannot track review read status

kidinov:
Fixes WOOMOB-3918

The unread reviews count on the Menu tab and on the Reviews row never goes down. Merchant opens every review one by one, count stays the same, and it only clears after killing and reopening the app. `UnseenReviewsCountHandler` counts unread review rows in the WPCom notifications table without checking `SupportsReviewsReadStatus`, but everything that could mark those rows as read is gated on it. That check returns false on a Jetpack site once Woo-driven push has taken over, and also on any site that is not Jetpack at all, so application password stores are hit too. Push notifications keep writing the rows in both cases, so the app was showing a count it could never clear. Now the badge is gated on the same check, same as the unread filter and the unread bars in the list already are.

- Known gap: the check is read once per selected site, like every other caller. If Woo push registers while the app is in foreground the badge can stay until the next foreground. Making it reactive means making `SupportsReviewsReadStatus` reactive everywhere, which is a bigger change.

### Before / after

Reproduced on a real tablet, on a Jetpack store with Woo-driven push registered, which is the reported setup:

```
SITE_ID 256978663 | ORIGIN 1 (WPCOM_REST) | IS_JETPACK_CONNECTED 1  -> connectionType = Jetpack
push_token_256978663 present                                        -> isWooRegistered = true
```

Three unread review rows in the local DB, three pending reviews on the store. Before is the build without the fix, after is the same device and the same three rows, still unread.

| Before | After |
|---|---|
| Reviews row and Menu tab both show 3 | No badge on the Reviews row or the Menu tab |

Before, after opening all three reviews one by one and going back, the count is still 3:

Count still 3 after opening every review

The application password case was checked separately on an emulator and behaves the same way.

### Test steps

You need a store where the app can't track read status, so either a Jetpack store with Woo-driven push, or a self hosted one with application password. It also needs at least one review notification already in the local DB.

1. Open the app, go to the Menu tab.
2. Before: Reviews row and the Menu tab both show unread count. Open Reviews, open each review, go back, count is still there.
3. After: no badge on the Reviews row and no badge on the Menu tab. Orders badge still shows, so badges in general still work.
4. On a Jetpack store still on WPCom push the badge should appear as before, and still go down when you open a review.


WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/UnseenReviewsCountHandler.kt
```diff
@@ -21,19 +23,23 @@ import javax.inject.Singleton
...
     @OptIn(ExperimentalCoroutinesApi::class)
     private val unseenReviewsCount: SharedFlow<Int> =
         selectedSite.observe()
             .filterNotNull()
             .flatMapLatest { site ->
-                wpComPushNotificationStore.observeNotificationsForSite(
-                    site = site,
-                    filterBySubtype = listOf(STORE_REVIEW.toString())
-                )
+                if (supportsReviewsReadStatus()) {
```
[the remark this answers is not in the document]
kidinov: @claude rejected: known and called out in the PR description. The window is one foreground session - Woo registration writes the token to DataStore, so the next launch reads the gate correctly. Making it reactive means exposing a Flow from SupportsReviewsReadStatus and subscribing to the plugin-support check for the whole activity lifetime, and the unread filter in ReviewListViewModel would need the same treatment. That belongs in its own change, not in this fix.

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/UnseenReviewsCountHandler.kt
```diff
@@ -21,19 +23,23 @@ import javax.inject.Singleton
...
+                    wpComPushNotificationStore.observeNotificationsForSite(
+                        site = site,
+                        filterBySubtype = listOf(STORE_REVIEW.toString())
+                    ).map { notifications -> notifications.count { !it.read } }
+                } else {
+                    flowOf(0)
```
malinajirka: ❓ On these stores, if the merchant has the Reviews list open and a new review comes in, the list won't refresh on its own any more — they'd have to pull to refresh. `ReviewListViewModel:236` re-fetches the list whenever this count changes, and now the count is always 0.

Intentional? Wdyt?

malinajirka (approved):

The code works as you intended, however, I'm honestly not sure this is the behavior users expect or want. Being notified about new product reviews feels like one of the important benefits of using the mobile app. The push notification still arrives, but there is nothing in the app itself which would tell the merchant something new is waiting. It seems like we are hiding it just because the server doesn't support telling us what has been read already. I understand the badge was wrong on these stores anyway, so this PR is definitely an improvement.

I think we should fix the root cause on the server side. If it's not possible for some reason, I'm wondering if we could somehow partially simulate the read/unread status. Perhaps, we could store last time reviews section was opened on the device and show the badge for all reviews with a newer creation date. Of course, it wouldn't work across devices, but that still feels like a more user-friendly behavior, than not showing the badges at all.

I don't want to block this PR, it's clearly better than a count which can never go down. But I think it'd be worth creating a follow up issue. Wdyt?

WooCommerce/src/main/kotlin/com/woocommerce/android/notifications/UnseenReviewsCountHandler.kt
kidinov: Yep, good catch - ad83e52

kidinov:

kidinov:
@malinajirka thanks for review.

> I think we should fix the root cause on the server side. If it's not possible for some reason, I'm wondering if we could somehow partially simulate the read/unread status. Perhaps, we could store last time reviews section was opened on the device and show the badge for all reviews with a newer creation date. Of course, it wouldn't work across devices, but that still feels like a more user-friendly behavior, than not showing the badges at all.

> I don't want to block this PR, it's clearly better than a count which can never go down. But I think it'd be worth creating a follow up issue. Wdyt?

Agree!  WOOMOB-3978

kidinov:

PR #16475, opened 2026-08-26
[WOOMOB-999] Make WCDataStore location functions suspendable

irfano:
Fixes WOOMOB-999

### Description

`WCDataStore.getCountries` and `getStates` wrapped their Room reads in `runBlocking`. Both are read from ViewModels on the main thread, so editing an order address, picking a customer during order creation and opening a customer's details each blocked the main thread on a location read.

This PR makes both suspend and removes the `runBlocking`. `GetLocations`, `OrderMapper.mapAddress`, `GetCustomerWithStats`, `CustomerListRepository.getState` and the customer selection chain became suspend as a result. `AddressViewModel` now resolves countries and states inside coroutines, which affects country and state selection and the two location selectors.

The legacy shipping labels screens keep blocking access through `getCountriesBlocking`, `getStatesBlocking` and `GetLocations.getBlocking`. Those screens read locations from property getters and adapter constructors, so converting them would mean restructuring a flow that WooCommerce Shipping is replacing. The three wrappers are documented as deletable with that flow. Behavior there is unchanged.

There is one commit per step.

#### 1. Order address country and state

1. Open an order from the **Orders** tab.
2. Tap **Edit** and open the billing address.
3. Tap the country field and pick a country that has states, for example United States.
4. Confirm the state field turns into a spinner and any previous state is cleared.
5. Tap the state field and pick a state.
6. Confirm the state field shows the state name.
7. Tap the country field again and pick a country with no states, for example Anguilla.
8. Confirm the state field turns into a free text field.
9. Tap **Done**.
10. Confirm the billing details on the order show the new country.

#### 2. Customer picker

Needs a registered customer with a billing address in a country that has states.

1. Tap the add order button on the **Orders** tab.
2. Tap **Add customer details**.
3. Pick the customer.
4. Confirm the billing and shipping addresses are filled in.
5. Tap the edit icon on the customer card.
6. Confirm the country and state fields show names and not codes.

#### 3. Customer details

1. Open the **More** menu and tap **Customers**.
2. Tap a customer that has an address.
3. Confirm the **Location** section shows the country and region names and not codes.

### Images/gif

N/A, no UI change.

- [x] I have considered if this change warrants release notes. This is an internal refactor with no behavior change, so none were added.

### DIFF
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/model/CustomerWithAnalytics.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/model/CustomerWithAnalytics.kt
@@ -30,7 +30,7 @@ data class CustomerWithAnalytics(
     fun getBillingAddress(): String = billingAddress.getFullAddress()
 }
 
-fun WCCustomerModel.toCustomerWithAnalytics(
+suspend fun WCCustomerModel.toCustomerWithAnalytics(
     repository: CustomerListRepository,
     mapper: CustomerListViewModelMapper,
 ): CustomerWithAnalytics {
--- a/WooCommerce/src/main/kotlin/com/woocommerce/android/model/GetLocations.kt
+++ b/WooCommerce/src/main/kotlin/com/woocommerce/android/model/GetLocations.kt
@@ -1,11 +1,12 @@
 package com.woocommerce.android.model
 
 import com.woocommerce.android.ui.orders.details.editing.address.LocationCode
+import kotlinx.coroutines.runBlocking
 import org.wordpress.android.fluxc.store.WCDataStore
 import javax.inject.Inject
 
 class GetLocations @Inject constructor(private val locationStore: WCDataStore) {
-    operator fun invoke(countryCode: LocationCode, stateCode: LocationCode): Pair<Location, AmbiguousLocation> {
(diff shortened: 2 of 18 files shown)


AdamGrzybkowski (approved):
Looks good! There's one thing worth mentioning, but this is not necesarirly a new bug - could be there before those changes. 

When going through these testing steps:

1. Open an order from the Orders tab.
2. Tap Edit and open the billing address.
3. Tap the country field and pick a country that has states, for example United States.
4. Confirm the state field turns into a spinner and any previous state is cleared.
5. Tap the state field and pick a state.
6. Confirm the state field shows the state name.
7. Tap the country field again and pick a country with no states, for example Anguilla.
8. Confirm the state field turns into a free text field.
9. Tap Done.
10. Confirm the billing details on the order show the new country.


And switching step 7 to: Tap the country field again and pick another country with states.

The app will keep the state from the previous country, making the form invalid - it would be best to clear the state in such scenario.

irfano:
Thanks for catching this. Checked it, and it's not from this PR, it's on trunk too.

The view model does clear the region, but the region spinner keeps its old text because we only set that text when there's a defined region. So the saved address ends up with no region while the field still shows the old one. Slightly worse than it looks, the form and the saved data disagree.

Filed it separately: WOOMOB-4007

PR #16476, opened 2026-08-26
Show attribute labels on order detail product rows

kidinov:
Order detail showed line item attributes as a comma joined list of values, without the labels. With Product Add-Ons this becomes unreadable: several text add-ons run together into one line and you can't tell which value belongs to which add-on. Now every attribute is rendered as `Key: Value` on its own line, and quantity/price moved to a separate line. Fixes WOOMOB-3247.

| Before | After |
|---|---|
| before | after |

Few things that are not visible in the diff:

- `attributesDescription` is kept and its output is byte for byte the same, so shipping label packaging, customs, order creation and POS are not touched. Only order detail uses the new `displayableAttributes`.
- The row layout is shared though, so the new look also lands on the grouped/bundle rows and on the product list inside the shipping labels card of order detail. Flat rows and grouped rows are both checked on a device. The shipping labels card I could only check by reading the code, it needs a really purchased label.
- SKU line was ellipsized before, now it wraps. Short SKU still stays on one line, and a row without attributes has no empty gap:

  short and long SKU in one order

- We label every attribute, not only add-ons. iOS labels add-ons only and tells them apart by the `_pao_ids` meta, but that one needs Product Add-Ons 5.0.2+. Labelling everything covers older plugin versions too.

### Test steps

1. Open an order with a variation product. The product row should show `Color: Red` on its own line, then `1 x $10.00`, then SKU.
2. To check the add-ons case, in wp-admin create an order in `Pending payment`, add a product, click the pencil on the line item and use `Add meta` twice with names like `Gift message` and `Second local text add-on` and some long values. Save, set status to Processing, Create.
3. Open this order in the app. Every meta is on its own line with its label, long values wrap and stay readable.
4. Open an order with a product that has no attributes. No empty gap, just name, quantity/price and SKU.
5. Product with a long SKU shows the full SKU now, wrapped to a second line.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 25.6
 -----
+- [*] Order detail: product line items now show each attribute with its label on its own line, so Product Add-Ons text values are readable instead of running together [https://github.com/woocommerce/woocommerce-android/pull/16476]
 - [Internal] Woo POS refunds: refund events now report whether totals were calculated locally or by the server, a new event records when a store falls back for lack of the server route, and refunds abandoned before submission are now reported instead of leaving the funnel open [https://github.com/woocommerce/woocommerce-android/pull/16419]
 - [*] The app name on the launcher is no longer translated - Brazilian Portuguese showed "Eba!" instead of "Woo"[https://github.com/woocommerce/woocommerce-android/pull/16452]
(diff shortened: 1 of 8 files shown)


samiuelson (approved):
LGTM!

kidinov:

PR #16480, opened 2026-08-27
Follow-up ProGuard rule cleanups from R8 PR review

malinajirka:
WOOMOB-3903

Follow-up to the review on #16308 — going through the ProGuard cleanup suggestions one by one. 

Original review comments:

- **OkHttp** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841471573)) — removed. OkHttp ships its own rules and `com.squareup.okhttp.**` is dead v2.
- **Event Bus 2** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841506814), [comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841501420)) — gone. We only use Event Bus 3, there's no `de.greenrobot.event` anywhere.
- **FluxC** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841469204)) — moved into `libs/fluxc/consumer-rules.pro` where it belongs. Same rules, so the merged config doesn't change.
- **Dagger errorprone** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841527435)) — removed, library rules already cover it.
- **Crypto Tink** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841544372)) — kept. Turns out we still pull it in transitively and R8 fails without these 9 lines (`KeysDownloader` references classes that aren't on the classpath). Added a comment explaining why.
- **AGP auto-generated** ([comment](https://github.com/woocommerce/woocommerce-android/pull/16308#discussion_r3841552686)) — removed, no longer needed.

While in here I also cleaned up a few more rules that turned out to be redundant or dead:

- **`-keepattributes Signature` / `*Annotation*`** — removed. Gson's own `gson.pro` ships `Signature` and EventBus/coordinatorlayout/FluxC ship `*Annotation*`; `-keepattributes` is global, so the merged config keeps both without ours.
- **Event Bus 3 keeps** (`ThreadMode`, `ThrowableFailureEvent` ctor) — removed. EventBus's own consumer rules already ship both, and the `* extends ThrowableFailureEvent` variant only matters with AsyncExecutor, which we don't use.
- **guava `-dontwarn com.google.common.**`** — removed, R8 builds clean without it.
- **Glide `GeneratedAppGlideModuleImpl` keep** — removed. The generated class is never actually in the release build (not in R8 seeds, on trunk either), so the rule matched nothing.

The file goes from 85 lines to ~24.

### Test Steps

Build-config only, nothing user-facing. `assembleWasabiRelease` and `lintWasabiRelease` are green, and I smoke-tested the logged-in release build — dashboard/stats, orders + detail, products + detail all load real data with no crashes or stripped-class errors.

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes. — N/A, internal.

### DIFF
--- a/WooCommerce/proguard-rules.pro
+++ b/WooCommerce/proguard-rules.pro
@@ -14,17 +14,6 @@
 }
 ###### Event Bus 3 - end
 
-###### Event Bus 2 - begin
--keepclassmembers class ** {
-    public void onEvent*(**);
-}
-
-# Only required if you use AsyncExecutor
--keepclassmembers class * extends de.greenrobot.event.util.ThrowableFailureEvent {
-    ** *(java.lang.Throwable);
-}
-###### Event Bus 2 - end
-
 ##### WooCommerce - begin
 # Gson populates fields reflectively, so keep fields and enum constants of our classes
 # (we don't obfuscate, so names are stable). Methods stay eligible for R8 optimization.
--- a/WooCommerce/proguard-rules.pro
+++ b/WooCommerce/proguard-rules.pro
@@ -11,8 +11,6 @@
 -keep class com.wellsql** { *; }
 ###### FluxC - end
 
-###### Zendesk (the SDK ships its own consumer rules; Gson/Retrofit/OkHttp ship theirs too)
-
 ###### SavedStateHandleExt - begin
 ###### We use reflection so we have to keep this method
 -keepclassmembers class * extends androidx.navigation.NavArgs {
@@ -20,8 +18,8 @@
 }
 ###### SavedStateHandleExt - end
 
-###### Google Crypto Tink (still transitively present; KeysDownloader references
-###### google-http-client + joda-time which aren't on the classpath) - begin
+# Crypto Tink is still transitively present; its KeysDownloader references google-http-client
+# and joda-time, which aren't on the classpath. R8 fails to build without these.
 -dontwarn com.google.api.client.http.GenericUrl
 -dontwarn com.google.api.client.http.HttpHeaders
 -dontwarn com.google.api.client.http.HttpRequest
@@ -31,4 +29,3 @@
 -dontwarn com.google.api.client.http.javanet.NetHttpTransport$Builder
 -dontwarn com.google.api.client.http.javanet.NetHttpTransport
 -dontwarn org.joda.time.Instant
-###### Google Crypto Tink - end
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/proguard-rules.pro
```diff
@@ -1,71 +1,25 @@
...
 ###### SavedStateHandleExt - begin
 ###### We use reflection so we have to keep this method
 -keepclassmembers class * extends androidx.navigation.NavArgs {
     fromSavedStateHandle(androidx.lifecycle.SavedStateHandle);
 }
 ###### SavedStateHandleExt - end
 
-###### Google Crypto Tink dependencies - begin
+# Crypto Tink is still transitively present; its KeysDownloader references google-http-client
+# and joda-time, which aren't on the classpath. R8 fails to build without these.
```
wzieba, line 22: omg 🤦 thanks for checking. Do you remember what's the origin / which dependency introduces it?

wzieba (approved):
Looks great, thank you @malinajirka !

WooCommerce/proguard-rules.pro
malinajirka: androidx.security:security-crypto:1.1.0 via two paths:

- FluxC → security-crypto (direct, for EncryptedSharedPreferences in ApplicationPasswordsStore)
- CardReader → Stripe Terminal → security-crypto


I considered moving this rule to card reader module and fluxc, but duplication here doesn't seem it's worth it. I don't understand why security-crypto doesn't have this rule in their proguard file, but I guess it is what it is.

wzieba: > I don't understand why security-crypto doesn't have this rule in their proguard file

Yeah, same 😕

PR #16483, opened 2026-08-27
[WOOMOB-3734] Fix SelectedSiteResetException and ClassCastException on application-password logout

AdamGrzybkowski:
### Description

Fixes WOOMOB-3734

Sentry: [WOOCOMMERCE-ANDROID-VB4](https://a8c.sentry.io/issues/WOOCOMMERCE-ANDROID-VB4) (the `SelectedSiteResetException`) and [WOOCOMMERCE-ANDROID-WDM](https://a8c.sentry.io/issues/WOOCOMMERCE-ANDROID-WDM) (the `ClassCastException`).

`AccountRepository.logout()` read `selectedSite.get()` after awaiting push-notification cleanup. That cleanup is a network round-trip, and `SelectedSite.reset()` sets a sticky flag, so anything that resets the site during that window — usually a second, concurrent logout — makes `get()` throw `SelectedSiteResetException` on resume. It became likely in 24.7, when #15667 hoisted push cleanup out of the two logout branches so it runs before them. The fix captures the site with the non-throwing `getOrNull()` before the suspension point.

The second commit fixes [WOOCOMMERCE-ANDROID-WDM](https://a8c.sentry.io/issues/WOOCOMMERCE-ANDROID-WDM), the fatal `ClassCastException` linked from the same Linear issue. `ApplicationPasswordsFetchResponse.uuid` was declared non-null, but Gson leaves it null when the server omits the field, so a `WPAPIResponse.Success` could carry a null payload and the old `if/else` cast that `Success` to `Error`. Making the field nullable also surfaced an NPE on the cookie-auth path. Both are guarded and covered by tests.

### Test Steps

**Sanity check — neither fix broke normal logout (no setup)**

This exercises the happy paths the two fixes touch, not the bugs themselves; both bug repros need a harness and are below.

1. Log in with store credentials: **Log In** → **No computer? Log in with site address** → enter a store address → **Log in with your site credentials**.
2. **Menu** → **Settings** → **Log out** → confirm.
3. Logout completes, the app lands on the login screen, and the app does not restart. `adb logcat | grep WooCommerce-LOGIN` shows `Application password deleted`.

For the first fix that confirms the captured site is still the one whose password gets deleted and that `cleanup()` runs. For the second it confirms the `credentials.uuid != null` short-circuit still skips the introspect call — the crash lives behind that call, so a normal logout never reaches it.

**WOOMOB-3734 — the race itself**

The window is one network round-trip wide and needs a second concurrent logout, so it isn't reproducible by hand. The attached patch stands in for a slow push cleanup and fires a second, unguarded logout; the first caller stays suspended long enough for the second to finish and reset the selected site. It touches only `PushNotificationRepository` and `AppSettingsPresenter`, so it applies to this branch **and** to `trunk`.

[woomob-3734-repro.patch](https://github.com/user-attachments/files/31788111/woomob-3734-repro.patch)

1. `git apply woomob-3734-repro.patch`, then `./gradlew :WooCommerce:installWasabiDebug`.
2. Log in with store credentials, then **Settings** → **Log out**.
3. Watch `adb logcat -v time | grep -aE "REPRO|SelectedSiteResetException|System.exit"`.
4. On this branch: `push cleanup started (call #1)` / `(call #2)`, then `second logout completed without throwing` — no exception, no exit, and two `account_logout` events, i.e. both logouts ran `cleanup()`.
5. On `trunk` (same patch): `SelectedSiteResetException` at `SelectedSite.kt:89` ← `AccountRepository.logoutApplicationPasswordAccount(AccountRepository.kt:130)`, followed by `System.exit called, status: 0` — the app disappears and restarts.

**WOOCOMMERCE-ANDROID-WDM**

A real server won't return a password list without a `uuid`, so this one is driven with the app's built-in ApiFaker (debug builds only, so use `installWasabiDebug`). Two mocks are needed: password creation returns `409`, which sends the app into `deleteConflictedApplicationPassword()`, and the password list returns `200` with a matching name but no `uuid`.

**Setup**

The faked entry's `name` has to match `WooApplicationPasswordsConfiguration.applicationName`, which is device-specific, so derive it first:

```bash
PKG=com.woocommerce.android.dev
RECEIVER="$PKG/com.woocommerce.android.apifaker.adb.ApiFakerBroadcastReceiver"

MAN=$(adb shell getprop ro.product.manufacturer | tr -d '\r')
MOD=$(adb shell getprop ro.product.model | tr -d '\r')
case "$MOD" in "$MAN"*) DEV="$MOD";; *) DEV="$MAN $MOD";; esac
APPNAME="$PKG.app-client.$(echo "$DEV" | tr ' /\\' '---')"
echo "$APPNAME"   # e.g. com.woocommerce.android.dev.app-client.Google-sdk_gphone16k_arm64
```

Then arm ApiFaker:

```bash
adb shell am broadcast -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS -n "$RECEIVER"
adb shell am broadcast -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled true -n "$RECEIVER"

# 1) application password creation -> 409 CONFLICT
adb shell am broadcast -a com.woocommerce.android.apifaker.ADD_ENDPOINT \
  --es api_type wp-api \
  --es path /wp/v2/users/me/application-passwords \
  --es http_method POST \
  --ei response_status_code 409 \
  --es response_body "'{\"code\":\"application_password_exists\"}'" \
  -n "$RECEIVER"

# 2) application password list -> 200, matching name, no uuid
adb shell am broadcast -a com.woocommerce.android.apifaker.ADD_ENDPOINT \
  --es api_type wp-api \
  --es path /wp/v2/users/me/application-passwords \
  --es http_method GET \
  --es response_body "'[{\"name\":\"$APPNAME\"}]'" \
  -n "$RECEIVER"
```

A red "ApiFaker Enabled" banner appears at the bottom of the app once it's on.

**Steps**

1. Log in with site credentials (site address → **Log in with your site credentials** → username/password).
2. Watch `adb logcat -v time | grep -aE "WCApiFaker|Application Password already exists|FATAL EXCEPTION"`.
3. On this branch: two `WCApiFaker: Matched request` lines, `Application Password already exists`, then login fails gracefully with an application-password error and the app stays up.
4. On `trunk` with the same mocks: same two lines, then `FATAL EXCEPTION: main` — `NullPointerException: Parameter specified as non-null is null` in `ApplicationPasswordUUIDFetchPayload.<init>` — and the process is killed.

**Cleanup**

```bash
adb shell am broadcast -a com.woocommerce.android.apifaker.SET_STATUS --ez enabled false -n "$RECEIVER"
adb shell am broadcast -a com.woocommerce.android.apifaker.CLEAR_ENDPOINTS -n "$RECEIVER"
```

### Images/gif

N/A

### DIFF
--- a/libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
+++ b/libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
@@ -1,0 +1,242 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+import com.android.volley.NetworkResponse
+import com.android.volley.Request
+import com.android.volley.RequestQueue
+import com.android.volley.VolleyError
+import com.google.gson.Gson
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.robolectric.RobolectricTestRunner
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.BaseRequest.BaseNetworkError
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.ResponseWithHeaders
+import org.wordpress.android.fluxc.network.rest.wpapi.CookieNonceAuthenticator
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIGsonRequest
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPINetworkError
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+@ExperimentalCoroutinesApi
+@RunWith(RobolectricTestRunner::class)
+class WPApiApplicationPasswordsRestClientTest {
+    private val testSite = SiteModel().apply {
+        url = "http://test-site.com"
+    }
+    private val credentialsWithoutUuid = ApplicationPasswordCredentials(
+        userName = "username",
+        password = "password",
+        uuid = null
+    )
+
+    private val noCookieRequestQueue: RequestQueue = mock()
+    private val cookieNonceAuthenticator: CookieNonceAuthenticator = mock()
+    private val restClient = WPApiApplicationPasswordsRestClient(
+        wpApiGsonRequestBuilder = mock(),
+        cookieNonceAuthenticator = cookieNonceAuthenticator,
+        noCookieRequestQueue = noCookieRequestQueue,
+        requestQueue = mock(),
+        dispatcher = mock<Dispatcher>(),
+        userAgent = mock<UserAgent>()
+    )
+
+    @Test
+    fun `given the introspect response has no uuid, when deleting a password, then return an error payload`() =
+        runTest {
+            // GIVEN the server answers 200 but omits the uuid field, so Gson leaves it null
+            givenSuccessResponse(
+                Gson().fromJson("""{"name":"woo-app"}""", ApplicationPasswordsFetchResponse::class.java)
+            )
+
+            // WHEN
+            val payload = restClient.deleteApplicationPassword(testSite, credentialsWithoutUuid)
+
+            // THEN
+            assertTrue(payload.isError)
+            assertFalse(payload.isDeleted)
+            assertEquals("UUID missing from response", payload.error.message)
+        }
+
+    @Test
+    fun `given the introspect response is empty, when deleting a password, then return an error payload`() =
+        runTest {
+            // GIVEN the server answers 200 with no body at all
+            givenSuccessResponse(null)
+
+            // WHEN
+            val payload = restClient.deleteApplicationPassword(testSite, credentialsWithoutUuid)
+
+            // THEN
+            assertTrue(payload.isError)
+            assertEquals("Response is empty", payload.error.message)
+        }
+
+    @Test
+    fun `given the introspect response has a uuid, when deleting a password, then the password is deleted`() =
+        runTest {
+            val introspectResponse = Gson().fromJson(
+                """{"uuid":"the-uuid","name":"woo-app"}""",
+                ApplicationPasswordsFetchResponse::class.java
+            )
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
```diff
@@ -0,0 +1,242 @@
...
+
+            // THEN only the DELETE is issued
+            assertFalse(payload.isError)
+            assertTrue(payload.isDeleted)
+            verify(noCookieRequestQueue, times(1)).add(any<WPAPIGsonRequest<Any>>())
+        }
+
+    private fun givenSuccessResponse(response: Any?) = givenSuccessResponses(response)
+
+    @Suppress("UNCHECKED_CAST", "SpreadOperator")
```
irfano: There's no spread call in this file, so `SpreadOperator` doesn't do anything here.
```suggestion
    @Suppress("UNCHECKED_CAST")
```

libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
```diff
@@ -0,0 +1,242 @@
...
+            val listedPassword = Gson().fromJson(
+                """{"name":"woo-app"}""",
+                ApplicationPasswordsFetchResponse::class.java
+            )
+            whenever(
+                cookieNonceAuthenticator.makeAuthenticatedWPAPIRequest<Array<ApplicationPasswordsFetchResponse>>(
+                    eq(testSite),
+                    any()
+                )
+            ).thenReturn(WPAPIResponse.Success(arrayOf(listedPassword), emptyList()))
```
irfano: np: This is the same stub as `givenCookieAuthReturns`. Can we use the helper here too?
```suggestion
            givenCookieAuthReturns(WPAPIResponse.Success(arrayOf(listedPassword), emptyList()))
```

irfano (approved):
LGTM! Left two small notes on the new fluxc test, not blocking.

libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
AdamGrzybkowski: Good catch — there's no spread call anywhere in the file, so that suppression was dead. Dropped it.

libs/fluxc/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/WPApiApplicationPasswordsRestClientTest.kt
AdamGrzybkowski: Done — that test predated `givenCookieAuthReturns`, it now uses the helper.

PR #16484, opened 2026-08-28
Fix Glide KSP processor so its generated module is applied

malinajirka:
WOOMOB-3933

Stacked on top of #16480 (R8 cleanup) — base it there, or retarget `trunk` once that merges.

The app wired `ksp(libs.bumptech.glide.compiler)`, but Glide's `compiler` is a kapt/javac annotation processor, not a KSP one — so KSP ran it, emitted nothing, and `GeneratedAppGlideModuleImpl` was never generated. That means our `WooCommerceGlideModule` (which routes Glide through FluxC's custom-SSL Volley `RequestQueue` and tunes the disk cache) was silently ignored, and Glide fell back to its default networking. Fix is to use Glide's actual KSP artifact, `com.github.bumptech.glide:ksp`.

**The warning I observed:** in a `wasabiRelease` build, `adb logcat` shows this at Glide init (reproduces on trunk too, so it predates the R8 work):

```
W Glide: Failed to find GeneratedAppGlideModule. You should include an annotationProcessor compile dependency on com.github.bumptech.glide:compiler in your application and a @GlideModule annotated AppGlideModule implementation or LibraryGlideModules will be silently ignored
```

After this change the warning is gone, and `GeneratedAppGlideModuleImpl` now shows up in R8's `mapping/wasabiRelease/seeds.txt`.

### Test Steps

1. `./gradlew :WooCommerce:assembleWasabiRelease`, install the APK, run `adb logcat | grep Glide`, open **Products** (loads images).
   - Before: the `Failed to find GeneratedAppGlideModule` warning appears.
   - After: it's gone; images still load, no crashes.
2. Optionally confirm `GeneratedAppGlideModuleImpl` is now in `WooCommerce/build/outputs/mapping/wasabiRelease/seeds.txt`.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

JorgeMucientes (approved):
Everything works as expected and didn't see any Glide warnings in the logcat 👍🏼

wzieba:
Great finding @malinajirka , thank you! I wonder

> `WooCommerceGlideModule` (which routes Glide through FluxC's custom-SSL Volley `RequestQueue` and tunes the disk cache) was silently ignored, and Glide fell back to its default networking. 

One question, what does it mean in practice, do you maybe know? Is our cache is more performant or the SSL is required for some users? Asking for the future, because if we could drop the custom Glide module it could be a win.

malinajirka:
> One question, what does it mean in practice, do you maybe know? Is our cache is more performant or the SSL is required for some users? Asking for the future, because if we could drop the custom Glide module it could be a win.

That's a great question! I assumed that it is required for self-hosted sites with self-signed certificate, but I honestly don't know.

> didn't see any Glide warnings in the logcat

Thanks for the review @JorgeMucientes! Have you tried building a release build - it doesn't happen on debug.

JorgeMucientes:
> Thanks for the review @JorgeMucientes! Have you tried building a release build - it doesn't happen on debug.

Yes @malinajirka . Compiled wasabiRelease variant and couldn't spot any Glide log while interacting with product images.

malinajirka:
> Yes @malinajirka . Compiled wasabiRelease variant and couldn't spot any Glide log while interacting with product images.

Ohh, I see - I don't see any such errors either. This is a build warning - produced once during build, not during runtime.

I don't know how to reproduce any issues at runtime - I think it requires some weird SSL setup on the site or something. But there is also a chance, this PR doesn't fix anything real and the custom glide module should be simply deleted.

PR #16485, opened 2026-08-28
Fix R8-stripped Gson classes breaking POS, analytics, subscriptions & shipping labels in release

malinajirka:
### Description

Fixes WOOMOB-3903, WOOMOB-3964

The R8 optimization migration (#16308) narrowed the app keep rule to `-keepclassmembers com.woocommerce.** { <fields>; }`, which keeps fields but not the class itself, so R8 full mode removes or abstracts classes that are only instantiated reflectively by Gson => this caused crashes in POS and silent errors in other features.

At first, I tried fixing this be manually excluding files in `com.woocommerce.android.network.**`, `**.networking.**`. However, this approach is IMO too risky. Some classes living outside these packages work just coincidentally - they are being references, but future refactor my change that.

Considering broadening the rule to `-keep class com.woocommerce.** { <fields>; }` increases the APK size just by 0.8MB and has no runtime performance impact (since methods optimization is still enabled for them), it feels like a much more robust and safe approach.

### Test Steps

Each path below crashes or logs a failure on a `wasabiRelease` build of `trunk` and works after this PR:

Each path below fails on a `wasabiRelease` build of `trunk` and works after this PR (verified on device — OnePlus Pad, Pressable staging store):

| Path | Steps | Trunk symptom (gone after fix) |
|---|---|---|
| POS | Open the POS tab (store with attributed variations) | `ClassCastException at WooPosLocalCatalogSyncWithFts.toFtsEntity` — catalog sync worker fails, POS shows "Unable to sync" and is unusable |
| Gift card analytics | My Store → View all store analytics | `JsonIOException: Abstract classes can't be instantiated … GiftCardsStatsApiResponse`; card silently shows "No gift cards this period" |
| Subscriptions | Open an order that has subscriptions | `JsonIOException … SubscriptionRestClient$SubscriptionDto`; Subscriptions section silently missing from order details |

Reproduced on device (OnePlus Pad, wasabiRelease) for POS, gift cards, and subscriptions — all clean after the fix.

### Images/gif

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosVariationMapper.kt
```diff
@@ -152,6 +153,9 @@ class WooPosVariationMapper @Inject constructor(
         }
     }
 
+    // @Keep: only instantiated reflectively by Gson via TypeToken, so R8 would otherwise
+    // strip the class and Gson would fall back to LinkedTreeMap, causing a ClassCastException.
+    @Keep
```
wzieba: Great finding @malinajirka here too, thanks for working on this. It's impossible to find places like this. I wish we could migrate from Gson or any reflection in the production code.

One suggestion - maybe we could move `AttributeJsonItem` and any other JSON mapping model outside to FluxC models that already have the R8 exclusion rule? I think keeping them all in one place is safer for the long run than keeping the `@Keep` annotation.

wzieba: > impossible

ok maybe it's far stretched, but I meant "difficult" 🙂

kidinov:
@malinajirka  I'd consider doing something to prevent this from happening in the future as well. Maybe:
* On R8 side, keep by package instead of by class e.g:

```
-keep class com.woocommerce.**.networking.** { <fields>; }
```
We'd need to move some classes there. (or maybe move all to fluxc)

* And a new detekt rule -  if `object : TypeToken<...>() {}` or `fromJson(x, Foo::class.java)`
used a type that is not in one of the kept packages, fail. So we enforce everyone to keep those "DTO" files in the specific "kept" packages

And btw, for these two POS classes we probably use another API - which removes both classes and there is nothing left to keep at all

```
            JsonParser.parseString(attributesJson).asJsonArray.map { element ->
                val attribute = element.asJsonObject
                WooPosVariation.WooPosVariationAttribute(
                    id = attribute.longOrNull("id"),
                    name = attribute.stringOrNull("name"),
                    option = attribute.stringOrNull("option"),
                )
            }
```

malinajirka:
Thank you both! 

> I'd consider doing something to prevent this from happening in the future as well. 

Yeah, something like this is exactly what I'm trying to do - I moved the PR to draft since I wanted to wrap it up before finalizing the PR.

I agree, we need some broader rule here. I like the idea of the lint/detekt rule! Will apply that too.

malinajirka:
@kidinov @wzieba  I've reworked this PR - I've updated the description. Each time I went over what else could go wrong, I found another example that is either not working or is working just coincidentally -> I opted for a general rule that excludes all woo classes -> the impact of the general rule is minimal and it's safer and future proof.

kidinov (approved):
LGTM but a couple of things

1.  The rule keeps fields but not the empty constructor. And I think Gson needs that constructor. Without it Gson builds the object in another way that skips all Kotlin default values, so `val items: List<X> = emptyList()` becomes null, and only in release.

Gson ships a rule for this, but it covers only classes that have `@SerializedName` on a field. For isntance The `AttributeJson` you delete here had none, so it was not covered. Fluxc has the same hole, and I think that is why you needed the `JsonObjectOrFalse` rule.

maybe then:
```
-keep class com.woocommerce.** { <fields>; <init>(); }
-keepclassmembers class org.wordpress.android.fluxc.** { <fields>; <init>(); }
```

2. Release note - the crash was in prod right? So this one is not internal but a public one

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosVariationMapper.kt
```diff
@@ -18,9 +21,7 @@ import javax.inject.Singleton
  * This centralizes all variation mapping logic, JSON parsing, and name generation.
  */
 @Singleton
-class WooPosVariationMapper @Inject constructor(
-    private val gson: Gson
-) {
+class WooPosVariationMapper @Inject constructor() {
```
wzieba, line 24: I know it's not in your edit, but this doesn't have to be a singleton (keeps no state, trivial to initialize)

```suggestion
class WooPosVariationMapper @Inject constructor() {
```

or, actually even better, make it `object` so it's not injectable (I doubt anyone wants to mock a static mapper - I see no point)

```suggestion
object WooPosVariationMapper {
```

It's not in scope of this PR, so feel free to ignore it.

WooCommerce/proguard-rules.pro
```diff
@@ -1,9 +1,23 @@
...
+# We considered scoping this to the packages Gson DTOs conventionally live in:
+#   -keep class com.woocommerce.android.network.** { <fields>; }
+#   -keep class com.woocommerce.android.**.networking.** { <fields>; }
+# but several Gson round-trip models live elsewhere (SharedPreferences/DataStore-persisted
+# models, the cardreader remote wire protocol) and only survived because their write paths
+# happen to construct them directly, and nothing would catch a covered DTO with a field
+# typed to a class outside these packages. The broad rule costs ~0.8 MB APK over the
+# per-package keeps, but can't be silently broken by a refactor and spares developers from
+# thinking about proguard when adding a Gson-deserialized class.
+-keep class com.woocommerce.** { <fields>; }
```
wzieba: I understand reasoning and I think it's the right choice. `~0.8MB` is negligible, my only concern is about performance difference. Have you been able to spot the difference in how snappy is the UI with and without this rule? I might be out of loop but I remember that Compose worked _significantly_ better with R8 - maybe it still does all the improvements to the Compose-generated graph and this `keep` rule doesn't affect it at all, but that's a thing I thought about seeing this.

> Gson-deserialized class

Idea: migrate to other parser? I really dislike that it uses reflection and IIRC it's the least performative option nowadays for Android. With LLMs it might be easier and less widely spread than, let's say, WellSql migration. Something to think about.

wzieba (approved):
Looks good! It'd be great if `gson` (or any JSON manipulation) would be limited to FluxC (data layer) only, but it is what it is.

WooCommerce/proguard-rules.pro
malinajirka: The rule keeps method optimizations, so runtime performance should not be impacted at all AFAICT - I mean, we still get the runtime performance benefits of R8. Wdyt?

malinajirka: > Idea: migrate to other parser? I really dislike that it uses reflection and IIRC it's the least performative option nowadays for Android. With LLMs it might be easier and less widely spread than, let's say, WellSql migration. Something to think about.

Agree, it'd be nice to migrate away. It feels a bit more risky compared to wellsql, but likely would be worth it.

WooCommerce/src/main/kotlin/com/woocommerce/android/ui/woopos/common/data/WooPosVariationMapper.kt
malinajirka: I agree, however, the optimization is minimal and I prefer to keep this PR focused on R8. Thanks for the suggestion though!

WooCommerce/proguard-rules.pro
wzieba: > The rule keeps method optimizations, so runtime performance should not be impacted at all AFAICT - I mean, we still get the runtime performance benefits of R8. Wdyt?

Got it, I thought it's all-or-nothing 👍

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/woocommerce/woocommerce-android/milestone/301">25.6</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

malinajirka:
Great catch @kidinov, thanks!

The crash never reached production, we merged R8 after the code-freeze.

PR #16366, opened 2026-08-05
[WOOMOB-3738] Migrate the More menu to the store design system

hichamboushaba:
### Description

Fixes WOOMOB-3738

Migrates the Android Menu tab from the legacy Compose theme and resources to the Store Design System.

The screen retains the existing Android information architecture and ordering: the store header is followed by Settings and General sections, with every action rendered as a full-width Design System cell. The layout remains a single column at all screen sizes, centered within a capped content width on larger displays.

The migration preserves the Fragment shell, Activity-owned bottom navigation, navigation and authenticated web flows, feature gating and loading positions, store switching, review/payment badges, analytics timing, and tab-reselection scroll-to-top behavior. It also makes the shared skeleton use Material 3 theme colors, aligns WooCell with the Design System's 24dp padding contract, strengthens RTL text constraints, adds focused unit coverage, and removes obsolete More color resources.

The status-bar inset color remains owned by the app-wide XML/window theme and is intentionally unchanged here; it will be aligned once the full app design-system migration can update window chrome consistently.

Figma: 50XIH5MmOf4xUYEkM6fAm6-fi-670_13821

### Test Steps

1. Log in to a store and open the Menu tab.
2. Verify the store header, followed by the Settings and General sections with full-width cells in their existing order.
3. Open Settings and return to Menu; also verify WC Admin and View Store still open their authenticated browser destinations.
4. Scroll down, reselect the Menu bottom-tab item, and verify the screen scrolls back to the top.
5. Check a store where optional actions resolve visible or hidden and verify loading placeholders disappear without gaps or reordering.
6. At a narrow width or 2× font size, verify cells remain readable and the store name/plan badge reflow without overlap.
7. In dark mode, RTL, landscape, and a tablet-sized window, verify the single-column content remains readable, correctly ordered, scrollable, and centered.

#### Themes

| Light | Dark |
| --- | --- |
| image | image |

- [x] I have considered if this change warrants release notes and have added them to `RELEASE-NOTES.txt` if necessary. Use the "[Internal]" label for non-user-facing changes.

PR #16487, opened 2026-08-28
[WOOMOB-3739] Add a scoped actions builder and WooOverflowMenu to the design system

hichamboushaba:
### Description

Fixes WOOMOB-3739

`WooTopAppBar` offers trailing actions two ways: a `List<WooTopAppBarAction>` descriptor, and a raw `@Composable RowScope.() -> Unit` escape hatch for anything the descriptor can't express. The escape hatch left spacing to the caller, so screens that reached for it re-implemented a spacing `Row` themselves. Product Detail additionally carried its own outlined ellipsis button and `DropdownMenu`, all of which is chrome the design system should own.

This replaces the raw `RowScope` escape hatch with a scoped builder, `@Composable WooTopAppBarActionsScope.() -> Unit`, exposing `IconAction`, `TextAction` and `OverflowAction`. The scope delegates to `RowScope`, so inline custom composables still work and stay source-ordered, but the component now owns the spacing container and callers no longer set gaps. It also adds `WooOverflowMenu` / `WooOverflowMenuItem` to the design system: a slot-based anchored dropdown that owns its expanded state, anchor and dismissal, which `OverflowAction` wires up to the standard outlined ellipsis trigger.

The descriptor overloads are untouched, so nothing else needs migrating. The scoped `actions` parameter deliberately has no default — a lambda can't match `List<WooTopAppBarAction>`, and with no default on the scoped overload, existing calls that omit `actions` still resolve to the descriptor one. Product Detail is the only consumer in this PR by design: it's the screen with the most caller-side chrome to delete, and it keeps its existing primary → share → overflow order, callbacks, menu anchor and test tags. It sheds ~100 lines of local top app bar code as a result.

A scoped API only helps if callers actually reach for it, so this also adds a detekt rule, `StoreTopAppBarIconButtonUsageRule`. `IconAction` is what gives an action the design system's outlined button, sizing and touch target; nothing previously stopped a caller dropping a raw Material `IconButton` in beside it, and the result looks close enough to survive review. The rule flags `androidx.compose.material` and `material3` `IconButton` calls that resolve inside a `WooTopAppBar` actions lambda — named, trailing or positional — and inside any function or property typed with `WooTopAppBarActionsScope`. Matching is import-aware, so aliased and wildcard imports are caught while a same-named component from another package is not. Overflow content is exempt, since menu items legitimately use `IconButton`: anything nested in `WooOverflowMenu`, a Material `DropdownMenu`, or an `OverflowAction` body is left alone.

The design system tests here cover what the API decides — the `require` checks, `onClick` and `enabled` surviving delegation, arbitrary inline content composing alongside scoped actions, and the overflow trigger opening and dismissing its menu. They deliberately assert no layout geometry. An earlier revision pinned the 40dp outline size, the 48dp touch target and the inter-action gap, but those numbers come from `WooOutlinedIconButton`, Compose's minimum interactive sizing and the row's `Arrangement.spacedBy` respectively, not from this API — so they'd have failed on a spacing token change rather than on a regression here.

Three things a reviewer might reasonably push back on:

- **Why not reuse `WCOverflowMenu`?** It lives in the app module (`ui/compose/component/OverflowMenu.kt`), so the design-system library can't depend on it, and its API takes an item-model list with a hardcoded vertical-dots `IconButton` trigger. `OverflowAction` needs to supply the trigger while the caller supplies the items, which is what the slot-based API gives us. `WCOverflowMenu` is unchanged and its callers are unaffected.
- **Two action APIs coexisting.** That's intentional for now rather than a migration left half-done — it keeps this PR reviewable and lets screens move over individually. `docs/design-system/component-catalog.md` documents when to use which.
- **Why the legacy `Toolbar` isn't a rule target.** `com.woocommerce.android.ui.compose.component.Toolbar` still exposes a plain `RowScope` actions lambda, so its callers have no `IconAction` to reach for and the rule would be unsatisfiable there. It joins the target set once that bridge is migrated.

One file in this diff is unrelated to the API. `WooCellTest.kt` switches its `createComposeRule` import to the `junit4.v2` one. The compose-bom bump in #16357 deprecated the v1 rule and migrated every usage that existed on trunk at the time; #16366 was branched before that bump and added a new v1 usage in that file. Both PRs were green in isolation, and the merge left `:libs:store-design-system:testDebugUnitTest` failing to compile under `-Werror` — which blocks CI for anything branched off `feature/store-design-system-migration`, not just this PR. Happy to split it into its own PR if you'd prefer it out of here.

No user-facing behaviour change: Product Detail renders and behaves as before.

### Test Steps

1. Open any product from the Products tab.
2. Check the top app bar: back chevron, centred product title, an outlined share icon and an outlined ellipsis, with even spacing between the actions.
3. Tap the ellipsis. The menu should anchor directly under it, right-aligned, with "Trash product" in the destructive red and the other items in the default colour.
4. Tap outside the menu — it should dismiss without the tap falling through to the row underneath.
5. Tap a menu item, e.g. "Product settings" — the menu should close and then navigate.
6. Edit the product title. The "Save" text action should appear in the bar; discard the change and confirm the bar returns to its previous state.
7. Repeat step 2 in dark mode and with a right-to-left locale to confirm the action spacing stays symmetric.
8. Run `./gradlew detektAll` — it should pass. Adding a Material `IconButton` inside any `WooTopAppBar` actions lambda should make it fail, while an `IconButton` inside an `OverflowAction` body should not.

### Images/gif

| Actions at rest | Overflow menu open | Primary text action |
| --- | --- | --- |
|  |  |  |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooOverflowMenu.kt
```diff
@@ -0,0 +1,102 @@
...
+            Text(
+                text = text,
+                color = if (isDestructive) {
+                    WooTheme.colors.error
+                } else {
+                    WooTheme.colors.surface.onDefault
+                },
+                style = WooTheme.text.bodyLarge.regular,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
```
AdamGrzybkowski: Previous drop-down menu item would wrap if the text was long - is this an intentional change?

libs/detektrules/src/main/kotlin/com/woocommerce/android/detektrules/store/StoreTopAppBarIconButtonUsageRule.kt
```diff
@@ -0,0 +1,225 @@
...
+                    if (ancestor.isTopAppBarActions(file)) return !isInsideMenuContent
+                    if (ancestor.isMenuContent(file)) isInsideMenuContent = true
+                }
+
+                is KtProperty -> {
+                    if (ancestor.hasTopAppBarActionsScopeType(file)) return !isInsideMenuContent
+                }
+
+                is KtNamedFunction -> {
+                    if (ancestor.hasTopAppBarActionsScopeReceiver(file)) return !isInsideMenuContent
```
AdamGrzybkowski: ⛏️ Imports should be constant per KtFile, so we should be able to check it once and not for each ancestor?

AdamGrzybkowski (approved):
Looks good! Two comments left but none blocking.

libs/store-design-system/src/main/kotlin/com/woocommerce/android/ui/compose/designsystem/component/WooOverflowMenu.kt
hichamboushaba: Thanks for catching it, it was just an oversight. Fixed in 5f09beebb1a

libs/detektrules/src/main/kotlin/com/woocommerce/android/detektrules/store/StoreTopAppBarIconButtonUsageRule.kt
hichamboushaba: Fixed in 0f25f52d1d6

PR #16488, opened 2026-08-29
[WOOMOB-3741] Migrate app chrome to the Store Design System

hichamboushaba:
### Description

Fixes WOOMOB-3741

Migrates the MainActivity shell, XML-owned destination app bars, and the remaining screen-level Compose app bars to the Store Design System, replacing legacy elevation shadows with the design-system divider and aligning app backgrounds and legacy toolbar colors. It routes the legacy Compose `Toolbar` bridge through `WooTopAppBar`, migrates direct Material `TopAppBar` call sites to that bridge, and uses the shared overflow menu so navigation and action controls have consistent chrome while preserving existing navigation, menu, SearchView, and screen ownership. The Settings activity shell and Feature Flags screen use the same design-system toolbars. The shared inset-only app bar now follows the Menu screen's surface while that destination is active and restores its themed surface when leaving, preventing a mismatched strip behind the status bar.

Because the bridge now hands its actions to `WooTopAppBar`, this also widens `StoreTopAppBarIconButtonUsageRule` to cover the legacy `Toolbar`. #16487 introduced that rule guarding `WooTopAppBar` only, deliberately: until this PR the legacy `Toolbar` took a plain `RowScope`, so its callers had no `IconAction` to reach for and the rule would have been unsatisfiable against them. One line in `TARGET_CALLABLES` and five `Toolbar` cases in the rule's tests are all that is left to add here.

Last of the WOOMOB-3741 stack, targeting #16509, which restyles bottom navigation — that is why bottom navigation is unchanged here. The colour token values this change consumes are corrected in #16497 and the scoped action and overflow APIs it builds on come from #16487, both further down.

**The top app bar treatment is not final.** Wagner is working on a more native app bar — dropping the outlined icon buttons, and probably left-aligning the title — so please don't block on how the current chrome looks. The clearest example is order details, where the centred title shares the bar with two outlined chevrons and the Edit action and ends up truncated to `Order…`. That resolves with the component redesign rather than with a fix here, and the same goes for any similar squeeze or spacing oddity you spot while testing.

What is worth reviewing is the routing: XML destinations and screen-level Compose destinations now go through the design-system toolbar components. When the component is finalised, the restyle lands in `WooDesignSystemToolbar` and `WooTopAppBar` — not in the ~80 layouts or individual screens this PR touches, which carry no toolbar styling of their own.

### Test Steps

1. Install the Wasabi debug build on a compact device and launch an authenticated store.
[...]

PR #16496, opened 2026-09-01
[WOOMOB-3741] Theme the bottom navigation with the Store Design System

hichamboushaba:
### Description

Part of WOOMOB-3741

Restyles the bottom navigation bar with the Store Design System. New `woo_ds_bottom_nav_*` tokens and a `Widget.Woo.DesignSystem.BottomNavigationView` style carry the container colour, the rounded active indicator, always-visible labels and 10sp label text, and the app theme points `bottomNavigationStyle` at it; the old `Widget.Woo.BottomNavigationView` and its `button_toggle_fg_selector` go with it. Tab icons move to the design system's paired solid/regular variants, solid on `state_checked`, so the replaced `as_menu_*` selectors, three `ic_menu_*` vectors and 21 `@animator` files are deleted. Badges use `woo_ds_color_accent_red` `#FF383C`, which is what the `tab-item-android` badge variants bind to in Figma.

Three costs worth naming. The `as_menu_*` drawables were `animated-selector`s that drew the icon in when a tab became selected, and a plain state-list has no equivalent, so that animation is gone — the selected tab is still marked by the indicator, the solid icon weight and the label tint. Figma gives the bar neither elevation nor a divider, so both are removed, which in light leaves `#F2EDFF` against a `#F6F7F7` list background and only weak separation; that one wants a designer's eye rather than silent acceptance. And always-visible labels cost per-label width: German renders `Verkaufsort (POS)` as `Verkaufsort (P…` at 360dp. That last one is accepted rather than worked around, because `onNavItemSelected` returns `false` for Point of Sale and launches the POS activity instead, so the tab is never the selected one and under the old auto-visibility that label was not drawn at all on a narrow phone.

`ic_menu_orders_list` was kept alive only by the create-order launcher shortcut, so the shortcut moves to the design-system file glyph and the vector goes with it. Design-system icons ship untinted and a launcher shortcut has no tint hook — `android:tint` is parsed on `vector` and `bitmap` roots but not on `selector`, `layer-list` or `inset` — so `ic_shortcut_create_order` copies the path data and carries its own fill, `woo_ds_color_palette_woo_purple_shade40`. A palette value rather than a semantic role because Launcher3 composites non-adaptive shortcut icons onto white, and a colour that flips under the night qualifier can end up white on white.

Point of Sale keeps its own icon, deliberately: that destination follows the POS theme rather than the design-system icon family. Its drawable is renamed `as_menu_pos` to `ic_menu_pos` — despite the prefix it is a plain vector rather than an `animated-selector`, and with its four namesakes deleted the `as_` prefix no longer matches anything in the tree.

Last of the WOOMOB-3741 stack, targeting #16488: the bar's container colour comes from the tokens in #16497 and the rest of the app chrome is migrated in #16488, so this lands once both are in.

### Test Steps

1. Install the Wasabi debug build and launch an authenticated store.
2. On the main screen, confirm the bar has a rounded indicator behind the selected tab, all five labels visible, and a flat surface with no shadow or divider along its top edge.
3. Move between My Store, Orders, Products and More. Confirm the selected tab's icon renders in the solid weight and the unselected ones in the regular weight.
4. Toggle dark mode and confirm the container colour, icons and labels adapt.
5. On a store with unfulfilled orders, confirm the Orders badge is red and that a three-digit count renders without overlapping the icon.
6. Open an order. Confirm the bar hides on the detail screen and returns on back with the same tab selected.
7. On a 360dp-wide phone with the system language set to German, confirm all five labels are drawn and that only Point of Sale truncates.
8. Set the system font size to its largest setting and confirm the labels stay legible and unclipped.
9. Long-press the app icon on the launcher and confirm the Create order shortcut shows the design-system file glyph, legible against both the light and the dark launcher theme.

### Images/gif

| Light | Dark |
|---|---|
| Bottom navigation, light | Bottom navigation, dark |


WooCommerce/src/main/res/menu/menu_bottom_bar.xml
```diff
@@ -2,26 +2,26 @@
...
     <item
         android:id="@+id/products"
-        android:icon="@drawable/as_menu_products_list"
+        android:icon="@drawable/sel_bottom_nav_products"
         android:title="@string/products" />
 
     <item
         android:id="@+id/point_of_sale"
-        android:icon="@drawable/as_menu_pos"
+        android:icon="@drawable/ic_menu_pos"
```
hichamboushaba, line 20: I renamed the icon because this wasn't an `animated vector`, so the `as_` prefix was misleading.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

hichamboushaba:
Superseded by #16509. This PR was auto-closed as merged when the WOOMOB-3741 stack was reordered — its base branch was pushed containing this branch's head, so GitHub marked it merged into its own base. Nothing was actually merged to trunk. The same branch and commits are up for review on #16509.

